PR #45, opened 2019-01-17
Cleaned up timers on server side

GuanLuo:
After this change, the scope of run timer is more aligned with the one in Run(), which is stopping the timer after `OnCompleteHandleInfer()`.

### DIFF
--- a/src/clients/c++/perf_client.cc
+++ b/src/clients/c++/perf_client.cc
@@ -1037,6 +1037,10 @@ Report(
   const uint64_t compute_time_us = summary.server_compute_time_ns / 1000;
   const uint64_t compute_avg_us = compute_time_us / cnt;
 
+  const uint64_t overhead = (cumm_avg_us > queue_avg_us + compute_avg_us)
+                                ? (cumm_avg_us - queue_avg_us - compute_avg_us)
+                                : 0;
+
   const uint64_t avg_latency_us = summary.client_avg_latency_ns / 1000;
   const uint64_t std_us = summary.std_us;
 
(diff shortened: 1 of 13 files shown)


GuanLuo:
But there is still a mismatch in terms of the compute timer between `AsyncRun()` and `Run()`. Take `NetDefBundle` as example, the compute timers stop after [`NetDefBundle::Run()`](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/servables/caffe2/netdef_bundle.cc#293) finishes. However, if it is asynchronous run, [`OnCompleteQueuedPayloads()`](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/servables/caffe2/netdef_bundle.cc#L312) includes `OnCompleteHandleInfer()` while in the synchronous case, it doesn't.

Depending on the range that we actually want the compute timer to record, we can either:
- Move `OnCompleteHandleInfer()` in [`Run()`](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/core/infer.cc#L975) into the callback function such that compute timers will record `OnCompleteHandleInfer()` in both cases
- Separate  [this](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/servables/caffe2/netdef_bundle.cc#L312) such that we can stop the compute timers before calling the callback function

deadeyegoodwin:
The RunTimer is not reported directly. It is just used to calculate the queue time (see server_status.cc). There is some legacy here that can be cleaned up. I would change RunTimer to QueueTimer and directly compute the queue time. The queue time can start where RunTimer does now (infer.cc:Run/RunAsync) and should be stopped when the payload is scheduled to run, which can be done, for example, inside this loop: https://github.com/NVIDIA/tensorrt-inference-server/blob/462960164e798e193fe052e29c18641ca8d9f622/src/servables/tensorrt/plan_bundle.cc#L413

You should be able to replace queued_timestamp_ in Payload (https://github.com/NVIDIA/tensorrt-inference-server/blob/462960164e798e193fe052e29c18641ca8d9f622/src/core/scheduler.h#L46) with the actual QueuedTimer so you can stop it in the bundles.

Issue #36, opened 2019-01-09
New labels file is not detected


bezero:
I deployed a project which contains model foder (version 1), config file and labels_1.txt for the given version. When new version of the model is introduced, TRTIS automatically detects new version. However, if labels file for the new version is changed, TRTIS does not update "label_filename" unless I restart TRTIS.

To update labels I add labels_2.txt file to the project folder and update "label_filename" in the config file. Is it possible to detect these changes without restarting TRTIS.

deadeyegoodwin:
Are you using the more recent version of TRTIS? Starting in the 18.12/0.9.0 release, TRTIS monitors the configuration file for changes and reloads the model, without requiring a server restart. See https://docs.nvidia.com/deeplearning/sdk/tensorrt-inference-server-guide/docs/model_repository.html#modifying-the-model-repository. If you are up-to-date you should see some console output showing model unload/load after making a configuration change (might be after a delay since TRTIS only checks the model repository at an interval).

bezero:
@deadeyegoodwin 
### Steps I have taken:
TRTIS version: 18.12/0.9.0
model repository:
```protobuf
models/
  mymodel/
  config.pbtxt
  labels_1.txt
  1/
    model.graphdef
```
config.pbtxt file:
```protobuf
name: "mymodel"
  platform: "tensorflow_graphdef"
  max_batch_size: 100
  input [
    {
      name: "input"
      data_type: TYPE_FP32
      format: FORMAT_NHWC
      dims: [ 299, 299, 3 ]
    }
  ]
  output [
    {
      name: "predictions"
      data_type: TYPE_FP32
      dims: [ 35 ]
    }
  ]
```
When TRTIS is up status is:
```protobuf
id: "inference:0" 
version: "0.9.0" 
uptime_ns: 19938682171 
model_status { 
  key: "mymodel" 
  value { 
    config { 
      name: "mymodel" 
      platform: "tensorflow_graphdef" 
      version_policy { latest { num_versions: 1 } } 
      max_batch_size: 100 
      input { 
        name: "input" 
        data_type: TYPE_FP32 
        format: FORMAT_NHWC 
        dims: 299 
        dims: 299 
        dims: 3 } 
      output { 
        name: "predictions" 
        data_type: TYPE_FP32 
        dims: 35 } 
      instance_group { 
        name: "mymodel" 
        count: 1 
        gpus: 0 
        kind: KIND_GPU } 
[...]

PR #52, opened 2019-01-23
Ensured ServerStatus to update stored model config on model modification

GuanLuo:
Addressed issue #36 , now the model config shown in status API is always up to date.

The error mentioned in the issue seems to be spurious (in my experiment, always show version 1 of the model). And the error message only happens on modified model. It may due to the fact that on modified model, the inference server first triggers its unload and then reload the model, which makes the model "unavailable" for a short period of time.

### DIFF
--- a/src/core/server.cc
+++ b/src/core/server.cc
@@ -1030,6 +1030,12 @@ InferenceServer::Wait()
                         << "': " << status.error_message();
               goto next;
             }
+            status = status_manager_->UpdateConfigForModel(name);
+            if (!status.ok()) {
+              LOG_ERROR << "Failed to reload model config for '" << name
+                        << "': " << status.error_message();
+              goto next;
+            }
           }
 
           status = core_->ReloadConfig(msc);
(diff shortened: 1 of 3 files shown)


deadeyegoodwin:
Put "Fix #36" at the beginning of the commit message and it will automatically close the issue when this commit is merged.

src/core/server.cc
```diff
@@ -1030,6 +1030,12 @@ InferenceServer::Wait()
                         << "': " << status.error_message();
               goto next;
             }
+            status = status_manager_->InitForModel(name);
```
deadeyegoodwin: Note the comment above about not wanting to re-init the status since that will delete all the stats collected for the model. What we need is a net ServerStatusManager method that just updates the model configuration in the status but doesn't clear the existing status.

deadeyegoodwin: What about **ms[model_name].Clear()** in InitForModel?

GuanLuo: Oh, you are right. I missed that.

PR #56, opened 2019-01-28
Use stub library if running on CPU-only device

GuanLuo:
Note that the libcuda used is from `/usr/local/cuda-10.0/compat/lib.real/`, which is not a stub library. Because `cudaGetDeviceCount()` is called multiple times to initialize servables on proper devices (CPU if return cudaErrorNoDevice), avoid adding new condition if a usable library is already in the image.

### DIFF
--- a/qa/L0_infer/test.sh
+++ b/qa/L0_infer/test.sh
@@ -40,6 +40,16 @@ rm -f $SERVER_LOG_BASE* $CLIENT_LOG_BASE*
 RET=0
 
 for TARGET in cpu gpu; do
+    if [ "$TENSORRT_SERVER_CPU_ONLY" == "1" ]; then
+        if [ "$TARGET" == "gpu" ]; then
+            echo -e "Skip GPU testing on CPU-only device"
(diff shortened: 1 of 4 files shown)


qa/L0_infer/test.sh
```diff
@@ -40,6 +40,16 @@ rm -f $SERVER_LOG_BASE* $CLIENT_LOG_BASE*
 RET=0
 
 for TARGET in cpu gpu; do
+    if [ "$TENSORRT_SERVER_CPU_ONLY" == "1" ]; then
+        if [ "$TARGET" == "gpu" ]; then
+            echo -e "Skip GPU testing on CPU-only device"
+            break
```
deadeyegoodwin: Use "continue"

deadeyegoodwin:
Can you confirm correct results for manual test of L0_infer on a non-CUDA system?

GuanLuo:
Yes, I ran it on Vipin's machine. Using just docker and with no CUDA installed.

PR #67, opened 2019-02-01
Adding TRTIS streaming API

GuanLuo:
- [x] Add StreamInfer API on server side
- [ ] Discuss how the RPC execution would be different from Infer API
- [ ] Add client side support (C++/Python image client, perf client)
- [ ] Add test cases

### DIFF
--- a/src/core/server.cc
+++ b/src/core/server.cc
@@ -100,6 +100,7 @@
 #include "src/nvrpc/ThreadPool.h"
 
 using nvrpc::Context;
+using nvrpc::StreamingContext;
 using nvrpc::ThreadPool;
 
 namespace nvidia { namespace inferenceserver {
@@ -153,6 +154,75 @@ class StatusContext final
 
 class InferContext final
     : public Context<InferRequest, InferResponse, AsyncResources> {
+  void ExecuteRPC(InferRequest& request, InferResponse& response) final override
+  {
+    auto server = GetResources()->GetServer();
+    auto infer_stats = std::make_shared<ModelInferStats>(
+        server->StatusManager(), request.model_name());
+    auto timer = std::make_shared<ModelInferStats::ScopedTimer>();
+    infer_stats->StartRequestTimer(timer.get());
+    infer_stats->SetRequestedVersion(request.model_version());
+
+    RequestStatus* request_status = response.mutable_request_status();
+
+    auto backend = std::make_shared<InferenceServer::InferBackendState>();
+    tensorflow::Status status = server->InitBackendState(
+        request.model_name(), request.model_version(), backend);
+    if (status.ok()) {
+      infer_stats->SetModelServable(backend->Backend());
+
+      std::shared_ptr<GRPCInferRequestProvider> request_provider;
+      status = GRPCInferRequestProvider::Create(
+          *(backend->Backend()), request, &request_provider);
+      if (status.ok()) {
+        infer_stats->SetBatchSize(
+            request_provider->RequestHeader().batch_size());
+
+        std::shared_ptr<GRPCInferResponseProvider> response_provider;
+        status = GRPCInferResponseProvider::Create(
+            request.meta_data(), &response, &response_provider);
+        if (status.ok()) {
+          server->HandleInfer(
+              request_status, backend, request_provider, response_provider,
+              infer_stats,
+              [this, request_status, &response, infer_stats, timer]() mutable {
+                // If the response is an error then clear the meta-data
+                // and raw output as they may be partially or
+                // un-initialized.
+                if (request_status->code() != RequestStatusCode::SUCCESS) {
+                  response.mutable_meta_data()->Clear();
+                  response.mutable_raw_output()->Clear();
+                }
+
+                this->FinishResponse();
+                timer.reset();
+              },
+              true  // async_frontend
+          );
+        }
+      }
+    }
+
+    if (!status.ok()) {
+      LOG_VERBOSE(1) << "Infer failed: " << status.error_message();
+      infer_stats->SetFailed(true);
+      RequestStatusFactory::Create(
+          request_status, 0 /* request_id */, server->Id(), status);
+
+      // If the response is an error then clear the meta-data and raw
+      // output as they may be partially or un-initialized.
+      response.mutable_meta_data()->Clear();
+      response.mutable_raw_output()->Clear();
+
+      this->FinishResponse();
+    }
+  }
+};
+
+// For now, this context ExecuteRPC in the same way as InferContext, except that
+// it actually maintains the stream to be able to process sequence of requests
+class StreamInferContext final
+    : public StreamingContext<InferRequest, InferResponse, AsyncResources> {
   void ExecuteRPC(InferRequest& request, InferResponse& response) final override
   {
     auto server = GetResources()->GetServer();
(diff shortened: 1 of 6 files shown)


src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
+// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: copyright year

src/core/server.cc
```diff
@@ -218,6 +219,75 @@ class InferContext final
   }
 };
 
+// For now, this context ExecuteRPC in the same way as InferContext, except that
```
deadeyegoodwin: If this code is (mostly) the same then should use a common function. How is the bi-directional streaming being used here?

deadeyegoodwin:
@ryanolson Guan is adding streaming grpc infer API and so is modifying nvrpc. Please take a look.

src/core/server.cc
GuanLuo: It is the same as the ExecuteRPC(), I didn't know if I will need to change the way to acquire servable if we want the requests to be processed by the same servable instance. If no changes are needed, I agree that we can extract it as a common function.

The bidirectional streaming part is in `BidirectionalStreamingLifeCycle.h`

src/nvrpc/Interfaces.h
```diff
@@ -50,9 +50,12 @@ class IContext {
   static IContext* Detag(void* tag) { return static_cast<IContext*>(tag); }
 
  protected:
-  IContext() = default;
+  IContext() : m_MasterContext(this) {}
+  IContext(IContext* master) : m_MasterContext(master) {}
   void* Tag() { return reinterpret_cast<void*>(this); }
 
+  IContext* m_MasterContext;
```
ryanolson: @GuanLuo - can you help me understand the use case for `m_MasterContext`?  I'd like to keep upstream nvRPC up-to-date.

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// The following are a set of functions used as function pointers
+// to keep track of the state of the context.
+template <class Request, class Response>
+bool
+BidirectionalStreamingLifeCycle<Request, Response>::StateInitializedDone(
+    bool ok)
+{
+  if (!ok)
```
ryanolson: probably should scope this with {}.  my bad as the reference didn't do it either.

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+    m_RequestQueue.emplace();
+  }
+  if (m_NextState == &BidirectionalStreamingLifeCycle<
+                         RequestType, ResponseType>::StateRequestDone)
+    m_NextState = &BidirectionalStreamingLifeCycle<
+        RequestType, ResponseType>::StateResponseDone;
+  m_ReaderWriter->Read(
+      &m_RequestQueue.back(), m_ReadStateContext.IContext::Tag());
+
+  if (should_execute)
```
ryanolson: {}

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+  bool should_write = true;
+  {
+    std::unique_lock<std::mutex> lock(m_QueueMutex);
+
+    m_ResponseWriteBackQueue.pop();
+    should_write = !m_ResponseWriteBackQueue.empty();
+  }
+  // Only call Write() if the write back queue is not empty,
+  // the FinishResponse() will call Write() otherwise.
+  if (should_write)
```
ryanolson: {}

src/nvrpc/Interfaces.h
GuanLuo: The original design of nvRPC uses the address of the context as "tag" for completion queue, and the context maintains its own inner state to keep track of what kind of event to be handled next, the life cycle must follow some pattern. 

But then the context can't handle multiple events at the same time, in the case of bidirectional streaming, we can't distinguish if an event is related to Read or Write. So for streaming context, I added two "StateContext" members (derived from IContext) as a way to provides different tags (their address) for handling different events, one for read and one for write.

m_MasterContext is just pointing to the actual streaming context, such that once we know what type of event it is, we call the corresponding functions in the actual context.

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+      should_execute = true;
+      m_Executing = true;
+    }
+
+    // Start reading the next request
+    m_RequestQueue.emplace();
+  }
+  if (m_NextState == &BidirectionalStreamingLifeCycle<
+                         RequestType, ResponseType>::StateRequestDone)
+    m_NextState = &BidirectionalStreamingLifeCycle<
```
ryanolson: {}

src/nvrpc/Interfaces.h
ryanolson: I think I get it.  

`m_MasterContext` is the streams context; however, you have two internal states for reading and writing that you want to independently manage events.  The tags for each of these `StateContext::Tag()` are different, so it allows you to have those two different tags in the completion queue at the same time, even though, they both point back to the same BiDirectionalStreamingContext object.  

LifeCycles are always a bit a mind-warp until you figured out just what's going one.  Good work.

GuanLuo: Yes, that is exactly what it is doing.

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+      if (m_Executing)
+        return true;
+    }
+    return false;
+  }
+
+  // Successfully receive request
+  bool should_execute = false;
+  {
+    std::unique_lock<std::mutex> lock(m_QueueMutex);
```
ryanolson: no condition variable, so we should probably be using `std::lock_guard` until we upgraded C++17 and then we can use `std::scoped_lock`

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+  // If write didn't go through, then the call is dead. Start reseting
+  if (!ok) {
+    CancelResponse();
+    return true;
+  }
+
+  // Done writing back one response
+  bool should_write = true;
+  {
+    std::unique_lock<std::mutex> lock(m_QueueMutex);
```
ryanolson: std::lock_guard

src/nvrpc/BidirectionalStreamingLifeCycle.h
```diff
@@ -0,0 +1,366 @@
...
+}
+
+template <class Request, class Response>
+void
+BidirectionalStreamingLifeCycle<Request, Response>::FinishResponse()
+{
+  bool should_write = false;
+  bool should_execute = true;
+  {
+    std::unique_lock<std::mutex> lock(m_QueueMutex);
```
ryanolson: std::lock_guard

src/nvrpc/Interfaces.h
ryanolson: 👍 very nice 

I had previously only using the streaming interface for batching, i.e. collecting inputs, compute, then returning a stream of outputs.

This is a great generalization for the more complicated case of independent BiDirectional Reads/Writes.

When you finalize your changes to the internal version of nvRPC, please upstream it to https://github.com/NVIDIA/tensorrt-laboratory/tree/master/tensorrt-laboratory/nvrpc

I'll build a bidirectional flatbuffers example to work out the kinks with the different Message objects.

PR #78, opened 2019-02-05
Adding client-side stream API

GuanLuo:


src/clients/c++/image_client.cc
```diff
@@ -224,6 +226,10 @@ Usage(char** argv, const std::string& msg = std::string())
   std::cerr << std::endl;
   std::cerr << "If -a is specified then asynchronous client API will be used. "
             << "Default is to use the synchronous API." << std::endl;
+  std::cerr << "The --streaming flag is only valid with gRPC protocol and"
+            << " without -a flag. Otherwise, the client will be run without"
```
deadeyegoodwin: The "Otherwise..." part doesn't match the code.

src/clients/c++/request.h
```diff
@@ -666,6 +667,29 @@ class InferContext {
   Error GetReadyAsyncRequest(
       std::shared_ptr<Request>* async_request, bool wait);
 
+  /// Send current request through the context's stream. All requests sent will
```
deadeyegoodwin: Don't say anything about being directed to the same model instance... since that is beyond the scope of what the grpc stream itself will do.

src/clients/c++/request.h
```diff
@@ -666,6 +667,29 @@ class InferContext {
...
+  /// be directed to the same model instance, be processed, and be returned
+  /// in the same order. Note that you need to retrieve the responses with
```
deadeyegoodwin: Does GRPC stream guarantee that responses are in same order as requests?

src/clients/c++/request.h
```diff
@@ -666,6 +667,29 @@ class InferContext {
...
+  /// GetStreamRunResults() method.
+  ///
+  /// \return Error object indicating success or failure.
+  virtual Error StreamRun() = 0;
```
deadeyegoodwin: Having a separate API for this doesn't seem right. Instead it seems like it should just be an option to Run and RunAsync. You can do either synchronous or asynchronous over a stream, right?

src/clients/c++/request.h
GuanLuo: Not exactly, gRPC guarantees that the receiving order will be the same as the order that the other side sends back. So here is based on the assumption that the server will process the requests in sequence.

deadeyegoodwin: I general that is not true... when you call into HandleInfer there is no guarentee that the completion callbacks will be called in that same order. You will need to add that ordering to your server-side implementation. Is seems like the stream / non-stream aspect needs to be part of the InferContext. You can still send sync or async through that context and it will just send those requests over a stream or not.

src/clients/c++/request.h
deadeyegoodwin: Actually I think the stream / no-stream should be on the InferContext.

src/clients/c++/request.h
GuanLuo: Right now in the server-side streaming, the requests will be queued up and only be processed if the previous request is finished in the stream. So if the client is not using correlation ID, then using the streaming API is not desired.

I think stream / non-stream aspect sounds right. And I think, if the difference between sync/async infer is whether multiple requests can be initiated at the same time, then the difference between sync/async streaming is in terms of streams, whether multiple streams can be maintained in the InferContext at the same time.

src/clients/c++/request.h
GuanLuo: The gRPC bidirectional streaming already have the decoupling between sending requests and receiving responses. So even the "sync streaming", it is similar to async infer in some sense, after sending a request, the client can perform other operations before receiving the response.

src/clients/python/__init__.py
```diff
@@ -535,7 +535,8 @@ class ResultFormat:
         RAW = 1,
         CLASS = 2
 
-    def __init__(self, url, protocol, model_name, model_version=None, verbose=False, correlation_id=0):
+    def __init__(self, url, protocol, model_name, model_version=None,
+                 streaming=False, verbose=False, correlation_id=0):
```
deadeyegoodwin: Need to put streaming as last parameter or you will break compatibility with code that wasn't using named parameters.

src/clients/python/image_client.py
```diff
@@ -231,14 +234,17 @@ def postprocess(results, filenames, batch_size):
 
     protocol = ProtocolType.from_str(FLAGS.protocol)
 
+    if FLAGS.streaming and protocol == ProtocolType.HTTP:
+        raise Exception("Streaming flag needs to be set with gRPC protocol.")
```
deadeyegoodwin: Fix to match others

src/core/api.proto
```diff
@@ -118,6 +118,14 @@ message InferRequestHeader
   //@@
   uint64 correlation_id = 4;
 
+  //@@  .. cpp:var:: uint64 id
```
deadeyegoodwin: Can you move this in the file so it is just before correlation_id

PR #82, opened 2019-02-08
Modified server-side streaming behavior 

GuanLuo:
With this commit, the TRTIS streaming API will:
1. handle (enqueue) the requests in the stream as they arrive (always run `ExecuteRPC()` on successful `Read()`)
2. queue up the responses in write-back queue as soon as they are ready (so they may be out of "order"), the lifecycle is no longer responsible for sending back responses in the same order as their requests arrived

In short, it will behave the same as what is in the non-streaming case, except that the requests and responses are transferred in the same connection, instead of one connection for one request/response pair.

### DIFF
--- a/src/core/server.cc
+++ b/src/core/server.cc
@@ -139,18 +139,20 @@ class StatusContext final
   void ExecuteRPC(
       StatusRequest& request, StatusResponse& response) final override
   {
-    GetResources()->GetMgmtThreadPool().enqueue([this, &request, &response] {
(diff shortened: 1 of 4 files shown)


deadeyegoodwin:
I'll leave it to @ryanolson to look at the nvrpc changes. But can you add a description to your commit message describing what you change does. I think it is allowing multiple outstanding requests from the stream to be processed instead of serializing them, but you should put some detail in the description.

PR #83, opened 2019-02-08
Fixed perf_client issue with model with max_batch_size == 0

GuanLuo:


deadeyegoodwin:
Looks good. Once testing completes I will merge.

PR #87, opened 2019-02-11
Updated stream inference documentation

GuanLuo:
in "Client Libraries and Examples" and "Inference Server API"


docs/http_grpc_api.rst
```diff
@@ -43,12 +43,18 @@ for each protocol.
 * :ref:`section-api-inference`: The inference API that accepts model
   inputs, runs inference and returns the requested outputs.
 
+GRPC only endpoint:
```
deadeyegoodwin: "The inference server also exposes an endpoint based on GRPC streams <or whatever the correct term is> that is only available when using the GRPC protocol:"

PR #92, opened 2019-02-14
Fixed resize in client examples

SlipknotTN:
Currently image client examples works correcly only with square inputs. Width and height must be swapped in resize functions, see https://pillow.readthedocs.io/en/3.1.x/reference/Image.html#PIL.Image.Image.resize


deadeyegoodwin:
Thanks for the PR. As soon as our internal CI completes we will merge.

PR #93, opened 2019-02-14
Increase default timeouts to account for cloud testing (slow nodes)

deadeyegoodwin:


qa/L0_lifecycle/test.sh
```diff
@@ -100,7 +100,7 @@ if [ "$SERVER_PID" == "0" ]; then
 fi
 
 # give plenty of time for model to load (and fail to load)
-wait_for_model_stable 30
+wait_for_model_stable 120
```
GuanLuo: Just set it to `$SERVER_TIMEOUT`?

PR #97, opened 2019-02-15
Adding python module dependency list and minor documentation changes.

gitvipin:
In some cases, it is convenient to have PDF version of documentation. Generating PDF documentation from Sphinx based documentation requires certain setup. This involves installing LaTex and dependent python modules. Instructions for that setup have been added here. List of dependent python modules has been added in a requirements.txt which makes it installing these modules easier in one go itself. 

Another section also has been added onto how to validate documentation changes locally by running a simple HTTP server locally.


docs/build.rst
```diff
@@ -120,3 +123,28 @@ directory::
...
   $ make clean html
+
+PDF Documentation
+^^^^^^^^^^^^^^^^^
+
+To build the PDF version of documentation, `TEX <https://www.tug.org/texlive/quickinstall.html>`_ needs
+to be installed first. Additional requirements of python modules can be met by simply running the following
+commands in docs directory::
+
+  $ pip install -r requirements.txt
```
deadeyegoodwin: Instead of having a separate file can you follow the pattern in the "HTML Documention" section and just list the command needed to install the requirements inline with the text. Also need to still mention that the TRTIS python package must be installed.

gitvipin: Sure. Let me make that change. 

Should TRTIS python package be installed for documentation as well ? Correct me if I am wrong but that doesn't seem to be necessary requirement.

deadeyegoodwin: TRTIS python package is required so that the API docs get generated correctly for it.

PR #98, opened 2019-02-18
Warning about http-thread-count and dynamic batching.

### LINKED ISSUE #95: Warning about Dynamic Batching and thread count
The Dynamic Batching docs should warn about having a sufficient `--http-thread-count`.

with the default of `8`, you can easily fail to hit the `preferred_batch_size` and will always hit the `max_queue_delay_microseconds` timeout.  

That makes performance awful.

I experienced this with 7 models each with target batch size of 4.  More than enough concurrent requests were inflight that each model should have filled batches within an msec or two.  But verbose logging indicated that wasn't happening:

```
I0215 11:16:51.400883 1 plan_bundle.cc:450] Running foo_0_gpu0 with 2 request payloads
I0215 11:16:51.400908 1 plan_bundle.cc:450] Running bar_0_gpu0 with 1 request payloads
```

That makes sense if each http thread only services a single request at a time.  You have 7 unfilled batches, 8 blocked request handling threads, and nothing more coming in to fill/release the batches until timeout.

Bumping `--http-thread-count` to 128 resolved the issue.  Performance was excellent and verbose logging shows full batches being made:

```
I0215 11:26:14.336939 1 plan_bundle.cc:450] Running foo_0_gpu0 with 4 request payloads
I0215 11:26:14.341266 1 plan_bundle.cc:450] Running bar_0_gpu0 with 4 request payloads
```

I think a warning in the docs makes sense.  Longer term, perhaps flip the http server from threads to an event loop architecture instead?

nieksand:
Fixes #95.

### DIFF
--- a/docs/model_configuration.rst
+++ b/docs/model_configuration.rst
@@ -368,6 +368,16 @@ batch sizes of 4 and 8, and a maximum delay time of 100 microseconds::
     max_queue_delay_microseconds: 100
   }
 
+The size of generated batches can be examined in aggregate using Count
+metrics, see :ref:`section-metrics`. TRTIS verbose logging can be used
+to examine the size of individual batches.
+
+The GRPC endpoint is recommended when using dynamic batching. With the
+HTTP endpoint, the --http-thread-count must be set larger than the
+preferred batch size. Otherwise the inference server will be unable to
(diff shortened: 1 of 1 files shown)


nieksand:
Repushed branch with signed commit.

PR #114, opened 2019-02-26
Enable HTTP async infer using evhtp library

GuanLuo:


src/core/server.h
```diff
@@ -42,7 +42,6 @@
 #include "src/core/server_status.pb.h"
 #include "tensorflow/core/lib/core/status.h"
 #include "tensorflow_serving/model_servers/server_core.h"
-#include "tensorflow_serving/util/net_http/server/public/httpserver.h"
```
deadeyegoodwin: In Dockerfile and tools/patch/tfs there are some patches we no longer need to apply because we don't use TFS http service anymore.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+  void Health(evhtp_request_t* req, const std::string& health_uri);
+  void Profile(evhtp_request_t* req, const std::string& profile_uri);
+  void Infer(evhtp_request_t* req, const std::string& infer_uri);
+  void Status(evhtp_request_t* req, const std::string& status_uri);
+  static std::string ToMethodString(htp_method method);
+
+  static void thread_init(evhtp_t* htp, evthr_t* thread, void* arg);
+  static void thread_exit(evhtp_t* htp, evthr_t* thread, void* arg);
+
+  // class object associated to evhtp thread, requests received are bounded
```
deadeyegoodwin: "Class..."
Capitalize comments.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+          evhtp_request_t* req, uint64_t id,
+          std::shared_ptr<HTTPInferRequestProvider> request_provider,
+          std::shared_ptr<HTTPInferResponseProvider> response_provider)
+          : done(false), req(req), id(id), request_provider(request_provider),
+            response_provider(response_provider){};
+
+     private:
+      friend class HTTPServiceImpl;
+      friend class ThreadWiseService;
+      bool done;
```
deadeyegoodwin: All these private members should end in underscore. done_, req_, etc.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+  // class object associated to evhtp thread, requests received are bounded
+  // with the thread that accepts it. Need to keep track of that and let the
+  // corresponding thread send back reply
+  class ThreadWiseService {
+   public:
+    class Request {
+     public:
+      Request(
+          evhtp_request_t* req, uint64_t id,
+          std::shared_ptr<HTTPInferRequestProvider> request_provider,
```
deadeyegoodwin: Since you are immediately making a copy of these shared pointers you should pass them by const reference. That will avoid an unnecessary shared-pointer copy (I know that some recommend never using const& for shared_ptrs but I don't think that is correct advice).

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+      friend class ThreadWiseService;
+      bool done;
+      evhtp_request_t* req;
+      uint64_t id;
+      RequestStatus request_status;
+      std::shared_ptr<HTTPInferRequestProvider> request_provider;
+      std::shared_ptr<HTTPInferResponseProvider> response_provider;
+    };
+
+    ThreadWiseService(evthr_t* thread) : thread_(thread){};
```
deadeyegoodwin: Add "explicit" since you don't want this used as a copy or assign constructor.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+    ~ThreadWiseService() = default;
+
+    static void ReplyCallback(evthr_t* thr, void* arg, void* shared)
```
deadeyegoodwin: Put the implementation of these methods outside of the class description.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+      Infer(req, rest);
     }
     // status
     else if (endpoint == "status") {
-      status = Status(req, rest);
+      Status(req, rest);
     }
   }
-
-  if (status != tfs::net_http::HTTPStatusCode::OK) {
```
deadeyegoodwin: What happens to the request if it is not one of the above? Don't you need to response with error status? Isn't there a test for this case? If not you should add something to L0_lifecycle that makes a request for an invalid endpoint.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
-    tfs::net_http::ServerRequestInterface* req, const std::string& health_uri)
+void
+HTTPServiceImpl::Health(evhtp_request_t* req, const std::string& health_uri)
 {
   ServerStatTimerScoped timer(
       server_->StatusManager(), ServerStatTimerScoped::Kind::HEALTH);
 
-  if (req->http_method() != "GET") {
-    return tfs::net_http::HTTPStatusCode::METHOD_NA;
+  if (ToMethodString(req->method) != "GET") {
```
deadeyegoodwin: Converting to string is inefficient. Why not directly compare to htp_method_GET?

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
-    tfs::net_http::ServerRequestInterface* req, const std::string& profile_uri)
+void
+HTTPServiceImpl::Profile(evhtp_request_t* req, const std::string& profile_uri)
 {
   ServerStatTimerScoped timer(
       server_->StatusManager(), ServerStatTimerScoped::Kind::PROFILE);
 
-  if (req->http_method() != "GET") {
-    return tfs::net_http::HTTPStatusCode::METHOD_NA;
+  if (ToMethodString(req->method) != "GET") {
```
deadeyegoodwin: Compare directly

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
 
-tfs::net_http::HTTPStatusCode
-HTTPServiceImpl::Infer(
-    tfs::net_http::ServerRequestInterface* req, const std::string& infer_uri)
+void
+HTTPServiceImpl::Infer(evhtp_request_t* req, const std::string& infer_uri)
 {
-  if (req->http_method() != "POST") {
-    return tfs::net_http::HTTPStatusCode::METHOD_NA;
+  if (ToMethodString(req->method) != "POST") {
```
deadeyegoodwin: Compare directly

src/core/server.cc
```diff
@@ -532,40 +705,48 @@ HTTPServiceImpl::Infer(
...
-    tfs::net_http::ServerRequestInterface* req, const std::string& status_uri)
+void
+HTTPServiceImpl::Status(evhtp_request_t* req, const std::string& status_uri)
 {
   ServerStatTimerScoped timer(
       server_->StatusManager(), ServerStatTimerScoped::Kind::STATUS);
 
-  if (req->http_method() != "GET") {
-    return tfs::net_http::HTTPStatusCode::METHOD_NA;
+  if (ToMethodString(req->method) != "GET") {
```
deadeyegoodwin: Compare directly

src/core/server.cc
```diff
@@ -472,58 +675,28 @@ HTTPServiceImpl::Infer(
...
-                    kInferResponseHTTPHeader,
-                    response_header->ShortDebugString());
-              }
+            &(request->request_status), backend, request->request_provider,
+            request->response_provider, infer_stats,
+            [requests, request]() mutable {
+              requests->FinalizeHTTPResponse(request);
             },
-            false  // async frontend
+            true  // async frontend
```
deadeyegoodwin: Not in this pull request, but in a follow-up pull request should remove this option and clean up the downstream synchronous path.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+      std::shared_ptr<HTTPInferRequestProvider> request_provider;
+      std::shared_ptr<HTTPInferResponseProvider> response_provider;
+    };
+
+    ThreadWiseService(evthr_t* thread) : thread_(thread){};
+    ~ThreadWiseService() = default;
+
+    static void ReplyCallback(evthr_t* thr, void* arg, void* shared)
+    {
+      ThreadWiseService* requests = (ThreadWiseService*)arg;
```
deadeyegoodwin: Calling ThreadWiseServer variables here and elsewhere "requests" is very confusing.

src/core/server.cc
```diff
@@ -294,150 +293,354 @@ class HTTPServiceImpl {
...
+      Request* request = nullptr;
+      {
+        std::lock_guard<std::mutex> lock(requests->mtx_);
+        // Using iterator will cause segfault
+        for (size_t i = 0; i < requests->requests_.size(); i++) {
```
deadeyegoodwin: I don't understand this loop. Why aren't you just doing the response for one request? If there are multiple done requests you just seem to erase them all and only remember the last one. Seems like there could be lost/leaked references. Using erase this way can be very inefficient.

Why do you even need requests_ at all? Doesn't the callback return the Request object?

GuanLuo: I didn't implement it correctly, one invocation of the callback should only send and erase one selected request. I am going to fix it like you said, directly passing the Request object to the callback.

src/core/server.cc
ryanolson: Depends on if you are capturing ownership or not.  Passing a reference to a shared_ptr is a little like passing a raw pointer since it could disappear out from under you.

But I agree, I pass references to shared_ptrs quite a bit, but only when I can guarantee the lifecycle of the object.

src/core/server.cc
```diff
@@ -294,150 +294,235 @@ class HTTPServiceImpl {
...
+        const std::shared_ptr<HTTPInferRequestProvider>& request_provider,
+        const std::shared_ptr<HTTPInferResponseProvider>& response_provider,
+        const std::shared_ptr<ModelInferStats>& infer_stats,
+        const std::shared_ptr<ModelInferStats::ScopedTimer>& timer);
+
+    void FinalizeHTTPResponse(Request* req);
+
+   private:
+    std::mutex mtx_;
+    std::set<Request*> requests_;
```
deadeyegoodwin: You don't need requests_ at all, do you? I don't think you need the mutex either.

src/core/server.cc
```diff
@@ -294,150 +294,235 @@ class HTTPServiceImpl {
...
+      std::shared_ptr<ModelInferStats> infer_stats_;
+      std::shared_ptr<ModelInferStats::ScopedTimer> timer_;
+    };
+
+    explicit ThreadWiseService(evthr_t* thread) : thread_(thread){};
+    ~ThreadWiseService() = default;
+
+    static void ReplyCallback(evthr_t* thr, void* arg, void* shared);
+
+    Request* Enqueue(
```
deadeyegoodwin: Enqueue is not a good name. You are just creating the request object.So CreateRequest would be a better name, or just new Request directly.

src/core/server.cc
```diff
@@ -472,58 +557,28 @@ HTTPServiceImpl::Infer(
...
       std::shared_ptr<HTTPInferResponseProvider> response_provider;
       status = HTTPInferResponseProvider::Create(
-          req->OutputBuffer(), *(backend->Backend()),
+          req->buffer_out, *(backend->Backend()),
           request_provider->RequestHeader(), &response_provider);
       if (status.ok()) {
+        evhtp_connection_t* htpconn = evhtp_request_get_connection(req);
+        evthr_t* thread = htpconn->thread;
+        ThreadWiseService* service = (ThreadWiseService*)evthr_get_aux(thread);
+        ThreadWiseService::Request* request = service->Enqueue(
```
deadeyegoodwin: Passing around a raw point is a little dangerous since it can get dropped and so cause a leak. Can you wrap it in a unique or shared ptr. unique would be best but that can be tricky getting it into the lamda. Then you don't need to explicitly delete...

GuanLuo: I used shared pointer, but the thread callback function requires `void*` for argument, so the evhtp request pointer is passed directly since it already has the content for the response at this point (end of `FinalizeHTTPResponse()`) and evhtp will free its resource for us.

PR #116, opened 2019-02-27
Add fields in model config for ensembling

GuanLuo:


src/core/model_config.proto
```diff
@@ -517,6 +517,62 @@ message ModelSequenceBatching
...
+  {
+    //@@    .. cpp:var:: string model_name
+    //@@
+    //@@       The name of the model.
+    //@@
+    string model_name = 1;
+
+    //@@  .. cpp:var:: map<string,string> input_alias_map
+    //@@
+    //@@     Map from the inputs specified in the model cofig to aliases
```
deadeyegoodwin: Map from ensemble tensor name to name of an input tensor on this step's model. The ensemble tensor must have the same datatype and shape as the model input. Each model input must be assigned to one ensemble tensor, but the same ensemble tensor can be assigned to multiple model inputs.

src/core/model_config.proto
```diff
@@ -517,6 +517,62 @@ message ModelSequenceBatching
...
+    //@@
+    //@@     Map from aliases that is used within the ensemble to the outputs
+    //@@     specified in the model cofig. The alias will have the same
+    //@@     data type and dimensions as the output. As implied in key value
+    //@@     order, each alias can only map to one output in one model, however,
+    //@@     one alias may be mapped to multiple outputs across models
+    //@@     as long as the data type and dimensions are matched,
+    //@@     the content of the alias will be overwritten by the newest output.
+    //@@
+    map<string, string> alias_output_map = 3;
```
deadeyegoodwin: output_map

src/core/model_config.proto
```diff
@@ -517,6 +517,62 @@ message ModelSequenceBatching
...
+    //@@     data type and dimensions as the input. As implied in key value
+    //@@     order, each input of the model should only map to one alias.
+    //@@     However, mapping multiple inputs to the same alias is allowed
+    //@@     as long as the data type and dimensions are matched.
+    //@@
+    map<string, string> input_alias_map = 2;
+
+    //@@  .. cpp:var:: map<string,string> alias_output_map
+    //@@
+    //@@     Map from aliases that is used within the ensemble to the outputs
```
deadeyegoodwin: Change to follow pattern of input message above.

src/core/model_config.proto
```diff
@@ -517,6 +517,62 @@ message ModelSequenceBatching
...
+    //@@     one alias may be mapped to multiple outputs across models
+    //@@     as long as the data type and dimensions are matched,
+    //@@     the content of the alias will be overwritten by the newest output.
+    //@@
+    map<string, string> alias_output_map = 3;
+  }
+
+  //@@  .. cpp:var:: Step step (repeated)
+  //@@
+  //@@     The model(s) and the input output mappings used within the ensemble,
```
deadeyegoodwin: input / output

src/core/model_config.proto
```diff
@@ -517,6 +517,62 @@ message ModelSequenceBatching
...
+  //@@     it needs to contain a Step message specifying the mapping between
```
deadeyegoodwin: Delete these lines. Just end with "within the ensemble."

src/core/model_config.proto
```diff
@@ -517,6 +517,62 @@ message ModelSequenceBatching
...
+//@@.. cpp:var:: message ModelEnsembling
+//@@
+//@@   Model ensembling configuration. These settings control how the inputs
+//@@   and outputs of the models are mapped together.
+//@@
+message ModelEnsembling
+{
+  //@@  .. cpp:var:: message Step
+  //@@
+  //@@     The step specifies a model included in the ensemble and how its
```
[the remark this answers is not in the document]
GuanLuo: Should we mention that the data type and shape of the ensemble tensors will be inferred from mapped output tensor?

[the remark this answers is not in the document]
deadeyegoodwin: You can put that in the description of the output_map below.

src/core/model_config.proto
```diff
@@ -517,6 +517,57 @@ message ModelSequenceBatching
...
+    //@@
+    map<string, string> input_map = 2;
+
+    //@@  .. cpp:var:: map<string,string> output_map
+    //@@
+    //@@     Map from ensemble tensor name to name of an output tensor on this
+    //@@     step's model. The data type and shape of the ensemble tensor will
+    //@@     be inferred from the model output. Each model output must be
+    //@@     assigned to one ensemble tensor, but the same ensemble tensor can
+    //@@     be assigned to multiple model outputs, where the content of
```
deadeyegoodwin: We could allow one model output to be assigned to multiple ensemble tensors, but there is no reason to do that.. so for now we should disallow it. And there can be no overwriting, each ensemble tensor can appear in an output map only once. Once we have conditionals we will need to revisit this (though we will still likely want to keep the define-once invariant and instead use some other technique to handle it)

src/core/model_config.proto
```diff
@@ -517,6 +517,57 @@ message ModelSequenceBatching
...
+    //@@     the ensemble tensor will be overwritten by the newest model output.
+    //@@
+    map<string, string> output_map = 3;
+  }
+
+  //@@  .. cpp:var:: Step step (repeated)
+  //@@
+  //@@     The model(s) and the input / output mappings used within the ensemble.
```
deadeyegoodwin: I would just say "models", there is no point in having an ensemble with 1 model.

src/core/model_config.proto
GuanLuo: For the overwriting, I was thinking about the same ensemble tensor are in different models' output maps. Then there will be conflict in the content of ensemble tensor if those models are run. But this can be avoided if we have conditionals, like making sure that only one of those model will be run.

PR #129, opened 2019-03-04
Fix conditionals on max_batch_size == 0 and add test with different batch sizes

GuanLuo:


qa/L0_perf_client/test.sh
```diff
@@ -0,0 +1,87 @@
...
+SERVER=/opt/tensorrtserver/bin/trtserver
+SERVER_ARGS=--model-store=$DATADIR
+SERVER_LOG="./inference_server.log"
+source ../common/util.sh
+
+rm -f $SERVER_LOG $CLIENT_LOG
+
+RET=0
+
+rm -fr models && \
```
deadeyegoodwin: You can just point DATADIR at qa_model_repository instead of copying. And choose a batch and nobatch model from there...

PR #138, opened 2019-03-08
Add Ensemble Model Config Validation and Unit Tests

GuanLuo:


src/core/utils.cc
```diff
@@ -37,6 +38,252 @@
...
+  TensorNode(std::string model, DataType type, DimsList dims)
+      : model_name(model), type(type), dims(dims), ready(false)
+  {
+  }
+
+  std::string model_name;
+  DataType type;
+  DimsList dims;
+  bool ready;
+  std::vector<struct TensorNode*> prev_nodes;
```
deadeyegoodwin: You shouldn't need the "struct" here or below.

src/core/utils.cc
```diff
@@ -37,6 +38,252 @@
 
 namespace nvidia { namespace inferenceserver {
 
+namespace {
```
deadeyegoodwin: These first set of functions should go in a new file, called "ensemble_utils.cc". Can you rename utils -> model_config_utils. In model_config_utils you should have the validation for a model config that can be performed in isolation (with just the one config). So you can have some validation on the structure of the Step inputs and outputs and their naming, etc. For the ensemble_utils routines those can only be used once all model configs are loaded and will need to be re-run whenever models are loaded and unloaded... That will probably have to happen in the ModelConfigManager. Not needed for this change since you can call directly in the unit tests, but something to think about.

src/core/utils.cc
```diff
@@ -328,6 +582,30 @@ ValidateModelConfig(
         "expected model of type ", expected_platform, " for ", config.name());
   }
 
+  // Validate ensemble schudling first as it will make other validations
```
deadeyegoodwin: Fix comment to align with below changes.

src/servables/trtis/BUILD
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Name this directory /src/servables/ensemble

src/test/model_config_test_base.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Are you missing some of the "simpler" tests? Like what if an undefined ensemble tensor name is specified, what if a model input or output is not used, what if a tensor (ensemble or output) is assigned to multiple times, etc.

src/core/model_config_utils.cc
```diff
@@ -382,48 +385,233 @@ ValidateModelConfig(
...
+  // Check redundant ensemble tensors
+  for (const auto& tensor : tensors) {
+    // skip ensemble outputs as they have been check and can have no next nodes
+    if (outputs.find(tensor.first) != outputs.end()) {
+      continue;
+    }
+    if (!tensor.second.ready) {
+      return tensorflow::errors::InvalidArgument(
+          "ensemble tensor ", tensor.first,
+          " is redundant as no data will be written to it under optimistic "
```
deadeyegoodwin: What is "optimistic assumption"?

src/servables/ensemble/BUILD
```diff
@@ -0,0 +1,61 @@
...
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+load("@protobuf_archive//:protobuf.bzl", "cc_proto_library")
```
deadeyegoodwin: not needed

src/core/model_config_utils.cc
GuanLuo: It means not considering conditionals, just assuming all nodes will be run if the inputs are ready.

deadeyegoodwin: We don't have any conditionals yet so it makes the error message confusing. The error messages will be important to help users understand why their config is not correct. So write them as accurately and helpfully as possible given the capabilities we have right now.

PR #143, opened 2019-03-08
Add stub ensemble backend

GuanLuo:


src/servables/ensemble/ensemble_bundle.proto
```diff
@@ -0,0 +1,37 @@
...
+syntax = "proto3";
+
+package nvidia.inferenceserver;
+
+// Config proto for EnsembleBundleSourceAdapter.
+message EnsembleBundleSourceAdapterConfig
+{
+  // An variable to store pointer to inference server, which will be used
+  // in ensemble scheduler
+  uint64 inference_server = 1;
```
deadeyegoodwin: Is uint64 the supported way to pass a pointer via protobuf. Generally that is bad practice...

GuanLuo: No, protobuf doesn't provide any instruction on storing pointers, maybe using "bytes" if we still want to passing the pointer via protobuf.

Another way is adding a new setter function in InferenceBackend and passing the pointer before calling `Run()`

src/core/server.cc
```diff
@@ -806,6 +806,7 @@ InferenceServer::HandleInfer(
     OnCompleteInferRPC();
   };
 
+  backend->Backend()->SetInferenceServer(this);
```
deadeyegoodwin: You should put a comment here: "Need to set 'this' in each backend even though it is redundant after the first time. Once we remove TFS dependency we can construct each backend in a way that makes it directly aware of the inference server"

deadeyegoodwin:
@GuanLuo You need to manually rebase this to master before it can be merged.

PR #148, opened 2019-03-12
protocol service Stop() fix

GuanLuo:
Add @ryanolson as part of the fix is on nvRPC

### DIFF
--- a/src/core/http_server.cc
+++ b/src/core/http_server.cc
@@ -108,6 +108,7 @@ class HTTPServerImpl : public HTTPServer {
   struct event_base* evbase_;
   std::thread worker_;
   int fds_[2];
+  event* break_ev_;
 };
 
 tensorflow::Status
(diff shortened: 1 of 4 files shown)


src/core/http_server.cc
```diff
@@ -129,15 +135,25 @@ tensorflow::Status
 HTTPServerImpl::Stop()
```
deadeyegoodwin: So does Stop() now actually cause the server to stop?

src/nvrpc/Server.cc
```diff
@@ -77,6 +77,7 @@ Server::AsyncRun()
 void
 Server::Shutdown()
 {
+  m_Server->Shutdown();
```
deadeyegoodwin: Does this cause Stop() to stop the GRPC service?

GuanLuo: Stop() now will actually stop the service and return

Stop() calls this function, and then procedure will follow the shutdown process suggested in [GRPC doc](https://grpc.io/grpc/cpp/classgrpc_1_1_server_builder.html#a960d55977e1aef56fd7b582037a01bbd)

deadeyegoodwin: So if --allow-poll-model-repository the server will just exit right away? Also I think on shutdown the main thread is going to exit before Close() is complete.

GuanLuo: Yes. And I think you are right, the main thread will exit before Close() is complete in this case. Do we have to call Close() on separate thread? If not, we can only set server state to SERVER_EXITING on receiving signal and call Close() after Wait() return.

deadeyegoodwin: For the library work I'm redoing the main.cc for trtserver so I will fix it as part of that change. I'll just hold off on merging this until my part of the change is ready.

deadeyegoodwin:
Subsumed by https://github.com/NVIDIA/tensorrt-inference-server/pull/159

PR #150, opened 2019-03-12
Add SystemMemoryProviders that can read from / write to given block

GuanLuo:


src/core/provider.h
```diff
@@ -35,6 +35,42 @@ namespace nvidia { namespace inferenceserver {
...
+
+  // Reserve a block of memory with size 'byte_size'
+  // 'byte_size' is the size of the memory to be reserved
+  // Return error if the SystemMemoryBlock has reserved memory
+  // with non-zero size
+  Status ReserveBlock(size_t byte_size);
+
+  // Return a pointer to the memory location. It is caller's responsibility
+  // to make sure the access to the memory is exclusive on write.
+  void* GetBlock() const;
```
deadeyegoodwin: We don't typically use "Get" prefix (and note you don't say GetByteSize). So I would change to just "Block".

src/core/provider.h
```diff
@@ -35,6 +35,42 @@ namespace nvidia { namespace inferenceserver {
...
+
+  // Return the byte size of the memory
+  size_t ByteSize() const { return byte_size_; }
+
+ private:
+  size_t byte_size_;
+  std::unique_ptr<char[]> buffer_;
+  void* referencing_block_;
```
deadeyegoodwin: This class seems more complicated than it needs to be. Why is it not just a size and buffer? A response provider will create it and wrap it in a shared_ptr. Then when the sceduler is creating the request providers that will provider this output to the "next" models, it simply makes a copy of the shared_ptr. I don't see what referencing_block_ and ReserveBlock are needed for.

src/core/provider.h
deadeyegoodwin: const void*
The returned block memory should not be modified as it may be shared.

src/core/provider.h
```diff
@@ -200,6 +236,36 @@ class HTTPInferRequestProvider : public InferRequestProvider {
...
+ private:
+  SystemMemoryInferRequestProvider(
+      const std::string& model_name, const int64_t version)
+      : InferRequestProvider(model_name, version)
+  {
+  }
+
+  std::unordered_map<
+      std::string, std::pair<std::shared_ptr<SystemMemoryBlock>, bool>>
+      input_map_;
```
deadeyegoodwin: Be consistent with naming. Here you say "input_map_", below it is output_block_. Other places you just say "blocks".

src/core/provider.h
GuanLuo: I see. referencing_block_ is for wrapping the data in the HTTP/GRPC request provider, so we don't need to copy them to create the provider for actual models.

But yes, ReserveBlock will not be needed if the response provide is creating the object.

src/core/provider.h
```diff
@@ -34,6 +34,37 @@
...
+class SystemMemoryBlock {
+ public:
+  // Create an SystemMemoryBlock object with 'byte_size'
+  SystemMemoryBlock(size_t byte_size);
+
+  // Create an SystemMemoryBlock that points to 'buffer' provided.
+  // 'buffer' is the memory that has been allocated. For instance,
+  // content returned by InferRequestProvider.GetNextInputContent().
+  // 'byte_size' is the size of the buffer
+  SystemMemoryBlock(const void* buffer, size_t byte_size);
```
deadeyegoodwin: I still don't understand the use case where this constructor is used. GetNextInputContent is used only by the backends to read input. The request provider content also has a lifetime tied to the provider, so you can't take a reference to it.

src/core/provider.h
```diff
@@ -34,6 +34,37 @@
...
+
+  // Return a pointer to the memory location.
+  const void* Buffer() const;
+
+  // Return the byte size of the memory
+  size_t ByteSize() const { return byte_size_; }
+
+ private:
+  friend class SystemMemoryInferResponseProvider;
```
deadeyegoodwin: Why does it need to be  a friend. That implies something is wrong with the abstraction.
Just add MutableBuffer() for non-const buffer access.

src/core/provider.cc
```diff
@@ -787,4 +809,148 @@ InferResponseProvider::FinalizeResponse(const InferenceBackend& is)
...
+
+  if (content_byte_size != output_blocks_[name]->ByteSize()) {
+    return Status(
+        RequestStatusCode::INVALID_ARG,
+        "unexpected size " + std::to_string(output_blocks_[name]->ByteSize()) +
+            " for output '" + name + "', expecting " +
+            std::to_string(content_byte_size));
+  }
+
+  *content = output_blocks_[name]->buffer_.get();
```
deadeyegoodwin: Call MutableBuffer

src/core/provider.cc
```diff
@@ -787,4 +809,148 @@ InferResponseProvider::FinalizeResponse(const InferenceBackend& is)
...
+{
+  *content = nullptr;
+
+  Output* output;
+  RETURN_IF_ERROR(CheckAndSetIfBufferedOutput(
+      name, content, content_byte_size, content_shape, &output));
+
+  // Always write output tensor to an output block no matter
+  // if output has cls field defined
+  if (output_blocks_.find(name) != output_blocks_.end()) {
```
deadeyegoodwin: Performance. Get the iterator to output_blocks_[name] once and then use it in all places below.

src/core/provider.h
```diff
@@ -35,16 +35,74 @@ namespace nvidia { namespace inferenceserver {
...
+//
+// SystemMemory used to access data in providers
+//
+class SystemMemory {
+ public:
+  // Get the 'idx'-th data block in the buffer. Using index to avoid
+  // maintaining internal state such that one buffer can be shared
+  // across multiple providers.
+  // 'idx' zero base index. Valid indices are continuous.
+  // 'byte_size' returns the byte size of the chunk of bytes. Returns 0 if
```
deadeyegoodwin: Returns nullptr if...

PR #151, opened 2019-03-12
Decouple from tensorflow::Status to avoid unnecessary TF dependency

deadeyegoodwin:
A large but shallow change to replace tensorflow::Status with nvidia::inferenceserver::Status except where we are actually interacting with the TF or TFS API.

### DIFF
--- a/src/core/request_status.h
+++ b/src/core/request_status.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
 //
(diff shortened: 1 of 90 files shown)


src/core/request_status.h
```diff
@@ -26,24 +26,24 @@
...
   static void Create(
       RequestStatus* status, uint64_t request_id, const std::string& server_id,
       RequestStatusCode code);
 
-  // Create a Status object from a TensorFlow status.
+  // Create a RequestStatus object a Status.
   static void Create(
       RequestStatus* status, uint64_t request_id, const std::string& server_id,
-      const tensorflow::Status& tf_status);
+      const Status& isstatus);
```
GuanLuo: is_status?

PR #153, opened 2019-03-13
Change L0_unit_test to get up-to-date TRT plan from dldata

deadeyegoodwin:


qa/L0_unit_test/test.sh
```diff
@@ -32,6 +32,24 @@ TEST_LOG="./unit_test.log"
...
+        testdata/autofill_sanity/bad_input_dims/1   \
+        testdata/autofill_sanity/unknown_input/1    \
+        testdata/autofill_sanity/empty_config/1     \
+        testdata/autofill_sanity/unknown_output/1   \
+        testdata/autofill_sanity/bad_output_dims/1  \
+        testdata/autofill_sanity/no_config/1        \
+        testdata/autofill_sanity/bad_output_type/1  \
+        testdata/autofill_sanity/too_few_inputs/1 ; do
+    mkdir -p /workspace/src/servables/tensorrt/$PATH
+    cp /data/inferenceserver/qa_model_repository/plan_float32_float32_float32/1/model.plan \
```
GuanLuo: Should we always perform `./gen_qa_model_repository` to make sure the test run outside of CI will also uses up-to-date plan?

deadeyegoodwin: We generate the CI models offline for two reasons:
1. Save time, since generating them for every CI is somewhat expensive.
2. Generating models requires running docker images, and since CI runs in docker that would require docker-in-docker which is sketchy (at least at the time the CI was set up).

GuanLuo: Ok, so as long as the users are following the [doc on testing](https://docs.nvidia.com/deeplearning/sdk/tensorrt-inference-server-guide/docs/test.html#), it should be fine.

PR #154, opened 2019-03-13
Add ensemble scheduler

GuanLuo:


src/core/backend.cc
```diff
@@ -241,9 +242,11 @@ InferenceBackend::SetConfiguredScheduler(
   if (config_.has_sequence_batching()) {
     RETURN_IF_ERROR(SequenceBatchScheduler::Create(
         config_, runner_cnt, OnInit, OnRun, &scheduler));
-  } else {
+  } else if (config_.has_dynamic_batching()) {
```
deadeyegoodwin: Dynamic batching has to be the default choice because if the model configuration does not list any scheduler then it should use the dynamic scheduler.

src/core/backend.h
```diff
@@ -102,6 +102,9 @@ class InferenceBackend {
       const uint32_t runner_cnt, Scheduler::StandardInitFunc OnInit,
       Scheduler::StandardRunFunc OnRun);
 
+  // Get the raw pointer to the scheduler of the servable.
+  Scheduler* GetRawScheduler() { return scheduler_.get(); }
```
deadeyegoodwin: Just "Scheduler". I know there are some Get prefixes on other functions but I plan to fix those.

src/servables/ensemble/ensemble_bundle.cc
```diff
@@ -57,10 +57,9 @@ EnsembleBundle::Init(const std::string& path, const ModelConfig& config)
...
 EnsembleBundle::SetInferenceServer(void* inference_server)
 {
-  // [TODO] Uncommon below after DLIS-290
-  // EnsembleScheduler* scheduler =
-  //    static_cast<EnsembleScheduler*>(scheduler_.get());
-  // RETURN_IF_ERROR(scheduler->SetInferenceServer(inference_server));
+  EnsembleScheduler* scheduler =
+      static_cast<EnsembleScheduler*>(GetRawScheduler());
+  RETURN_IF_ERROR(scheduler->SetInferenceServer(inference_server));
   return InferenceBackend::SetInferenceServer(inference_server);
```
deadeyegoodwin: It seems like you don't actually have to store inference_server_ in backend.h. The SetInferencServer() method can just do nothing by default.

src/core/ensemble_scheduler.h
```diff
@@ -0,0 +1,70 @@
...
+
+  // \see Scheduler::Enqueue()
+  void Enqueue(
+      const std::shared_ptr<ModelInferStats>& stats,
+      const std::shared_ptr<InferRequestProvider>& request_provider,
+      const std::shared_ptr<InferResponseProvider>& response_provider,
+      std::function<void(Status)> OnComplete) override;
+
+  // Set the inference server that the scheduler is communicating with
+  Status SetInferenceServer(void* inference_server)
```
deadeyegoodwin: Override

src/core/ensemble_scheduler.cc
```diff
@@ -0,0 +1,245 @@
...
+  static void Proceed(
+      const std::shared_ptr<EnsembleContext>& context,
+      const std::shared_ptr<EnsembleStep>& completed_step);
+
+  // Helper function that updates ensemble state given 'completed_step'
+  void UpdateEnsembleState(const std::shared_ptr<EnsembleStep>& completed_step);
+
+  // Helper function that returns a list of step that should be run
+  // under current state.
+  // 'ok' will be set to false if the ensemble context will not make any
```
deadeyegoodwin: 'ok' is not a good name. Do you 'deadlocked'? Or 'complete'?

src/core/ensemble_scheduler.cc
```diff
@@ -0,0 +1,245 @@
...
+    InferenceServer* is, const ModelConfig& config,
+    const std::unordered_map<std::string, EnsembleTensor>& ensemble_graph,
+    const std::shared_ptr<ModelInferStats>& stats,
+    const std::shared_ptr<InferRequestProvider>& request_provider,
+    const std::shared_ptr<InferResponseProvider>& response_provider,
+    std::function<void(Status)> OnComplete)
+    : is_(is), config_(config), ensemble_graph_(ensemble_graph), stats_(stats),
+      request_provider_(request_provider),
+      response_provider_(response_provider), OnComplete_(OnComplete)
+{
```
deadeyegoodwin: At some point, perhaps here, you need to grab servable handles (backends) for all the models you will need in the ensemble. Otherwise they could be unloaded while the ensemble is executing.

src/core/ensemble_scheduler.cc
```diff
@@ -0,0 +1,531 @@
...
+      *((*(*step)->backend_)()), (*step)->request_provider_->RequestHeader(),
+      &((*step)->response_provider_)));
+
+  return Status::Success;
+}
+
+Status
+EnsembleContext::FinishEnsemble()
+{
+  if (stats_ == nullptr) {
```
deadeyegoodwin: Overloading stats for this purpose is confusing...

src/core/ensemble_scheduler.cc
```diff
@@ -0,0 +1,531 @@
...
+    StepList res;
+    if (ensemble_status_.IsOk()) {
+      std::vector<size_t> updated_tensors;
+      ensemble_status_ = UpdateEnsembleState(completed_step, updated_tensors);
+      if (ensemble_status_.IsOk()) {
+        ensemble_status_ = GetNextSteps(updated_tensors, res);
+      }
+    }
+    // Error or no more progress (completed or deadlock)
+    if ((!ensemble_status_.IsOk()) || (inflight_request_counter_ == 0)) {
```
deadeyegoodwin: inflight_steps_counter_ would be a more meaningful name. I assume this is just to be extra safe since you verify that the emsemble graph is fully connected and has no loops?

Also, shouldn't you set some kind of error if you do finish the ensemble because of a deadlock?

src/core/ensemble_scheduler.cc
```diff
@@ -0,0 +1,531 @@
...
+}
+
+Status
+EnsembleContext::GetNextSteps(
+    const std::vector<size_t>& updated_tensors, StepList& steps)
+{
+  steps.clear();
+
+  std::set<size_t> next_step_idx;
+  // Get steps whose tensors used for input are set
```
deadeyegoodwin: This is not a very efficient way to execute the graph. We could have a more direct represntation of the graph so we don't have to iterate over all the tensors. It is fine for a first implementation but at some point we may need to refine it.

GuanLuo:
New changes since last review starts from this commit https://github.com/NVIDIA/tensorrt-inference-server/pull/154/commits/fc784b60833bb698ef4fc52fe43063d0d79ae482

src/core/ensemble_scheduler.h
deadeyegoodwin: Override is still missing on SetInferenceServer

GuanLuo: This is a new function for EnsembleScheduler, the Scheduler interface doesn't have this function. Should I also add this to Scheduler and default to do nothing?

deadeyegoodwin: My mistake, I thought it was the override from Backend.

PR #164, opened 2019-03-18
Improve metric documentation

deadeyegoodwin:


deadeyegoodwin:
@tripti-singhal please take a look.

PR #172, opened 2019-03-22
Allow wildcard dimension to have zero size

deadeyegoodwin:


deadeyegoodwin:
> What is the intention of allowing zero size dimension? Doesn't a zero size dimension just make the input redundant?

If any dimension is zero-sized then the entire tensor size is 0 and so no data is expected for the input. Some models have variable-sized input tensors where it is valid to provide no data for some of the inputs, and so we need to support that case.

PR #177, opened 2019-03-26
Add input/output reshape option to allow zero-rank tensors

deadeyegoodwin:


src/core/model_config.proto
```diff
@@ -206,9 +220,19 @@ message ModelInput
...
+  //@@     The dimensions/shape of the input tensor that must be provided
+  //@@     when invoking the inference API for this model.
   //@@
   repeated int64 dims = 4;
+
+  //@@  .. cpp:var:: ModelTensorReshape reshape
+  //@@
+  //@@     The shape expected for this input by the backend. The input will
+  //@@     be reshaped to this before being presented to the backend. The
+  //@@     reshape must have the same number of elements as the input shape
```
GuanLuo: It seems like reshape can have no dimension, then this statement is not accurate.

src/core/model_config.proto
```diff
@@ -236,6 +260,15 @@ message ModelOutput
   //@@
   repeated int64 dims = 3;
 
+  //@@  .. cpp:var:: ModelTensorReshape reshape
+  //@@
+  //@@     The shape produed for this output by the backend. The output will
```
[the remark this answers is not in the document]
deadeyegoodwin: Fixed

src/core/model_config.proto
GuanLuo: I see, so [] is like an implicit [1]. i.e. dims: [1] and reshape: [] is ok, but dims:[2] and reshape: [] is not ok?

deadeyegoodwin: You have to reshape into something that has the same number of elements. So going from [ bs, 1 ] to [ bs ] is ok (in this case dims is [1] and reshape is []). Going from dims [2,2] to reshape [4,1,1] is also ok. But going from dims [2] to reshape [ ] is not ok since with batch that is [ bs, 2 ] to [ bs ] which has different number of elements (also not ok with non-batching model)

PR #178, opened 2019-03-26
Add custom backend for image preprocessing

GuanLuo:


qa/L0_custom_image_preprocess/models/image_preprocess/config.pbtxt
```diff
@@ -0,0 +1,54 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "image_preprocess"
```
deadeyegoodwin: You should use a more precise name, like "image_preprocess_nhwc_224x224x3".

src/clients/c++/simple_ensemble_client.cc
```diff
@@ -0,0 +1,294 @@
...
+        break;
+      case '?':
+        Usage(argv);
+        break;
+    }
+  }
+
+  nic::Error err;
+
+  // We use a simple model that takes 2 input tensors of 16 strings
```
deadeyegoodwin: Comment is wrong

qa/L0_custom_image_preprocess/preprocess_test.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/python
```
deadeyegoodwin: I see that this is a test just of the custom image preprocess backend. That is fine. I don't see why it also have "simple_ensemble_client" in the test. The should be a separate L0_ensemble_image_client test that runs the simple client that uses the ensemble model to perform image classification.

src/clients/c++/BUILD
```diff
@@ -135,6 +144,17 @@ cc_binary(
     ],
 )
 
+cc_binary(
+    name = "simple_ensemble_client",
```
deadeyegoodwin: Call this "ensemble_image_client" as a parallel to image_client.

src/clients/c++/simple_ensemble_client.cc
```diff
@@ -0,0 +1,294 @@
...
+  std::cerr << "\t-v" << std::endl;
+  std::cerr << "\t-i <Protocol used to communicate with inference service>"
+            << std::endl;
+  std::cerr << "\t-u <URL for inference service>" << std::endl;
+  std::cerr << "\t-p <proprocessed output filename prefix>" << std::endl;
+  std::cerr << std::endl;
+  std::cerr
+      << "For -i, available protocols are 'grpc' and 'http'. Default is 'http."
+      << std::endl;
+  std::cerr
```
deadeyegoodwin: I would get rid of -p. The preprocess aspect of the inference request is going to be hidden from the client because it is going to be part of the ensemble.

src/clients/c++/simple_ensemble_client.cc
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  nic::Error err;
+
+  // We use a simple model that takes 2 input tensors of 16 strings
+  // each and returns 2 output tensors of 16 strings each. The input
+  // strings must represent integers. One output tensor is the
+  // element-wise sum of the inputs and one output is the element-wise
+  // difference.
+  std::string model_name = "image_preprocess";
```
deadeyegoodwin: This should use the ensemble model that includes image preprocessing followed by RN50 (or some other image classification model). This test will have to go on gitlab since it uses a RN50 model and those aren't available on github. We have a problem in that for someone to run this as an example they will have to create a model store with the custom preprocess model, the ensemble model, and a RN50 model. We should talk about the best way to add this CI test.

src/custom/image_preprocess/image_preprocess.cc
```diff
@@ -0,0 +1,533 @@
...
+  if (model_config_.output_size() != 1) {
+    return kOutput;
+  }
+
+  // The output shape must have three dims for width, height and color channel
+  if (model_config_.output(0).dims_size() != 3) {
+    return kOutput;
+  } else {
+    for (const auto& dim : model_config_.output(0).dims()) {
+      output_shape_.push_back(dim);
```
deadeyegoodwin: Need to make sure each dimension is >= 1.

src/clients/c++/simple_ensemble_client.cc
```diff
@@ -0,0 +1,294 @@
...
+    Usage(argv, "image file or image folder must be specified");
+  }
+
+  if (!err.IsOk()) {
+    std::cerr << "error: unable to create inference context: " << err
+              << std::endl;
+    exit(1);
+  }
+
+  // Create the data for the two input tensors. Initialize the first
```
deadeyegoodwin: Review all comments...

src/clients/c++/simple_ensemble_client.cc
```diff
@@ -0,0 +1,294 @@
...
+    file >> std::noskipws;
+    image_str.emplace_back((std::istreambuf_iterator<char>(file)),
+                  std::istreambuf_iterator<char>());
+    if (image_str.back().empty()) {
+      std::cerr << "error: unable to read image file " << fn << std::endl;
+      exit(1);
+    }
+  }
+
+  // only send one request for simplicity
```
deadeyegoodwin: Should print warning if the number of images exceeds what will be processed.

src/clients/c++/ensemble_image_client.cc
```diff
@@ -0,0 +1,263 @@
...
+        break;
+      case '?':
+        Usage(argv);
+        break;
+    }
+  }
+
+  nic::Error err;
+
+  // The ensemble model takes 1 input tensor with shape [ 1 ] and STRING
```
deadeyegoodwin: It would be good to add the -c flag so that the number of classification results can be selected.

PR #190, opened 2019-03-29
Extra defensive wiring for CUDA compat lib

deadeyegoodwin:


Dockerfile
```diff
@@ -243,6 +243,12 @@ COPY --from=trtserver_caffe2 /opt/pytorch/pytorch/LICENSE LICENSE.pytorch
 COPY --from=trtserver_build /opt/tensorrtserver/bin/trtserver bin/
 COPY --from=trtserver_build /opt/tensorrtserver/lib lib
 
+# Extra defensive wiring for CUDA Compat lib
+RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
```
GuanLuo: why don't you just use "lib.real"?

cliffwoolley: Because the directory must not exist on the PATH until we test that the library in it can actually initialize at runtime.  It's a bit of a mess, but just trust me - please do not change this.

PR #191, opened 2019-03-29
Example showing how to add a TF custom op

deadeyegoodwin:


docs/custom_operation.rst
```diff
@@ -0,0 +1,99 @@
...
+  # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+Custom Operations
+=================
+
+Modeling frameworks that allow custom operations are partially
+supported by the TensorRT Inference Server. Custom operations can be
+added to the server at build time or at server startup. Currently
+there is no way to associate custom operations with the models that
```
GuanLuo: Do you mean "the custom ops can't be used in models yet" or "they can't be loaded/unloaded dynamically"?

deadeyegoodwin: I updated the doc to be less confusing.

PR #196, opened 2019-04-01
Add basic addsub-like ensemble models in L0_infer / L0_infer_variable CI ...

GuanLuo:
Fix output shape of non-batching ensemble in ensemble scheduler

### DIFF
--- a/qa/L0_infer/test.sh
+++ b/qa/L0_infer/test.sh
@@ -71,7 +71,7 @@ for TARGET in cpu gpu; do
         cp -r ../custom_models/custom_int32_* models/. && \
         cp -r ../custom_models/custom_nobatch_* models/.
 
-    create_nop_modelfile ../L0_infer_reshape/libidentity.so `pwd`/models
+    create_nop_modelfile `pwd`/libidentity.so `pwd`/models
 
     KIND="KIND_GPU" && [[ "$TARGET" == "cpu" ]] && KIND="KIND_CPU"
     for FW in graphdef savedmodel netdef custom; do
--- a/qa/L0_infer/test.sh
+++ b/qa/L0_infer/test.sh
@@ -31,7 +31,8 @@ INFER_TEST=infer_test.py
 DATADIR=`pwd`/models
 
 SERVER=/opt/tensorrtserver/bin/trtserver
-SERVER_ARGS=--model-store=$DATADIR
+# Allow more time to exit. Ensemble brings in too many models
+SERVER_ARGS="--model-store=$DATADIR --exit-timeout-secs=120"
 SERVER_LOG_BASE="./inference_server"
 source ../common/util.sh
 
@@ -57,7 +58,7 @@ for TARGET in cpu gpu; do
         fi
         # set strict readiness=false on CPU-only device to allow
         # unsuccessful load of TensorRT plans, which require GPU.
-        SERVER_ARGS="--model-store=$DATADIR --strict-readiness=false"
+        SERVER_ARGS="--model-store=$DATADIR --exit-timeout-secs=120 --strict-readiness=false"
     fi
 
     SERVER_LOG=$SERVER_LOG_BASE.${TARGET}.log
@@ -65,9 +66,12 @@ for TARGET in cpu gpu; do
 
     rm -fr models && \
         cp -r /data/inferenceserver/qa_model_repository models && \
+        cp -r /data/inferenceserver/qa_ensemble_repository/* models/. && \
         cp -r ../custom_models/custom_float32_* models/. && \
         cp -r ../custom_models/custom_int32_* models/. && \
         cp -r ../custom_models/custom_nobatch_* models/.
+
+    create_nop_modelfile ../L0_infer_reshape/libidentity.so `pwd`/models
 
     KIND="KIND_GPU" && [[ "$TARGET" == "cpu" ]] && KIND="KIND_CPU"
     for FW in graphdef savedmodel netdef custom; do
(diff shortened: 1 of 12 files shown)


qa/common/gen_qa_model_repository
```diff
@@ -77,11 +80,13 @@ cp ./gen_qa_models.py $HOST_SRCDIR/.
 cp ./gen_qa_zero_models.py $HOST_SRCDIR/.
 cp ./gen_qa_reshape_models.py $HOST_SRCDIR/.
 cp ./gen_qa_sequence_models.py $HOST_SRCDIR/.
+cp ./gen_ensemble_model_utils.py $HOST_SRCDIR/.
```
deadeyegoodwin: Why isn't it just called "gen_qa_ensemble_models.py"? Its confusing that the utils used by other scripts is also a main script itself. Can you break apart that functionality.

qa/common/gen_qa_model_repository
```diff
@@ -210,3 +216,32 @@ if [ $? -ne 0 ]; then
...
+chown -R $(id -u):$(id -g) $ENSEMBLEDESTDIR
+EOF
+
+chmod a+x $HOST_SRCDIR/$ENSEMBLESCRIPT
+if [ $? -ne 0 ]; then
+    echo -e "Failed: chmod"
+    exit 1
+fi
+
+# Doesn't have to use TensorFlow image, only creating model configs of util
```
deadeyegoodwin: Do you need to run the container again? Can you just include this extra work in the existing TF container launch?

qa/common/gen_ensemble_model_utils.py
```diff
@@ -0,0 +1,428 @@
...
+    # Create utility models used in ensemble
+    # nop (only creates model config, should add model file before use)
+    model_dtypes = ["TYPE_BOOL", "TYPE_STRING"]
+    for s in [8, 16, 32, 64]:
+      for t in ["INT", "UINT", "FP"]:
+        if t == "FP" and s == 8:
+          continue
+        model_dtypes.append("TYPE_{}{}".format(t, s))
+
+    for model_dtype in model_dtypes:
```
deadeyegoodwin: How come this variable-size is done here and not in gen_qa_models.py like you do for the other models?

qa/common/gen_qa_models.py
```diff
@@ -870,3 +915,19 @@ def create_fixed_models(
                       (8,-1), (8,4), (8,4), 32)
         create_models(FLAGS.models_dir, np_dtype_string, np.int32, np_dtype_string,
                       (-1,8,-1), (2,8,2), (2,8,2), 32)
+
+    if FLAGS.ensemble:
```
deadeyegoodwin: Shouldn't this be indented in so that it is inside "if FLAGS.variable"?

GuanLuo: These nops are used in both variable and non-variable size test. For the non-variable size test, the ensemble models are fixed size, but internally it will pass the tensors to variable size nop with the same dims size, so that we don't need to create fixed size version of the nops.

deadeyegoodwin: ok

deadeyegoodwin:
I will re-generate the qa model repositories and then you can run the CI to validate.

qa/L0_infer/test.sh
```diff
@@ -57,18 +58,21 @@ for TARGET in cpu gpu; do
...
     CLIENT_LOG=$CLIENT_LOG_BASE.${TARGET}.log
 
     rm -fr models && \
         cp -r /data/inferenceserver/qa_model_repository models && \
+        cp -r /data/inferenceserver/qa_ensemble_model_repository/qa_model_repository/* models/. && \
         cp -r ../custom_models/custom_float32_* models/. && \
         cp -r ../custom_models/custom_int32_* models/. && \
         cp -r ../custom_models/custom_nobatch_* models/.
 
+    create_nop_modelfile ../L0_infer_reshape/libidentity.so `pwd`/models
```
deadeyegoodwin: Don't copy from L0_infer_reshape. That is very brittle. If you need libidentity.so in the test then change Dockerfile.QA to copy it in when create the QA container.

qa/L0_infer_variable/test.sh
```diff
@@ -45,10 +45,13 @@ for TARGET in cpu gpu; do
 
     rm -fr models && \
         cp -r /data/inferenceserver/qa_variable_model_repository models && \
+        cp -r /data/inferenceserver/qa_ensemble_model_repository/qa_model_repository/* models/. && \
         cp -r ../custom_models/custom_float32_* models/. && \
         cp -r ../custom_models/custom_int32_* models/. && \
         cp -r ../custom_models/custom_nobatch_* models/.
 
+    create_nop_modelfile ../L0_infer_reshape/libidentity.so `pwd`/models
```
deadeyegoodwin: Use Dockerfile.QA

qa/common/gen_qa_model_repository
```diff
@@ -59,16 +59,25 @@ HOST_ZERODESTDIR=/tmp/qa_zero_model_repository
...
 mkdir -p $HOST_VARDESTDIR
 mkdir -p $HOST_ZERODESTDIR
 mkdir -p $HOST_RESHAPEDESTDIR
 mkdir -p $HOST_SEQDESTDIR
 mkdir -p $HOST_VARSEQDESTDIR
+# Since the models required by ensemble models may not be available
+# at this point, storing ensemble models separately so that other qa directories
+# are proper model stores
+mkdir -p $HOST_ENSEMBLEDESTDIR/qa_model_repository
+mkdir -p $HOST_ENSEMBLEDESTDIR/qa_variable_model_repository
```
deadeyegoodwin: You don't seem to generate/use variable, zero, or reshape ensemble repositories. Is that coming later?

GuanLuo: Yes. The models in both reshape and zero are essentially nop, so I am going to add them in a different PR since they are different from infer / variable in this sense.

PR #197, opened 2019-04-01
Restructure build to create trtserver shared library and main executable

deadeyegoodwin:
In preparation for an in-process API.

Decouple autofill backend from core autofill, avoid circular build dependency

Move GRPC/HTTP service code out of core


deadeyegoodwin:
@GuanLuo Sorry for the huge PR... it is mostly just to get TF and TFS untangled enough so we can build a library and link against it while allowing the main program to include the headers without bazel pulling all of TF into the main executable.

src/servables/tensorrt/BUILD
```diff
@@ -102,10 +104,9 @@ cc_library(
         ":plan_bundle",
         ":plan_bundle_proto",
         "//src/core:constants",
-        "//src/core:logging",
         "//src/core:model_config",
```
GuanLuo: Since `//src/core:server_header` contains `model_config.h`, can we also remove this?

deadeyegoodwin: Yes, we probably could. I likely have some redundant dependencies here and there in BUILD files. Feel free to clean them up if you come across them in an are you are working on.

PR #198, opened 2019-04-02
Add model input / output validation in ValidateModelConfig(). Since ...

GuanLuo:
custom backend and ensemble backend will not call the corresponding validation functions explicitly.


src/core/model_config_utils.cc
```diff
@@ -492,6 +492,20 @@ ValidateModelConfig(
     }
   }
 
+  Status status;
```
deadeyegoodwin: I think this is a good change, but should be moved to line 360, and the input/output checks there should be moved into ValidateModelInputOutput (likely need to pass in max_batch_size).

The history is that input/output validation used to be in here but I forget why it moved out and probably this is now better. But you need to check where the current usage is in the bundles and remove the direct calls to ValidateModelInput/Output. Also I recently added explicit calls to these in the unit-test harness and those can now be removed.

src/core/model_config_utils.h
```diff
@@ -104,33 +102,33 @@ Status BuildEnsembleGraph(
...
 
-/// Validate that an input is specified correctly in a model
-/// configuration and matches one of the allowed input names.
+/// Validate that an input matches one of the allowed input names.
 /// \param io The model input.
 /// \param allowed The set of allowed input names.
 /// \return The error status. A non-OK status indicates the input
 /// is not valid.
-Status ValidateModelInput(
+Status AllowModelInput(
```
deadeyegoodwin: Can you change name to "CheckAllowedModelInput" so it is clear that this is a validation and not an assignment.

src/core/model_config_utils.h
```diff
@@ -104,33 +102,33 @@ Status BuildEnsembleGraph(
...
 
-/// Validate that an output is specified correctly in a model
-/// configuration and matches one of the allowed output names.
+/// Validate that an output matches one of the allowed output names.
 /// \param io The model output.
 /// \param allowed The set of allowed output names.
 /// \return The error status. A non-OK status indicates the output
 /// is not valid.
-Status ValidateModelOutput(
+Status AllowModelOutput(
```
deadeyegoodwin: CheckAllowedModelOutput

PR #215, opened 2019-04-09
set libjpeg symbols to local

GuanLuo:


src/core/lib.ldscript
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Can you call this libtrtserver.ldscript?

PR #226, opened 2019-04-12
 Allow label lookup in ensemble if label file is not provided 

GuanLuo:


src/core/label_provider.h
```diff
@@ -38,17 +38,28 @@ class LabelProvider {
  public:
   LabelProvider() = default;
 
+  using Labels = std::vector<std::string>;
+
   // Return the label for a given 'index' of a 'name'. Return empty
   // string if no label is available.
   const std::string& GetLabel(const std::string& name, size_t index) const;
 
+  // Return the labels for a given 'name'. Return nullptr
```
deadeyegoodwin: LabelProvider is only the labels, so it doesn't make sense to just shared_ptr the labels part. Change backend.h to use a shared_ptr for LabelProvider there and just use shared_ptr<LabelProvider> everywhere to share it. I don't think LabelProvider needs any changes (except the header comments need updating... I actually have a pending change for that).

src/core/provider.h
```diff
@@ -242,7 +243,13 @@ class InferResponseProvider {
   Status OutputBufferContents(
       const std::string& name, void** content, size_t* content_byte_size) const;
 
-  // Finialize response based on a servable.
+  // Set label provider for classification output.
+  void SetLabelProvider(const LabelProvider& p) { label_provider_ = &p; }
```
deadeyegoodwin: Don't have a Set method, but add shared_ptr<LabelProvider> to the constructor.

src/core/provider.h
```diff
@@ -276,6 +283,9 @@ class InferResponseProvider {
 
   // Ordered list of outputs as they "added" by AllocateOutputBuffer().
   std::vector<Output> outputs_;
+
+  // label provider used to generate classification results.
+  const LabelProvider* label_provider_;
```
deadeyegoodwin: This will be a shared_ptr<LabelProvider>

src/core/ensemble_scheduler.cc
```diff
@@ -143,6 +143,20 @@ class EnsembleContext {
   std::shared_ptr<InferRequestProvider> request_provider_;
   std::shared_ptr<InferResponseProvider> response_provider_;
   std::function<void(Status)> OnComplete_;
+
+  // Each context should have its own label provider that is a superset of
```
deadeyegoodwin: I think "superset" is the wrong word here. I'm not sure why the context needs to carry a label provider. Each ResponseProvider has a label provider...

src/core/ensemble_scheduler.cc
```diff
@@ -143,6 +143,20 @@ class EnsembleContext {
...
+  // the ensemble's label provider.
+  // Note that the context doesn't use the ensemble's label provider directly,
+  // because each context is a "snapshot" of the models when a request arrives.
+  // Different contexts of the same ensemble may have different labels due to
+  // model unload/reload, and this won't be reflected if they use the
+  // same label provider.
+  LabelProvider context_label_provider_;
+
+  // Output tensors whose labels are not provided by the ensemble servable
+  std::unordered_map<size_t, std::string> no_label_tensors_;
```
deadeyegoodwin: I think you  need a map from each ensemble tensor to the LabelProvider associated with that tensor, if any.

PR #227, opened 2019-04-12
 Add ensemble tests in L0_sequence_batcher.  Fix ensemble bug with sequence model

GuanLuo:
Ensemble context wasn't passing the flags field to internal request header.

### DIFF
--- a/qa/L0_sequence_batcher/sequence_batcher_test.py
+++ b/qa/L0_sequence_batcher/sequence_batcher_test.py
@@ -50,6 +50,16 @@ elif os.environ['BATCHER_TYPE'] == "VARIABLE":
     _trials = ("savedmodel", "graphdef", "netdef")
 else:
     _trials = ("custom", "savedmodel", "graphdef", "netdef", "plan")
+# Add ensemble to the _trials
+ENSEMBLE_PREFIXES = ["simple_", "sequence_", "fan_"]
+res = []
+for trial in _trials:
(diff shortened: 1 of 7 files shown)


qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -272,6 +282,9 @@ def check_setup(self, model_name):
         self.assertEqual(len(ss.model_status), 1)
         self.assertTrue(model_name in ss.model_status,
                         "expected status for model " + model_name)
+        # Skip the sequence batching check on ensemble model
+        if ss.model_status[model_name].config.platform == "ensemble":
```
deadeyegoodwin: Instead of returning early just reverse the condition

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -291,6 +304,11 @@ def check_status(self, model_name, static_bs, exec_cnt, infer_cnt):
         for b in static_bs:
             self.assertTrue(b in infer,
                             "expected batch-size " + str(b) + ", got " + str(vs[1]))
+
+        # Skip checking on ensemble because its execution count isn't modified like
+        # sequence batcher.
+        if ss.model_status[model_name].config.platform == "ensemble":
```
deadeyegoodwin: Don't return early

PR #229, opened 2019-04-15
Add helm chart to deploy single inference server

deadeyegoodwin:


deploy/single_server/values.yaml
```diff
@@ -0,0 +1,36 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+replicaCount: 1
+
+image:
+  imageName: nvcr.io/nvidia/tensorrtserver:19.03-py3
```
GuanLuo: And we need to update the version per release?

deadeyegoodwin: Yes, ideally. If we miss it sometimes it is not a big deal.

PR #236, opened 2019-04-17
Hemantj cleanup

CoderHam:
Fixing permission for all shared objects to be only read and executable.


Dockerfile
```diff
@@ -254,6 +254,8 @@ COPY --from=trtserver_build /opt/tensorrtserver/bin/trtserver bin/
 COPY --from=trtserver_build /opt/tensorrtserver/lib lib
 COPY --from=trtserver_build /opt/tensorrtserver/include include
 
+RUN chmod ugo-w+x /opt/tensorrtserver/lib/*.so
```
GuanLuo: `+rx` to ensure all users has both read and execute permission.

PR #241, opened 2019-04-20
Add port options for Status, Health, profile, and Infer APIs

CoderHam:
Adding port options for API endpoints to be able to serve status, health, profile and infer at different (unique) ports. 
E.g.: --http-status-port, --grpc--status-port ... (status, health, profile, infer)
PS: retained naming convention for HTTP requests for consistency.

### DIFF
--- a/src/servers/grpc_server.cc
+++ b/src/servers/grpc_server.cc
@@ -26,6 +26,7 @@
 
 #include "src/servers/grpc_server.h"
 
+#include <map>
 #include "grpc++/security/server_credentials.h"
 #include "grpc++/server.h"
 #include "grpc++/server_builder.h"
@@ -252,98 +253,65 @@ GRPCServer::~GRPCServer()
 
 Status
 GRPCServer::Create(
-    InferenceServer* server, uint16_t port, int infer_thread_cnt,
-    int stream_infer_thread_cnt, std::unique_ptr<GRPCServer>* grpc_server)
+    InferenceServer* server,
+    const std::map<int32_t, std::vector<std::string>>& port_map,
+    int infer_thread_cnt, int stream_infer_thread_cnt,
+    std::vector<std::unique_ptr<GRPCServer>>* grpc_servers)
 {
-  // DLIS-162 - provide global defaults and cli overridable options
+  if (port_map.empty()) {
+    return Status(
+        RequestStatusCode::INVALID_ARG,
+        "GRPC is enabled but none of the service endpoints have a valid port "
+        "assignment");
+  }
   g_Resources = std::make_shared<AsyncResources>(
-      server,  // InferenceServer*,
-      1,       // infer threads
-      1        // mgmt threads
-  );
+      server, 1 /* infer threads */, 1 /* mgmt threads */);
+  grpc_servers->clear();
 
-  LOG_INFO << "Building nvrpc server";
-  const std::string addr = "0.0.0.0:" + std::to_string(port);
-  grpc_server->reset(
-      new GRPCServer(addr, infer_thread_cnt, stream_infer_thread_cnt));
+  for (auto const& ep_map : port_map) {
+    std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+    LOG_INFO << "Starting a GRPCService at " << addr;
 
-  (*grpc_server)->GetBuilder().SetMaxMessageSize(MAX_GRPC_MESSAGE_SIZE);
+    grpc_servers->emplace_back(
+        new GRPCServer(addr, infer_thread_cnt, stream_infer_thread_cnt));
+    auto& grpc_server = grpc_servers->back();
 
-  LOG_INFO << "Register TensorRT GRPCService";
-  auto inferenceService = (*grpc_server)->RegisterAsyncService<GRPCService>();
+    grpc_server->GetBuilder().SetMaxMessageSize(MAX_GRPC_MESSAGE_SIZE);
 
-  LOG_INFO << "Register Infer RPC";
-  (*grpc_server)->rpcInfer_ = inferenceService->RegisterRPC<InferContext>(
-      &GRPCService::AsyncService::RequestInfer);
+    LOG_INFO << "Register TensorRT GRPCService";
+    auto inferenceService = grpc_server->RegisterAsyncService<GRPCService>();
 
-  LOG_INFO << "Register StreamInfer RPC";
-  (*grpc_server)->rpcStreamInfer_ =
-      inferenceService->RegisterRPC<StreamInferContext>(
-          &GRPCService::AsyncService::RequestStreamInfer);
+    for (auto ep_name : ep_map.second) {
+      if (ep_name == "infer") {
+        LOG_INFO << "Register Infer RPC";
+        grpc_server->rpcInfer_ = inferenceService->RegisterRPC<InferContext>(
+            &GRPCService::AsyncService::RequestInfer);
 
-  LOG_INFO << "Register Status RPC";
-  (*grpc_server)->rpcStatus_ = inferenceService->RegisterRPC<StatusContext>(
-      &GRPCService::AsyncService::RequestStatus);
+        LOG_INFO << "Register StreamInfer RPC";
+        grpc_server->rpcStreamInfer_ =
+            inferenceService->RegisterRPC<StreamInferContext>(
+                &GRPCService::AsyncService::RequestStreamInfer);
+      }
 
-  LOG_INFO << "Register Profile RPC";
-  (*grpc_server)->rpcProfile_ = inferenceService->RegisterRPC<ProfileContext>(
-      &GRPCService::AsyncService::RequestProfile);
+      if (ep_name == "status") {
+        LOG_INFO << "Register Status RPC";
+        grpc_server->rpcStatus_ = inferenceService->RegisterRPC<StatusContext>(
+            &GRPCService::AsyncService::RequestStatus);
+      }
 
-  LOG_INFO << "Register Health RPC";
-  (*grpc_server)->rpcHealth_ = inferenceService->RegisterRPC<HealthContext>(
-      &GRPCService::AsyncService::RequestHealth);
+      if (ep_name == "profile") {
+        LOG_INFO << "Register Profile RPC";
+        grpc_server->rpcProfile_ =
+            inferenceService->RegisterRPC<ProfileContext>(
+                &GRPCService::AsyncService::RequestProfile);
+      }
 
-  return Status::Success;
-}
-
-Status
-GRPCServer::CreateUniqueEndpointPorts(
-    InferenceServer* server, std::string endpoint_name, uint16_t port,
-    std::unique_ptr<GRPCServer>* grpc_server)
-{
-  // DLIS-162 - provide global defaults and cli overridable options
-  g_Resources = std::make_shared<AsyncResources>(
-      server,  // InferenceServer*,
-      1,       // infer threads
-      1        // mgmt threads
-  );
-
-  LOG_INFO << "Building nvrpc server";
-  const std::string addr = "0.0.0.0:" + std::to_string(port);
-  grpc_server->reset(new GRPCServer(addr));
-
-  (*grpc_server)->GetBuilder().SetMaxMessageSize(MAX_GRPC_MESSAGE_SIZE);
-
-  LOG_INFO << "Register TensorRT GRPCService";
-  auto inferenceService = (*grpc_server)->RegisterAsyncService<GRPCService>();
-
-  if (endpoint_name == "infer"){
-    LOG_INFO << "Register Infer RPC";
-    (*grpc_server)->rpcInfer_ = inferenceService->RegisterRPC<InferContext>(
-        &GRPCService::AsyncService::RequestInfer);
-
-    LOG_INFO << "Register StreamInfer RPC";
-    (*grpc_server)->rpcStreamInfer_ =
-        inferenceService->RegisterRPC<StreamInferContext>(
-            &GRPCService::AsyncService::RequestStreamInfer);
-  }
-
-  if (endpoint_name == "status"){
-    LOG_INFO << "Register Status RPC";
-    (*grpc_server)->rpcStatus_ = inferenceService->RegisterRPC<StatusContext>(
-        &GRPCService::AsyncService::RequestStatus);
-  }
-
-  if (endpoint_name == "profile"){
-    LOG_INFO << "Register Profile RPC";
-    (*grpc_server)->rpcProfile_ = inferenceService->RegisterRPC<ProfileContext>(
-        &GRPCService::AsyncService::RequestProfile);
-  }
-
-  if (endpoint_name == "health"){
-    LOG_INFO << "Register Health RPC";
-    (*grpc_server)->rpcHealth_ = inferenceService->RegisterRPC<HealthContext>(
-        &GRPCService::AsyncService::RequestHealth);
+      if (ep_name == "health") {
+        LOG_INFO << "Register Health RPC";
+        grpc_server->rpcHealth_ = inferenceService->RegisterRPC<HealthContext>(
+            &GRPCService::AsyncService::RequestHealth);
+      }
+    }
   }
 
   return Status::Success;
--- a/src/servers/grpc_server.cc
+++ b/src/servers/grpc_server.cc
@@ -297,6 +297,59 @@ GRPCServer::Create(
 }
 
 Status
+GRPCServer::CreateUniqueEndpointPorts(
+    InferenceServer* server, std::string endpoint_name, uint16_t port,
+    std::unique_ptr<GRPCServer>* grpc_server)
+{
+  // DLIS-162 - provide global defaults and cli overridable options
+  g_Resources = std::make_shared<AsyncResources>(
+      server,  // InferenceServer*,
+      1,       // infer threads
+      1        // mgmt threads
+  );
+
+  LOG_INFO << "Building nvrpc server";
+  const std::string addr = "0.0.0.0:" + std::to_string(port);
+  grpc_server->reset(new GRPCServer(addr));
+
+  (*grpc_server)->GetBuilder().SetMaxMessageSize(MAX_GRPC_MESSAGE_SIZE);
+
+  LOG_INFO << "Register TensorRT GRPCService";
+  auto inferenceService = (*grpc_server)->RegisterAsyncService<GRPCService>();
+
+  if (endpoint_name == "infer"){
+    LOG_INFO << "Register Infer RPC";
+    (*grpc_server)->rpcInfer_ = inferenceService->RegisterRPC<InferContext>(
+        &GRPCService::AsyncService::RequestInfer);
+
+    LOG_INFO << "Register StreamInfer RPC";
+    (*grpc_server)->rpcStreamInfer_ =
+        inferenceService->RegisterRPC<StreamInferContext>(
+            &GRPCService::AsyncService::RequestStreamInfer);
+  }
+
+  if (endpoint_name == "status"){
+    LOG_INFO << "Register Status RPC";
+    (*grpc_server)->rpcStatus_ = inferenceService->RegisterRPC<StatusContext>(
+        &GRPCService::AsyncService::RequestStatus);
+  }
+
+  if (endpoint_name == "profile"){
+    LOG_INFO << "Register Profile RPC";
+    (*grpc_server)->rpcProfile_ = inferenceService->RegisterRPC<ProfileContext>(
+        &GRPCService::AsyncService::RequestProfile);
+  }
+
+  if (endpoint_name == "health"){
+    LOG_INFO << "Register Health RPC";
+    (*grpc_server)->rpcHealth_ = inferenceService->RegisterRPC<HealthContext>(
+        &GRPCService::AsyncService::RequestHealth);
+  }
+
+  return Status::Success;
+}
+
+Status
 GRPCServer::Start()
 {
   if (!running_) {
--- a/src/servers/http_server.cc
+++ b/src/servers/http_server.cc
@@ -47,7 +47,7 @@ namespace nvidia { namespace inferenceserver {
 class HTTPServerImpl : public HTTPServer {
  public:
   explicit HTTPServerImpl(
-      InferenceServer* server, uint16_t port, int thread_cnt)
+      InferenceServer* server, std::string endpoint_name, uint16_t port, int thread_cnt)
       : server_(server), port_(port), thread_cnt_(thread_cnt),
         api_regex_(R"(/api/(health|profile|infer|status)(.*))"),
         health_regex_(R"(/(live|ready))"),
@@ -57,7 +57,7 @@ class HTTPServerImpl : public HTTPServer {
 
   ~HTTPServerImpl() { Stop(); }
 
-  static void Dispatch(evhtp_request_t* req, void* arg);
+  static void Dispatch(evhtp_request_t* req, std::string endpoint_name, void* arg);
 
   Status Start() override;
   Status Stop() override;
@@ -171,13 +171,13 @@ HTTPServerImpl::StopCallback(int sock, short events, void* arg)
 }
 
 void
-HTTPServerImpl::Dispatch(evhtp_request_t* req, void* arg)
+HTTPServerImpl::Dispatch(evhtp_request_t* req, std::string endpoint_name, void* arg)
 {
-  (static_cast<HTTPServerImpl*>(arg))->Handle(req);
+  (static_cast<HTTPServerImpl*>(arg))->Handle(req, endpoint_name);
 }
 
 void
-HTTPServerImpl::Handle(evhtp_request_t* req)
+HTTPServerImpl::Handle(evhtp_request_t* req, std::string endpoint_name)
 {
   LOG_VERBOSE(1) << "HTTP request: " << req->method << " "
                  << req->uri->path->full;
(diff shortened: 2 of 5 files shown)


Makefile.client
```diff
@@ -146,7 +146,8 @@ grpc: $(LIBGRPCDIR)/libgrpc_indicator
 
 $(LIBGRPCDIR)/libgrpc_indicator:
 	mkdir -p $(LIBGRPCDIR)
-	git clone -b $(shell curl -L https://grpc.io/release) \
+  # temporarily set to 1.19 due to build problems with 1.20
```
deadeyegoodwin: This is going in as part of https://github.com/NVIDIA/tensorrt-inference-server/pull/240 so need to remove it from here before merging.

src/servers/grpc_server.cc
```diff
@@ -292,6 +292,59 @@ GRPCServer::Create(
   return Status::Success;
 }
 
+Status
+GRPCServer::CreateUniqueEndpointPorts(
+    InferenceServer* server, std::string endpoint_name, uint16_t port,
+    std::unique_ptr<GRPCServer>* grpc_server)
+{
+  // DLIS-162 - provide global defaults and cli overridable options
```
deadeyegoodwin: Don't include the DLIS since that is our internal tracking and so not useful to GitHub users.

src/servers/grpc_server.cc
```diff
@@ -292,6 +292,59 @@ GRPCServer::Create(
...
+  g_Resources = std::make_shared<AsyncResources>(
+      server,  // InferenceServer*,
```
deadeyegoodwin: We don't use this commenting style. Put a comment before the statement if necessary. For constants like the 1's use an inline comment, ... 1 /* infer threads */, ...

src/servers/main.cc
```diff
@@ -74,6 +76,15 @@ std::unique_ptr<prometheus::Exposer> exposer_;
...
+// Unique Ports for Health, Status, Profile and Infer Requests on HTTP and GRPC servers
+int32_t http_status_port_ = -1;
+int32_t http_health_port_ = -1;
+int32_t http_profile_port_ = -1;
+int32_t http_infer_port_ = -1;
+int32_t grpc_status_port_ = -1;
+int32_t grpc_health_port_ = -1;
+int32_t grpc_profile_port_ = -1;
+int32_t grpc_infer_port_ = -1;
 // The metric port. Initialized to default values and modifyied based
```
deadeyegoodwin: Blank line here

src/servers/main.cc
```diff
@@ -104,6 +115,14 @@ enum OptionId {
   OPTION_ALLOW_GPU_METRICS,
   OPTION_GRPC_PORT,
   OPTION_HTTP_PORT,
+  OPTION_HTTP_STATUS_PORT,
```
deadeyegoodwin: maintain existing convention, GRPC first followed by HTTP.

src/servers/main.cc
```diff
@@ -247,31 +282,105 @@ StartHttpService(nvidia::inferenceserver::InferenceServer* server)
   return std::move(service);
 }
 
+std::unique_ptr<nvidia::inferenceserver::GRPCServer>
```
deadeyegoodwin: You need to "auto format" to fix these long lines. See https://docs.nvidia.com/deeplearning/sdk/tensorrt-inference-server-master-branch-guide/docs/contribute.html#coding-convention

src/servers/main.cc
```diff
@@ -247,31 +282,105 @@ StartHttpService(nvidia::inferenceserver::InferenceServer* server)
...
+  // Start normally if common port else spawn different servers for different endpoints
   // Enable gRPC endpoint if requested...
-  if (grpc_port_ != -1) {
-    LOG_INFO << " localhost:" << std::to_string(grpc_port_)
-             << " for gRPC requests";
-    grpc_service_ = StartGrpcService(server);
-    if (grpc_service_ == nullptr) {
-      LOG_ERROR << "Failed to start gRPC service";
-      return false;
+  if (grpc_status_port_ == grpc_health_port_ || grpc_health_port_ == grpc_profile_port_ || grpc_profile_port_ == grpc_infer_port_ || grpc_infer_port_ == grpc_status_port_) {
```
deadeyegoodwin: Parentheses around the sub-expressions in the conditional.

src/servers/main.cc
```diff
@@ -247,31 +282,105 @@ StartHttpService(nvidia::inferenceserver::InferenceServer* server)
...
+
+  return std::move(service);
+}
+
 bool
 StartEndpoints(nvidia::inferenceserver::InferenceServer* server)
 {
   LOG_INFO << "Starting endpoints, '" << server->Id() << "' listening on";
 
+  // Start normally if common port else spawn different servers for different endpoints
```
deadeyegoodwin: We're not sure if/what additional overhead is introduced by having multiple sevice instances (grpc or http) so I'd like to minimize as much as possible. So can you change the existing Create functions to take all 4 ports and be smart to look at those ports and start the minimum number of services needed. This will keep the same behavior as currently when all 4 ports are the same, but will (for example) only start one additional service is only one port is different.

src/servers/main.cc
```diff
@@ -63,7 +63,9 @@ bool exit_on_failed_init_ = true;
 
 // The HTTP and GRPC services
 std::unique_ptr<nvidia::inferenceserver::HTTPServer> http_service_;
+std::unique_ptr<nvidia::inferenceserver::HTTPServer> http_endpoint_services_[4];
```
deadeyegoodwin: Don't use "magic constant". Add a "constexpr size_t unique_service_endpoint_count_ = 4;" or something similar.

GuanLuo:
As a side node, if your local master branch is up to date with remote master,  you can use `git rebase master` to avoid the merge commit `Merge branch 'master' into ...`

src/servers/grpc_server.cc
```diff
@@ -249,14 +252,13 @@ GRPCServer::~GRPCServer()
...
-    InferenceServer* server, uint16_t port,
+    InferenceServer* server, int32_t port,
     std::unique_ptr<GRPCServer>* grpc_server)
 {
-  // DLIS-162 - provide global defaults and cli overridable options
   g_Resources = std::make_shared<AsyncResources>(
-      server,  // InferenceServer*,
-      1,       // infer threads
-      1        // mgmt threads
+      server, /* InferenceServer*, */
```
deadeyegoodwin: Closer, don't comment the server argument, only constants. I'm surprised the auto-formatter didn't but these on one line.

src/servers/main.cc
```diff
@@ -60,26 +60,44 @@ std::condition_variable exit_cv_;
 // exit if inference-server doesn't completely initialize (e.g. will
 // exit if even one model fails to load).
 bool exit_on_failed_init_ = true;
+constexpr size_t unique_service_endpoint_count_ = 4;
```
deadeyegoodwin: This doesn't really make sense anymore. It could be the maximum number of unique endpoints... but I don;t think we need that (see other comments).

src/servers/main.cc
```diff
@@ -427,6 +560,31 @@ Parse(nvidia::inferenceserver::InferenceServer* server, int argc, char** argv)
...
+      case OPTION_HTTP_HEALTH_PORT:
+        http_health_port = ParseIntOption(optarg);
+        break;
+      case OPTION_HTTP_PROFILE_PORT:
+        http_profile_port = ParseIntOption(optarg);
+        break;
+      case OPTION_HTTP_INFER_PORT:
+        http_infer_port = ParseIntOption(optarg);
+        break;
+      case OPTION_GRPC_STATUS_PORT:
```
deadeyegoodwin: These GPRC case statements should be with OPTION_GRPC_PORT

src/servers/main.cc
```diff
@@ -346,15 +472,22 @@ Parse(nvidia::inferenceserver::InferenceServer* server, int argc, char** argv)
...
-  bool allow_http = true;
-  bool allow_grpc = true;
   bool allow_metrics = true;
   bool allow_gpu_metrics = true;
   int32_t http_port = http_port_;
   int32_t grpc_port = grpc_port_;
   int32_t metrics_port = metrics_port_;
   int32_t http_thread_cnt = http_thread_cnt_;
 
+  int32_t http_status_port = -1;
```
deadeyegoodwin: To get the correct in-order command-line parsing behavior you need to initialize these http_port_ or grpc_port_. Then in the case statements update them appropriately. For example if --http-port=-1 is parsed then you need to set all 4 http port variables to -1. After this first initialization you shouldn't use http_port_ or grpc_port_ anywhere else.

src/servers/grpc_server.cc
```diff
@@ -292,6 +294,82 @@ GRPCServer::Create(
   return Status::Success;
 }
 
+Status
+GRPCServer::CreateUniqueEndpointPorts(
```
deadeyegoodwin: You shouldn't need to create these new functions (or in main.cc). Just following the existing flow but pass in 4 ports instead of 1. Also, use std::vector for endpoint_names, endpoint_ports, and grpc_servers. We typically avoid raw arrays because they are not as safe.

grpc_servers should be std::vector<unique_ptr<GRPServer>>* since when returning values we typically pass in an address.

src/servers/grpc_server.cc
```diff
@@ -249,14 +252,13 @@ GRPCServer::~GRPCServer()
 
 Status
 GRPCServer::Create(
```
deadeyegoodwin: Create with all 4 ports and put CreateUniqueEndpointPorts logic here.

src/servers/main.cc
```diff
@@ -60,26 +60,44 @@ std::condition_variable exit_cv_;
 // exit if inference-server doesn't completely initialize (e.g. will
 // exit if even one model fails to load).
 bool exit_on_failed_init_ = true;
+constexpr size_t unique_service_endpoint_count_ = 4;
 
 // The HTTP and GRPC services
-std::unique_ptr<nvidia::inferenceserver::HTTPServer> http_service_;
 std::unique_ptr<nvidia::inferenceserver::GRPCServer> grpc_service_;
```
deadeyegoodwin: Why are there still the individual http_server_ and grpc_server_?

src/servers/main.cc
```diff
@@ -60,26 +60,44 @@ std::condition_variable exit_cv_;
...
+std::unique_ptr<nvidia::inferenceserver::HTTPServer> http_service_;
+std::unique_ptr<nvidia::inferenceserver::GRPCServer>
```
deadeyegoodwin: use std::vector<std::unque_ptr<GRPCServer>>

src/servers/main.cc
CoderHam: This is for when the http-port is set to some value. If it is set to -1 then the multi-port option is triggered

src/servers/grpc_server.cc
CoderHam: Will do

src/servers/main.cc
CoderHam: It is for the four unique (or common) ports for status, health etc

deadeyegoodwin: Yes, but given what we talked about yesterday they will not each (necessarily) have their own service. The common case (as it is today) is that all 4 will share the same service.

src/servers/main.cc
deadeyegoodwin: Don't think about 2 different modes. There is only one mode that has 4 inputs (the ports). There are 5 command-line flags (plus the allow flag) that are used to initialize those 4 inputs.

CoderHam: Alright. Will combine the functionality to support the above.

src/servers/main.cc
CoderHam: Got it

src/servers/grpc_server.cc
```diff
@@ -249,46 +252,75 @@ GRPCServer::~GRPCServer()
...
+          LOG_INFO << "Register Health RPC";
+          (*grpc_servers)[i]->rpcHealth_ =
+              inferenceService->RegisterRPC<HealthContext>(
+                  &GRPCService::AsyncService::RequestHealth);
+        }
+      }
+      i++;
+    }
+  } else {
+    return Status(RequestStatusCode::INVALID_ARG, "must specify valid ports");
```
deadeyegoodwin: Better error message: "GRPC is enabled but none of the service endpoints have a valid port assignment"

src/servers/http_server.cc
```diff
@@ -553,10 +566,34 @@ HTTPServerImpl::InferRequest::FinalizeResponse()
...
+      std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+      LOG_INFO << "Starting HTTPService at " << addr;
+      ((*http_servers)[i])
+          .reset(new HTTPServerImpl(
+              server, ep_map.second, ep_map.first, thread_cnt));
+      i++;
+    }
+  } else {
+    return Status(
+        RequestStatusCode::INVALID_ARG, "must specify ports if http-port -1");
```
deadeyegoodwin: Better error message...

src/servers/main.cc
```diff
@@ -213,63 +253,100 @@ SignalHandler(int signum)
...
+    std::vector<int32_t> endpoint_ports)
 {
-  std::unique_ptr<nvidia::inferenceserver::GRPCServer> service;
   nvidia::inferenceserver::Status status =
-      nvidia::inferenceserver::GRPCServer::Create(server, grpc_port_, &service);
+      nvidia::inferenceserver::GRPCServer::Create(
+          server, endpoint_names, endpoint_ports, &grpc_endpoint_services_);
   if (status.IsOk()) {
-    status = service->Start();
+    for (auto&& grpc_eps : grpc_endpoint_services_) {
```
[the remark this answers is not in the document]
CoderHam: That was a typo. Sorry.

src/servers/http_server.cc
```diff
@@ -553,10 +566,27 @@ HTTPServerImpl::InferRequest::FinalizeResponse()
 
 Status
 HTTPServer::Create(
-    InferenceServer* server, uint16_t port, int thread_cnt,
-    std::unique_ptr<HTTPServer>* http_server)
+    InferenceServer* server,
+    std::map<int32_t, std::vector<std::string>> port_map, int thread_cnt,
```
deadeyegoodwin: const&

src/servers/http_server.cc
```diff
@@ -553,10 +566,27 @@ HTTPServerImpl::InferRequest::FinalizeResponse()
...
+    for (auto const& ep_map : port_map) {
+      std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+      LOG_INFO << "Starting HTTPService at " << addr;
+      ((*http_servers)[i])
+          .reset(new HTTPServerImpl(
+              server, ep_map.second, ep_map.first, thread_cnt));
+      i++;
+    }
+  } else {
+    return Status(
```
deadeyegoodwin: Check move above under port_map.empty() and avoid having else.

src/servers/http_server.cc
```diff
@@ -553,10 +566,27 @@ HTTPServerImpl::InferRequest::FinalizeResponse()
...
+    std::vector<std::unique_ptr<HTTPServer>>* http_servers)
 {
-  http_server->reset(new HTTPServerImpl(server, port, thread_cnt));
+  size_t i = 0;
+
+  if (!port_map.empty()) {
+    for (auto const& ep_map : port_map) {
+      std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+      LOG_INFO << "Starting HTTPService at " << addr;
+      ((*http_servers)[i])
```
deadeyegoodwin: Same comment as above, don't require caller to pre-allocator http_servers. Clear it and push_back.

src/servers/http_server.h
```diff
@@ -34,8 +34,10 @@ class InferenceServer;
 class HTTPServer {
  public:
   static Status Create(
-      InferenceServer* server, uint16_t port, int thread_cnt,
-      std::unique_ptr<HTTPServer>* http_server);
+      InferenceServer* server,
+      std::map<int32_t, std::vector<std::string>> port_map, int thread_cnt,
```
deadeyegoodwin: const&

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
   exit_cv_.notify_all();
 }
 
-std::unique_ptr<nvidia::inferenceserver::GRPCServer>
-StartGrpcService(nvidia::inferenceserver::InferenceServer* server)
+nvidia::inferenceserver::Status
+StartMultipleGrpcService(
+    nvidia::inferenceserver::InferenceServer* server,
+    std::map<int32_t, std::vector<std::string>> port_map)
```
deadeyegoodwin: const&

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
-  return std::move(service);
+  return status;
 }
 
-std::unique_ptr<nvidia::inferenceserver::HTTPServer>
-StartHttpService(nvidia::inferenceserver::InferenceServer* server)
+nvidia::inferenceserver::Status
+StartMultipleHttpService(
+    nvidia::inferenceserver::InferenceServer* server,
+    std::map<int32_t, std::vector<std::string>> port_map)
```
deadeyegoodwin: const&

src/servers/grpc_server.cc
```diff
@@ -249,45 +250,71 @@ GRPCServer::~GRPCServer()
...
-
-  LOG_INFO << "Register Profile RPC";
-  (*grpc_server)->rpcProfile_ = inferenceService->RegisterRPC<ProfileContext>(
-      &GRPCService::AsyncService::RequestProfile);
-
-  LOG_INFO << "Register Health RPC";
-  (*grpc_server)->rpcHealth_ = inferenceService->RegisterRPC<HealthContext>(
-      &GRPCService::AsyncService::RequestHealth);
+      server, 1 /* infer threads */, 1 /* mgmt threads */);
+  (*grpc_servers).clear();
```
deadeyegoodwin: Why not use grpc_servers->clear()? And similar changes below...

src/servers/grpc_server.cc
```diff
@@ -249,45 +250,71 @@ GRPCServer::~GRPCServer()
...
+
+  for (auto const& ep_map : port_map) {
+    std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+    LOG_INFO << "Starting a GRPCService at " << addr;
+
+    (*grpc_servers).push_back(nullptr);
+    ((*grpc_servers).back()).reset(new GRPCServer(addr));
```
deadeyegoodwin: It will be easier to read if instead of using back() everywhere below you just use:
auto& grpc_server = grpc_servers->back();

src/servers/grpc_server.cc
```diff
@@ -249,45 +250,71 @@ GRPCServer::~GRPCServer()
...
-  (*grpc_server)->rpcHealth_ = inferenceService->RegisterRPC<HealthContext>(
-      &GRPCService::AsyncService::RequestHealth);
+      server, 1 /* infer threads */, 1 /* mgmt threads */);
+  (*grpc_servers).clear();
+
+  for (auto const& ep_map : port_map) {
+    std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+    LOG_INFO << "Starting a GRPCService at " << addr;
+
+    (*grpc_servers).push_back(nullptr);
```
deadeyegoodwin: grpc_servers->push_back(new GRPCServer(addr))  ?
Might need to use emplace_back since it is unique_ptr

src/servers/http_server.cc
```diff
@@ -41,14 +42,17 @@
...
+
 //
 // Handle HTTP requests
-//
 class HTTPServerImpl : public HTTPServer {
  public:
   explicit HTTPServerImpl(
-      InferenceServer* server, uint16_t port, int thread_cnt)
-      : server_(server), port_(port), thread_cnt_(thread_cnt),
+      InferenceServer* server, std::vector<std::string> endpoints, int32_t port,
```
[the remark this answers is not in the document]
deadeyegoodwin: You missed const& on "endpoints"

src/servers/http_server.cc
```diff
@@ -553,10 +563,26 @@ HTTPServerImpl::InferRequest::FinalizeResponse()
...
+    std::vector<std::unique_ptr<HTTPServer>>* http_servers)
 {
-  http_server->reset(new HTTPServerImpl(server, port, thread_cnt));
+  if (port_map.empty()) {
+    return Status(
+        RequestStatusCode::INVALID_ARG,
+        "HTTP is enabled but none of the service endpoints have a valid port "
+        "assignment");
+  }
+  (*http_servers).clear();
```
deadeyegoodwin: http_severs->clear()

src/servers/http_server.cc
```diff
@@ -553,10 +563,26 @@ HTTPServerImpl::InferRequest::FinalizeResponse()
...
+  for (auto const& ep_map : port_map) {
+    std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+    LOG_INFO << "Starting HTTPService at " << addr;
+    (*http_servers).push_back(nullptr);
```
deadeyegoodwin: Use emplace_back as above?

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
-  return std::move(service);
+  return status;
 }
 
-std::unique_ptr<nvidia::inferenceserver::HTTPServer>
-StartHttpService(nvidia::inferenceserver::InferenceServer* server)
+nvidia::inferenceserver::Status
+StartMultipleHttpService(
+    nvidia::inferenceserver::InferenceServer* server,
+    std::map<int32_t, std::vector<std::string>>& port_map)
```
deadeyegoodwin: const-reference, not just reference

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
 
-  return std::move(service);
+  return status;
 }
 
 bool
 StartEndpoints(nvidia::inferenceserver::InferenceServer* server)
 {
+  size_t i;
+  std::map<int32_t, std::vector<std::string>> port_map;
```
deadeyegoodwin: If you declare port_map here isn't it being used for both grpc and http? And then won't it have grpc entries in it when being used for http? Move the declaration down into "allow_grpc" and "allow_http" conditionals.

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
 
-  // Enable gRPC endpoint if requested...
-  if (grpc_port_ != -1) {
-    LOG_INFO << " localhost:" << std::to_string(grpc_port_)
-             << " for gRPC requests";
-    grpc_service_ = StartGrpcService(server);
-    if (grpc_service_ == nullptr) {
+  // Enable gRPC endpoints if requested...
+  if (allow_grpc) {
+    std::vector<std::string> endpoint_names = {"status", "health", "profile",
```
deadeyegoodwin: Just write:
port_map[grpc_status_port_] = "status";
...

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
+                                           grpc_infer_port_};
+
+    // Group by port numbers
+    for (i = 0; i < endpoint_ports.size(); i++) {
+      if (endpoint_ports[i] != -1) {
+        port_map[endpoint_ports[i]].push_back(endpoint_names[i]);
+      }
+    }
+
+    for (i = 0; i < port_map.size(); i++) {
```
deadeyegoodwin: What is this for?

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
     }
   }
 
-  // Enable HTTP endpoint if requested...
-  if (http_port_ != -1) {
-    LOG_INFO << " localhost:" << std::to_string(http_port_)
-             << " for HTTP requests";
+  // Enable HTTP endpoints if requested...
+  if (allow_http) {
+    std::vector<std::string> endpoint_names = {"status", "health", "profile",
```
deadeyegoodwin: Just write:
port_map[http_status_port_] = "status";
...

src/servers/main.cc
```diff
@@ -212,64 +252,115 @@ SignalHandler(int signum)
...
+                                           http_infer_port_};
+
+    // Group by port numbers
+    for (i = 0; i < endpoint_ports.size(); i++) {
+      if (endpoint_ports[i] != -1) {
+        port_map[endpoint_ports[i]].push_back(endpoint_names[i]);
+      }
+    }
+
+    for (i = 0; i < port_map.size(); i++) {
```
deadeyegoodwin: What is this for?

src/servers/main.cc
```diff
@@ -469,21 +600,37 @@ Parse(nvidia::inferenceserver::InferenceServer* server, int argc, char** argv)
     LOG_ERROR << "At least one of the following options must be true: "
               << "--allow-http, --allow-grpc";
     return false;
-  } else if (allow_http && allow_grpc && (http_port == grpc_port)) {
+  } else if (
+      allow_http && allow_grpc &&
+      ((http_port == grpc_port) && ((http_port != -1) && (grpc_port != -1)))) {
```
deadeyegoodwin: These checks are not really sufficient anymore. Based on the error messages is it clear what actually needs to be checked. If not we can discuss...

src/servers/main.cc
CoderHam: I need to iterate through all ports and group based on port numbers. I also need to filter out ports that are set to -1.

src/servers/main.cc
CoderHam: No longer needed since this is done in grpc/http_server.cc now

src/servers/main.cc
CoderHam: redundant. removed

src/servers/main.cc
CoderHam: Moved these tests and added more to CheckPortCollision()

deadeyegoodwin:
@CoderHam You need to manually rebase this PR onto master and resolve the conflicts.

PR #242, opened 2019-04-20
perf_client report percentile latency and use it as stable criteria

GuanLuo:
sample output:
```
./perf_client -v -d -i grpc -u localhost:8001 -m resnet_v1_50_graphdef -p20000 -b1 -t4 --percentile 99

*** Measurement Settings ***
  Batch size: 1
  Measurement window: 20000 msec
  Latency limit: 0 msec

Request concurrency: 4
  Pass [1] throughput: 195 infer/sec. 99-th percentile latency: 32811 usec
  Pass [2] throughput: 237 infer/sec. 99-th percentile latency: 24978 usec
  Pass [3] throughput: 237 infer/sec. 99-th percentile latency: 24965 usec
  Pass [4] throughput: 235 infer/sec. 99-th percentile latency: 25492 usec
  Client: 
    Request count: 4712
    Throughput: 235 infer/sec
    99-th percentile latency: 25492 usec
    Avg gRPC time: 0 usec (marshal 0 usec + response wait 0 usec + unmarshal 0 usec)
  Server: 
    Request count: 5652
    Avg request latency: 16355 usec (overhead 20 usec + queue 5209 usec + compute 11126 usec)

[ 0] SUCCESS
Inferences/Second vs. Client Average Batch Latency
Concurrency: 4, 235 infer/sec, latency 16960 usec
```

PR #243, opened 2019-04-22
Fix Client Infer Timer. Returning reference of the timer for starting / stopping the timer

GuanLuo:


src/clients/c++/request_common.h
```diff
@@ -331,7 +331,7 @@ class RequestImpl : public InferContext::Request {
   bool IsReady() const { return ready_; }
   void SetIsReady(bool r) { ready_ = r; }
 
-  RequestTimers Timer() const { return timer_; }
+  RequestTimers& Timer() { return timer_; }
```
deadeyegoodwin: Why is this change needed all by itself?

GuanLuo: The timer is used to record time like the following:
`current_context->Timer().Record(RequestTimers::Kind::REQUEST_START);`
The original function call returns a copy of the timer, and thus the timer doesn't get set properly. See #242 description, gRPC time is all zeros.

deadeyegoodwin: I don't see any description in that other pull request but perhaps I'm looking in the wrong place. We should have some perf_client testing that makes sure that the timers don't just return all zeros. Can you add that to this PR.

PR #245, opened 2019-04-23
adding tests for multi-port options

CoderHam:


Makefile.client
```diff
@@ -146,7 +146,10 @@ grpc: $(LIBGRPCDIR)/libgrpc_indicator
 
 $(LIBGRPCDIR)/libgrpc_indicator:
 	mkdir -p $(LIBGRPCDIR)
-	git clone -b $(shell curl -L https://grpc.io/release) \
+	# Build gRPc v1.19.0 since it is a known good build. Change to
```
deadeyegoodwin: Remove this file from your commit. Just rebase to master and you will pick up the change.

Dockerfile.client
```diff
@@ -88,6 +88,7 @@ RUN mkdir -p /tmp/client/bin && \
     cp src/clients/python/ensemble_image_client.py /tmp/client/python/. && \
     cp src/clients/python/grpc_image_client.py /tmp/client/python/. && \
     cp src/clients/python/simple_client.py /tmp/client/python/. && \
+    cp src/clients/python/multi_port_client.py /tmp/client/python/. && \
```
deadeyegoodwin: src/clients/... is not for test clients. It is for clients that have some more general use or that are used for examples. So just rename it to multi_port_test.py and put it in L0_multiple_ports

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,89 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# MULTI_PORT_CLIENT=../clients/multi_port_client
```
deadeyegoodwin: Just delete this. We don't need to test multiple ports with c++ client since it is not API related.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,89 @@
...
+CLIENT_LOG="./client.log"
+
+DATADIR=`pwd`/models
+HSP_ARR=(8008 8009 8010 8011)
+HHP_ARR=(8008 8010 8010 8009)
+HPP_ARR=(8008 8011 8011 8010)
+HIP_ARR=(8008 8011 8008 8008)
+SERVER=/opt/tensorrtserver/bin/trtserver
+for (( n=0; n<4; n++ ))
+do
```
deadeyegoodwin: This is strange formatting, and the comment seems incorrect and doesn't really say anything anyway.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,89 @@
...
+:
+# do whatever on $i
+  SERVER_ARGS_ADD_HTTP="--http-status-port ${HSP_ARR[n]} --http-health-port ${HHP_ARR[n]} --http-profile-port ${HPP_ARR[n]}\
+    --http-infer-port ${HIP_ARR[n]} --http-port -1"
+  SERVER_ARGS_ADD_GRPC="--grpc-status-port $((${HSP_ARR[n]}+10)) --grpc-health-port $((${HHP_ARR[n]}+10)) \
+    --grpc-profile-port $((${HPP_ARR[n]}+10)) --grpc-infer-port $((${HIP_ARR[n]}+10)) --grpc-port -1"
+  SERVER_ARGS="--model-store=$DATADIR $SERVER_ARGS_ADD_GRPC $SERVER_ARGS_ADD_HTTP"
+  SERVER_LOG="./inference_server.log"
+  SERVER_TIMEOUT=20
```
deadeyegoodwin: Don't set SERVER_TIMEOUT unless the test requires it specifically. The default value is set to account for some slow environments where we run manual QA (cloud service providers).

src/clients/python/multi_port_client.py
```diff
@@ -0,0 +1,117 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--verbose', action="store_true", required=False, default=False,
+                        help='Enable verbose output')
+    parser.add_argument('-u', '--url', type=str, required=False, default='localhost:',
+                        help='Inference server URL. Default is localhost:.')
+    parser.add_argument('-i', '--protocol', type=str, required=False, default='http',
+                        help='Protocol ("http"/"grpc") used to ' +
+                        'communicate with inference service. Default is "http".')
+    parser.add_argument(-'hsp', "http_status_port", type=int,
```
deadeyegoodwin: Why 'http' in the name when these are used for both http and grpc? Why not juse --status_port?

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,89 @@
...
+HSP_ARR=(8008 8009 8010 8011)
+HHP_ARR=(8008 8010 8010 8009)
+HPP_ARR=(8008 8011 8011 8010)
+HIP_ARR=(8008 8011 8008 8008)
+SERVER=/opt/tensorrtserver/bin/trtserver
+for (( n=0; n<4; n++ ))
+do
+:
+# do whatever on $i
+  SERVER_ARGS_ADD_HTTP="--http-status-port ${HSP_ARR[n]} --http-health-port ${HHP_ARR[n]} --http-profile-port ${HPP_ARR[n]}\
```
deadeyegoodwin: We need to discuss the cases that need testing. It seems you are missing a lot of important ones. Also, since the last arg is "--http-port -1" (and "--grpc-port -1") then aren't all the ports disabled and so the server shouldn't even start?

qa/L0_multiple_ports/multi_port_tests.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Create the inference context for the model.
+    if FLAGS.protocol == "http":
+        # infer
+        if FLAGS.infer_port !=-1:
+            ctx = InferContext(FLAGS.url+str(FLAGS.infer_port), protocol, model_name, model_version, FLAGS.verbose)
+        # health
+        if FLAGS.health_port !=-1:
+            hctx = ServerHealthContext(FLAGS.url+str(FLAGS.health_port), protocol, True)
+            assert hctx.is_ready() == False
```
deadeyegoodwin: Why do you expect the server to be not-ready and not-live? It should be both ready and live.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,92 @@
...
+SERVER=/opt/tensorrtserver/bin/trtserver
+len=${#SP_ARR[@]}
+
+for (( n=0; n<$len; n++ ))
+do
+:
+  SERVER_ARGS_ADD_GRPC="--grpc-status-port ${SP_ARR[n]} --grpc-health-port ${HP_ARR[n]} \
+    --grpc-profile-port ${PP_ARR[n]} --grpc-infer-port ${IP_ARR[n]} --allow-http 0"
+  SERVER_ARGS_ADD_HTTP="--http-status-port ${SP_ARR[n]} --http-health-port ${HP_ARR[n]} --http-profile-port ${PP_ARR[n]}\
+    --http-infer-port ${IP_ARR[n]} --allow-grpc 0"
```
deadeyegoodwin: Since you are disallowing both HTTP and GRPC isn't the server going to fail to start? I think you must not have all your changes in here to test all the use-cases we discussed.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,92 @@
...
+
+  run_server
+  if [ "$SERVER_PID" == "0" ]; then
+      echo -e "\n***\n*** Failed to start $SERVER\n***"
+      cat $SERVER_LOG
+      exit 1
+  fi
+
+  RET=0
+
```
deadeyegoodwin: You need a "set +e" here or else python returning non-zero will cause script to exit immediately.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,92 @@
...
+  python $MULTI_PORT_TESTS_PY -v >>$CLIENT_LOG 2>&1 -sp ${SP_ARR[n]} -hp ${HP_ARR[n]} -pp ${PP_ARR[n]} -ip ${IP_ARR[n]} -i grpc
+  if [ $? -ne 0 ]; then
+      RET=1
+  fi
+
```
deadeyegoodwin: set -e

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,92 @@
...
+  SERVER_ARGS="--model-store=$DATADIR $SERVER_ARGS_ADD_HTTP"
+
+  run_server
+  if [ "$SERVER_PID" == "0" ]; then
+      echo -e "\n***\n*** Failed to start $SERVER\n***"
+      cat $SERVER_LOG
+      exit 1
+  fi
+
```
deadeyegoodwin: set +e

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,92 @@
...
+  python $MULTI_PORT_TESTS_PY -v >>$CLIENT_LOG 2>&1 -sp ${SP_ARR[n]} -hp ${HP_ARR[n]} -pp ${PP_ARR[n]} -ip ${IP_ARR[n]}
+  if [ $? -ne 0 ]; then
+      RET=1
+  fi
+
```
deadeyegoodwin: set -e

qa/L0_multiple_ports/multi_port_tests.py
CoderHam: Changed both to True. That was an error

qa/L0_multiple_ports/test.sh
CoderHam: I have yet to add more test cases. Also I was testing HTTP +/ GRPC Individually hence the above. I can combine the tests

Dockerfile.QA
```diff
@@ -53,6 +53,9 @@ RUN mkdir -p qa/common && \
     mkdir -p qa/L0_simple_ensemble/models/simple/1 && \
     cp /opt/tensorrtserver/custom/libaddsub.so \
        qa/L0_simple_ensemble/models/simple/1/. && \
+    mkdir -p qa/L0_muliple_ports/models/simple/1 && \
```
deadeyegoodwin: multiple is misspelled

qa/L0_multiple_ports/multi_port_tests.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # We use a simple model that takes 2 input tensors of 16 integers
+    # each and returns 2 output tensors of 16 integers each. One
+    # output tensor is the element-wise sum of the inputs and one
+    # output is the element-wise difference.
+    model_name = "simple"
+    model_version = -1
+    batch_size = 1
+
+    # Create the inference context for the model.
```
deadeyegoodwin: This comment seems like from an outdated cur and paste since you are creating many contexts here not just infer

qa/L0_multiple_ports/multi_port_tests.py
```diff
@@ -0,0 +1,115 @@
...
+    # infer
+    if FLAGS.infer_port !=-1:
+        ctx = InferContext(FLAGS.url+str(FLAGS.infer_port), protocol, model_name, model_version, FLAGS.verbose)
```
deadeyegoodwin: Why is this separate from the rest of FLAGS.infer_port != -1 handling below?

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,152 @@
...
+PP_ARR=(8008 8011 8011 8010 8005 8005 -1 8006 -1 -1 8004 -1)
+IP_ARR=(8008 8011 8008 8008 8006 8006 8006 -1 -1 -1 -1 8004)
+SERVER=/opt/tensorrtserver/bin/trtserver
+len=${#SP_ARR[@]}
+
+# HTTP + GRPC w/o Interleaved
+for (( n=0; n<$len; n++ ))
+do
+:
+  SERVER_ARGS_ADD_GRPC="--grpc-status-port ${SP_ARR[n]} --grpc-health-port ${HP_ARR[n]} \
```
deadeyegoodwin: You are not testing --grpc-port=-1 (or non -1) mixed in with the specific port flags.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,152 @@
...
+
+  run_server
+  if [ "$SERVER_PID" == "0" ]; then
+      echo -e "\n***\n*** Failed to start $SERVER\n***"
+      cat $SERVER_LOG
+      exit 1
+  fi
+
+  set +e
+  python $MULTI_PORT_TESTS_PY -v >>$CLIENT_LOG 2>&1 -p 8005 -sp ${SP_ARR[n]} -hp ${HP_ARR[n]} -pp ${PP_ARR[n]} -ip ${IP_ARR[n]} -i grpc
```
deadeyegoodwin: This isn't interleaved, it is just -p at the beginning. You can probably just do a couple of very directed test cases here to cover what needs testing.

qa/L0_multiple_ports/test.sh
```diff
@@ -0,0 +1,152 @@
...
+SP_ARR=(8008 8009 8010 8011 -1 8004 8004 8004 8004 -1 -1 -1)
+HP_ARR=(8008 8010 8010 8009 8005 -1 8005 8005 -1 8004 -1 -1)
+PP_ARR=(8008 8011 8011 8010 8005 8005 -1 8006 -1 -1 8004 -1)
+IP_ARR=(8008 8011 8008 8008 8006 8006 8006 -1 -1 -1 -1 8004)
+SERVER=/opt/tensorrtserver/bin/trtserver
+len=${#SP_ARR[@]}
+
+# HTTP + GRPC w/o Interleaved
+for (( n=0; n<$len; n++ ))
+do
```
deadeyegoodwin: "do" and ":" on separate lines is strange formatting. Is the colon even needed? We would typically write:

for (( n=0; n<$len; n++ )) ; do

src/servers/main.cc
```diff
@@ -655,19 +655,16 @@ Parse(nvidia::inferenceserver::InferenceServer* server, int argc, char** argv)
...
-    http_infer_port_ = http_infer_port;
-  }
 
-  if (allow_grpc) {
-    grpc_status_port_ = grpc_status_port;
-    grpc_health_port_ = grpc_health_port;
-    grpc_profile_port_ = grpc_profile_port;
-    grpc_infer_port_ = grpc_infer_port;
-  }
+  http_status_port_ = http_status_port;
```
deadeyegoodwin: This part of the change seems outdated. I dont' think you need any changes to main.cc

CoderHam: Rebased with master to fix changes.

src/servers/main.cc
```diff
@@ -86,6 +86,7 @@ int32_t metrics_port_ = 8002;
 
 bool allow_http = true;
 bool allow_grpc = true;
+bool allow_metrics_ = true;
 
 // endpoint names for http/gRPC
 std::vector<std::string> endpoint_names = {"status", "health", "profile",
```
deadeyegoodwin: Should end in underscore like other globals? Why is this different?

deadeyegoodwin: You missed this one...

src/servers/main.cc
```diff
@@ -323,7 +325,7 @@ StartEndpoints(nvidia::inferenceserver::InferenceServer* server)
   LOG_INFO << "Starting endpoints, '" << server->Id() << "' listening on";
 
   // Enable gRPC endpoints if requested...
-  if (allow_grpc && grpc_port_!=-1) {
+  if (allow_grpc_ && grpc_port_ != -1) {
```
deadeyegoodwin: parens around sub-expression

deadeyegoodwin:
@CoderHam You need to manuallly rebase to master

PR #253, opened 2019-04-26
Expose gRPC (stream) infer context count as option

GuanLuo:


src/servers/main.cc
```diff
@@ -105,6 +111,8 @@ enum OptionId {
   OPTION_GRPC_PORT,
   OPTION_HTTP_PORT,
   OPTION_METRICS_PORT,
+  OPTION_GRPC_INFER_COUNT,
```
deadeyegoodwin: ...THREAD_COUNT

src/servers/main.cc
```diff
@@ -169,6 +177,10 @@ std::vector<Option> options_{
      "The port for the server to listen on for HTTP requests."},
     {OPTION_METRICS_PORT, "metrics-port",
      "The port reporting prometheus metrics."},
+    {OPTION_GRPC_INFER_COUNT, "grpc-infer-count",
```
deadeyegoodwin: grpc-infer-thread-count

src/servers/main.cc
```diff
@@ -169,6 +177,10 @@ std::vector<Option> options_{
...
+     "Number of threads handling GRPC infer requests."},
+    {OPTION_GRPC_STREAM_INFER_COUNT, "grpc-stream-infer-count",
```
deadeyegoodwin: grpc-stream-infer-thread-count

src/servers/main.cc
```diff
@@ -169,6 +177,10 @@ std::vector<Option> options_{
      "The port for the server to listen on for HTTP requests."},
     {OPTION_METRICS_PORT, "metrics-port",
      "The port reporting prometheus metrics."},
+    {OPTION_GRPC_INFER_COUNT, "grpc-infer-count",
+     "Number of threads handling GRPC infer requests."},
```
deadeyegoodwin: Spell out inferrence.
"... GRPC inference requests."
Same for stream option.

src/servers/main.cc
```diff
@@ -82,6 +82,12 @@ int metrics_port_ = 8002;
 // Should GPU metrics be reported.
 bool allow_gpu_metrics_ = false;
 
+// The number of threads to initialize for handling GRPC infer requests.
+int grpc_infer_cnt_ = 1000;
```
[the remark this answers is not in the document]
GuanLuo: I think around 100 will be good. The thread will not be released until the request is completed, so the value for this will be a bit tricky. Say if the value is set to 16, and there is a big model with max batch size 32 and dynamic batching. If the client send 32 batch-1 requests, only every 16 requests will be batched and processed.

In the future, we will need to decouple the gRPC thread's life cycle a bit more so that it is released once the request is transferred to the scheduler and another thread will pick up if the response is ready.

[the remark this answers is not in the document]
deadeyegoodwin: I see. Then leave them both at 1000 then since for some cases batching > 100 is used.

PR #256, opened 2019-04-27
Tests for proper response error when non-existent output name is passed

CoderHam:


GuanLuo:
You don't have to use image classification model in tests, if you look into the other test cases in qa directory, you will see that we tend to use "addsub" model, a custom light weight model that doesn't require too much pre/post-processing , to verify the correctness.

And using Python's unit test framework will be better.

qa/L0_output_name/output_name_test.py
```diff
@@ -0,0 +1,107 @@
...
+        # Prepare request for Status gRPC
+        request = grpc_service_pb2.StatusRequest(model_name=self.model_name)
+        response = grpc_stub.Status(request)
+
+        request_ = self.requestGenerator("DUMMY", FLAGS)
+        # Send request
+        response_ = grpc_stub.Infer(request_)
+        return str(response_).find("code: SUCCESS")==-1
+
+    def TestHTTP(self):
```
deadeyegoodwin: I don't see how you can test HTTP using grpc? Just using grpc (to avoid using our client library) is suffuicient, so you should need this test at all.

qa/L0_output_name/output_name_test.py
```diff
@@ -0,0 +1,107 @@
...
+import json
+import requests
+import unittest
+FLAGS = None
+
+class OutputNameValidationTest(unittest.TestCase):
+    def TestGRPC(self):
+        channel = grpc.insecure_channel(self.url)
+        grpc_stub = grpc_service_pb2_grpc.GRPCServiceStub(channel)
+        # Prepare request for Status gRPC
```
deadeyegoodwin: Why run a status request? Not need for this test AFAIK.

qa/L0_output_name/output_name_test.py
```diff
@@ -0,0 +1,107 @@
...
+        response = grpc_stub.Status(request)
+
+        input0_data = np.arange(start=0, stop=16, dtype=np.int32).tobytes()
+        headers = {'NV-InferRequest': 'batch_size: 1 input { name: "INPUT0" dims: 16} \
+                    output { name: "DUMMY"}} cls { count: 3 } }'}
+        # Send request
+        r = requests.post(url_, data=input0_data, headers=headers)
+        return r.status_code!=200
+
+    def requestGenerator(self, output_name, FLAGS):
```
deadeyegoodwin: Why is FLAGS argument all caps?

qa/L0_output_name/output_name_test.py
```diff
@@ -0,0 +1,107 @@
...
+        # Prepare request for Infer gRPC
+        # The meta data part can be reused across requests
+        request = grpc_service_pb2.InferRequest()
+        request.model_name = self.model_name
+        request.model_version = -1
+
+        request.meta_data.batch_size = 1
+        output_message = api_pb2.InferRequestHeader.Output()
+        output_message.name = output_name
+        output_message.cls.count = 1
```
deadeyegoodwin: I would not set cls.count since not all models (including the one you are using) provide class results.

qa/L0_output_name/output_name_test.py
```diff
@@ -0,0 +1,107 @@
...
+        channel = grpc.insecure_channel(self.url)
+        grpc_stub = grpc_service_pb2_grpc.GRPCServiceStub(channel)
+        # Prepare request for Status gRPC
+        request = grpc_service_pb2.StatusRequest(model_name=self.model_name)
+        response = grpc_stub.Status(request)
+
+        request_ = self.requestGenerator("DUMMY", FLAGS)
+        # Send request
+        response_ = grpc_stub.Infer(request_)
+        return str(response_).find("code: SUCCESS")==-1
```
deadeyegoodwin: You should check RequestStatus in the response more explicitly against the RequestStatusCode.

qa/L0_output_name/output_name_test.py
CoderHam: That was the convention followed in all the other tests

qa/L0_output_name/output_name_test.py
CoderHam: Agreed. I used that for testing. Removing it now

qa/L0_output_name/output_name_test.py
CoderHam: Should I remove the HTTP test entirely then?

deadeyegoodwin: yes

PR #265, opened 2019-05-02
Improve ModelRepositoryManager to replace TFS functionalities

GuanLuo:


GuanLuo:
Will create a separate PR for clean up as mentioned in TODOs.

- Change `ServerbleBundle` classes to be just `ServerbleBackend`
- Add function to update backend state in `ServerStatusManager` so that `ServerStatusManager` doesn't need to be aware of `ModelRepositoryManager`

deadeyegoodwin:
You need to manually rebase.
Also, don't use "servable" in any part of a name, that is a TFS-ism.
Use BackendFactory instead of ...SourceAdapter...
Rename servables directory to backends.
Of course do all that in a separate change after we get this one through

src/core/model_repository_manager.cc
```diff
@@ -257,92 +226,565 @@ IsModified(const std::string& path, int64_t* last_ns)
...
+  // [TODO] fix InferenceBackend destructor
+  // The expectation is that the InferenceBackend can be released directly when
+  // ~BackendHandleImpl() is called. But in fact, it is not "ready" to be
+  // released if it just finished a request (Resource Deadlock error in
+  // Scheduler's destructor). Now we have to defer InferenceBackend destruction.
+
+  // Call callback with separate thread because the callback will try to acquire
+  // the mutex of the corresponding backend_info. And this destructor is likely
+  // to to be called within BackendLifeCycle class where the mutex is being hold
+  std::thread worker(std::move(OnDestroyBackend_), std::move(is_));
```
deadeyegoodwin: I think spawning a thread for every request is going to be too expensive...

PR #275, opened 2019-05-07
Add onnx runtime build

GuanLuo:


Dockerfile
```diff
@@ -31,6 +31,7 @@
 ARG BASE_IMAGE=nvcr.io/nvidia/tensorrtserver:19.04-py3
 ARG PYTORCH_IMAGE=nvcr.io/nvidia/pytorch:19.04-py3
 ARG TENSORFLOW_IMAGE=nvcr.io/nvidia/tensorflow:19.04-py3
+ARG CUDA_IMAGE=nvcr.io/nvidia/cuda:10.1-cudnn7-devel-ubuntu16.04
```
deadeyegoodwin: Use BASE_IMAGE and don't add CUDA_IMAGE. I know it looks like BASE_IMAGE is tensorrtserver but in our CI build it actually uses the correct cuda base image (which isn't available on nvcr.

PR #284, opened 2019-05-10
Add build instructions for Libtorch

CoderHam:
Build from GitHub source


Dockerfile
```diff
@@ -123,6 +123,26 @@ RUN python3 /workspace/onnxruntime/tools/ci_build/build.py --build_dir /workspac
             --update \
             --build
 
+############################################################################
```
deadeyegoodwin: I thought you could just do an extra build step in the caffe2 container above?

CoderHam: Yes. Moving the build to inside the caffe2 container

CoderHam: Solved the build error. The libtorch build failed because it was already in the pytorch container. Tested that and it was compatible.

Dockerfile
```diff
@@ -67,6 +67,8 @@ RUN cd pytorch && \
       NO_DISTRIBUTED=1 NO_TEST=1 NO_MIOPEN=1 USE_MKLDNN=0 USE_OPENCV=OFF USE_LEVELDB=OFF \
       python setup.py install && python setup.py clean
 
+# PS: No need to build LibTorch from source as it is already present in the docker
```
deadeyegoodwin: Remove this comment and instead in at line 59 add something like. "This will build both the caffe2 libraries needed by the caffe2 netdef backand and the libtorch library needed by the pytorch backend.

Dockerfile
```diff
@@ -177,6 +179,10 @@ COPY --from=trtserver_caffe2 /opt/conda/lib/libmkl_intel_lp64.so /opt/tensorrtse
 COPY --from=trtserver_caffe2 /opt/conda/lib/libmkl_rt.so /opt/tensorrtserver/lib/
 COPY --from=trtserver_caffe2 /opt/conda/lib/libmkl_vml_def.so /opt/tensorrtserver/lib/
 
+# LibTorch library
+COPY --from=trtserver_caffe2 /opt/conda/lib/python3.6/site-packages/torch/lib/libtorch.so* \
```
deadeyegoodwin: How many libtorch libraries are there? Why the wildcard?

CoderHam: libtorch.so and libtorch.so.1

Dockerfile
CoderHam: Will do

Dockerfile
deadeyegoodwin: You don't want to copy both of those since one is just a link and then you will end up with two copies of the file. Instead you want to copy one and create the other as a link. You can do this in the docker file, ask Guan for help if you need it since it has done it several times.

CoderHam: Sure :)

PR #286, opened 2019-05-11
Add procedure for execution. Add skeleton for input/output conversion.

GuanLuo:


src/backends/onnx/onnx_backend.cc
```diff
@@ -276,7 +299,171 @@ OnnxBackend::Context::Run(
...
+  // total_batch_size can be 1 for models that don't support batching
+  // (i.e. max_batch_size_ == 0).
+  if ((total_batch_size != 1) && (total_batch_size > (size_t)max_batch_size_)) {
+    return Status(
+        RequestStatusCode::INTERNAL,
+        "dynamic batch size " + std::to_string(total_batch_size) + " for '" +
+            name_ + "', max allowed is " + std::to_string(max_batch_size_));
+  }
+
+  // Hold reference to each buffer of input data to that it stays
```
deadeyegoodwin: "[so] that is stays"

PR #319, opened 2019-05-31
Fix fetch_models.sh for kubeflow example setup

CoderHam:
**TRTIS r19.05 fetch_models issue may block kubeflow example setup.** 
Adding back the custom models download code for `custom_models.tar.gz`

### DIFF
--- a/docs/examples/fetch_models.sh
+++ b/docs/examples/fetch_models.sh
@@ -41,10 +41,13 @@ wget -O /tmp/inception_v3_2016_08_28_frozen.pb.tar.gz \
 (cd /tmp && tar xzf inception_v3_2016_08_28_frozen.pb.tar.gz)
 mv /tmp/inception_v3_2016_08_28_frozen.pb model_repository/inception_graphdef/1/model.graphdef
 
-# Custom models. Only the source code is provided, users must build
+# Custom models. The source code is also provided, users may also build
 # the model definition files in order to use them (see "Building" section in
 # the documentation).
 mkdir -p model_repository/image_preprocess_nchw_3x224x224_inception/1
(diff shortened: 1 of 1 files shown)


docs/examples/fetch_models.sh
```diff
@@ -41,10 +41,13 @@ wget -O /tmp/inception_v3_2016_08_28_frozen.pb.tar.gz \
 (cd /tmp && tar xzf inception_v3_2016_08_28_frozen.pb.tar.gz)
 mv /tmp/inception_v3_2016_08_28_frozen.pb model_repository/inception_graphdef/1/model.graphdef
 
-# Custom models. Only the source code is provided, users must build
+# Custom models. The source code is also provided, users may also build
 # the model definition files in order to use them (see "Building" section in
 # the documentation).
 mkdir -p model_repository/image_preprocess_nchw_3x224x224_inception/1
+wget -O /tmp/custom_models.tar.gz `curl https://api.github.com/repos/NVIDIA/tensorrt-inference-server/releases/latest | grep -oP -m 1 '(?<=browser_download_url\":\ \")https.*\.tar\.gz(?=\")'` | head -1
```
deadeyegoodwin: Did you test this with 19.05 release? We have 2 tar.gz files there now, one for 16.04 and one for 18.04. I guess you just take the first one...

GuanLuo: Why reverting this change e3c75f8d2f0627139c184f6fa12670247f126e96?

deadeyegoodwin: Good point. I think we no longer ship the custom backends as part of the client tar.gz.

deadeyegoodwin: I'm confused. I think change e3c75f8 never made it into 19,05.... perhaps it was done on master after 19.05 had already branched and we forgot to cherry-pick it? Anyway, @CoderHam can you verify that e3c75f8 is in the r19.06 branch.

PR #327, opened 2019-06-03
Generate models and test for LibTorch Sequence (+ variable) models

CoderHam:
- [x]  Generate **sequence** models (+variable)
- [x] Test **sequence** models (+ variable)

### DIFF
--- a/src/backends/pytorch/libtorch_backend.cc
+++ b/src/backends/pytorch/libtorch_backend.cc
@@ -553,12 +553,22 @@ LibTorchBackend::Context::Run(
             name_ + "', max allowed is " + std::to_string(max_batch_size_));
   }
 
+  // Additional inputs added to the provider...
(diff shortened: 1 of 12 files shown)


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -553,12 +553,22 @@ LibTorchBackend::Context::Run(
...
   std::vector<std::unique_ptr<char[]>> input_buffers;
 
+  size_t overide_inputs = 0;
+  if (input_override_map != nullptr) {
+    overide_inputs = input_override_map->size();
+  }
+
   // Store input and output tensors
-  std::vector<torch::jit::IValue> inputs_(input_request_provider->RequestHeader().input().size());
+  std::vector<torch::jit::IValue> inputs_(input_request_provider->RequestHeader().input().size() \
```
deadeyegoodwin: Don't put a '\' in C file. Just let the auto-formatter layout the code...

PR #332, opened 2019-06-04
Update documentation for PyTorch Backend

CoderHam:


docs/model_configuration.rst
```diff
@@ -158,6 +158,14 @@ minimal configuration above. Specifically:
   batching, the initial batch dimension must be variable-size for all inputs
   and outputs.
 
+* :ref:`PyTorch TorchScript <section-pytorch-models>` models have an optional
+  output configuration in the model configuration file to support cases where
+  there are variable number and/or datatypes of output. Both the input and
+  outputs in the configuration must follow a specific naming convention i.e.
+  "\<name\>__\<index\>". Where <index> is the index of the corresponding input or
```
deadeyegoodwin: What is <name> in the naming convention? Can it be any string I want to use?

docs/model_configuration.rst
```diff
@@ -158,6 +158,14 @@ minimal configuration above. Specifically:
...
+  output. If the model supports variable size input tensors, their dimensions
+  in the configuration must be -1 for all variable-size inputs and outputs.
```
deadeyegoodwin: Is this true? For other model types the model configuration doesn't need to be -1 if the underlying model is variable sized. It is ok for the model config to require a specific size even if the model itself could handle other sizes.

docs/model_configuration.rst
CoderHam: <name> can be any string but must be followed by "__<index>" + where <index is the corresponding position of the I/O>

docs/model_configuration.rst
CoderHam: The provider_utils.cc and provider.cc perform a check while creating the input/output map. 

provider_utils has a check to see if the number of bytes in the request is equal to the number of bytes expected and throws an error otherwise. The -1 is to inform that it is a variable size input.

Similarly provider requires a check to verify the o/p shape is the same as the expected shape and the 1 is to inform that it is a variable size o/p. 

This is not native to libtorch and should be consistent with all other backends.

docs/model_configuration.rst
deadeyegoodwin: You need to explain that better somewhere. Perhaps in model_configuration.rst at the top where it is talking about inputs and outputs. You could refer to that section from here.

docs/model_configuration.rst
deadeyegoodwin: Yes. But that is not what you said in your description. You can have a model that supports variable size in one or more dimension of one or more input or output tensors, but the model configuration doesn't need to be specified as -1 in (any or all of) those dimensions. That is, it is ok for the model configuration to be more restrictive then the actual model itself. The last paragraph in section "Model Configuration" says this. So I don't think you actually need to say anything specific for pytorch at all.

CoderHam: Ok. Will remove this

docs/model_configuration.rst
CoderHam: Ok. Will do

PR #334, opened 2019-06-04
Implement Prometheus Service Endpoint

GuanLuo:


deadeyegoodwin:
@GuanLuo You need to rebase to master and change to use the cmake build.

GuanLuo:
rebased

src/servers/http_server.cc
```diff
@@ -574,9 +627,25 @@ HTTPServer::Create(
...
     LOG_INFO << "Starting HTTPService at " << addr;
     http_servers->emplace_back(
-        new HTTPServerImpl(server, ep_map.second, ep_map.first, thread_cnt));
+        new HTTPAPIServer(server, ep_map.second, ep_map.first, thread_cnt));
   }
 
   return Status::Success;
 }
+
+Status
```
deadeyegoodwin: Can we change Create() to CreateAPIServer() to keep the function names consistent.

src/core/metrics.cc
```diff
@@ -367,6 +368,14 @@ Metrics::GetRegistry()
   return singleton->registry_;
 }
 
+const std::string
+Metrics::SerializedMetrics()
+{
+  auto singleton = Metrics::GetSingleton();
+  return singleton->serializer_->Serialize(
```
deadeyegoodwin: Is there a serializer that doesn't include all the comments in the outout?

GuanLuo: No, this is the only serializer provided by prometheus-cpp and there is no option to exclude the comments. To do so, we can either strip the comments out from the serialized metrics or provide our own serializer implementation.

PR #338, opened 2019-06-05
Make output optional for LibTorch model config

CoderHam:
- [x] Make the specification of Output in config.pbtxt optional for the LibTorch models
> This allows support for models with variable number of Outputs.
- [ ] Add tests for both Raw and Class based Output

**Note:** If no output is specified in model configuration all the outputs of the model are processed and made available. The client still needs to request for the output he wants. The output naming convention is **OUTPUT_0** ... **OUTPUT_<n-1>** when there are **n** outputs.

### DIFF
--- a/src/backends/pytorch/libtorch_backend.cc
+++ b/src/backends/pytorch/libtorch_backend.cc
@@ -115,8 +115,9 @@ ConvertTorchTypeToDataType(const torch::ScalarType& ttype)
       return TYPE_FP32;
     case torch::kDouble:
       return TYPE_FP64;
+    default:
+      return TYPE_FP32;
   }
-  return TYPE_FP32;
 }
 
 Status
--- a/src/backends/pytorch/libtorch_backend.cc
+++ b/src/backends/pytorch/libtorch_backend.cc
@@ -93,6 +93,32 @@ ConvertDataTypeToTorchType(const DataType& dtype)
   return std::make_pair(true, type);
 }
 
+DataType
+ConvertTorchTypeToDataType(const torch::ScalarType& ttype)
+{
+  switch (ttype) {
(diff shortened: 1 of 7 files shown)


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -605,6 +632,20 @@ LibTorchBackend::Context::Run(
 
   // Make sure each output is of the expected size and copy it into
   // the payload responses.
+
+  //  When no Output in Model Config:
```
deadeyegoodwin: The outputs that we want to read and return in the response should be determined by what is in the request, not what is in the model config. I don't think this code should even look at the model config, should it? The model config should be used to make sure that only allowed outputs are requested.

src/core/backend.h
```diff
@@ -121,6 +121,9 @@ class InferenceBackend {
 
   // Map from output name to the model configuration for that output.
   std::unordered_map<std::string, ModelOutput> output_map_;
+
+  // Platform name
+  std::string platform_name_;
```
deadeyegoodwin: platform name is availabe in 'config_' so don't need to have it here explicitly.

src/backends/pytorch/libtorch_backend.cc
CoderHam: Based on the behavior of the other backends I was under the assumption that all outputs in the model config are passed to the server to finalize the response. Only while finalizing the response does the server pick only the outputs that were requested for in the request header.

Yes, it would make sense to return only those outputs to the server that the user requested. The changes would have to be made to all backends for consistency.

src/core/backend.h
CoderHam: Removed

PR #342, opened 2019-06-05
 Add checking on number of versions to load

GuanLuo:


src/core/model_repository_manager.cc
```diff
@@ -871,22 +871,20 @@ ModelRepositoryManager::Create(
...
-    RETURN_IF_ERROR(
-        local_manager->status_manager_->InitForModel(name, model_config));
-
-    std::vector<int64_t> versions;
-    RETURN_IF_ERROR(
-        local_manager->VersionsToLoad(name, model_config, versions));
-    RETURN_IF_ERROR(local_manager->backend_life_cycle_->AsyncLoad(
-        name, versions, model_config));
+    // If there is error on model loading, just report it and move to next model
+    Status update_status = local_manager->Update(name, true);
```
deadeyegoodwin: Is this something that we should have caught in testing? Can you add a lifecycle test to make sure it doesn't regress?

src/core/model_repository_manager.cc
```diff
@@ -1179,6 +1191,14 @@ ModelRepositoryManager::VersionsToLoad(
     }
   }
 
+  if (versions.empty()) {
```
deadeyegoodwin: Add a lifecycle test for this

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -127,6 +127,56 @@ def test_parse_error_modelfail(self):
                 ex.message().startswith(
                     "Inference request for unknown model 'graphdef_float32_float32_float32'"))
 
+def test_parse_error_model_no_version(self):
```
deadeyegoodwin: Need to fix indent (see test failure)

PR #344, opened 2019-06-06
Allow TF signature key other than "serving_default"

GuanLuo:


src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -698,12 +698,25 @@ TRTISTF_ModelCreateFromSavedModel(
...
-    return TRTISTF_ErrorNew(
-        "unable to load model '" + std::string(model_name) + "', expected '" +
-        DEFAULT_SERVING_SIGNATURE_DEF_KEY + "' signature");
+    // If default serving signature_def key is not found, maybe it is named
+    // something else, use one that is neither init_op key nor train_op key
+    for (sig_itr = bundle->meta_graph_def.signature_def().begin();
+         sig_itr != bundle->meta_graph_def.signature_def().end(); sig_itr++) {
+      if ((sig_itr->first != INIT_OP_SIGNATURE_DEF_KEY) &&
+          (sig_itr->first != TRAIN_OP_SIGNATURE_DEF_KEY)) {
+        break;
```
deadeyegoodwin: Can you log a warning here saying "unable to find default serving signature xyz, using signature abc"

deadeyegoodwin: Replace "serving_default" with DEFAULT_SERVING_SIGNATURE_DEF_KEY as in the error message.

PR #345, opened 2019-06-06
Add TRTIS_ENABLE_GPU Check in LibTorch backend

CoderHam:


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -429,10 +434,15 @@ LibTorchBackend::Context::ReadFixedSizedOutputTensor(
...
         } else {
+#ifdef TRTIS_ENABLE_GPU
           cudaMemcpy(
               buffer, (char*)content + content_offset, expected_byte_size,
               cudaMemcpyDeviceToHost);
         }
+#else
+          return Status(
+              RequestStatusCode::INTERNAL, "GPU instances not supported");
+#endif  // TRTIS_ENABLE_GPU
```
GuanLuo: move this block up, seems to be in wrong scope

CoderHam: That's right. rectified.

PR #349, opened 2019-06-07
Fix output validation for LibTorch backend

CoderHam:
- [x] Extract model outputs only for the **Requested outputs** (Subset of those in model config).
- [x] Check **Datatype** of outputs after execution with that from model config.
- [x] Check **Shape** of outputs after execution with that from model config.
- [x] Check **Index** of outputs is valid for the number of outputs after execution.

### DIFF
--- a/src/backends/pytorch/libtorch_backend.cc
+++ b/src/backends/pytorch/libtorch_backend.cc
@@ -613,8 +613,21 @@ LibTorchBackend::Context::Run(
   // Run...
   RETURN_IF_ERROR(Execute(&inputs_, &outputs_));
 
-  // TODO Verify Output shape and number are valid after model execution
+  // TODO Verify Output shape and datatype after model execution
 
+  for (const auto& output : base->Config().output()) {
+    const std::string& name = output.name();
+    std::string index_str = name.substr(name.find(deliminator) + 2);
+    int op_index = std::atoi(index_str.c_str());
+    if ((op_index < 0) || (op_index >= outputs_.size())) {
+      return Status(
+          RequestStatusCode::INVALID_ARG,
+          "The output " + name +
+              " in the model config refers to an output index which doesn't "
+              "exist. This model has " +
+              std::to_string(outputs_.size() - 1) + " outputs");
+    }
+  }
   // Prepare set of Outputs requested for
   std::set<std::string> required_outputs;
   for (auto& payload : *payloads) {
--- a/src/backends/pytorch/libtorch_backend.cc
+++ b/src/backends/pytorch/libtorch_backend.cc
@@ -421,13 +421,31 @@ Status
 LibTorchBackend::Context::ReadFixedSizedOutputTensor(
     std::vector<torch::Tensor>* outputs_, const std::string& name,
     const int& op_index, const DataType dtype, const size_t dtype_byte_size,
-    const size_t total_batch_size, std::vector<Scheduler::Payload>* payloads)
+    const size_t total_batch_size, std::vector<Scheduler::Payload>* payloads,
+    const DimsList& dims)
 {
   std::vector<int64_t> content_shape;
   void* content = nullptr;
   size_t byte_size = 0;
   RETURN_IF_ERROR(GetOutputTensor(
       outputs_, op_index, dtype, &content, &byte_size, &content_shape));
+
+  // verify shape of output matches shape from model config
+  int8_t dim_i = 0;
+  if (max_batch_size_ != NO_BATCHING) {
+    dim_i++;
+  }
+  for (int i = 0; i < dims.size(); i++) {
+    if (dims[i] != -1) {
+      if (dims[i] != content_shape[i + dim_i]) {
+        return Status(
+            RequestStatusCode::INTERNAL,
+            "unexpected shape for output '" + name + "', dimension " +
+                std::to_string(content_shape[i + dim_i]) + " does not equal " +
+                std::to_string(dims[i]) + " specified in config");
+      }
+    }
+  }
 
   const size_t total_byte_size =
       GetElementCount(content_shape) * dtype_byte_size;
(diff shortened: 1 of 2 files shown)


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -613,12 +669,36 @@ LibTorchBackend::Context::Run(
...
   for (const auto& output : base->Config().output()) {
     const std::string& name = output.name();
     std::string index_str = name.substr(name.find(deliminator) + 2);
+    int op_index = std::atoi(index_str.c_str());
+    int max_index = outputs_.size() - 1;
+    if ((op_index < 0) || (op_index > max_index)) {
+      return Status(
+          RequestStatusCode::INVALID_ARG,
+          "The output " + name +
+              " in the model config refers to an output index which doesn't "
```
deadeyegoodwin: Haven't we already tested (in the provider_utils) that the request asks only for outputs that are in the model config. That means that here we just have to make sure that every output required by the request is actually provided by the result. Isn't it ok if the result provides extra outputs that the request doesn't need (and/or that are listed in the model config). Perhaps unexpected? Perhaps we can log an error? But the request can still complete.

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -627,10 +707,17 @@ LibTorchBackend::Context::Run(
     // Checked at initialization time to make sure that STRING is not
     // being used for an output, so can just assume fixed-sized here.
     const DataType dtype = output_config->data_type();
+
+    // If a reshape is provided for the input then use that when
```
deadeyegoodwin: input -> output

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -627,10 +707,17 @@ LibTorchBackend::Context::Run(
     // Checked at initialization time to make sure that STRING is not
```
deadeyegoodwin: This comment belongs above the call to ReadFixedSizedOutputTensor below, not up here...

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -464,6 +509,17 @@ LibTorchBackend::Context::GetOutputTensor(
...
   try {
     torch::Tensor output_flat = (*outputs_)[op_index].flatten();
+
+    // verify output datatype matches datatype from model config
+    DataType rec_dtype = ConvertTorchTypeToDataType(output_flat.scalar_type());
+    if (dtype != rec_dtype) {
+      return Status(
+          RequestStatusCode::INVALID_ARG,
+          "unexpected datatype " + DataType_Name(rec_dtype) +
+              " for inference output 'OUTPUT__" + std::to_string(op_index) +
```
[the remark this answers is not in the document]
CoderHam: You are correct. Will fix this

src/backends/pytorch/libtorch_backend.cc
CoderHam: Yes we test that the outputs asked for are present in the model config. However we do not test that the outputs have a valid *__\<index\> i.e. an the model config can have *_1 but the model can have only one output i.e. *__0. 
However now I see that these checks should be present in the earlier validation section as they need only be performed once. I will move it there.

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -613,12 +669,36 @@ LibTorchBackend::Context::Run(
...
   RETURN_IF_ERROR(Execute(&inputs_, &outputs_));
 
-  // Make sure each output is of the expected size and copy it into
-  // the payload responses.
+  // TODO Validate Output shape after model execution
+
   for (const auto& output : base->Config().output()) {
     const std::string& name = output.name();
     std::string index_str = name.substr(name.find(deliminator) + 2);
+    int op_index = std::atoi(index_str.c_str());
```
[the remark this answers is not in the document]
CoderHam: Done

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -244,13 +277,19 @@ Status
...
     const auto pr = ConvertDataTypeToTorchType(io.data_type());
     if (!pr.first) {
       return Status(
           RequestStatusCode::INTERNAL,
           "unsupported datatype " + DataType_Name(io.data_type()) +
               " for output '" + io.name() + "' for model '" + name_ + "'");
+    } else {
+      const std::string& name = io.name();
+      int op_index = std::atoi(name.substr(name.find(deliminator) + 2).c_str());
+      io_index_map_[name] = op_index;
```
deadeyegoodwin: I don't think you can use the same map for both input and output. What if I have an input with the same name as an output?

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -394,21 +433,52 @@ Status
...
+      outputs_, op_index, name, dtype, &content, &byte_size, &content_shape));
+
+  // verify shape of output matches shape from model config
+  const int batch_offset = ((max_batch_size_ == NO_BATCHING) ? 0 : 1);
+
+  for (int i = 0; i < dims.size(); i++) {
+    if (dims[i] != -1) {
+      if (dims[i] != content_shape[i + batch_offset]) {
+
+        // create print friendly array of shapes
```
deadeyegoodwin: See src/core/model_config.h DimsListToString... you don't need to write your own versions.

src/backends/pytorch/libtorch_backend.cc
CoderHam: Right. Will fix that

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -226,13 +253,19 @@ Status
...
   for (const auto& io : ios) {
     const auto pr = ConvertDataTypeToTorchType(io.data_type());
     if (!pr.first) {
       return Status(
           RequestStatusCode::INTERNAL,
           "unsupported datatype " + DataType_Name(io.data_type()) +
               " for input '" + io.name() + "' for model '" + name_ + "'");
+    } else {
+      const std::string& name = io.name();
+      int op_index = std::atoi(name.substr(name.find(deliminator) + 2).c_str());
```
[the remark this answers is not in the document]
CoderHam: Nope. Will throw an exception and add a unit test for this

src/backends/pytorch/libtorch_backend.cc
CoderHam: Got it

PR #350, opened 2019-06-07
Fix memory leaks found with ASAN

deadeyegoodwin:


src/servers/libtrtserver.ldscript
```diff
@@ -30,5 +30,7 @@
     *nvidia15inferenceserver10GRPCServer*;
     *nvidia15inferenceserver10HTTPServer*;
     *nvidia15inferenceserver15InferenceServer*;
+    *nvidia15inferenceserver15InferenceServer*;
```
GuanLuo: Repeated

src/servers/grpc_server.cc
```diff
@@ -57,18 +57,19 @@ namespace {
 class AsyncResources : public nvrpc::Resources {
  public:
   explicit AsyncResources(
-      InferenceServer* server, int infer_threads, int mgmt_threads)
+      const std::shared_ptr<InferenceServer>& server, int infer_threads,
```
dzier: Why are you passing a shared pointer as a const reference. That kind of defeats a shared pointer. It should be treated just like a pointer so that reference counting works correctly.
```
std::shared_ptr<InferenceServer> server, int infer_threads,
```
This would be analogous to `InferenceServer *`

src/servers/grpc_server.cc
```diff
@@ -57,18 +57,19 @@ namespace {
...
   }
 
-  InferenceServer* GetServer() { return m_Server; }
+  const std::shared_ptr<InferenceServer>& GetServer() { return m_Server; }
   ThreadPool& GetMgmtThreadPool() { return m_MgmtThreadPool; }
   ThreadPool& GetInferThreadPool() { return m_InferThreadPool; }
 
  private:
-  InferenceServer* m_Server;
+  std::shared_ptr<InferenceServer> m_Server;
```
dzier: question about consistency in style here. Shouldn't this be `server_` and not `m_Server`?

src/servers/grpc_server.cc
deadeyegoodwin: For efficiency we always pass (non-primitive) arguments as const& (arguments that are actually returns we pass as pointers... we never pass arguments by non-const ref... well we do a few places but we shouldn't).

If the callee needs to take ownership of the argument it must make a copy.

shared_ptr (and unique_ptr) are just classes.  If the callee needs their own copy of the class (which in this case means taking ownership of the shared pointer) then they make a copy. In this case I assume in the class you will see:
 std::shared_ptr<InferenceServer> server_;
and server_ is assigned to server parameter and at that point the shared_ptr is copied and reference count updated.

If you don't use const& there is an extra copy of the shared_ptr, increment of reference count, destruction or shared_ptr and decrement of reference count.

src/servers/grpc_server.cc
deadeyegoodwin: Yes, those names are not in the correct style.

PR #357, opened 2019-06-10
Consistent validation checks across all Backends

CoderHam:
- [x] **TensorFlow** - Add verification of shape of output after execution 
- [x] **Caffe2** - Add verification of shape of output after execution 
- [x] **TensorFlow** - Add validation of input + output datatype in model config
- [x] **TensorRT** - Add validation of input + output datatype in model config 
 

- Init time validation for datatype/shape for **GraphDef** models not possible for all graphdef models. The graphdef model has missing or incorrect shape/datatype for some tensors and is therefore not stable for model validation.

**Additionally:**
- [x] **Onnx** model batch shape validation and model generation

### DIFF
--- a/src/backends/caffe2/netdef_backend.cc
+++ b/src/backends/caffe2/netdef_backend.cc
@@ -459,26 +459,12 @@ NetDefBackend::Context::ReadFixedSizedOutputTensor(
   for (int i = 0; i < dims.size(); i++) {
     if (dims[i] != -1) {
       if (dims[i] != content_shape[i + batch_offset]) {
-
-        // create print friendly array of shapes
-        std::string expected_shape = "[ ";
-        std::string output_shape = "[ ";
-        for (size_t i = batch_offset; i < content_shape.size() - 1; i++) {
-          output_shape += std::to_string(content_shape[i]) + ", ";
-        }
-        output_shape +=
-            std::to_string(content_shape[content_shape.size() - 1]) + " ]";
-
-        for (int i = 0; i < dims.size() - 1; i++) {
-          expected_shape += std::to_string(dims[i]) + ", ";
-        }
-        expected_shape += std::to_string(dims[dims.size() - 1]) + " ]";
-
         return Status(
             RequestStatusCode::INVALID_ARG,
             "unexpected shape for output '" + name +
-                "', model configuration shape is " + expected_shape +
-                ", inference shape is " + output_shape);
+                "', model configuration shape is " +
+                DimsListToString(content_shape) + ", inference shape is " +
+                DimsListToString(dims));
       }
     }
   }
--- a/src/backends/caffe2/netdef_backend.cc
+++ b/src/backends/caffe2/netdef_backend.cc
@@ -442,7 +442,7 @@ Status
 NetDefBackend::Context::ReadFixedSizedOutputTensor(
     const std::string& name, const Caffe2Workspace::DataType dtype,
     const size_t dtype_byte_size, const size_t total_batch_size,
-    std::vector<Scheduler::Payload>* payloads)
+    std::vector<Scheduler::Payload>* payloads, const DimsList& dims)
 {
   std::vector<int64_t> content_shape;
   const char* content = nullptr;
@@ -451,6 +451,36 @@ NetDefBackend::Context::ReadFixedSizedOutputTensor(
       name, dtype, &content, &byte_size, &content_shape);
   if (!err.IsOk()) {
     return Status(RequestStatusCode::INTERNAL, err.Message());
+  }
+
+  // verify shape of output matches shape from model config
+  const int batch_offset = ((max_batch_size_ == NO_BATCHING) ? 0 : 1);
+
+  for (int i = 0; i < dims.size(); i++) {
+    if (dims[i] != -1) {
+      if (dims[i] != content_shape[i + batch_offset]) {
+
+        // create print friendly array of shapes
+        std::string expected_shape = "[ ";
+        std::string output_shape = "[ ";
+        for (size_t i = batch_offset; i < content_shape.size() - 1; i++) {
+          output_shape += std::to_string(content_shape[i]) + ", ";
+        }
+        output_shape +=
+            std::to_string(content_shape[content_shape.size() - 1]) + " ]";
+
+        for (int i = 0; i < dims.size() - 1; i++) {
+          expected_shape += std::to_string(dims[i]) + ", ";
+        }
+        expected_shape += std::to_string(dims[dims.size() - 1]) + " ]";
+
+        return Status(
+            RequestStatusCode::INVALID_ARG,
+            "unexpected shape for output '" + name +
+                "', model configuration shape is " + expected_shape +
+                ", inference shape is " + output_shape);
+      }
+    }
   }
 
   const size_t total_byte_size =
@@ -628,9 +658,14 @@ NetDefBackend::Context::Run(
     // being used for an output, so can just assume fixed-sized here.
     const Caffe2Workspace::DataType dtype =
         ConvertDataType(output_config->data_type());
+
+    const DimsList& output_dims = (output_config->has_reshape())
+                                      ? output_config->reshape().shape()
+                                      : output_config->dims();
+
     RETURN_IF_ERROR(ReadFixedSizedOutputTensor(
         name, dtype, GetDataTypeByteSize(output_config->data_type()),
-        total_batch_size, payloads));
+        total_batch_size, payloads, output_dims));
   }
 
   return Status::Success;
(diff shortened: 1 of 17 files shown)


src/backends/caffe2/netdef_backend.cc
```diff
@@ -453,6 +453,36 @@ NetDefBackend::Context::ReadFixedSizedOutputTensor(
     return Status(RequestStatusCode::INTERNAL, err.Message());
   }
 
+  // verify shape of output matches shape from model config
+  const int batch_offset = ((max_batch_size_ == NO_BATCHING) ? 0 : 1);
+
+  for (int i = 0; i < dims.size(); i++) {
+    if (dims[i] != -1) {
+      if (dims[i] != content_shape[i + batch_offset]) {
```
deadeyegoodwin: Use utility methods from model_config.h to create string representation of dims.

src/backends/onnx/onnx_utils.cc
```diff
@@ -274,8 +274,8 @@ CompareDimsSupported(
     const int max_batch_size)
 {
   // If the model configuration expects batching support in the model,
-  // then the model shape first dimension must be either -1 (unbounded)
-  // or value that larger than max_batch_size.
+  // then the model shape's first dimension must be either -1 (unbounded)
```
deadeyegoodwin: Is this true. If the model shape's first dimension is not -1, then how can be know if it supports batching or not? @GuanLuo what does this mean?

src/backends/tensorrt/plan_backend.cc
```diff
@@ -111,6 +111,24 @@ PlanBackend::Context::~Context()
   }
 }
 
+bool
```
deadeyegoodwin: Use ConvertDataType from plan_utils.cc

src/backends/tensorflow/base_backend.cc
```diff
@@ -649,6 +687,26 @@ BaseBackend::Context::Run(
       skip_element_cnt = false;
     }
 
+    const DimsList& output_dims = (output_config->has_reshape())
```
deadeyegoodwin: So can we not validate the shape at init time? For saved-model I know we can, what about graphdef?

src/backends/onnx/onnx_utils.cc
GuanLuo: Here just saying that the batch dimension doesn't have to be -1. By the point where this function is called, the model config should be normalized. So if the model config says `max_batch_size : 8` and input dims is [16]. Then the actual input shape (retrieved from model) can be [-1, 16] or [9, 16] (as long as the value is >= 8)

deadeyegoodwin: For TF we are more conservative in that we only allow TRTIS to assume the model understand batching (max_batch_size > 0) if the first dimension of every tensor is -1. It seems dangerous to assume that other tensors (without -1 batch dim) could support batching.

src/backends/tensorflow/base_backend.cc
CoderHam: I tested it, it works for graphdef. The QA models include some graphdef models as well and it runs successfully with them.

src/backends/tensorrt/plan_backend.cc
CoderHam: plan_utils.cc contains a function `DataType ConvertDatatype(nvinfer1::DataType trt_type)` but not a function to test convert the DataType in config file to a nvinfer1::DataType. 
I can move this function to the plan_utils.cc and create the appropriate fucntion: `nvinfer1::DataType ConvertDatatype(DataType dtype)`

src/backends/tensorflow/base_backend.cc
deadeyegoodwin: I was asking if we can test the graphdef tensor shapes against the model config at init time instead of runtime (like your change here). Since we already can check saved-model at init time it would be good if we could also check graphdef at init time, then we wouldn't need this runtime check.

src/backends/onnx/onnx_utils.cc
GuanLuo: I see your concern now. If the first dimension has to be -1, then people who doesn't know the detail of the model can change the `max_batch_size` without any concern. However, in the case here, they may find that the model suddenly can't be loaded after they change it to some larger value.

So I agree with you, I think being more conservative will be better.

src/backends/tensorrt/plan_backend.cc
deadeyegoodwin: Yes, replace this with a new ConvertDatatype in plan_utils.

src/backends/tensorflow/base_backend.cc
CoderHam: Got it. Will try to do this at init time.

CoderHam: Not possible for all graphdef models.

CoderHam: When possible the model's tensor shape and datatype are used for validation of GraphDef models in Init. If not, the same is done during runtime.

src/backends/onnx/onnx_utils.cc
CoderHam: Updated the generated models and fixed this check to be more conservative (as discussed above).

src/backends/onnx/onnx_utils.cc
```diff
@@ -274,18 +274,10 @@ CompareDimsSupported(
     const int max_batch_size)
 {
   // If the model configuration expects batching support in the model,
-  // then the model shape first dimension must be either -1 (unbounded)
-  // or value that larger than max_batch_size.
+  // then the tensorflow shape first dimension must be -1.
```
deadeyegoodwin: This is onnx, not tensorflow...

src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -625,9 +625,61 @@ TRTISTF_ModelCreateFromGraphDef(
     if (node.op() == "Placeholder") {
       potential_inputs =
           TRTISTF_IOListNew(node.name().c_str(), nullptr, potential_inputs);
+
+      TRTISTF_IO* io = potential_inputs->io_;
+
+      TRTISTF_DataType dt;
+      tensorflow::DataType tf_dt;
```
deadeyegoodwin: You never use this value

src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -625,9 +625,61 @@ TRTISTF_ModelCreateFromGraphDef(
...
+      io->data_type_ = dt;
+      io->shape_ = TRTISTF_ShapeNew(tensor_shape.size(), tensor_shape.data());
     } else {
       potential_outputs =
           TRTISTF_IOListNew(node.name().c_str(), nullptr, potential_outputs);
+
+      TRTISTF_IO* io = potential_outputs->io_;
+
+      TRTISTF_DataType dt;
+      tensorflow::DataType tf_dt;
```
deadeyegoodwin: Never used

src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -625,9 +625,61 @@ TRTISTF_ModelCreateFromGraphDef(
...
+              auto shape = value.shape();
+              for (int i=0; i<shape.dim_size(); ++i) {
+                  auto dim = shape.dim(i);
+                  auto dim_size = dim.size();
+                  tensor_shape.push_back(dim_size);
+              }
+          }
+      }
+
+      io->data_type_ = dt;
```
deadeyegoodwin: What is dt is undefined? Should initialize it to invalid. Also, you don't ever seem to use this calculated type and shape. For savedmodel we collect these and then in savedmodel_backend.cc we use them to check against the configuration. Here you collect them but you didn't change graphdef_backend.cc to use them to check against the configuration. Also, need some way to represent unknown datatype or shape so that you know when to check and when you can't (for savedmodel we also assume there is type and shape but for shape at least that may change in future so will need this not just for graphdef).

src/backends/tensorflow/tensorflow_backend_tf.cc
CoderHam: Was a part of earlier changes. remove.

src/backends/tensorflow/tensorflow_backend_tf.cc
CoderHam: Was a part of earlier changes. remove. 

This was reverted because for GraphDef models the datatype stored is sometimes a valid datatype but it is not _correct_. 
For example, for the Resnet GraphDef model, the shape of input was [224,3] while the true shape expected is [224, 224, 3], similarly the datatype for the Softmax layer in some networks was **INT32** but it is actually supposed to be **FLOAT32**.

src/backends/onnx/onnx_utils.cc
CoderHam: Fixed.

src/backends/tensorflow/tensorflow_backend_tf.cc
CoderHam: Was a part of earlier changes. removed.

PR #362, opened 2019-06-11
Always check input/output signature to verify batching support

GuanLuo:
Before this change, server that starts with `--strict-model-config=false` will fail to load a non-batch model that contains full config.pbtxt. Its `max_batch_size` will be overwritten to 1. Can reproduce with one of the QA non-batch saved model


deadeyegoodwin:
@GuanLuo You need to manually rebase this to pick up the build fix on master.

PR #364, opened 2019-06-12
Implement AsyncRun C++ client API that uses callback

GuanLuo:


src/clients/c++/request_common.cc
```diff
@@ -925,12 +925,29 @@ InferContextImpl::GetReadyAsyncRequest(
...
+          static_cast<RequestImpl*>(ongoing_async_request.second.get());
+      if (!request_impl->HasCallback()) {
+        has_dangling_request = true;
+        break;
+      }
+    }
+    if (!has_dangling_request) {
+      return Error(
+          RequestStatusCode::UNAVAILABLE,
+          "No asynchronous requests should be returned, all requests will be "
```
deadeyegoodwin: "No asynchronous requests can be returned, all outstanding requests will signal completion via their callback function"

src/clients/c++/request.h
```diff
@@ -616,6 +618,17 @@ class InferContext {
   /// \return Error object indicating success or failure.
   virtual Error AsyncRun(std::shared_ptr<Request>* async_request) = 0;
 
+  /// Similar to AsyncRun() above. However, this function does not return the
```
deadeyegoodwin: For the above AsyncRun put a deprecation notice at the top of the description.
"DEPRECATED: This function is deprecated and will be removed in a future version of this API. Use AsyncRun(OnCompleteFn)".

On GetReadyAsyncRequest also should put a deprecation notice.
"DEPRECATED: This function is deprecated and will be removed in a future version of this API. This function is only useful with the deprecated version of AsyncRun(). Instead use AsyncRun(OnCompleteFn)"

PR #369, opened 2019-06-12
Add Autofill for LibTorch

CoderHam:


src/backends/pytorch/autofill.cc
```diff
@@ -0,0 +1,97 @@
...
+  // case where there is one version directory.
+  if (version_dirs.size() != 1) {
+    return Status(
+        RequestStatusCode::INTERNAL,
+        "unable to autofill for '" + model_name + "' due to multiple versions");
+  }
+
+  const auto version_path = JoinPath({model_path, *(version_dirs.begin())});
+
+  // There must be a single pt model (which is spread across two
```
deadeyegoodwin: This comment seems wrong... it is 1 file right? Not "spread across two files"

CoderHam: Yes that was a typo. Fixed in next commit.

PR #371, opened 2019-06-12
Implement GCS Filesystem support

xprotobeast2:
Cmake build and FileSystem interface for google cloud storage using C++ client API.

### DIFF
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -36,6 +36,7 @@ option(TRTIS_ENABLE_CAFFE2 "Include Caffe2 backend in server" OFF)
 option(TRTIS_ENABLE_ONNXRUNTIME "Include ONNXRuntime backend in server" OFF)
 option(TRTIS_ENABLE_PYTORCH "Include PyTorch backend in server" OFF)
 option(TRTIS_ENABLE_CUSTOM "Include Custom backend in server" OFF)
+option(TRTIS_ENABLE_GCS "Include GCS Filesystem support in server" OFF)
 
 # Multiple paths may be specified by separating them with semicolon
 set(TRTIS_ONNXRUNTIME_INCLUDE_PATHS "" CACHE PATH "Paths to ONNXRuntime includes")
@@ -202,6 +203,49 @@ ExternalProject_Add(prometheus-cpp
 )
 
 #
+# Build crc32c
+#
+ExternalProject_Add(crc32c
+  PREFIX crc32c
+  GIT_REPOSITORY "https://github.com/google/crc32c.git"
+  GIT_TAG "1.0.6"
+  SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/crc32c/src/crc32c"
+  CMAKE_CACHE_ARGS
+    -DBUILD_SHARED_LIBS:STRING=yes
+    -DCRC32C_BUILD_TESTS:BOOL=OFF
+    -DCRC32C_BUILD_BENCHMARKS:BOOL=OFF
+    -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON
+    -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_CURRENT_BINARY_DIR}/crc32c/install
+)
+
+set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH} 
+   ${CMAKE_CURRENT_BINARY_DIR}/c-ares/lib/cmake/c-ares
+   ${CMAKE_CURRENT_BINARY_DIR}/grpc/lib/cmake/grpc
+   ${CMAKE_CURRENT_BINARY_DIR}/
+   ${_FINDPACKAGE_PROTOBUF_CONFIG_DIR})
+
+#
+# Build google-cloud-cpp
+#
+ExternalProject_Add(google-cloud-cpp
+  PREFIX google-cloud-cpp
+  GIT_REPOSITORY "https://github.com/googleapis/google-cloud-cpp.git"
+  GIT_TAG "v0.10.x"
+  SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/google-cloud-cpp/src/google-cloud-cpp"
+  CMAKE_CACHE_ARGS
+    -DGOOGLE_CLOUD_CPP_DEPENDENCY_PROVIDER:STRING=package
+    -DGOOGLE_CLOUD_CPP_PROTOBUF_PROVIDER:STRING=package
+    -DBUILD_TESTING:BOOL=OFF
+    -DBUILD_SHARED_LIBS:BOOL=ON
+    -DgRPC_DIR:PATH=${CMAKE_CURRENT_BINARY_DIR}/grpc/lib/cmake/grpc
+    -DCMAKE_PREFIX_PATH:PATH=${CMAKE_PREFIX_PATH}
+    -DCrc32c_DIR:PATH=${CMAKE_CURRENT_BINARY_DIR}/crc32c/install/lib/cmake/Crc32c
+    -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON
+    -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_CURRENT_BINARY_DIR}/google-cloud-cpp/install
+  DEPENDS grpc protobuf crc32c
+)
+
+#
 # Build TRTIS test utilities
 #
 if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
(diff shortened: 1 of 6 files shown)


build/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ option(TRTIS_ENABLE_CAFFE2 "Include Caffe2 backend in server" OFF)
 option(TRTIS_ENABLE_ONNXRUNTIME "Include ONNXRuntime backend in server" OFF)
 option(TRTIS_ENABLE_PYTORCH "Include PyTorch backend in server" OFF)
 option(TRTIS_ENABLE_CUSTOM "Include Custom backend in server" OFF)
+option(TRTIS_ENABLE_GCS "Include GCS Filesystem support in server" OFF)
```
deadeyegoodwin: Move this down to after line 48. This group of enables is for backends.

build/CMakeLists.txt
```diff
@@ -201,6 +202,49 @@ ExternalProject_Add(prometheus-cpp
...
+#
+# Build crc32c
+#
+ExternalProject_Add(crc32c
+  PREFIX crc32c
+  GIT_REPOSITORY "https://github.com/google/crc32c.git"
+  GIT_TAG "1.0.6"
+  SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/crc32c/src/crc32c"
+  CMAKE_CACHE_ARGS
+    -DBUILD_SHARED_LIBS:STRING=yes
```
deadeyegoodwin: Can this be built static and have it be built-into GCS library so we don't need to deliver it as separate shared library?

build/CMakeLists.txt
```diff
@@ -201,6 +202,49 @@ ExternalProject_Add(prometheus-cpp
...
+    -DCRC32C_BUILD_TESTS:BOOL=OFF
+    -DCRC32C_BUILD_BENCHMARKS:BOOL=OFF
+    -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON
+    -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_CURRENT_BINARY_DIR}/crc32c/install
+)
+
+set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH}
```
deadeyegoodwin: Why is this needed? You should be able to use just -DProtobuf_DIR, -DgRPC_DIR and -Dc-ares_DIR in CMAKE_CACHE_ARGS. Doing it this way changes the path for all the projects, not just GCS.

build/CMakeLists.txt
```diff
@@ -201,6 +202,49 @@ ExternalProject_Add(prometheus-cpp
...
+ExternalProject_Add(google-cloud-cpp
+  PREFIX google-cloud-cpp
+  GIT_REPOSITORY "https://github.com/googleapis/google-cloud-cpp.git"
+  GIT_TAG "v0.10.x"
+  SOURCE_DIR "${CMAKE_CURRENT_BINARY_DIR}/google-cloud-cpp/src/google-cloud-cpp"
+  CMAKE_CACHE_ARGS
+    -DGOOGLE_CLOUD_CPP_DEPENDENCY_PROVIDER:STRING=package
+    -DGOOGLE_CLOUD_CPP_PROTOBUF_PROVIDER:STRING=package
+    -DBUILD_TESTING:BOOL=OFF
+    -DBUILD_SHARED_LIBS:BOOL=ON
```
deadeyegoodwin: It would be worth trying with static libraries to see what happens... that would save us having to install the extra shared libraries.

build/trtis/CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wno-unused-parameter -Werror")
+set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Werror")
```
deadeyegoodwin: We should add this flag only when compiling the files that need it...

src/core/filesystem.cc
```diff
@@ -205,10 +210,317 @@ LocalFileSystem::WriteTextFile(
...
+  Status GetDirectorySubdirs(
+      const std::string& path, std::set<std::string>* subdirs) override;
+  Status GetDirectoryFiles(
+      const std::string& path, std::set<std::string>* files) override;
+  Status ReadTextFile(const std::string& path, std::string* contents) override;
+  Status WriteTextFile(
+      const std::string& path, const std::string& contents) override;
+
+ private:
+  google::cloud::StatusOr<gcs::Client> client_;
```
deadeyegoodwin: Put members after functions, separated by blank line (so move to line 238)

src/core/filesystem.cc
```diff
@@ -205,10 +210,317 @@ LocalFileSystem::WriteTextFile(
...
+GCSFileSystem::ParsePath(
+    const std::string& path, std::string* bucket, std::string* object)
+{
+  // Get the bucket name and the object path. Return error if input is malformed
+  size_t bucket_start = path.find("gs://") + 5;
+  size_t bucket_end = path.find("/", bucket_start);
+  *bucket = path.substr(bucket_start, bucket_end - bucket_start);
+  *object = path.substr(bucket_end + 1);
+
+
```
deadeyegoodwin: Coding style... don't use multiple blank lines

src/core/filesystem.cc
```diff
@@ -205,10 +210,317 @@ LocalFileSystem::WriteTextFile(
...
+  if (object->empty()) {
+    return Status(
+        RequestStatusCode::INTERNAL, "No file name found in path: " + path);
+  }
+
+  return Status::Success;
+}
+
+std::string
+AppendSlash(std::string& name)
```
deadeyegoodwin: const std::string&

src/core/filesystem.cc
```diff
@@ -205,10 +210,317 @@ LocalFileSystem::WriteTextFile(
...
+{
+  // Helper function to take care of lack of trailing slashes
```
deadeyegoodwin: Function description goes before the function (line 282)

src/core/filesystem.cc
```diff
@@ -205,10 +210,317 @@ LocalFileSystem::WriteTextFile(
...
+  google::cloud::StatusOr<gcs::BucketMetadata> bucket_metadata =
+      client_->GetBucketMetadata(bucket);
+
+  if (!bucket_metadata) {
+    return Status(
+        RequestStatusCode::INTERNAL,
+        "Could not get MetaData for bucket with name " + bucket);
+  }
+
+  // Check if whether it has children. If at least one child, it is a directory
```
deadeyegoodwin: So GCS doesn't allow empty directories?

src/core/filesystem.cc
```diff
@@ -205,10 +210,317 @@ LocalFileSystem::WriteTextFile(
...
+  for (auto&& object_metadata :
+       client_->ListObjects(bucket, gcs::Prefix(AppendSlash(object_path)))) {
+    if (object_metadata) {
+      *is_dir = true;
+      return Status::Success;
```
deadeyegoodwin: Use break here instead of having early exit.

src/core/filesystem.cc
```diff
@@ -27,13 +27,16 @@
 #include "src/core/filesystem.h"
 
 #include <dirent.h>
+#include <google/cloud/storage/client.h>
```
deadeyegoodwin: Need ifdef TRTIS_ENABLE_GCS to protect the parts of the file that depend on GCS. You can do a build to test and make sure you did it correctly.

deadeyegoodwin:
@xprotobeast2 You also need to rebase to master since there have been recent changes to some common files.

src/core/filesystem.cc
xprotobeast2: It does but the way I've done it here, empty directories will just have themselves as a child. I added a check to remove these is GetDirectoryContents().

src/core/filesystem.cc
```diff
@@ -205,11 +212,327 @@ LocalFileSystem::WriteTextFile(
...
+}
+
+#endif  // TRTIS_ENABLE_GCS
+
+Status
+GetFileSystem(const std::string& path, FileSystem** file_system)
+{
+#ifdef TRTIS_ENABLE_GCS
+  // Check if this is a GCS path (gs://$BUCKET_NAME)
+  if (!path.empty() && !path.rfind("gs://", 0)) {
```
deadeyegoodwin: Here you need to still detect gs:// but if TRTIS_ENABLE_GCS is false then raise an appropriate error about trtis not being compiled with support for gcs.

build/CMakeLists.txt
```diff
@@ -201,6 +204,49 @@ ExternalProject_Add(prometheus-cpp
...
+    -DCRC32C_BUILD_BENCHMARKS:BOOL=OFF
+    -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON
+    -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_CURRENT_BINARY_DIR}/crc32c
+)
+
+
+# This is necessary because google-cloud-cpp 
+# does not pass protobuf_DIR and gRPC_DIR etc to its 
+# external projects but expects them in CMAKE_PREFIX_PATH
+set(GCS_CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH}
```
deadeyegoodwin: Can you guard this with TRTIS_ENABLE_GCS

PR #372, opened 2019-06-13
Improve perf_client to use AsyncRun with callback API 

GuanLuo:


GuanLuo:
@deadeyegoodwin I only tested it on my local machine (1 GPU) and the results are the same

PR #377, opened 2019-06-13
 Implement AysncRun Python client API that uses callback

GuanLuo:


src/clients/python/__init__.py
```diff
@@ -952,6 +974,80 @@ def async_run(self, inputs, outputs, batch_size=1, flags=0):
...
+        inside the callback function or deferring it to a different thread so that
+        the InferContext is unblocked.
+
+        Parameters
+        ----------
+        callback : function
+            Python function that accepts an InferContext object that sends the
+            request and an integer identifier as arguments. This function will
+            be invoked once the request is completed.
+        inputs : dict
```
deadeyegoodwin: Insert blank line

src/clients/python/__init__.py
```diff
@@ -952,6 +974,80 @@ def async_run(self, inputs, outputs, batch_size=1, flags=0):
...
+        # Wrap over the provided callback
+        wrapped_cb = partial(self._async_callback_wrapper, self._callback_resources_dict_id, callback)
+        c_cb = _async_run_callback_prototype(wrapped_cb)
+
+        # Run asynchronous inference...
+        _raise_if_error(
+            c_void_p(
+                _crequest_infer_ctx_async_run_with_cb(self._ctx, c_cb)))
+
+        self._callback_resources_dict[self._callback_resources_dict_id] = (outputs, batch_size, contiguous_input, c_cb, wrapped_cb)
```
deadeyegoodwin: long line

src/clients/python/simple_callback_client.py
```diff
@@ -0,0 +1,157 @@
...
+    ctx = InferContext(FLAGS.url, protocol, model_name, model_version, FLAGS.verbose)
+
+    # Create the data for the two input tensors. Initialize the first
+    # to unique integers and the second to all ones.
+    input0_data = np.arange(start=0, stop=16, dtype=np.int32)
+    input1_data = np.ones(shape=16, dtype=np.int32)
+
+    # Send inference request to the inference server. Get results for
+    # both output tensors.
+    class Context:
```
deadeyegoodwin: Very confusing to have something else called context. Can you instead call it CompletionUserData or something like that.

src/clients/python/simple_callback_client.py
```diff
@@ -0,0 +1,157 @@
...
+            self._cv = threading.Condition(self._lock)
+            self._input0_data = input0_data
+            self._input1_data = input1_data
+            self._request_cnt = 2
+            self._done_cnt = 0
+            self._recent_ctx_obj = None
+            self._recent_request_id = None
+            self._success = True
+    
+    context = Context(input0_data, input1_data)
```
deadeyegoodwin: user_data = ...

src/clients/python/simple_callback_client.py
```diff
@@ -0,0 +1,157 @@
...
+    def active_callback(context, idx, ctx_obj, request_id):
```
deadeyegoodwin: Why is it called active_callback instead of completion_callback. 'ctx_obj' is also generic, how about 'infer_ctx'. This example is also confusing because there is a lot of stuff in that user data and it doesn't seem to be best practice. Can you instead just grab infer_ctx and request_id and schedule a future or enqueue or something that then performs the checking and error reporting.

src/clients/python/simple_callback_client.py
```diff
@@ -0,0 +1,157 @@
...
+                              { 'OUTPUT0' : InferContext.ResultFormat.RAW,
+                                'OUTPUT1' : InferContext.ResultFormat.RAW },
+                              batch_size)
+
+    context._cv.acquire()
+    while context._recent_ctx_obj is None:
+        context._cv.wait()
+    context._cv.release()
+
+    # Retrieve InferContext and request id from the shared context
```
deadeyegoodwin: Lots of common code with above that should be combined into a "checking" thread or future.

GuanLuo: @deadeyegoodwin I simplified the callback example. Although GtiHub is still showing "change requested"

PR #379, opened 2019-06-14
Add perf_client option to use user provided data

GuanLuo:


src/clients/c++/perf_client.cc
```diff
@@ -181,6 +181,25 @@ typedef struct PerformanceStatusStruct {
...
+  if (!in) {
+    return nic::Error(
+        ni::RequestStatusCode::INVALID_ARG,
+        "failed to open file '" + path + "'");
+  }
+
+  in.seekg(0, std::ios::end);
+  contents->resize(in.tellg());
+  in.seekg(0, std::ios::beg);
+  in.read(&(*contents)[0], contents->size());
```
nieksand: If `path` points to an empty file, the vector size will be zero.  Is it valid to take the address of a non-existent 0th element?

(It's been many years since I worked in C++)

src/clients/c++/perf_client.cc
```diff
@@ -1665,10 +1719,17 @@ Usage(char** argv, const std::string& msg = std::string())
...
-      << " is stable instead of average latency."
-      << " Default is -1 to indicate no percentile will be used or reported."
-      << std::endl;
+      << " of latency will also be reported and used to detemine if the"
+      << " measurement is stable instead of average latency."
+      << " Default is -1 to indicate no percentile will be used." << std::endl;
+  std::cerr
+      << "For --data-directory, it indicates that the perf client will use user"
+      << " provided data instead of synthetic data for model inputs. For each"
+      << " input, there must be a binary file in the way that its relative path"
```
deadeyegoodwin: "... must be a binary file for each input required by the model. The file must be named the same as the input and must contain data required for sending the input...

src/clients/c++/perf_client.cc
GuanLuo: Added check on return value of `in.tellg()`

PR #380, opened 2019-06-14
Add library interface for inferencing

deadeyegoodwin:


src/core/trtserver.cc
```diff
@@ -196,6 +289,89 @@ TRTSERVER_ProtobufSerialize(
...
 
+//
+// TRTSERVER_MemoryAllocator
+//
+TRTSERVER_Error*
+TRTSERVER_MemoryAllocatorNew(
+    TRTSERVER_MemoryAllocator** allocator, TRTSERVER_MemoryAllocFn_t alloc_fn,
+    TRTSERVER_MemoryDeleteFn_t delete_fn)
+{
+  return nullptr;  // Success
```
GuanLuo: Is this supposed to be implemented?

src/core/trtserver.cc
```diff
@@ -311,6 +487,63 @@ TRTSERVER_ServerModelStatus(
...
+      *request_header, backend->GetInferenceBackend()->GetLabelProvider(),
+      &infer_response_provider));
+
+  auto request_status = std::make_shared<ni::RequestStatus>();
+  lserver->HandleInfer(
+      request_status.get(), backend, infer_request_provider,
+      infer_response_provider, infer_stats,
+      [infer_stats, timer, infer_response_provider, server, request_status,
+       complete_fn, userp]() mutable {
+        infer_stats->SetFailed(false);
```
GuanLuo: `infer_stats` will be set in [HandleInfer()](https://github.com/NVIDIA/tensorrt-inference-server/blob/81bbd906e73f6bf3604604d8bcc8e9eefc2af214/src/core/server.cc#L350). And the response may need to include request status so that user can check if the inference is successful.

src/core/trtserver.cc
deadeyegoodwin: Not yet... it is going to require some provider changes that I will make in a future change. But I will add a note in the code

src/core/trtserver.cc
deadeyegoodwin: Added TRTSERVER_InferenceResponse* methods to get status and response header.

PR #392, opened 2019-06-18
Allow saved-model to not specify shape on input/output

CoderHam:
If no shape is specified then the shape from the model configuration will be used.


src/backends/tensorflow/savedmodel_backend.cc
```diff
@@ -133,8 +133,19 @@ SavedModelBackend::CreateTRTISTFModel(
...
         (io.has_reshape()) ? io.reshape().shape() : io.dims();
 
-    RETURN_IF_ERROR(CompareDimsSupported(
-        Name(), io.name(), input->shape_, dims, Config().max_batch_size() > 0));
+    if (input->shape_->rank_ != 0) {
+      RETURN_IF_ERROR(CompareDimsSupported(
+          Name(), io.name(), input->shape_, dims, Config().max_batch_size() > 0));
+    }
+    else {
+      // Use input shape from model configuration if empty
```
deadeyegoodwin: "The savedmodel doesn't specify a shape for the input so use the shape from the model configuration"

src/backends/tensorflow/savedmodel_backend.cc
```diff
@@ -162,9 +173,19 @@ SavedModelBackend::CreateTRTISTFModel(
...
 
-    RETURN_IF_ERROR(CompareDimsSupported(
-        Name(), io.name(), output->shape_, dims,
-        Config().max_batch_size() > 0));
+    if (output->shape_->rank_ != 0) {
+      RETURN_IF_ERROR(CompareDimsSupported(
+          Name(), io.name(), output->shape_, dims, Config().max_batch_size() > 0));
+    }
+    else {
+      // Use output shape from model configuration if empty
```
deadeyegoodwin: "The savedmodel doesn't specify a shape for the output so use the shape from the model configuration"

qa/L0_savedmodel_shape/gen_qa_models_savedmodel.py
```diff
@@ -0,0 +1,328 @@
+# Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: This file should move to qa/common/gen_qa_noshape_models.py and be invoked by gen_qa_model_repository as we do with the other qa models. You will need to run it on a system with TitanV (or v100) and then copy over to dbcluster with the other qa model repositories.

CoderHam: Already generated and put on the dbcluster. Will add to the gen_qa_model_repository and in qa/common/gen_qa_noshape_models.py

src/backends/tensorflow/savedmodel_backend.cc
CoderHam: Fixed comment.

src/backends/tensorflow/savedmodel_backend.cc
CoderHam: Fixed comment.

PR #394, opened 2019-06-19
Fix pytorch backend to use GCS compatible API to load model

xprotobeast2:


deadeyegoodwin:
Looks ok to me. @CoderHam please review.

PR #403, opened 2019-06-24
Modify client library and its build to be compatible with Windows

GuanLuo:


src/clients/c++/CMakeLists.txt
```diff
@@ -30,23 +30,47 @@ cmake_minimum_required (VERSION 3.5)
...
-  request_common.cc request_common.h
-  request_http.cc request_grpc.cc
-  $<TARGET_OBJECTS:model-config-library>
-  $<TARGET_OBJECTS:proto-library>
-  $<TARGET_OBJECTS:grpc-library>
-)
+
+# use different syntax for Windows
+if(WIN32)
+  # using static so that linker does not complain on unresolved external symbol
```
deadeyegoodwin: What symbol is missing? librequest.so shouldn't have any undefined symbols

GuanLuo: It complains about Error::Success [here](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/clients/c%2B%2B/request.h#L719), it works if I change it to `err` instead of using static class member, but then CI complains with [maybe-uninitialized error](https://gitlab-master.nvidia.com/dl/dgx/tensorrtserver/-/jobs/7009505)

deadeyegoodwin: Did you try using dllimport as described here: https://stackoverflow.com/a/2479921

src/clients/c++/CMakeLists.txt
```diff
@@ -30,23 +30,37 @@ cmake_minimum_required (VERSION 3.5)
...
+    request.cc request.h
+    request_common.cc request_common.h
+    request_http.cc request_grpc.cc
+    $<TARGET_OBJECTS:model-config-library>
+    $<TARGET_OBJECTS:proto-library>
+    $<TARGET_OBJECTS:grpc-library>
+  )
+
+# use different syntax for Windows
+if(WIN32)
```
deadeyegoodwin: You can list target_link_libraries multiple times, so can everything be common except -lpthread (use ${CURL_LIBRARY_RELEASE} for both??). Then add:

if(NOT WIN32)
target_link_libraries(
  request
  -lpthread
)

src/clients/python/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ add_library(crequest SHARED crequest.cc)
 target_link_libraries(
   crequest
   PUBLIC request
+  PRIVATE protobuf::libprotobuf
```
deadeyegoodwin: Why is this needed?

src/clients/python/setup.py
```diff
@@ -54,6 +54,10 @@ def get_tag(self):
 except ImportError:
     bdist_wheel = None
 
+platform_package_data= [ 'libcrequest.so', 'librequest.so', ]
```
deadeyegoodwin: If then else?
Also space before '='

src/clients/python/CMakeLists.txt
GuanLuo: Otherwise there will be linker error: `unresolved external symbol "public: bool __cdecl google::protobuf::MessageLite::SerializeToString()`, this function is called [here](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/clients/python/crequest.cc#L249).
But this error can also be resolved if the "request" library link gRPC as PUBLIC instead of [PRIVATE](https://github.com/NVIDIA/tensorrt-inference-server/pull/403/files#diff-a4364faad3475ef0a2269ab8f0f606acR48), perhaps gRPC uses protobuf and thus exposes the symbol?

deadeyegoodwin: We don't want librequest to expose the protobuf symbols since that can cause problems when the main application is using their own protobuf... I see why we need it, but I wonder why we don't for linux?

GuanLuo: I am not sure, perhaps they use different strategy on exporting symbols? If I `nm librequest.so`, this symbol is included, and it is not the case on Windows (`dumpbin /EXPORTS request.dll`)

```
nm librequest.so | grep SerializeToString
000000000078eeec T _ZNK6google8protobuf11MessageLite17SerializeToStringEPNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```

deadeyegoodwin: I guess we should include an ldscript for librequest.so like we do for libtrtserver.so in src/servers/CMakeList.txt... but I'm a little hesitant because users will need our exact version of protobuf to use the library. Perhaps for consistency between linux and win32 we should just make libprotobuf public for librequest and then we don't need it explicitly in crequest.

PR #404, opened 2019-06-24
Add test on sending large payload. Fix exposed bugs

GuanLuo:


qa/L0_large_payload/large_payload_test.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_onnx(self):
+        tensor_shape = (self.input_size_,)
+        small_tensor_shape = (1,)
+
+        # onnx_nobatch_zero_1_float32 is identity model with input shape [-1]
+        for protocol, url in self.protocols_:
+            model_name = tu.get_zero_model_name("onnx_nobatch", 1, self.data_type_)
+            ctx = InferContext(url, protocol, model_name, None, True)
+            self._test_helper(ctx, tensor_shape, small_tensor_shape)
+
```
deadeyegoodwin: What about pytorch, tensorrt, custom?

qa/L0_large_payload/test.sh
```diff
@@ -0,0 +1,84 @@
...
+source ../common/util.sh
+
+rm -f $SERVER_LOG_BASE* $CLIENT_LOG_BASE*
+
+RET=0
+
+# Restart server before every test to make sure server state
+# is invariant to previous test
+#
+# [TODO] add LibTorch model and Plan model
```
deadeyegoodwin: Need to add these...

src/backends/tensorflow/base_backend.cc
```diff
@@ -593,6 +602,8 @@ BaseBackend::Context::Run(
   // Create a tensor for each input sized correctly for the total
   // payload batch size. Concatenate input values from each payload
   // into the corresponding tensor.
+  // [TODO] should be responsible for calling TRTISTF_TensorListDelete
+  //        if there is error and returning before calling TRTISTF_ModelRun
   TRTISTF_TensorList* input_tensors = nullptr;
```
deadeyegoodwin: Can you just wrap it in a unique_ptr?

qa/L0_large_payload/test.sh
```diff
@@ -0,0 +1,102 @@
...
+            echo "default_model_filename: \"libidentity.so\"" >> config.pbtxt && \
+            echo "instance_group [ { kind: KIND_CPU }]" >> config.pbtxt && \
+            sed -i "s/dims: \[ 1 \]/dims: \[ -1 \]/" config.pbtxt)
+
+# Restart server before every test to make sure server state
+# is invariant to previous test
+#
+# Skipping TensorRT Plan model for now as it only supports fixed size
+# tensor and it fails to generate layer with large dimension size
+# [TODO] Revisit this once TensorRT supports variable size tensor
```
deadeyegoodwin: Make sure to file a ticket for the TRT testing and refer to this file and how it needs to be updated.

qa/common/gen_qa_identity_models
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
```
deadeyegoodwin: So the gen_qa_zero_models.py is actually generating identity models. The "zero" name comes from its original use in only the zero testing, but the identity models are now used for much more. So you don't need a whole new file for this... you just need to add identity model for pytorch to existing file (or perhaps as zero file contents to this file so that the naming is better.. "identity" makes more sense than "zero").

qa/common/gen_qa_identity_models.py
```diff
@@ -0,0 +1,188 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Can you also ask Hemant about why he didn't already create an identity model for PyTorch. Perhaps he thought these models were only for "zero-sized tensor" testing and pytorch doesn't support that so he didn't create it. But he needs to make sure that we have pytorch testing in all the other places that use the identity model.

CoderHam: On it!

qa/L0_large_payload/test.sh
GuanLuo: Opened ticket DLIS-521

PR #408, opened 2019-06-25
Add support for TensorRT provided plugins

CoderHam:
- [x] Register provided plugins
- [x] Create custom model that uses one of the provided plugins
- [x] Add tests for plugin model registration


qa/L0_trt_plugin/test.sh
```diff
@@ -0,0 +1,82 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+CLIENT_LOG="./client_trt_plugin.log"
+PLUGIN_TEST=trt_plugin_test.py
+
+DATADIR=`pwd`/models
+
+SERVER=/opt/tensorrtserver/bin/trtserver
+# Allow more time to exit. Ensemble brings in too many models
```
GuanLuo: Using enesmble?

qa/L0_trt_plugin/test.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+# Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
```
GuanLuo: 2019

qa/L0_trt_plugin/trt_plugin_test.py
```diff
@@ -0,0 +1,61 @@
...
+        config = ("localhost:8000", ProtocolType.HTTP, False)
+        ctx = InferContext(config[0], config[1], model_name,
+                           correlation_id=0, streaming=config[2],
+                           verbose=True)
+        results = ctx.run(
+            { "INPUT0" : input_list }, { "OUTPUT0" : InferContext.ResultFormat.RAW},
+            batch_size=batch_size)
+
+        self.assertEqual(len(results), 1)
+        self.assertTrue("OUTPUT0" in results)
```
GuanLuo: Can you also verify output values here?

qa/common/gen_qa_plugin_models.py
```diff
@@ -0,0 +1,180 @@
...
+    elif np_dtype == np.float16:
+        return trt.infer.DataType.HALF
+    elif np_dtype == np.float32:
+        return trt.infer.DataType.FLOAT
+    return None
+
+def get_trt_plugin(plugin_name):
+        plugin = None
+        for plugin_creator in PLUGIN_CREATORS:
+            if plugin_creator.name == plugin_name:
```
GuanLuo: The following seems to be "LReLU_TRT" specific? What if you pass other plugin's name which is also in plugin_creator?

qa/L0_trt_plugin/test.sh
CoderHam: Nope. That was a typo. Removed

qa/common/gen_qa_plugin_models.py
CoderHam: We can create models that use other plugins but since all plugins are registered in one step, it is safe to say that if one plugin works then the others will as well.

qa/L0_trt_plugin/trt_plugin_test.py
CoderHam: Sure.

qa/common/gen_qa_plugin_models.py
GuanLuo: I understand that. What I was saying is that `get_trt_plugin` is rather general name and it accepts `plugin_name` which imply that it will returns the plugin that corresponds to the name, while the actual implementation is specific to ReLu...

src/backends/tensorrt/loader.cc
```diff
@@ -39,6 +42,23 @@ LoadPlan(
...
 
+  // Register all the default plugins that come with TensorRT: RPROI_TRT
+  // Normalize_TRT, PriorBox_TRT, GridAnchor_TRT, NMS_TRT, LReLU_TRT, Reorg_TRT,
+  // Region_TRT and Clip_TRT
+  std::once_flag onceFlag;
+  {
+    std::call_once(onceFlag, [] {
+      bool success = initLibNvInferPlugins(&tensorrt_logger.getTRTLogger(), "");
+      if (!success) {
+        return Status(
```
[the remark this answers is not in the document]
CoderHam: Understood

src/backends/tensorrt/logging.cc
```diff
@@ -54,4 +54,10 @@ TensorRTLogger::log(Severity severity, const char* msg)
   }
 }
 
+nvinfer1::ILogger&
```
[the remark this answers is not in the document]
CoderHam: Yes. Remnant of older iteration. Fixed now.

PR #417, opened 2019-06-26
Add http header option to client library

deadeyegoodwin:


src/clients/c++/request_http.cc
```diff
@@ -1130,6 +1233,10 @@ InferHttpContextImpl::PreRunProcessing(std::shared_ptr<Request>& request)
   list = curl_slist_append(list, "Expect:");
   list = curl_slist_append(list, "Content-Type: application/octet-stream");
   list = curl_slist_append(list, infer_request_str_.c_str());
+  for (const auto& pr : headers_) {
+    std::string hdr = pr.first + ": " + pr.second;
```
GuanLuo: What will happen if users pass in their own infer request header? Should we add check to prevent them from doing so?

src/clients/c++/request_http.h
```diff
@@ -117,6 +159,26 @@ class InferHttpContext {
       const std::string& model_name, int64_t model_version = -1,
       bool verbose = false);
 
+  /// Create context that performs inference for a non-sequence model
+  /// using HTTP protocol and custom headers.
+  ///
+  /// \param ctx Returns a new InferHttpContext object.
+  /// \param server_url The inference server name and port.
+  /// \param headers Map of HTTP headers to use with the status
```
GuanLuo: "... use with the infer request"

src/clients/c++/request_http.h
```diff
@@ -136,6 +198,31 @@ class InferHttpContext {
...
+  /// Create context that performs inference for a sequence model
+  /// using a given correlation ID and the HTTP protocol using custom
+  /// headers.
+  ///
+  /// \param ctx Returns a new InferHttpContext object.
+  /// \param correlation_id The correlation ID to use for all
+  /// inferences performed with this context. A value of 0 (zero)
+  /// indicates that no correlation ID should be used.
+  /// \param server_url The inference server name and port.
+  /// \param headers Map of HTTP headers to use with the status
```
GuanLuo: infer

src/clients/python/crequest.cc
```diff
@@ -103,27 +117,32 @@ struct ServerHealthContextCtx {
...
-    if (protocol == ProtocolType::HTTP) {
-      err = nic::ServerHealthHttpContext::Create(
-          &(lctx->ctx), std::string(url), verbose);
-    } else {
-      err = nic::ServerHealthGrpcContext::Create(
-          &(lctx->ctx), std::string(url), verbose);
-    }
 
+  std::map<std::string, std::string> http_headers;
+  err = ParseHttpHeaders(&http_headers, headers, num_headers);
```
GuanLuo: Shouldn't parse headers be done after checking that HTTP protocol is used?

src/clients/python/crequest.cc
```diff
@@ -167,37 +186,43 @@ struct ServerStatusContextCtx {
...
     ServerStatusContextCtx** ctx, const char* url, int protocol_int,
-    const char* model_name, bool verbose)
+    const char** headers, int num_headers, const char* model_name, bool verbose)
 {
   nic::Error err;
-  ProtocolType protocol;
-  err = ParseProtocol(&protocol, protocol_int);
+
+  std::map<std::string, std::string> http_headers;
+  err = ParseHttpHeaders(&http_headers, headers, num_headers);
```
GuanLuo: after ParseProtocol

src/clients/python/crequest.cc
```diff
@@ -243,38 +268,44 @@ struct InferContextCtx {
...
-          &(lctx->ctx), correlation_id, std::string(url),
-          std::string(model_name), model_version, verbose);
-    } else {
-      err = nic::InferGrpcContext::Create(
-          &(lctx->ctx), correlation_id, std::string(url),
-          std::string(model_name), model_version, verbose);
-    }
 
+  std::map<std::string, std::string> http_headers;
+  err = ParseHttpHeaders(&http_headers, headers, num_headers);
```
GuanLuo: after ParseProtocol

src/clients/c++/request_http.cc
deadeyegoodwin: I'll check for it during initialization of the context

src/clients/python/crequest.cc
deadeyegoodwin: Yes, fixed

PR #422, opened 2019-06-28
Create CustomInstance SDK and apply to Identity Custom Backend

dzier:
completed custom instance and identity backend


src/custom/sdk/custom_instance.cc
```diff
@@ -0,0 +1,145 @@
...
+  return InitContext();
+}
+
+int
+CustomInstance::Execute(
+    const uint32_t payload_cnt, CustomPayload* payloads,
+    CustomGetNextInputFn_t input_fn, CustomGetOutputFn_t output_fn)
+{
+  for (uint32_t pidx = 0; pidx < payload_cnt; ++pidx) {
+    ExecutePayload(payloads[pidx], input_fn, output_fn);
```
dzier: I think I need to add a check here after ExecutePayload, to see if `payloads[pidx].error_code == kSuccess`. Or better, just have ExecutePayload return the error and break the for loop on error.

src/custom/sdk/custom_instance.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace nvidia { namespace inferenceserver { namespace custom {
+
+// Integer error codes. TRTIS requires that success must be 0. All
+// other codes are interpreted by TRTIS as failures.
+enum ErrorCodes : int {
+  kSuccess = 0,
+  kUnknown,
+  kInvalidModelConfig,
+  kNumErrorCodes
```
dzier: Enums are one of those C++ features that cannot easily be extended. So my solution here is to have a _last_ error code `kNumErrorCodes` which is used to start another list of error codes. You can see this in the `identity.cc` file. Another alternative is to create a custom error class that is extensible, but this works for now.

PR #423, opened 2019-06-28
Create CustomInstance and apply to Identity

dzier:
completed custom instance and identity backend

Cleanup


src/custom/identity/identity.cc
```diff
@@ -154,138 +145,88 @@ Context::Init()
...
-      if (itr == copy_map_.end()) {
-        payload.error_code = kRequestOutput;
+    const std::string& input_name = itr->second.input_name_;
+    const DataType datatype = itr->second.datatype_;
+
+    std::vector<int64_t> shape;
+    if (model_config_.max_batch_size() != 0) {
+      shape.push_back(payload.batch_size);
+    }
+    for (uint32_t input_idx = 0; input_idx < payload.input_cnt; ++input_idx) {
```
GuanLuo: The step of "given an input name, extract its shape (and content)" can probably be abstracted to a helper function in the SDK. This should be two functions, one function only operates on one payload while the other operates on multiple payloads for dynamic batching.

.gitignore
```diff
@@ -1 +1,2 @@
 /builddir
+.vscode
```
GuanLuo: You can hide IDE configuration from git by adding another .gitignore file inside the ".vscode" folder, with "/*" as content. In this way, you hide the folder without modifying the top level .gitignore file.

src/custom/sdk/custom_instance.cc
```diff
@@ -0,0 +1,145 @@
...
+  return InitContext();
+}
+
+int
+CustomInstance::Execute(
+    const uint32_t payload_cnt, CustomPayload* payloads,
+    CustomGetNextInputFn_t input_fn, CustomGetOutputFn_t output_fn)
+{
+  for (uint32_t pidx = 0; pidx < payload_cnt; ++pidx) {
+    ExecutePayload(payloads[pidx], input_fn, output_fn);
```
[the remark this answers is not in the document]
dzier: @GuanLuo, is there an example of how payloads can be executed with dynamic batching? It seems like an easy feature to add.

src/custom/identity/identity.cc
dzier: I will look again. Originally I went down that route but found some of this code was not shared enough between the custom backends. Perhaps I can abstract some virtual methods for any customizations.

.gitignore
dzier: I will try that. Thank you.

src/custom/sdk/custom_instance.cc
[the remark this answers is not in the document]
GuanLuo: You can take a look at one of the framework backend, i.e. https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/backends/onnx/onnx_backend.cc#L610
The idea is basically concatenating the data of the same input name across payloads as one big input buffer.

.gitignore
dzier: reverted the changes

src/custom/sdk/custom_instance.cc
```diff
@@ -0,0 +1,145 @@
...
+  // Convert the serialized model config to a ModelConfig object.
+  ModelConfig model_config;
+  if (!model_config.ParseFromString(std::string(
+          data->serialized_model_config, data->serialized_model_config_size))) {
+    return kInvalidModelConfig;
+  }
+
+  // Create the instance and validate that the model configuration is
+  // something that we can handle.
+  CustomInstance* instance = CustomInstance::Create(data, model_config);
```
deadeyegoodwin: You want to have Create return an error code. I don't think you need a separate Create and Init. We should discuss this when you visit.

src/custom/identity/identity.cc
dzier: Ok, I will move it up to execute and let the child class define if is per payload or batch. I did notice in other backend, this top level for loop doesn't exist.

src/custom/identity/identity.cc
```diff
@@ -61,29 +59,14 @@ enum ErrorCodes {
...
  private:
-  // The name of this instance of the backend.
-  const std::string instance_name_;
-
-  // The model configuration.
-  const ModelConfig model_config_;
-
   // The GPU device ID to execute on or CUSTOM_NO_GPU_DEVICE if should
   // execute on CPU.
   const int gpu_device_;
```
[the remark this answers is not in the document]
dzier: The gpu_device? Because in the examples it is only identity that requires it. So it is unique to the identity class.

src/custom/identity/identity.cc
```diff
@@ -30,6 +30,7 @@
 #include "src/backends/custom/custom.h"
```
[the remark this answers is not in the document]
dzier: Correct, will remove

src/custom/identity/identity.cc
```diff
@@ -96,20 +79,28 @@ class Context {
...
-      gpu_device_(gpu_device)
+    : CustomInstance(instance_name, model_config), gpu_device_(gpu_device)
 {
 }
 
-Context::~Context() {}
-
 int
-Context::Init()
+Context::InitContext()
```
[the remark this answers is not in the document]
dzier: True, however, I was looking for a hook to init any objects that may be common to all instances. The top level CustomInstance class has public Init() already that will call the common init (if any) and then call the private InitContext, which is overridden by any derived class.  

However, right now, this is more forward thinking as there is nothing that is commonly initiated. See custom_instance.cc:38.  So I could revert this and just keep the public method as a virtual method.

Thoughts?

src/custom/identity/identity.cc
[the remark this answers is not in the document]
deadeyegoodwin: In general the device tells you what CPU/GPU device the instance should be created on. It is needed by all backends.

src/custom/sdk/CMakeLists.txt
```diff
@@ -62,12 +64,12 @@ install(
 # backend
 install(
```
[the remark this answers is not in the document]
dzier: Isn't the library installed on line 58? Where is the best place to install custom_instance.*?

src/custom/identity/identity.cc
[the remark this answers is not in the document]
deadeyegoodwin: I'd like to discuss in person since you are coming down.

src/custom/sdk/custom_instance.cc
```diff
@@ -0,0 +1,145 @@
...
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "custom_instance.h"
+
+namespace nvidia { namespace inferenceserver { namespace custom {
+
+CustomInstance::CustomInstance(
+    const std::string& instance_name, const ModelConfig& model_config)
```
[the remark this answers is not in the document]
dzier: So I see some examples with device and some without. I can add device easily enough, however, I may follow up with you on how to handle backends without devices.

src/custom/sdk/custom_instance.cc
dzier: I agree. No need for a separate instance. I can fix that. Do you prefer returning the object as a pointer parameter, or return a `std::pair<CustomInstance *, int>`?

src/custom/sdk/custom_instance.cc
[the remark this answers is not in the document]
dzier: Going to go with David's suggestion and move the virtual method to Execute. It is up to the sub-class to decide if they want to handle the code individually or in one execution.

src/custom/sdk/custom_instance.cc
dzier: Actually, I am going with a member error_code_ variable that can be set during construction.

deadeyegoodwin: Preferred style will be to return Error, and pass in a CustomInstance* which the callee is expected to use to return the new instance object.

src/custom/identity/identity.cc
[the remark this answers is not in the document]
dzier: Removed all together

src/custom/identity/identity.cc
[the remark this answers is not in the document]
dzier: removed

src/custom/identity/identity.cc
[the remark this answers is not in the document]
dzier: added to base class

src/custom/sdk/custom_instance.cc
dzier: Create now returns an error and passes the instance as output return pointer.

src/custom/sdk/custom_instance.cc
```diff
@@ -0,0 +1,107 @@
...
+    return ErrorCodes::Unknown;
+  }
+
+  CustomInstance* instance = static_cast<CustomInstance*>(custom_instance);
+  return instance->Execute(payload_cnt, payloads, input_fn, output_fn);
+}
+
+}  // extern "C"
+
+}}}  // namespace nvidia::inferenceserver::custom
```
deadeyegoodwin: Put CR/LF at end of file

src/custom/sdk/custom_instance.h
```diff
@@ -0,0 +1,86 @@
...
+#include "src/core/model_config.h"
+#include "src/core/model_config.pb.h"
+#include "src/custom/sdk/error_codes.h"
+
+namespace nvidia { namespace inferenceserver { namespace custom {
+
+// Base class for custom backend instances
+// Responsible for the state of the instance and used to provide a
+// C++ wrapper around the C-API
+class CustomInstance {
```
deadeyegoodwin: You need to document this entire thing using our doc format (see custom.h or request.h) since we are going to want to generate online documentation for this interface.

src/custom/sdk/custom_instance.h
```diff
@@ -0,0 +1,86 @@
...
+  // The GPU device ID to execute on or CUSTOM_NO_GPU_DEVICE if should
+  // execute on CPU.
+  const int gpu_device_;
+
+ private:
+  // Error code manager
+  ErrorCodes errors_;
+};
+
+}}}  // namespace nvidia::inferenceserver::custom
```
deadeyegoodwin: CR/LF

src/custom/sdk/error_codes.cc
```diff
@@ -0,0 +1,55 @@
...
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#include "src/custom/sdk/error_codes.h"
+
+namespace nvidia { namespace inferenceserver { namespace custom {
+
+ErrorCodes::ErrorCodes() {
```
deadeyegoodwin: Need to run formatter

src/custom/sdk/error_codes.cc
```diff
@@ -0,0 +1,55 @@
...
+  RegisterError("success");
+  RegisterError("unknown error");
+  RegisterError("failed to create instance");
+  RegisterError("invalid model configuration");
+}
+
+const char *
+ErrorCodes::ErrorString(int error) const
+{
+    if (error < 0 || error >= static_cast<int>(err_names_.size())) {
```
deadeyegoodwin: parens around if sub-expressions

Shouldn't this be (error >=) && (error < ...)

src/custom/sdk/error_codes.h
```diff
@@ -0,0 +1,58 @@
...
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace nvidia { namespace inferenceserver { namespace custom {
+
+// Integer error codes. TRTIS requires that success must be 0. All
```
deadeyegoodwin: Document for online docs

src/custom/sdk/error_codes.h
```diff
@@ -0,0 +1,58 @@
...
+// other codes are interpreted by TRTIS as failures.
+class ErrorCodes {
+ public:
+  // Default error codes
+  static const int Success = 0;
```
deadeyegoodwin: This is a little brittle since you are relying on the initialization order in ErrorCodes constructor to match this order. At the very least you should put that in the comment.

src/custom/sdk/custom_instance.cc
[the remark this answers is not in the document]
dzier: I will be pulling these helper methods out in a separate PR

src/custom/sdk/error_codes.cc
dzier: Yes it should. Good catch. Fixed

src/custom/sdk/error_codes.h
dzier: done

src/custom/sdk/error_codes.h
dzier: I attempted to make is less brittle, Please review and comment. Still not 100% happy, but I need the static ints.

src/custom/sdk/error_codes.h
```diff
@@ -0,0 +1,87 @@
...
+  ///
+  /// \param error The error code
+  /// \param error_string The error message
+  void RegisterError(int error, const std::string& error_string);
+
+  /// \param error Error code.
+  /// \return True if error code is registered
+  inline bool ErrorIsRegistered(int error) const
+  {
+    return (error > 0) && (error < static_cast<int>(err_messages_.size()));
```
deadeyegoodwin: Isn't it error >= 0?

src/custom/sdk/error_codes.cc
```diff
@@ -0,0 +1,64 @@
...
+#include "src/custom/sdk/error_codes.h"
+
+namespace nvidia { namespace inferenceserver { namespace custom {
+
+ErrorCodes::ErrorCodes()
+{
+  RegisterError(Success, "success");
+  RegisterError(CreationFailure, "failed to create instance");
+  RegisterError(InvalidModelConfig, "invalid model configuration");
+  RegisterError(Unknown, "Unknown Error");
```
deadeyegoodwin: Don't capitialize "unknown" or "error"

src/custom/sdk/error_codes.h
dzier: yes, fixed

src/custom/sdk/error_codes.cc
dzier: fixed

PR #426, opened 2019-07-01
Updates to client build doc

deadeyegoodwin:


docs/client.rst
```diff
@@ -109,43 +112,54 @@ scripts is at /workspace/v<version>.clients.tar.gz.
...
-The actual client build is performed using CMake. The build
-dependencies and requirements are shown in Dockerfile.client. To build
-without Docker you must first install those dependencies. The CMake
-build can also be targeted for other OSes and platforms. We welcome
-any updates that expand the build functionality and allow the clients
-to be built on additional platforms.
+The client build is performed using CMake. The build dependencies and
+requirements are shown in Dockerfile.client. To build without Docker
+you must first install those dependencies. This section describes the
+client build for Ubuntu 16.04, Ubuntu 18.04, and Windows 10
```
GuanLuo: Not sure if we need to mention it. I install dependencies on Windows using [vcpkg](https://github.com/microsoft/vcpkg), which is basically a "apt-get" style tool for installing packages.

deadeyegoodwin: Not for now. If people are having problems we can add more detail.

PR #429, opened 2019-07-02
Delete finished message to release allocated memory for gRPC requests

GuanLuo:
From [protobuf's documentation](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.Clear.details), `Clear()` will not release the allocated memory as it assumes that the message will be reused for next request soon, which is not the case in our application.


seovchinnikov:
Thanks, that was a big problem

PR #431, opened 2019-07-02
Support for using system Shared Memory for I/O 

CoderHam:
Allow option to:

- Read (selected) Inputs from Shared Memory
- Write (selected) Outputs to Shared Memory

- [x] Update API request / response 
Support batching using offsets for each element with the same shm_key
Support for registering / unregistering SHM regions

### DIFF
--- a/src/core/api.proto
+++ b/src/core/api.proto
@@ -94,6 +94,24 @@ message InferRequestHeader
     //@@       for tensors with a non-fixed-size datatype (like STRING).
     //@@
     uint64 batch_byte_size = 3;
+
+    //@@  .. cpp:var:: message SharedMemoryIO
+    //@@
+    //@@     Unique identity of the location in shared memory to read input
+    //@@     tensors. If present, do no provide the data directly in the request
+    //@@     payload, it will be ignored. User responsible for the creation,
+    //@@     deletion and writing to shared memory.
+    //@@
+    message SharedMemoryIO
+    {
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The unique key used to refer to a specific location in shared
+      //@@     memory. Optional for tensors. Required to read tensors from
+      //@@     shared memory. If provided, tensor data read from this location.
+      //@@
+      key_t shm_key = 1;
+    }
   }
 
   //@@  .. cpp:var:: message Output
@@ -132,6 +150,23 @@ message InferRequestHeader
     //@@       highest probabilities will be returned.
     //@@
     Class cls = 3;
+
+    //@@  .. cpp:var:: message SharedMemoryIO
+    //@@
+    //@@     Unique identity of the location in shared memory to write output
+    //@@     tensors. User responsible for the creation, deletion and reading
+    //@@     from shared memory.
+    //@@
+    message SharedMemoryIO
+    {
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The unique key used to refer to a specific location in shared
+      //@@     memory. Optional for tensors. Required to write tensors to
+      //@@     shared memory. If provided, tensor data read from this location.
+      //@@
+      key_t shm_key = 1;
+    }
   }
 
   //@@  .. cpp:var:: uint64 id
@@ -269,7 +304,8 @@ message InferResponseHeader
     //@@       If specified deliver results for this output as raw tensor data.
     //@@       The actual output data is delivered in the HTTP body for an HTTP
     //@@       request, or in the :cpp:var:`InferResponse` message for a gRPC
-    //@@       request. Only one of 'raw' and 'batch_classes' may be specified.
+    //@@       request. Only one of 'raw', 'batch_classes' and 'SharedMemoryIO'
+    //@@       may be specified.
     //@@
     Raw raw = 2;
 
@@ -277,10 +313,27 @@ message InferResponseHeader
     //@@
     //@@       If specified deliver results for this output as classifications.
     //@@       There is one :cpp:var:`Classes` object for each batch entry in
-    //@@       the output. Only one of 'raw' and 'batch_classes' may be
-    //@@       specified.
+    //@@       the output. Only one of 'raw', 'batch_classes' and
+    //@@       'SharedMemoryIO' may be specified.
     //@@
     repeated Classes batch_classes = 3;
+
+    //@@  .. cpp:var:: message SharedMemoryIO
+    //@@
+    //@@     Unique identity of the location in shared memory to write output
+    //@@     tensors. User responsible for the creation, deletion and reading
+    //@@     from shared memory.
+    //@@
+    message SharedMemoryIO
+    {
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The unique key used to refer to a specific location in shared
+      //@@     memory. Optional for tensors. Required to write tensors to
+      //@@     shared memory. If provided, tensor data read from this location.
+      //@@
+      key_t shm_key = 1;
+    }
   }
 
   //@@  .. cpp:var:: uint64 id
--- a/src/core/api.proto
+++ b/src/core/api.proto
@@ -111,6 +111,16 @@ message InferRequestHeader
       //@@     shared memory. If provided, tensor data read from this location.
       //@@
       key_t shm_key = 1;
+
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The offset for the next element in the batch. Only required if
+      //@@     batch_size > 1. n-1 offsets required (we know first element has
+      //@@     offset of 0), where n is number of elements in the batch.
+      //@@     element0 start = 0, end = offset0;
+      //@@     element1 start = offset0, end = offset1; and so on.
+      //@@
+      repeated uint64 offset = 1;
     }
   }
 
@@ -166,6 +176,16 @@ message InferRequestHeader
       //@@     shared memory. If provided, tensor data read from this location.
       //@@
       key_t shm_key = 1;
+
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The offset for the next element in the batch. Only required if
+      //@@     batch_size > 1. n-1 offsets required (we know first element has
+      //@@     offset of 0), where n is number of elements in the batch.
+      //@@     element0 start = 0, end = offset0;
+      //@@     element1 start = offset0, end = offset1; and so on.
+      //@@
+      repeated uint64 offset = 1;
     }
   }
 
@@ -333,6 +353,16 @@ message InferResponseHeader
       //@@     shared memory. If provided, tensor data read from this location.
       //@@
       key_t shm_key = 1;
+
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The offset for the next element in the batch. Only required if
+      //@@     batch_size > 1. n-1 offsets required (we know first element has
+      //@@     offset of 0), where n is number of elements in the batch.
+      //@@     element0 start = 0, end = offset0;
+      //@@     element1 start = offset0, end = offset1; and so on.
+      //@@
+      repeated uint64 offset = 1;
     }
   }
 
--- a/src/core/api.proto
+++ b/src/core/api.proto
@@ -95,14 +95,11 @@ message InferRequestHeader
     //@@
     uint64 batch_byte_size = 3;
 
-    //@@  .. cpp:var:: message SharedMemoryIO
+    //@@      .. cpp:var:: message SharedMemoryElement
     //@@
-    //@@     Unique identity of the location in shared memory to read input
-    //@@     tensors. If present, do no provide the data directly in the request
-    //@@     payload, it will be ignored. User responsible for the creation,
-    //@@     deletion and writing to shared memory.
+    //@@         The Meta-data for the Input shared memory to read element from.
     //@@
-    message SharedMemoryIO
+    message SharedMemoryElement
     {
       //@@    .. cpp:var:: key_t shm_key
       //@@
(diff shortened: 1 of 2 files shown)


src/core/api.proto
```diff
@@ -94,6 +94,55 @@ message InferRequestHeader
     //@@       for tensors with a non-fixed-size datatype (like STRING).
     //@@
     uint64 batch_byte_size = 3;
+
+    //@@      .. cpp:var:: message SharedMemoryElement
+    //@@
+    //@@         The Meta-data for the Input shared memory to read element from.
+    //@@
+    message SharedMemoryElement
```
deadeyegoodwin: SharedMemoryElement and SharedMemoryIO should move to top-level so they can be shared for both input and output. Call them InferSharedMemoryElement and InferSharedMemory.

src/core/api.proto
```diff
@@ -94,6 +94,55 @@ message InferRequestHeader
...
+    {
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The unique key used to refer to a specific location in shared
+      //@@     memory. Optional for tensors. Required to read tensors from
+      //@@     shared memory. If provided, tensor data read from this location.
+      //@@
+      key_t shm_key = 1;
```
deadeyegoodwin: Is key_t a protobuf type? Should it just be "string"?

src/core/api.proto
```diff
@@ -94,6 +94,55 @@ message InferRequestHeader
...
+
+    //@@      .. cpp:var:: message SharedMemoryElement
+    //@@
+    //@@         The Meta-data for the Input shared memory to read element from.
+    //@@
+    message SharedMemoryElement
+    {
+      //@@    .. cpp:var:: key_t shm_key
+      //@@
+      //@@     The unique key used to refer to a specific location in shared
```
deadeyegoodwin: It's the name of the shared memory region that holds the input data (or where the output data should be written). It is not a specific location. You should drop the optional part. Specifying shared memory at all is optional, but if you add this message you must fill in all fields.

src/core/api.proto
```diff
@@ -94,6 +94,55 @@ message InferRequestHeader
...
+      //@@     memory. Optional for tensors. Required to read tensors from
+      //@@     shared memory. If provided, tensor data read from this location.
+      //@@
+      key_t shm_key = 1;
+
+      //@@    .. cpp:var:: uint64 offset
+      //@@
+      //@@     The offset for the each element.
```
deadeyegoodwin: offset from the start of the shared memory region. Not for each element... this comment needs updating.

src/core/api.proto
```diff
@@ -94,6 +94,55 @@ message InferRequestHeader
...
+      //@@      .. cpp:var:: SharedMemoryElement element (repeated)
+      //@@
+      //@@         No. of elements required is number of elements in the batch.
+      //@@
+      repeated SharedMemoryElement element = 1;
+    }
+
+    //@@    .. cpp:var:: SharedMemoryIO shm_io
+    //@@
+    //@@       Unique identity of the location in shared memory to read input
```
deadeyegoodwin: This should say optional. And then just something like "location of the input tensor in shared memory". Don't talk about responsibilities here.

src/core/api.proto
```diff
@@ -269,7 +367,9 @@ message InferResponseHeader
     //@@       If specified deliver results for this output as raw tensor data.
     //@@       The actual output data is delivered in the HTTP body for an HTTP
     //@@       request, or in the :cpp:var:`InferResponse` message for a gRPC
-    //@@       request. Only one of 'raw' and 'batch_classes' may be specified.
+    //@@       request. Only one of 'raw' and 'batch_classes' may be
```
deadeyegoodwin: Raw still needs to be used if results are delivered via shared memory since it delivers meta-data associated with those results. This comment and the one below don't need to change.

src/core/api.proto
CoderHam: No it isn't. Fixed.

src/core/api.proto
```diff
@@ -30,6 +30,47 @@ package nvidia.inferenceserver;
 
 //@@.. cpp:namespace:: nvidia::inferenceserver
 
+//@@      .. cpp:var:: message InferSharedMemoryElement
```
deadeyegoodwin: Comment indenting (spaces between //@@ and text) is wrong here. Look at an existing message to see how it should be at each level of your new messages. We have to be precise about this because of the way the doc generator works for protobuf files.

src/core/api.proto
```diff
@@ -30,6 +30,47 @@ package nvidia.inferenceserver;
...
+  //@@    .. cpp:var:: uint64 offset
+  //@@
+  //@@       The offset from the start of the shared memory region.
+  //@@       element0 start = offset0, end = offset0 + size;
+  //@@
+  uint64 offset = 2;
+
+  //@@    .. cpp:var:: uint64 byte_size
+  //@@
+  //@@       Size of the element in the batch of the input tensor, in bytes.
```
deadeyegoodwin: "Size of the element, in bytes". Don't talk about batch or input since that is specific and this is a generic structure.

src/core/api.proto
```diff
@@ -30,6 +30,47 @@ package nvidia.inferenceserver;
...
+
+//@@    .. cpp:var:: message InferSharedMemory
+//@@
+//@@       Meta-data for I/O of batch from shared memory.
+//@@
+message InferSharedMemory
+{
+  //@@      .. cpp:var:: InferSharedMemoryElement element (repeated)
+  //@@
+  //@@         No. of elements required is the size of the batch.
```
deadeyegoodwin: "One or more contiguous shared memory blocks that compose the entire shared memory"

How about change InferSharedMemoryElement to InferSharedMemoryBlock since that more accurately captures what it is.

src/core/api.proto
```diff
@@ -94,6 +135,14 @@ message InferRequestHeader
     //@@       for tensors with a non-fixed-size datatype (like STRING).
     //@@
     uint64 batch_byte_size = 3;
+
+    //@@    .. cpp:var:: InferSharedMemory shm_io
+    //@@
+    //@@       It is the location in shared memory to read the input tensors.
```
deadeyegoodwin: "... location in shared memory that contains the tensor data for this input"

shm_io -> shared_memory

src/core/api.proto
```diff
@@ -132,6 +181,14 @@ message InferRequestHeader
     //@@       highest probabilities will be returned.
     //@@
     Class cls = 3;
+
+    //@@    .. cpp:var:: InferSharedMemory shm_io
+    //@@
+    //@@       It is the location in shared memory to write the output tensors.
```
deadeyegoodwin: ".. location in shared memory that the result tensor data for this output will be written"

shm_io -> shared_memory

src/core/api.proto
CoderHam: Yes I think Block would capture the meaning better

src/core/api.proto
CoderHam: Fixed

src/core/api.proto
```diff
@@ -94,6 +122,15 @@ message InferRequestHeader
     //@@       for tensors with a non-fixed-size datatype (like STRING).
     //@@
     uint64 batch_byte_size = 3;
+
+    //@@    .. cpp:var:: InferSharedMemory shared_memory
+    //@@
+    //@@       It is the location in shared memory that contains the tensor data
+    //@@       for this input. Using shared memory is optional but if this
+    //@@       message is used, all fields are required. The memory must be
```
deadeyegoodwin: It is too confusing to try to explain contiguous vs non-contiguous here. So drop everything after "... all fields are required".

src/core/api.proto
```diff
@@ -132,6 +169,15 @@ message InferRequestHeader
     //@@       highest probabilities will be returned.
     //@@
     Class cls = 3;
+
+    //@@    .. cpp:var:: InferSharedMemory shared_memory
+    //@@
+    //@@       It is the location in shared memory that the result tensor data
+    //@@       for this output will be written. Using shared memory is optional
+    //@@       but if this message is used, all fields are required. The memory
+    //@@       must be continuous. For non-contiguous, use Dynamic Batching.
```
deadeyegoodwin: It is too confusing to try to explain contiguous vs non-contiguous here. So drop everything after "... all fields are required".

src/core/api.proto
```diff
@@ -94,6 +94,13 @@ message InferRequestHeader
     //@@       for tensors with a non-fixed-size datatype (like STRING).
     //@@
     uint64 batch_byte_size = 3;
+
+    //@@    .. cpp:var:: string shm_name
+    //@@
+    //@@       The name given by the user during registration of a shared memory
+    //@@       region that holds the input data.
+    //@@
+    string shm_name = 4;
```
deadeyegoodwin: We still need the offset and size. User can register a single shared memory region and use (parts of) it for different inputs and outputs, even across multiple requests. So I thing you still need the InferSharedMemory message but the with "key" being the registered shared memory region name.

src/core/grpc_service.proto
```diff
@@ -65,6 +65,12 @@ service GRPCService
   //@@
   rpc Control(ControlRequest) returns (ControlResponse) {}
 
+  //@@  .. cpp:var:: rpc ShmControl(ShmControlRequest) returns (ShmControlResponse)
+  //@@
+  //@@     Request to register / unregister a specified shared memory region.
+  //@@
+  rpc ShmControl(ShmControlRequest) returns (ShmControlResponse) {}
```
deadeyegoodwin: SharedMemoryControl.. don't use Shm abbreviation anywhere, just spell out SharedMemory.

src/core/grpc_service.proto
```diff
@@ -65,6 +65,12 @@ service GRPCService
   //@@
   rpc Control(ControlRequest) returns (ControlResponse) {}
```
deadeyegoodwin: This should change to ModelControl, ModelControlResponse, etc. And the endpoint should be /api/modelcontrol. You should do that in a separate change and Guan and review to make sure you get all the cases.

src/core/grpc_service.proto
```diff
@@ -250,6 +256,100 @@ message ControlResponse
   RequestStatus request_status = 1;
 }
 
+//@@.. cpp:var:: message InferSharedMemory
+//@@
+//@@   The meta-data for the shared memory from which to read the input
+//@@   data and/or write the output data.
+//@@
+message InferSharedMemory
```
deadeyegoodwin: Perhaps InferSharedMemoryRegion. Need a "string name" field so that the user can give a name to the region that they then refer to in their input/output requests.

src/core/grpc_service.proto
```diff
@@ -250,6 +256,100 @@ message ControlResponse
...
+    REGISTER = 1;
+  }
+
+  //@@
+  //@@  .. cpp:var:: string model_name
+  //@@
+  //@@     The target shared memory region name (different from the shm_key)
+  //@@     that the user will use to refer to the specified region.
+  //@@
+  string shm_name = 1;
```
deadeyegoodwin: I would put this inside InferSharedMemoryRegion. The reason is that we need a "status" API as well that returns all the shared memory regions and the response of that can just be a repeated of InferSharedMemoryRegion.

src/core/grpc_service.proto
```diff
@@ -250,6 +256,100 @@ message ControlResponse
...
+  enum Type {
+    //@@    .. cpp:enumerator:: Type::UNREGISTER = 0
+    //@@
+    //@@       To unregister the specified shared memory region.
+    //@@
+    UNREGISTER = 0;
+
+    //@@    .. cpp:enumerator:: Type::REGISTER = 1
+    //@@
+    //@@       To register the specified shared memory region. If used the
```
deadeyegoodwin: Remove "If used ..." sentence and instead in the description of shared_memory=4 below can mention what is required for REGISTER and UNREGISTER.

src/core/api.proto
CoderHam: Understood

src/core/grpc_service.proto
CoderHam: will do

src/core/grpc_service.proto
CoderHam: Sure

CoderHam: PR for the rename from Control to ModelControl is here: #488

src/core/api.proto
```diff
@@ -30,6 +30,34 @@ package nvidia.inferenceserver;
...
+//@@   data and/or write the output data.
+//@@
+message InferSharedMemory
+{
+  //@@  .. cpp:var:: string shm_name
+  //@@
+  //@@     The name given during registration of a shared memory region that
+  //@@     holds the input data (or where the output data should be written).
+  //@@
+  string shm_name = 1;
```
deadeyegoodwin: Just "name"

src/core/grpc_service.proto
```diff
@@ -250,6 +260,99 @@ message ControlResponse
...
+//@@
+message InferSharedMemoryRegion
+{
+  //@@
+  //@@  .. cpp:var:: string model_name
+  //@@
+  //@@     The target shared memory region name (different from the shm_key)
+  //@@     that the user will use to refer to the specified region.
+  //@@
+  string shm_name = 1;
```
deadeyegoodwin: Just  "name"

src/core/grpc_service.proto
```diff
@@ -250,6 +260,99 @@ message ControlResponse
...
+//@@
+//@@   The meta-data for the shared memory region from which to read the input
+//@@   data and/or write the output data.
+//@@
+message InferSharedMemoryRegion
+{
+  //@@
+  //@@  .. cpp:var:: string model_name
+  //@@
+  //@@     The target shared memory region name (different from the shm_key)
```
deadeyegoodwin: "The name for this shared memory region."

src/core/grpc_service.proto
```diff
@@ -250,6 +260,99 @@ message ControlResponse
...
 
+//@@.. cpp:var:: message InferSharedMemoryRegion
+//@@
+//@@   The meta-data for the shared memory region from which to read the input
+//@@   data and/or write the output data.
+//@@
+message InferSharedMemoryRegion
+{
+  //@@
+  //@@  .. cpp:var:: string model_name
```
deadeyegoodwin: model_name -> name

src/core/grpc_service.proto
```diff
@@ -250,6 +260,99 @@ message ControlResponse
...
+    //@@    .. cpp:enumerator:: Type::REGISTER = 1
+    //@@
+    //@@       To register the specified shared memory region.
+    //@@
+    REGISTER = 1;
+  }
+
+  //@@  .. cpp:var:: InferSharedMemoryRegion shared_memory_region
+  //@@
+  //@@     It is the region in shared memory that the user wishes to register.
```
deadeyegoodwin: Change first sentence to: "The shared memory region to register or unregister."

src/core/grpc_service.proto
```diff
@@ -250,6 +260,99 @@ message ControlResponse
...
+  //@@     All fields are needed to REGISTER the shared memory region.
+  //@@     Only 'shm_name' is needed to UNREGISTER the shared memory region.
```
deadeyegoodwin: "Only 'name' is ..."

src/core/grpc_service.proto
```diff
@@ -250,6 +260,99 @@ message ControlResponse
...
+  //@@
+  InferSharedMemoryRegion shared_memory_region = 1;
+
+  //@@
+  //@@  .. cpp:var:: Type type
+  //@@
+  //@@     The control type that states whether to register or unregister the
+  //@@     specified shared memory region.
+  //@@
+  Type type = 2;
```
deadeyegoodwin: Can you make 'type' first in the message and give it value 1.

deadeyegoodwin:
@CoderHam Do you have a CI running for this? Once that passes I can merge.

CoderHam:
> @CoderHam Do you have a CI running for this? Once that passes I can merge.

I re-ran with the new changes.

deadeyegoodwin:
@CoderHam The CI failed

CoderHam:
Fixed and re-ran

PR #436, opened 2019-07-03
Allow autofill for multiple model versions

CoderHam:
Relax Autofill check to allow for multiple model versions. 
PS: Need to add aggressive checks across versions to make it more robust. Currently uses only the first version of the model.

### DIFF
--- a/src/backends/onnx/autofill.cc
+++ b/src/backends/onnx/autofill.cc
@@ -299,10 +299,6 @@ AutoFillOnnx::Create(
   std::set<std::string> version_dirs;
   RETURN_IF_ERROR(GetDirectorySubdirs(model_path, &version_dirs));
 
-  // There must be at least one version directory that we can inspect to
-  // attempt to determine the platform. For now we allow multiple versions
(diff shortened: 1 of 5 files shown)


src/backends/onnx/autofill.cc
```diff
@@ -299,6 +299,10 @@ AutoFillOnnx::Create(
   std::set<std::string> version_dirs;
   RETURN_IF_ERROR(GetDirectorySubdirs(model_path, &version_dirs));
 
+  // There must be at least one version directory that we can inspect to
+  // attempt to determine the platform. For now we allow multiple versions
+  // and only inspect the first verison directory to ensure it is valid.
```
GuanLuo: Not true for ONNX, see code below.

deadeyegoodwin: @CoderHam You should just remove this comment... the ONNX code is ok as is.

CoderHam: Sure. Will do.

PR #437, opened 2019-07-03
Add virtual device support to tensorflow backend

xprotobeast2:


src/backends/tensorflow/graphdef_backend_factory.cc
```diff
@@ -69,10 +69,18 @@ GraphDefBackendFactory::CreateBackend(
         std::make_tuple(graphdef_path));
   }
 
+  // Initialize virtual device counter
```
deadeyegoodwin: I forgot that there is both graphdef and savedmodel factories. You don't want 2 vgpu id maps, just one that is shared. So you should make a singleton utility class that is shared that holds the gpu_device -> vgpu_device logic.

src/backends/tensorflow/graphdef_backend_factory.cc
```diff
@@ -69,10 +69,18 @@ GraphDefBackendFactory::CreateBackend(
...
+  for (size_t gpu_idx = 0; gpu_idx < backend_config_->memory_limit_mb.size();
+       gpu_idx++) {
+    virtual_device_ids_.emplace(
```
deadeyegoodwin: We need a map from physical GPU -> virtual GPU. This seems to be a map from virtual GPU -> ?.

src/backends/tensorflow/graphdef_backend_factory.cc
```diff
@@ -69,10 +69,18 @@ GraphDefBackendFactory::CreateBackend(
...
+        std::piecewise_construct, std::forward_as_tuple(gpu_idx),
+        std::forward_as_tuple(0));
+  }
+
   std::unique_ptr<GraphDefBackend> local_backend(new GraphDefBackend);
   RETURN_IF_ERROR(local_backend->Init(path, model_config));
-  RETURN_IF_ERROR(
-      local_backend->CreateExecutionContexts(backend_config_, graphdef_paths));
+  RETURN_IF_ERROR(local_backend->CreateExecutionContexts(
+      backend_config_, graphdef_paths, &virtual_device_ids_));
```
deadeyegoodwin: Don't pass virtual_device_ids_ to CreateExecutionContexts. virtual_device_ids_ should be hide inside the utility class and it should expose a method can can be called to get the "next" virtual id for a given pysical gpu_device.

src/backends/tensorflow/autofill.cc
```diff
@@ -295,7 +295,8 @@ AutoFillSavedModel::Create(
       TRTISTF_NO_GPU_DEVICE, false /* have_graph */, 0 /* graph_level */,
       graphdef_backend_config->allow_gpu_memory_growth,
       graphdef_backend_config->per_process_gpu_memory_fraction,
-      graphdef_backend_config->allow_soft_placement));
+      graphdef_backend_config->allow_soft_placement,
+      graphdef_backend_config->memory_limit_mb));
```
deadeyegoodwin: This can't be just a single value. The user is going to need to be able to specify a value for each physical GPU. We can update this as part of the other ticket.

src/backends/tensorflow/base_backend.cc
```diff
@@ -167,11 +171,19 @@ BaseBackend::CreateExecutionContext(
...
+      graphdef_backend_config->memory_limit_mb[gpu_device].size();
+  int virtual_device =
+      ((*virtual_device_ids)[gpu_device]++) % num_virtual_devices_on_device;
+
   RETURN_IF_ERROR(context->ValidateInputs(Config().input()));
   RETURN_IF_ERROR(context->ValidateOutputs(Config().output()));
 
   RETURN_IF_ERROR(CreateTRTISTFModel(
-      graphdef_backend_config, gpu_device, Config().optimization().has_graph(),
+      graphdef_backend_config, virtual_device,
```
deadeyegoodwin: virtual_device -> vgpu_device for consistency

src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -515,7 +528,7 @@ TRTISTF_TensorNew(
   ConvertShape(shape, &tfshape);
 
   TensorImpl* tensor = new TensorImpl(name, dtype, shape, tfshape);
-  // If data type is non-string, make sure TensorImpl contains valid TF tensor
+    // If data type is non-string, make sure TensorImpl contains valid TF tensor
```
deadeyegoodwin: indent?

src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -165,14 +165,27 @@ NewSessionOptions(
     const bool allow_gpu_memory_growth,
     const float per_process_gpu_memory_fraction,
     const bool allow_soft_placement,
+    const std::vector<std::vector<float>>& memory_limit_mb,
     tensorflow::SessionOptions* session_options)
 {
+  tensorflow::ConfigProto* conf = &(session_options->config);
```
deadeyegoodwin: Why create "conf"? Rest of function just uses session_options->config.

src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -165,14 +165,27 @@ NewSessionOptions(
...
+  (*conf->mutable_device_count())["GPU"] = memory_limit_mb.size();
```
deadeyegoodwin: Move this down below "create virtual devices" comment.

src/backends/tensorflow/autofill.cc
xprotobeast2: memory_limit_mb should be a vector of vectors of floats. Could we parse this from the user somehow?

src/backends/tensorflow/graphdef_backend_factory.cc
xprotobeast2: This does map [0, #gpus) ->  [0, #vgpus). The memory_limit_mb vector has #gpus elements in it.

src/backends/tensorflow/tf_virtual_device.cc
```diff
@@ -0,0 +1,56 @@
...
+
+#include "src/backends/tensorflow/tf_virtual_device.h"
+
+#include <unordered_map>
+#include <vector>
+
+namespace nvidia { namespace inferenceserver {
+
+Status
+VirtualDeviceTracker::Create(
```
deadeyegoodwin: A common pattern would be to have a private GetInstance() method that has the static tracker and just returns a pointer to it. Then GetNextDeviceId would just call that to get the singleton instance.

src/backends/tensorflow/tf_virtual_device.cc
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Only 2019

src/backends/tensorflow/tf_virtual_device.h
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Only 2019

src/backends/tensorflow/base_backend.cc
```diff
@@ -164,14 +175,11 @@ BaseBackend::CreateExecutionContext(
   contexts_.emplace_back(new Context(instance_name, gpu_device, mbs));
   const std::unique_ptr<Context>& context = contexts_.back();
 
-  auto graphdef_backend_config =
```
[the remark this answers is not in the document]
xprotobeast2: I need this pointer on line 155 to access graphdef_backend_config->memory_limit_mb.

src/backends/tensorflow/base_backend.cc
```diff
@@ -143,8 +147,15 @@ BaseBackend::CreateExecutionContext(
                             ? Config().default_model_filename()
                             : cc_itr->second;
 
+    // Get virtual device tracker instance, and get next device id
+    VirtualDeviceTracker* device_tracker;
+    RETURN_IF_ERROR(VirtualDeviceTracker::Create(
+        graphdef_backend_config->memory_limit_mb, &device_tracker));
+
+    RETURN_IF_ERROR(device_tracker->GetNextDeviceId(gpu_device, &vgpu_device));
```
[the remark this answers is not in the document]
xprotobeast2: Will put this inside a check for whether the memory limits vector is empty. I think an empty vector should be the default, indicating that vgpus are not enabled.

src/backends/tensorflow/base_backend.cc
```diff
@@ -120,6 +120,10 @@ BaseBackend::CreateExecutionContext(
   // For a GPU context, determine the model file to use for device
   // compute capability. CPU always uses the default model file.
   std::string cc_model_filename;
+  int vgpu_device = gpu_device;
```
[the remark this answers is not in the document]
xprotobeast2: vgpu_device gets passed into the CreateExecutionContext() function. It needs to be the same as gpu_device in the case that vGPUs are not enabled, or its a CPU context. the alternative would be to have an extra if (gpu_device==Context::NO_GPU_DEVICE) check and then set the vgpu_device in the else clause or something.

src/backends/tensorflow/graphdef_backend_factory.cc
```diff
@@ -69,6 +70,8 @@ GraphDefBackendFactory::CreateBackend(
         std::make_tuple(graphdef_path));
   }
 
+  VirtualDeviceTracker::Init(backend_config_->memory_limit_mb);
```
deadeyegoodwin: This should be in GraphDefBackendFactory::Create

src/backends/tensorflow/savedmodel_backend_factory.cc
```diff
@@ -69,6 +70,8 @@ SavedModelBackendFactory::CreateBackend(
         std::make_tuple(savedmodel_path));
   }
 
+  VirtualDeviceTracker::Init(backend_config_->memory_limit_mb);
```
deadeyegoodwin: This should be in SavedModelBackendFactory::Create

src/backends/tensorflow/tf_virtual_device.cc
```diff
@@ -0,0 +1,87 @@
...
+}
+
+Status
+VirtualDeviceTracker::GetNextVirtualDevice(
+    const int gpu_device, int* vgpu_device)
+{
+  // Get the tracker
+  auto device_tracker = GetInstance();
+  // Check physical device index
+  if (gpu_device < 0 || static_cast<unsigned int>(gpu_device) >=
```
deadeyegoodwin: Put parens around sub-expressions in if

src/backends/tensorflow/tf_virtual_device.cc
```diff
@@ -0,0 +1,87 @@
...
+                            device_tracker->num_virtual_per_physical_.size()) {
+    return Status(
+        RequestStatusCode::INTERNAL, "Encountered invalid physical device ID " +
```
deadeyegoodwin: remove "Encountered"

src/backends/tensorflow/tf_virtual_device.cc
```diff
@@ -0,0 +1,87 @@
...
+  }
+  return Status::Success;
+}
+
+Status
+VirtualDeviceTracker::GetNextVirtualDevice(
+    const int gpu_device, int* vgpu_device)
+{
+  // Get the tracker
+  auto device_tracker = GetInstance();
```
deadeyegoodwin: Check for nullptr and return internal error

src/backends/tensorflow/tf_virtual_device.cc
```diff
@@ -0,0 +1,87 @@
...
+#include <unordered_map>
+#include <vector>
+
+namespace nvidia { namespace inferenceserver {
+
+Status
+VirtualDeviceTracker::Init(
+    const std::vector<std::vector<float>>& memory_limit_mb)
+{
+  // Instantiate tracker object
```
deadeyegoodwin: Don't initialize again if already initialized.

src/backends/tensorflow/tf_virtual_device.h
```diff
@@ -0,0 +1,54 @@
...
+#pragma once
+
+#include "src/core/constants.h"
+#include "src/core/status.h"
+
+namespace nvidia { namespace inferenceserver {
+
+class VirtualDeviceTracker {
+ public:
+  static Status Init(const std::vector<std::vector<float>>& memory_limit_mb);
```
deadeyegoodwin: Document these methods.

src/backends/tensorflow/base_backend.cc
```diff
@@ -143,8 +148,14 @@ BaseBackend::CreateExecutionContext(
                             ? Config().default_model_filename()
                             : cc_itr->second;
 
+    // Get virtual device tracker instance, and get next device id
+    if (!graphdef_backend_config->memory_limit_mb.empty()) {
```
deadeyegoodwin: Create VirtualDeviceTracker::HasVirtualDevice() and use that to determine if vgpus should be used

src/backends/tensorflow/autofill.cc
xprotobeast2: Changed to map from int to vector of floats

src/servers/main.cc
```diff
@@ -422,6 +429,28 @@ ParseFloatOption(const std::string arg)
   return std::stof(arg);
 }
 
+std::pair<int, std::vector<float>>
+ParseVGPUOption(const std::string arg)
```
CoderHam: Do you check if the user does not follow the mentioned format: **"\<physical GPU\>;\<number of virtual GPUs\>;\<memory limit for physical GPU\>"**
You should check and report an appropriate error in this case. For example arg.find(";) may be -1 if there is no ";"

src/backends/tensorflow/tensorflow_backend_tf.h
```diff
@@ -27,6 +27,8 @@
 
 #include <stddef.h>
 #include <stdint.h>
+#include <map>
```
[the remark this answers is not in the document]
xprotobeast2: I included them because of line 195. Will it be sufficient just to put them in the .cc file?

src/core/server.h
```diff
@@ -134,6 +135,17 @@ class InferenceServer {
   float TensorFlowGPUMemoryFraction() const { return tf_gpu_memory_fraction_; }
   void SetTensorFlowGPUMemoryFraction(float f) { tf_gpu_memory_fraction_ = f; }
 
+  // Get / set Tensorflow vGPU memory limits
+  std::map<int, std::vector<float>> TensorFlowVGPUMemoryLimits() const
+  {
+    return tf_vgpu_memory_limits_;
+  }
+  void SetTensorFlowVGPUMemoryLimits(
```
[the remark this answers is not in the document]
xprotobeast2: I just tried to follow the format of the functions defined from line 125-132. Should I deviate from the pattern?

src/backends/tensorflow/tensorflow_backend_tf.h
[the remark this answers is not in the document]
dzier: Yes, you are correct. I didn't see the map or vector earlier

src/core/server.h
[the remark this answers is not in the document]
dzier: Personally, I think they all should have inline, but it may be better to check with @deadeyegoodwin. Should still add the extra line.

src/servers/main.cc
dzier: To add to this, how does this handle the case where a user defines virtual GPUs on multiple physical GPUs?

src/backends/tensorflow/tensorflow_backend_tf.h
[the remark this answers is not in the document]
dzier: I must of have missed those lines earlier. This is correct.

src/servers/main.cc
xprotobeast2: Hemant asked me about this as well. You just use the option multiple times. I'm currently adding this to the description of the option. Also changing the option name to "--tf-add-vgpu".

src/backends/tensorflow/savedmodel_backend_factory.cc
```diff
@@ -49,6 +50,12 @@ SavedModelBackendFactory::Create(
   auto savedmodel_backend_config =
       std::static_pointer_cast<GraphDefBackendFactory::Config>(backend_config);
   factory->reset(new SavedModelBackendFactory(savedmodel_backend_config));
+
+  // Initialize VGPUs if required
+  if (!savedmodel_backend_config->memory_limit_mb.empty()) {
```
deadeyegoodwin: You are putting vgpu knowledge outside of VirtualDeviceTracker. You should always call Init and let the tracker determine how to initialize and if any gpu -> vgpu is needed or not.

src/servers/main.cc
```diff
@@ -218,7 +219,13 @@ std::vector<Option> options_{
      "Reserve a portion of GPU memory for TensorFlow models. Default "
      "value 0.0 indicates that TensorFlow should dynamically allocate "
      "memory as needed. Value of 1.0 indicates that TensorFlow should "
-     "allocate all of GPU memory."}};
+     "allocate all of GPU memory."},
+    {OPTION_TF_ADD_VGPU, "tf-add-vgpu",
+     "Add a tensorflow virtual GPU instance on a physical GPU. This option can"
```
deadeyegoodwin: "Add TensorFlow virtual GPU instances on a physical GPU..."

src/servers/main.cc
```diff
@@ -218,7 +219,13 @@ std::vector<Option> options_{
...
+     "be used multiple times.Input should be 2 integers and 1 float separated"
```
deadeyegoodwin: Space after '.'

src/servers/main.cc
```diff
@@ -422,6 +429,54 @@ ParseFloatOption(const std::string arg)
...
+{
+  std::vector<float> mem_limits_per_vgpu(0);
+
+  int delim_gpu = arg.find(";");
+  int delim_num_vgpus = arg.find(";", delim_gpu + 1);
+
+  // Check for 2 semicolons
+  if ((delim_gpu < 0) || (delim_num_vgpus < 0)) {
+    *malformed = true;
+    LOG_ERROR << "Cannot add virtual devices due to malformed TF-VGPU setting "
```
deadeyegoodwin: Be specific and don't use terms (TF_VGPU) that don't necessarily mean anything to the user.
"--tf-add-vgpu argument requires format blah;blah;blah"

src/servers/main.cc
```diff
@@ -422,6 +429,54 @@ ParseFloatOption(const std::string arg)
...
+  std::string gpu_string = arg.substr(0, delim_gpu);
+  std::string vgpu_string =
+      arg.substr(delim_gpu + 1, delim_num_vgpus - delim_gpu - 1);
+  std::string mem_limit_string = arg.substr(delim_num_vgpus + 1);
+
+  // Ensure that options are non-empty otherwise calling stoi/stof will throw an
+  // exception
+  if (gpu_string.empty() || vgpu_string.empty() || mem_limit_string.empty()) {
+    *malformed = true;
+    LOG_ERROR << "Cannot add virtual devices due to malformed TF-VGPU setting "
```
deadeyegoodwin: Be specific

src/servers/main.cc
```diff
@@ -422,6 +429,54 @@ ParseFloatOption(const std::string arg)
...
+              << arg;
+    return std::make_pair(-1, mem_limits_per_vgpu);
+  }
+
+  int gpu_device = std::stoi(gpu_string);
+  int num_vgpus_on_device = std::stoi(vgpu_string);
+  float mem_limit = std::stof(mem_limit_string);
+
+  if (num_vgpus_on_device <= 0) {
+    LOG_ERROR << "Cannot add virtual devices due to invalid number of VGPUs "
```
deadeyegoodwin: Specific... ".. number of virtual GPUs must be > 0"

src/servers/main.cc
```diff
@@ -422,6 +429,54 @@ ParseFloatOption(const std::string arg)
...
+              << arg;
+    *malformed = true;
+    return std::make_pair(-1, mem_limits_per_vgpu);
+  }
+
```
deadeyegoodwin: Should also check that gpu_device and mem_limit are positive

src/servers/main.cc
```diff
@@ -422,6 +429,54 @@ ParseFloatOption(const std::string arg)
...
+  // Compute map element
+  float limit_per_vgpu = mem_limit / num_vgpus_on_device;
```
deadeyegoodwin: Memory limit that the user provides should be the memory for each virtual GPU

src/servers/main.cc
```diff
@@ -553,6 +610,14 @@ Parse(nvidia::inferenceserver::InferenceServer* server, int argc, char** argv)
       case OPTION_TF_GPU_MEMORY_FRACTION:
         tf_gpu_memory_fraction = ParseFloatOption(optarg);
         break;
+      case OPTION_TF_ADD_VGPU:
+        bool malformed;
+        std::pair<int, std::vector<float>> map_element =
+            ParseVGPUOption(optarg, &malformed);
+        if (!malformed) {
+          tf_vgpu[map_element.first] = map_element.second;
```
deadeyegoodwin: You can't just continue... if something went wrong with the parsing then you should return false.

src/backends/tensorflow/graphdef_backend_factory.cc
```diff
@@ -49,6 +50,12 @@ GraphDefBackendFactory::Create(
   auto graphdef_backend_config =
       std::static_pointer_cast<Config>(backend_config);
   factory->reset(new GraphDefBackendFactory(graphdef_backend_config));
+
+  // Initalize VGPUs if required
+  if (!graphdef_backend_config->memory_limit_mb.empty()) {
```
deadeyegoodwin: Always call Init (see related comment below)

src/core/server.h
```diff
@@ -129,6 +130,18 @@ class InferenceServer {
   float TensorFlowGPUMemoryFraction() const { return tf_gpu_memory_fraction_; }
   void SetTensorFlowGPUMemoryFraction(float f) { tf_gpu_memory_fraction_ = f; }
 
+  // Get / set Tensorflow vGPU memory limits
+  std::map<int, std::vector<float>> TensorFlowVGPUMemoryLimits() const
```
deadeyegoodwin: const &

src/core/trtserver.cc
```diff
@@ -222,6 +222,17 @@ class TrtServerOptions {
   float TensorFlowGpuMemoryFraction() const { return tf_gpu_mem_fraction_; }
   void SetTensorFlowGpuMemoryFraction(float f) { tf_gpu_mem_fraction_ = f; }
 
+  std::map<int, std::vector<float>> TensorFlowVgpuMemoryLimits() const
```
deadeyegoodwin: const&

src/core/trtserver.h
```diff
@@ -281,6 +281,13 @@ TRTSERVER_EXPORT TRTSERVER_Error*
 TRTSERVER_ServerOptionsSetTensorFlowGpuMemoryFraction(
     TRTSERVER_ServerOptions* options, float fraction);
 
+// Add Tensorflow virtual GPU instances to physical GPU. Specify limit
+// for total memory available for use on that physical GPU.
+TRTSERVER_EXPORT TRTSERVER_Error*
+TRTSERVER_ServerOptionsAddTensorFlowVgpuMemoryLimits(
+    TRTSERVER_ServerOptions* options, int gpu_device, int num_vgpus,
+    float mem_limit);
```
deadeyegoodwin: Memory should be per-vgpu memory, so how about name it "per_vgpu_memory_mbytes". And it should be a uint64_t, not a float.

src/servers/main.cc
```diff
@@ -223,7 +224,16 @@ std::vector<Option> options_{
      "Reserve a portion of GPU memory for TensorFlow models. Default "
      "value 0.0 indicates that TensorFlow should dynamically allocate "
      "memory as needed. Value of 1.0 indicates that TensorFlow should "
-     "allocate all of GPU memory."}};
+     "allocate all of GPU memory."},
+    {OPTION_TF_ADD_VGPU, "tf-add-vgpu",
+     "Add a tensorflow virtual GPU instances on a physical GPU. Input "
+     "should be 2 integers and 1 float separated by semicolons in the format "
+     "<physical GPU>;<number of virtual GPUs>;<memory limit per VGPU>. "
```
deadeyegoodwin: Memory limit should be specified in bytes? Mbytes? I think Mbytes is best but you should document.

src/servers/main.cc
```diff
@@ -502,6 +512,63 @@ ParseFloatOption(const std::string arg)
   return std::stof(arg);
 }
 
+struct VgpuOption {
+  int gpu_device_;
+  int num_vgpus_;
+  float mem_limit_;
```
deadeyegoodwin: uint64_t... make name explicit "mem_limit_bytes_" or "mem_limit_mbytes_"

src/servers/main.cc
```diff
@@ -502,6 +512,63 @@ ParseFloatOption(const std::string arg)
...
+  int delim_gpu = arg.find(";");
+  int delim_num_vgpus = arg.find(";", delim_gpu + 1);
+
+  // Check for 2 semicolons
+  if ((delim_gpu < 0) || (delim_num_vgpus < 0)) {
+    LOG_ERROR << "Cannot add virtual devices due to incorrect number of inputs."
+                 "--tf-add-vgpu argument requires format <physical "
+                 "GPU>;<number of virtual GPUs>;<memory limit per VGPU>. "
+              << "Found: " << arg;
+    return {-1, 0, 0.0};
```
deadeyegoodwin: Follow pattern of ParseBoolOption, just print Usage() and exit.

src/servers/main.cc
```diff
@@ -648,6 +716,19 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
       case OPTION_TF_GPU_MEMORY_FRACTION:
         tf_gpu_memory_fraction = ParseFloatOption(optarg);
         break;
+      case OPTION_TF_ADD_VGPU:
+        tf_vgpu = ParseVGPUOption(optarg);
+        if ((tf_vgpu.gpu_device_ >= 0) && (tf_vgpu.num_vgpus_ > 0) &&
```
deadeyegoodwin: If parsing returns is was successful so don't need to check or return false here.

src/core/server.h
deadeyegoodwin: Strange to have a vector, why not a std::pair<int, uint64_t>?

xprotobeast2: I have to turn it into a vector somewhere. I suppose I could do it when constructing the backend config?

src/servers/main.cc
```diff
@@ -505,6 +516,69 @@ ParseFloatOption(const std::string arg)
   return std::stof(arg);
 }
 
+struct VgpuOption {
+  int gpu_device_;
+  int num_vgpus_;
+  uint64_t mem_limit_mbytes_;
```
CoderHam: While `mem_limit_mbytes_` is `uint64_t`, you later use a `float` for the memory limit in mb: `std::map<int, std::vector<float>>& memory_limit_mb`. I might be missing something here but is there a reason for using different datatypes?

xprotobeast2: The tensorflow API requires a float, but we're restricting the use to uint64_t.

CoderHam: Understood

PR #440, opened 2019-07-09
Shared Memory Client API + Image client example

CoderHam:
- [ ] Image Client example using Shared memory for Input and Output - (in a different PR now)
- [x] Update Client API logic for shared memory I/O - (Input side only)
- [x] Update Client API Docs

**Note**: For using shared memory outside docker `--ipc=host` or `--ipc=container:<id>`


src/clients/c++/CMakeLists.txt
```diff
@@ -109,6 +109,23 @@ install(
   RUNTIME DESTINATION bin
 )
 
+#
+# image_shm_client
+#
+find_package(OpenCV REQUIRED)
+add_executable(image_shm_client image_shm_client.cc)
```
deadeyegoodwin: Instead of copying image_client should just add an option to the existing image_client to use shared memory for input and outputs.

src/clients/c++/request.h
```diff
@@ -329,6 +329,13 @@ class InferContext {
     /// \param input The vector holding tensor string values.
     /// \return Error object indicating success or failure.
     virtual Error SetFromString(const std::vector<std::string>& input) = 0;
+
+    // Set the shard memory object for the batch tensor input
```
deadeyegoodwin: Follow existing documentation "Set tensor values for this input by reference into a shared memory region. The vector is not copied and so it must not be modified or destroyed until this input is no longer needed (that is until the Run() call(s) that use the input have completed). For batched inputs this function must be called batch-size times to provide all tensor values for a batch of this input.

src/clients/c++/request.h
```diff
@@ -329,6 +329,13 @@ class InferContext {
...
+    /// \param shm_key The shared memory location holding the input tensors.
```
deadeyegoodwin: If "shared memory location" the terminology used in the API documentation for shm_open and mmap? We should follow whatever naming convention is used there. Also it is not the location "holding the input tensors". It is "holding tensor values for this input".

deadeyegoodwin: "Offset into the shared memory <whatever> to the start of the input tensor values"

src/clients/c++/request.h
```diff
@@ -329,6 +329,13 @@ class InferContext {
...
+    /// \param offset The offset for the batch input tensor.
+    /// \param byte_size The byte size of the batch input tensor.
```
deadeyegoodwin: "The size, in bytes of the input tensor data. Must match the size expected by the input".

src/clients/c++/request.h
```diff
@@ -329,6 +329,13 @@ class InferContext {
...
+    /// \return Error object indicating success or failure.
+    virtual Error SetSharedMemoryObject(const std::string &shm_key, size_t offset, size_t byte_size) = 0;
```
deadeyegoodwin: Just "SetSharedMemory"

src/clients/c++/request.h
```diff
@@ -347,6 +354,13 @@ class InferContext {
     /// model configuration. Variable-size dimensions are reported as
     /// -1.
     virtual const DimsList& Dims() const = 0;
+
+    // Set the shard memory object for the batch tensor output
```
deadeyegoodwin: Same comments as for input. Till now Output has only returned information about the model configuration so using it to set something may take some work in the implementation.

src/clients/c++/request_common.cc
```diff
@@ -214,6 +214,20 @@ InputImpl::SetRaw(const std::vector<uint8_t>& input)
...
 }
 
+Error
+InputImpl::SetSharedMemoryObject(const std::string &shm_key, size_t offset, size_t byte_size)
+{
+  // TODO Add Logic
+  return Error::Success;
+}
+
+Error
```
deadeyegoodwin: Put this with other OutputImpl functions...

src/clients/c++/CMakeLists.txt
CoderHam: Done

src/clients/c++/image_client.cc
```diff
@@ -173,44 +216,57 @@ void
 Postprocess(
     const std::map<std::string, std::unique_ptr<nic::InferContext::Result>>&
         results,
-    const std::vector<std::string>& filenames, const size_t batch_size)
+    const std::vector<std::string>& filenames, const size_t batch_size,
+    const float* shm_addr, size_t offset, const size_t& byte_size, bool use_shm)
 {
-  if (results.size() != 1) {
+  if ((results.size() != 1) && (!use_shm)) {
```
deadeyegoodwin: It is an option to read directly from the shared memory where the outputs were written, but the GetRawAtCursor, etc. should also work just as well. So I would just have a comment saying something like "Can read outputs directly from shared memory but can also use the client APIs", and then use the client APIs.

src/clients/c++/image_client.cc
```diff
@@ -388,6 +445,7 @@ ParseModel(
 
     if (dim > 1) {
       non_one_cnt++;
+      *op_size = dim;
```
deadeyegoodwin: Not sure what op_size is. Looks like the size of the output tensor. If so needs a better name, like output_size

src/clients/c++/image_client.cc
```diff
@@ -468,7 +526,8 @@ void
 FileToInputData(
```
deadeyegoodwin: This function should still just return the input_data. Make another function to get it into shared memory.

src/clients/c++/image_client.cc
```diff
@@ -490,6 +549,16 @@ FileToInputData(
 
   // Pre-process the image to match input size expected by the model.
   Preprocess(img, format, type1, type3, c, cv::Size(w, h), scale, input_data);
+
+  if (use_shm) {
+    *byte_size = (size_t)(img.total() * img.elemSize());
+
+    // place image data into memory in the form [channel][width][height]
```
deadeyegoodwin: This comment isn't generally true since image_client supports both CHW and HWC.

src/clients/c++/image_client.cc
```diff
@@ -640,22 +717,43 @@ main(int argc, char** argv)
...
+    if (!use_shm) {
+      if ((image_data.size() == 1) && !preprocess_output_filename.empty()) {
+        std::ofstream output_file(preprocess_output_filename);
+        std::ostream_iterator<uint8_t> output_iterator(output_file);
+        std::copy(image_data[0].begin(), image_data[0].end(), output_iterator);
+      }
+    } else {
+      offset += byte_size;
     }
   }
```
deadeyegoodwin: Defer shared_memory initialize till here and you don't need many of the above changes. At this point you know the batch size and size of an image (from image_data), so you can allocate the shared memory and copy the images into it.

src/clients/c++/image_client.cc
```diff
@@ -604,16 +677,20 @@ main(int argc, char** argv)
...
+  ParseModel(
+      ctx, batch_size, &c, &h, &w, &format, &type1, &type3, &op_size, verbose);
 
-  //
-
-  // Collect the names of the image(s).
+  // Collect the names of the image(s) and their sizes (for shared memory).
   std::vector<std::string> image_filenames;
 
+  size_t input_byte_size = c * h * w * sizeof(float);
```
deadeyegoodwin: Can't assume float... see below as to why these changes are not needed.

src/clients/c++/request.h
```diff
@@ -329,6 +329,23 @@ class InferContext {
     /// \param input The vector holding tensor string values.
     /// \return Error object indicating success or failure.
     virtual Error SetFromString(const std::vector<std::string>& input) = 0;
+
+    /// Set tensor values for this input by reference into a shared memory
+    /// region. The values are not copied and so it must not be modified or
```
deadeyegoodwin: ... and so the shared memory region must not ...

src/clients/c++/request.h
```diff
@@ -329,6 +329,23 @@ class InferContext {
...
+    /// destroyed until this input is no longer needed (that is until the Run()
+    //// call(s) that use the input have completed). For batched inputs this
+    /// function must be called batch-size times to provide all tensor values
+    /// for a batch of this input.
+    /// \param shm_key The identifier for the shared memory object that is used
```
deadeyegoodwin: With Register function this changes to the user-supplied name for the register shared memory region.

src/clients/c++/request.h
```diff
@@ -347,6 +364,23 @@ class InferContext {
...
+    /// for a batch of this output.
+    /// \param shm_key The identifier for the shared memory object that is used
+    /// to get the region in shared memory where the tensor values for this
+    /// output should be stored.
+    /// \param offset The offset into the shared memory to the start of the
+    /// output tensor values.
+    /// \param byte_size The size, in bytes of the output tensor data. Must
+    /// match the size expected by the output.
+    /// \return Error object indicating success or failure.
+    virtual Error SetSharedMemory(
```
deadeyegoodwin: Delete? We don't set output values...

src/clients/c++/image_client.cc
CoderHam: While using shared memory for output, why is it necessary to populate that output as a part of the results in the response?

src/clients/c++/image_client.cc
CoderHam: Removed

src/clients/c++/image_client.cc
CoderHam: Shared memory needs to be allocated before this since the image data needs to be copied into the right place here and it requires the pointer from mmap

src/clients/c++/request.h
CoderHam: Yes

src/clients/c++/request.h
CoderHam: This just sets the shared memory region to be used for the output tensors. it does not set their values

src/clients/c++/image_client.cc
deadeyegoodwin: GetRawAtCursor should just maintain a pointer into the shared memory region and use the data directly from there. We want the client APIs to work for both "raw" and shared-memory outputs.

src/clients/c++/image_client.cc
deadeyegoodwin: Can't we copy it below just before calling SetSharedMemory?

src/clients/c++/request.h
```diff
@@ -298,8 +298,8 @@ class InferContext {
     virtual Error SetShape(const std::vector<int64_t>& dims) = 0;
 
     /// Set tensor values for this input from a byte array. The array
-    /// is not copied and so it must not be modified or destroyed
-    /// until this input is no longer needed (that is until the Run()
+    /// is not copied and so  shared memory region must not be modified or
```
deadeyegoodwin: Why does this mention shared memory? I don't think this comment needs to change.

src/clients/c++/request.h
```diff
@@ -310,8 +310,8 @@ class InferContext {
     virtual Error SetRaw(const uint8_t* input, size_t input_byte_size) = 0;
 
     /// Set tensor values for this input from a byte vector. The vector
-    /// is not copied and so it must not be modified or destroyed
-    /// until this input is no longer needed (that is until the Run()
+    /// is not copied and so  shared memory region must not be modified or
```
deadeyegoodwin: I don't think this comment needs to change

src/clients/c++/request.h
```diff
@@ -329,6 +329,22 @@ class InferContext {
     /// \param input The vector holding tensor string values.
     /// \return Error object indicating success or failure.
     virtual Error SetFromString(const std::vector<std::string>& input) = 0;
+
+    /// Set tensor values for this input by reference into a shared memory
+    /// region. The values are not copied and so it must not be modified or
```
deadeyegoodwin: "... and so the shared memory region and its contents must not be..."

src/clients/c++/request.h
deadeyegoodwin: Right. The comment needs to be changed to indicate that this is just providing shared memory buffer where output is to be written... as it is now it mentions output values.

src/clients/python/__init__.py
```diff
@@ -161,6 +161,13 @@ def from_param(cls, value):
 _crequest_infer_ctx_input_set_raw = _crequest.InferContextInputSetRaw
 _crequest_infer_ctx_input_set_raw.restype = c_void_p
 _crequest_infer_ctx_input_set_raw.argtypes = [c_void_p, c_void_p, c_uint64]
+_crequest_infer_ctx_input_set_shared_memory = _crequest.InferContextInputSetSharedMemory
```
deadeyegoodwin: We're not doing any python changes yet. Can you move this and crequest changes out of this commit.

src/core/api.proto
```diff
@@ -94,6 +94,13 @@ message InferRequestHeader
     //@@       for tensors with a non-fixed-size datatype (like STRING).
     //@@
     uint64 batch_byte_size = 3;
+
+    //@@  .. cpp:var:: string shm_name
```
deadeyegoodwin: Remove, this file is in the other commit, isn't it?

src/core/grpc_service.proto
```diff
@@ -65,6 +65,12 @@ service GRPCService
   //@@
   rpc Control(ControlRequest) returns (ControlResponse) {}
 
+  //@@  .. cpp:var:: rpc ShmControl(ShmControlRequest) returns (ShmControlResponse)
```
deadeyegoodwin: Remove, this file is in the other commit, isn't it?

src/core/api.proto
CoderHam: It is needed to build and test successfully. I thought it would be ignored after I do a rebase.

src/core/grpc_service.proto
CoderHam: Same

src/clients/python/__init__.py
CoderHam: Sure. Will do

src/clients/c++/image_client.cc
CoderHam: Yes we can

src/core/api.proto
CoderHam: Rebased and overwritten.

src/core/grpc_service.proto
CoderHam: Rebased and overwritten.

src/clients/c++/request.h
```diff
@@ -329,6 +329,22 @@ class InferContext {
     /// \param input The vector holding tensor string values.
     /// \return Error object indicating success or failure.
     virtual Error SetFromString(const std::vector<std::string>& input) = 0;
+
+    /// Set tensor values for this input by reference into a shared memory
+    /// region. The values are not copied and so the shared memory region and
+    /// its contents must not be modified or destroyed until this input is no
+    /// longer needed (that is until the Run() call(s) that use the input have
```
deadeyegoodwin: Missing end of sentence "... have completed."

src/clients/c++/request.h
```diff
@@ -329,6 +329,22 @@ class InferContext {
...
+    /// For batched inputs, all tensor values must be contiguous in a single
```
deadeyegoodwin: "This function must be called a single time for an input that is using shared memory. For batched inputs, the tensor values for the entire batch must be contiguous..."

src/clients/c++/request_common.cc
```diff
@@ -204,6 +205,7 @@ InputImpl::SetRaw(const uint8_t* input, size_t input_byte_size)
 
   bufs_.push_back(input);
   buf_byte_sizes_.push_back(input_byte_size);
+  io_type_ = RAW;
```
deadeyegoodwin: You need to check at the start of the function if io_type_ is SHARED_MEMORY and if it is return an error.

src/clients/c++/request_common.cc
```diff
@@ -214,6 +216,17 @@ InputImpl::SetRaw(const std::vector<uint8_t>& input)
   return SetRaw(&input[0], input.size());
 }
 
+Error
+InputImpl::SetSharedMemory(
+    const std::string& name, size_t offset, size_t byte_size)
+{
+  shm_name_ = name;
```
deadeyegoodwin: Check if io_type_ is RAW and return error.
Check if io_type_ is SHARED_MEMORY and return error that can only set once.

src/clients/c++/request_common.cc
```diff
@@ -339,6 +352,26 @@ InputImpl::PrepareForRequest()
 
 //==============================================================================
 
+Error
+OutputImpl::SetSharedMemory(
+    const std::string& name, size_t offset, size_t byte_size)
+{
+  shm_name_ = name;
```
deadeyegoodwin: Check if io_type_ is SHARED_MEMORY and return error that can only set once.

src/clients/c++/request_http.cc
```diff
@@ -982,6 +1159,7 @@ HttpRequestImpl::CreateResult(
   }
 
   std::unique_ptr<ResultImpl> result(new ResultImpl(infer_output, batch_size));
+  // Check if output uses shared memory. If so handle differently
```
deadeyegoodwin: Remove this line

PR #442, opened 2019-07-10
opencv4 compatibility fix in image_client

tanmayv25:
With this change image_client can run off from openCV 4 library. The execution was verified on client container with openCV 4 installed.

### DIFF
--- a/src/clients/c++/image_client.cc
+++ b/src/clients/c++/image_client.cc
@@ -49,7 +49,7 @@
 #include <opencv2/highgui.hpp>
 #include <opencv2/imgproc.hpp>
 #define GET_TRANSFORMATION_CODE(x) CV_##x
-#elif CV_MAJOR_VERSION == 4
+#elif CV_MAJOR_VERSION >= 4
 #include <opencv2/core.hpp>
 #include <opencv2/highgui.hpp>
 #include <opencv2/imgproc.hpp>
@@ -88,7 +88,8 @@ Preprocess(
   } else if ((img.channels() == 1) && (img_channels == 3)) {
     cv::cvtColor(img, sample, GET_TRANSFORMATION_CODE(GRAY2RGB));
   } else {
-    std::cerr << "unexpected number of channels in input image or model"
+    std::cerr << "unexpected number of channels " << img.channels()
+              << " in input image, model expects " << img_channels << "."
               << std::endl;
     exit(1);
   }
--- a/src/clients/c++/image_client.cc
+++ b/src/clients/c++/image_client.cc
@@ -43,10 +43,17 @@
 #include <opencv2/core/core.hpp>
 #include <opencv2/highgui/highgui.hpp>
 #include <opencv2/imgproc/imgproc.hpp>
-#elif CV_MAJOR_VERSION >= 3
+#define GET_TRANSFORMATION_CODE(x) CV_##x
+#elif CV_MAJOR_VERSION == 3
 #include <opencv2/core.hpp>
 #include <opencv2/highgui.hpp>
 #include <opencv2/imgproc.hpp>
+#define GET_TRANSFORMATION_CODE(x) CV_##x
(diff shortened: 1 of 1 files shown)


src/clients/c++/image_client.cc
```diff
@@ -85,6 +86,27 @@ Preprocess(
...
+  if ((img.channels() == 3) && (img_channels == 1)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGR2GRAY);
+  } else if ((img.channels() == 4) && (img_channels == 1)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGRA2GRAY);
+  } else if ((img.channels() == 3) && (img_channels == 3)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGR2RGB);
+  } else if ((img.channels() == 4) && (img_channels == 3)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGRA2RGB);
+  } else if ((img.channels() == 1) && (img_channels == 3)) {
+    cv::cvtColor(img, sample, cv::COLOR_GRAY2RGB);
```
GuanLuo: perhaps you can use `#define` macro to unify the color conversion code

src/clients/c++/image_client.cc
```diff
@@ -85,6 +86,27 @@ Preprocess(
...
+  } else {
+    std::cerr << "unexpected number of channels in input image or model"
+              << std::endl;
+    exit(1);
+  }
+#else
+  std::cerr << "The installed openCV version is not supported yet" << std::endl;
```
CoderHam: If possible try to log out the version here "The installed openCV version <version> ..."

src/clients/c++/image_client.cc
```diff
@@ -85,6 +86,27 @@ Preprocess(
...
+  } else if ((img.channels() == 4) && (img_channels == 1)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGRA2GRAY);
+  } else if ((img.channels() == 3) && (img_channels == 3)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGR2RGB);
+  } else if ((img.channels() == 4) && (img_channels == 3)) {
+    cv::cvtColor(img, sample, cv::COLOR_BGRA2RGB);
+  } else if ((img.channels() == 1) && (img_channels == 3)) {
+    cv::cvtColor(img, sample, cv::COLOR_GRAY2RGB);
+  } else {
+    std::cerr << "unexpected number of channels in input image or model"
```
CoderHam: Again you can do the same for the error message for image channels. 
"unexpected number of channels \<x\> in input image, model expects \<y\>".

src/clients/c++/image_client.cc
```diff
@@ -43,10 +43,17 @@
...
 #elif CV_MAJOR_VERSION == 3
 #include <opencv2/core.hpp>
 #include <opencv2/highgui.hpp>
 #include <opencv2/imgproc.hpp>
+#define GET_TRANSFORMATION_CODE(x) CV_##x
+#elif CV_MAJOR_VERSION == 4
+#include <opencv2/core.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
+#define GET_TRANSFORMATION_CODE(x) cv::COLOR_##x
```
GuanLuo: Didn't notice this in the first review. Since the include files for both OpenCV 3 and 4 are in the same names, you can even do:
```c++
#if CV_MAJOR_VERSION >= 3
#include ...
#else 
#include ...
#endif 

#if CV_MAJOR_VERSION == 4
#define GET...
#else 
#define GET...
#endif
```

src/clients/c++/image_client.cc
tanmayv25: OpenCV may offer backwards compatibility. Hence, I am not putting a validation check as of now.

PR #445, opened 2019-07-10
Fix Autofill errors 

CoderHam:
Follow up to earlier autofill fix - model config test failure resolved

### DIFF
--- a/src/backends/tensorflow/autofill.cc
+++ b/src/backends/tensorflow/autofill.cc
@@ -330,6 +330,11 @@ AutoFillGraphDef::Create(
   // attempt to determine the platform. For now we allow multiple versions
   // and only inspect the first verison directory to ensure it is valid.
   // We can add more aggressive checks later.
+  if (version_dirs.size() == 0) {
+      return Status(
+          RequestStatusCode::INTERNAL, "unable to autofill for '" + model_name +
+                                           "' due to no version directories");
+  }
 
   const auto version_path = JoinPath({model_path, *(version_dirs.begin())});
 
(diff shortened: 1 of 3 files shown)


qa/L0_model_config/autofill_noplatform/ensemble/tensor_to_input_overmapped/expected_2
```diff
@@ -0,0 +1 @@
+ensemble tensor 'temp_tensor_5' is unused in ensemble 'tensor_to_input_overmapped'
```
GuanLuo: Why touching ensemble test? Was this removed accidentally in earlier changes?

CoderHam: Very likely. This was not failing due to my changes. The error message expected itself was incorrect - i.e. the one in qa/L0_model_config/autofill_noplatform/ensemble/tensor_to_input_overmapped/expected

GuanLuo: There is no reason to keep expected as the error message is changed, you should just modify expected and remove the empty expected_2.

CoderHam: Sure

src/backends/tensorflow/autofill.cc
```diff
@@ -330,6 +330,11 @@ AutoFillGraphDef::Create(
   // attempt to determine the platform. For now we allow multiple versions
   // and only inspect the first verison directory to ensure it is valid.
   // We can add more aggressive checks later.
+  if (version_dirs.size() == 0) {
+      return Status(
+          RequestStatusCode::INTERNAL, "unable to autofill for '" + model_name +
```
dzier: I think you need to run the formatter on this file

CoderHam: Done

PR #451, opened 2019-07-13
Ctrl+C prints complete concurrency

tanmayv25:
With these changes the user can get the summary of the run even if they interrupts the execution. If printing/writing of the information is time consuming or the user wants to immediately stop the execution, they can again interrupt the execution, in which case the perf_client will exit without completing the summary dump.

### DIFF
--- a/src/clients/c++/perf_client.cc
+++ b/src/clients/c++/perf_client.cc
@@ -141,15 +141,24 @@ volatile bool finished_inferencing = false;
 void
 SignalHandler(int signum)
 {
-  std::cout << "Interrupt signal (" << signum << ") received." << std::endl
-            << "Waiting for in-flight inferences to complete." << std::endl;
   // Upon invoking the SignalHandler for the first time early_exit flag is
   // invoked. On the subsequent invocations, if there are no in-flight inference
   // requests, the program exits.
-  if (!early_exit)
+  if (!early_exit) {
+    std::cout << "Interrupt signal (" << signum << ") received." << std::endl
+              << "Initializing early exit..." << std::endl;
     early_exit = true;
-  else if (finished_inferencing)
+  } else if (finished_inferencing) {
+    std::cout << "Another Interrupt signal (" << signum << ") received."
+              << std::endl
+              << "No in-flight inferences. Exitting immediately..."
+              << std::endl;
     exit(0);
+  } else {
+    std::cout << "Interrupt signal (" << signum << ") received." << std::endl
+              << "Still waiting for in-flight inferences to complete."
+              << std::endl;
+  }
 }
 
 typedef struct PerformanceStatusStruct {
@@ -590,7 +599,7 @@ ConcurrencyManager::ChangeConcurrencyLevel(
         new std::vector<nic::InferContext::Stat>());
     threads_concurrency_.emplace_back(new size_t(0));
 
-    // Worker maintians concurrency in different ways.
+    // Worker maintains concurrency in different ways.
     // For sequence models, multiple contexts must be created for multiple
     // concurrent sequences.
     // For non-sequence models, one context can send out multiple requests
--- a/src/clients/c++/perf_client.cc
+++ b/src/clients/c++/perf_client.cc
@@ -135,14 +135,21 @@ using TimestampVector =
 namespace {
 
 volatile bool early_exit = false;
+// The flag is set when there are no in-flight inferences to complete.
+volatile bool finished_inferencing = false;
 
 void
 SignalHandler(int signum)
 {
   std::cout << "Interrupt signal (" << signum << ") received." << std::endl
             << "Waiting for in-flight inferences to complete." << std::endl;
-
-  early_exit = true;
+  // Upon invoking the SignalHandler for the first time early_exit flag is
+  // invoked. On the subsequent invocations, if there are no in-flight inference
+  // requests, the program exits.
+  if (!early_exit)
+    early_exit = true;
+  else if (finished_inferencing)
+    exit(0);
 }
 
 typedef struct PerformanceStatusStruct {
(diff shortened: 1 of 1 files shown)


src/clients/c++/perf_client.cc
```diff
@@ -135,14 +135,21 @@ using TimestampVector =
...
 SignalHandler(int signum)
 {
   std::cout << "Interrupt signal (" << signum << ") received." << std::endl
             << "Waiting for in-flight inferences to complete." << std::endl;
-
-  early_exit = true;
+  // Upon invoking the SignalHandler for the first time early_exit flag is
+  // invoked. On the subsequent invocations, if there are no in-flight inference
+  // requests, the program exits.
+  if (!early_exit)
```
GuanLuo: brackets around conditional

src/clients/c++/perf_client.cc
```diff
@@ -135,14 +135,21 @@ using TimestampVector =
...
+    early_exit = true;
```
deadeyegoodwin: Probably need to print something so that the user knows that perf_client got the ctrl-c and is now just finishing up outstanding requests...

tanmayv25: Done

src/clients/c++/perf_client.cc
tanmayv25: Done

src/clients/c++/perf_client.cc
```diff
@@ -1991,9 +2007,13 @@ main(int argc, char** argv)
       }
     }
   }
+  finished_inferencing = true;
   if (!err.IsOk()) {
     std::cerr << err << std::endl;
-    return 1;
+    // In the case of early_exit, the thread doesn't returns to continue
+    // reporting the summary.
+    if (!early_exit)
```
GuanLuo: bracket

tanmayv25: Done

src/clients/c++/perf_client.cc
```diff
@@ -135,14 +135,30 @@ using TimestampVector =
...
+  // invoked. On the subsequent invocations, if there are no in-flight inference
+  // requests, the program exits.
+  if (!early_exit) {
+    std::cout << "Interrupt signal (" << signum << ") received." << std::endl
+              << "Initializing early exit..." << std::endl;
+    early_exit = true;
+  } else if (finished_inferencing) {
+    std::cout << "Another Interrupt signal (" << signum << ") received."
+              << std::endl
+              << "No in-flight inferences. Exitting immediately..."
```
deadeyegoodwin: "Exiting"

src/clients/c++/perf_client.cc
```diff
@@ -135,14 +135,30 @@ using TimestampVector =
...
 {
-  std::cout << "Interrupt signal (" << signum << ") received." << std::endl
-            << "Waiting for in-flight inferences to complete." << std::endl;
-
-  early_exit = true;
+  // Upon invoking the SignalHandler for the first time early_exit flag is
+  // invoked. On the subsequent invocations, if there are no in-flight inference
+  // requests, the program exits.
+  if (!early_exit) {
+    std::cout << "Interrupt signal (" << signum << ") received." << std::endl
```
deadeyegoodwin: Can you just unconditionally print "Interrupt signal ..."  at the start of SignalHandler

src/clients/c++/perf_client.cc
```diff
@@ -135,14 +135,30 @@ using TimestampVector =
...
+              << "Initializing early exit..." << std::endl;
+    early_exit = true;
+  } else if (finished_inferencing) {
+    std::cout << "Another Interrupt signal (" << signum << ") received."
+              << std::endl
+              << "No in-flight inferences. Exitting immediately..."
+              << std::endl;
+    exit(0);
+  } else {
+    std::cout << "Interrupt signal (" << signum << ") received." << std::endl
```
deadeyegoodwin: I thought on the second Ctrl-C we should just exit immediately.

src/clients/c++/perf_client.cc
```diff
@@ -135,14 +135,30 @@ using TimestampVector =
...
-  std::cout << "Interrupt signal (" << signum << ") received." << std::endl
-            << "Waiting for in-flight inferences to complete." << std::endl;
-
-  early_exit = true;
+  // Upon invoking the SignalHandler for the first time early_exit flag is
+  // invoked. On the subsequent invocations, if there are no in-flight inference
+  // requests, the program exits.
+  if (!early_exit) {
+    std::cout << "Interrupt signal (" << signum << ") received." << std::endl
+              << "Initializing early exit..." << std::endl;
```
deadeyegoodwin: Would "Waiting for in-flight inferences to complete..." be more useful?

tanmayv25: The client was originally waiting for the inflight inferences before exiting so I decided to carry on the behavior. But it makes more sense to exit immediately when a second interrupt is received. Hence, now perf_client would not wait for all in-flight inferences if second request is raised.

deadeyegoodwin: On the first Ctrl-C perf_client does/should wait for all in-flight inferences to complete. Can we just keep the original "Waiting for in-flight inferences to complete."?

tanmayv25: Added the message.

src/clients/c++/perf_client.cc
```diff
@@ -1991,9 +1998,14 @@ main(int argc, char** argv)
       }
     }
   }
+
   if (!err.IsOk()) {
     std::cerr << err << std::endl;
-    return 1;
+    // In the case of early_exit, the thread doesn't returns to continue
```
CoderHam: Fix typo

tanmayv25: Done

PR #455, opened 2019-07-14
Disallow zero-prefixed version directories

jobegrabber:


src/core/model_repository_manager.cc
```diff
@@ -1102,6 +1102,15 @@ ModelRepositoryManager::VersionsToLoad(
   for (const auto& subdir : subdirs) {
     try {
       int64_t version = std::stoll(subdir);
+
+      if (subdir.length() > 1 && subdir.front() == '0') {
```
deadeyegoodwin: Add parens around if condition sub-expressions.

PR #460, opened 2019-07-15
Modify model manager to load / unload model based on dependency. Enable ...

GuanLuo:
ensemble config tests


deadeyegoodwin:
@GuanLuo You need to manually rebase...

PR #461, opened 2019-07-15
Fix #458 Autofill should not overwrite max_batch_size if it is set

### LINKED ISSUE #458: Explicitly configured max_batch_size is ignored by Autofill (ONNX, TensorFlow)
In the current state Autofill seems to set the `max_batch_size` of a model to 1 if TRTIS determined the model supports batching, and otherwise to 0 (for [ONNX](https://github.com/NVIDIA/tensorrt-inference-server/blob/93627b0021ee7cf9e3e381bb01e41ec45d594154/src/backends/onnx/autofill.cc#L211) and [TensorFlow](https://github.com/NVIDIA/tensorrt-inference-server/blob/a07509b5456f5502a4761be1b40cf0dad62ccda4/src/backends/tensorflow/autofill.cc#L192) models). If there's an explicitly configured `max_batch_size` it is silently ignored.
This I validated with a TensorFlow SavedModel having some inputs of shape `-1` and a single output of shape `-1`. In the model's `config.pbtxt` I set the `max_batch_size` to 30, but when picked up and served by the TRTIS the `max_batch_size` was set to 1.

There should be at least a warning about this behaviour.
I'm don't know if this is possible, but if the model supports batching TRTIS should be able to keep the `max_batch_size` that was explicitly configured.

GuanLuo:


jobegrabber:
Thanks!

PR #463, opened 2019-07-15
Cleanup usage of RequestStatus object from core server

deadeyegoodwin:


src/core/trtserver.cc
```diff
@@ -791,16 +790,16 @@ TRTSERVER_ServerModelStatus(
...
+      lserver->StatusManager(), ni::ServerStatTimerScoped::Kind::STATUS);
+
   ni::ServerStatus server_status;
-  lserver->HandleStatus(
-      &request_status, &server_status, std::string(model_name));
-  if (request_status.code() == ni::RequestStatusCode::SUCCESS) {
-    TrtServerProtobuf* protobuf = new TrtServerProtobuf(server_status);
-    *status = reinterpret_cast<TRTSERVER_Protobuf*>(protobuf);
-  }
+  lserver->GetStatus(&server_status, std::string(model_name));
```
GuanLuo: Check return value?

src/core/trtserver.cc
```diff
@@ -852,17 +851,16 @@ TRTSERVER_ServerInferAsync(
     infer_response_provider = del_response_provider;
   }
 
-  auto request_status = std::make_shared<ni::RequestStatus>();
-  lserver->HandleInfer(
-      request_status.get(), lprovider->Backend(), infer_request_provider,
-      infer_response_provider, infer_stats,
-      [infer_stats, timer, request_status, infer_response_provider, server,
-       complete_fn, userp]() mutable {
+  lserver->Infer(
```
GuanLuo: check return value, trigger `complete_fn` is not okay since the lambda function will not be called.

src/core/sequence_batch_scheduler.cc
```diff
@@ -845,12 +845,13 @@ SequenceBatchScheduler::SequenceBatch::SchedulerThread(
     }
 
     if ((payloads != nullptr) && !payloads->empty()) {
-      auto OnCompleteQueuedPayloads = [payloads](Status status) {
+      auto OnCompleteQueuedPayloads = [payloads](const Status& rstatus) {
```
dzier: I have to admit, this is something new to me. What is going on with `[ ]` around `payloads`? Is this a type of `shared_ptr` dereference?

src/core/server.cc
```diff
@@ -75,7 +74,7 @@ class ScopedAtomicIncrement {
 // InferenceServer
 //
 InferenceServer::InferenceServer()
-    : ready_state_(ServerReadyState::SERVER_INVALID), next_request_id_(1)
+    : ready_state_(ServerReadyState::SERVER_INVALID)
```
dzier: Is there a reason you don't set this as the default assignment in the header file?

src/core/sequence_batch_scheduler.cc
deadeyegoodwin: Its a lambda (anonymous function). The square brackets capture variables from the local env so that they are available when the function executes. It is up to dev to make sure that the captured variables are still valid when the lambda executes. payloads is a shared_ptr in this case which ensures that it remains live for the lambda to use.

src/core/server.cc
deadeyegoodwin: Stylistically we don't typically use default assignment.

PR #464, opened 2019-07-15
Add L0_shared_memory test + simple client

CoderHam:
PS: Wait till after Shared Memory Server changes to merge


src/clients/c++/simple_shm_client.cc
```diff
@@ -0,0 +1,374 @@
...
+
+  options->SetBatchSize(1);
+  for (const auto& output : infer_ctx->Outputs()) {
+    options->AddRawResult(output);
+  }
+
+  FAIL_IF_ERR(
+      infer_ctx->SetRunOptions(*options), "unable to set inference options");
+
+  // Create the data for the two input tensors. Initialize the first
```
deadeyegoodwin: Can you allocate and create the data directly in the shared memory. Also it will be more informative if you use a single shared memory region (or perhaps one for inputs and one for outputs).

src/clients/c++/simple_shm_client.cc
```diff
@@ -0,0 +1,374 @@
...
+
+  size_t input_byte_size = 16 * sizeof(int32_t);
+  size_t output_byte_size = 16 * sizeof(int32_t);
+  std::string shm_key;
+  int shm_fd, res;
+  void* shm_addr;
+
+  // Create  Input0 and Input1 in Shared Memory
+  // get shared memory file descriptor (NOT a file)
+  for (size_t n = 0; n < 2; n++) {
```
deadeyegoodwin: Opening a shared memory region and memmapping a certain size should be encapsulated in a function.

src/clients/c++/simple_shm_client.cc
```diff
@@ -0,0 +1,374 @@
...
+
+
+  // Send inference request to the inference server.
+  std::map<std::string, std::unique_ptr<nic::InferContext::Result>> results;
+  FAIL_IF_ERR(infer_ctx->Run(&results), "unable to run model");
+
+  // We expect there to be 2 results and them to be written by TRTIS into the
+  // corresponding shared memory location provided. The 16 result elements
+  // in both outputs are the sum and difference calculated by the model.
+  for (size_t n = 0; n < 2; n++) {
```
deadeyegoodwin: Don't open and memmap the region again...

src/clients/c++/simple_shm_client.cc
CoderHam: Will do

src/clients/c++/simple_shm_client.cc
CoderHam: Yes, I'll allocate the data directly in shared memory.
I kept two different shared memory regions (1 for each input tensor name). Similarly for the output tensors.

src/clients/c++/simple_shm_client.cc
CoderHam: Right. I can re-used the `shm_addr` from the previous mmap

src/clients/c++/simple_shm_client.cc
```diff
@@ -0,0 +1,331 @@
...
+    *(input1_shm + i) = 1;
+  }
+
+  // Create Output0 and Output1 in Shared Memory
+  shm_key = "/output_simple";
+  shm_fd = create_shared_region(shm_key, output_byte_size * 2);
+  int* output0_shm = (int*)(get_shm_addr(shm_fd, 0, output_byte_size * 2));
+  int* output1_shm = (int*)(output0_shm + 16);
+
+  err = input0->SetSharedMemory("/input_simple", 0, input_byte_size);
```
deadeyegoodwin: Needs update with Register/Unregister.

CoderHam: Yup working on it now

src/clients/c++/simple_shm_client.cc
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  // Create Output0 and Output1 in Shared Memory
+  shm_key = "/output_simple";
+  shm_fd = create_shared_region(shm_key, output_byte_size * 2);
+  int* output0_shm = (int*)(get_shm_addr(shm_fd, 0, output_byte_size * 2));
+  int* output1_shm = (int*)(output0_shm + 16);
+
+  // Register all Inputs and Outputs with TRTIS
+  infer_ctx->RegisterSharedMemory("input0_data", "/input_simple", 0, input_byte_size);
```
deadeyegoodwin: Just register 1 shared memory region for inputs and 1 for outputs. Then in SetSharedMemory use offsets.

deadeyegoodwin:
@CoderHam Just one small change. After that I am ok with merging (@GuanLuo can do that).

CoderHam:
The Unregister/RegisterSharedMemory() and SetSharedMemory are not defined here. You'll have to wait till the server side implementation is in before merging this in as it won't compile.

GuanLuo:
If you want, you can create stub functions which return UNSUPPORTED error code so it compiles without waiting for the actual server side implementation. Just that your L0_shared_memory test will fail, but that's okay.

qa/L0_shared_memory/test.sh
```diff
@@ -0,0 +1,81 @@
...
+    RET=1
+fi
+
+if [ `grep -c "localhost:8000" client_c++.log` != "5" ]; then
+    echo -e "\n***\n*** Failed. Expected 5 Host:localhost:8000 headers for C++ client\n***"
+    RET=1
+fi
+
+# Run with custom host header...
+$SIMPLE_SHM_CLIENT -v -H"Host:my_host_" >>client_c++_host.log 2>&1
```
GuanLuo: Any reason to also test with custom header?

CoderHam: To make it consistent with L0_simple_example but I don't think it is needed to be repeated again here, hence removing.

Dockerfile.QA
```diff
@@ -61,6 +61,8 @@ RUN mkdir -p qa/common && \
     cp -r docs/examples/model_repository/simple qa/L0_simple_callback_example/models/. && \
     mkdir qa/L0_simple_string_example/models && \
     cp -r docs/examples/model_repository/simple_string qa/L0_simple_string_example/models/. && \
+    mkdir qa/L0_shared_memory/models && \
```
deadeyegoodwin: You should call this L0_simple_shared_memory_example to match the naming of the other simple example tests.

src/clients/c++/CMakeLists.txt
```diff
@@ -103,13 +103,13 @@ target_link_libraries(
   image_client
   PRIVATE request
   PRIVATE ${OpenCV_LIBS}
+  rt
```
deadeyegoodwin: This shouldn't be needed?

src/clients/c++/image_client.cc
```diff
@@ -64,6 +66,47 @@ enum ScaleType { NONE = 0, VGG = 1, INCEPTION = 2 };
 
 enum ProtocolType { HTTP = 0, GRPC = 1 };
 
+int
+CreateSharedMemoryRegion(std::string shm_key, size_t batch_byte_size)
```
deadeyegoodwin: I thought you were removing image_client from this PR

src/clients/c++/request.h
```diff
@@ -277,6 +277,11 @@ class InferContext {
     /// -1.
     virtual const DimsList& Dims() const = 0;
 
+    virtual bool IsSharedMemory() const = 0;
```
deadeyegoodwin: The Input class provides methods to query the model configuration. Not methods to learn about setting you've made to the input. So these IsShared... GetShared... methods should be removed.

src/clients/c++/request.h
```diff
@@ -329,6 +334,23 @@ class InferContext {
     /// \param input The vector holding tensor string values.
     /// \return Error object indicating success or failure.
     virtual Error SetFromString(const std::vector<std::string>& input) = 0;
+
+    /// Set tensor values for this input by reference into a shared memory
+    /// region. The values are not copied and so the shared memory region and
+    /// its contents must not be modified or destroyed until this input is no
+    /// longer needed (that is until the Run()
+    //// call(s) that use the input have completed). For batched inputs this
```
deadeyegoodwin: Fix comment

src/clients/c++/request.h
```diff
@@ -347,6 +369,21 @@ class InferContext {
     /// model configuration. Variable-size dimensions are reported as
     /// -1.
     virtual const DimsList& Dims() const = 0;
+
+    /// The values for this output are set by reference in a shared memory
```
deadeyegoodwin: "Indicate that the result values for this output should be placed in a shared memory region instead of being returned in the inference response. The shared memory region must not be modified or..."

I don't think we call this multiple times for batched output, do we? Our entire batch of output must be contiguous in a single region.

Fix param formatting.

src/clients/c++/request.h
```diff
@@ -329,6 +334,23 @@ class InferContext {
     /// \param input The vector holding tensor string values.
     /// \return Error object indicating success or failure.
     virtual Error SetFromString(const std::vector<std::string>& input) = 0;
+
+    /// Set tensor values for this input by reference into a shared memory
+    /// region. The values are not copied and so the shared memory region and
+    /// its contents must not be modified or destroyed until this input is no
+    /// longer needed (that is until the Run()
+    //// call(s) that use the input have completed). For batched inputs this
+    /// function must be called batch-size times to provide all tensor values
```
deadeyegoodwin: Not called multiple times for batched input, correct?

src/clients/c++/request.h
```diff
@@ -745,6 +782,60 @@ class ModelControlContext {
...
+///   ctx->UnregisterAllSharedMemory();
+///   ...
+/// \endcode
+///
+class SharedMemoryControlContext {
+ public:
+  virtual ~SharedMemoryControlContext() = 0;
+
+  /// Register a shared memory region on the inference server. If the shared
+  /// memory region is already registered, it will return an error message
```
deadeyegoodwin: "If the shared memory region is already register, return error 'TRTSERVER_ERROR_ALREADY_EXISTS'.

src/clients/c++/request_common.cc
```diff
@@ -214,6 +216,18 @@ InputImpl::SetRaw(const std::vector<uint8_t>& input)
...
 
+Error
+InputImpl::SetSharedMemory(
+    const std::string& name, size_t offset, size_t byte_size)
+{
+  name_ = name;
+  offset_ = offset;
+  shm_byte_size_ = byte_size;
+  use_shm_ = true;
+  // total_byte_size_ += byte_size;
```
deadeyegoodwin: Delete, don't just comment out

src/clients/c++/request_common.cc
```diff
@@ -204,6 +205,7 @@ InputImpl::SetRaw(const uint8_t* input, size_t input_byte_size)
 
   bufs_.push_back(input);
   buf_byte_sizes_.push_back(input_byte_size);
+  use_shm_ = false;
```
deadeyegoodwin: Need better checking around use_shm_. It needs to be a tri-state. Initially it is "NONE". Once SetRaw is called change it to RAW and make sure that there are no calls to SetSharedMemory. Once SetSharedMemory is called set to "SHARED_MEMORY" and make sure no calls to SetRaw or SetSharedMemory (can only be one call to that). There is a Reset() function that should set it back to "NONE". Also should update the documentation on the Reset() function to indicate this.

src/clients/c++/request_common.cc
```diff
@@ -339,6 +353,19 @@ InputImpl::PrepareForRequest()
 
 //==============================================================================
 
+Error
```
deadeyegoodwin: You'll need to add a Reset() to the output to unset the shared memory settings.

src/clients/c++/request_common.h
```diff
@@ -195,21 +205,29 @@ class InputImpl : public InferContext::Input {
   // reallocs that could invalidate the pointer references into the
   // std::string objects.
   std::list<std::string> str_bufs_;
+
+  // Used only if working with Shared Memory
+  bool use_shm_;
+  std::string name_;
```
deadeyegoodwin: Use shm_name_, shm_offset_.

src/clients/c++/request_common.h
```diff
@@ -220,9 +238,18 @@ class OutputImpl : public InferContext::Output {
...
+  Error SetSharedMemory(
+      const std::string& name, size_t offset, size_t byte_size) override;
+
  private:
   const ModelOutput mio_;
   InferContext::Result::ResultFormat result_format_;
+
+  // Used only if working with Shared Memory
+  bool use_shm_;
+  std::string name_;
```
deadeyegoodwin: shm_name_
shm_offset_

src/clients/c++/request_grpc.cc
```diff
@@ -530,12 +630,13 @@ GrpcRequestImpl::GetResults(
 
     std::unique_ptr<GrpcResultImpl> result(
         new GrpcResultImpl(grpc_response_, infer_output));
+    // Check if output uses shared memory. If so handle differently
```
deadeyegoodwin: Remove comment since no corresponding code

src/clients/c++/request_grpc.cc
```diff
@@ -873,6 +984,20 @@ InferGrpcContext::Create(
   return err;
 }
 
+bool
+InferGrpcContextImpl::HasSharedMemory(std::string output_name) const
```
deadeyegoodwin: Who calls this?

src/clients/c++/request_http.cc
```diff
@@ -1505,6 +1692,20 @@ InferHttpContextImpl::AsyncTransfer()
   } while (!exiting_);
 }
 
+bool
+InferHttpContextImpl::HasSharedMemory(std::string output_name) const
```
deadeyegoodwin: Who calls this?

src/clients/c++/request.h
CoderHam: Right. This should be once per batch.

src/clients/c++/request.h
CoderHam: Single call per batch, thats right.

src/clients/c++/request_http.cc
CoderHam: No longer required

src/clients/c++/request_grpc.cc
CoderHam: No longer required

PR #468, opened 2019-07-16
Implement model control API

GuanLuo:


src/core/grpc_service.proto
```diff
@@ -188,6 +194,62 @@ message HealthResponse
...
+    //@@    .. cpp:enumerator:: Type::UNLOAD = 0
+    //@@
+    //@@       To unload the specified model.
+    //@@
+    UNLOAD = 0;
+
+    //@@    .. cpp:enumerator:: Type::LOAD = 1 << 0
+    //@@
+    //@@       To load the specified model. Or to reload if the model has
+    //@@       been loaded.
```
CoderHam: How about: "To load (or reload) the specified model."

src/core/grpc_service.proto
```diff
@@ -188,6 +194,62 @@ message HealthResponse
...
+    //@@
+    LOAD = 1;
+  }
+
+  //@@
+  //@@  .. cpp:var:: string model_name
+  //@@
+  //@@     The target model.
```
CoderHam: The target model _name_.

src/core/grpc_service.proto
GuanLuo: it seems to be a bit confusing without elaborating on when "reload" is going to be performed

qa/L0_lifecycle/test.sh
```diff
@@ -373,6 +373,45 @@ wait $SERVER_PID
...
+SERVER_LOG="./inference_server_$LOG_IDX.log"
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+set +e
+code=`curl -s -w %{http_code} -X POST localhost:8000/api/control/unload/graphdef_float32_float32_float32`
```
deadeyegoodwin: You should test "load" as well

src/core/grpc_service.proto
```diff
@@ -188,6 +194,62 @@ message HealthResponse
...
+  //@@     Types of control operation
+  //@@
+  enum Type {
+    //@@    .. cpp:enumerator:: Type::UNLOAD = 0
+    //@@
+    //@@       To unload the specified model.
+    //@@
+    UNLOAD = 0;
+
+    //@@    .. cpp:enumerator:: Type::LOAD = 1 << 0
```
deadeyegoodwin: Using "1 << 0" implies that this is a bit-mask where multiple control options can be specified. I don't think that is the intent.

src/core/server.cc
```diff
@@ -388,6 +388,32 @@ InferenceServer::HandleStatus(
   }
 }
 
+void
+InferenceServer::HandleControl(
```
deadeyegoodwin: This needs to be updated to match the new style with TRTSERVER API. There should be both a ModelLoad and ModelUnload function instead of combining them.

For example should return Status, remove RequestStatus arg, ...

src/core/trtserver.h
```diff
@@ -336,6 +336,14 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerModelStatus(
     TRTSERVER_Server* server, TRTSERVER_Protobuf** status,
     const char* model_name);
 
+// Load the requested model.
```
deadeyegoodwin: Does not return until model is loaded? Returned error indicates if model loaded successfully or not?

src/core/trtserver.h
```diff
@@ -336,6 +336,14 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerModelStatus(
...
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_LoadModel(
+    TRTSERVER_Server* server, const char* model_name);
+
+// Unload the requested model.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_UnloadModel(
```
deadeyegoodwin: Does not return until model is unloaded? Returned error indicates if model unloaded successfully or not?

deadeyegoodwin:
@GuanLuo You need to manually rebase and do some updates to conform the RequestStatus changes.

src/core/trtserver.cc
```diff
@@ -923,6 +923,32 @@ TRTSERVER_ServerModelStatus(
   return nullptr;  // success
 }
 
+TRTSERVER_Error*
+TRTSERVER_LoadModel(TRTSERVER_Server* server, const char* model_name)
+{
+  ni::InferenceServer* lserver = reinterpret_cast<ni::InferenceServer*>(server);
+
+  ni::ServerStatTimerScoped timer(
+      lserver->StatusManager(), ni::ServerStatTimerScoped::Kind::CONTROL);
```
deadeyegoodwin: It seems that we should record both success and failed control requests. This can be done as a separate change.

PR #474, opened 2019-07-18
Checks the compute capapbility of GPUs before loading the models

tanmayv25:
The minimum supported compute capability of GPU for TRTIS can now be specified as a cmake setting while building it. By default, it is set to 6.0. Next, only the GPUs having required compute capability will be loaded with the model.


src/core/model_config_utils.cc
```diff
@@ -905,4 +919,75 @@ CheckAllowedModelOutput(
...
+        "The compute capability of gpu id " + std::to_string(gpu_id) + " is '" +
+            std::to_string(cuprops.major) + "." +
+            std::to_string(cuprops.minor) +
+            "' which is lesser than the minimum supported of '" +
+            TRTIS_MIN_COMPUTE_CAPABILITY + "' by TRTIS.");
+  }
+}
+
+Status
+ParseMinComputeCapability(std::pair<int, int>& min_compute_capability)
```
CoderHam: Use std::stof instead of custom parser. Remember to check for failure and report error if invalid compute compatibility string.

src/core/model_config_utils.cc
```diff
@@ -905,4 +919,75 @@ CheckAllowedModelOutput(
...
+  // gpu compute capability.
+  std::pair<int, int> min_compute_capability;
+  // Validates and populates the minimum supported Compute Capability by the
+  // Server.
+  Status status = ParseMinComputeCapability(min_compute_capability);
+  if (!status.IsOk()) {
+    return status;
+  }
+
+  if ((cuprops.major > min_compute_capability.first) ||
```
CoderHam: You can use float comparisons directly here.

nieksand:
Is there some compelling reason to use sm 6 as the minimum?  

AWS g3s instances have M60 GPUs (Maxwell architecture, sm 5) and are the most cost-effective option for inference on my workloads.  With this change it sounds like TRTIS won't work on those anymore.

AWS g4 instances with T4 cards, despite having been announced at GTC back in March, still haven't been released by AWS.  These would be the ideal replacement for the geriatric g3 series.

AWS p2 instances have K80s (Kepler architecture, sm 3).

AWS p3 instances have V100s (Volta, sm 7) and will be the only instances to still work with TRTIS if the minimum sm becomes 6.  Cost/perf for these is sadly garbage for my use case.

Or perhaps I'm misunderstanding what this PR is changing... also quite possible!

src/servers/main.cc
```diff
@@ -734,6 +759,15 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
 int
 main(int argc, char** argv)
 {
+#ifdef TRTIS_ENABLE_GPU
+  // Validate the format of specified minimum supported GPU compute capability
+  if (!ValidateMinComputeCapability()) {
+    LOG_ERROR << "TRTIS_MIN_COMPUTE_CAPABILITY should be specified in the "
```
GuanLuo: What happens if running on CPU-only system? And what if the system contains low C.C. GPU but all models are specified to be on CPU

build/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ option(TRTIS_ENABLE_GCS "Include GCS Filesystem support in server" OFF)
 set(TRTIS_ONNXRUNTIME_INCLUDE_PATHS "" CACHE PATH "Paths to ONNXRuntime includes")
 set(TRTIS_PYTORCH_INCLUDE_PATHS "" CACHE PATH "Paths to PyTorch includes")
 set(TRTIS_EXTRA_LIB_PATHS "" CACHE PATH "Extra library paths for TRTIS build")
+set(TRTIS_MIN_COMPUTE_CAPABILITY "6.0" CACHE STRING "The minimum compute capability supported by TRTIS" )
```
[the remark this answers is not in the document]
deadeyegoodwin: The default value is 6.0 because that is what TRTIS officially supports. If someone wants to build the framework backends with support for earlier CCs that can do that and then set TRTIS_MIN_COMPUTE_CAPABILITY appropriately so that TRTIS allows those GPUs to be used.

src/servers/main.cc
GuanLuo: oh, this is just checking the format.

src/core/model_config_utils.cc
```diff
@@ -505,6 +518,7 @@ ValidateModelConfig(
                     std::to_string(gid) + ", valid GPUs are 0 - " +
                     std::to_string(dcnt - 1));
           }
+          RETURN_IF_ERROR(CheckGPUCompatibility(gid));
```
deadeyegoodwin: Replace dcnt above with GetSupportGPUs and then you can just check gid for membership in the set and if not there print the error that is currently in CheckGPUCompatibility. In general in the normalize and validate routines don't want to replace cudaGetDeviceCount and device_cnt with supported_gpus.

src/servers/main.cc
```diff
@@ -734,6 +759,15 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
 int
 main(int argc, char** argv)
 {
+#ifdef TRTIS_ENABLE_GPU
+  // Validate the format of specified minimum supported GPU compute capability
+  if (!ValidateMinComputeCapability()) {
```
deadeyegoodwin: We want a compile-time failure. Can you see if you can use static_assert? Perhaps static_assert(TRTIS_MIN_COMPUTE_CAPABILITY >= 1.0) will be sufficient.

src/core/model_config_utils.h
```diff
@@ -131,4 +131,12 @@ Status ValidateModelOutput(const ModelOutput& io, int32_t max_batch_size);
 Status CheckAllowedModelOutput(
     const ModelOutput& io, const std::set<std::string>& allowed);
 
+#ifdef TRTIS_ENABLE_GPU
+/// Validates the compute capability of the GPU indexed
+/// \param The index of the target GPU.
+/// \return The error status. A non-OK status means the target GPU is
+/// \ not supported
```
CoderHam: The '\' before not supported is not needed.

src/core/model_config_utils.cc
```diff
@@ -905,4 +919,32 @@ CheckAllowedModelOutput(
...
+  float compute_compability = cuprops.major + (cuprops.minor/10.0);
+  if (compute_compability >= std::stof(TRTIS_MIN_COMPUTE_CAPABILITY)) {
+    return Status::Success;
+  } else {
+    return Status(
+        RequestStatusCode::UNSUPPORTED,
+        "The compute capability of gpu id " + std::to_string(gpu_id) + " is '" +
+            std::to_string(cuprops.major) + "." +
+            std::to_string(cuprops.minor) +
+            "' which is lesser than the minimum supported of '" +
```
CoderHam: .. lesser than the minimum (<>) supported by TRTIS.

deadeyegoodwin:
> Is there some compelling reason to use sm 6 as the minimum?
> 
> AWS g3s instances have M60 GPUs (Maxwell architecture, sm 5) and are the most cost-effective option for inference on my workloads. With this change it sounds like TRTIS won't work on those anymore.
> 
> AWS g4 instances with T4 cards, despite having been announced at GTC back in March, still haven't been released by AWS. These would be the ideal replacement for the geriatric g3 series.
> 
> AWS p2 instances have K80s (Kepler architecture, sm 3).
> 
> AWS p3 instances have V100s (Volta, sm 7) and will be the only instances to still work with TRTIS if the minimum sm becomes 6. Cost/perf for these is sadly garbage for my use case.
> 
> Or perhaps I'm misunderstanding what this PR is changing... also quite possible!

@nieksand Not sure if you saw this in the review comment:

The default value is 6.0 because that is what TRTIS officially supports. If someone wants to build the framework backends with support for earlier CCs that can do that and then set TRTIS_MIN_COMPUTE_CAPABILITY appropriately so that TRTIS allows those GPUs to be used.

nieksand:
Thanks for the heads up.  I'm hoping AWS gets the g4 series out before I need a newer version of TRTIS.

PR #475, opened 2019-07-18
Add lazy model handling mode (disable periodic model polling)

GuanLuo:


qa/L0_lifecycle/test.sh
```diff
@@ -227,8 +227,8 @@ for i in graphdef netdef plan ; do
 done
 cp -r $DATADIR/qa_model_repository/savedmodel_float32_float32_float32 .
 
-SERVER_ARGS="--model-store=`pwd`/models --allow-poll-model-repository=false \
-             --repository-poll-secs=1 --exit-timeout-secs=5"
+SERVER_ARGS="--model-store=`pwd`/models --allow-poll-model-repository=true \
```
deadeyegoodwin: This is confusing because --allow seems to turn it on but --repository-poll-secs=0 turns if off. Can you just use --allow-poll-model-repository=false and remove --repository-poll-secs?

qa/L0_lifecycle/test.sh
```diff
@@ -286,8 +286,8 @@ for i in graphdef ; do
     cp -r $DATADIR/qa_model_repository/${i}_int32_int32_int32 models/.
 done
 
-SERVER_ARGS="--model-store=`pwd`/models --repository-poll-secs=1 \
-             --allow-poll-model-repository=false --exit-timeout-secs=5"
+SERVER_ARGS="--model-store=`pwd`/models --repository-poll-secs=0 \
```
deadeyegoodwin: same comment as above

qa/L0_lifecycle/test.sh
```diff
@@ -373,15 +373,53 @@ wait $SERVER_PID
...
 # Send HTTP request to invalid endpoints
 rm -fr models
 mkdir models
 for i in graphdef ; do
     cp -r $DATADIR/qa_model_repository/${i}_int32_int32_int32 models/.
 done
 
-SERVER_ARGS="--model-store=`pwd`/models --repository-poll-secs=1 \
-             --allow-poll-model-repository=false --exit-timeout-secs=5"
+SERVER_ARGS="--model-store=`pwd`/models --repository-poll-secs=0 \
```
deadeyegoodwin: Same comment as above

src/core/model_repository_manager.cc
```diff
@@ -842,17 +842,20 @@ ModelRepositoryManager::Create(
...
+      return Status(
+          RequestStatusCode::INTERNAL,
+          "Unexpected initial state for model repository");
+    }
 
-  // model loading / unloading error will be printed but ignored
-  local_manager->LoadModelByDependency();
+    RETURN_IF_ERROR(local_manager->Update(added, deleted, modified));
+
+    // [TODO] fix this. On init, should return loading error and let
```
deadeyegoodwin: Is this DLIS-506?

src/core/trtserver.cc
```diff
@@ -225,6 +228,7 @@ class TrtServerOptions {
  private:
   std::string server_id_;
   std::string repo_path_;
+  bool model_polling_;
```
deadeyegoodwin: Update constructor to default this to true. This may conflict with the in-flight TRTSERVER metrics change...

qa/L0_lifecycle/test.sh
```diff
@@ -373,15 +373,53 @@ wait $SERVER_PID
 LOG_IDX=$((LOG_IDX+1)) 
 
 
+# disable model polling, no model in the repository should be loaded
```
deadeyegoodwin: Model polling is separate from lazy model and we can't change that behavior now. Even if polling is disabled we still need to load the model repository on startup. Perhaps --allow-model-control which defaults to false. We need to check in the parser to make sure that both are not enabled.

src/core/trtserver.h
```diff
@@ -248,6 +248,11 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerOptionsSetServerId(
 TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerOptionsSetModelRepositoryPath(
     TRTSERVER_ServerOptions* options, const char* model_repository_path);
 
+// Enable or disable periodic polling on model repository in a server
+// options.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerOptionsSetModelPolling(
```
deadeyegoodwin: Instead of a true/false I think we need a ServerOptionsSetModelControlMode with options None, Poll, and Explicit. None and poll mean that the models are loaded on startup. Poll means the Polling API can be used. Explicit means that the load/unload API can be used.

qa/L0_lifecycle/test.sh
GuanLuo: This is how we can tell the server to only poll model repository once. We should also update this usage then (i.e. whether loading the model repository on startup depends on `allow-model-control`)
> --repository-poll-secs	Interval in seconds between each poll of the model repository to check for changes. A value of zero indicates that the repository is checked only a single time at startup. Valid only when --allow-poll-model-repository=true is specified.

src/core/model_repository_manager.cc
GuanLuo: Yes.

qa/L0_lifecycle/test.sh
deadeyegoodwin: Oh, that makes sense.... But --allow-poll-model-repository=false is redundant with true and poll-secs=0... but that is just the way it is.
Yes we need to add the allow-model-control flag and update the descriptions. But default values should be set so current behavior is maintained

qa/L0_lifecycle/test.sh
```diff
@@ -373,15 +373,55 @@ wait $SERVER_PID
...
+SERVER_LOG="./inference_server_$LOG_IDX.log"
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+set +e
+code=`curl -s -w %{http_code} localhost:8000/api/health/ready`
```
deadeyegoodwin: Isn't this redundant since run_server waits until the server is ready

src/core/model_repository_manager.cc
deadeyegoodwin: I think it is ok to but the DLIS-506 in the comment. It doesn't mean anything externally but it helps us remember.

src/core/server.cc
```diff
@@ -124,10 +124,12 @@ InferenceServer::Init()
 
   // Create the global manager for the repository. For now, all models are
```
deadeyegoodwin: Need to update this comment

src/core/trtserver.h
```diff
@@ -280,6 +287,23 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerOptionsSetServerId(
 TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerOptionsSetModelRepositoryPath(
     TRTSERVER_ServerOptions* options, const char* model_repository_path);
 
+// Set the model control mode in a server options. For each mode the models
+// will be managed as the following:
+//
+//   TRTSERVER_MODEL_CONTROL_NONE: the models in model repository will be loaded
+//   on startup. The model states will not be modified after startup.
```
deadeyegoodwin: Change the last sentence to: "After startup any changes to the model repository will be ignored. Calling TRTSERVER_ServerPollModelRepository will result in an error."

src/core/trtserver.h
```diff
@@ -280,6 +287,23 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_ServerOptionsSetServerId(
...
+//
+//   TRTSERVER_MODEL_CONTROL_POLL: the models in model repository will be loaded
+//   on startup. The model repository will be polled periodically and the models
```
deadeyegoodwin: "The model repository can be polled periodically using TRTSERVER_ServerPollModelRepository and the server will load, unload, and updated models according to ..."

src/servers/main.cc
```diff
@@ -213,9 +217,12 @@ std::vector<Option> options_{
      "controlled by 'repository-poll-secs'."},
     {OPTION_POLL_REPO_SECS, "repository-poll-secs",
      "Interval in seconds between each poll of the model repository to check "
-     "for changes. A value of zero indicates that the repository is checked "
-     "only a single time at startup. Valid only when "
+     "for changes. Valid only when "
      "--allow-poll-model-repository=true is specified."},
+    {OPTION_ALLOW_MODEL_CONTROL, "allow-model-control",
+     "Allow to load or to unload models explicitly using model control API. "
+     "If true the model repository will not be polled at startup. Ignored "
```
deadeyegoodwin: "If true the models in the model repository will not be loaded at startup. Cannot be specified if --allow-pool-model-repository is true."

src/servers/main.cc
```diff
@@ -669,6 +680,22 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
...
+    LOG_ERROR << "--allow-model-control and --allow-poll-model-repository "
+              << "can not be both set to true";
+    LOG_ERROR << Usage();
+    return false;
+  }
+
+  TRTSERVER_Model_Control_Mode control_mode;
+  if (allow_model_control) {
+    control_mode = TRTSERVER_MODEL_CONTROL_EXPLICIT;
+  } else if (allow_poll_model_repository) {
```
deadeyegoodwin: repository_poll_secs > 0

src/core/model_repository_manager.cc
```diff
@@ -835,24 +838,28 @@ ModelRepositoryManager::Create(
...
       new ModelRepositoryManager(
           status_manager, repository_path, backend_config_map,
-          !strict_model_config, polling_enabled, std::move(life_cycle)));
+          !strict_model_config, polling_enabled, model_control_enabled,
+          std::move(life_cycle)));
 
   // Similar to PollAndUpdate(), but simplier
   std::set<std::string> added, deleted, modified, unmodified;
-  if (polling_enabled) {
+  if (!model_control_enabled) {
```
deadeyegoodwin: Don't want to allow polling if control mode is NONE.

GuanLuo: This is the poll for startup, so always poll as long as the mode is not EXPLICIT

GuanLuo:
Rebased and update trtserver.h to format in #485

CoderHam:
Looks good to me

PR #476, opened 2019-07-18
Add TRTSERVER API for metrics

deadeyegoodwin:


src/core/trtserver.cc
```diff
@@ -188,13 +189,37 @@ TrtServerProtobuf::Serialize(const char** base, size_t* byte_size) const
...
+  void Serialize(const char** base, size_t* byte_size);
+
+ private:
+  std::string serialized_;
+};
+
+void
+TrtServerMetrics::Serialize(const char** base, size_t* byte_size)
+{
+  serialized_ = ni::Metrics::SerializedMetrics();
```
GuanLuo: Wrap parts related to `ni::Metrics` with `#ifdef TRTIS_ENABLE_METRICS` and return error message otherwise?

PR #479, opened 2019-07-18
Add logging options to TRTSERVER API

deadeyegoodwin:


src/servers/main.cc
```diff
@@ -717,6 +716,19 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
...
+      "setting log info enable");
+  FAIL_IF_ERR(
+      TRTSERVER_ServerOptionsSetLogWarn(server_options, log_warn),
+      "setting log warn enable");
+  FAIL_IF_ERR(
+      TRTSERVER_ServerOptionsSetLogError(server_options, log_error),
+      "setting log error enable");
+  FAIL_IF_ERR(
+      TRTSERVER_ServerOptionsSetLogVerbose(server_options, log_verbose),
+      "setting log verbose level");
```
GuanLuo: The old implementation in line 663 can be removed

deadeyegoodwin: That's needed to enable the logging in main, grpc_server, http_server. They create their own logger object that is distinct from the one in libtrtserver.so. Ultimately we want to have a proper logging API that routes all logging output to the client to do with as they like but for now just doing this.

PR #480, opened 2019-07-19
Fix L0_sequence_batcher

deadeyegoodwin:


src/core/server.cc
```diff
@@ -298,21 +298,13 @@ InferenceServer::Infer(
   // it goes out of scope which can cause the model to be unloaded,
   // and we don't want that to happen when a request is in flight.
   auto OnCompleteHandleInfer = [this, OnCompleteInfer, backend,
-                                response_provider, infer_stats,
+                                response_provider,
```
GuanLuo: can also remove `infer_stats` from `Infer()`? Since it is captured one layer up

deadeyegoodwin: Needs to be passed into backend.Run() so that stats can be recorded.

GuanLuo: Oh right, I missed that line.

PR #482, opened 2019-07-19
Add control context in client library

GuanLuo:
This change depends on #468 (implemented based on `grpc_service.proto` in that change)

### DIFF
--- a/src/clients/c++/request.h
+++ b/src/clients/c++/request.h
@@ -705,8 +705,43 @@ class ProfileContext {
   virtual Error StartProfile() = 0;
 
   /// Stop profiling on the inference server.
-  // \return Error object indicating success or failure.
+  /// \return Error object indicating success or failure.
   virtual Error StopProfile() = 0;
+};
+
+//==============================================================================
+/// A ControlContext object is used to control the model loading / unloading
+/// on the inference server. Once created a ControlContext object can be used
+/// repeatedly.
+///
+/// A ControlContext object can use either HTTP protocol or GRPC protocol
+/// depending on the Create function (ControlHttpContext::Create or
+/// ControlGrpcContext::Create). For example:
+///
+/// \code
+///   std::unique_ptr<ControlContext> ctx;
+///   ControlGrpcContext::Create(&ctx, "localhost:8000");
+///   std::string model_name = "model";
+///   ctx->Load(model_name);
+///   ...
+///   ctx->Unload(model_name);
+///   ...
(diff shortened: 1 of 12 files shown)


src/clients/c++/request.h
```diff
@@ -705,10 +705,45 @@ class ProfileContext {
...
+///   ...
+///   ctx->Unload(model_name);
+///   ...
+/// \endcode
+///
+class ControlContext {
+ public:
+  virtual ~ControlContext() = 0;
+
+  /// Load a model on the inference server. If the model has been loaded,
```
deadeyegoodwin: "If the model is already loaded, ..."

src/clients/c++/request.h
```diff
@@ -705,10 +705,45 @@ class ProfileContext {
...
+  /// it will be reloaded to use the latest configuration.
+  /// \param model_name The name of the model to be loaded.
+  /// \return Error object indicating success or failure.
+  virtual Error Load(const std::string& model_name) = 0;
+
+  /// Unload a model on the inference server.
```
deadeyegoodwin: "... model from the ..."

src/clients/c++/request.h
```diff
@@ -705,10 +705,45 @@ class ProfileContext {
...
+  /// \param model_name The name of the model to be unloaded.
+  /// \return Error object indicating success or failure.
```
deadeyegoodwin: Is it an error to attempt to unload a model that is not loaded on the server? If so then it should be a error code that is not otherwise returned by the server and we should document it here and in trtserver.h.

src/clients/c++/request_http.h
```diff
@@ -139,6 +139,25 @@ class ProfileHttpContext {
       const std::map<std::string, std::string>& headers, bool verbose = false);
 };
 
+//==============================================================================
+/// ControlHttpContext is the HTTP instantiation of ControlContext.
+///
+class ControlHttpContext {
+ public:
+  /// Create context that controls model ready state on a server using HTTP
```
deadeyegoodwin: "controls mode ready state" is a strange way to say it. Can you just directly say that it allows models to be loaded and unloaded.

src/clients/c++/request.h
GuanLuo: It shouldn't be an error since the model becomes unavailable anyway. I updated to be "unloading a model that is not loaded on server has no affect and success code will be returned" and update the server side implementation accordingly

PR #488, opened 2019-07-22
Rename Model Load/Unload from Control to ModelControl

CoderHam:


src/clients/c++/request.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
+ // Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
```
GuanLuo: space

src/core/server_status.proto
```diff
@@ -350,9 +350,9 @@ message ServerStatus
...
   HealthRequestStats health_stats = 8;
 
-  //@@  .. cpp:var:: ControlRequestStats control_stats
+  //@@  .. cpp:var:: ModelControlRequestStats control_stats
   //@@
-  //@@     Statistics for Control requests.
+  //@@     Statistics for ModelControl requests.
   //@@
-  ControlRequestStats control_stats = 9;
+  ModelControlRequestStats control_stats = 9;
```
GuanLuo: Also rename `control_stats`? Is your new API going to add similar thing here?

CoderHam: No, I don't think I will need to add Stats for Register/Unregistering Shared Memory Regions.

deadeyegoodwin: We should have stats for all the APIs. There is no reason to treat register/unregister differently. So control_stats needs to change to model_control_stats.

CoderHam: Already done. Will incorporate the stats functionality in my changes for the shared memory control stats

PR #491, opened 2019-07-23
Fix compute timer scope

GuanLuo:


src/backends/caffe2/netdef_backend.h
```diff
@@ -128,7 +128,7 @@ class NetDefBackend : public InferenceBackend {
     std::unique_ptr<Caffe2Workspace> workspace_;
   };
 
-  std::vector<Context> contexts_;
+  std::vector<std::unique_ptr<Context>> contexts_;
```
deadeyegoodwin: Because you did this you can delete the Context move constructor, and add DISALLOW_MOVE(Context).
See plan_backend.h for example.

PR #493, opened 2019-07-24
Add TRTSERVER callback for result tensor buffer allocation

deadeyegoodwin:


src/core/trtserver.h
```diff
@@ -108,6 +109,64 @@ TRTSERVER_EXPORT const char* TRTSERVER_ErrorCodeString(TRTSERVER_Error* error);
...
+/// memory block of size 'byte_size' for result tensor called
+/// 'tensor_name'. The buffer must be allocated in the memory region
+/// identified by 'region' and 'region_id'. The 'userp' data is the
+/// same as what is supplied in the call to
+/// TRTSERVER_ServerInferAsync.
+///
+/// The function may return 'buffer' == nullptr to indicate that an
+/// allocation in the requested 'region' is not possible. In this case
+/// the function may be called again for the same 'tensor_name' but
+/// with a different 'region'.
```
GuanLuo: Is the returning Error object a nullptr in this case?

src/core/provider.cc
```diff
@@ -801,10 +804,26 @@ DelegatingInferResponseProvider::AllocateOutputBuffer(
...
+    }
+    if (buffer == nullptr) {
+      return Status(
+          RequestStatusCode::UNAVAILABLE,
+          "unable to allocate memory for result tensor '" + name + "'");
+    }
+
+    *content = buffer;
+    output->ptr_ = buffer;
+    output->buffer_.reset(reinterpret_cast<char*>(buffer));
```
GuanLuo: So the allocated memory will be managed by the response provider. For memory that are allocated on GPU, is the smart pointer able to release it properly?

src/core/trtserver.h
deadeyegoodwin: Yes, any non-null error returned will indicate that something went wrong and the server will fail the request.

src/core/provider.cc
deadeyegoodwin: No, this change is only an intermediate state for the provider. After I move the grpc and http logic out I plan on simplifying the provider since buffer lifetime management must be handled by the client of TRTSERVER API. There is likely to be some changes in the API to allow that.

PR #494, opened 2019-07-24
Add model control test. Fix bugs

GuanLuo:


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -826,5 +826,196 @@ def test_dynamic_file_delete(self):
...
+                ctx = ModelControlContext(pair[0], pair[1], True)
+                ctx.load("unknown_model")
+                self.assertTrue(False, "expected unknown model failure")
+            except InferenceServerException as ex:
+                self.assertEqual("inference:0", ex.server_id())
+                self.assertGreater(ex.request_id(), 0)
+                self.assertTrue(
+                    ex.message().startswith("failed to stat directory for model"))
+
+        # Load ensemble model first, should fail due to depending model is not loaded
```
deadeyegoodwin: If they load an ensemble model we should automatically  load all the dependent models. Otherwise user has to discover all the necessary models and figure out the right order to load them.

src/clients/c++/request_http.cc
```diff
@@ -652,7 +652,7 @@ ModelControlHttpContextImpl::SendRequest(
   curl_easy_setopt(curl, CURLOPT_URL, full_url.c_str());
   curl_easy_setopt(curl, CURLOPT_USERAGENT, "libcurl-agent/1.0");
   // use POST method
-  curl_easy_setopt(curl, CURLOPT_POST, 1L);
+  curl_easy_setopt(curl, CURLOPT_POSTFIELDS, "");
```
deadeyegoodwin: What does this change do?

GuanLuo: To provide an empty request body for POST request. Otherwise libcurl will use the default reader, which reads from stdin.

qa/L0_lifecycle/lifecycle_test.py
GuanLuo: Sounds good. Should I make the change on a separate PR?

src/clients/c++/request_http.cc
deadeyegoodwin: ok

qa/L0_lifecycle/lifecycle_test.py
deadeyegoodwin: Yes, that will be ok.

PR #495, opened 2019-07-24
Partial sequence batcher fix (Pytorch)

CoderHam:


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -50,6 +51,10 @@ LibTorchBackend::Context::Context(
 
 LibTorchBackend::Context::~Context()
 {
+  torch_model_.reset();
+  c10::cuda::CUDACachingAllocator::emptyCache();
```
GuanLuo: What is the cache used for in libtorch? If there are other pytorch models and only one of them is to be unloaded, which calls emptyCache(), will this affects other models?

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -50,6 +51,10 @@ LibTorchBackend::Context::Context(
...
+  input_index_map_.clear();
+  output_index_map_.clear();
```
GuanLuo: I don't think you need to call `clear()` explicitly. The same for `torch_model_` unless it needs to be destroyed before calling `emptyCache()`?

CoderHam: Yes the clear is not needed. No, it isn't necessary to reset the `torch_model_` before calling `emptyCache()`.

src/backends/pytorch/libtorch_backend.cc
CoderHam: The cache is used for some CUDA context and internal pytorch optimization.
Here is the relevant discussion: https://github.com/pytorch/pytorch/issues/17433#issuecomment-479102003
It might not but I will run a test for this and verify.

CoderHam: Ran the following tests:
1. `perf_client` on a pytorch model (A) while removing another pytorch model (B) from the directory.
- This unloads B and therefore calls the `emptyCache()` method inside `~Context()`
2. The follow `image_client` script (on a single coffee mug image) and unloaded the model during the test. Even after emptyCache the results were unaffected.
```
for i in {1..200}; do install/bin/image_client -m resnet50_libtorch -s INCEPTION -c 1 -b 1 images/mug.jpg; done > temp.txt
grep -c COFFEE temp.txt
```
> 200

PR #499, opened 2019-07-25
Allow variable-size reshape if the variable-size dimensions are aligned

GuanLuo:


src/core/provider.cc
```diff
@@ -449,12 +450,28 @@ InferResponseProvider::FinalizeResponse(const InferenceBackend& is)
...
+          }
+        }
+
+        for (const auto& dim : output_config->dims()) {
+          if (dim == -1) {
+            poutput->mutable_raw()->add_dims(variable_size_values.front());
+            variable_size_values.pop_front();
+          } else {
+            poutput->mutable_raw()->add_dims(dim);
+            ;
```
deadeyegoodwin: remove

qa/L0_infer_reshape/test.sh
```diff
@@ -44,7 +44,13 @@ for i in \
         zero_1_float32 \
         zero_2_float32 \
         zero_3_float32 \
-        zero_4_float32 ; do
+        zero_4_float32 \
```
deadeyegoodwin: Did you add a passing test for Boris' exact case of 1 or more pre-pended '1' dimensions?

GuanLuo: Yes, the `zero_2_int32` and `zero_3_int32` models are set up for that case

PR #503, opened 2019-07-26
Server side implementation for System Shared Memory

CoderHam:
- **Unregister/RegisterSharedMemory**: register shared memory (shm_opens and mmaps when needed/registered. close and munmap when not needed/unregistered).
- **UnregisterAllSharedMemory**: unregister _all_ shared memory regions. 
- **GetLiveSharedMemory**: get list (names) of  all shared memory regions that are currently registered.

### DIFF
--- a/src/clients/c++/request.h
+++ b/src/clients/c++/request.h
@@ -770,18 +770,25 @@ class SharedMemoryControlContext {
  public:
   virtual ~SharedMemoryControlContext() = 0;
 
-  /// Register a model on the inference server. If the shared memory is already
-  /// loaded, it will be reloaded to update the LRU policy. \param name The
-  /// name of the shared memory region to be registered. \return Error object
-  /// indicating success or failure.
+  /// Register a shared memory region on the inference server. If the shared
+  /// memory region is already registered, it will be unregistered and then
+  /// re-registered.
+  /// \param name The user-given name for the shared memory region to be
+  /// registered.
+  /// \param shm_key The unique name of the location in shared memory being
+  /// registered.
+  /// \param offset The offset into the shared memory region.
+  /// \param byte_size The size, in bytes of the tensor data.
+  /// \return Error object indicating success or failure.
   virtual Error RegisterSharedMemory(
       const std::string& name, const std::string& shm_key, size_t offset,
       size_t byte_size) = 0;
 
-  /// Unload a model from the inference server. unregistering a shared memory
-  /// region that is not registered on server has no affect and success code
-  /// will be returned. \param name The name of the shared memory region to
-  /// be unregistered. \return Error object indicating success or failure.
+  /// Unregister a registered shared memory region on the inference server. If
+  /// the shared memory region is not registered, do nothing and return failure.
+  /// \param name The user-given name for the shared memory region to be
+  /// unregistered.
+  /// \return Error object indicating success or failure.
   virtual Error UnregisterSharedMemory(const std::string& name) = 0;
 };
 
--- a/src/clients/c++/request.h
+++ b/src/clients/c++/request.h
@@ -771,8 +771,8 @@ class SharedMemoryControlContext {
   virtual ~SharedMemoryControlContext() = 0;
 
   /// Register a shared memory region on the inference server. If the shared
-  /// memory region is already registered, it will be unregistered and then
-  /// re-registered.
+  /// memory region is already registered, it will return an error message
+  /// with the error code 'TRTSERVER_ERROR_ALEADY_EXISTS'.
   /// \param name The user-given name for the shared memory region to be
   /// registered.
   /// \param shm_key The unique name of the location in shared memory being
@@ -785,7 +785,7 @@ class SharedMemoryControlContext {
       size_t byte_size) = 0;
 
   /// Unregister a registered shared memory region on the inference server. If
-  /// the shared memory region is not registered, do nothing and return failure.
+  /// the shared memory region is not registered, do nothing and return success.
   /// \param name The user-given name for the shared memory region to be
   /// unregistered.
   /// \return Error object indicating success or failure.
--- a/src/clients/c++/request.h
+++ b/src/clients/c++/request.h
@@ -747,8 +747,9 @@ class ModelControlContext {
 
 //==============================================================================
 /// A SharedMemoryControlContext object is used to control the registration /
-/// unregistration of shared memory regions on the inference server. Once
-/// created, a SharedMemoryControlContext object can be used repeatedly.
+/// unregistration and get the status of shared memory regions on the inference
+/// server. Once created, a SharedMemoryControlContext object can be used
+/// repeatedly.
 ///
 /// A SharedMemoryControlContext object can use either HTTP protocol or GRPC
 /// protocol depending on the Create function
@@ -763,6 +764,11 @@ class ModelControlContext {
 ///   ctx->RegisterSharedMemory(name, shm_key, 0, 104);
 ///   ...
 ///   ctx->UnregisterSharedMemory(name);
+///   ...
+///   ctx->UnregisterAllSharedMemory();
+///   ...
+///   std::vector<std::string> active_shm_regions;
+///   ctx->GetSharedMemoryStatus(active_shm_regions);
 ///   ...
 /// \endcode
 ///
@@ -790,6 +796,17 @@ class SharedMemoryControlContext {
   /// unregistered.
   /// \return Error object indicating success or failure.
   virtual Error UnregisterSharedMemory(const std::string& name) = 0;
+
+  /// Unregisters all registered shared memory regions on the inference server.
+  /// \return Error object indicating success or failure.
+  virtual Error UnregisterAllSharedMemory() = 0;
+
+  /// Get the list of names of all registered (active) shared memory regions on
+  /// the inference server. It is empty if there are none.
+  /// \param active_shm_regions The vector of names to be populated with the
+  /// list of active shared memory regions.
+  /// \return Error object indicating success or failure.
+  virtual Error GetSharedMemoryStatus() = 0;
 };
 
 //==============================================================================
--- a/src/clients/c++/request.h
+++ b/src/clients/c++/request.h
@@ -715,7 +715,7 @@ class ProfileContext {
 /// can be used repeatedly.
 ///
 /// A ModelControlContext object can use either HTTP protocol or GRPC protocol
-/// depending on the Create function (ControlHttpContext::Create or
+/// depending on the Create function (ModelControlHttpContext::Create or
 /// ModelControlGrpcContext::Create). For example:
 ///
 /// \code
@@ -743,6 +743,46 @@ class ModelControlContext {
   /// \param model_name The name of the model to be unloaded.
   /// \return Error object indicating success or failure.
   virtual Error Unload(const std::string& model_name) = 0;
+};
+
+//==============================================================================
+/// A SharedMemoryControlContext object is used to control the registration /
+/// unregistration of shared memory regions on the inference server. Once
+/// created, a SharedMemoryControlContext object can be used repeatedly.
+///
+/// A SharedMemoryControlContext object can use either HTTP protocol or GRPC
+/// protocol depending on the Create function
+/// (SharedMemoryControlHttpContext::Create or
+/// SharedMemoryControlGrpcContext::Create). For example:
+///
+/// \code
+///   std::unique_ptr<SharedMemoryControlContext> ctx;
+///   SharedMemoryControlGrpcContext::Create(&ctx, "localhost:8000");
+///   std::string name = "shared_memory";
+///   std::string shm_key = "/input";
+///   ctx->RegisterSharedMemory(name, shm_key, 0, 104);
+///   ...
+///   ctx->UnregisterSharedMemory(name);
+///   ...
+/// \endcode
+///
+class SharedMemoryControlContext {
+ public:
+  virtual ~SharedMemoryControlContext() = 0;
+
+  /// Register a model on the inference server. If the shared memory is already
+  /// loaded, it will be reloaded to update the LRU policy. \param name The
+  /// name of the shared memory region to be registered. \return Error object
+  /// indicating success or failure.
+  virtual Error RegisterSharedMemory(
+      const std::string& name, const std::string& shm_key, size_t offset,
+      size_t byte_size) = 0;
+
+  /// Unload a model from the inference server. unregistering a shared memory
+  /// region that is not registered on server has no affect and success code
+  /// will be returned. \param name The name of the shared memory region to
+  /// be unregistered. \return Error object indicating success or failure.
+  virtual Error UnregisterSharedMemory(const std::string& name) = 0;
 };
 
 //==============================================================================
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -129,6 +129,7 @@ set(
   metrics.cc
   model_config_utils.cc
   model_repository_manager.cc
+  shared_memory_manager.cc
   profile.cc
   provider.cc
   provider_utils.cc
@@ -154,6 +155,7 @@ set(
   metrics.h
   model_config_utils.h
   model_repository_manager.h
+  shared_memory_manager.h
   profile.h
   provider.h
   provider_utils.h
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -129,13 +129,13 @@ set(
   metrics.cc
   model_config_utils.cc
   model_repository_manager.cc
-  shared_memory_manager.cc
   profile.cc
   provider.cc
   provider_utils.cc
   sequence_batch_scheduler.cc
   server.cc
   server_status.cc
+  shared_memory_manager.cc
   status.cc
   trtserver.cc
 )
@@ -155,7 +155,6 @@ set(
   metrics.h
   model_config_utils.h
   model_repository_manager.h
-  shared_memory_manager.h
   profile.h
   provider.h
   provider_utils.h
@@ -163,6 +162,7 @@ set(
   sequence_batch_scheduler.h
   server.h
   server_status.h
+  shared_memory_manager.h
   status.h
   trtserver.h
 )
--- a/src/core/constants.h
+++ b/src/core/constants.h
@@ -38,6 +38,7 @@ constexpr char kStatusRESTEndpoint[] = "api/status";
 constexpr char kProfileRESTEndpoint[] = "api/profile";
 constexpr char kHealthRESTEndpoint[] = "api/health";
 constexpr char kModelControlRESTEndpoint[] = "api/modelcontrol";
+constexpr char kSharedMemoryControlRESTEndpoint[] = "api/shmcontrol";
 
 #ifdef TRTIS_ENABLE_TENSORFLOW
 constexpr char kTensorFlowGraphDefPlatform[] = "tensorflow_graphdef";
--- a/src/core/constants.h
+++ b/src/core/constants.h
@@ -38,7 +38,7 @@ constexpr char kStatusRESTEndpoint[] = "api/status";
 constexpr char kProfileRESTEndpoint[] = "api/profile";
 constexpr char kHealthRESTEndpoint[] = "api/health";
 constexpr char kModelControlRESTEndpoint[] = "api/modelcontrol";
-constexpr char kSharedMemoryControlRESTEndpoint[] = "api/shmcontrol";
+constexpr char kSharedMemoryControlRESTEndpoint[] = "api/sharedmemorycontrol";
 
 #ifdef TRTIS_ENABLE_TENSORFLOW
 constexpr char kTensorFlowGraphDefPlatform[] = "tensorflow_graphdef";
--- a/src/core/provider.cc
+++ b/src/core/provider.cc
@@ -117,6 +117,11 @@ InferRequestProvider::Create(
               (*provider)->model_name_ + "'");
     }
     (*provider)->input_buffer_[io.name()] = std::make_pair(it->second, 0);
+    // If using shared memory set shm_regions
+    if (false) {
+      // (*provider)->io_shm_regions_[io.name()] = std::make_tuple(shm_key_,
+      // offset_, byte_size_);
+    }
   }
 
   return Status::Success;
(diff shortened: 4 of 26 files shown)


src/clients/c++/request.h
```diff
@@ -745,6 +745,53 @@ class ModelControlContext {
...
+      const std::string& name, const std::string& shm_key, size_t offset,
+      size_t byte_size) = 0;
+
+  /// Unregister a registered shared memory region on the inference server. If
+  /// the shared memory region is not registered, do nothing and return failure.
+  /// \param name The user-given name for the shared memory region to be
+  /// unregistered.
+  /// \return Error object indicating success or failure.
+  virtual Error UnregisterSharedMemory(const std::string& name) = 0;
+};
```
deadeyegoodwin: UnregisterAll and GetSharedMemoryStatus not declared here. They should be done in separate changes but you likely want a ticket(s) to track them and update the comment in this pull request.

src/core/CMakeLists.txt
```diff
@@ -129,6 +129,7 @@ set(
   metrics.cc
   model_config_utils.cc
   model_repository_manager.cc
+  shared_memory_manager.cc
```
deadeyegoodwin: alpha order

src/core/CMakeLists.txt
```diff
@@ -154,6 +155,7 @@ set(
   metrics.h
   model_config_utils.h
   model_repository_manager.h
+  shared_memory_manager.h
```
deadeyegoodwin: alpha order

src/core/constants.h
```diff
@@ -38,6 +38,7 @@ constexpr char kStatusRESTEndpoint[] = "api/status";
 constexpr char kProfileRESTEndpoint[] = "api/profile";
 constexpr char kHealthRESTEndpoint[] = "api/health";
 constexpr char kModelControlRESTEndpoint[] = "api/modelcontrol";
+constexpr char kSharedMemoryControlRESTEndpoint[] = "api/shmcontrol";
```
deadeyegoodwin: "api/sharedmemorycontrol"

src/core/server.cc
```diff
@@ -145,6 +145,15 @@ InferenceServer::Init()
     return status;
   }
 
+  // Create the shared memory manager that registers / unregisters and returns
+  // the current registered shared memory regions.
+  status =
```
deadeyegoodwin: Do this before creating model repository manager (above)

src/core/server.cc
```diff
@@ -367,6 +376,49 @@ InferenceServer::UnloadModel(const std::string& model_name)
...
+
+  ScopedAtomicIncrement inflight(inflight_request_counter_);
+
+  auto action_type = SharedMemoryManager::ActionType::REGISTER;
+  return shared_memory_manager_->RegisterUnregisterSharedMemory(
+      name, shm_key, offset, byte_size, action_type);
+}
+
+Status
+InferenceServer::GetLiveSharedMemory(
```
deadeyegoodwin: GetSharedMemoryStatus()

src/core/server.cc
```diff
@@ -367,6 +376,49 @@ InferenceServer::UnloadModel(const std::string& model_name)
...
+    const size_t byte_size)
+{
+  if (ready_state_ != ServerReadyState::SERVER_READY) {
+    return Status(RequestStatusCode::UNAVAILABLE, "Server not ready");
+  }
+
+  ScopedAtomicIncrement inflight(inflight_request_counter_);
+
+  auto action_type = SharedMemoryManager::ActionType::REGISTER;
+  return shared_memory_manager_->RegisterUnregisterSharedMemory(
```
deadeyegoodwin: Can shared memory manager just have separate register and unregister functions?

src/core/server.cc
```diff
@@ -367,6 +376,49 @@ InferenceServer::UnloadModel(const std::string& model_name)
...
+  if (shared_memory_manager_ == nullptr) {
+    LOG_INFO << "No shared memory manager is available. Exiting immediately.";
+    return Status::Success;
+  }
+  return shared_memory_manager_->GetLiveSharedMemory(active_shm_regions);
+
+  return Status::Success;
+}
+
+Status
```
deadeyegoodwin: Move this up next to register function

src/core/server_status.cc
```diff
@@ -190,13 +190,21 @@ ServerStatusManager::UpdateServerStat(
       break;
     }
 
-    case ServerStatTimerScoped::Kind::CONTROL: {
+    case ServerStatTimerScoped::Kind::MODELCONTROL: {
```
deadeyegoodwin: MODEL_CONTROL

src/core/server_status.cc
```diff
@@ -190,13 +190,21 @@ ServerStatusManager::UpdateServerStat(
...
       StatDuration* d =
           server_status_.mutable_model_control_stats()->mutable_success();
       d->set_count(d->count() + 1);
       d->set_total_time_ns(d->total_time_ns() + duration);
       break;
     }
+
+    case ServerStatTimerScoped::Kind::SHMCONTROL: {
```
deadeyegoodwin: SHARED_MEMORY_CONTROL

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+#include <deque>
+#include <exception>
+#include <future>
+#include "src/core/constants.h"
+#include "src/core/logging.h"
+#include "src/core/server_status.h"
+
+namespace nvidia { namespace inferenceserver {
+
+Status
```
deadeyegoodwin: Put non-class functions in anonymous namespace

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+open_shm_region(const std::string& shm_key, int* shm_fd)
```
deadeyegoodwin: Use regular naming convention for these functions.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+    std::unique_ptr<SharedMemoryInfo> shm_info(new SharedMemoryInfo(
+        name, shm_key, offset, byte_size, shm_fd, tmp_addr));
+    shared_memory_map_.insert(std::make_pair(
+        name, std::unique_ptr<SharedMemoryInfo>(new SharedMemoryInfo(
+                  name, shm_key, offset, byte_size, shm_fd, tmp_addr))));
+  }
+  catch (std::exception& ex) {
+    return Status(
+        RequestStatusCode::INVALID_ARG,
+        "Unable to register shared memory region successfully.");
```
deadeyegoodwin: "failed to register shared memory region '" + name + "'"

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+
+Status
+SharedMemoryManager::UnregisterSharedMemory(const std::string& name)
+{
+  LOG_VERBOSE(1) << "Unregister() '" << name << "'";
+
+  auto it = shared_memory_map_.find(name);
+  if (it != shared_memory_map_.end()) {
+    void* tmp_addr;
+    RETURN_IF_ERROR(get_shm_addr(
```
deadeyegoodwin: Don't we have all the information we need from the map?

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+        &tmp_addr));
+    RETURN_IF_ERROR(mmap_cleanup(tmp_addr, it->second->byte_size_));
+
+    // if no other region with same shm_key then close
+    bool last_one = true;
+    for (auto itr = shared_memory_map_.begin(); itr != shared_memory_map_.end();
+         ++itr) {
+      if (itr->second->shm_key_ == it->second->shm_key_)
+        last_one = false;
+    }
```
deadeyegoodwin: Always branches around then-else.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+        it->second->shm_fd_, it->second->offset_, it->second->byte_size_,
+        &tmp_addr));
+    RETURN_IF_ERROR(mmap_cleanup(tmp_addr, it->second->byte_size_));
+
+    // if no other region with same shm_key then close
+    bool last_one = true;
+    for (auto itr = shared_memory_map_.begin(); itr != shared_memory_map_.end();
+         ++itr) {
+      if (itr->second->shm_key_ == it->second->shm_key_)
+        last_one = false;
```
deadeyegoodwin: break after setting last_one=false

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+Status
+SharedMemoryManager::RegisterSharedMemory(
+    const std::string& name, const std::string& shm_key, const size_t offset,
+    const size_t byte_size)
+{
+  LOG_VERBOSE(1) << "Register() shared memory region: '" << name << "'";
+  Status status = Status::Success;
+
+  // check if key is in shared_memory_map_ then remove
+  if (shared_memory_map_.find(name) != shared_memory_map_.end()) {
```
deadeyegoodwin: This can be dangerous if client makes an mistake are reuses name for 2 different regions. Instead we should return an error.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+  // Not setting the smart pointer directly to simplify clean up
+  std::unique_ptr<SharedMemoryManager> tmp_manager(
+      new SharedMemoryManager(status_manager));
+  *shared_memory_manager = std::move(tmp_manager);
+
+  return Status::Success;
+}
+
+Status
+SharedMemoryManager::RegisterUnregisterSharedMemory(
```
deadeyegoodwin: Remove, just have register and unregister functions.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  return Status::Success;
+}
+
+Status
+SharedMemoryManager::UnregisterAllSharedMemory()
+{
+  for (const auto& shm_info : shared_memory_map_) {
+    Status unregister_status = UnregisterSharedMemory(shm_info.first);
```
deadeyegoodwin: Should go through and unregister as many as possible and just collect the names of those that don't unregister and return a single error message listing all that didn't unregister... and just use the Message from one of them I guess.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+          "Failed to gracefully unregister all shared memory regions: " +
+              unregister_status.Message());
+    }
+  }
+  return Status::Success;
+}
+
+// Note: change to return entire SharedMemoryStateMap if needed
+Status
+SharedMemoryManager::GetLiveSharedMemory(
```
deadeyegoodwin: Need to discuss what to return as the status...

src/core/shared_memory_manager.h
```diff
@@ -0,0 +1,117 @@
...
+  /// Creates a SharedMemoryManager object that uses the given status_manager
+  static Status Create(
+      const std::shared_ptr<ServerStatusManager>& status_manager,
+      std::unique_ptr<SharedMemoryManager>* shared_memory_manager);
+
+ private:
+  SharedMemoryManager(
+      const std::shared_ptr<ServerStatusManager>& status_manager);
+
+  std::mutex shm_info_mu_;
```
deadeyegoodwin: Probably need to use a single mutex for everything.. to protect the map.

src/core/trtserver.cc
```diff
@@ -1017,13 +1017,42 @@ TRTSERVER_UnloadModel(TRTSERVER_Server* server, const char* model_name)
...
-      lserver->StatusManager(), ni::ServerStatTimerScoped::Kind::CONTROL);
+      lserver->StatusManager(), ni::ServerStatTimerScoped::Kind::MODELCONTROL);
 
   RETURN_IF_STATUS_ERROR(lserver->UnloadModel(std::string(model_name)));
 
   return nullptr;  // success
 }
 
+TRTSERVER_Error*
+TRTSERVER_RegisterSharedMemory(
```
deadeyegoodwin: This should be TRTSERVER_ServerRegisterSharedMemory to follow the naming convention. The model load/unload is also wrong, I'll ask @GuanLuo to fix that.

src/core/trtserver.h
```diff
@@ -646,6 +646,33 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_LoadModel(
 TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_UnloadModel(
     TRTSERVER_Server* server, const char* model_name);
 
+/// Register a shared memory region on the inference server. If the shared
+/// memory region is already registered, it will be unregistered and then
```
deadeyegoodwin: if already registered => error

src/core/trtserver.h
```diff
@@ -646,6 +646,33 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_LoadModel(
...
+/// registered.
+/// \param offset The offset into the shared memory region.
+/// \param byte_size The size, in bytes of the tensor data.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_RegisterSharedMemory(
+    TRTSERVER_Server* server, const char* name, const char* shm_key,
+    const size_t offset, const size_t byte_size);
+
+/// Unregister a registered shared memory region on the inference server. If
+/// the shared memory region is not registered, do nothing and return failure.
```
deadeyegoodwin: If not register don't return error, just silently ignore and return success. They can use status API if they want to know what is registered.

src/servers/main.cc
```diff
@@ -80,8 +80,8 @@ bool allow_http_ = true;
 int32_t http_port_ = 8000;
 int32_t http_health_port_ = -1;
 std::vector<int32_t> http_ports_;
-std::vector<std::string> endpoint_names = {"status", "health", "profile",
-                                           "infer", "modelcontrol"};
+std::vector<std::string> endpoint_names = {
+    "status", "health", "profile", "infer", "modelcontrol", "shmcontrol"};
```
deadeyegoodwin: sharedmemorycontrol

src/servers/main.cc
```diff
@@ -802,8 +802,8 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
 #ifdef TRTIS_ENABLE_HTTP
   http_port_ = http_port;
   http_health_port_ = http_health_port;
-  http_ports_ = {http_port_, http_health_port_, http_port_, http_port_,
-                 http_port_};
+  http_ports_ = {http_port_, http_health_port_, http_port_,
+                 http_port_, http_port_,        http_port_};
```
deadeyegoodwin: format

CoderHam: This was caused by the python format command. Should I remove the extra spacing even though our formatter would revert back to this every time it is run?

deadeyegoodwin: No, if the formatter does it then leave it... I didn't know it did that...

src/core/server.cc
CoderHam: Yes. There already are two separate functions. This is not necessary.

src/core/shared_memory_manager.cc
CoderHam: Yes we do. fixed this

src/clients/c++/request.h
CoderHam: Created a new ticket

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+         ++itr) {
+      if (itr->second->shm_key_ == shm_key)
+        shm_fd = itr->second->shm_fd_;
+      else {
+        RETURN_IF_ERROR(OpenSharedMemoryRegion(shm_key, &shm_fd));
+      }
+    }
+
+    RETURN_IF_ERROR(
+        GetSharedMemoryAddress(shm_fd, offset, byte_size, &tmp_addr));
```
deadeyegoodwin: How about MapSharedMemory? Can you use something more descriptive than tmp_addr. Perhaps mapped_addr.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+    shared_memory_map_.insert(std::make_pair(
+        name, std::unique_ptr<SharedMemoryInfo>(new SharedMemoryInfo(
+                  name, shm_key, offset, byte_size, shm_fd, tmp_addr))));
+  }
+  catch (std::exception& ex) {
```
deadeyegoodwin: What can throw an exception?

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+SharedMemoryManager::UnregisterSharedMemory(const std::string& name)
+{
+  LOG_VERBOSE(1) << "Unregister() shared memory region: '" << name << "'";
+
+  // Serialize all operations that write/read current shared memory regions
+  std::lock_guard<std::mutex> lock(register_mu_);
+
+  auto it = shared_memory_map_.find(name);
+  if (it != shared_memory_map_.end()) {
+    RETURN_IF_ERROR(MmapCleanup(it->second->shm_addr_, it->second->byte_size_));
```
deadeyegoodwin: UnmapSharedMemory?

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+
+    // if no other region with same shm_key then close
```
deadeyegoodwin: You need to remove the region from the map before this or you will always find it

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+  // Not setting the smart pointer directly to simplify clean up
+  std::unique_ptr<SharedMemoryManager> tmp_manager(
+      new SharedMemoryManager(status_manager));
+  *shared_memory_manager = std::move(tmp_manager);
+
+  return Status::Success;
+}
+
+Status
+SharedMemoryManager::UnregisterAllSharedMemory()
```
deadeyegoodwin: Group this with unregister function above.

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+  // Serialize all operations that write/read current shared memory regions
+  std::lock_guard<std::mutex> lock(register_mu_);
+
+  std::string error_message =
+      "Failed to unregister the following shared memory regions: ";
+  std::vector<std::string> unregister_fails;
+  for (const auto& shm_info : shared_memory_map_) {
+    Status unregister_status = UnregisterSharedMemory(shm_info.first);
+    if (!unregister_status.IsOk()) {
+      unregister_fails.push_back(shm_info.first);
```
deadeyegoodwin: Since we aren't returning the actual 'unregister_status' error message you should log it with LOG_ERROR

src/core/shared_memory_manager.cc
```diff
@@ -0,0 +1,244 @@
...
+    for (auto unreg_fail : unregister_fails) {
+      error_message += unreg_fail + " ,";
+    }
+    return Status(RequestStatusCode::INTERNAL, error_message);
+  }
+
+  return Status::Success;
+}
+
+// Note: change to return entire SharedMemoryStateMap if needed
```
deadeyegoodwin: Not in this change

src/core/shared_memory_manager.h
```diff
@@ -0,0 +1,116 @@
...
+    size_t offset_;
+    size_t byte_size_;
+    int shm_fd_;
+    void* shm_addr_;
+  };
+
+  using SharedMemoryStateMap =
+      std::map<std::string, std::unique_ptr<SharedMemoryInfo>>;
+
+  enum ActionType { NO_ACTION, REGISTER, UNREGISTER };
```
deadeyegoodwin: Don't need

src/core/shared_memory_manager.cc
CoderHam: I believe that all cases for exception are covered by my earlier checks

src/core/server.cc
```diff
@@ -367,6 +376,46 @@ InferenceServer::UnloadModel(const std::string& model_name)
...
+    return Status(RequestStatusCode::UNAVAILABLE, "Server not ready");
+  }
+
+  ScopedAtomicIncrement inflight(inflight_request_counter_);
+
+  return shared_memory_manager_->UnregisterSharedMemory(name);
+}
+
+Status
+InferenceServer::GetSharedMemoryStatus(
```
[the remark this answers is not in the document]
CoderHam: Are you asking me to move GetSharedMemoryStatus and UnregisterAll into a different PR?

src/core/shared_memory_manager.cc
CoderHam: Returning string of shared memory region names. Will later modify to return protobuf form response like for server status

src/core/provider.cc
```diff
@@ -176,39 +181,46 @@ InferRequestProvider::GetNextInputContent(
       return Status(
           RequestStatusCode::INTERNAL, "unexpected input '" + name + "'");
     }
-
```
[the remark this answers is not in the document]
CoderHam: In the provider the raw buffer if fetched from the input_buffer content. We'll need a check here to read from the shared memory buffer instead of from the input_buffer here. Similarly for the raw output.

src/core/server.cc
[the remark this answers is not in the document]
CoderHam: Added it back to this PR

src/core/provider.cc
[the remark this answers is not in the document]
CoderHam: Fixed shared memory input for grpc. Fixing for http as well.

src/core/trtserver.cc
```diff
@@ -610,6 +612,35 @@ TRTSERVER_InferenceRequestProviderSetInputData(
   return nullptr;  // Success
 }
 
+TRTSERVER_Error*
+TRTSERVER_InferenceRequestProviderSetSharedMemoryInputData(
```
deadeyegoodwin: We don't need this function. Just use TRTSERVER_InferenceRequestProviderSetInputData along with TRTSERVER_ServerGetSharedMemoryInfo.

src/core/trtserver.cc
```diff
@@ -1017,13 +1048,91 @@ TRTSERVER_ServerUnloadModel(TRTSERVER_Server* server, const char* model_name)
...
+      lserver->StatusManager(),
+      ni::ServerStatTimerScoped::Kind::SHARED_MEMORY_CONTROL);
+
+  RETURN_IF_STATUS_ERROR(lserver->GetSharedMemoryStatus(active_shm_regions));
+
+  return nullptr;  // success
+}
+
+TRTSERVER_Error*
+TRTSERVER_ServerGetSharedMemoryInfo(
```
deadeyegoodwin: This needs a different name. It should just return an address.

src/servers/grpc_server.cc
```diff
@@ -311,24 +391,28 @@ class InferBaseContext : public BaseContext<LifeCycle, AsyncResources> {
       if (err == nullptr) {
         err = GRPCToInput(request.meta_data(), request, request_provider);
```
deadeyegoodwin: GRPCToInput should handle the shared memory case in addition to the existing raw case.

src/core/trtserver.cc
CoderHam: Renamed to GetSharedMemoryAddress and returning calculated `mapped_address + offset`

src/core/trtserver.cc
CoderHam: TRTSERVER_InferenceRequestProviderSetInputData and TRTSERVER_InferenceRequestProviderSetSharedMemoryInputData require significantly different arguments. Should I still merge them?

src/servers/grpc_server.cc

src/core/trtserver.cc
CoderHam: Merged. Can undo if needed.

PR #504, opened 2019-07-26
Add zero-prefixed version test in L0_lifecycle

GuanLuo:


GuanLuo:
@jobegrabber I modified the code to log and ignore version directories that has zero-prefixed, so that the server will try to load the model as long as there is at least one well-formed version directory.

src/core/model_repository_manager.cc
```diff
@@ -1565,17 +1565,13 @@ ModelRepositoryManager::VersionsToLoad(
   RETURN_IF_ERROR(GetDirectorySubdirs(model_path, &subdirs));
   std::set<int64_t, std::greater<int64_t>> existing_versions;
   for (const auto& subdir : subdirs) {
+    if ((subdir.length() > 1) && (subdir.front() == '0')) {
+      LOG_WARNING << "ignore version directory '" << subdir
```
deadeyegoodwin: Need to update the documentation as well to mention this behavior.

PR #512, opened 2019-07-30
Enhance model control to try loading model dependencies for ensemble

GuanLuo:


src/core/model_repository_manager.h
```diff
@@ -165,6 +165,24 @@ class ModelRepositoryManager {
...
+  /// \param models The set of models to be polled
+  /// \param added The names of the models added to the repository.
+  /// \param deleted The names of the models removed from the repository.
+  /// \param modified The names of the models remaining in the
+  /// repository that have been changed.
+  /// \param unmodified The names of the models remaining in the
+  /// repository that have not changed.
+  /// \param updated_infos The model infos retrieved from the poll.
+  /// \return The error status.
+  Status PollModels(
```
deadeyegoodwin: The different between Poll and PollModels is not clear. Perhaps the functions just need better names, or perhaps there is some refactoring to avoid such similar functionality?

PR #526, opened 2019-08-05
Enhance ensemble scheduler to reshape tensors and to adjust batch size between batchable and non-batchable models

GuanLuo:


deadeyegoodwin:
@GuanLuo Did you try this on Boris' case?

GuanLuo:
No, will try it now.

PR #533, opened 2019-08-08
Add S3 filesystem support

xprotobeast2:


Dockerfile
```diff
@@ -283,6 +284,8 @@ RUN LIBCUDA_FOUND=$(ldconfig -p | grep -v compat | awk '{print $1}' | grep libcu
             cp -r trtis/install/bin /opt/tensorrtserver/. && \
             cp -r trtis/install/lib /opt/tensorrtserver/. && \
             cp -r trtis/install/include /opt/tensorrtserver/include/trtserver) && \
+            # Copy aws-sdk-cpp libs to opt/tensorrtserver/lib
```
deadeyegoodwin: Don't need this comment

build/CMakeLists.txt
```diff
@@ -355,6 +373,15 @@ if(${TRTIS_ENABLE_METRICS})
   set(TRTIS_DEPENDS ${TRTIS_DEPENDS} prometheus-cpp)
 endif() # TRTIS_ENABLE_METRICS
 
+if(${TRTIS_ENABLE_S3})
```
deadeyegoodwin: Put this at line 365 just after GCS (and follow style of no blank lines before/after)

build/CMakeLists.txt
```diff
@@ -355,6 +373,15 @@ if(${TRTIS_ENABLE_METRICS})
...
+  set(TRTIS_DEPENDS ${TRTIS_DEPENDS} aws-sdk-cpp)
+endif() # TRTIS_ENABLE_S3
+
+set(CMAKE_PREFIX_PATH ${CMAKE_PREFIX_PATH}
```
deadeyegoodwin: Is this really necessary. Why doesn't AWSSDK_DIR take care of this?

qa/L0_storage_S3/test.sh
```diff
@@ -0,0 +1,179 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+export CUDA_VISIBLE_DEVICES=0
+
+CLIENT_LOG_BASE="./client"
+INFER_TEST=infer_test.py
+
+# Google cloud variables (Point to bucket when testing cloud storage)
```
deadeyegoodwin: Not google

src/core/filesystem.cc
```diff
@@ -541,6 +837,23 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
 #endif  // TRTIS_ENABLE_GCS
   }
 
+  // Check if this is a GCS path (gs://$BUCKET_NAME)
```
deadeyegoodwin: Comment is wrong

src/core/filesystem.cc
```diff
@@ -541,6 +837,23 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
+    Aws::SDKOptions options;
+    Aws::InitAPI(options);
+    static S3FileSystem s3_fs(options);
+    // RETURN_IF_ERROR(s3_fs.CheckClient());
+    *file_system = &s3_fs;
+    return Status::Success;
+#endif  // TRTIS_ENABLE_S3
+  }
+
   // For now assume all paths are local...
```
deadeyegoodwin: Change this comment: "Assume path is for local filesystem"

src/core/filesystem.cc
```diff
@@ -497,31 +514,310 @@ Status
...
 
+Status
+S3FileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
+{
+  // We don't need to worry about the case when this is a GCS directory
+  bool is_dir;
+
+  RETURN_IF_ERROR(IsDirectory(path, &is_dir));
+  if (is_dir) {
+    return Status::Success;
```
GuanLuo: Is it possible to get the modification time for directory? In model repository manager, I used the modification time of the directory for the case of file deletion ([here](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/core/model_repository_manager.cc#L194)). If the way I am doing only make sense on local filesystem, then I will need to use a different approach

src/core/filesystem.cc
```diff
@@ -497,31 +514,310 @@ Status
 GCSFileSystem::WriteTextFile(
     const std::string& path, const std::string& contents)
 {
+  return Status(
+      RequestStatusCode::INTERNAL,
+      "Write text file operation not yet implemented " + path);
```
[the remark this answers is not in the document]
xprotobeast2: We never use this function anywhere, and so we'd have to write it and test it for no specific purpose.

src/core/filesystem.cc
xprotobeast2: I can define it as the most recently modified child in the directory, and return that.

xprotobeast2: Seems like these cases should be added to the CI somehow.

PR #534, opened 2019-08-08
Tool for performance comparison across releases

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo I am going to generate some baseline results from older releases so that we can compare against them. It requires some manual work since perf_client didn't support all the flags we used and some backends were not available.

PR #537, opened 2019-08-09
Support for output tensors in shared memory

CoderHam:
Support for Output Tensors in Shared memory


src/clients/c++/request_http.cc
```diff
@@ -1598,6 +1598,22 @@ InferHttpContextImpl::PreRunProcessing(std::shared_ptr<Request>& request)
     }
   }
 
+  for (const auto& io : outputs_) {
+    if (reinterpret_cast<OutputImpl*>(io.get())->IsSharedMemory()) {
+      for (int i=0; i<infer_request_.output_size(); i++) {
```
deadeyegoodwin: You need to run the formatter

src/backends/tensorflow/base_backend.cc
```diff
@@ -449,6 +449,7 @@ ReadFixedSizedOutputTensor(
       void* content = nullptr;
       Status status = payload.response_provider_->AllocateOutputBuffer(
           output_name, &content, expected_byte_size, shape);
+      LOG_VERBOSE(1) << "Allocated Output Buffer";
```
deadeyegoodwin: Either add more information to this log, like tensor name, output buffer address and size.
Or remove this line

src/clients/c++/request_common.h
```diff
@@ -155,7 +155,7 @@ class InputImpl : public InferContext::Input {
   bool IsSharedMemory() const { return (io_type_ == SHARED_MEMORY); }
   const std::string& GetSharedMemoryName() const { return shm_name_; }
   size_t GetSharedMemoryOffset() const { return shm_offset_; }
-  size_t GetSharedMemoryByteSize() const { return byte_size_; }
+  size_t GetSharedMemoryByteSize() const { return shm_byte_size_; }
```
deadeyegoodwin: This looks like a bug in the input shared memory...
But we shouldn't really have shm_byte_size_ at all... instead in SetSharedMemory we just need to validate that the shared memory size == byte_size_ like is done in SetRaw.
You should put this fix in a different PR so we can get it submitted ASAP since it is an important change.

src/clients/c++/request_http.cc
```diff
@@ -1598,6 +1598,22 @@ InferHttpContextImpl::PreRunProcessing(std::shared_ptr<Request>& request)
     }
   }
 
+  for (const auto& io : outputs_) {
```
deadeyegoodwin: Output::SetSharedMemory needs to move from Output to Options::AddSharedMemoryResult. This is consistent with how we handle other results. Then this code to initialize infer_request_.outputs should be put in SetRunOptions.

src/clients/c++/simple_shm_client.cc
```diff
@@ -261,22 +261,6 @@ main(int argc, char** argv)
     }
   }
 
-  // Set the context options to do batch-size 1 requests. Also request that
```
deadeyegoodwin: Need to remove -I and -O option from simple_shm_client. Make this example as simple as possible

src/servers/http_server.cc
```diff
@@ -726,6 +726,21 @@ HTTPAPIServer::EVBufferToInput(
             .c_str());
   }
 
+  // Initialize System Memory for Output if it uses shared memory
```
deadeyegoodwin: I think the changes to ensembler, provider or trtserver can be much simplied... we should discuss.

src/backends/tensorflow/base_backend.cc
CoderHam: Added for debugging. Removing.

src/clients/c++/simple_shm_client.cc
CoderHam: Will remove once the output is functional before finalizing this PR

src/servers/http_server.cc
CoderHam: Sure

src/clients/c++/request_common.h
CoderHam: Will do this ASAP

CoderHam: Resolved by #541

src/clients/c++/request_http.cc
CoderHam: Done

src/clients/c++/request.h
```diff
@@ -401,7 +380,11 @@ class InferContext {
       /// CLASS format is the top-k highest probability values of the
       /// result and the associated class label (if provided by the
       /// model).
-      CLASS = 1
+      CLASS = 1,
+
+      /// SHARED_MEMORY format, the entire result tensor of values are written
```
deadeyegoodwin: I don't think we need this. The format of the result is still "RAW" even if it is in shared memory.

src/clients/c++/request.h
```diff
@@ -553,6 +536,22 @@ class InferContext {
...
     virtual Error AddClassResult(
         const std::shared_ptr<InferContext::Output>& output, uint64_t k) = 0;
+
+    /// Indicate that the result values for this output should be placed in a
+    /// shared memory region instead of being returned in the inference
+    /// response. The shared memory region must not be modified or destroyed
+    //  until this output is ready (that is until after the Run() call(s) have
+    /// written the output completely). For batched outputs, all tensor values
+    /// are copied into a contiguous space in a single shared memory region.
+    /// \param name The user-given name for the registered shared memory region
```
deadeyegoodwin: Need to add "output" param

src/clients/c++/request.h
```diff
@@ -553,6 +536,22 @@ class InferContext {
...
+    /// where the tensor values for this output should be stored.
+    /// \param offset The offset into the shared memory region upto the start
+    /// of the output tensor values.
+    /// \param byte_size The size, in bytes of the output tensor data.
+    /// Must match the size expected by the output.
+    /// \return Error object indicating success or failure.
+    virtual Error AddSharedMemoryResult(
```
deadeyegoodwin: Run formatter

src/clients/c++/request_common.cc
```diff
@@ -909,6 +895,12 @@ InferContextImpl::SetRunOptions(const InferContext::Options& boptions)
 
     auto routput = infer_request_.add_output();
     routput->set_name(output->Name());
+    if (ooptions.result_format == Result::ResultFormat::SHARED_MEMORY) {
```
deadeyegoodwin: Use some other way to detect that you need to add these shared memory values.

src/clients/c++/request_common.cc
```diff
@@ -114,6 +114,15 @@ OptionsImpl::AddClassResult(
   return Error::Success;
 }
 
+Error
+OptionsImpl::AddSharedMemoryResult(const std::shared_ptr<InferContext::Output>& output, const std::string& name, size_t offset, size_t byte_size)
+{
+  outputs_.emplace_back(std::make_pair(
+      output, OutputOptions(InferContext::Result::ResultFormat::SHARED_MEMORY, 0, name, offset, byte_size)));
```
deadeyegoodwin: Use RAW

src/clients/c++/request_common.h
```diff
@@ -219,16 +222,15 @@ class InputImpl : public InferContext::Input {
...
+      : mio_(mio), result_format_(InferContext::Result::ResultFormat::RAW)
   {
   }
   ~OutputImpl() = default;
 
   const std::string& Name() const override { return mio_.name(); }
   DataType DType() const override { return mio_.data_type(); }
   const DimsList& Dims() const override { return mio_.dims(); }
-  bool IsSharedMemory() const { return (io_type_ == SHARED_MEMORY); }
+  bool IsSharedMemory() const { return (result_format_ == InferContext::Result::ResultFormat::SHARED_MEMORY); }
```
deadeyegoodwin: Remove this

src/clients/c++/request_common.cc
CoderHam: Will do

src/clients/c++/request_common.cc
```diff
@@ -425,10 +411,10 @@ ResultImpl::GetRawShape(std::vector<int64_t>* shape) const
 {
   shape->clear();
 
-  if (result_format_ != InferContext::Result::ResultFormat::RAW) {
```
deadeyegoodwin: Now that using RAW format can revert this change

src/clients/c++/request_common.cc
```diff
@@ -438,11 +424,10 @@ ResultImpl::GetRawShape(std::vector<int64_t>* shape) const
 Error
 ResultImpl::GetRaw(size_t batch_idx, const std::vector<uint8_t>** buf) const
 {
-  if (result_format_ != InferContext::Result::ResultFormat::RAW) {
```
deadeyegoodwin: Now that using RAW format can revert this change

src/clients/c++/request_common.cc
```diff
@@ -469,11 +454,10 @@ Error
 ResultImpl::GetRaw(
     size_t batch_idx, const uint8_t** buf, size_t* byte_size) const
 {
-  if (result_format_ != InferContext::Result::ResultFormat::RAW) {
+  if (result_format_ == InferContext::Result::ResultFormat::CLASS) {
```
deadeyegoodwin: revert

src/clients/c++/request_common.cc
```diff
@@ -499,11 +483,10 @@ Error
 ResultImpl::GetRawAtCursor(
     size_t batch_idx, const uint8_t** buf, size_t adv_byte_size)
 {
-  if (result_format_ != InferContext::Result::ResultFormat::RAW) {
```
deadeyegoodwin: revert

src/clients/c++/request_common.cc
```diff
@@ -916,6 +899,12 @@ InferContextImpl::SetRunOptions(const InferContext::Options& boptions)
 
     auto routput = infer_request_.add_output();
     routput->set_name(output->Name());
+    if (ooptions.shm_name != "") {
```
deadeyegoodwin: !shm_name.empty()

src/servers/grpc_server.cc
```diff
@@ -133,7 +133,12 @@ AsyncResources::ResponseAlloc(
     const char* tensor_name, size_t byte_size,
     TRTSERVER_Memory_Type memory_type, int64_t memory_type_id, void* userp)
 {
-  InferResponse* response = reinterpret_cast<InferResponse*>(userp);
+  auto userp_pair = reinterpret_cast<std::pair<
+      InferResponse*,
+      std::unordered_map<std::string, std::pair<const char*, size_t>>>*>(userp);
+  InferResponse* response = reinterpret_cast<InferResponse*>(userp_pair->first);
+  std::unordered_map<std::string, std::pair<const char*, size_t>>
```
deadeyegoodwin: const&, you are making a copy of the map with this statement.

src/servers/grpc_server.cc
```diff
@@ -150,9 +155,15 @@ AsyncResources::ResponseAlloc(
   // that the number and order of raw output entries equals the output
   // meta-data.
   std::string* raw_output = response->add_raw_output();
-  if (byte_size > 0) {
-    raw_output->resize(byte_size);
-    *buffer = static_cast<void*>(&((*raw_output)[0]));
+  auto pr = output_shm_map.find(tensor_name);
+  if (pr != output_shm_map.end()) {
+    *buffer =
```
deadeyegoodwin: Why is it stored as const char in the map?

deadeyegoodwin: You should check to make sure that the requested byte size matches what is provided by the shared memory.

src/servers/http_server.cc
```diff
@@ -308,42 +321,49 @@ HTTPAPIServer::ResponseAlloc(
...
-          TRTSERVER_ERROR_INTERNAL,
-          std::string(
-              "reserved " + std::to_string(output_iovec.iov_len) +
-              " bytes in output tensor buffer, need " +
-              std::to_string(byte_size))
-              .c_str());
-    }
+  auto pr = output_shm_map.find(tensor_name);
+  if (pr != output_shm_map.end()) {
+    *buffer =
```
deadeyegoodwin: Check size...

src/core/trtserver.h
```diff
@@ -352,6 +352,11 @@ TRTSERVER_InferenceRequestProviderSetInputData(
     TRTSERVER_InferenceRequestProvider* request_provider, const char* name,
     const void* base, size_t byte_size);
 
+TRTSERVER_EXPORT TRTSERVER_Error*
```
deadeyegoodwin: Remove this change

src/servers/grpc_server.cc
CoderHam: Fixed

src/servers/http_server.cc
CoderHam: Done

src/clients/c++/request.h
```diff
@@ -337,8 +337,10 @@ class InferContext {
     /// completed. This function must be called a single time for an input that
     /// is using shared memory. For batched inputs, the tensor values for the
     /// entire batch must be contiguous in a single shared memory region.
-    /// \param name The user-given name for the registered shared memory
-    /// region where the tensor values for this input is stored.
+    /// Note: The options should be set using SetRunOptions before calling the
```
deadeyegoodwin: "should be set" -> "must be set"

src/clients/c++/request_http.cc
```diff
@@ -1127,8 +1127,9 @@ HttpRequestImpl::SetNextRawResult(
       }
     }
 
+    // std::cerr << "Output size - ob: " << size << " - " << ob << '\n';
```
deadeyegoodwin: Remove

src/clients/c++/request_http.cc
```diff
@@ -1127,8 +1127,9 @@ HttpRequestImpl::SetNextRawResult(
...
     // If output couldn't accept any more bytes then move to the next.
-    if (ob == 0) {
+    if (ob == 0 || io->UsesSharedMemory()) {
```
deadeyegoodwin: This seems like a strange place to put this check. Do you find that 'ob' !=0 for a shared-memory result? That would mean that some raw bytes were read for the result, and that shouldn't happen.

src/clients/c++/request_http.cc
```diff
@@ -1159,9 +1160,15 @@ HttpRequestImpl::CreateResult(
   }
 
   std::unique_ptr<ResultImpl> result(new ResultImpl(infer_output, batch_size));
-  result->SetBatch1Shape(output.raw().dims());
-  if (IsFixedSizeDataType(infer_output->DType())) {
-    result->SetBatchnByteSize(output.raw().batch_byte_size());
+
+  if (!ctx.UsesSharedMemory(output.name())) {
+    result->SetBatch1Shape(output.raw().dims());
```
deadeyegoodwin: We need to set the shape (and batch-n size??) even if using shared memory... why can't we? How will user know the shape of an variable-sized output if we don't set this?

src/clients/c++/request_grpc.cc
```diff
@@ -560,34 +560,36 @@ GrpcRequestImpl::InitResult(
...
-    result->SetBatchnByteSize(output.raw().batch_byte_size());
-  }
-
-  if (result->ResultFormat() == InferContext::Result::ResultFormat::RAW) {
-    if (grpc_response_->raw_output_size() <= (int)idx) {
-      return Error(
-          RequestStatusCode::INVALID,
-          "Expected RAW output for result '" + output.name() + "'");
+  if (!result->UsesSharedMemory()) {
+    result->SetBatch1Shape(output.raw().dims());
```
deadeyegoodwin: Still need shape for shared memory...

src/servers/grpc_server.cc
```diff
@@ -387,6 +407,8 @@ class InferBaseContext : public BaseContext<LifeCycle, AsyncResources> {
           grpc_infer_request->exec_ctx_);
     }
 
+    void* response_pair_;
```
deadeyegoodwin: How does this ever get deleted? Can you wrap in a unique_ptr?

src/servers/http_server.cc
```diff
@@ -361,6 +392,10 @@ HTTPAPIServer::ResponseRelease(
   LOG_VERBOSE(1) << "HTTP release: "
                  << "size " << byte_size << ", addr " << buffer;
 
+  // hack to delete evbuffer pair structure during release
+  auto response_pair = reinterpret_cast<evbuffer_pair*>(buffer_userp);
+  delete response_pair;
```
deadeyegoodwin: ResponseRelease will be called for each output buffer... so I don't think this is correct. You will get double-free errors I believe.

src/clients/c++/request_http.cc
CoderHam: The result size is used for extracting the request header. For the shared memory case although the result has a byte size the same is not used in calculation of the header. Moving this check to SetNextRawResult()

src/clients/c++/request_http.cc
CoderHam: Fixed

src/clients/c++/request_grpc.cc
CoderHam: Fixed

src/clients/c++/request_common.cc
```diff
@@ -432,6 +419,13 @@ ResultImpl::GetRawShape(std::vector<int64_t>* shape) const
 {
   shape->clear();
 
+  if (use_shm_) {
```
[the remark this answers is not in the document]
CoderHam: Done

src/servers/grpc_server.cc
```diff
@@ -151,9 +156,25 @@ AsyncResources::ResponseAlloc(
   // that the number and order of raw output entries equals the output
   // meta-data.
   std::string* raw_output = response->add_raw_output();
-  if (byte_size > 0) {
```
deadeyegoodwin: When byte_size is 0 you shouldn't do anything with shared memory. Keep the (byte_size > 0) as condition around all logic.

src/servers/http_server.cc
```diff
@@ -296,54 +304,76 @@ HTTPAPIServer::ResponseAlloc(
...
+  auto userp_pair = reinterpret_cast<EVBufferPair*>(userp);
+  evbuffer* evhttp_buffer = reinterpret_cast<evbuffer*>(userp_pair->first);
+  const std::unordered_map<std::string, std::pair<const void*, size_t>>&
+      output_shm_map = userp_pair->second;
 
   *buffer = nullptr;
   *buffer_userp = nullptr;
 
-  // Don't need to do anything if no memory was requested.
-  if (byte_size > 0) {
```
deadeyegoodwin: Keep (byte_size > 0) as guard around all logic.

src/servers/http_server.cc
```diff
@@ -296,54 +304,76 @@ HTTPAPIServer::ResponseAlloc(
...
-    if (output_iovec.iov_len < byte_size) {
-      return TRTSERVER_ErrorNew(
-          TRTSERVER_ERROR_INTERNAL,
-          std::string(
-              "reserved " + std::to_string(output_iovec.iov_len) +
-              " bytes in output tensor buffer, need " +
-              std::to_string(byte_size))
-              .c_str());
+
+  if (!output_shm_map.empty()) {
```
deadeyegoodwin: This logic isn't right. If some output is in map, but not 'tensor_name' then we want to use existing evbuffer logic for the output. Just use find() and only if you find the output do you use shared memory.

CoderHam: Fixed

src/servers/http_server.cc
CoderHam: Fixed

src/servers/grpc_server.cc
CoderHam: Fixed

PR #540, opened 2019-08-09
Support models with variable-sized inputs in perf_client

GuanLuo:


src/clients/c++/concurrency_manager.cc
```diff
@@ -49,28 +49,62 @@ nic::Error
...
+      if (it != input_shapes.end()) {
+        bool valid = true;
+
+        const auto& dims = it->second;
+        const auto& config_dims = input->Dims();
+        if (((int64_t)dims.size()) != config_dims.size()) {
+          valid = false;
+        } else {
+          for (size_t idx = 0; idx < dims.size(); idx++) {
+            if ((dims[idx] <= 0) ||
```
deadeyegoodwin: Seems like it would be better to check that user-specified dims are all positive when we process the arg

src/clients/c++/concurrency_manager.cc
```diff
@@ -49,28 +49,62 @@ nic::Error
...
+                ((dims[idx] != config_dims[idx]) && (config_dims[idx] != -1))) {
+              valid = false;
+              break;
+            }
+          }
+        }
+        if (!valid) {
```
deadeyegoodwin: Don't need the valid flag. Just put the return in the loop. Also would be better error mesage to print the use-supplied and model shapes so user can clearly see what they did wrong.

src/clients/c++/concurrency_manager.cc
```diff
@@ -49,28 +49,62 @@ nic::Error
...
+                  "' does not match model config shape");
+        }
+      }
+    }
+    local_manager->input_shapes_ = input_shapes;
+  }
+
   // Read provided data
   if (!data_directory.empty()) {
     // Get all input names
```
deadeyegoodwin: This comment isn't very accurate or useful (same above)

src/clients/c++/concurrency_manager.cc
```diff
@@ -49,28 +49,62 @@ nic::Error
...
+    // Get all input names
+    for (const auto& input : ctx->Inputs()) {
+      auto it = input_shapes.find(input->Name());
+      if (it != input_shapes.end()) {
+        bool valid = true;
+
+        const auto& dims = it->second;
+        const auto& config_dims = input->Dims();
+        if (((int64_t)dims.size()) != config_dims.size()) {
+          valid = false;
```
deadeyegoodwin: Just return an appropriate error here about number of dimensions not matching (and give the user-supplied and expected shapes)

src/clients/c++/perf_client.cc
```diff
@@ -342,6 +343,11 @@ Usage(char** argv, const std::string& msg = std::string())
       << " of latency will also be reported and used to detemine if the"
       << " measurement is stable instead of average latency."
       << " Default is -1 to indicate no percentile will be used." << std::endl;
+  std::cerr << "For --shape, it set the shape used for the specified input. The"
```
deadeyegoodwin: --shape, the shape...

src/clients/c++/perf_client.cc
```diff
@@ -342,6 +343,11 @@ Usage(char** argv, const std::string& msg = std::string())
...
+               " argument must be specified as 'name:shape' where the shape is"
```
deadeyegoodwin: .. where the shape is a comma-separate list of dimension sizes, for example --shape input:1,2,3.

src/clients/c++/perf_client.cc
```diff
@@ -407,6 +417,31 @@ main(int argc, char** argv)
...
+              pos = comma_pos;
+            } else {
+              shape.emplace_back(
+                  std::stoll(shape_str.substr(pos, comma_pos - pos)));
+              pos = comma_pos + 1;
+            }
+          }
+        }
+        catch (const std::invalid_argument& ia) {
+          Usage(argv, "failed to parse input shape");
```
deadeyegoodwin: .. shape: " + optarg

PR #541, opened 2019-08-09
Fix bug for shared memory byte size in C++ Client API

CoderHam:


src/clients/c++/request_common.cc
```diff
@@ -244,7 +243,16 @@ InputImpl::SetSharedMemory(
 
   shm_name_ = name;
   shm_offset_ = offset;
-  shm_byte_size_ = byte_size;
+
+  // verify byte size of shared memory matches that of expected byte size
```
deadeyegoodwin: Put this error check at line 243 so that it is next to the other checks.

CoderHam: Done

PR #548, opened 2019-08-12
Add "--model-repository" option. Change terminology used to "model repository"

GuanLuo:


src/servers/main.cc
```diff
@@ -179,7 +179,11 @@ std::vector<Option> options_{
     {OPTION_LOG_WARNING, "log-warning", "Enable/disable warning-level logging"},
     {OPTION_LOG_ERROR, "log-error", "Enable/disable error-level logging"},
     {OPTION_ID, "id", "Identifier for this server"},
-    {OPTION_MODEL_STORE, "model-store", "Path to model repository directory"},
+    {OPTION_MODEL_REPOSITORY, "model-store",
+     "Path to model repository directory. This option is deprecated, the "
+     "prefer usage is --model-repository"},
```
CoderHam: prefer -> preferred

CoderHam:
You should add these changes to the docs and the QA tests as well as `--model-store` will now deprecated.

deadeyegoodwin:
> You should add these changes to the docs and the QA tests as well as `--model-store` will now deprecated.

Yes, should update docs/ directory. For CI should change most tests to use --model-repository but should leave at least 1 using --model-store.

Issue #544, opened 2019-08-10
Shared Memory client fails for batch size != 1


philipp-schmidt:
Requests with batch size 1 and correct input_byte_size calculation work correctly.

Increasing batch size to two and multiply input_byte_size by two leads to the following exception (modified version of simple_shm_client.cc):

```
int batch_size = 2;
options->SetBatchSize(batch_size);
size_t input_byte_size = 608 * 608 * 3 * sizeof(float) * batch_size;
```
> **failed setting shared memory input: [ 0] INVALID_ARG - The input '000_net' has shared memory of size 8871936 bytes while the expected size is 4435968 bytes**

So the expected size does not take batch size into account.
Matching the expected size then fails during a different sanity check:

```
int batch_size = 2;
options->SetBatchSize(batch_size);
size_t input_byte_size = 608 * 608 * 3 * sizeof(float);
```
> **error: unable to run model: [inference:0 6] INVALID_ARG - unexpected shared-memory size 4435968 for input '000_net', expecting 8871936 for model 'yolov3'**

So now batch size is taken into account for the expected byte size.

philipp-schmidt:
Probably related to #541

CoderHam:
The `input_byte_size` must be the **size of the entire batch** and not just that of one element in the batch since this is the total size in shared memory that will be allocated for that corresponding input. 

In this case you should do:
```
size_t input_byte_size = 2 * 608 * 608 * 3 * sizeof(float) * batch_size;
```

philipp-schmidt:
No matter which size I set, the client side sanity check complains about the shared memory region being too large and tells me it expects a input_byte_size equal to the byte size of batch size 1.

> int batch_size = 2;
options->SetBatchSize(batch_size);
size_t input_byte_size = 608 * 608 * 3 * sizeof(float) * batch_size;

This is equal to the equation you provided, except that the leading 2 is missing, because I'm dealing with a single input and no double input via offsets like the simple_shm_client does.

The error shows, that while I set the correct input_byte_size of the full batch (~8.8MB), the client sanity checks always expect batch size 1 (~4.4MB). I can do the very same for any batch size other than 1.

> **failed setting shared memory input: [ 0] INVALID_ARG - The input '000_net' has shared memory of size 8871936 bytes while the expected size is 4435968 bytes**

My current fix for this is to hardcode the byte size in request_grpc.cc starting from line 849:

```
    // set shared memory
[...]

PR #549, opened 2019-08-12
Fix for batch size != 1

CoderHam:
Fix for #544. Batch size is now taken into account for Inputs in shared memory

### DIFF
--- a/src/clients/c++/request_common.h
+++ b/src/clients/c++/request_common.h
@@ -157,7 +157,7 @@ class InputImpl : public InferContext::Input {
   size_t GetSharedMemoryOffset() const { return shm_offset_; }
 
   void SetBatchSize(size_t batch_size) { batch_size_ = batch_size; }
-  size_t BatchSize() const { return batch_size_; }
+  size_t TotalSendByteSize() const { return total_send_byte_size_; }
 
   const std::vector<int64_t>& Shape() const override { return shape_; }
   Error SetShape(const std::vector<int64_t>& dims) override;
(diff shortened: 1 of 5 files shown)


src/clients/c++/request_grpc.cc
```diff
@@ -854,7 +854,8 @@ InferGrpcContextImpl::PreRunProcessing(std::shared_ptr<Request>& request)
       rshared_memory->set_offset(
           reinterpret_cast<InputImpl*>(io.get())->GetSharedMemoryOffset());
       rshared_memory->set_byte_size(
-          reinterpret_cast<InputImpl*>(io.get())->ByteSize());
+          reinterpret_cast<InputImpl*>(io.get())->ByteSize() *
```
deadeyegoodwin: This should be TotalByteSize... but you've changed that to mean just the total size of the raw bytes. I think you need to fix TotalByteSize to be what it originally was... the total size of an input (i.e. entire batch). See above the call to set_batch_byte_size() is going to be wrong for the shared memory case (and that is needed to correctly support string datatype. You also need a TotalSendByteSize (for http content size and maybe something else but you likely need to make another class member for that).

CoderHam: Done

src/clients/c++/request_common.h
```diff
@@ -157,6 +157,7 @@ class InputImpl : public InferContext::Input {
   size_t GetSharedMemoryOffset() const { return shm_offset_; }
 
   void SetBatchSize(size_t batch_size) { batch_size_ = batch_size; }
+  size_t TotalSendByteSize() const { return total_send_byte_size_; }
```
deadeyegoodwin: Move this up under TotalByteSize()

CoderHam: Done

PR #555, opened 2019-08-14
Quick Start 

maggiezha:
the Quick start needs to update based on my hands on using NGC container version19.07
https://docs.nvidia.com/deeplearning/sdk/tensorrt-inference-server-guide/docs/quickstart.html 
1. "$ /tmp/client/bin/image_client -m resnet50_netdef -s INCEPTION images/mug.jpg"
 it cannot work, as there is nothing in /tmp, so what the user needs to do is: 
The client libraries and examples are found in a tar file in /workspace in the tensorrtserver_client image, for example, v1.4.0.clients.tar.gz, and they should: 
$ tar xzf v1.4.0.clients.tar.gz 
Then the client examples are at /workspace/bin
$ bin/image_client -m resnet50_netdef -s INCEPTION images/mug.jpg 
2.	"To send a request for the resnet50_netdef (Caffe2) model from the example model repository for an image from the qa/images directory:"
there is no folder called qa in this container, so just remove this qa in this sentence and the boxes below, the images are at /workspace/images
3.	In the end of section " Run TensorRT Inference Server" we probably should tell users that they will see a message like this to tell them the model is loaded successful:
successfully loaded 'resnet50_netdef' version 1
Then I got stuck here, and I saw others asked the same question in the Slack channel, as we thought it stuck here and we weren't sure how to do the next, we were expecting it works like you start a container and it will start an interactive session, but this is not how it works, so I suggest we add sth like this to users:
When you see the successful message, your trtis is working properly and ready to get request from clients. Since it is a service running on the backend, now please start a new console to execute the next steps.


GuanLuo:
Are you meant to open an issue? For a pull request, you should create a separate branch from the branch which you want to make changes on, then make the pull request against the new branch and the target branch. This one just tries to merge the current master into r19.07 branch, which doesn't seem right.

PR #558, opened 2019-08-14
Refactor backend contexts to inherit from common base class. Move set input buffer logic to base class

GuanLuo:


src/core/backend_context.h
```diff
@@ -0,0 +1,87 @@
...
+  // set the status of the payload accordingly.
+  // Return true if cudaMemcpyAsync is called, and the caller should call
+  // cudaStreamSynchronize before using the data. Otherwise, return false.
+  bool SetInputBuffer(
+      const std::string& name, const std::vector<size_t>& expected_byte_sizes,
+      std::vector<Scheduler::Payload>* payloads,
+      TRTSERVER_Memory_Type dst_memory_type, char* input_buffer);
+
+  // Helper function to set output buffer of fixed size data type to payloads
+  void SetFixedSizeOutputBuffer(
```
deadeyegoodwin: This is only used by onnx? Are you going to change other backends to use as well?

GuanLuo: Yes. This will be done in a separate PR

PR #559, opened 2019-08-15
Add mem_type parameter in SetFixedSizeOutputBuffer. Modify contexts to use SetFixedSizeOutputBuffer

GuanLuo:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -734,54 +733,18 @@ PlanBackend::Context::Run(std::vector<Scheduler::Payload>* payloads)
...
+    if (byte_sizes_[bindex] < (batch1_byte_size * total_batch_size)) {
+      return Status(
+          RequestStatusCode::INTERNAL,
+          "unexpected size for output '" + name + "', byte-size " +
+              std::to_string(byte_sizes_[bindex]) + " is less than " +
+              std::to_string(total_batch_size) + " * " +
+              std::to_string(batch1_byte_size));
     }
+
+    SetFixedSizeOutputBuffer(
```
deadeyegoodwin: Don't you need to check the return type here?

src/core/backend_context.h
```diff
@@ -78,8 +86,9 @@ struct BackendContext {
   int max_batch_size_;
 
 #ifdef TRTIS_ENABLE_GPU
-  // The stream where data transfer operations are executed.
-  // nullptr if the context is created on CPU.
+  // The stream where data transfer operations are executed on.
+  // It must be set explicitly after the context instance is created,
+  // the default is nullptr.
   cudaStream_t stream_;
```
deadeyegoodwin: I only see a stream created for libtorch backend (except plan which has always created one). I assume other backends will do the same once they support tensors on GPU?

src/backends/caffe2/netdef_backend.cc
```diff
@@ -436,32 +436,23 @@ NetDefBackend::Context::ReadFixedSizedOutputTensor(
...
+  // auto content_memory_type = (gpu_device_ == NO_GPU_DEVICE)
+  //                                ? TRTSERVER_MEMORY_CPU
+  //                                : TRTSERVER_MEMORY_GPU;
+  auto content_memory_type = TRTSERVER_MEMORY_CPU;
+  bool cuda_copy = SetFixedSizeOutputBuffer(
+      name, batch1_byte_size, content, content_shape, content_memory_type,
+      payloads);
+  if (cuda_copy) {
+#ifdef TRTIS_ENABLE_GPU
+    cudaStreamSynchronize(stream_);
```
deadeyegoodwin: You should defer the sync until after all output reads are complete, so that the CPU can run ahead as much as possible.

Does netdef backend even create a stream?

src/core/backend_context.h
GuanLuo: I should have done the same for all backends, will do that. And I think all backend contexts have to create its own stream even if they only support tensors on CPU, since the input data may be in GPU and they need to do a device-to-host copy in that case.

src/backends/caffe2/netdef_backend.cc
GuanLuo: OK. No, netdef backend doesn't create a stream, and will fix that.

PR #568, opened 2019-08-20
Fix TRTIS_ENABLE_GPU=OFF builds

dougn:
* fix #ifdef TRTIS_ENABLE_GPU logic in backend_context.cc

Discovered this while working on build automation. Very simple fix for the build.


deadeyegoodwin:
@dougn Thanks for the fix!

dougn:
Looks like this was reintroduced in another pull request merge from a non-up-to-date branch

https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/core/ensemble_scheduler.cc#L640

PR #579, opened 2019-08-23
Extend simple lib example to support GPU inputs / outputs

GuanLuo:


src/servers/simple.cc
```diff
@@ -63,18 +85,35 @@ ResponseAlloc(
...
 
   // If 'byte_size' is zero just return 'buffer'==nullptr, we don't
-  // need to do any other book-keeping. Only handle allocation in the
-  // CPU region.
-  if ((byte_size == 0) || (memory_type == TRTSERVER_MEMORY_CPU)) {
-    *buffer = (byte_size == 0) ? nullptr : malloc(byte_size);
+  // need to do any other book-keeping.
+  if (byte_size == 0) {
+    *buffer = nullptr;
     *buffer_userp = new std::string(tensor_name);
```
deadeyegoodwin: *buffer_userp = nullptr when byte_size is 0. Release won't be called (will it?) and buffer_userp is just passed to release.

src/servers/simple.cc
```diff
@@ -24,6 +24,7 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#include <cuda_runtime_api.h>
```
deadeyegoodwin: Need some ifdef GPU

src/servers/simple.cc
```diff
@@ -92,9 +131,19 @@ ResponseRelease(
     name = new std::string("<unknown>");
   }
 
-  LOG_INFO << "Releasing buffer " << buffer << " of size " << byte_size
-           << " for result '" << *name << "'";
-  free(buffer);
+  std::string memory_type_str =
```
deadeyegoodwin: Make this global and use it in the logging in the alloc function as well.

src/servers/simple.cc
```diff
@@ -48,6 +69,7 @@ Usage(char** argv, const std::string& msg = std::string())
   }
 
   LOG_ERROR << "Usage: " << argv[0] << " [options]";
+  LOG_ERROR << "\t-g store input data in GPU";
```
deadeyegoodwin: "Use GPU memory for input and output tensors"

src/servers/simple.cc
```diff
@@ -63,18 +85,35 @@ ResponseAlloc(
...
              << tensor_name;
   } else {
-    *buffer = nullptr;
-    *buffer_userp = nullptr;
-    LOG_INFO << "failed to allocated " << byte_size
-             << " bytes for result tensor " << tensor_name;
+    void* allocated_ptr = nullptr;
+    if (memory_type == TRTSERVER_MEMORY_CPU) {
+      allocated_ptr = malloc(byte_size);
+    } else {
```
deadeyegoodwin: Only use GPU if -g flag was specified

src/servers/simple.cc
```diff
@@ -118,11 +167,15 @@ int
 main(int argc, char** argv)
 {
   std::string model_repository_path;
+  bool gpu_input = false;
```
deadeyegoodwin: gpu_input implies only using for input. How about "use_gpu_memory"

PR #605, opened 2019-08-29
Update custom APIs to version 2 which is aware of inputs / outputs memory type

GuanLuo:


src/backends/custom/custom.h
```diff
@@ -40,6 +40,14 @@ extern "C" {
 #define TRTIS_CUSTOM_EXPORT
 #endif
 
+#ifndef CUSTOM_VERSION
```
deadeyegoodwin: I don't think we need this (see other comments). Instead we can just have the version be a uint32_t.

src/backends/custom/custom.h
```diff
@@ -197,6 +208,33 @@ typedef int (*CustomExecuteFn_t)(
     void*, uint32_t, CustomPayload*, CustomGetNextInputFn_t,
     CustomGetOutputFn_t);
 
+/// See CustomGetNextInputFn_t
+///
+/// \param memory_type Acts as both input and output. On input
+/// gives the buffer memory type preferred by the function caller.
+/// Returns the actual memory type of 'content'.
+typedef bool (*CustomGetNextInputVer2Fn_t)(
```
deadeyegoodwin: Lets use "V2" for versioning, so CustomGetNextInputV2Fn_t.

src/backends/custom/custom.h
```diff
@@ -197,6 +208,33 @@ typedef int (*CustomExecuteFn_t)(
     void*, uint32_t, CustomPayload*, CustomGetNextInputFn_t,
     CustomGetOutputFn_t);
 
+/// See CustomGetNextInputFn_t
```
deadeyegoodwin: CustomGetNextInputV2Fn_t

src/backends/custom/custom.h
```diff
@@ -197,6 +208,33 @@ typedef int (*CustomExecuteFn_t)(
...
+///
+/// \param memory_type Acts as both input and output. On input
+/// gives the buffer memory type preferred by the function caller.
+/// Returns the actual memory type of 'content'.
+typedef bool (*CustomGetNextInputVer2Fn_t)(
+    void* input_context, const char* name, const void** content,
+    uint64_t* content_byte_size, CustomMemoryType* memory_type);
+
+/// See CustomGetOutputFn_t
```
deadeyegoodwin: V2

src/backends/custom/custom.h
```diff
@@ -197,6 +208,33 @@ typedef int (*CustomExecuteFn_t)(
...
+///
+/// \param memory_type Acts as both input and output. On input
+/// gives the buffer memory type preferred by the function caller.
+/// Returns the actual memory type of 'content'.
+typedef bool (*CustomGetOutputVer2Fn_t)(
+    void* output_context, const char* name, size_t shape_dim_cnt,
+    int64_t* shape_dims, uint64_t content_byte_size, void** content,
+    CustomMemoryType* memory_type);
+
+/// Type for the CustomExecuteVer2 function.
```
deadeyegoodwin: V2

src/backends/custom/custom.h
```diff
@@ -197,6 +208,33 @@ typedef int (*CustomExecuteFn_t)(
...
+typedef int (*CustomExecuteVer2Fn_t)(
+    void*, uint32_t, CustomPayload*, CustomGetNextInputVer2Fn_t,
+    CustomGetOutputVer2Fn_t);
+
+/// Type for the CustomVersion function.
```
deadeyegoodwin: Move this up above CustomInitializeFn_t. Return type uint32_t.

src/backends/custom/custom.h
```diff
@@ -249,6 +287,15 @@ TRTIS_CUSTOM_EXPORT int CustomExecute(
     void* custom_context, uint32_t payload_cnt, CustomPayload* payloads,
```
deadeyegoodwin: In the description of CustomExecute.
"Execute the custom model using the version 1 implementation of the execute interface. This function must be defined when the custom backend returns 1 from CustomVersion (or when CustomVersion is not defined). ..."

src/backends/custom/custom.h
```diff
@@ -249,6 +287,15 @@ TRTIS_CUSTOM_EXPORT int CustomExecute(
...
     CustomGetNextInputFn_t input_fn, CustomGetOutputFn_t output_fn);
 
+/// See CustomExecute.
```
deadeyegoodwin: "Execute the custom model using the version 2 implementation of the execute interface. This function must be defined when the custom backend returns 2 from CustomVersion. See CustomExecute for description of the parameters."

src/backends/custom/custom.h
```diff
@@ -249,6 +287,15 @@ TRTIS_CUSTOM_EXPORT int CustomExecute(
...
+TRTIS_CUSTOM_EXPORT int CustomExecuteVer2(
+    void* custom_context, uint32_t payload_cnt, CustomPayload* payloads,
+    CustomGetNextInputVer2Fn_t input_fn, CustomGetOutputVer2Fn_t output_fn);
+
+/// Return the custom version. For custom backend that doesn't define this
```
deadeyegoodwin: Move this above CustomInitialize. Here is where we can define what the versions are.

"Get the custom version. For a custom backend that doesn't define this entry point, the inference server will assume the backend version is 1. The currently supported versions are defined below, returning any other version is an error:

Version 1: Input and output tensors must be communicated via system memory (i.e. CPU memory). The CustomExecute function must be defined and CustomGetNextInputFn_t and CustomGetOutputFn_t define the function signature for the input and output callbacks.

Version 2: Input and output tensors may be communicated by both system memory and GPU memory. The CustomExecuteV2 function must be defined and CustomGetNextInputV2Fn_t and CustomGetOutputV2Fn_t define the function signature for the input and output callbacks.

\return the custom version.

src/custom/sdk/custom_instance.cc
```diff
@@ -100,6 +100,25 @@ CustomExecute(
   return instance->Execute(payload_cnt, payloads, input_fn, output_fn);
 }
 
+int
+CustomExecuteVer2(
```
deadeyegoodwin: V2

src/custom/sdk/custom_instance.cc
```diff
@@ -100,6 +100,25 @@ CustomExecute(
...
+{
+  if (custom_instance == nullptr) {
+    return ErrorCodes::Unknown;
+  }
+
+  CustomInstance* instance = static_cast<CustomInstance*>(custom_instance);
+  return instance->Execute(payload_cnt, payloads, input_fn, output_fn);
+}
+
+int
```
deadeyegoodwin: Don't provide a default implementation, force the custom backend writer to think about which version they are implementing.

src/custom/sdk/custom_instance.h
```diff
@@ -73,7 +74,20 @@ class CustomInstance {
   /// output (see CustomGetOutputFn_t).
```
deadeyegoodwin: Add
"\return Error code indicating success or the type of failure"
Instead of returning -1 (which is an non-defined error code) you should add a predefined error code to error_codes.h and return that . The description for that error code should be something about how a V1 version of a function was called when custom backend is not V1.

src/custom/sdk/custom_instance.h
```diff
@@ -73,7 +74,20 @@ class CustomInstance {
...
+  /// Execute the custom instance. User should override this function
+  /// if version 2 of the custom interface is used.
+  ///
+  /// \param payload_cnt The number of payloads to execute.
+  /// \param payloads The payloads to execute.
+  /// \param input_fn The callback function to get tensor input (see
+  /// CustomGetNextInputVer2Fn_t).
+  /// \param output_fn The callback function to get buffer for tensor
+  /// output (see CustomGetOutputVer2Fn_t).
+  virtual int Execute(
```
deadeyegoodwin: Add
"\return Error code indicating success or the type of failure"
Instead of returning -1 (which is an non-defined error code) you should add a predefined error code to error_codes.h and return that . The description for that error code should be something about how a V2 version of a function was called when custom backend is not V2.

src/backends/custom/custom.h
deadeyegoodwin: Comment still needs fixing

PR #606, opened 2019-08-29
Shared Memory tests using Python Client API

CoderHam:
- [x] - L0_infer
- [x] - L0_batcher
- [x] - L0_sequence_batcher
- [x] L0_shared_memory - test shared memory client API functionality

### DIFF
--- a/qa/L0_batcher/batcher_test.py
+++ b/qa/L0_batcher/batcher_test.py
@@ -59,7 +59,7 @@ class BatcherTest(unittest.TestCase):
 
     def check_response(self, trial, bs, thresholds,
                        requested_outputs=("OUTPUT0", "OUTPUT1"), input_size=16,
-                       use_shared_memory=False):
+                       include_shared_memory_test=False):
         global _check_exception
         try:
             start_ms = int(round(time.time() * 1000))
@@ -71,14 +71,14 @@ class BatcherTest(unittest.TestCase):
                                np.float32, np.float32, np.float32, swap=False,
                                model_version=1, outputs=requested_outputs,
                                use_grpc=False, skip_request_id_check=True,
-                               use_streaming=False, use_shared_memory=use_shared_memory)
+                               use_streaming=False, include_shared_memory_test=include_shared_memory_test)
             elif trial == "plan":
                 tensor_shape = (input_size,1,1)
                 iu.infer_exact(self, trial, tensor_shape, bs,
                                np.float32, np.float32, np.float32, swap=False,
                                model_version=1, outputs=requested_outputs,
                                use_grpc=False, skip_request_id_check=True,
-                               use_streaming=False, use_shared_memory=use_shared_memory)
+                               use_streaming=False, include_shared_memory_test=include_shared_memory_test)
             else:
                 self.assertFalse(True, "unknown trial type: " + trial)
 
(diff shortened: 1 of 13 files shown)


qa/L0_batcher/batcher_test.py
```diff
@@ -151,6 +152,26 @@ def test_static_batch_preferred(self):
             except InferenceServerException as ex:
                 self.assertTrue(False, "unexpected error {}".format(ex))
 
+    def test_static_batch_preferred_shm(self):
```
GuanLuo: Only this case is tested with shared memory?

qa/common/infer_util.py
```diff
@@ -55,15 +57,21 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
                 skip_request_id_check=False, use_streaming=True,
-                correlation_id=0):
+                correlation_id=0, use_shared_memory=False):
     tester.assertTrue(use_http or use_grpc or use_streaming)
     configs = []
     if use_http:
-        configs.append(("localhost:8000", ProtocolType.HTTP, False))
+        if use_shared_memory:
+            configs.append(("localhost:8000", ProtocolType.HTTP, False, True))
+        configs.append(("localhost:8000", ProtocolType.HTTP, False, False))
```
GuanLuo: So if `use_shared_memory == True`, the "no shared memory" case will still be tested?

qa/common/infer_util.py
```diff
@@ -143,24 +170,38 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
-                output_req[OUTPUT0] = InferContext.ResultFormat.RAW
-            else:
-                output_req[OUTPUT0] = (InferContext.ResultFormat.CLASS, num_classes)
-        if "OUTPUT1" in outputs:
-            if output1_raw:
-                output_req[OUTPUT1] = InferContext.ResultFormat.RAW
-            else:
-                output_req[OUTPUT1] = (InferContext.ResultFormat.CLASS, num_classes)
+        if config[3]:
+            if "OUTPUT0" in outputs:
```
GuanLuo: change the order:
```
if "OUTPUT0" in outputs:
    if config[3]:
        ...
    elif output0_raw:
        ...
    else:
        ...
```

src/clients/python/crequest.cc
```diff
@@ -29,6 +29,7 @@
 #include <iostream>
 #include "src/clients/c++/request_grpc.h"
 #include "src/clients/c++/request_http.h"
+#include "src/clients/python/shared_memory/shared_memory_handle.h"
```
GuanLuo: Isn't this the wrapper for the Python only? The request API doesn't need to know anything about it, right?

qa/common/infer_util.py
deadeyegoodwin: include_shared_memory_test might be a better name to imply that in addition to regular testing also want to do shared-memory testing.

deadeyegoodwin:
@CoderHam Need to rebase this to master to get rid of the already committed changes.

src/clients/python/crequest.cc
CoderHam: This contains the shared_memory_handle struct used to get the data inside it from the void ptr handle that was returned to python when the shared memory region was created.

qa/common/infer_util.py

qa/common/infer_util.py
CoderHam: Sure

qa/L0_batcher/batcher_test.py
CoderHam: Modified for greater test coverage.

PR #611, opened 2019-09-03
Adding support for the qa model versions

tanmayv25:
These changes are needed to maintain separate sets of models for the different versions of TRTIS.


docs/test.rst
```diff
@@ -73,9 +73,9 @@ Run QA Container
 Now run the QA container and mount the QA model repositories into the
 container so the tests will be able to access them::
 
-  $ nvidia-docker run -it --rm -v/tmp:/data/inferenceserver tensorrtserver_qa
+  $ nvidia-docker run -it --rm -v/tmp:/data/inferenceserver/<TRTIS_CONTAINER_VERSION> tensorrtserver_qa
```
deadeyegoodwin: I don't think this is quite correct. The gen scripts will leave files in, for example, /tmp/qa_model_repository. There is no version directory there.

qa/common/run_all_tests.sh
```diff
@@ -30,9 +30,10 @@ DIRS=(../L*/)
 
 passed=0
 failed=0
+TRTIS_CONTAINER_VERSION= $1
```
deadeyegoodwin: Remove space after '='

qa/common/run_all_tests.sh
```diff
@@ -30,9 +30,10 @@ DIRS=(../L*/)
...
 for dir in "${DIRS[@]}"; do
     echo -e "Running $dir...\n"
-    (cd $dir && ./test.sh)
+    (cd $dir && ./test.sh ${TRTIS_CONTAINER_VERSION})
```
GuanLuo: This helper script is used for running all tests on local machine, and it is not used in CI (CI runs each test individually). You need to change `.yaml` so that all tests will be run with version argument. If you look at your CI, you will see that `TRTIS_CONTAINER_VERSION` is always an empty string

GuanLuo: Perhaps instead of setting `TRTIS_CONTAINER_VERSION` via argument, just make it an environment variable that is set on building the container?

tanmayv25: Yes. Currently it is an empty string. Once the TRT6 changes are pushed in we can have all these things well set up. This is needed for the transition phase. Have a look at https://gitlab-master.nvidia.com/dl/dgx/tensorrtserver/merge_requests/38. The YAML changes can be found in
 https://gitlab-master.nvidia.com/dl/dgx/tensorrtserver/commit/3700b2a93d27df34e03ce1b03d996d5c408540b8. As these might break the current CIs, the YAML changes are not yet included.

docs/test.rst
tanmayv25: Yes, there is no version directory in /tmp but the tests after modification will expect the models to be in /data/inferenceserver/<version>/qa_*  path. Hence, mounting the qa_* directories to the /data/inferenceserver/<version>.

qa/common/run_all_tests.sh
tanmayv25: For the same reason not using the environment variable. Can not force the TRTIS to run off from the new model plans until and unless the TRT6 changes are in. Additionally, a separate parameter to the test script will give user freedom to run TRTIS on earlier model version. For instance, TRT6 is backwards compatible with TRT5 model plans.

PR #636, opened 2019-09-06
Decouple setting input buffer and tensor creation in torch backend

GuanLuo:


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -371,82 +378,29 @@ LibTorchBackend::Run(
   OnCompleteQueuedPayloads(status);
 }
 
-Status
-LibTorchBackend::Context::SetFixedSizedInputTensor(
```
GuanLuo: This function is modified and moved to [SetFixedSizedInputBuffer()](https://github.com/NVIDIA/tensorrt-inference-server/pull/636/files#diff-bd1504eb0ecb5a2546e4ec040de4bdd1R538) below

PR #648, opened 2019-09-11
SharedMemory Status server API support

CoderHam:
- [x] fix shared memory unregisterAll server implementation

### DIFF
--- a/src/clients/c++/request_grpc.cc
+++ b/src/clients/c++/request_grpc.cc
@@ -310,10 +310,18 @@ class SharedMemoryControlGrpcContextImpl : public SharedMemoryControlContext {
   Error UnregisterSharedMemory(const std::string& name) override;
   Error UnregisterAllSharedMemory() override;
 
+  enum ControlType {
+    REGISTER = 1,
+    UNREGISTER = 2,
+    UNREGISTER_ALL = 3,
+    GET_STATUS = 4
+  };
+
  private:
   Error SendRequest(
-      const std::string& name, const SharedMemoryControlRequest::Type action,
-      const std::string& shm_key, const size_t offset, const size_t byte_size);
+      const std::string& name, const ControlType action,
+      const std::string& shm_key, const size_t offset, const size_t byte_size,
+      SharedMemoryStatus* shm_status);
 
   // GRPC end point.
   std::unique_ptr<GRPCService::Stub> stub_;
@@ -333,58 +341,79 @@ SharedMemoryControlGrpcContextImpl::RegisterSharedMemory(
     const std::string& name, const std::string& shm_key, const size_t offset,
     const size_t byte_size)
 {
-  return SendRequest(
-      name, SharedMemoryControlRequest::REGISTER, shm_key, offset, byte_size);
+  return SendRequest(name, REGISTER, shm_key, offset, byte_size, nullptr);
 }
 
 Error
 SharedMemoryControlGrpcContextImpl::UnregisterSharedMemory(
     const std::string& name)
 {
-  return SendRequest(name, SharedMemoryControlRequest::UNREGISTER, "", 0, 0);
+  return SendRequest(name, UNREGISTER, "", 0, 0, nullptr);
 }
 
 Error
 SharedMemoryControlGrpcContextImpl::UnregisterAllSharedMemory()
 {
-  return SendRequest("", SharedMemoryControlRequest::UNREGISTER_ALL, "", 0, 0);
+  return SendRequest("", UNREGISTER_ALL, "", 0, 0, nullptr);
 }
 
 Error
 SharedMemoryControlGrpcContextImpl::SendRequest(
-    const std::string& name, const SharedMemoryControlRequest::Type action,
-    const std::string& shm_key, const size_t offset, const size_t byte_size)
+    const std::string& name, const ControlType action,
+    const std::string& shm_key, const size_t offset, const size_t byte_size,
+    SharedMemoryStatus* shm_status)
 {
   SharedMemoryControlRequest request;
   SharedMemoryControlResponse response;
   grpc::ClientContext context;
+  if (shm_status != nullptr) {
+    shm_status->Clear();
+  }
 
-  if (action == SharedMemoryControlRequest::REGISTER) {
-    auto rshm_region = request.mutable_shared_memory_region();
+  if (action == REGISTER) {
+    auto register_info = request.mutable_register_();
+    auto rshm_region = register_info->mutable_shared_memory_region();
     rshm_region->set_name(name);
     rshm_region->set_shared_memory_key(shm_key);
     rshm_region->set_offset(offset);
     rshm_region->set_byte_size(byte_size);
-    request.set_type(action);
-  } else if (action == SharedMemoryControlRequest::UNREGISTER) {
-    auto rshm_region = request.mutable_shared_memory_region();
-    rshm_region->set_name(name);
-    request.set_type(action);
-  } else if (action == SharedMemoryControlRequest::UNREGISTER_ALL) {
-    request.set_type(action);
+  } else if (action == UNREGISTER) {
+    auto unregister_info = request.mutable_unregister();
+    unregister_info->set_name(name);
+  } else if (action == UNREGISTER_ALL) {
+    request.mutable_unregister_all();
+  } else if (action == GET_STATUS) {
+    request.mutable_get_status();
   }
 
+  Error grpc_status;
   grpc::Status status =
       stub_->SharedMemoryControl(&context, request, &response);
   if (status.ok()) {
-    return Error(response.request_status());
+    grpc_status = Error(response.request_status());
+    if (action == GET_STATUS) {
+      std::string response_str;
+      response.clear_request_status();
+      response.SerializeToString(&response_str);
+      if (!shm_status->ParseFromString(response_str)) {
+        return Error(
+            RequestStatusCode::INTERNAL,
+            "failed to parse shared memory status");
+      }
+    }
   } else {
     // Something wrong with the GRPC conncection
-    return Error(
+    grpc_status = Error(
         RequestStatusCode::INTERNAL,
         "GRPC client failed: " + std::to_string(status.error_code()) + ": " +
             status.error_message());
   }
+
+  // Log shared memory status if request is SUCCESS and verbose is true.
+  if ((grpc_status.IsOk()) && verbose_ && (action == GET_STATUS)) {
+    std::cerr << shm_status->DebugString() << std::endl;
+  }
+  return grpc_status;
 }
 
 Error
--- a/src/clients/c++/request_grpc.cc
+++ b/src/clients/c++/request_grpc.cc
@@ -310,19 +310,7 @@ class SharedMemoryControlGrpcContextImpl : public SharedMemoryControlContext {
   Error UnregisterSharedMemory(const std::string& name) override;
   Error UnregisterAllSharedMemory() override;
 
-  enum ControlType {
-    REGISTER = 1,
-    UNREGISTER = 2,
-    UNREGISTER_ALL = 3,
-    GET_STATUS = 4
-  };
-
  private:
-  Error SendRequest(
-      const std::string& name, const ControlType action,
-      const std::string& shm_key, const size_t offset, const size_t byte_size,
-      SharedMemoryStatus* shm_status);
-
   // GRPC end point.
   std::unique_ptr<GRPCService::Stub> stub_;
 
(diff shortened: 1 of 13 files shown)


src/core/grpc_service.proto
```diff
@@ -325,6 +325,12 @@ message SharedMemoryControlRequest
     //@@       To unregister all active shared memory regions.
     //@@
     UNREGISTER_ALL = 2;
+
```
deadeyegoodwin: Having one request type for all the control requests doesn't work if the different requests need different inputs. Instead look at the TraceControl messages to see how you can use oneof to define different request (and response) for different request types.

src/core/server.cc
```diff
@@ -415,6 +416,12 @@ InferenceServer::SharedMemoryAddress(
       name, offset, byte_size, shm_mapped_addr);
 }
 
+Status
+InferenceServer::GetSharedMemoryStatus(SharedMemoryControlResponse* shm_status)
+{
+  return shared_memory_manager_->GetSharedMemoryStatus(shm_status);
```
deadeyegoodwin: Need to check ready_state_ and update inflight_request_counter_ as in UnregisterAllSharedMemory.

src/core/server.h
```diff
@@ -38,6 +38,7 @@
 #include "src/core/provider.h"
 #include "src/core/server_status.h"
 #include "src/core/server_status.pb.h"
+#include "src/core/grpc_service.pb.h"
```
deadeyegoodwin: You need to run the formatter

src/core/shared_memory_manager.cc
```diff
@@ -205,6 +205,17 @@ SharedMemoryManager::UnregisterAllSharedMemory()
   return Status::Success;
 }
 
+Status
+SharedMemoryManager::GetSharedMemoryStatus(SharedMemoryControlResponse* shm_status)
```
deadeyegoodwin: This needs an implementation.

src/core/trtserver.cc
```diff
@@ -1154,6 +1156,36 @@ TRTSERVER_ServerSharedMemoryAddress(
   return nullptr;  // success
 }
 
+TRTSERVER_Error*
+TRTSERVER_ServerGetSharedMemoryStatus(
```
deadeyegoodwin: ServerSharedMemoryStatus

src/core/grpc_service.proto
CoderHam: Done

src/clients/c++/request_grpc.cc
```diff
@@ -333,58 +341,79 @@ SharedMemoryControlGrpcContextImpl::RegisterSharedMemory(
...
 
 Error
 SharedMemoryControlGrpcContextImpl::UnregisterAllSharedMemory()
 {
-  return SendRequest("", SharedMemoryControlRequest::UNREGISTER_ALL, "", 0, 0);
+  return SendRequest("", UNREGISTER_ALL, "", 0, 0, nullptr);
 }
 
 Error
 SharedMemoryControlGrpcContextImpl::SendRequest(
```
deadeyegoodwin: Why use a common routine? There is not that much common code across the request types and it just makes it confusing.

src/core/grpc_service.proto
```diff
@@ -370,47 +370,69 @@ message InferSharedMemoryRegion
...
 {
-  //@@  .. cpp:enum:: Type
+  //@@  .. cpp:var:: message RegisterInfo
   //@@
-  //@@     Types of control operations for shared memory
+  //@@     The information needed to register the shared memory region.
   //@@
-  enum Type {
-    //@@    .. cpp:enumerator:: Type::REGISTER = 0
+  message RegisterInfo
```
deadeyegoodwin: Drop "info" here and elsewhere. Just "Register".

src/core/grpc_service.proto
```diff
@@ -370,47 +370,69 @@ message InferSharedMemoryRegion
 //@@
 message SharedMemoryControlRequest
 {
-  //@@  .. cpp:enum:: Type
+  //@@  .. cpp:var:: message RegisterInfo
   //@@
-  //@@     Types of control operations for shared memory
+  //@@     The information needed to register the shared memory region.
```
deadeyegoodwin: "Register a shared memory region".

src/core/grpc_service.proto
```diff
@@ -370,47 +370,69 @@ message InferSharedMemoryRegion
...
   //@@
-  //@@  .. cpp:var:: Type type
+  //@@     Information needed to UnregisterAll and GetStatus of shared memory.
   //@@
-  //@@     The control type that states whether to register/unregister the
-  //@@     specified shared memory region. Unregister all active shared memory
-  //@@     regions or get the list of active shared memory regions
-  //@@
-  Type type = 1;
+  message NoInfo {}
```
deadeyegoodwin: Make an "UnregisterAll" and "Status" message. Both will be empty.

src/core/grpc_service.proto
```diff
@@ -370,47 +370,69 @@ message InferSharedMemoryRegion
...
+
+    //@@    .. cpp:var:: NoInfo unregister_all
+    //@@
+    //@@       To unregister all active shared memory regions.
+    //@@
+    NoInfo unregister_all = 3;
+
+    //@@    .. cpp:var:: NoInfo get_status
+    //@@
+    //@@       To get the list of all active shared memory regions.
```
deadeyegoodwin: Not a "list". "Get the status of all active shared memory regions".

src/core/grpc_service.proto
```diff
@@ -426,6 +448,13 @@ message SharedMemoryControlResponse
   //@@     The status of the request, indicating success or failure.
   //@@
   RequestStatus request_status = 1;
+
+  //@@
+  //@@  .. cpp:var:: SharedMemoryRegion shared_memory_region
+  //@@
+  //@@     The list of active/registered shared memory regions.
+  //@@
+  repeated InferSharedMemoryRegion shared_memory_region = 2;
```
deadeyegoodwin: This is only relevant for "Status" so again so a oneof with a "Status" message where in that message is this repeated. Also status results should be returned as a SharedMemoryStatus message, right?

src/servers/grpc_server.cc
```diff
@@ -1476,40 +1476,56 @@ SharedMemoryControlHandler::Process(Handler::State* state, bool rpc_ok)
...
+      err = TRTSERVER_ServerSharedMemoryStatus(
+          trtserver_.get(), &shm_status_protobuf);
+      if (err == nullptr) {
+        const char* status_buffer;
+        size_t status_byte_size;
+        err = TRTSERVER_ProtobufSerialize(
+            shm_status_protobuf, &status_buffer, &status_byte_size);
         if (err == nullptr) {
-          err = TRTSERVER_ServerUnregisterAllSharedMemory(trtserver_.get());
+          if (!response.ParseFromArray(status_buffer, status_byte_size)) {
```
deadeyegoodwin: How can this be correct. The response is not just the SharedMemoryStatus message.

src/core/grpc_service.proto
```diff
@@ -370,47 +370,69 @@ message InferSharedMemoryRegion
...
-    //@@    .. cpp:enumerator:: Type::REGISTER = 0
+  message RegisterInfo
+  {
+    //@@  .. cpp:var:: InferSharedMemoryRegion shared_memory_region
     //@@
-    //@@       To register the specified shared memory region.
+    //@@     The shared memory region to register.
     //@@
-    REGISTER = 0;
+    InferSharedMemoryRegion shared_memory_region = 5;
```
deadeyegoodwin: InferSharedMemoryRegion is no longer needed as a separate mesage. The contents of that message should become the contents of Register.

src/core/shared_memory_manager.cc
```diff
@@ -149,11 +149,8 @@ SharedMemoryManager::RegisterSharedMemory(
 }
 
 Status
-SharedMemoryManager::UnregisterSharedMemory(const std::string& name)
+SharedMemoryManager::UnregisterSharedMemoryHelper(const std::string& name)
 {
```
deadeyegoodwin: Add a comment that the lock must be held when this function is called.

src/core/shared_memory_manager.cc
```diff
@@ -205,6 +211,23 @@ SharedMemoryManager::UnregisterAllSharedMemory()
   return Status::Success;
```
deadeyegoodwin: Above you shouldn't just log an error you should also return the same message as an error status.

src/core/shared_memory_manager.h
```diff
@@ -83,8 +83,15 @@ class SharedMemoryManager {
       const std::string& name, const std::string& shm_key, const size_t offset,
       const size_t byte_size);
 
+  /// Helper function that unregisters a specified shared memory region if
+  /// registered else do nothing and return success.
+  /// \param name The user-given name for the shared memory region to be
+  /// registered.
+  /// \return error status.
+  Status UnregisterSharedMemoryHelper(const std::string& name);
```
deadeyegoodwin: This should be private.

src/servers/grpc_server.cc
```diff
@@ -1476,40 +1476,56 @@ SharedMemoryControlHandler::Process(Handler::State* state, bool rpc_ok)
...
-        err = smb_manager_->Remove(&smb, request.shared_memory_region().name());
-        if ((err == nullptr) && (smb != nullptr)) {
-          err = TRTSERVER_ServerUnregisterSharedMemory(trtserver_.get(), smb);
-          TRTSERVER_Error* del_err = TRTSERVER_SharedMemoryBlockDelete(smb);
-          if (del_err != nullptr) {
-            LOG_ERROR << "failed to delete shared memory block: "
-                      << TRTSERVER_ErrorMessage(del_err);
-            TRTSERVER_ErrorDelete(del_err);
-          }
+    if (request.has_register_()) {
```
deadeyegoodwin: has_register_? Does that function even exist?

src/servers/grpc_server.cc
```diff
@@ -1476,40 +1476,56 @@ SharedMemoryControlHandler::Process(Handler::State* state, bool rpc_ok)
...
+          }
         }
-        break;
-      default:
-        err = TRTSERVER_ErrorNew(
-            TRTSERVER_ERROR_UNKNOWN,
-            "unknown sharedmemorycontrol request type");
-        break;
+      }
+      if (err != nullptr) {
```
deadeyegoodwin: Why log an error here? We don't log for other request failures. We just return the error status.

src/servers/http_server.cc
```diff
@@ -687,8 +700,49 @@ HTTPAPIServer::HandleSharedMemoryControl(
...
+                "failed to parse shared memory status");
+          } else {
+            std::string shm_status_str = shm_status.DebugString();
+            evbuffer_add(
+                req->buffer_out, shm_status_str.c_str(), shm_status_str.size());
+          }
+        }
+      }
+    }
+    if (err != nullptr) {
```
deadeyegoodwin: Remove this logging

src/core/grpc_service.proto
CoderHam: Sure

src/clients/c++/request_grpc.cc
CoderHam: Will split this into different components and remove the SendRequest function

src/servers/grpc_server.cc
CoderHam: register gets named as register_ due to a keyword issue or conflict.

src/core/grpc_service.proto
CoderHam: That is correct

src/servers/grpc_server.cc
CoderHam: I add the request status later. Not entirely sure why the ParseFromArray does not throw an error though.

src/core/grpc_service.proto
CoderHam: Deleted InferSharedMemoryRegion. Reuses SharedMemoryRegion from server_status.proto

src/core/grpc_service.proto
```diff
@@ -370,47 +370,75 @@ message InferSharedMemoryRegion
...
-  //@@     specified shared memory region. Unregister all active shared memory
-  //@@     regions or get the list of active shared memory regions
+  message UnregisterAll {}
+
+  //@@  .. cpp:var:: message GetStatus
+  //@@
+  //@@     Get the status of all active shared memory regions.
   //@@
-  Type type = 1;
+  message GetStatus {}
```
deadeyegoodwin: Just "Status"

src/core/grpc_service.proto
deadeyegoodwin: This comment seems wrong. I still see InferSharedMemoryRegion. It should be deleted and contents moved to Register. SharedMemoryRegion is the status and in the future we may add more things to the status so don't reuse that here.

src/core/grpc_service.proto
deadeyegoodwin: So this needs to be a oneof which (currently) just has a "Status" message which includes SharedMemoryStatus.

CoderHam:
Register is re-used in the SharedMemoryStatus. Also the SharedMemoryStatus Response should be Status + repeated SharedMemoryRegions

src/core/grpc_service.proto
```diff
@@ -314,12 +328,37 @@ message SharedMemoryControlRequest
 //@@
 message SharedMemoryControlResponse
 {
+  //@@
+  //@@.. cpp:var:: message SharedMemoryStatusResponse
+  //@@
+  //@@   Status of all active shared memory regions.
+  //@@
+  message SharedMemoryStatusResponse
```
deadeyegoodwin: Just "Status" to parallel what is done on the request side.

src/core/grpc_service.proto
```diff
@@ -314,12 +328,37 @@ message SharedMemoryControlRequest
...
+  }
+
   //@@
   //@@  .. cpp:var:: RequestStatus request_status
   //@@
   //@@     The status of the request, indicating success or failure.
   //@@
   RequestStatus request_status = 1;
+
+  oneof status_response
```
deadeyegoodwin: oneof shared_memory_control

Issue #603, opened 2019-08-29
Changing batch size between requests with shared memory fails


philipp-schmidt:
When running two clients with different batch sizes after each other, I get server errors on running inference (using shared memory) regarding the expected byte sizes.

Everything works well as long as batch sizes are consistent between clients or when the client is the very first client, but whenever a client has already set a batch size on the server before and the batch size differs for the currently requested one, I run into something similar to this:

Client error:
```
Server error message: expected buffer size to be 2945760bytes but gets 5891520 bytes in output tensor
```
Server error:
```
[trtserver.cc:1212] Infer failed: expected buffer size to be 2945760bytes but gets 5891520 bytes in output tensor
```

This happens when trying batch size 16, after another client successfully ran batch size 8, is done and unregistered his shared memory. So registering the shared memory with the new size seems to be okay, but running inference fails. All clients use the same model of course.

CoderHam:
Tracking with [this](https://jirasw.nvidia.com/browse/DLIS-676).

CoderHam:
@philipp-schmidt are you setting the run options (outputs as well) when you are running with the new batch_size?

```
  options->SetBatchSize(16);
  options->AddSharedMemoryResult(output0, "output_data", 0, output_byte_size);
  options->AddSharedMemoryResult(
      output1, "output_data", output_byte_size, output_byte_size);

  FAIL_IF_ERR(infer_ctx->SetRunOptions(*options), "unable to set inference options");
```
It is important to note that the output_byte_size must be modified based on the batch size.

philipp-schmidt:
Yes, in the following order and in more detail:

```cpp
// Set batch size
options->SetBatchSize(batch_size);

// Calculate byte sizes
size_t frame_byte_size = GetByteSize(output->DType(), output->Dims());
size_t total_output_size = frame_byte_size * batch_size;

// Create output in shared memory
...

// Register output shared memory with TRTIS
shared_memory_ctx->RegisterSharedMemory(shm_key, shm_key, 0, total_output_size);

// Add result
options->AddSharedMemoryResult(output, shm_key, 0, total_output_size);

// Set options
inference_ctx->SetRunOptions(*options)
```

For me it seems weird to communicate the batch size in the options after doing registrations with said batch size, but other than running SetRunOptions() twice (once after the batch size and then once after AddSharedMemoryResult) I don't see how that could be changed.

philipp-schmidt:
[...]

PR #649, opened 2019-09-11
Fix GRPC shm_map life cycle 

CoderHam:
Caused the failure: #603


src/servers/grpc_server.cc
```diff
@@ -711,6 +711,7 @@ InferAllocatorPayload(
     AllocPayload* alloc_payload)
 {
   alloc_payload->response_ = &response;
+  alloc_payload->shm_map_ = new AllocPayload::TensorShmMap;
```
deadeyegoodwin: We don't want to allocate this everytime, and also not for all the requests that don't even use shared memory. So use:
if (alloc_payload->shm_map_ != nullptr) { alloc_payload->shm_map_.clear(); }

src/servers/grpc_server.cc
```diff
@@ -726,10 +727,6 @@ InferAllocatorPayload(
           trtserver.get(), smb, io.shared_memory().offset(),
           io.shared_memory().byte_size(), &base));
 
-      if (alloc_payload->shm_map_ == nullptr) {
```
deadeyegoodwin: Keep this.

src/servers/grpc_server.cc
CoderHam: Fixed

src/servers/grpc_server.cc

PR #661, opened 2019-09-16
Add option to use TensorRT execution accelerator in ONNX Runtime backend

GuanLuo:


qa/L0_onnx_execution_provider/test.sh
```diff
@@ -0,0 +1,151 @@
...
+    echo -e "\n***\n*** Failed. Expected OpenVINO is not supported\n***"
+    RET=1
+fi
+grep "CUDA Execution Accelerator is set for onnx_float32_float32_float32_openvino" $SERVER_LOG
+if [ $? -ne 0 ]; then
+    echo -e "\n***\n*** Failed. Expected CUDA Execution Accelerator is set\n***"
+    RET=1
+fi
+
+grep "OpenVINO Execution Accelerator is not supported for onnx_float32_float32_float32_cpu_openvino" $SERVER_LOG
```
deadeyegoodwin: Why isn't openvino supported for a CPU context?

src/backends/onnx/onnx_backend.cc
```diff
@@ -204,15 +205,42 @@ OnnxBackend::CreateExecutionContext(
   OrtResourceWrapper<OrtSessionOptions*> options_wrapper(
       session_options, &OrtReleaseSessionOptions);
 
+  // Set execution execution_accelerators (execution providers in ONNX Runtime)
   if (gpu_device != Context::NO_GPU_DEVICE) {
 #ifdef TRTIS_ENABLE_GPU
+    if (Config().optimization().has_execution_accelerators()) {
+      // Don't need to ensure uniqueness of the providers,
+      // ONNX Runtime will check it.
+      for (const auto& execution_accelerator : Config().optimization().execution_accelerators().gpu_execution_accelerator()) {
```
deadeyegoodwin: Need to format?

qa/L0_onnx_execution_provider/test.sh
GuanLuo: This PR hasn't included build with OpenVINO, will work on OpenVINO and change this test accordingly in another PR.

src/backends/onnx/onnx_backend.cc
GuanLuo: Will do

PR #662, opened 2019-09-17
Narrow TRTSERVER trace interface to support minimal required for trace.

deadeyegoodwin:
Remove open-trace style tracing for now.

Leave the decision of what to do with the trace to the client of the
TRTSERVER API. In trtserver executable report the trace using zipkin
library.


src/core/tracing.cc
```diff
@@ -28,87 +28,63 @@
...
-  cppkin::Init(cppkin_params);
+  LOG_INFO << str << ts.tv_sec * NANOS_PER_SECOND + ts.tv_nsec;
 }
-
-Status
-TraceManager::SetLevel(uint32_t level, uint32_t rate)
+#endif
+#if 0
+void
+Trace::Submit(const std::vector<struct timespec>& timestamps)
```
GuanLuo: What will this two functions be used for? Logging to console in addition to writing to file?

deadeyegoodwin: Those were left in by mistake. I will remove them in the next PR.

PR #667, opened 2019-09-18
Add option to use OpenVINO execution accelerator in ONNX Runtime

GuanLuo:
We can see where to put the OpenVINO directory to make the docker image nicer, right now it is put into `/opt/tensorrt/lib`, ONNX Runtime looks up OpenVINO module by directory hierarchy so we couldn't keep it into a flat structure


Dockerfile
```diff
@@ -367,6 +406,26 @@ COPY --from=trtserver_build /opt/tensorrtserver/include include
...
+ENV INTEL_CVSDK_DIR /opt/tensorrtserver/lib/openvino_2019.1.144
+ENV INTEL_OPENVINO_DIR /opt/tensorrtserver/lib/openvino_2019.1.144
+
+ENV LD_LIBRARY_PATH $INTEL_CVSDK_DIR/deployment_tools/inference_engine/lib/intel64:$INTEL_CVSDK_DIR/deployment_tools/inference_engine/external/tbb/lib:$LD_LIBRARY_PATH
+
+ENV PATH $INTEL_CVSDK_DIR/deployment_tools/model_optimizer:$PATH
+ENV PYTHONPATH $INTEL_CVSDK_DIR/deployment_tools/model_optimizer:$INTEL_CVSDK_DIR/tools:$PYTHONPATH
+ENV IE_PLUGINS_PATH $INTEL_CVSDK_DIR/deployment_tools/inference_engine/lib/intel64
+
+# ONNX Runtime uses Python3 syntax
```
deadeyegoodwin: ONNX runtime needs python to run? Is this only for openvino?

Dockerfile
```diff
@@ -367,6 +406,26 @@ COPY --from=trtserver_build /opt/tensorrtserver/include include
...
+RUN apt-get update && apt-get install -y --no-install-recommends python3-pip
+RUN pip3 install --upgrade wheel setuptools && \
+    (cd $INTEL_CVSDK_DIR/deployment_tools/model_optimizer && \
+        pip3 install -r requirements_onnx.txt)
+# Python scripts that need to be in 'pwd' on running the application
+# https://github.com/microsoft/onnxruntime/issues/1861
+COPY --from=trtserver_onnx /workspace/build/Release/openvino_* \
```
deadeyegoodwin: We don't want these scripts at the top-level of /opt/tensorrtserver

Dockerfile
```diff
@@ -248,18 +269,32 @@ COPY --from=trtserver_onnx /workspace/onnxruntime/include/onnxruntime/core/provi
...
+RUN cd /opt/tensorrtserver/lib && ln -sf libtbb.so.2 libtbb.so
+
 # Copy entire repo into container even though some is not needed for
 # build itself... because we want to be able to copyright check on
 # files that aren't directly needed for build.
 WORKDIR /workspace
 RUN rm -fr *
 COPY . .
 
+RUN find / -name libtbb.so.2
```
deadeyegoodwin: Remove this

Dockerfile
```diff
@@ -367,6 +404,22 @@ COPY --from=trtserver_build /opt/tensorrtserver/include include
...
+COPY --from=trtserver_onnx /workspace/build/Release/openvino_* \
+     /opt/openvino_scripts/
+COPY --from=trtserver_onnx /data/dldt/openvino_2019.1.144/deployment_tools/model_optimizer \
+     /opt/openvino_scripts/openvino_2019.1.144/deployment_tools/model_optimizer/
+COPY --from=trtserver_onnx /data/dldt/openvino_2019.1.144/tools \
+     /opt/openvino_scripts/openvino_2019.1.144/tools
+ENV INTEL_CVSDK_DIR /opt/openvino_scripts/openvino_2019.1.144
+ENV PYTHONPATH $INTEL_CVSDK_DIR:$INTEL_CVSDK_DIR/deployment_tools/model_optimizer:$INTEL_CVSDK_DIR/tools:$PYTHONPATH
+
+# ONNX Runtime uses Python3 syntax
```
deadeyegoodwin: Say something about ONNX Runtime requiring python3 to generate OpenVINO models.

src/backends/onnx/onnx_backend.cc
```diff
@@ -237,22 +241,37 @@ OnnxBackend::CreateExecutionContext(
...
-        LOG_ERROR << "OpenVINO Execution Accelerator is not supported for "
-                  << instance_name;
+#ifdef TRTIS_ENABLE_ONNXRUNTIME_OPENVINO
+        need_lock = true;
+        RETURN_IF_ORT_ERROR(OrtSessionOptionsAppendExecutionProvider_OpenVINO(
+            session_options, "CPU"));
+        LOG_VERBOSE(1) << "OpenVINO Execution Accelerator is set for "
+                        << instance_name << " on device CPU"; 
+#else
+        LOG_ERROR << "OpenVINO Execution Accelerator is not enabled";
```
deadeyegoodwin: Return an error status here. If model configuration is invalid in any way we want to fail.

src/backends/onnx/onnx_backend.cc
```diff
@@ -237,22 +241,37 @@ OnnxBackend::CreateExecutionContext(
...
+#endif  // TRTIS_ENABLE_ONNXRUNTIME_OPENVINO
       } else {
         LOG_ERROR << "Ignore unknown Execution Accelerator '"
```
deadeyegoodwin: Seems like this should be a status error as well.

PR #671, opened 2019-09-19
Updates for performance analysis and tracing

deadeyegoodwin:


dzier:
Are the logs used in the test as a base line comparison?

deadeyegoodwin:
@dzier The files in baseline/ and legacy_baseline/ are used as comparison against current version.

PR #673, opened 2019-09-19
Add the shape validation for dynamic shapes

tanmayv25:
TRT engine->setBindingDimensions() doesn't validate the shape it is trying to set. This shape should be in the range of selected optimization profile. Hence, adding the validation in TRTIS to ensure the sanity.

### DIFF
--- a/qa/L0_trt_dynamic_shape/test.sh
+++ b/qa/L0_trt_dynamic_shape/test.sh
@@ -1,0 +1,102 @@
+#!/bin/bash
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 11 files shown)


qa/L0_trt_dynamic_shape/test.sh
```diff
@@ -0,0 +1,84 @@
...
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+# Shape beyond the limits of optimization profile
+set +e
+$PERF_CLIENT -v -i grpc -u localhost:8001 -m plan_float32_float32_float32 --shape INPUT0:33 --shape INPUT1:33 -t 1 -p2000 -b 1 > ${CLIENT_LOG} 2>&1
```
deadeyegoodwin: You should test both too-small dimension and too-large dimension. And check for both a runtime failure and a load-time failure. You added code for both kinds of checks so you need to cover them all in your test.

PR #680, opened 2019-09-20
Updated API for CUDA Shared Memory support

CoderHam:
- Should SharedMemoryStatus return KIND?

### DIFF
--- a/src/core/grpc_service.proto
+++ b/src/core/grpc_service.proto
@@ -242,25 +242,82 @@ message SharedMemoryControlRequest
     //@@
     string name = 1;
 
-    //@@  .. cpp:var:: string shared_memory_key
     //@@
-    //@@     The name of the shared memory region that holds the input data
-    //@@     (or where the output data should be written).
+    //@@  .. cpp:var:: message SystemSharedMemoryIdentifier
     //@@
-    string shared_memory_key = 2;
+    //@@     The identifier for this system shared memory region.
+    //@@
+    message SystemSharedMemoryIdentifier
+    {
+      //@@  .. cpp:var:: string shared_memory_key
+      //@@
+      //@@     The name of the shared memory region that holds the input data
+      //@@     (or where the output data should be written).
+      //@@
+      string shared_memory_key = 2;
+    }
+
+    //@@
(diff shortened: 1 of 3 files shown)


src/core/grpc_service.proto
```diff
@@ -242,25 +242,82 @@ message SharedMemoryControlRequest
...
+      //@@     The offset from the start of the CUDA shared memory region.
+      //@@     start = offset, end = offset + size;
+      //@@
+      uint64 offset = 2;
+
+      //@@  .. cpp:var:: uint64 byte_size
+      //@@
+      //@@     Size of the CUDA shared memory block, in bytes.
+      //@@
+      uint64 byte_size = 3;
```
GuanLuo: Are the 'offset' and 'byte_size' here the same as those in the outer message? It seems to be a bit confusing without further explanation

src/servers/grpc_server.cc
```diff
@@ -1559,8 +1559,8 @@ SharedMemoryControlHandler::Process(Handler::State* state, bool rpc_ok)
     if (request.has_register_()) {
       err = smb_manager_->Create(
           &smb, request.register_().name(),
-          request.register_().shared_memory_key(), request.register_().offset(),
-          request.register_().byte_size());
+          request.register_().system_shared_memory().shared_memory_key(),
```
GuanLuo: For now, should add a check on 'system_shared_memory' is set, return unimplemented error if 'cuda_shared_memory' is set instead.

CoderHam: Thanks! Added a has_system_shared_memory check

src/core/grpc_service.proto
CoderHam: **offset**: The offset of the cudaIPC handle from the start of the shared memory region.
**byte_size**: Size of the cudaIPC handle in the shared memory block, in bytes.

The cudaIPC handle is stored in shared memory and the 'key' we used for System Shared memory instead must be a specific shared memory region that contains this cudaIPC handle in the case of the CUDA shared memory

deadeyegoodwin:
@CoderHam Did you run CI for this?

CoderHam:
Was waiting on your final review suggestions, ran it now.

PR #683, opened 2019-09-20
Fix HTTP inefficiency by using larger, contiguous buffers

deadeyegoodwin:


build/libevhtp/evhtp.patch
```diff
@@ -0,0 +1,42 @@
+diff --git a/evhtp.c b/evhtp.c
```
CoderHam: Did you add the patch file intentionally?

deadeyegoodwin: Yes. We must patch evhtp so that it collects the HTTP body into a contiguous memory buffer.

src/clients/c++/CMakeLists.txt
```diff
@@ -29,7 +29,8 @@ cmake_minimum_required (VERSION 3.5)
 #
 # librequest.so and librequest_static.a
 #
-find_package(CURL REQUIRED)
+find_package(CURL CONFIG REQUIRED)
```
GuanLuo: Since we are building libcurl, should we just remove libcurl installation in Dockerfile and update the documentation?

src/clients/c++/request_http.cc
```diff
@@ -1453,6 +1453,10 @@ InferHttpContextImpl::PreRunProcessing(std::shared_ptr<Request>& request)
     curl_easy_setopt(curl, CURLOPT_VERBOSE, 1L);
   }
 
+  const long buffer_byte_size = 16 * 1024 * 1024;
```
GuanLuo: Is this an empirical value?

src/clients/c++/CMakeLists.txt
deadeyegoodwin: libcurl is needed by GCS and perhaps ONNX. Maybe we could change those builds to instead use our version of libcurl, but that doesn't seem worth the effort.

deadeyegoodwin: But for the client container we should be able to remove libcurl-dev. I will try.

src/clients/c++/request_http.cc
deadeyegoodwin: No. But the maximum that curl allows is 2MB so I just wanted to be bigger than that.

PR #687, opened 2019-09-24
Update client timestamp collection

deadeyegoodwin:
Use server-style timestamp collection. Adjust timestamp collection to
be more accurate.

### DIFF
--- a/src/clients/c++/library/request_common.h
+++ b/src/clients/c++/library/request_common.h
@@ -28,6 +28,7 @@
 #include <chrono>
 #include <condition_variable>
 #include <deque>
+#include <limits>
 #include <list>
 #include <mutex>
 #include <string>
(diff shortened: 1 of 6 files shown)


src/clients/c++/library/request_common.h
```diff
@@ -40,61 +41,87 @@
...
+  /// \return The timestamp in nanoseconds.
+  uint64_t Timestamp(Kind kind) const { return timestamps_[(size_t)kind]; }
+
+  /// Set a timestamp to the current time, in nanoseconds.
+  /// \param kind The timestamp kind.
+  /// \return The timestamp in nanoseconds.
+  uint64_t CaptureTimestamp(Kind kind)
+  {
+    struct timespec spec;
+    clock_gettime(CLOCK_MONOTONIC, &spec);
```
GuanLuo: We used `<chrono>` previously because `timespec` etc. are POSIX functionalities, which is not portable to Windows. Need to make sure client library builds on Windows...

PR #688, opened 2019-09-24
Variable Shape tensors and String support for shared memory + Tests

CoderHam:
Add String and Variable size tensor support for I/O using shared memory
> Can create shared memory regions larger than size of the expected output


qa/L0_batcher/test.sh
```diff
@@ -169,67 +166,64 @@ for model_type in $types; do
...
-          RET=1
-      fi
-      set -e
-
-      unset TRTSERVER_DELAY_SCHEDULER
-      kill $SERVER_PID
-      wait $SERVER_PID
-  done
-fi
+export BATCHER_TYPE=VARIABLE
```
deadeyegoodwin: Is this putting it back to how it was before shared memory?

qa/L0_infer/infer_test.py
```diff
@@ -168,29 +168,31 @@ def test_raw_ihs(self):
...
+    def test_raw_iio(self):
+        self._full_exact(np.int32, np.int32, np_dtype_string,
+                         output0_raw=True, output1_raw=True, swap=False)
+    def test_raw_ioi(self):
+        self._full_exact(np.int32, np_dtype_string, np.int32,
+                         output0_raw=True, output1_raw=True, swap=False)
+
+    # shared memory does not support class output
+    if not TEST_SHARED_MEMORY:
         def test_class_bbb(self):
```
deadeyegoodwin: The indent for this and many following tests is wrong. Shouldn't it be one indent to the left.

qa/L0_sequence_batcher/test.sh
```diff
@@ -105,36 +105,30 @@ for m in \
...
-          $DATADIR/qa_ensemble_model_repository/qa_variable_sequence_model_repository/*_savedmodel_sequence_float32 \
-          $DATADIR/qa_variable_sequence_model_repository/libtorch_sequence_int32 ; do
-      cp -r $m modelsv/. && \
-          (cd modelsv/$(basename $m) && \
-              sed -i "s/^max_batch_size:.*/max_batch_size: 4/" config.pbtxt && \
-              sed -i "s/kind: KIND_GPU/kind: KIND_GPU\\ncount: 1/" config.pbtxt && \
-              sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 1/" config.pbtxt)
-  done
-fi
+# # Setup variable-size model repository for non shared memory case.
```
deadeyegoodwin: Remove these commented lines

src/clients/c++/library/request_common.cc
```diff
@@ -219,13 +219,15 @@ InputImpl::SetSharedMemory(
...
 
   // verify byte size of shared memory equals that of expected batch byte size
-  if (byte_size != (batch_size_ * byte_size_)) {
-    return Error(
-        RequestStatusCode::INVALID_ARG,
-        "The input '" + Name() + "' has shared memory of size " +
-            std::to_string(byte_size) + " bytes while the expected size is " +
-            std::to_string(batch_size_) + " * " + std::to_string(byte_size_) +
-            " = " + std::to_string(batch_size_ * byte_size_) + " bytes");
+  if (byte_size_ != -1) {
```
deadeyegoodwin: Use compound if here instead of two if statements

src/clients/python/__init__.py
```diff
@@ -987,6 +987,12 @@ def _prepare_request(self, inputs, outputs,
                              "' values must be specified as a list of numpy arrays" \
                              " or as a single c_void_p representing the shared memory handle")
             if type(inp) != c_void_p:
+                flag = False
```
deadeyegoodwin: What is this checking for? Put a comment. Also just have a compound if statement that has "continue" as the body.

src/clients/python/__init__.py
```diff
@@ -1085,6 +1091,16 @@ def _prepare_request(self, inputs, outputs,
                                         _crequest_infer_ctx_input_set_raw(
                                             input, input_value.ctypes.data_as(c_void_p),
                                             c_uint64(input_value.size * input_value.itemsize))))
+                    elif isinstance(input_values[1], (list, tuple)) and type(input_values[0]) == c_void_p:
```
deadeyegoodwin: What is this case? Add a comment. Use parens around sub-expressions in the if

src/clients/python/__init__.py
```diff
@@ -1218,31 +1234,45 @@ def _get_results(self, outputs, batch_size, request_id=None):
                         c_void_p(_crequest_get_shared_memory_handle_info(output_format[1], \
                                 byref(shm_addr), byref(shm_key), byref(shm_fd), \
                                 byref(offset), byref(byte_size))))
+                    if (np.prod(shape) * np.dtype(result_dtype).itemsize) < int(byte_size.value/batch_size):
```
deadeyegoodwin: What is this? The size of the returned result must exactly match what is expected from the result shape.

src/servers/grpc_server.cc
```diff
@@ -668,7 +668,7 @@ InferResponseAlloc(
         // If the output is in shared memory then just need to check
         // that the requested size matches what is expected by the
```
deadeyegoodwin: Fix comment

src/servers/http_server.cc
```diff
@@ -336,7 +336,7 @@ HTTPAPIServer::ResponseAlloc(
     auto pr = output_shm_map.find(tensor_name);
     if (pr != output_shm_map.end()) {
       // check for byte size mismatch
```
deadeyegoodwin: Fix comment and error string. Here and elsewhere.

src/servers/http_server.cc
```diff
@@ -813,7 +813,7 @@ HTTPAPIServer::EVBufferToInput(
       // correct and set input from the shared memory.
       if (io.has_shared_memory()) {
         LOG_VERBOSE(1) << io.name() << " has shared memory";
-        if (byte_size != io.shared_memory().byte_size()) {
+        if (byte_size > io.shared_memory().byte_size()) {
```
deadeyegoodwin: For input the buffer should be the exact size.

qa/L0_batcher/test.sh
CoderHam: Yes

src/clients/python/__init__.py
CoderHam: The output shared memory region may have a byte size of X but the size of the output not always be X, i.e. if the output shared memory region is too large we still want to allocate the appropriate number of byte size for each element in the batch to read.

src/clients/python/__init__.py
CoderHam: For variable size tensors, need the input shape as well as the shared memory handle

src/clients/python/__init__.py
CoderHam: Checks for the case in variable size tensors when (handle, tensor size) is passed

qa/L0_infer/infer_test.py
CoderHam: These tests are not needed for shared memory since they use CLASS/mixed output instead of RAW. Shared memory currently does not support class output.

src/clients/c++/library/request_common.cc
```diff
@@ -218,8 +218,9 @@ InputImpl::SetSharedMemory(
         "The input '" + Name() + "' can only be set once with SetSharedMemory");
   }
 
-  // verify byte size of shared memory equals that of expected batch byte size
-  if (byte_size != (batch_size_ * byte_size_)) {
+  // Verify byte size of shared memory equals that of expected batch byte size
+  // Skip check for string and variable size tensors
+  if ((byte_size_ != -1) && (byte_size > (batch_size_ * byte_size_))) {
```
deadeyegoodwin: This is for input so we should be able to have exact size testing. So shouldn't the ">" here be "!=" ?

src/clients/python/__init__.py
```diff
@@ -982,11 +982,13 @@ def _prepare_request(self, inputs, outputs,
         # An input's data may be specified as a list of numpy arrays,
         # or as a shared memory handle.
         for inp_name, inp in inputs.items():
-            if not isinstance(inp, (list, tuple)) and type(inp) != c_void_p:
+            if (not isinstance(inp, (list, tuple))) and (type(inp) != c_void_p):
                 _raise_error("input '" + inp_name +
                              "' values must be specified as a list of numpy arrays" \
                              " or as a single c_void_p representing the shared memory handle")
             if type(inp) != c_void_p:
+                if (len(inp) == 2) and (type(inp[0]) == c_void_p) and (isinstance(inp[1], (list, tuple))):
```
deadeyegoodwin: Put a comment here explaining that this case is that allows you to skip further processing for this input.

src/servers/grpc_server.cc
```diff
@@ -665,10 +665,10 @@ InferResponseAlloc(
...
-        // that the requested size matches what is expected by the
-        // request.
-        if (byte_size != pr->second.byte_size_) {
+        // If the output is in shared memory then check that the output byte
+        // size is equal to or less than the size of the requested registered
+        // shared memory region.
+        if (byte_size > pr->second.byte_size_) {
           return TRTSERVER_ErrorNew(
               TRTSERVER_ERROR_INTERNAL,
               std::string(
```
deadeyegoodwin: Below, "...expected requested buffer size to be at least ...

src/servers/http_server.cc
```diff
@@ -335,8 +335,10 @@ HTTPAPIServer::ResponseAlloc(
...
     if (pr != output_shm_map.end()) {
-      // check for byte size mismatch
-      if (byte_size != pr->second.second) {
+      // If byte size of output is greater than requested registered shared
+      // memory region then throw an error. Allow regions that are larger or
+      // equal to the byte size of output.
+      if (byte_size > pr->second.second) {
         return TRTSERVER_ErrorNew(
             TRTSERVER_ERROR_INTERNAL,
             std::string(
```
deadeyegoodwin: " ... expected buffer size to be at least ... "

src/clients/python/__init__.py

src/clients/c++/library/request_common.cc
CoderHam: Yes, fixed it in the other batcher fix branch, reverted it back here as well.

src/servers/grpc_server.cc
deadeyegoodwin: I meant that you should change the reported error in the lines below (line 675). You still need to change it.

src/servers/http_server.cc
deadeyegoodwin: Again here need to change text in line 344

CoderHam:
Understood, fixed it now

PR #689, opened 2019-09-25
Add TFTRT optimization path for Tensorflow backend

GuanLuo:


src/core/model_config.proto
```diff
@@ -448,27 +448,63 @@ message ModelOptimizationPolicy
   //@@
   message ExecutionAccelerators
```
deadeyegoodwin: The Graph message above with XLA should really be replaced with this interface. Can you file a ticket for that.

src/core/model_config.proto
```diff
@@ -448,27 +448,63 @@ message ModelOptimizationPolicy
...
+    }
+
+    //@@    .. cpp:var:: Accelerator gpu_execution_accelerator (repeated)
     //@@
     //@@       The preferred execution provider to be used if the model instance
     //@@       is deployed on GPU. The order priority implies priority, i.e. the
     //@@       provider at the front has highest priority.
     //@@
-    //@@       For ONNX Runtime backend, possible value is "tensorrt".
+    //@@       For ONNX Runtime backend, possible value is "tensorrt" as name,
```
deadeyegoodwin: ONNXruntime has no way to control the TRT parameters like TF? We should file a ticket with ONNXRuntime for that.

src/backends/tensorflow/base_backend.cc
```diff
@@ -191,11 +191,110 @@ BaseBackend::CreateExecutionContext(
...
+      }
+    }
+    for (const auto& io : Config().output()) {
+      const auto& dims = io.has_reshape() ? io.reshape().shape() : io.dims();
+      for (const auto& dim : dims) {
+        tftrt_config.is_dynamic_op_ |= (dim == -1);
+      }
+    }
+
+    // [TODO] Use non-default value.
```
deadeyegoodwin: Remove this TODO?

src/core/model_config.proto
GuanLuo: DLIS-765

src/core/model_config.proto
GuanLuo: DLIS-766 is the tracker ticket, will open an issue on ONNX Runtime later.

PR #694, opened 2019-09-26
Add more precise tracing for GRPC endpoint

deadeyegoodwin:


qa/common/trace_summary.py
```diff
@@ -50,54 +53,55 @@ def summarize(traces):
...
+            if ("http recv start" in timestamps) and ("http send end" in timestamps):
+                add_span(model_span_map[key], timestamps,
+                         "http infer", "http recv start", "http send end")
+                add_span(model_span_map[key], timestamps,
+                         "http recv", "http recv start", "http recv end")
+                add_span(model_span_map[key], timestamps,
+                         "http send", "http send start", "http send end")
+            elif ("grpc wait/read start" in timestamps) and ("grpc send end" in timestamps):
+                add_span(model_span_map[key], timestamps,
+                         "grpc infer", "grpc wait/read start", "grpc send end")
```
GuanLuo: `"grpc wait/read start"` is recorded when a GRPC request is created, so GRPC infer time here will be longer than it actually is. Do we need to point this out somewhere?

CoderHam:
Not related to changes from this PR but at [trace_summary.py#L166](https://github.com/NVIDIA/tensorrt-inference-server/blob/b2ff130faf0b0ea6d2353be2cd8d54678fb87559/qa/common/trace_summary.py#L166), the print statement does not use braces and may cause errors when running on python3.

deadeyegoodwin:
> Not related to changes from this PR but at [trace_summary.py#L166](https://github.com/NVIDIA/tensorrt-inference-server/blob/b2ff130faf0b0ea6d2353be2cd8d54678fb87559/qa/common/trace_summary.py#L166), the print statement does not use braces and may cause errors when running on python3.

I will fix

qa/common/trace_summary.py
deadeyegoodwin: Unfortunately, GRPC doesn't provide a hook where we can get just the request read timestamp. I will document the tracer in general and include this type of information.

PR #701, opened 2019-09-27
CUDA shared memory simple client

CoderHam:
- [x] Create C++ CUDA shared memory client example
- [x] Add tests for simple CUDA shared memory client examples

### DIFF
--- a/src/clients/c++/examples/CMakeLists.txt
+++ b/src/clients/c++/examples/CMakeLists.txt
@@ -63,6 +63,9 @@ install(
 # ensemble_image_client
 #
 add_executable(ensemble_image_client ensemble_image_client.cc)
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(ensemble_image_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   ensemble_image_client
   PRIVATE TRTIS::request_static
@@ -77,6 +80,9 @@ install(
 #
 find_package(OpenCV REQUIRED)
 add_executable(image_client image_client.cc)
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(image_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_include_directories(image_client PRIVATE ${OpenCV_INCLUDE_DIRS})
 target_link_libraries(
   image_client
@@ -92,6 +98,9 @@ install(
 # simple_callback_client
 #
 add_executable(simple_callback_client simple_callback_client.cc)
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(simple_callback_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   simple_callback_client
   PRIVATE TRTIS::request_static
@@ -108,6 +117,9 @@ add_executable(
   simple_client
   ${CMAKE_CURRENT_BINARY_DIR}/simple_client.cc
 )
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(simple_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   simple_client
   PRIVATE TRTIS::request_static
@@ -136,6 +148,9 @@ set(
 )
 add_executable(simple_shm_client simple_shm_client.cc
   ${SIMPLE_SHM_SRCS} ${SIMPLE_SHM_HDRS})
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(simple_shm_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   simple_shm_client
   PRIVATE request_static
@@ -148,29 +163,20 @@ install(
 #
 # simple_cuda_shm_client
 #
-if($(TRTIS_ENABLE_CUDA_EXAMPLES))
-  find_package(CUDA REQUIRED)
-  message(STATUS "Using CUDA ${CUDA_VERSION}")
-  set(CUDA_NVCC_FLAGS -std=c++11)
+if(${TRTIS_ENABLE_GPU})
   set(
     SIMPLE_CUDA_SHM_SRCS
     simple_cuda_shm_client.cc
-    shm_utils.cc
   )
   set(
     SIMPLE_CUDA_SHM_HDRS
-    shm_utils.h
   )
   add_executable(simple_cuda_shm_client simple_cuda_shm_client.cc
     ${SIMPLE_CUDA_SHM_SRCS} ${SIMPLE_CUDA_SHM_HDRS})
-  target_link_libraries(
-    simple_cuda_shm_client
-    PRIVATE request_static
-  )
-
   target_include_directories(simple_cuda_shm_client PRIVATE ${CUDA_INCLUDE_DIRS})
   target_link_libraries(
     simple_cuda_shm_client
+    PRIVATE request_static
     PRIVATE -L/usr/local/cuda/lib64/stubs
     PRIVATE -lnvidia-ml
     PRIVATE ${CUDA_LIBRARIES}
@@ -179,12 +185,15 @@ if($(TRTIS_ENABLE_CUDA_EXAMPLES))
     TARGETS simple_cuda_shm_client
     RUNTIME DESTINATION bin
   )
-endif() # TRTIS_ENABLE_CUDA_EXAMPLES
+endif() # TRTIS_ENABLE_GPU
 
 #
 # simple_sequence_client
 #
 add_executable(simple_sequence_client simple_sequence_client.cc)
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(simple_sequence_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   simple_sequence_client
   PRIVATE TRTIS::request_static
@@ -199,6 +208,9 @@ install(
 # simple_string_client
 #
 add_executable(simple_string_client simple_string_client.cc)
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(simple_string_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   simple_string_client
   PRIVATE TRTIS::request_static
@@ -212,6 +224,9 @@ install(
 # simple_perf_client
 #
 add_executable(simple_perf_client simple_perf_client.cc)
+if(${TRTIS_ENABLE_GPU})
+  target_include_directories(simple_perf_client PRIVATE ${CUDA_INCLUDE_DIRS})
+endif() # TRTIS_ENABLE_GPU
 target_link_libraries(
   simple_perf_client
   PRIVATE TRTIS::request_static
--- a/src/clients/c++/examples/CMakeLists.txt
+++ b/src/clients/c++/examples/CMakeLists.txt
@@ -148,7 +148,7 @@ install(
 #
 # simple_cuda_shm_client
 #
-#if($(TRTIS_ENABLE_CUDA_EXAMPLES))
+if($(TRTIS_ENABLE_CUDA_EXAMPLES))
   find_package(CUDA REQUIRED)
   message(STATUS "Using CUDA ${CUDA_VERSION}")
   set(CUDA_NVCC_FLAGS -std=c++11)
@@ -166,7 +166,6 @@ install(
   target_link_libraries(
     simple_cuda_shm_client
     PRIVATE request_static
-    rt
   )
 
   target_include_directories(simple_cuda_shm_client PRIVATE ${CUDA_INCLUDE_DIRS})
@@ -180,7 +179,7 @@ install(
     TARGETS simple_cuda_shm_client
     RUNTIME DESTINATION bin
   )
-#endif() # TRTIS_ENABLE_CUDA_EXAMPLES
+endif() # TRTIS_ENABLE_CUDA_EXAMPLES
 
 #
 # simple_sequence_client
--- a/src/clients/c++/examples/CMakeLists.txt
+++ b/src/clients/c++/examples/CMakeLists.txt
@@ -123,6 +123,38 @@ install(
 )
 
 #
+# shm_utils
+#
+if(TRTIS_ENABLE_CUDA_EXAMPLES)
+  find_package(CUDA REQUIRED)
+  message(STATUS "Using CUDA ${CUDA_VERSION}")
+  set(CUDA_NVCC_FLAGS -std=c++11)
+  add_library(shm_utils shm_utils.cc)
+  target_include_directories(shm_utils PRIVATE ${CUDA_INCLUDE_DIRS})
+  target_link_libraries(
+    shm_utils
+    PRIVATE -L/usr/local/cuda/lib64/stubs
+    PRIVATE -lnvidia-ml
+    PRIVATE ${CUDA_LIBRARIES}
+    rt
+  )
+  install(
+    TARGETS shm_utils
+    ARCHIVE DESTINATION lib
+  )
+else()
+  add_library(shm_utils shm_utils.cc)
+  target_link_libraries(
+    shm_utils
+    rt
+  )
+  install(
+    TARGETS shm_utils
+    ARCHIVE DESTINATION lib
+  )
+endif() #TRTIS_ENABLE_CUDA_EXAMPLES
+
+#
 # simple_shm_client
 #
 set(
@@ -138,8 +170,7 @@ add_executable(simple_shm_client simple_shm_client.cc
   ${SIMPLE_SHM_SRCS} ${SIMPLE_SHM_HDRS})
 target_link_libraries(
   simple_shm_client
-  PRIVATE TRTIS::request_static
-  rt
+  PRIVATE request_static
 )
 install(
   TARGETS simple_shm_client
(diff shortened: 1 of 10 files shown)


src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+      << "For -i, available protocols are 'grpc' and 'http'. Default is 'http."
+      << std::endl;
+
+  exit(1);
+}
+
+}  // namespace
+
+int
+CreateSharedMemoryRegion(std::string shm_key, size_t batch_byte_size)
```
GuanLuo: `batch_byte_size` is not accurate here. It is just some byte size that can hold cuda handle..

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+  checkCudaErrors(cudaMalloc((void**)&d_ptr, 16 * sizeof(int)));
+  checkCudaErrors(cudaMemcpy(
+      (void*)d_ptr, (void*)input_data, 16 * sizeof(int),
+      cudaMemcpyHostToDevice));
+
+  CreateCUDAIPCHandle(input0_cuda_rep, input0_d_ptr);
+  CreateCUDAIPCHandle(input1_cuda_rep, input1_d_ptr);
+
+  // Register Input shared memory with TRTIS
+  err = shared_memory_ctx->RegisterSharedMemory(
```
GuanLuo: Need to extend the API here or in option so that the server can know what type of shared memory it is working with as input / output

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+  err = shared_memory_ctx->UnregisterSharedMemory("output_data");
+  if (!err.IsOk()) {
+    std::cerr << "error: unable to unregister shared memory output region: "
+              << err << std::endl;
+    exit(1);
+  }
+
+  // cleanup cuda shared memory
+  checkCudaErrors(cudaIpcCloseMemHandle(input0_d_ptr));
+  checkCudaErrors(cudaIpcCloseMemHandle(input1_d_ptr));
```
GuanLuo: `cudaFree`?

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+  checkCudaErrors(cudaMalloc((void**)&d_ptr, 16 * sizeof(int)));
+  checkCudaErrors(cudaIpcGetMemHandle(
+      (cudaIpcMemHandle_t*)&shm_cuda_rep->memHandle, (void*)d_ptr));
+  checkCudaErrors(cudaMemcpy(
+      (void*)output_data, (void*)d_ptr, 16 * sizeof(int),
+      cudaMemcpyDeviceToHost));
+
+  // close IPC handle and free output data on GPU
+  checkCudaErrors(cudaIpcCloseMemHandle(d_ptr));
+  checkCudaErrors(cudaFree(d_ptr));
```
GuanLuo: Either change the function name to suggest that the handle will also be closed, or move this out to where it closes the input handles

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+  checkCudaErrors(
+      cudaEventCreate(&event, cudaEventDisableTiming | cudaEventInterprocess));
+  checkCudaErrors(cudaIpcGetEventHandle(
+      (cudaIpcEventHandle_t*)&shm_cuda_rep->eventHandle, event));
+  cudaIpcOpenMemHandle(
+      (void**)&d_ptr, *(cudaIpcMemHandle_t*)&shm_cuda_rep->memHandle,
+      cudaIpcMemLazyEnablePeerAccess)
+
+      // set device to default GPU - GPU0
+      checkCudaErrors(cudaSetDevice(DEVICE_ID));
```
GuanLuo: Indent. And it is necessary to set the device back to 0?

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+
+  // cleanup cuda shared memory
+  checkCudaErrors(cudaIpcCloseMemHandle(input0_d_ptr));
+  checkCudaErrors(cudaIpcCloseMemHandle(input1_d_ptr));
+
+  // Cleanup system shared memory
+  UnmapSharedMemory(input0_cuda_rep, cuda_ipc_byte_size * 2);
+  UnlinkSharedMemoryRegion("/input_simple");
+  UnmapSharedMemory(output0_cuda_rep, cuda_ipc_byte_size * 2);
+  UnlinkSharedMemoryRegion("/output_simple");
```
GuanLuo: missing `output1_cuda_rep`

src/clients/c++/examples/simple_cuda_shm_client.cc
deadeyegoodwin: This should be fixed in simple_shm_client.cc as well

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+                     "' to requested size " + std::to_string(batch_byte_size) +
+                     " bytes"
+              << std::endl;
+    exit(1);
+  }
+  return shm_fd;
+}
+
+void*
+MapSharedMemory(int shm_fd, size_t offset, size_t batch_byte_size)
```
deadeyegoodwin: Fix batch_byte_size

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,497 @@
...
+#include <sys/mman.h>
+#include <unistd.h>
+#include <iostream>
+#include <string>
+#include "src/clients/c++/library/request_grpc.h"
+#include "src/clients/c++/library/request_http.h"
+
+#include <cuda_runtime_api.h>
+// CUDA utilities and system includes
+#include <helper_cuda.h>
```
deadeyegoodwin: Where is helper_cuda.h from?

src/clients/c++/examples/simple_cuda_shm_client.cc
deadeyegoodwin: The sharedmemory functions are the same as in simple_shm_client.cc. Should make a shm_utils.cc and put these as well as the CUDA functions in there.

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Yes, adding a kind field.

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Used for testing, removed.

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Done

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Yes, it is needed

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Moving to where input handles are destoryed

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Not currently since there is only GPU 0 but it is needed in the case where there are multiple GPUs

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: output1_cuda_rep is deleted at the same time since output1_cuda_rep and output0_cuda_rep both reside in continuous regions in the same shared memory region. Similarly for input0_cuda_rep and input1_cuda_rep.

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Done, will move CUDA functions shortly as well

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Not needed anymore due to different Cuda Register function

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: The utils file is included by both system and Cuda shared memory clients. If this function is added to the utils file, CUDA libraries will be needed to build both clients.

qa/L0_simple_cuda_shared_memory_example/test.sh
```diff
@@ -0,0 +1,86 @@
...
+
+python $SIMPLE_CUDA_SHM_CLIENT_PY -v >>client_py.log 2>&1
+if [ $? -ne 0 ]; then
+    RET=1
+fi
+
+if [ `grep -c "localhost:8000" client_c++.log` != "10" ]; then
+    echo -e "\n***\n*** Failed. Expected 10 Host: localhost:8000 headers for C++ client\n***"
+    RET=1
+fi
```
GuanLuo: Why you need to check header here?

src/clients/c++/examples/CMakeLists.txt
```diff
@@ -125,17 +125,53 @@ install(
...
+set(
+  SIMPLE_SHM_SRCS
+  simple_shm_client.cc
+  shm_utils.cc
+)
+set(
+  SIMPLE_SHM_HDRS
+  shm_utils.h
+)
+add_executable(simple_shm_client simple_shm_client.cc
```
GuanLuo: drop `simple_shm_client.cc` as it is included in `${SIMPLE_SHM_SRCS}`

src/clients/c++/examples/CMakeLists.txt
```diff
@@ -125,17 +125,53 @@ install(
...
+  ${SIMPLE_SHM_SRCS} ${SIMPLE_SHM_HDRS})
 target_link_libraries(
   simple_shm_client
-  PRIVATE TRTIS::request_static
-  rt
+  PRIVATE request_static
```
GuanLuo: Why removing `TRTIS::` namespace?

src/clients/c++/examples/CMakeLists.txt
```diff
@@ -125,17 +125,53 @@ install(
...
+#
+if(${TRTIS_ENABLE_GPU})
+  set(
+    SIMPLE_CUDA_SHM_SRCS
+    simple_cuda_shm_client.cc
+  )
+  set(
+    SIMPLE_CUDA_SHM_HDRS
+  )
+  add_executable(simple_cuda_shm_client simple_cuda_shm_client.cc
```
GuanLuo: same

src/clients/c++/examples/CMakeLists.txt
```diff
@@ -125,17 +125,53 @@ install(
...
+    ${SIMPLE_CUDA_SHM_SRCS} ${SIMPLE_CUDA_SHM_HDRS})
+  target_include_directories(simple_cuda_shm_client PRIVATE ${CUDA_INCLUDE_DIRS})
+  target_link_libraries(
+    simple_cuda_shm_client
+    PRIVATE request_static
```
GuanLuo: same

src/clients/c++/examples/shm_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+CreateSharedMemoryRegion(std::string shm_key, size_t byte_size)
+{
+  // get shared memory region descriptor
+  int shm_fd = shm_open(shm_key.c_str(), O_RDWR | O_CREAT, S_IRUSR | S_IWUSR);
+  if (shm_fd == -1) {
+    std::cerr << "error: unable to get shared memory descriptor for "
+                 "shared-memory key '" +
+                     shm_key + "'"
+              << std::endl;
+    exit(1);
```
GuanLuo: Since you are making this a general utility, use error code or other indicator instead of exiting on error. Same for the rest

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,398 @@
...
+
+#define CudaRTCheck(FUNC, previous_device)                         \
+  {                                                                \
+    const cudaError_t result = FUNC;                               \
+    if (result != cudaSuccess) {                                   \
+      std::cout << "CUDA exception (line " << __LINE__             \
+                << "): " << cudaGetErrorName(result) << " ("       \
+                << cudaGetErrorString(result) << ")" << std::endl; \
+      cudaSetDevice(previous_device);                              \
+      exit(1);                                                     \
```
GuanLuo: Name the macro to `FAIL_IF_CUDA_ERR` to match what we have above.
Print through `std::cerr`. And you don't need to set device since you are going to exit right away.

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,398 @@
...
+{
+  int previous_device;
+  cudaGetDevice(&previous_device);
+  CudaRTCheck(cudaSetDevice(device_id), previous_device);
+
+  //  Create IPC handle for data on the gpu
+  CudaRTCheck(cudaIpcGetMemHandle(cuda_handle, input_d_ptr), previous_device);
+
+  // set device to previous GPU
+  cudaSetDevice(previous_device);
```
GuanLuo: Do you need to recover current device to the previous one?

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,398 @@
...
+  // Get the size of the inputs and outputs from the Shape and DataType
+  int input_byte_size =
+      infer_ctx->ByteSize(input0->Dims(), ni::DataType::TYPE_INT32);
+  int output_byte_size =
+      infer_ctx->ByteSize(output0->Dims(), ni::DataType::TYPE_INT32);
+
+  // Create Output0 and Output1 in CUDA Shared Memory
+  int *output0_d_ptr, *output1_d_ptr;
+  cudaMalloc((void**)&output0_d_ptr, output_byte_size * 2);
+  cudaMemset((void*)output0_d_ptr, 0, output_byte_size * 2);
```
GuanLuo: We don't need to initialize the output buffer as it will be overwritten anyway.

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,398 @@
...
+  CreateCUDAIPCHandle(&input_cuda_handle, (void*)input0_d_ptr);
+
+  // Register Input shared memory with TRTIS
+  err = shared_memory_ctx->RegisterCudaSharedMemory(
+      "input_data", input_cuda_handle, input_byte_size * 2, 0);
+  if (!err.IsOk()) {
+    std::cerr << "error: unable to register shared memory input region: " << err
+              << std::endl;
+    exit(1);
+  }
```
GuanLuo: A lot of this kind of error handling can be simplified with `FAIL_IF_ERR` macro

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,398 @@
...
+              << std::endl;
+  }
+
+  // Copy input and output data back to the CPU
+  int input0_data[16], input1_data[16];
+  int output0_data[16], output1_data[16];
+  cudaMemcpy(
+      input0_data, input0_d_ptr, input_byte_size, cudaMemcpyDeviceToHost);
+  cudaMemcpy(
+      input1_data, input1_d_ptr, input_byte_size, cudaMemcpyDeviceToHost);
```
GuanLuo: We already have the input data in system memory (`input_data`)

src/clients/c++/examples/simple_cuda_shm_client.cc
```diff
@@ -0,0 +1,398 @@
...
+  cudaError_t cuerr = cudaFree(input0_d_ptr);
+  if (cuerr != cudaSuccess) {
+    std::cerr << "CUDA free (input0_d_ptr) error: " << cudaGetErrorName(cuerr)
+              << "\n";
+  }
+  cuerr = cudaFree(output0_d_ptr);
+  if (cuerr != cudaSuccess) {
+    std::cerr << "CUDA free (output0_d_ptr) error: " << cudaGetErrorName(cuerr)
+              << "\n";
+  }
```
GuanLuo: CUDA error handling can be wrapped with the new CUDA macro.

src/clients/c++/examples/shm_utils.h
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2018-2019, NVIDIA CORPORATION. All rights reserved.
```
GuanLuo: Are the utils here similar to the one we have for Python shared memory library? If so, we can probably keep one instead of having duplicates.

qa/L0_simple_cuda_shared_memory_example/test.sh
CoderHam: Similar to the L0_simple_example this checks to see if all 10 HTTP requests made by the client are successful.

src/clients/c++/examples/CMakeLists.txt
CoderHam: Fixed

src/clients/c++/examples/CMakeLists.txt
CoderHam: Fixed

src/clients/c++/examples/CMakeLists.txt
CoderHam: Fixed

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Will do

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Removed

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Done

src/clients/c++/examples/shm_utils.h
CoderHam: Yes these are similar to the ones used in the Python client but both do slightly different things. The python side implements one function each for creation, setting and deletion and uses the shm handle. These functions are in some way the building blocks of the python shared memory client library

src/clients/c++/examples/shm_utils.cc
CoderHam: Done

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: No this was not necessary. Removed for now

src/clients/c++/examples/CMakeLists.txt
```diff
@@ -125,17 +125,51 @@ install(
...
+  )
+  set(
+    SIMPLE_CUDA_SHM_HDRS
+  )
+  add_executable(simple_cuda_shm_client ${SIMPLE_CUDA_SHM_SRCS} ${SIMPLE_CUDA_SHM_HDRS})
+  target_include_directories(simple_cuda_shm_client PRIVATE ${CUDA_INCLUDE_DIRS})
+  target_link_libraries(
+    simple_cuda_shm_client
+    PRIVATE TRTIS::request_static
+    PRIVATE -L/usr/local/cuda/lib64/stubs
```
deadeyegoodwin: I thought we found that we don't need stubs or nvidia-ml.

src/clients/c++/examples/simple_cuda_shm_client.cc
deadeyegoodwin: Was something removed? I still see the previous_device being get and restored?

src/clients/c++/examples/CMakeLists.txt
CoderHam: True. Removing this.

qa/L0_simple_cuda_shared_memory_example/test.sh
CoderHam: Removed from this PR

src/clients/c++/examples/simple_cuda_shm_client.cc
CoderHam: Fixed.

PR #702, opened 2019-09-28
Always load pytorch (libtorch) model in eval mode

CoderHam:


src/backends/pytorch/libtorch_backend.cc
```diff
@@ -254,6 +254,10 @@ LibTorchBackend::CreateExecutionContext(
     std::istringstream model_stream(lp_itr->second);
     context->torch_model_ = std::make_shared<torch::jit::script::Module>(
         torch::jit::load(model_stream, context->device_));
+
+    // load the torch model in eval mode (Solve PyTorch bug during saving model)
+    // https://github.com/pytorch/pytorch/issues/26884
```
GuanLuo: Does the linking issue here imply that this workaround can be removed once the Pytorch issue is resolved? If so, we should have a JIRA ticket to keep track of it.

CoderHam: Yes, we can remove this once the issue is resolved, will file a JIRA ticket for the same.

PR #709, opened 2019-09-30
Make CURL dependency PRIVATE in CMake

bencsikandrei:
It seems like CURL should be a private dep(?).
Even though that means nothing for a static library, at least
it makes it clearer that the interface used by librequest(_static)
is not tied to CURL in any way and its consumers don't need 
to see CURL headers.

Seems to me that CURL is not used in the public C++ API and is
hidden in code. It should be the same for CMake, since
it's the same for gRPC (although that seems to be linked
statically).

Signed-off-by: Andrei-Florin BENCSIK <andrei.bencsik@gmail.com>

### DIFF
--- a/src/clients/c++/examples/CMakeLists.example
+++ b/src/clients/c++/examples/CMakeLists.example
@@ -61,8 +61,8 @@ target_link_libraries(
   PRIVATE request_static
   PRIVATE gRPC::grpc++
   PRIVATE gRPC::grpc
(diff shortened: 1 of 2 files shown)


bencsikandrei:
I might be terribly wrong on this one but it seemes weird to have the CURL target (and all its include dirs and whatnot) propagated when I compiling and linking my application with librequest.so.

PR #718, opened 2019-10-02
cleaning the perclient cli arguments

tanmayv25:
The usage message will look like:
[Note: pasting here removed some indentation in the message]

> 
> Usage: ../install/bin/perf_client [options]
> ==== SYNOPSIS ====
>  
> 	-m <model name>
> 	-x <model version>
> 	-v
> 
> I. MEASUREMENT PARAMETERS: 
> 	--measurement-interval (-p) <measurement window (in msec)>
> 	--concurrency-range <start:end:step>
> 	--latency-threshold (-l) <latency threshold (in msec)>
> 	--max-threads <thread counts>
> 	--stability-percentage (-s) <deviation threshold for stable measurement (in percentage)>
> 	--max-trials (-r)  <maximum number of measurements for each profiling>
> 	--percentile <percentile>
> 	DEPRECATED OPTIONS
> 	-t <number of concurrent requests>
> 	-c <maximum concurrency>
> 	-d
> 	-a
> 
> II. INPUT DATA OPTIONS: 
> 	-b <batch size>
> 	--input-data <"zero"|"random"|<path>>
> 	--shape <name:shape>
> 	--sequence-length <length>
> 	DEPRECATED OPTIONS
> 	-z
> 	--data-directory <path>
> 
> III. SERVER DETAILS: 
> 	-u <URL for inference service>
> 	-i <Protocol used to communicate with inference service>
> 
> IV. OTHER OPTIONS: 
> 	-f <filename for storing report in csv format>
> 	-H <HTTP header>
> 	--streaming
> 
> ==== OPTIONS ==== 
>  
>  -m:     This is a required argument and is used to specify the model against
> 	 which to run perf_client.
>  -x:     The version of the above model to be used. If not specified the most
> 	 recent version (that is, the highest numbered version) of the model
> 	 will be used.
>  -v:     Enables the verbose mode.
> 
> I. MEASUREMENT PARAMETERS: 
>  --measurement-interval (-p): Indicates the time interval used for each
> 	 measurement in milliseconds. The perf client will sample a time interval
> 	 specified by -p and take measurement over the requests completed
> 	 within that time interval. The default value is 5000 msec.
>  --concurrency-range <start:end:step>: Determines the range of concurrency
> 	 levels covered by the perf_client. The perf_client will start from the
> 	 concurrency level of 'start' and go till 'end' with a stride of
> 	 'step'. The default value of 'end' and 'step' are 1. If 'end' is not
> 	 specified then perf_client will run for a single concurrency level
> 	 determined by 'start'. If 'end' is set as 0, then the concurrency limit
> 	 will be incremented by 'step' till latency threshold is met. 'end'
> 	 and --latency-threshold can not be both 0 simultaneously.
>  --latency-threshold (-l): Sets the limit on the observed latency. Client will
> 	 terminate the concurrency search once the measured latency exceeds
> 	 this threshold. By default, latency threshold is set 0 and the
> 	 perf_client will run for entire --concurrency-range.
>  --max-threads: Sets the maximum number of threads that will be created for
> 	 providing desired concurrency. Default is 16.
>  --stability-percentage (-s): Indicates the allowed variation in latency
> 	 measurements when determining if a result is stable. The measurement is
> 	 considered as stable if the recent 3 measurements are within +/-
> 	 (stability percentage)% of their average in terms of both infer per
> 	 second and latency. Default is 10(%).
>  --max-trials (-r): Indicates the maximum number of measurements for each
> 	 concurrency level visited during search. The perf client will take
> 	 multiple measurements and report the measurement until it is stable. The
> 	 perf client will abort if the measurement is still unstable after
> 	 the maximum number of measurements. The default value is 10.
>  --percentile: Indicates the confidence value as a percentile that will be
> 	 used to determine if a measurement is stable. For example, a value of
> 	 85 indicates that the 85th percentile latency will be used to
> 	 determine stability. The percentile will also be reported in the results.
> 	 The default is -1 indicating that the average latency is used to
> 	 determine stability
> 
> II. INPUT DATA OPTIONS: 
>  -b:     Batch size for each request sent.
>  --input-data: Select the type of data that will be used for input in
> 	 inference requests. The available options are "zero", "random" or path to a
> 	 directory. If the option is path to a directory then the directory
> 	 must contain a binary file for each input, named the same as the
> 	 input. Each file must contain the data required for that input for a
> 	 batch-1 request. Each file should contain the raw binary representation
> 	 of the input in row-major order. Default is "random".
>  --shape: The shape used for the specified input. The argument must be
> 	 specified as 'name:shape' where the shape is a comma-separated list for
> 	 dimension sizes, for example '--shape input_name:1,2,3' indicate tensor
> 	 shape [ 1, 2, 3 ]. --shape may be specified multiple times to
> 	 specify shapes for different inputs.
>  --sequence-length: Indicates the base length of a sequence used for sequence
> 	 models. A sequence with length x will be composed of x requests to
> 	 be sent as the elements in the sequence. The length of the actual
> 	 sequence will be within +/- 20% of the base length.
> 
> III. SERVER DETAILS: 
>  -u:     Specify URL to the server. Default is "localhost:8000" if using HTTP
> 	 and "localhost:8001" if using gRPC. 
>  -i:     The communication protocol to use. The available protocols are gRPC
> 	 and HTTP. Default is HTTP.
> 
> IV. OTHER OPTIONS: 
>  -f:     The latency report will be stored in the file named by this option.
> 	 By default, the result is not recorded in a file.
>  -H:     The header will be added to HTTP requests (ignored for GRPC
> 	 requests). The header must be specified as 'Header:Value'. -H may be
> 	 specified multiple times to add multiple headers.
>  --streaming: Enables the use of streaming API. This flag is only valid with
> 	 gRPC protocol. By default, it is set false.

### DIFF
--- a/qa/L0_perf_client/test.sh
+++ b/qa/L0_perf_client/test.sh
@@ -58,7 +58,7 @@ fi
 
 # Sanity check on measurements are not all zero
 set +e
-$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 -t 1 -p2000 -b 1 >$CLIENT_LOG 2>&1
+$PERF_CLIENT -v -i grpc -m graphdef_int32_int32_int32 -t 1 -p2000 -b 1 >$CLIENT_LOG 2>&1
 if [ $? -ne 0 ]; then
     cat $CLIENT_LOG
     echo -e "\n***\n*** Test Failed\n***"
@@ -75,7 +75,7 @@ set -e
 for MODEL in graphdef_nobatch_int32_int32_int32 graphdef_int32_int32_int32; do
     # Valid batch size
     set +e
-    $PERF_CLIENT -v -i grpc -u localhost:8001 -m $MODEL -t 1 -p20000 -b 1 >$CLIENT_LOG 2>&1
+    $PERF_CLIENT -v -i grpc -m $MODEL -t 1 -p20000 -b 1 >$CLIENT_LOG 2>&1
     if [ $? -ne 0 ]; then
         cat $CLIENT_LOG
         echo -e "\n***\n*** Test Failed\n***"
@@ -86,7 +86,7 @@ for MODEL in graphdef_nobatch_int32_int32_int32 graphdef_int32_int32_int32; do
     # Invalid batch sizes
     for STATIC_BATCH in 0 10; do
         set +e
-        $PERF_CLIENT -v -i grpc -u localhost:8001 -m $MODEL -t 1 -p20000 -b $STATIC_BATCH >$CLIENT_LOG 2>&1
+        $PERF_CLIENT -v -i grpc -m $MODEL -t 1 -p20000 -b $STATIC_BATCH >$CLIENT_LOG 2>&1
         if [ $? -eq 0 ]; then
             cat $CLIENT_LOG
             echo -e "\n***\n*** Test Failed\n***"
@@ -98,7 +98,7 @@ done
 
 # Testing with the new arguments
 set +e
-$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 >$CLIENT_LOG 2>&1
+$PERF_CLIENT -v -i grpc -m graphdef_int32_int32_int32 >$CLIENT_LOG 2>&1
 if [ $? -ne 0 ]; then
     cat $CLIENT_LOG
     echo -e "\n***\n*** Test Failed\n***"
@@ -112,7 +112,7 @@ fi
 set -e
 
 set +e
-$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 --search-range 1:5:2 >$CLIENT_LOG 2>&1
+$PERF_CLIENT -v -i grpc -m graphdef_int32_int32_int32 --concurrency-range 1:5:2 >$CLIENT_LOG 2>&1
 if [ $? -ne 0 ]; then
     cat $CLIENT_LOG
     echo -e "\n***\n*** Test Failed\n***"
--- a/qa/L0_perf_client/test.sh
+++ b/qa/L0_perf_client/test.sh
@@ -96,6 +96,36 @@ for MODEL in graphdef_nobatch_int32_int32_int32 graphdef_int32_int32_int32; do
     done
 done
 
+# Testing with the new arguments
+set +e
+$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 >$CLIENT_LOG 2>&1
+if [ $? -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+if [ $(cat $CLIENT_LOG | grep ": 0 infer/sec\|: 0 usec" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+set -e
+
+set +e
+$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 --search-range 1:5:2 >$CLIENT_LOG 2>&1
+if [ $? -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+if [ $(cat $CLIENT_LOG | grep ": 0 infer/sec\|: 0 usec|Request concurrency: 2" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+set -e
+
+
 kill $SERVER_PID
 wait $SERVER_PID
 
--- a/src/clients/c++/perf_client/perf_client.cc
+++ b/src/clients/c++/perf_client/perf_client.cc
@@ -79,16 +79,19 @@ SignalHandler(int signum)
 // There are two settings (see -d option) for the data collection:
 // - Fixed concurrent request mode:
 //     In this setting, the client will maintain a fixed number of concurrent
-//     requests sent to the server (see -t option). See ConcurrencyManager for
-//     more detail. The number of requests will be the total number of requests
-//     sent within the time interval for measurement (see -p option) and
-//     the latency will be the average latency across all requests.
+//     requests sent to the server (see --concurrency-range option). See
+//     ConcurrencyManager for more detail. The number of requests will be the
+//     total number of requests sent within the time interval for measurement
+//     (see --measurement-interval option) and the latency will be the average
+//     latency across all requests.
 //
 //     Besides throughput and latency, which is measured in client side,
 //     the following data measured by the server will also be reported
 //     in this setting:
 //     - Concurrent request: the number of concurrent requests as specified
-//         in -t option
+//         in --concurrency-range option. Note, for fixed concurrent request
+//         mode, user must specify --concurrency-range <'start'>,
+//         omitting 'end' and 'step' values.
 //     - Batch size: the batch size of each request as specified in -b option
 //     - Inference count: batch size * number of inference requests
 //     - Cumulative time: the total time between request received and
@@ -106,10 +109,13 @@ SignalHandler(int signum)
 // - Dynamic concurrent request mode:
 //     In this setting, the client will perform the following procedure:
 //       1. Follows the procedure in fixed concurrent request mode using
-//          k concurrent requests (k starts at 1).
+//          k concurrent requests (k starts at 'start').
 //       2. Gathers data reported from step 1.
-//       3. Increases k by 1 and repeats step 1 and 2 until latency from current
-//          iteration exceeds latency threshold (see -l option)
+//       3. Increases k by 'step' and repeats step 1 and 2 until latency from
+//          current iteration exceeds latency threshold (see --latency-threshold
+//          option) or concurrency level reaches 'end'. Note, by setting
+//          --latency-threshold or 'end' to 0 the effect of each threshold can
+//          be removed. However, both can not be 0 simultaneously.
 //     At each iteration, the data mentioned in fixed concurrent request mode
 //     will be reported. Besides that, after the procedure above, a collection
 //     of "throughput, latency, concurrent request count" tuples will be
@@ -117,11 +123,10 @@ SignalHandler(int signum)
 //
 // Options:
 // -b: batch size for each request sent.
-// -t: number of concurrent requests sent. If -d is set, -t indicate the number
-//     of concurrent requests to start with ("starting concurrency" level).
-// -d: enable dynamic concurrent request mode.
-// -l: latency threshold in msec, will have no effect if -d is not set.
-// -p: time interval for each measurement window in msec.
+// --concurrency-range: The range of concurrency levels perf_client will use.
+//     A concurrency level indicates the number of concurrent requests in queue.
+// --latency-threshold: latency threshold in msec.
+// --measurement-interval: time interval for each measurement window in msec.
 //
 // For detail of the options not listed, please refer to the usage.
 //
@@ -283,11 +288,11 @@ Usage(char** argv, const std::string& msg = std::string())
   std::cerr << "I. MEASUREMENT PARAMETERS: " << std::endl;
   std::cerr << "\t--measurement-interval (-p) <measurement window (in msec)>"
             << std::endl;
-  std::cerr << "\t--search-range <m:n:o>" << std::endl;
+  std::cerr << "\t--concurrency-range <start:end:step>" << std::endl;
   std::cerr << "\t--latency-threshold (-l) <latency threshold (in msec)>"
             << std::endl;
   std::cerr << "\t--max-threads <thread counts>" << std::endl;
-  std::cerr << "\t--stability-offset (-s) <deviation threshold for stable "
+  std::cerr << "\t--stability-percentage (-s) <deviation threshold for stable "
                "measurement (in percentage)>"
             << std::endl;
   std::cerr << "\t--max-trials (-r)  <maximum number of measurements for each "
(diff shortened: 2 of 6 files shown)


src/clients/c++/perf_client/perf_client.cc
```diff
@@ -427,6 +602,7 @@ main(int argc, char** argv)
         data_directory = optarg;
         break;
       case 5: {
+        using_search_range = true;
```
GuanLuo: should move to `case 7`?

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +277,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+  std::cerr << std::endl;
+  std::cerr << "IV. OTHER OPTIONS: " << std::endl;
+  std::cerr << "\t-f <filename for storing report in csv format>" << std::endl;
   std::cerr << "\t-H <HTTP header>" << std::endl;
-  std::cerr << "\t--sequence-length <length>" << std::endl;
-  std::cerr << "\t--percentile <percentile>" << std::endl;
-  std::cerr << "\t--shape <name:shape>" << std::endl;
-  std::cerr << "\t--data-directory <path>" << std::endl;
+  std::cerr << "\t--streaming" << std::endl;
+  std::cerr << "\t-a" << std::endl;
```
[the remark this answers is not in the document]
tanmayv25: See the description.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +277,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
-      << " within that time interval." << std::endl;
-  std::cerr << "For -r, it indicates the maximum number of measurements for "
-               "each profiling setting. The perf client will take multiple "
-               "measurements and report the measurement until it is stable. "
-               "The perf client will abort if the measurement is still "
-               "unstable after the maximum number of measuremnts."
+      << FormatMessage(
+             " --percentile: It indicates that the specified percentile in "
+             "terms of latency will also be reported and used to detemine if "
+             "the  measurement is stable instead of average latency. Default "
```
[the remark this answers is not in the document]
tanmayv25: done

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -251,6 +251,24 @@ Report(
...
+  int width = 60;
+  int index = 1;
+  int current_pos = offset;
+  while (current_pos + width < int(str.length())) {
+    int n = str.rfind(' ', current_pos + width);
+    if (n != int(std::string::npos)) {
+      str.replace(n, 1, "\n\t ");
+      current_pos += (width + 10);
+    }
+    index++;
```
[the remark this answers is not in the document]
tanmayv25: removed

qa/L0_perf_client/test.sh
```diff
@@ -96,6 +96,36 @@ for MODEL in graphdef_nobatch_int32_int32_int32 graphdef_int32_int32_int32; do
     done
 done
 
+# Testing with the new arguments
+set +e
+$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 >$CLIENT_LOG 2>&1
```
deadeyegoodwin: We should make the default -u value be based on the -i. So if -u is not specified explicitly the default for http should be localhost:8000 and the default for grpc should be localhost:8001

qa/L0_perf_client/test.sh
```diff
@@ -96,6 +96,36 @@ for MODEL in graphdef_nobatch_int32_int32_int32 graphdef_int32_int32_int32; do
...
+fi
+if [ $(cat $CLIENT_LOG | grep ": 0 infer/sec\|: 0 usec" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+set -e
+
+set +e
+$PERF_CLIENT -v -i grpc -u localhost:8001 -m graphdef_int32_int32_int32 --search-range 1:5:2 >$CLIENT_LOG 2>&1
```
deadeyegoodwin: Can we call --search-range to be --concurrency-range since that is what it is controlling.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
-  std::cerr << "\t-b <batch size>" << std::endl;
+  std::cerr << std::endl;
+  std::cerr << "I. MEASUREMENT PARAMETERS: " << std::endl;
+  std::cerr << "\t--measurement-interval (-p) <measurement window (in msec)>"
+            << std::endl;
+  std::cerr << "\t--search-range <m:n:o>" << std::endl;
+  std::cerr << "\t--latency-threshold (-l) <latency threshold (in msec)>"
+            << std::endl;
+  std::cerr << "\t--max-threads <thread counts>" << std::endl;
+  std::cerr << "\t--stability-offset (-s) <deviation threshold for stable "
```
deadeyegoodwin: I don't think "offset" is the correct term. Why not --stability-percentage

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   " version) of the model will be used.",
+                   9)
+            << std::endl;
+  std::cerr << std::setw(9) << std::left
+            << " -v: " << FormatMessage("Enables the verbose mode.", 9)
+            << std::endl;
+  std::cerr << std::endl;
+  std::cerr << "I. MEASUREMENT PARAMETERS: " << std::endl;
+  std::cerr << FormatMessage(
+                   " --measurement-interval (-p): It indicates the time "
```
deadeyegoodwin: Somewhere say that the value is specified in milliseconds.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
-      << "For -t, it indicates the number of starting concurrent requests if -d"
-      << " flag is set." << std::endl;
+      << FormatMessage(
+             " --search-range <m:n:o>: Determines the range of concurrency "
+             "levels covered by the perf_client. The perf_client will start "
+             "from the concurrency level of 'm' to 'n' with steps of 'o'. The "
+             "default value of 'm' and 'o' are 1. If 'n' is not specified then "
+             "perf_client will run for a single concurrency level determined "
+             "by 'm'. If 'n' is set as 0, then the concurrency limit will be "
+             "incremented by 'o' till latency threshold is met. 'n' and -l can "
```
deadeyegoodwin: Use --latency-threshold here instead of -l

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+             "not be both 0 simultaneously.",
+             18)
+      << std::endl;
+  std::cerr << FormatMessage(
+                   " --latency-threshold (-l): Sets the limit on the observed "
+                   "latency. Client will top incrementing the concurrency and "
```
deadeyegoodwin: "will stop"
".. and terminate the concurrency search ..."

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+             18)
       << std::endl;
   std::cerr
-      << "For -c, it indicates the maximum number of concurrent requests "
-         "allowed if -d flag is set. Once the number of concurrent requests "
-         "exceeds the maximum, the perf client will stop and exit regardless "
-         "of the latency threshold. Default is 0 to indicate that no limit is "
-         "set on the number of concurrent requests."
+      << FormatMessage(
+             " --stability-offset (-s): It indicates the deviation threshold "
```
deadeyegoodwin: Drop "it". "Indicates the deviation..."

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+             "for the measurements. The measurement is considered as stable if "
+             "the recent 3 measurements are within +/- (deviation threshold)% "
```
deadeyegoodwin: Don't say "deviation threshold" when the flag is called "stability percentage", use "stability percentage" in the description.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+             "of their average in terms of both infer per second and latency. "
+             "Default is 10(%).",
+             18)
       << std::endl;
+  std::cerr << FormatMessage(
+                   " --max-trials (-r): It indicates the maximum number of "
```
deadeyegoodwin: Drop "It"

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "measurements for each profiling setting. The perf client "
```
deadeyegoodwin: "profile setting" -> "each concurrency level visited during the search".

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
-      << " The perf client will sample a time interval specified by -p and"
-      << " take measurement over the requests completed"
-      << " within that time interval." << std::endl;
-  std::cerr << "For -r, it indicates the maximum number of measurements for "
-               "each profiling setting. The perf client will take multiple "
-               "measurements and report the measurement until it is stable. "
-               "The perf client will abort if the measurement is still "
-               "unstable after the maximum number of measuremnts."
+      << FormatMessage(
+             " --percentile: It indicates that the specified percentile in "
```
deadeyegoodwin: "Indicates the confidence value as a percentile that will be used to determine if a measurement is stable. For example, a value of 85 indicates that the 85th percentile latency will be used to determine stability. The percentile will also be reported in the results. The default is -1 indicating that the average latency is used to determine stability"

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "It indicates the number of starting concurrent requests if "
+                   "-d flag is set.",
+                   9)
             << std::endl;
-  std::cerr << "For -l, it has no effect unless -d flag is set." << std::endl;
   std::cerr
-      << "If -x is not specified the most recent version (that is, the highest "
-      << "numbered version) of the model will be used." << std::endl;
-  std::cerr << "For -i, available protocols are gRPC and HTTP. Default is HTTP."
+      << std::setw(9) << std::left << " -c: "
```
deadeyegoodwin: Perhaps we should not document the deprecated options to discourage their use.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "This flag is deprecated and will not affect client.", 9)
+            << std::endl;
+
+  std::cerr << std::endl;
+  std::cerr << "II. INPUT DATA OPTIONS: " << std::endl;
+  std::cerr << std::setw(9) << std::left
+            << " -b: " << FormatMessage("Batch size for each request sent.", 9)
+            << std::endl;
+  std::cerr << FormatMessage(
+                   " --input-data: This option can be used to select data, "
```
deadeyegoodwin: "Select the type of data that will be used for input in inference requests."

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "client will use in it's inference requests. The available "
+                   "options are \"zero\", \"random\" or path to a directory. "
+                   "If the option is path to a directory then there must be a "
```
deadeyegoodwin: "The directory must contain a binary file for each input, named the same as the input. Each file must contain the data required for that input for a batch-1 request. Each file should contain the raw binary representation of the input in row-major order. Default is \"random\"."

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "size. Note that the files should contain only the raw "
+                   "binary representation of the data in row major order. By "
+                   "default, the client will use random data in its requests.",
+                   18)
+            << std::endl;
+  std::cerr << FormatMessage(
+                   " --shape: The shape used for the specified input. The "
+                   "argument must be specified as 'name:shape' where the shape "
+                   "is a comma-separated list for dimension sizes, for example "
+                   "'--shape input_name:1,2,3'. --shape may be specified "
```
deadeyegoodwin: ...:1,2,3' indicate tensor shape [ 1, 2, 3 ].

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -259,142 +275,293 @@ Usage(char** argv, const std::string& msg = std::string())
...
+             "the raw binary representation of the data in row major order.",
+             18)
       << std::endl;
-  std::cerr << "The -z flag causes input tensors to be initialized with zeros "
-               "instead of random data"
+  std::cerr << std::endl;
+  std::cerr << "III. SERVER DETAILS: " << std::endl;
+  std::cerr << std::setw(9) << std::left << " -u: "
+            << FormatMessage(
+                   "This option is used to specify URL to the server. Default "
```
deadeyegoodwin: Using "it" or "this option" is redundant. Just say "Specify the URL ..."
Update to show the default for both http and grpc protocols

src/clients/c++/perf_client/perf_client.cc
deadeyegoodwin: top -> stop

src/clients/c++/perf_client/perf_client.cc
deadeyegoodwin: You still say "deviation threshold" which doesn't have any standard meaning. "Indicates the allowed variation in latency measurements when determining if a result is stable. The measurement is ... within +/- (stability percentage) ..."

src/clients/c++/perf_client/perf_client.cc
deadeyegoodwin: Still need this change.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -555,6 +749,28 @@ main(int argc, char** argv)
...
+
+  if (concurrency_range[CONCURRENCY_RANGE::kEND] == NO_LIMIT &&
+      latency_threshold_ms == NO_LIMIT) {
+    Usage(
+        argv,
+        "The end of the search range and the latency limit can not be both 0 "
+        "simultaneously");
+  }
+
+  if ((!url_specified) && protocol == perfclient::ProtocolType::GRPC) {
```
deadeyegoodwin: You don't need parens around the !url_specified but you should have parens around the "protocol == ...". In general always put parens are conditional sub-expressions to avoid any precedence ambiguity.

tanmayv25: fixed

PR #721, opened 2019-10-02
Improve error message when no GPU available

deadeyegoodwin:


src/core/model_config_utils.cc
```diff
@@ -503,10 +503,18 @@ ValidateModelConfig(
                 " has kind KIND_GPU but server does not support GPUs");
 #else
         if (group.gpus().size() == 0) {
-          return Status(
-              RequestStatusCode::INVALID_ARG,
-              "instance group " + group.name() + " of model " + config.name() +
-                  " has kind KIND_GPU but specifies no GPUs");
+          if (supported_gpus.size() == 0) {
```
GuanLuo: Shouldn't this check be outside of `if (group.gpus().size() == 0)`?

deadeyegoodwin: If there are no GPUs then group.gpus() will be empty at this point. If someone tries to set explicit GPUs on a system without GPUs it will fail earlier.

GuanLuo: I see, this will be hit if one only specifies KIND_GPU while the system has no GPU present

PR #738, opened 2019-10-08
Add model config option to enable GPU I/O for TensorFlow model

GuanLuo:


src/backends/tensorflow/tensorflow_backend_tf.cc
```diff
@@ -399,35 +501,82 @@ ModelImpl::~ModelImpl()
...
-  for (std::vector<tensorflow::Tensor>::reverse_iterator ri =
-           tfoutputs.rbegin();
-       ri != tfoutputs.rend(); ++ri) {
-    TRTISTF_Tensor* tensor =
-        reinterpret_cast<TRTISTF_Tensor*>(new TensorImpl(std::move(*ri)));
-    *output_tensors = TRTISTF_TensorListNew(tensor, *output_tensors);
+    // Have to sync device to ensure output tensors are ready,
+    // because the Callable is created with 'fetch_skip_sync = true'
+    // (false is not implemented).
+    device_->Sync();
```
deadeyegoodwin: This is potentially very bad for performance of not just this model but other things running on the GPU. Does the Run return before the outputs are computed?

src/core/model_config.proto
```diff
@@ -442,6 +442,27 @@ message ModelOptimizationPolicy
     bool graphs = 1;
   }
 
+  //@@
+  //@@  .. cpp:var:: message Tensorflow
+  //@@
+  //@@     TensorFlow-specific optimization settings.
+  //@@
+  message Tensorflow
```
deadeyegoodwin: Can we just use a new accelerator instead of making a TF specific message?

GuanLuo: This can be moved to accelerator section, perhaps in `gpu_execution_accelerator` with the name `gpu_io`?

deadeyegoodwin: Sounds good

src/backends/tensorflow/tensorflow_backend_tf.cc
GuanLuo: That's possible.. I am not sure whether the outputs are actually ready when `RunCallable()` returns, the `Device::Sync()` call is suggested by the [Callable doc](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto#L752), but I have also done some image classifications without the sync call and the client responses are consistent.

I can remove the sync and add on the test to verify the correctness on large payload (identity model with large input), if the output is consistent, then we can be confident to remove the sync call?

deadeyegoodwin: I don't know... but I assume that after the run call the results are sitting somewhere on the GPU. Hopefully the callable doesn't move them around async to the actual Run call. Or does the callable make the Run async somehow?

GuanLuo: Removed, and added inline comment on why

PR #739, opened 2019-10-09
Fix string serialization to tensor to allow null characters

CoderHam:
Fixes the TRTISTF_TensorSetString function to also use string length to include intermediate null characters.
PS: This does not have any test coverage Should I add the same.

### DIFF
--- a/src/backends/tensorflow/base_backend.cc
+++ b/src/backends/tensorflow/base_backend.cc
@@ -377,7 +377,7 @@ void
 FillStringTensor(TRTISTF_Tensor* tensor, const size_t idx, const size_t cnt)
 {
   for (size_t c = 0; c < cnt; ++c) {
-    TRTISTF_TensorSetString(tensor, idx + c, nullptr);
+    TRTISTF_TensorSetString(tensor, idx + c, nullptr, 0);
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/backends/tensorflow/base_backend.cc
```diff
@@ -609,7 +609,7 @@ BaseBackend::Context::SetStringInputTensor(
       content_byte_size -= len;
 
       TRTISTF_TensorSetString(
-          tensor, tensor_element_idx + element_idx, str.c_str());
+          tensor, tensor_element_idx + element_idx, str.c_str(), len);
```
deadeyegoodwin: Don't need 'str' anymore, just use "content, len" here.
And move lines 608 and 609 to be just before 613

CoderHam: Done

deadeyegoodwin:
@CoderHam Can you file a ticket for 19.11 to add a test case for string containing a null character. We should be able to leverage some test we already have.

CoderHam:
> @CoderHam Can you file a ticket for 19.11 to add a test case for string containing a null character. We should be able to leverage some test we already have.
Done

PR #740, opened 2019-10-09
Adding thread protection for the local data structures

tanmayv25:


src/clients/python/__init__.py
```diff
@@ -1501,6 +1504,7 @@ def async_run_with_cb(self, callback, inputs, outputs, batch_size=1, flags=0):
         wrapped_cb = partial(self._async_callback_wrapper, self._callback_resources_dict_id, callback)
         c_cb = _async_run_callback_prototype(wrapped_cb)
 
+        self._lock.acquire()
         # Run asynchronous inference...
```
deadeyegoodwin: Use a try .. finally to make sure you always release the lock. Here is an exception is raised below the lock won't be released.

src/clients/python/__init__.py
```diff
@@ -1548,10 +1553,11 @@ def get_async_run_results(self, request_id, wait):
         err = c_void_p(_crequest_infer_ctx_get_async_run_results(
             self._ctx, byref(c_is_ready), request_id, wait))
 
-
+        self._lock.acquire()
```
deadeyegoodwin: Use try...finally.

Why does this acquire need to be here. Can't it be moved down to just before "requested_outputs = ... "

src/clients/python/__init__.py
tanmayv25: using "with self._lock:" block to ensure the lock is released in all cases.

src/clients/python/__init__.py
tanmayv25: Yes, we can allow threads to race for updating _last_request_id. Fixed.

deadeyegoodwin:
@tanmayv25 you need to rerun CI.

PR #741, opened 2019-10-09
Feature/cmake minor improvements

bencsikandrei:
Some minor improvements and additions to the cmake build are made in 2 commits.

The first one is quite the small change: use Threads::Threads instead of relying on the if(NOT WIN32) check. This does the right thing behind the scenes and makes the CMake code the same on both WIN32 and Linux. Also, use the GNUInstallDirs from CMake to provide certain packages that also use it with the proper library directory on certain distros (eg. Fedora/CentOS use lib64 instead of lib). Also these resolve to an empty string on windows, which was the case before as well. There is still a long (or maybe not?) way to go until it compiles on these aforementioned operating systems.

Second, is a more complex change involving the addition of some cmake TrtisConfig/TrtisTargets configuration files. First, added targets in the namespace Trtis:: (i.e. request, request_static).
There is a chance that someone might use it as a subdirectory in cmake. Also I find it useful only for the client as I doubt anything similar would be done with the server, but who knows? If you wish to do the same for that I will add another commit. There is also a TrtisVersion file being generated and it uses the VERSION file present in the repo. Please correct me if this is not the best approach

I could always split these in smaller chunks if you find it would be better.


deadeyegoodwin:
@bencsikandrei Why did you close this?

deadeyegoodwin:
> @bencsikandrei Why did you close this?

I see, you have another PR open as well...

PR #743, opened 2019-10-09
Supporting ensemble with sequential models in perf_client

tanmayv25:


src/clients/c++/perf_client/context_factory.cc
```diff
@@ -59,6 +66,31 @@ ContextFactory::Create(
   return nic::Error::Success;
 }
 
+nic::Error
+ContextFactory::GetEnsembleSchedulerType(
+    const std::string& model_name, const int64_t model_version,
+    ni::ServerStatus& server_status, bool& is_sequential)
```
deadeyegoodwin: Pass server_status by const&. By coding convention don't use ref to return values, use a pointer. So "bool* is_sequential".

tanmayv25: fixed

PR #747, opened 2019-10-09
L0_perf -> L0_simple_perf

deadeyegoodwin:
Update to track nightly performance.

### DIFF
--- a/qa/L0_simple_perf/test.sh
+++ b/qa/L0_simple_perf/test.sh
@@ -36,11 +36,15 @@ if [ -z "$REPO_VERSION" ]; then
 fi
 
 CLIENT=../clients/simple_perf_client
+REPORTER=../common/reporter.py
+
+# The tensorrt identity model only allows variable size up to 32 so we
+# can't use it for the large tensor size runs we do here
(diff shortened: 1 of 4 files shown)


qa/L0_simple_perf/test.sh
```diff
@@ -36,6 +36,10 @@ if [ -z "$REPO_VERSION" ]; then
 fi
 
 CLIENT=../clients/simple_perf_client
+REPORTER=../common/reporter.py
+
+# The tensorrt identity model only allows variable size up to 32 so we
```
tanmayv25: 32 is the max shape we set in the optimization profile in our model generation script. Can't we add an optimization profile with kOPT shape set as 16384 for this test?

deadeyegoodwin: Potentially, if that doesn't interfere with other testing. We need a provide that allows very large input tensors (~ 1384 * 1024) for some additional testing that I will be adding for large tensors. Is that something we can add?

tanmayv25: I was able to build the plan identity model for 1417216 FP32 elements (0.045 GB) per tensor. However, there seems to be a limit for other reformat-free data types. As far as I know, It should not impact the other tests. We use add-sub models for range base testing.

PR #751, opened 2019-10-12
Preallocate and mange pinned memory pool to avoid (de)allocation overhead

GuanLuo:


src/core/pinned_memory_manager.cc
```diff
@@ -50,60 +51,102 @@ PointerToString(void* ptr)
...
+    : pinned_memory_buffer_(pinned_memory_buffer)
 {
-  if (instance_ == nullptr) {
-    return Status(
-        RequestStatusCode::UNAVAILABLE,
-        "PinnedMemoryManager has not been created");
+  if (pinned_memory_buffer_ != nullptr) {
+    managed_pinned_memory_ = boost::interprocess::managed_external_buffer(
+        boost::interprocess::create_only_t{}, pinned_memory_buffer_, size);
+    // Sanity check on whether we need to configure 'size_type'
```
deadeyegoodwin: What check is this comment referring to?

src/core/pinned_memory_manager.cc
```diff
@@ -50,60 +51,102 @@ PointerToString(void* ptr)
 
 std::unique_ptr<PinnedMemoryManager> PinnedMemoryManager::instance_;
 
+class PinnedMemoryManagerImpl : public PinnedMemoryManager {
```
deadeyegoodwin: Why did you make a Impl class instead of just directly implementing PinnedMemoryManager?

src/core/pinned_memory_manager.cc
```diff
@@ -50,60 +51,102 @@ PointerToString(void* ptr)
...
+
+  Status AllocInternal(
+      void** ptr, uint64_t size,
+      bool allow_nonpinned_fallback = false) override;
+  Status FreeInternal(void* ptr) override;
+
+ private:
+  PinnedMemoryManagerImpl(void* pinned_memory_buffer, uint64_t size);
+  std::mutex info_mtx_;
+  std::map<void*, std::pair<bool, uint64_t>> memory_info_;
```
deadeyegoodwin: Can this just be a map from addr -> bool? I don't see the size used anywhere

src/core/pinned_memory_manager.cc
GuanLuo: This check can be removed. This is for debugging purpose, Boost uses 'size_type' which is obscured to me, so print the size seen by Boost to make sure it is larger than 32 bits.

src/core/pinned_memory_manager.cc
GuanLuo: There is no strong intention to do this. I was thinking that then a different implementation, if needed, can be added easily.

PR #765, opened 2019-10-16
Adjust grpc defaults for improved performance

deadeyegoodwin:


src/servers/main.cc
```diff
@@ -106,18 +106,17 @@ int32_t trace_rate_ = 1000;
...
+int grpc_stream_infer_thread_cnt_ = 1;
 
 // The maximum number of inference request/response objects that
 // remain allocated for reuse. As long as the number of in-flight
 // requests doesn't exceed this value there will be no
-// allocation/deallocation of request/response objects. Higher values
-// trade-off increased memory usage for higher performance.
-int grpc_infer_allocation_pool_size_ = 128;
+// allocation/deallocation of request/response objects.
+int grpc_infer_allocation_pool_size_ = 0;
```
GuanLuo: So reusing the inference object actually has negative impact on performance?

src/servers/grpc_server.cc
```diff
@@ -340,14 +338,17 @@ class Handler : public GRPCServer::HandlerBase {
 
   void StateRelease(State* state)
   {
-    std::unique_lock<std::mutex> lock(alloc_mu_);
+    if (max_state_bucket_count_ > 0) {
+      std::unique_lock<std::mutex> lock(alloc_mu_);
```
GuanLuo: Can just use `lock_guard` for scoped lock

src/servers/main.cc
deadeyegoodwin: For the case I'm testing. Maybe not for request with large input/output tensors. We may need to do something dynamic based on the size of the message.

src/servers/grpc_server.cc
deadeyegoodwin: Fixed

PR #774, opened 2019-10-18
 Expose more options for TF-TRT optimization

GuanLuo:


src/core/model_config.proto
```diff
@@ -496,6 +496,18 @@ message ModelOptimizationPolicy
     //@@         The value can be one of "FP32", "FP16" and "INT8".
     //@@         Default value is "FP32".
     //@@
+    //@@         "use_calibration" Whether to perform INT8 quantization with
+    //@@         TensorRT calibration. If set to False, quantization nodes are
+    //@@         expected for every tensor in the graph. Default value is true.
+    //@@         This argument is ignored if 'precision_mode' is not 'INT8'.
+    //@@
+    //@@         "is_dynamic_op" Whether at least one input has dynamic shape.
```
deadeyegoodwin: If we infer this value from the model shape why does it need to be specified explicitly? Perhaps the description is incorrect?

GuanLuo: The edge case that I am considering is that the model input is dynamic shape but the user specify it as fixed-size tensor in model config. In that case, `is_dynamic_op` should be true but inferring from model config says false.

deadeyegoodwin: OK to keep it then

src/backends/tensorflow/base_backend.cc
```diff
@@ -237,48 +239,47 @@ BaseBackend::CreateExecutionContext(
...
+                &tftrt_config.is_dynamic_op_));
           } else {
             return Status(
                 RequestStatusCode::INVALID_ARG,
                 "unknown parameter '" + parameter.first +
                     "' is provided for TensorRT Execution Accelerator");
           }
         }
+        // If needed, set 'use_calibration' at hindsight
+        if (!is_calibration_specified) {
```
deadeyegoodwin: Remove this and other calibration-related settings

PR #778, opened 2019-10-22
Support 'memory_type_id' along with 'memory_type'

GuanLuo:
Major change is let TRTIS internal be aware of 'memory_type_id', this is done by tying it with all occurrences of 'memory_type'.

Fix custom backend implementation to match description (return acutal memory type and type id)

### DIFF
--- a/src/core/provider.cc
+++ b/src/core/provider.cc
@@ -48,24 +48,28 @@ SystemMemoryReference::SystemMemoryReference() : SystemMemory() {}
 
 const char*
 SystemMemoryReference::BufferAt(
-    size_t idx, size_t* byte_size, TRTSERVER_Memory_Type* memory_type) const
+    size_t idx, size_t* byte_size, TRTSERVER_Memory_Type* memory_type,
+    int64_t* memory_type_id) const
 {
   if (idx >= buffer_.size()) {
     *byte_size = 0;
     *memory_type = TRTSERVER_MEMORY_CPU;
+    *memory_type_id = 0;
     return nullptr;
   }
-  *memory_type = std::get<2>(buffer_[idx]);
-  *byte_size = std::get<1>(buffer_[idx]);
-  return std::get<0>(buffer_[idx]);
+  *memory_type = buffer_[idx].memory_type_;
+  *memory_type_id = buffer_[idx].memory_type_id_;
+  *byte_size = buffer_[idx].byte_size_;
+  return buffer_[idx].buffer_;
 }
 
 size_t
 SystemMemoryReference::AddBuffer(
-    const char* buffer, size_t byte_size, TRTSERVER_Memory_Type memory_type)
+    const char* buffer, size_t byte_size, TRTSERVER_Memory_Type memory_type,
+    int64_t memory_type_id)
 {
   total_byte_size_ += byte_size;
-  buffer_.emplace_back(std::make_tuple(buffer, byte_size, memory_type));
+  buffer_.emplace_back(buffer, byte_size, memory_type, memory_type_id);
   return buffer_.size() - 1;
 }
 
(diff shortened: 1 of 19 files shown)


src/core/provider.cc
```diff
@@ -84,7 +88,10 @@ AllocatedSystemMemory::AllocatedSystemMemory(
       }
     } else {
 #ifdef TRTIS_ENABLE_GPU
-      cudaError_t err = cudaMalloc((void**)&buffer_, byte_size);
+      auto err = cudaSetDevice(memory_type_id_);
```
deadeyegoodwin: Are callers of AllocatedSystemMemory all ok with the cuda device changing? I think so but we should check to make sure...

src/servers/simple.cc
```diff
@@ -548,10 +551,11 @@ main(int argc, char** argv)
   const void* output0_content;
   size_t output0_byte_size;
   TRTSERVER_Memory_Type output0_memory_type;
+  int64_t memory_type_id;
```
deadeyegoodwin: output0_memory_type_id

src/servers/simple.cc
```diff
@@ -573,7 +578,7 @@ main(int argc, char** argv)
   FAIL_IF_ERR(
       TRTSERVER_InferenceResponseOutputData(
           response, output1->name().c_str(), &output1_content,
-          &output1_byte_size, &output1_memory_type),
+          &output1_byte_size, &output1_memory_type, &memory_type_id),
```
deadeyegoodwin: output1_memory_type_id

src/servers/simple.cc
```diff
@@ -111,11 +111,25 @@ ResponseAlloc(
       allocated_ptr = malloc(byte_size);
 #ifdef TRTIS_ENABLE_GPU
     } else if (use_gpu_memory) {
-      auto err = cudaMalloc(&allocated_ptr, byte_size);
+      int current_device;
+      auto err = cudaGetDevice(&current_device);
```
deadeyegoodwin: We should not push this requirement into the user callback code. Instead in the provider where we invoke this callback we should remember the current device and set it back after returning from the callback. I think get and set take about the same time so we should just set the device on return and not bother to check if it has changed.

src/servers/simple.cc
```diff
@@ -150,11 +164,25 @@ ResponseRelease(
     free(buffer);
 #ifdef TRTIS_ENABLE_GPU
   } else if (use_gpu_memory) {
-    auto err = cudaFree(buffer);
+    int current_device;
+    auto err = cudaGetDevice(&current_device);
```
deadeyegoodwin: Here again... they should be able to just set the device and cudaFree...

src/core/provider.cc
```diff
@@ -698,15 +744,39 @@ InferResponseProvider::AllocateOutputBuffer(
...
+#endif  // TRTIS_ENABLE_GPU
   TRTSERVER_Error* err = alloc_fn_(
       allocator_, &buffer, &buffer_userp, name.c_str(), alloc_byte_size,
-      preferred_memory_type, 0 /* region_id */, alloc_userp_);
+      preferred_memory_type, preferred_memory_type_id, alloc_userp_);
+  Status status;
+#ifdef TRTIS_ENABLE_GPU
+  // [TODO] change to check actual memory type after
+  // https://github.com/NVIDIA/tensorrt-inference-server/pull/780
+  if (preferred_memory_type == TRTSERVER_MEMORY_GPU) {
```
deadeyegoodwin: I don't think we should actually guard this with preferred_memory_type since we don't know what happened during the alloc function. But it is ok for now since I think this code will be changing due to callback changes for mem_type_id.

GuanLuo: After some thinking, even after the allocation function improvement, we still can't rely on the actual memory type returned by the allocation function, because we should treat it as a black box and then the returned memory type doesn't necessarily tell whether the current device has been changed.

I think we should just always call cudaGet/SetDevice() regardless of their returned error code.

src/core/provider.cc
```diff
@@ -617,7 +661,7 @@ InferResponseProvider::~InferResponseProvider()
     if (output.release_buffer_ != nullptr) {
       TRTSERVER_Error* err = release_fn_(
```
deadeyegoodwin: Here we also need to get/restore cuda device in the same was as with alloc_fn_.

PR #780, opened 2019-10-22
Fix Response Allocation

CoderHam:
- [x] keep track of memory type for output shared memory 
- [x] Call response allocation only once and return actual_memory_type (and actual_device_id) of where the output was finally allocated/resides.

- [x] Temporarily remove custom plugin from test due to unexpected behavior.

### DIFF
--- a/src/backends/custom/custom_backend.cc
+++ b/src/backends/custom/custom_backend.cc
@@ -586,9 +586,10 @@ CustomBackend::Context::GetOutput(
 
     auto dst_memory_type = ToTRTServerMemoryType(*memory_type);
     auto actual_memory_type = dst_memory_type;
+    int64_t device_id;
     Status status = payload->response_provider_->AllocateOutputBuffer(
         name, content, content_byte_size, shape, dst_memory_type,
-        *memory_type_id, &actual_memory_type);
+        *memory_type_id, &actual_memory_type, &device_id);
 
     // Done with this output if 'content_byte_size' is 0
     if (content_byte_size == 0) {
--- a/src/backends/custom/custom_backend.cc
+++ b/src/backends/custom/custom_backend.cc
@@ -583,28 +583,23 @@ CustomBackend::Context::GetOutput(
     if (shape_dim_cnt > 0) {
       shape.assign(shape_dims, shape_dims + shape_dim_cnt);
     }
+
+    auto dst_memory_type = ToTRTServerMemoryType(*memory_type);
+    auto actual_memory_type = dst_memory_type;
     Status status = payload->response_provider_->AllocateOutputBuffer(
-        name, content, content_byte_size, shape,
-        ToTRTServerMemoryType(*memory_type), *memory_type_id);
+        name, content, content_byte_size, shape, dst_memory_type,
+        *memory_type_id, &actual_memory_type);
 
     // Done with this output if 'content_byte_size' is 0
     if (content_byte_size == 0) {
       *content = nullptr;
     } else if (*content == nullptr) {
-      // If first attempt is CPU and failed, then allocation failed
-      if ((!status.IsOk()) || (*memory_type == CUSTOM_MEMORY_CPU)) {
-        return false;
-      }
-
-      *memory_type = CUSTOM_MEMORY_CPU;
-      *memory_type_id = 0;
-      status = payload->response_provider_->AllocateOutputBuffer(
-          name, content, content_byte_size, shape,
-          ToTRTServerMemoryType(*memory_type), *memory_type_id);
-      if (*content == nullptr) {
-        return false;
-      }
+      return false;
     }
+
+    // Update memory type with actual memory type
+    *memory_type = ToCustomMemoryType(actual_memory_type);
+
     return status.IsOk();
   }
 
--- a/src/backends/custom/custom_backend.cc
+++ b/src/backends/custom/custom_backend.cc
@@ -586,10 +586,10 @@ CustomBackend::Context::GetOutput(
 
     auto dst_memory_type = ToTRTServerMemoryType(*memory_type);
     auto actual_memory_type = dst_memory_type;
-    int64_t device_id;
+    int64_t actual_memory_type_id;
     Status status = payload->response_provider_->AllocateOutputBuffer(
         name, content, content_byte_size, shape, dst_memory_type,
-        *memory_type_id, &actual_memory_type, &device_id);
+        *memory_type_id, &actual_memory_type, &actual_memory_type_id);
 
     // Done with this output if 'content_byte_size' is 0
     if (content_byte_size == 0) {
--- a/src/backends/custom/custom_backend.cc
+++ b/src/backends/custom/custom_backend.cc
@@ -584,12 +584,12 @@ CustomBackend::Context::GetOutput(
       shape.assign(shape_dims, shape_dims + shape_dim_cnt);
     }
 
-    auto dst_memory_type = ToTRTServerMemoryType(*memory_type);
-    auto actual_memory_type = dst_memory_type;
+    TRTSERVER_Memory_Type actual_memory_type;
     int64_t actual_memory_type_id;
     Status status = payload->response_provider_->AllocateOutputBuffer(
-        name, content, content_byte_size, shape, dst_memory_type,
-        *memory_type_id, &actual_memory_type, &actual_memory_type_id);
+        name, content, content_byte_size, shape,
+        ToTRTServerMemoryType(*memory_type), *memory_type_id,
+        &actual_memory_type, &actual_memory_type_id);
 
     // Done with this output if 'content_byte_size' is 0
     if (content_byte_size == 0) {
@@ -600,6 +600,7 @@ CustomBackend::Context::GetOutput(
 
     // Update memory type with actual memory type
     *memory_type = ToCustomMemoryType(actual_memory_type);
+    *memory_type_id = actual_memory_type_id;
 
     return status.IsOk();
   }
--- a/src/backends/onnx/onnx_backend.cc
+++ b/src/backends/onnx/onnx_backend.cc
@@ -883,8 +883,11 @@ OnnxBackend::Context::SetStringOutputBuffer(
           data_byte_size + sizeof(uint32_t) * expected_element_cnt;
 
       void* buffer;
+      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
+      TRTSERVER_Memory_Type actual_memory_type;
+      int64_t device_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
-          name, &buffer, expected_byte_size, content_shape);
+          name, &buffer, expected_byte_size, content_shape, preferred_memory_type, &actual_memory_type, &device_id);
       if (status.IsOk()) {
         size_t copied_byte_size = 0;
         for (size_t e = 0; e < expected_element_cnt; ++e) {
--- a/src/backends/onnx/onnx_backend.cc
+++ b/src/backends/onnx/onnx_backend.cc
@@ -887,20 +887,57 @@ OnnxBackend::Context::SetStringOutputBuffer(
       TRTSERVER_Memory_Type actual_memory_type;
       int64_t device_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
-          name, &buffer, expected_byte_size, content_shape, preferred_memory_type, &actual_memory_type, &device_id);
+          name, &buffer, expected_byte_size, content_shape,
+          preferred_memory_type, &actual_memory_type, &device_id);
       if (status.IsOk()) {
         size_t copied_byte_size = 0;
         for (size_t e = 0; e < expected_element_cnt; ++e) {
           const uint32_t len =
               offsets[element_idx + e + 1] - offsets[element_idx + e];
-          memcpy(
-              static_cast<char*>(buffer) + copied_byte_size,
-              static_cast<const void*>(&len), sizeof(uint32_t));
+          if (actual_memory_type == TRTSERVER_MEMORY_GPU) {
+#ifdef TRTIS_ENABLE_GPU
+            cudaError_t err = cudaMemcpyAsync(
+                static_cast<char*>(buffer) + copied_byte_size,
+                static_cast<const void*>(&len), sizeof(uint32_t),
+                cudaMemcpyDeviceToHost, stream_);
+            if (err != cudaSuccess) {
+              payload.status_ = Status(
+                  RequestStatusCode::INTERNAL,
+                  "output '" + name + "' is in GPU memory and failed to use" +
+                      " CUDA copy from CPU memory to get String output data: " +
+                      std::string(cudaGetErrorString(err)));
+            } else {
+              cudaStreamSynchronize(stream_);
+            }
+#endif  // TRTIS_ENABLE_GPU
+          } else {
+            memcpy(
+                static_cast<char*>(buffer) + copied_byte_size,
+                static_cast<const void*>(&len), sizeof(uint32_t));
+          }
           copied_byte_size += sizeof(uint32_t);
 
-          memcpy(
-              static_cast<char*>(buffer) + copied_byte_size,
-              content + offsets[element_idx + e], len);
+          if (actual_memory_type == TRTSERVER_MEMORY_GPU) {
+#ifdef TRTIS_ENABLE_GPU
+            cudaError_t err = cudaMemcpyAsync(
+                static_cast<char*>(buffer) + copied_byte_size,
+                content + offsets[element_idx + e], len, cudaMemcpyDeviceToHost,
+                stream_);
+            if (err != cudaSuccess) {
+              payload.status_ = Status(
+                  RequestStatusCode::INTERNAL,
+                  "output '" + name + "' is in GPU memory and failed to use" +
+                      " CUDA copy from CPU memory to get String output data: " +
+                      std::string(cudaGetErrorString(err)));
+            } else {
+              cudaStreamSynchronize(stream_);
+            }
+#endif  // TRTIS_ENABLE_GPU
+          } else {
+            memcpy(
+                static_cast<char*>(buffer) + copied_byte_size,
+                content + offsets[element_idx + e], len);
+          }
           copied_byte_size += len;
         }
       } else {
--- a/src/backends/onnx/onnx_backend.cc
+++ b/src/backends/onnx/onnx_backend.cc
@@ -888,7 +888,7 @@ OnnxBackend::Context::SetStringOutputBuffer(
       int64_t device_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
           name, &buffer, expected_byte_size, content_shape,
-          preferred_memory_type, &actual_memory_type, &device_id);
+          preferred_memory_type, 0, &actual_memory_type, &device_id);
       if (status.IsOk()) {
         size_t copied_byte_size = 0;
         for (size_t e = 0; e < expected_element_cnt; ++e) {
--- a/src/backends/onnx/onnx_backend.cc
+++ b/src/backends/onnx/onnx_backend.cc
@@ -817,7 +817,7 @@ OnnxBackend::Context::ReadOutputTensors(
       // Mark "passed end byte offset"
       offsets[element_count] = total_length;
 
-      SetStringOutputBuffer(
+      cuda_copy |= SetStringOutputBuffer(
           name, batch1_element_cnt, content, content_shape, offsets, payloads);
     } else {
       // Fixed size data type...
@@ -858,13 +858,14 @@ OnnxBackend::Context::ReadOutputTensors(
   return Status::Success;
 }
 
-void
+bool
 OnnxBackend::Context::SetStringOutputBuffer(
     const std::string& name, const size_t batch1_element_cnt,
     const char* content, const std::vector<int64_t>& content_shape,
     const size_t* offsets, std::vector<Scheduler::Payload>* payloads)
 {
   size_t element_idx = 0;
+  bool cuda_copy = false;
   for (auto& payload : *payloads) {
     const InferRequestHeader& request_header =
         payload.request_provider_->RequestHeader();
@@ -885,10 +886,11 @@ OnnxBackend::Context::SetStringOutputBuffer(
       void* buffer;
       TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
       TRTSERVER_Memory_Type actual_memory_type;
-      int64_t device_id;
+      int64_t actual_memory_type_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
           name, &buffer, expected_byte_size, content_shape,
-          preferred_memory_type, 0, &actual_memory_type, &device_id);
+          preferred_memory_type, 0 /* preferred_memory_type_id */,
+          &actual_memory_type, &actual_memory_type_id);
       if (status.IsOk()) {
         size_t copied_byte_size = 0;
         for (size_t e = 0; e < expected_element_cnt; ++e) {
@@ -907,7 +909,7 @@ OnnxBackend::Context::SetStringOutputBuffer(
                       " CUDA copy from CPU memory to get String output data: " +
                       std::string(cudaGetErrorString(err)));
             } else {
-              cudaStreamSynchronize(stream_);
+              cuda_copy = true;
             }
 #endif  // TRTIS_ENABLE_GPU
           } else {
@@ -930,7 +932,7 @@ OnnxBackend::Context::SetStringOutputBuffer(
                       " CUDA copy from CPU memory to get String output data: " +
                       std::string(cudaGetErrorString(err)));
             } else {
-              cudaStreamSynchronize(stream_);
+              cuda_copy = true;
             }
 #endif  // TRTIS_ENABLE_GPU
           } else {
@@ -947,6 +949,8 @@ OnnxBackend::Context::SetStringOutputBuffer(
 
     element_idx += expected_element_cnt;
   }
+
+  return cuda_copy;
 }
 
 void
--- a/src/backends/tensorflow/base_backend.cc
+++ b/src/backends/tensorflow/base_backend.cc
@@ -442,7 +442,7 @@ ReadStringOutputTensor(
       int64_t device_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
           output_name, &content, serialized.size(), shape,
-          preferred_memory_type, &actual_memory_type, &device_id);
+          preferred_memory_type, 0, &actual_memory_type, &device_id);
       if (status.IsOk()) {
         if (actual_memory_type == TRTSERVER_MEMORY_GPU) {
 #ifdef TRTIS_ENABLE_GPU
--- a/src/backends/tensorflow/base_backend.cc
+++ b/src/backends/tensorflow/base_backend.cc
@@ -441,11 +441,31 @@ ReadStringOutputTensor(
       TRTSERVER_Memory_Type actual_memory_type;
       int64_t device_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
-          output_name, &content, serialized.size(), shape, preferred_memory_type, &actual_memory_type, &device_id);
+          output_name, &content, serialized.size(), shape,
+          preferred_memory_type, &actual_memory_type, &device_id);
       if (status.IsOk()) {
-        memcpy(
-            content, reinterpret_cast<const void*>(serialized.c_str()),
-            serialized.size());
+        if (actual_memory_type == TRTSERVER_MEMORY_GPU) {
+#ifdef TRTIS_ENABLE_GPU
+          cudaStream_t stream = nullptr;
+          cudaError_t err = cudaMemcpyAsync(
+              content, reinterpret_cast<const void*>(serialized.c_str()),
+              serialized.size(), cudaMemcpyDeviceToHost, stream);
+          if (err != cudaSuccess) {
+            payload.status_ = Status(
+                RequestStatusCode::INTERNAL,
+                "output '" + output_name +
+                    "' is in GPU memory and failed to use" +
+                    " CUDA copy from CPU memory to get String output data: " +
+                    std::string(cudaGetErrorString(err)));
+          } else {
+            cudaStreamSynchronize(stream);
+          }
+#endif  // TRTIS_ENABLE_GPU
+        } else {
+          memcpy(
+              content, reinterpret_cast<const void*>(serialized.c_str()),
+              serialized.size());
+        }
       } else {
         payload.status_ = status;
       }
--- a/src/backends/tensorflow/base_backend.cc
+++ b/src/backends/tensorflow/base_backend.cc
@@ -437,8 +437,11 @@ ReadStringOutputTensor(
       }
 
       void* content;
+      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
+      TRTSERVER_Memory_Type actual_memory_type;
+      int64_t device_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
-          output_name, &content, serialized.size(), shape);
+          output_name, &content, serialized.size(), shape, preferred_memory_type, &actual_memory_type, &device_id);
       if (status.IsOk()) {
         memcpy(
             content, reinterpret_cast<const void*>(serialized.c_str()),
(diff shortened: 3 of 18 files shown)


src/servers/grpc_server.cc
```diff
@@ -635,7 +637,9 @@ TRTSERVER_Error*
 InferResponseAlloc(
     TRTSERVER_ResponseAllocator* allocator, void** buffer, void** buffer_userp,
     const char* tensor_name, size_t byte_size,
-    TRTSERVER_Memory_Type memory_type, int64_t memory_type_id, void* userp)
+    TRTSERVER_Memory_Type preferred_memory_type, int64_t memory_type_id,
+    void* userp, TRTSERVER_Memory_Type* actual_memory_type,
+    int64_t* actual_device_id)
```
deadeyegoodwin: Fix

src/servers/grpc_server.cc
```diff
@@ -685,6 +683,15 @@ InferResponseAlloc(
...
 
     if (!use_shm) {
+      // Can't allocate for any memory type other than CPU. If asked to
+      // allocate on GPU memory then force allocation on CPu instead.
+      if (*actual_memory_type != TRTSERVER_MEMORY_CPU) {
+        LOG_VERBOSE(1) << "GRPC allocation failed for type "
+                       << *actual_memory_type << " for " << tensor_name
+                       << ", will use type " << TRTSERVER_MEMORY_CPU;
+        *actual_memory_type = TRTSERVER_MEMORY_CPU;
+      }
```
deadeyegoodwin: Set actual_memory_type_id = 0

src/servers/grpc_server.cc
```diff
@@ -727,21 +734,24 @@ InferAllocatorPayload(
   // invoked.
   for (const auto& io : request_header.output()) {
     if (io.has_shared_memory()) {
-      TRTSERVER_SharedMemoryBlock* smb = nullptr;
-      RETURN_IF_ERR(smb_manager->Get(&smb, io.shared_memory().name()));
-
       if (alloc_payload->shm_map_ == nullptr) {
```
deadeyegoodwin: Move this as late as possible, otherwise the RETURN_IF_ERR below will cause a leak.

src/core/ensemble_scheduler.cc
```diff
@@ -658,33 +660,25 @@ EnsembleContext::CheckAndSetEnsembleOutput()
...
 
     // Use the memory type of the memory block as preferred memory type
-    TRTSERVER_Memory_Type dst_memory_type;
+    TRTSERVER_Memory_Type dst_memory_type, allocated_memory_type;
     int64_t memory_type_id;
     size_t content_size;
     memory_block->BufferAt(0, &content_size, &dst_memory_type, &memory_type_id);
 
     void* buffer;
+    int64_t device_id;
```
deadeyegoodwin: allocated_memory_type_id

src/core/ensemble_scheduler.cc
```diff
@@ -658,33 +660,25 @@ EnsembleContext::CheckAndSetEnsembleOutput()
...
-            output_pair.first, &buffer, expected_byte_size, shape,
-            dst_memory_type, memory_type_id));
-      }
-      if (buffer == nullptr) {
-        return Status(
-            RequestStatusCode::INTERNAL,
-            "all attempts to allocate buffer for output '" + output_pair.first +
-                "' failed");
-      }
+      return Status(
```
deadeyegoodwin: This isn't the right message anymore. There is only one allocation attempt. AllocateOutputBuffer and its callers need to be changed to this new way of working.

src/core/provider.h
```diff
@@ -297,8 +297,10 @@ class InferResponseProvider {
   Status AllocateOutputBuffer(
       const std::string& name, void** content, size_t content_byte_size,
       const std::vector<int64_t>& content_shape,
-      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU,
-      int64_t preferred_memory_type_id = 0);
+      const TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU,
```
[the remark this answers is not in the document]
CoderHam: For the preferred_memory_type as well?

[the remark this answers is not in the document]
deadeyegoodwin: Yes

src/core/provider.cc
```diff
@@ -705,8 +705,9 @@ Status
 InferResponseProvider::AllocateOutputBuffer(
     const std::string& name, void** content, size_t content_byte_size,
     const std::vector<int64_t>& content_shape,
-    TRTSERVER_Memory_Type preferred_memory_type,
-    int64_t preferred_memory_type_id)
+    const TRTSERVER_Memory_Type preferred_memory_type,
+    int64_t preferred_memory_type_id, TRTSERVER_Memory_Type* actual_memory_type,
+    int64_t* actual_device_id)
 {
   *content = nullptr;
```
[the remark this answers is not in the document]
GuanLuo: I added line 722 to make this function behaves like the allocation callback: call again with other memory type if the previous type doesn't work. This PR should just remove 722 and related section

src/core/provider.cc
```diff
@@ -743,15 +746,16 @@ InferResponseProvider::AllocateOutputBuffer(
...
     }
-    if (preferred_memory_type == TRTSERVER_MEMORY_CPU) {
-      loutput->cls_count_ = pr->second.cls().count();
-      char* buffer = new char[content_byte_size];
-      *content = static_cast<void*>(buffer);
-      loutput->ptr_ = static_cast<void*>(buffer);
-      loutput->buffer_.reset(buffer);
-    } else {
-      return Status::Success;
+    if (preferred_memory_type != TRTSERVER_MEMORY_CPU) {
```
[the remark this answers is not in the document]
CoderHam: If the output is class type instead of Raw and memory is GPU then we set the actual memory type to CPU and device id to 0 here. It needs to be done before the alloc_fn is called

src/core/provider.cc
```diff
@@ -776,7 +780,8 @@ InferResponseProvider::AllocateOutputBuffer(
 #endif  // TRTIS_ENABLE_GPU
   TRTSERVER_Error* err = alloc_fn_(
       allocator_, &buffer, &buffer_userp, name.c_str(), alloc_byte_size,
-      preferred_memory_type, preferred_memory_type_id, alloc_userp_);
+      preferred_memory_type, preferred_memory_type_id, alloc_userp_,
+      actual_memory_type, actual_device_id);
```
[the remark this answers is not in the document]
CoderHam: Sure

src/core/provider.cc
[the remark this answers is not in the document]
deadeyegoodwin: actual* are outputs from the alloc_fn so we don't need to set them before. And the preferred doesn't matter since we are requesting 0 size.

src/core/provider.cc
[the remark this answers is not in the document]
CoderHam: Understood, will remove

src/servers/simple.cc
```diff
@@ -109,24 +110,32 @@ ResponseAlloc(
...
+#ifdef TRTIS_ENABLE_GPU
       auto err = cudaSetDevice(memory_type_id);
       if (err == cudaSuccess) {
         err = cudaMalloc(&allocated_ptr, byte_size);
       }
+
       if (err != cudaSuccess) {
         LOG_INFO << "cudaMalloc failed: " << cudaGetErrorString(err);
         allocated_ptr = nullptr;
+        *actual_memory_type = TRTSERVER_MEMORY_GPU;
```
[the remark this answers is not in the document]
CoderHam: Done

src/servers/grpc_server.cc
CoderHam: Agreed

src/core/ensemble_scheduler.cc
CoderHam: Done. Will fix this message

src/core/trtserver.h
```diff
@@ -197,9 +206,11 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_SharedMemoryBlockDelete(
 ///
 /// If the function is called with 'byte_size' non-zero but an
```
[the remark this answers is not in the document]
CoderHam: Fixed

src/core/trtserver.h
```diff
@@ -197,9 +206,11 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_SharedMemoryBlockDelete(
...
-/// indicate that an allocation in the requested 'memory_type' is not
-/// possible. In this case the function may be called again for the
+/// function will try a different memory type and return that as the
+/// 'actual_memory_type'. If this too fails then it returns failure and
+/// sets 'buffer' == nullptr to indicate that an allocation in the requested
+/// 'memory_type' as well as the 'actual_memory_type' is not possible.
+/// In this case the function may be called again for the
 /// same 'tensor_name' but with a different 'memory_type'.
 ///
 /// The function should return a TRTSERVER_Error object if a failure
```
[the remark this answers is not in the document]
CoderHam: Done

src/core/provider.cc
```diff
@@ -729,11 +730,13 @@ InferResponseProvider::AllocateOutputBuffer(
   loutput->byte_size_ = content_byte_size;
   loutput->memory_type_ = preferred_memory_type;
```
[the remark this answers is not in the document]
CoderHam: Done

src/core/provider.cc
```diff
@@ -743,15 +746,16 @@ InferResponseProvider::AllocateOutputBuffer(
...
+      *actual_memory_type = TRTSERVER_MEMORY_CPU;
+      loutput->memory_type_ = *actual_memory_type;
+      loutput->memory_type_id_ = 0;
     }
+    loutput->cls_count_ = pr->second.cls().count();
+    char* buffer = new char[content_byte_size];
+    *content = static_cast<void*>(buffer);
+    loutput->ptr_ = static_cast<void*>(buffer);
+    loutput->buffer_.reset(buffer);
   }
```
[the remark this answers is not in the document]
CoderHam: Done

src/core/provider.cc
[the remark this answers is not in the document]
CoderHam: Understood

src/servers/simple.cc
```diff
@@ -109,24 +110,32 @@ ResponseAlloc(
     void* allocated_ptr = nullptr;
     if (memory_type == TRTSERVER_MEMORY_CPU) {
       allocated_ptr = malloc(byte_size);
-#ifdef TRTIS_ENABLE_GPU
+      *actual_memory_type = TRTSERVER_MEMORY_CPU;
+      *actual_device_id = 0;
     } else if (use_gpu_memory) {
+#ifdef TRTIS_ENABLE_GPU
       auto err = cudaSetDevice(memory_type_id);
       if (err == cudaSuccess) {
```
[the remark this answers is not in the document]
CoderHam: Done

src/servers/simple.cc
```diff
@@ -109,24 +110,32 @@ ResponseAlloc(
...
         err = cudaMalloc(&allocated_ptr, byte_size);
       }
+
       if (err != cudaSuccess) {
```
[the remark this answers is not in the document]
CoderHam: Done

src/backends/custom/custom_backend.cc
```diff
@@ -583,28 +583,24 @@ CustomBackend::Context::GetOutput(
     if (shape_dim_cnt > 0) {
       shape.assign(shape_dims, shape_dims + shape_dim_cnt);
     }
+
+    auto dst_memory_type = ToTRTServerMemoryType(*memory_type);
```
GuanLuo: `dst_memory_type` is not used anywhere, why change from directly using `ToTRTServerMemoryType(*memory_type)` in `AllocateOutputBuffer()`

src/backends/custom/custom_backend.cc
```diff
@@ -583,28 +583,24 @@ CustomBackend::Context::GetOutput(
...
-          name, content, content_byte_size, shape,
-          ToTRTServerMemoryType(*memory_type), *memory_type_id);
-      if (*content == nullptr) {
-        return false;
-      }
+      return false;
     }
+
+    // Update memory type with actual memory type
+    *memory_type = ToCustomMemoryType(actual_memory_type);
```
GuanLuo: Should also set `*memory_type_id = actual_*`

src/backends/onnx/onnx_backend.cc
```diff
@@ -883,21 +883,61 @@ OnnxBackend::Context::SetStringOutputBuffer(
           data_byte_size + sizeof(uint32_t) * expected_element_cnt;
 
       void* buffer;
+      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
+      TRTSERVER_Memory_Type actual_memory_type;
+      int64_t device_id;
```
GuanLuo: `actual_memory_type_id`

src/backends/onnx/onnx_backend.cc
```diff
@@ -883,21 +883,61 @@ OnnxBackend::Context::SetStringOutputBuffer(
...
+                content + offsets[element_idx + e], len, cudaMemcpyDeviceToHost,
+                stream_);
+            if (err != cudaSuccess) {
+              payload.status_ = Status(
+                  RequestStatusCode::INTERNAL,
+                  "output '" + name + "' is in GPU memory and failed to use" +
+                      " CUDA copy from CPU memory to get String output data: " +
+                      std::string(cudaGetErrorString(err)));
+            } else {
+              cudaStreamSynchronize(stream_);
```
GuanLuo: Call `cudaStreamSynchronize()` at the end, or do it like for fixed size tensor: propagating flag all the way up to `Run()` and call `cudaStreamSynchronize()` once for all

src/backends/tensorflow/base_backend.cc
```diff
@@ -437,12 +437,35 @@ ReadStringOutputTensor(
       }
 
       void* content;
+      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
+      TRTSERVER_Memory_Type actual_memory_type;
+      int64_t device_id;
```
GuanLuo: same

src/backends/tensorflow/base_backend.cc
```diff
@@ -437,12 +437,35 @@ ReadStringOutputTensor(
...
+              serialized.size(), cudaMemcpyDeviceToHost, stream);
+          if (err != cudaSuccess) {
+            payload.status_ = Status(
+                RequestStatusCode::INTERNAL,
+                "output '" + output_name +
+                    "' is in GPU memory and failed to use" +
+                    " CUDA copy from CPU memory to get String output data: " +
+                    std::string(cudaGetErrorString(err)));
+          } else {
+            cudaStreamSynchronize(stream);
```
GuanLuo: Move to end

src/core/backend_context.cc
```diff
@@ -165,34 +165,26 @@ BackendContext::SetFixedSizeOutputBuffer(
...
-            bool cuda_used = false;
-            status = CopyBuffer(
-                name, src_memory_type, dst_memory_type, expected_byte_size,
-                content + content_offset, buffer, &cuda_used);
-            cuda_copy |= cuda_used;
-          }
+        if (buffer == nullptr) {
+          status = Status(
+              RequestStatusCode::INTERNAL,
+              "all attempts to allocate buffer for output '" + name +
```
GuanLuo: same as David's earlier comment

src/core/ensemble_scheduler.cc
```diff
@@ -658,33 +660,24 @@ EnsembleContext::CheckAndSetEnsembleOutput()
...
     int64_t memory_type_id;
     size_t content_size;
     memory_block->BufferAt(0, &content_size, &dst_memory_type, &memory_type_id);
 
     void* buffer;
+    int64_t allocated_memory_type_id;
     RETURN_IF_ERROR(response_provider_->AllocateOutputBuffer(
         output_pair.first, &buffer, expected_byte_size, shape, dst_memory_type,
-        memory_type_id));
+        memory_type_id, &allocated_memory_type, &allocated_memory_type_id));
```
GuanLuo: This is not right, either let `AllocateOutputBuffer()` write actual type into `dst_memory_type` or change the code below to condition on `allocated_memory_type` instead of `dst_memory_type`.

src/core/trtserver.cc
```diff
@@ -570,6 +570,31 @@ TRTSERVER_SharedMemoryBlockDelete(
...
+TRTSERVER_SharedMemoryBlockMemoryTypeId(
+    TRTSERVER_SharedMemoryBlock* shared_memory_block, int64_t* memory_type_id)
+{
+  TrtServerSharedMemoryBlock* lsmb =
+      reinterpret_cast<TrtServerSharedMemoryBlock*>(shared_memory_block);
+#ifdef TRTIS_ENABLE_GPU
+  *memory_type_id = lsmb->DeviceId();
+#else
+  *memory_type_id = 0;
+#endif             // TRTIS_ENABLE_GPU
```
GuanLuo: Is `DeviceId()` not defined if `TRTIS_ENABLE_GPU=OFF`? If it is defined, you don't need `#ifdef` directive.

src/core/trtserver.h
```diff
@@ -169,6 +169,21 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_SharedMemoryBlockGpuNew(
...
+/// \param shared_memory_block The object whose memory type is required.
+/// \param memory_type Returns the memory type of the shared memory block.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_Error* TRTSERVER_SharedMemoryBlockMemoryType(
+    TRTSERVER_SharedMemoryBlock* shared_memory_block,
+    TRTSERVER_Memory_Type* memory_type);
+
+/// Get the memory type id of a shared memory block object.
+/// \param shared_memory_block The object whose memory type is required.
+/// \param memory_type_id The CPU/GPU number the shared memory region is in.
```
GuanLuo: CPU number seems strange here. Currently we don't have meaning for `memory_type_id` in the case of system memory. Perhaps just say the device ID if the region is in GPU memory.

src/backends/custom/custom_backend.cc
CoderHam: Fixed. Was using it to initialize actual_memory_type. Assigned directly instead.

src/backends/onnx/onnx_backend.cc
CoderHam: Sure

src/core/trtserver.cc
CoderHam: Yes, `DeviceId()` is not defined when `TRTIS_ENABLE_GPU_=OFF`

src/core/trtserver.h
CoderHam: Sure

src/core/backend_context.cc
CoderHam: Fixed

src/core/ensemble_scheduler.cc
CoderHam: Fixed.

src/backends/custom/custom_backend.cc
```diff
@@ -583,28 +583,24 @@ CustomBackend::Context::GetOutput(
     if (shape_dim_cnt > 0) {
       shape.assign(shape_dims, shape_dims + shape_dim_cnt);
     }
+
+    auto actual_memory_type = ToTRTServerMemoryType(*memory_type);
```
deadeyegoodwin: Don't initialize actual_memory_type... just causes overhead and it is an "out" parameter,  not an "inout"

src/backends/onnx/onnx_backend.cc
```diff
@@ -883,21 +884,61 @@ OnnxBackend::Context::SetStringOutputBuffer(
           data_byte_size + sizeof(uint32_t) * expected_element_cnt;
 
       void* buffer;
+      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
+      TRTSERVER_Memory_Type actual_memory_type;
+      int64_t actual_memory_type_id;
       Status status = payload.response_provider_->AllocateOutputBuffer(
-          name, &buffer, expected_byte_size, content_shape);
+          name, &buffer, expected_byte_size, content_shape,
+          preferred_memory_type, 0, &actual_memory_type, &actual_memory_type_id);
```
deadeyegoodwin: Add /* preferred_memory_type_id */ next to magic 0

src/backends/tensorflow/base_backend.cc
```diff
@@ -437,19 +442,47 @@ ReadStringOutputTensor(
       }
 
       void* content;
+      TRTSERVER_Memory_Type preferred_memory_type = TRTSERVER_MEMORY_CPU;
```
deadeyegoodwin: All this tensor copy logic (including for string datatype), should be in a common shared area and handle all the possible cases. Should also return the bool indicating if sync is needed. @GuanLuo is this some of the "common code" you were mentioning earlier. @CoderHam Don't need to fix for this change be we do need to do this cleanup so please file a ticket for it.

src/backends/tensorflow/base_backend.cc
```diff
@@ -409,6 +409,11 @@ ReadStringOutputTensor(
     std::vector<Scheduler::Payload>* payloads)
 {
   size_t tensor_element_idx = 0;
+  bool cuda_copy = false;
+
+#ifdef TRTIS_ENABLE_GPU
+  cudaStream_t stream = nullptr;
```
GuanLuo: There should be a `stream_` member in context. Just move this function to the context class and use it. Or if you don't want to move function around, you still need to create the stream, otherwise the default stream will be used

src/servers/simple.cc
```diff
@@ -108,27 +108,40 @@ ResponseAlloc(
...
 
       if (err != cudaSuccess) {
-        LOG_INFO << "cudaMalloc failed: " << cudaGetErrorString(err);
+        return TRTSERVER_ErrorNew(
+            TRTSERVER_ERROR_INTERNAL,
+            std::string(
+                "cudaMalloc failed: " + std::string(cudaGetErrorString(err)))
+                .c_str());
         allocated_ptr = nullptr;
         *actual_memory_type = TRTSERVER_MEMORY_GPU;
```
GuanLuo: Code after return is not necessary

src/servers/simple.cc
```diff
@@ -108,27 +108,40 @@ ResponseAlloc(
...
-      if (err == cudaSuccess) {
+      if ((err != cudaSuccess) && (err != cudaErrorNoDevice) &&
+          (err != cudaErrorInsufficientDriver)) {
+        return TRTSERVER_ErrorNew(
+            TRTSERVER_ERROR_INTERNAL,
+            std::string(
+                "unable to recover current CUDA device: " +
+                std::string(cudaGetErrorString(err)))
+                .c_str());
+      } else if (err == cudaSuccess) {
```
GuanLuo: Don't need `else if` here since the other condition will just return now.

src/core/backend_context.cc
```diff
@@ -165,34 +165,26 @@ BackendContext::SetFixedSizeOutputBuffer(
     if (payload.status_.IsOk() && (payload.response_provider_ != nullptr) &&
         payload.response_provider_->RequiresOutput(name)) {
       auto dst_memory_type = src_memory_type;
+      int64_t dst_device_id;
```
deadeyegoodwin: dst_memory_type_id

src/core/backend_context.cc
```diff
@@ -165,34 +165,26 @@ BackendContext::SetFixedSizeOutputBuffer(
...
       void* buffer = nullptr;
 
       // try to get buffer with the same memory type as the output tensor
       Status status = payload.response_provider_->AllocateOutputBuffer(
           name, &buffer, expected_byte_size, content_shape, src_memory_type,
-          src_memory_type_id);
-
+          src_memory_type_id, &dst_memory_type, &dst_device_id);
+      LOG_VERBOSE(1) << "src_memory_type: " << src_memory_type;
```
deadeyegoodwin: Make a single, more expressive log message or remove.

src/core/backend_context.cc
```diff
@@ -165,34 +165,26 @@ BackendContext::SetFixedSizeOutputBuffer(
...
-            cuda_copy |= cuda_used;
-          }
+        if (buffer == nullptr) {
+          status = Status(
+              RequestStatusCode::INTERNAL,
+              "failed to allocate buffer for output '" + name + "'");
+        } else {
+          bool cuda_used = false;
+          status = CopyBuffer(
+              name, src_memory_type, dst_memory_type, expected_byte_size,
```
deadeyegoodwin: How can we copy correctly without knowing the memory_type_ids?

src/servers/simple.cc
```diff
@@ -107,26 +108,47 @@ ResponseAlloc(
              << tensor_name;
   } else {
     void* allocated_ptr = nullptr;
-    if (memory_type == TRTSERVER_MEMORY_CPU) {
+    if (!use_gpu_memory || (memory_type == TRTSERVER_MEMORY_CPU)) {
       allocated_ptr = malloc(byte_size);
-#ifdef TRTIS_ENABLE_GPU
+      *actual_memory_type = TRTSERVER_MEMORY_CPU;
+      *actual_memory_type_id = 0;
     } else if (use_gpu_memory) {
```
deadeyegoodwin: This can be unconditional else?

src/servers/simple.cc
```diff
@@ -107,26 +108,47 @@ ResponseAlloc(
...
+            TRTSERVER_ERROR_INTERNAL,
+            std::string(
+                "cudaMalloc failed: " + std::string(cudaGetErrorString(err)))
+                .c_str());
         allocated_ptr = nullptr;
+        *actual_memory_type = TRTSERVER_MEMORY_GPU;
+        *actual_memory_type_id = 0;
       }
 #endif  // TRTIS_ENABLE_GPU
+    } else {
```
deadeyegoodwin: Remove this else clause?

src/backends/tensorflow/base_backend.cc
GuanLuo: This is not the part that I think of originally. I don't think we need to generalize it yet, since currently only ONNX and TensorFlow backends are accepting string data type and the two backends handle string type differently. But still, we can unify the logic and move it to common area if other frameworks are going to support string data type.

src/servers/simple.cc
CoderHam: Yes. It can. For all other cases CPU will be used except when use_gpu_memory = true and memory_type = TRTSERVER_MEMORY_GPU

src/core/backend_context.cc
CoderHam: Used this for logging. Removed

src/backends/tensorflow/base_backend.cc
CoderHam: Filled a ticket for the cleanup of the tensor copy logic

src/backends/custom/custom_backend.cc
CoderHam: Done. Moved this initialization inside the function.

src/core/backend_context.cc
CoderHam: cudaMemcpyDeviceToDevice is used. According to the docs it is able to detect source and destination GPU device IDS based on the address. @GuanLuo can you confirm?

CoderHam: It appears that DeviceToDevice copy should only be used for copies on the same GPU. For different GPU devices, `cudaMemcpyPeerAsync` should be used. 
Here we should also check if PeerAccess is allowed between those two GPUs else do DeviceToHost + HostToDevice (don't see a better route) I believe this would tie into the tensor copy logic cleanup.

GuanLuo: I didn't realize `cudaMemcpyDeviceToDevice` flag is limited to copy within the same device, this may explain the bug that I am working on. We should address this ASAP as data transfer between GPUs are clearly broken.

Doing  "DeviceToHost + HostToDevice" should be the default choice. But it seems to have too much an overhead. Then we can optimize in the case if PeerAccess is allowed. And in the case that it's not, we should warn user that using CUDA shared memory across devices is not suggested and for ensemble, the intermediate data should always be allocated in system memory (until it gets smarter).

CoderHam: This makes sense. What I have done right now should be enough for now. We can add Peer Access to avoid transfer in the copy tensor cleanup ticket

src/core/backend_context.cc
```diff
@@ -210,14 +211,25 @@ BackendContext::CopyBuffer(
     memcpy(dst, src, byte_size);
   } else {
 #ifdef TRTIS_ENABLE_GPU
+    // [TODO] Add check for Peer Access between two GPUs and use
```
GuanLuo: Not clear if we need to check for peer access here, the copy is necessary as we want data to be in contiguous chunk for input and output to be moved to allocated buffer.

src/backends/tensorflow/base_backend.cc
```diff
@@ -410,6 +410,16 @@ ReadStringOutputTensor(
 {
   size_t tensor_element_idx = 0;
 
+#ifdef TRTIS_ENABLE_GPU
+  bool cuda_copy = false;
+  cudaStream_t stream;
+  cudaError_t cuerr = cudaStreamCreate(&stream);
```
deadeyegoodwin: This is really bad to be creating a stream every time we copy a string output tensor. Don't we already have a stream created for each backend for other copy tasks? This should be dealt with in the copy cleanup.

src/backends/tensorflow/base_backend.cc
```diff
@@ -437,19 +447,48 @@ ReadStringOutputTensor(
...
         payload.status_ = status;
       }
     }
 
     tensor_element_idx += expected_element_cnt;
   }
+
+#ifdef TRTIS_ENABLE_GPU
+  if (cuda_copy) {
+    cudaStreamSynchronize(stream);
```
deadeyegoodwin: You need to delete the stream.

src/core/provider.cc
```diff
@@ -762,6 +756,8 @@ InferResponseProvider::AllocateOutputBuffer(
 
   void* buffer = nullptr;
   void* buffer_userp = nullptr;
+  TRTSERVER_Memory_Type raw_actual_memory_type = preferred_memory_type;
```
deadeyegoodwin: Don't set these... it is confusing and they are output values that must be set by alloc_fn_.

deadeyegoodwin:
@CoderHam Also looks like you need to manually rebase to resolve some conflicts.

src/backends/tensorflow/base_backend.cc
CoderHam: Done

src/core/backend_context.cc
CoderHam: Removing Peer Access comment for now. This is true, a memcpy would be necessary

src/core/backend_context.cc
```diff
@@ -209,8 +200,9 @@ BackendContext::SetFixedSizeOutputBuffer(
 Status
 BackendContext::CopyBuffer(
     const std::string& name, const TRTSERVER_Memory_Type src_memory_type,
-    const TRTSERVER_Memory_Type dst_memory_type, const size_t byte_size,
-    const void* src, void* dst, bool* cuda_used)
+    const int src_memory_type_id, const TRTSERVER_Memory_Type dst_memory_type,
```
deadeyegoodwin: Don't use "int" as the type of memory_type_id (and don't cast to int in the calls to this function). The the type of memory_type_id, which is int64_t.

src/core/trtserver.h
```diff
@@ -195,23 +210,23 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_SharedMemoryBlockDelete(
...
+/// 'actual_memory_type_id' the memory where the buffer is allocated.
 ///
 /// The function should return a TRTSERVER_Error object if a failure
 /// occurs while attempting an allocation. If an error object is
-/// returned, or if 'buffer' == nullptr is returned on all attempts
-/// for a result tensor, the inference server will assume allocation
+/// returned for a result tensor, the inference server will assume allocation
 /// is not possible for the result buffer and will abort the inference
 /// request.
 typedef TRTSERVER_Error* (*TRTSERVER_ResponseAllocatorAllocFn_t)(
```
deadeyegoodwin: Sorry I didn't notice this before but the order of args in this function are incorrect. To be consistent with other API (and also not to be so jumbled even within this call) should have object (allocator in this case), followed by inputs, followed by outputs. So (allocator, tensor_name, byte_size, memory_type, memory_type_id, userp, buffer, buffer_userp, actual_memory_type, actual_memory_type_id)

CoderHam: Fixed

src/core/backend_context.cc
CoderHam: Fixed

PR #784, opened 2019-10-23
Client Lib clean-up

tanmayv25:
Removing the support of older API.

### DIFF
--- a/qa/L0_sequence_stress/sequence_stress.py
+++ b/qa/L0_sequence_stress/sequence_stress.py
@@ -59,7 +59,7 @@ class UserData:
     def __init__(self):
         self._completed_requests = queue.Queue()
 
-# Callback function used for async_run_with_cb()
+# Callback function used for async_run()
 def completion_callback(value, expected_result, user_data, infer_ctx, request_id):
     user_data._completed_requests.put((request_id, value, expected_result))
 
@@ -104,7 +104,7 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
                 in0 = np.full(tensor_shape, value, dtype=input_dtype)
             input_list.append(in0)
 
-        ctx.async_run_with_cb(partial(completion_callback, value, expected_result, user_data), 
+        ctx.async_run(partial(completion_callback, value, expected_result, user_data), 
                             { 'INPUT' :input_list }, { 'OUTPUT' : InferContext.ResultFormat.RAW},
                                batch_size=batch_size, flags=flags)
         i+=1
--- a/qa/L0_sequence_stress/sequence_stress.py
+++ b/qa/L0_sequence_stress/sequence_stress.py
@@ -85,7 +85,7 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
     seq_start_ms = int(round(time.time() * 1000))
     user_data = UserData()
 
-    i=0
+    sent_count=0
     for flag_str, value, expected_result, delay_ms in steps:
         flags = InferRequestHeader.FLAG_NONE
         if flag_str is not None:
@@ -107,7 +107,7 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
         ctx.async_run(partial(completion_callback, value, expected_result, user_data), 
                             { 'INPUT' :input_list }, { 'OUTPUT' : InferContext.ResultFormat.RAW},
                                batch_size=batch_size, flags=flags)
-        i+=1
+        sent_count += 1
 
         if delay_ms is not None:
             time.sleep(delay_ms / 1000.0)
@@ -115,7 +115,7 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
     # Process the results in order that they were sent
     result = None
     processed_count = 0
-    while processed_count < i:
+    while processed_count < sent_count:
         (id, value, expected) = user_data._completed_requests.get()
         processed_count += 1
         results = None
--- a/qa/L0_sequence_stress/sequence_stress.py
+++ b/qa/L0_sequence_stress/sequence_stress.py
@@ -120,7 +120,7 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
         processed_count += 1
         results = None
         while results == None:
-            results = ctx.get_async_run_results(id, False)
+            results = ctx.get_async_run_results(id)
             if results == None:
                 if timeout_ms != None:
                     now_ms = int(round(time.time() * 1000))
--- a/qa/L0_sequence_stress/sequence_stress.py
+++ b/qa/L0_sequence_stress/sequence_stress.py
@@ -37,8 +37,14 @@ import threading
 import traceback
 import numpy as np
 import test_util as tu
+from functools import partial
 from tensorrtserver.api import *
 import tensorrtserver.api.server_status_pb2 as server_status
+
+if sys.version_info >= (3, 0):
+  import queue
+else:
+  import Queue as queue
 
 FLAGS = None
 CORRELATION_ID_BLOCK_SIZE = 100
@@ -48,6 +54,14 @@ SEQUENCE_LENGTH_STDEV = 8
 
 _thread_exceptions = []
 _thread_exceptions_mutex = threading.Lock()
+
+class UserData:
+    def __init__(self):
+        self._completed_requests = queue.Queue()
+
+# Callback function used for async_run_with_cb()
+def completion_callback(value, expected_result, user_data, infer_ctx, request_id):
+    user_data._completed_requests.put((request_id, value, expected_result))
 
 class TimeoutException(Exception):
     pass
@@ -69,10 +83,9 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
 
     # Execute the sequence of inference...
     seq_start_ms = int(round(time.time() * 1000))
-    input_values = list()
-    result_ids = list()
-    expected_results = list()
+    user_data = UserData()
 
+    i=0
     for flag_str, value, expected_result, delay_ms in steps:
         flags = InferRequestHeader.FLAG_NONE
         if flag_str is not None:
@@ -91,18 +104,20 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
                 in0 = np.full(tensor_shape, value, dtype=input_dtype)
             input_list.append(in0)
 
-        input_values.append(value)
-        expected_results.append(expected_result)
-        result_ids.append(ctx.async_run(
-            { "INPUT" : input_list }, { "OUTPUT" : InferContext.ResultFormat.RAW},
-            batch_size=batch_size, flags=flags))
+        ctx.async_run_with_cb(partial(completion_callback, value, expected_result, user_data), 
+                            { 'INPUT' :input_list }, { 'OUTPUT' : InferContext.ResultFormat.RAW},
+                               batch_size=batch_size, flags=flags)
+        i+=1
 
         if delay_ms is not None:
             time.sleep(delay_ms / 1000.0)
 
-    # Wait for the results in the order sent
+    # Process the results in order that they were sent
     result = None
-    for id, expected, value in zip(result_ids, expected_results, input_values):
+    processed_count = 0
+    while processed_count < i:
+        (id, value, expected) = user_data._completed_requests.get()
+        processed_count += 1
         results = None
         while results == None:
             results = ctx.get_async_run_results(id, False)
--- a/src/clients/c++/examples/image_client.cc
+++ b/src/clients/c++/examples/image_client.cc
@@ -689,7 +689,7 @@ main(int argc, char** argv)
   std::mutex mtx;
   std::condition_variable cv;
 
-  std::queue<std::shared_ptr<nic::InferContext::Request>> request_buffer_;
+  std::queue<std::shared_ptr<nic::InferContext::Request>> async_request_queue_;
 
   while (!last_request) {
     // Already verified that there is 1 input...
@@ -733,11 +733,12 @@ main(int argc, char** argv)
       err = ctx->AsyncRun(
           [&](nic::InferContext* ctx,
               const std::shared_ptr<nic::InferContext::Request>& request) {
-            std::lock_guard<std::mutex> lk(mtx);
-            done_cnt++;
-            request_buffer_.push(request);
+            {
+              std::lock_guard<std::mutex> lk(mtx);
+              done_cnt++;
+              async_request_queue_.push(request);
+            }
             cv.notify_all();
-            return;
           });
       if (!err.IsOk()) {
         std::cerr << "failed sending asynchronous infer request: " << err
@@ -760,14 +761,15 @@ main(int argc, char** argv)
         }
       });
     }
-    while (!request_buffer_.empty()) {
+    while (!async_request_queue_.empty()) {
       results.emplace_back();
-      err = ctx->GetAsyncRunResults(request_buffer_.front(), &(results.back()));
+      err = ctx->GetAsyncRunResults(
+          async_request_queue_.front(), &(results.back()));
       if (!err.IsOk()) {
         std::cerr << "failed receiving infer response: " << err << std::endl;
         exit(1);
       }
-      request_buffer_.pop();
+      async_request_queue_.pop();
     }
   }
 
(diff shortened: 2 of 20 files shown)


src/clients/c++/library/request_grpc.cc
```diff
@@ -944,31 +927,32 @@ InferGrpcContextImpl::AsyncTransfer()
...
-      // wake up if at least one request is not ready
-      for (auto& ongoing_async_request : this->ongoing_async_requests_) {
-        if (std::static_pointer_cast<GrpcRequestImpl>(
-                ongoing_async_request.second)
-                ->IsReady() == false) {
-          return true;
-        }
+      // wake up if an async request has been generated
+      if (!this->ongoing_async_requests_.empty()) {
+        return true;
```
GuanLuo: can just be `return !this->ongoing_async_requests_.empty()`

src/clients/c++/library/request_grpc.cc
```diff
@@ -944,31 +927,32 @@ InferGrpcContextImpl::AsyncTransfer()
...
+      if (!ok) {
+        // The lock would prevent the jumbling up of the error messages.
+        // Additionally, doesn't impact performance as it is outside the
+        // execution path.
+        std::lock_guard<std::mutex> lock(mutex_);
+        fprintf(stderr, "Unexpected not ok on client side.");
+      }
+      if (!status) {
+        std::lock_guard<std::mutex> lock(mutex_);
+        fprintf(stderr, "Completion queue is closed.");
```
GuanLuo: We are just logging error here, do we need to handle them? The part below will still be executed on error...

src/clients/c++/library/request_grpc.cc
```diff
@@ -1073,12 +1053,27 @@ InferGrpcStreamContextImpl::Run(ResultMap* results)
...
   if (!err.IsOk()) {
     return err;
   }
-  bool is_ready;
-  return GetAsyncRunResults(results, &is_ready, req, true);
+  // Wait for the callback
+  {
+    std::unique_lock<std::mutex> lk(mtx);
+    cv.wait(lk, [&callback_invoked]() { return callback_invoked; });
+  }
```
GuanLuo: The above can be simplified with promise and future if you want.

src/clients/c++/library/request_http.cc
```diff
@@ -1622,29 +1601,24 @@ InferHttpContextImpl::AsyncTransfer()
...
+        http_request->callback_(this, request);
       }
     }
 
-    lock.unlock();
-
     // Were task(s) completed? If so send signal in case the main
     // thread is waiting
     if (has_completed) {
       cv_.notify_all();
```
GuanLuo: same as in grpc

src/clients/c++/library/request_grpc.cc
```diff
@@ -1073,26 +1046,41 @@ InferGrpcStreamContextImpl::Run(ResultMap* results)
...
 
 Error
 InferGrpcStreamContextImpl::AsyncRun(
     std::shared_ptr<Request>* async_request, OnCompleteFn callback)
 {
   GrpcRequestImpl* current_context;
+  current_context =
+      new GrpcRequestImpl(async_request_id_++, std::move(callback));
+  async_request->reset(static_cast<Request*>(current_context));
+  uintptr_t run_index = current_context->Id();
```
deadeyegoodwin: Don't use run_index as the object that grpc returns in AsyncTransfer. Just use the shared_ptr<Request> object directly. That should allow you to not need to use ongoing_async_requests_

src/clients/c++/library/request_http.cc
```diff
@@ -1351,24 +1344,12 @@ InferHttpContextImpl::AsyncRun(
...
       std::static_pointer_cast<HttpRequestImpl>(async_request);
 
-  {
-    std::lock_guard<std::mutex> lock(mutex_);
-    ongoing_async_requests_.erase(http_request->RunIndex());
-    curl_multi_remove_handle(multi_handle_, http_request->easy_handle_);
-  }
-
-  err = UpdateStat(http_request->Timer());
+  Error err = UpdateStat(http_request->Timer());
```
deadeyegoodwin: This should happen in AsyncTransfer when the response is received and REQUEST_END timestamp is captured.

src/clients/c++/library/request_http.cc
deadeyegoodwin: Remove. Main thread doesn't wait on cv_

src/clients/python/__init__.py
```diff
@@ -1481,6 +1412,12 @@ def async_run_with_cb(self, callback, inputs, outputs, batch_size=1, flags=0):
             The batch size of the inference. Each input must provide
             an appropriately sized batch of inputs.
 
+        corr_id : int
+            The correlation id of the inference. If non-zero this
+            correlation ID overrides the context's correlation ID for
+            all subsequent inference requests, else the context retains
```
deadeyegoodwin: ".. else the inference request uses the context's correlation ID"

src/clients/python/image_client.py
```diff
@@ -271,6 +285,8 @@ def postprocess(results, filenames, batch_size):
     request_ids = []
     image_idx = 0
     last_request = False
+    user_data = UserData()
+    i=0
```
deadeyegoodwin: sent_count or similar

src/clients/c++/library/request_grpc.cc
tanmayv25: done

src/clients/c++/library/request_grpc.cc
tanmayv25: Fixed

src/clients/c++/library/request_grpc.cc
```diff
@@ -977,23 +961,19 @@ InferGrpcContextImpl::AsyncTransfer()
...
-        }
+      // GRPC ensures that only a single polling thread is woken per tag_id
+      std::shared_ptr<GrpcRequestImpl> grpc_request =
+          std::static_pointer_cast<GrpcRequestImpl>(request);
+      grpc_request->Timer().CaptureTimestamp(RequestTimers::Kind::REQUEST_END);
+      if (grpc_request->HasCallback()) {
+        grpc_request->callback_(this, request);
       }
       // send signal in case the main thread is waiting
       cv_.notify_all();
```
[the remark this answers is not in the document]
tanmayv25: Fixed

src/clients/c++/library/request_grpc.cc
```diff
@@ -977,23 +961,19 @@ InferGrpcContextImpl::AsyncTransfer()
...
-            RequestTimers::Kind::REQUEST_END);
-        grpc_request->SetIsReady(true);
-        if (grpc_request->HasCallback()) {
-          request_with_callback = itr->second;
-        }
+      // GRPC ensures that only a single polling thread is woken per tag_id
+      std::shared_ptr<GrpcRequestImpl> grpc_request =
+          std::static_pointer_cast<GrpcRequestImpl>(request);
+      grpc_request->Timer().CaptureTimestamp(RequestTimers::Kind::REQUEST_END);
+      if (grpc_request->HasCallback()) {
```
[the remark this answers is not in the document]
tanmayv25: removed the explicit function but still need to validate if the user hasn't given nullptr as the callback function.

src/clients/c++/library/request_grpc.cc
tanmayv25: Thanks! Done.

src/clients/c++/library/request_grpc.cc
```diff
@@ -1144,23 +1139,19 @@ InferGrpcStreamContextImpl::AsyncTransfer()
...
-      grpc_request->Timer().CaptureTimestamp(RequestTimers::Kind::REQUEST_END);
-      grpc_request->SetIsReady(true);
-      if (grpc_request->HasCallback()) {
-        request_with_callback = itr->second;
-      }
+    std::shared_ptr<GrpcRequestImpl> grpc_request =
+        std::static_pointer_cast<GrpcRequestImpl>(request);
+    grpc_request->grpc_response_->Swap(&response);
+    grpc_request->Timer().CaptureTimestamp(RequestTimers::Kind::REQUEST_END);
+    if (grpc_request->HasCallback()) {
```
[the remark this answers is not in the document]
tanmayv25: fixed

src/clients/c++/library/request_http.cc
```diff
@@ -1585,13 +1566,9 @@ InferHttpContextImpl::AsyncTransfer()
...
-      // wake up if at least one request is not ready
-      for (auto& ongoing_async_request : this->ongoing_async_requests_) {
-        if (!std::static_pointer_cast<HttpRequestImpl>(
-                 ongoing_async_request.second)
-                 ->IsReady()) {
-          return true;
-        }
+      // wake up if an async request has been generated
+      if (!this->ongoing_async_requests_.empty()) {
+        return true;
```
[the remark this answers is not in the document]
tanmayv25: fixed

src/clients/c++/library/request_http.cc
```diff
@@ -1622,29 +1601,24 @@ InferHttpContextImpl::AsyncTransfer()
         http_request->Timer().CaptureTimestamp(
             RequestTimers::Kind::REQUEST_END);
       }
-
       http_request->http_status_ = msg->data.result;
-      http_request->SetIsReady(true);
+    }
+    lock.unlock();
```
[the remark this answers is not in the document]
tanmayv25: The worker and the runner threads can not use the same handle simultaneously. Hence, needed to have locks in entirety.

src/clients/c++/library/request_http.cc
tanmayv25: fixed

qa/L0_sequence_stress/sequence_stress.py
```diff
@@ -69,10 +83,9 @@ def check_sequence_async(ctx, trial, model_name, input_dtype, steps,
 
     # Execute the sequence of inference...
     seq_start_ms = int(round(time.time() * 1000))
-    input_values = list()
-    result_ids = list()
-    expected_results = list()
+    user_data = UserData()
 
+    i=0
```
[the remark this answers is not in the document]
tanmayv25: renamed

src/clients/c++/examples/image_client.cc
```diff
@@ -681,7 +684,13 @@ main(int argc, char** argv)
   std::vector<std::vector<std::string>> result_filenames;
   std::vector<std::shared_ptr<nic::InferContext::Request>> requests;
   size_t image_idx = 0;
+  size_t done_cnt = 0;
   bool last_request = false;
+  std::mutex mtx;
+  std::condition_variable cv;
+
+  std::queue<std::shared_ptr<nic::InferContext::Request>> request_buffer_;
```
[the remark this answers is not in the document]
tanmayv25: renamed

src/clients/c++/examples/simple_sequence_client.cc
```diff
@@ -164,9 +171,16 @@ AsyncSend(
...
-
-  return request;
+  FAIL_IF_ERR(
+      ctx->AsyncRun(
+          [&](nic::InferContext* ctx,
+              const std::shared_ptr<nic::InferContext::Request>& request) {
+            std::lock_guard<std::mutex> lk(mutex_);
+            request_list.push_back(request);
+            cv_.notify_all();
+            return;
```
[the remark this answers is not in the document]
tanmayv25: removed

src/clients/c++/library/request_grpc.cc
```diff
@@ -977,23 +954,19 @@ InferGrpcContextImpl::AsyncTransfer()
...
-        }
+      // GRPC ensures that only a single polling thread is woken per tag_id
+      std::shared_ptr<GrpcRequestImpl> grpc_request =
+          std::static_pointer_cast<GrpcRequestImpl>(request);
+      grpc_request->Timer().CaptureTimestamp(RequestTimers::Kind::REQUEST_END);
+      if (grpc_request->HasCallback()) {
+        grpc_request->callback_(this, request);
       }
       // send signal in case the main thread is waiting
       cv_.notify_all();
```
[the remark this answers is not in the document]
tanmayv25: Fixed

src/clients/c++/library/request_grpc.cc
```diff
@@ -1073,26 +1046,41 @@ InferGrpcStreamContextImpl::Run(ResultMap* results)
...
+  std::mutex mtx;
+  std::condition_variable cv;
+  Error err =
+      AsyncRun([&](InferContext* ctx, std::shared_ptr<Request> request) {
+        // Defer the response retrieval to main thread
+        std::lock_guard<std::mutex> lk(mtx);
+        callback_invoked = true;
+        req = std::move(request);
+        cv.notify_all();
+        return;
```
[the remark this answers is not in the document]
tanmayv25: using future and promises as Guan suggested.

src/clients/c++/library/request_grpc.cc
```diff
@@ -1073,26 +1046,41 @@ InferGrpcStreamContextImpl::Run(ResultMap* results)
...
+  bool callback_invoked = false;
+  std::mutex mtx;
+  std::condition_variable cv;
+  Error err =
+      AsyncRun([&](InferContext* ctx, std::shared_ptr<Request> request) {
+        // Defer the response retrieval to main thread
+        std::lock_guard<std::mutex> lk(mtx);
+        callback_invoked = true;
+        req = std::move(request);
+        cv.notify_all();
```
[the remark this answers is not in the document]
tanmayv25: done

src/clients/c++/library/request_grpc.cc
tanmayv25: Unfortunately, grpc ClientReaderWriter sync_stream (current service implementation) interface does not allow sending void* pointer to Write and Read interface. ClientAsyncReaderWriter async_stream.h  on the other hand does allow sending and receiving void* tag.

I have removed the ongoing requests from the non-streaming case but not yet for streaming case. The order of the requests and repsonses should be the same for streaming, with this assumption we might be able to replace ongoing_async_requests_ (with a loss in generality) with a lock-free queue storing the callbacks . Shall I do that?

src/clients/c++/library/request_http.cc
tanmayv25: Moved

src/clients/c++/library/request_http.cc
tanmayv25: Fixed

src/clients/python/__init__.py
tanmayv25: fixed

src/clients/python/image_client.py
tanmayv25: renamed

src/clients/c++/library/request_grpc.cc
deadeyegoodwin: I wouldn't since we may be using this same client for the future unordered response case. For now just keep ongoing_async_requests for streaming case.

src/clients/c++/library/request_grpc.cc
```diff
@@ -938,61 +900,27 @@ void
...
-      // send signal in case the main thread is waiting
-      cv_.notify_all();
-      if (request_with_callback != nullptr) {
-        GrpcRequestImpl* request =
-            static_cast<GrpcRequestImpl*>(request_with_callback.get());
-        request->callback_(this, request_with_callback);
+    } else if (grpc_request_ptr == nullptr) {
+       fprintf(stderr, "Unexpected null tag received at client.\n");
+      } else {
+      grpc_request_ptr->Timer().CaptureTimestamp(
```
deadeyegoodwin: You should put grpc_request_ptr into a shared_ptr as soon as you know it is not nullptr (in fact you could do it immeidately after the Next() call above, it doesn't matter if it is null or not. As it is now you have a leak when the callback_ is nullptr.

src/clients/c++/library/request_grpc.cc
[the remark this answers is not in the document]
GuanLuo: You can check that when the user initiates the request (when the user provides callback), shouldn't need to wait until here.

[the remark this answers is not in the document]
GuanLuo: I think it is better to check early, we shouldn't send a request if the user has no incentive to collect the result.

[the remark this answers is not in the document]
deadeyegoodwin: I'd even say that we should require a non-null callback, and have the AsynRun call return an error if that is not the case.

PR #791, opened 2019-10-24
Add model warm up option

GuanLuo:


nieksand:
Cool!

src/backends/pytorch/libtorch_backend.cc
```diff
@@ -175,6 +175,10 @@ LibTorchBackend::CreateExecutionContexts(
         Run(runner_idx, payloads, func);
       }));
 
+  if (Config().has_model_warm_up()) {
+    RETURN_IF_ERROR(WarmUp());
```
CoderHam: Based on the thread based caching we see here (https://github.com/pytorch/pytorch/issues/27902), it is possible that this will not be reliable. We should document or call this out somewhere.

docs/model_configuration.rst
```diff
@@ -555,3 +555,20 @@ or when it first receives inference requests. TensorRT optimizations
...
+.. _section-model-warm-up:
+
+Model Warm Up
+-------------------
+
+For some framework backends, model initialization may be delayed until the
+first inference is requested, TF-TRT optimization for example, which introduces
+unexpected latency seen by the client. The model configuration
+:cpp:var:`ModelWarmUp <nvidia::inferenceserver::ModelWarmUp>` is used to specify
+warm up settings for a model. The settings define an sample inference request
```
deadeyegoodwin: "The settings define the inference request that the inference server should create to warm-up each model instance."

qa/L0_model_config/autofill_noplatform/ensemble/model_warm_up_set/expected
```diff
@@ -0,0 +1 @@
+model warm up should not be specified for ensemble 'model_warm_up_set'
```
deadeyegoodwin: model_warm_up can not be specified...

qa/L0_warmup/test.sh
```diff
@@ -0,0 +1,112 @@
...
+SERVER=/opt/tensorrtserver/bin/trtserver
+SERVER_ARGS="--model-repository=$DATADIR --log-verbose=1 --exit-timeout-secs=120"
+SERVER_LOG="./inference_server.log"
+source ../common/util.sh
+
+rm -f $SERVER_LOG
+
+RET=0
+
+# [TODO] use a model that takes time for first time inference and compare
```
deadeyegoodwin: If you use the ONNX inception_v1 model and enable ONNX-TRT optimization it takes quite a while to load.

docs/model_configuration.rst
```diff
@@ -555,3 +555,20 @@ or when it first receives inference requests. TensorRT optimizations
 include specializing and fusing model layers, and using reduced
 precision (for example 16-bit floating-point) to provide significant
 throughput and latency improvements.
+
+.. _section-model-warm-up:
+
+Model Warm Up
```
deadeyegoodwin: Warmup is one word so throughout should change

src/backends/caffe2/netdef_backend.cc
```diff
@@ -137,6 +137,10 @@ NetDefBackend::CreateExecutionContexts(
         Run(runner_idx, payloads, func);
       }));
 
+  if (Config().has_model_warm_up()) {
```
deadeyegoodwin: I think we will need to have warmup as one of the "phases" for SetConfiguredScheduler because we need to init to happen before warmup and also need the warmup runs to be trigger by the scheduler thread (as normal runs are done).

src/core/model_config.proto
```diff
@@ -909,4 +935,13 @@ message ModelConfig
...
   map<string, ModelParameter> parameters = 14;
+
+  //@@  .. cpp:var:: ModelWarmUp model_warm_up
+  //@@
+  //@@     Warm up setting of this model. If specified, all instances
+  //@@     will be run with a request sample before serving the model.
+  //@@     This field can only be specified if the model is not an ensemble
+  //@@     model.
+  //@@
+  ModelWarmUp model_warm_up = 16;
```
deadeyegoodwin: I think this should be "repeated" so that users can specify multiple warmup trials... for example, with different input dimensions.

src/core/model_config.proto
```diff
@@ -764,6 +766,30 @@ message ModelParameter
...
+//@@
+message ModelWarmUp
+{
+  //@@  .. cpp:var:: InferRequestHeader request_header
+  //@@
+  //@@     The header message of the request sample.
+  //@@
+  InferRequestHeader request_header = 1;
+
+  //@@  .. cpp:var:: bool use_zero_value
```
deadeyegoodwin: We should have a oneof with a message for each of the initialization. Like here: https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/core/grpc_service.proto#L292.

And I would start with just zero and random. Actual data case can be done but we should discuss.

src/core/model_config.proto
```diff
@@ -764,6 +766,30 @@ message ModelParameter
...
 }
 
+//@@
+//@@.. cpp:var:: message ModelWarmUp
+//@@
+//@@   Settings for warming up the model instances.
+//@@
+message ModelWarmUp
+{
+  //@@  .. cpp:var:: InferRequestHeader request_header
```
deadeyegoodwin: Give a name... will be useful for logging and I think it will also be necessary for the "actual data" case.

src/core/model_config.proto
```diff
@@ -764,6 +766,30 @@ message ModelParameter
...
+  //@@
+  //@@     The header message of the request sample.
```
deadeyegoodwin: "The request meta-data to use for the warm-up inference request."
But I'm actually not sure we should do this. There is a lot of information in the request header that just doesn't apply here. Really all we need are the shapes for inputs that have variable-size dimensions (other than batch dimension). So can we just have a map from tensor name to dims (and even that is required only for inputs that need it). We should be able to generate the rest of the meta-data.

qa/L0_warmup/test.sh
GuanLuo: I believe ONNX-TRT optimization is done during model loading, not when the first inference is initiated. We need TF-TRT optimization instead, but I remember that the inception model, the one in example model repository, is not working with TF-TRT?

src/backends/caffe2/netdef_backend.cc
GuanLuo: Do you mean just moving this part into `SetConfiguredScheduler()`? Or actually sending the sample requests to backend context via `Scheduler::Enqueue()`? I wasn't doing the latter because we don't have strong guarantee that all contexts will be run if the request is dispatched by scheduler.

src/core/backend.cc
```diff
@@ -189,4 +235,217 @@ InferenceBackend::Run(
   OnCompleteQueuedPayloads(status);
 }
 
+Status
+InferenceBackend::GenerateWarmupData(std::vector<WarmupData>* samples)
+{
+  static std::string warmup_data_folder = "warmup";
```
deadeyegoodwin: This should go in constants.h

src/core/backend.cc
```diff
@@ -189,4 +235,217 @@ InferenceBackend::Run(
...
+  samples->clear();
+  for (const auto& warmup_setting : config_.model_warmup()) {
+    LOG_VERBOSE(1) << "Generating warmup sample data for '"
+                   << warmup_setting.name() << "'";
+
+    // Two passes. First pass to get max byte size for synthetic data
+    int64_t max_zero_byte_size = 0;
+    int64_t max_random_byte_size = 0;
+    for (const auto& input_meta : warmup_setting.inputs()) {
+      auto element_count = GetElementCount(input_meta.second.dims());
```
deadeyegoodwin: There's a GetByteSize that takes element count and datatype... can you use that?

src/core/backend.cc
```diff
@@ -189,4 +235,217 @@ InferenceBackend::Run(
...
+    warmup_data.zero_data_.reset(new AllocatedSystemMemory(
+        max_zero_byte_size, TRTSERVER_MEMORY_CPU /* memory_type */,
+        0 /* memory_type_id */));
+    char* zero_buffer = warmup_data.zero_data_->MutableBuffer(&type, &type_id);
+    memset(zero_buffer, 0, max_zero_byte_size);
+
+    warmup_data.random_data_.reset(new AllocatedSystemMemory(
+        max_random_byte_size, TRTSERVER_MEMORY_CPU /* memory_type */,
+        0 /* memory_type_id */));
+    char* random_buffer =
```
deadeyegoodwin: Where do you actually set random data into this buffer?

src/core/model_config.proto
```diff
@@ -764,6 +764,86 @@ message ModelParameter
...
 }
 
+//@@
+//@@.. cpp:var:: message ModelWarmup
+//@@
+//@@   Settings used to construct the request sample for model warmup.
+//@@
+message ModelWarmup
+{
+  message InputMetaData
```
deadeyegoodwin: Just call it "Input"

src/core/model_config.proto
```diff
@@ -764,6 +764,86 @@ message ModelParameter
...
+    //@@     data type and 'random_data' is set, the data generation will fall
+    //@@     back to 'zero_data'.
+    //@@
+    oneof input_data_type
+    {
+      //@@
+      //@@  .. cpp:var:: bool zero_data
+      //@@
+      //@@     The identifier for using zeros as input data. Note that the
+      //@@     field data will not be checked, instead, zero data will be used
```
deadeyegoodwin: What does "field data" mean?

deadeyegoodwin:
@GuanLuo can you file a separate ticket to add more coverage for warmup. I think we need to test string datatype more directly.

src/core/backend.cc
GuanLuo: I didn't use `GetByteSize()` since the element count will be needed for the STRING data type

src/core/model_config.proto
GuanLuo: I should have just said "value of 'zero_data'"

PR #798, opened 2019-10-29
Add test on covering different instance group with various I/O memory type and id

GuanLuo:


src/servers/CMakeLists.txt
```diff
@@ -229,6 +229,43 @@ install(
   RUNTIME DESTINATION bin
 )
 
+#
+# io_test_util (only build when TRTIS_ENABLE_GPU given the purpose)
```
deadeyegoodwin: It is ok to only build when GPU enabled, but let's call it something non-test related since I think we can also use it as an example in the documentation at some point. Ideally make an examples/ sub-directory and put simple.cc in there and also this file and call it memory.cc or memory_alloc.cc.

src/servers/io_test_util.cc
```diff
@@ -0,0 +1,697 @@
...
+    if (err != cudaSuccess) {                                             \
+      LOG_ERROR << "error: " << (MSG) << ": " << cudaGetErrorString(err); \
+      exit(1);                                                            \
+    }                                                                     \
+  } while (false)
+
+
+static auto gpu_data_deleter = [](void* data) {
+  if (data != nullptr) {
+    auto err = cudaSetDevice(io_spec.input_type_id_);
```
deadeyegoodwin: Can we just FAIL_IF_CUDA_ERROR for these?

src/servers/io_test_util.cc
```diff
@@ -0,0 +1,697 @@
...
+    void* allocated_ptr = nullptr;
+    if (memory_type == TRTSERVER_MEMORY_CPU) {
+      allocated_ptr = malloc(byte_size);
+    } else if (io_spec.output_type_ == TRTSERVER_MEMORY_GPU) {
+      auto err = cudaSetDevice(memory_type_id);
+      if (err == cudaSuccess) {
+        err = cudaMalloc(&allocated_ptr, byte_size);
+      }
+      if (err != cudaSuccess) {
+        LOG_INFO << "cudaMalloc failed: " << cudaGetErrorString(err);
```
deadeyegoodwin: Should just return a TRTSERVER_Error here.

src/servers/io_test_util.cc
```diff
@@ -0,0 +1,697 @@
...
+        allocated_ptr = nullptr;
+      }
+    }
+
+    if (allocated_ptr != nullptr) {
```
deadeyegoodwin: If allocated_ptr is nullptr here it means allocation failed so should return an appropriate error code.

src/servers/io_test_util.cc
```diff
@@ -0,0 +1,697 @@
...
+
+  return nullptr;  // Success
+}
+
+TRTSERVER_Error*
+ResponseRelease(
+    TRTSERVER_ResponseAllocator* allocator, void* buffer, void* buffer_userp,
+    size_t byte_size, TRTSERVER_Memory_Type memory_type, int64_t memory_type_id)
+{
+  std::string* name = nullptr;
```
deadeyegoodwin: Use shared_ptr/unique_ptr to ensure deallocation when adding early exits below

src/servers/io_test_util.cc
```diff
@@ -0,0 +1,697 @@
...
+  }
+
+  LOG_INFO << "Releasing buffer " << buffer << " of size " << byte_size
+           << " in " << MemoryTypeString(memory_type) << " for result '"
+           << *name << "'";
+  if (memory_type == TRTSERVER_MEMORY_CPU) {
+    free(buffer);
+  } else if (io_spec.output_type_ == TRTSERVER_MEMORY_GPU) {
+    auto err = cudaSetDevice(memory_type_id);
+    if (err == cudaSuccess) {
```
deadeyegoodwin: Just return TRTSERVER_Error for these failure cases.

src/servers/io_test_util.cc
```diff
@@ -0,0 +1,697 @@
...
+    LOG_INFO << "'" + model_name + "' model is "
+             << ni::ModelReadyState_Name(vitr->second.ready_state());
+    if (vitr->second.ready_state() == ni::ModelReadyState::MODEL_READY) {
+      FAIL_IF_ERR(
+          ParseModelConfig(itr->second.config(), &is_int, &is_torch_model),
+          "parsing model config");
+      break;
+    }
+
+    // [TODO] do so before break
```
deadeyegoodwin: Fix this?

src/core/ensemble_scheduler.cc
```diff
@@ -664,9 +664,13 @@ EnsembleContext::CheckAndSetEnsembleOutput()
...
     void* buffer;
-    RETURN_IF_ERROR(response_provider_->AllocateOutputBuffer(
+    // Don't return on error if we haven't try all types
+    auto status = response_provider_->AllocateOutputBuffer(
         output_pair.first, &buffer, expected_byte_size, shape, dst_memory_type,
-        memory_type_id));
+        memory_type_id);
+    if (dst_memory_type == TRTSERVER_MEMORY_CPU) {
+      return status;
+    }
```
GuanLuo: Here and above is just for correctness, "still tries other memory type if the current one returns error or nullptr". This will be changed after Hemant's PR is merged.

PR #804, opened 2019-10-30
Clarify why ONNX is not in L0_nullchar_string

GuanLuo:
Not going to run CI as it is just documentation change.

### DIFF
--- a/qa/L0_nullchar_string/test.sh
+++ b/qa/L0_nullchar_string/test.sh
@@ -59,7 +59,9 @@ RET=0
 
 set +e
 
-# Add onnx_nobatch_zero_1_object once Onnx backend is fixed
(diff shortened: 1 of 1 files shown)


qa/L0_nullchar_string/test.sh
```diff
@@ -59,7 +59,9 @@ RET=0
 
 set +e
 
-# Add onnx_nobatch_zero_1_object once Onnx backend is fixed
+# Ignore ONNX backend because even though ONNX supports string data type,
+# strings that contain null character in the middle is not considered.
```
CoderHam: supported instead of considered?

deadeyegoodwin: I would say "not allowed"

PR #805, opened 2019-10-30
Fix race condition in Python async client

deadeyegoodwin:


src/clients/python/crequest.cc
```diff
@@ -724,6 +733,8 @@ InferContextAsyncResultNew(
 {
   InferContextResultCtx* lctx = new InferContextResultCtx;
```
GuanLuo: Not related to this change, but this should be managed by smart pointer in case of early return.

deadeyegoodwin: Fixed by moving the creation till after the early returns.

PR #806, opened 2019-10-30
Make sure to create the backend context CUDA stream on the correct de…

deadeyegoodwin:
…vice


deadeyegoodwin:
@GuanLuo it seems like without this change we would likely always be creating streams on device 0 for the backend_context (except plan which I think will have already set device correctly).

PR #812, opened 2019-11-01
Clean up InferenceBackend

GuanLuo:
This is part of adding warmup functionality, the main purpose is to expose `Run()` in base class so that warmup functionality can utilize it without knowing underlying backends, and to remove redundant code.


deadeyegoodwin:
@GuanLuo You need to manually rebase.

PR #822, opened 2019-11-05
Synchronous API support in perf_client

tanmayv25:
Besides adding support for synchronous API, added the timestamp reporting in the callback of AsyncRun() call for more precise latency values.

### DIFF
--- a/src/clients/c++/perf_client/concurrency_manager.h
+++ b/src/clients/c++/perf_client/concurrency_manager.h
@@ -112,10 +112,7 @@ class ConcurrencyManager : public LoadManager {
     InferContextMetaData(const InferContextMetaData&) = delete;
 
     std::unique_ptr<nic::InferContext> ctx_;
-    // mutex to guard 'inflight_request_cnt_' which will be acessed by
-    // both the main thread and callback thread
-    std::mutex mtx_;
-    size_t inflight_request_cnt_;
+    std::atomic<size_t> inflight_request_cnt_;
   };
 
  private:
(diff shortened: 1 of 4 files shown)


src/clients/c++/perf_client/concurrency_manager.h
```diff
@@ -108,28 +112,39 @@ class ConcurrencyManager : public LoadManager {
     InferContextMetaData(const InferContextMetaData&) = delete;
 
     std::unique_ptr<nic::InferContext> ctx_;
-    size_t inflight_request_cnt_;
-    // mutex to guard 'completed_requests_' which will be acessed by
+    // mutex to guard 'inflight_request_cnt_' which will be acessed by
     // both the main thread and callback thread
     std::mutex mtx_;
-    std::vector<RequestMetaData> completed_requests_;
+    size_t inflight_request_cnt_;
```
deadeyegoodwin: Can this just be an atomic?

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -286,6 +286,7 @@ Usage(char** argv, const std::string& msg = std::string())
   std::cerr << "\t-v" << std::endl;
   std::cerr << std::endl;
   std::cerr << "I. MEASUREMENT PARAMETERS: " << std::endl;
+  std::cerr << "\t--async (-a)" << std::endl;
```
deadeyegoodwin: For sequence models in sync mode... does that mean that only one request for a sequence is made at a time?

src/clients/c++/perf_client/concurrency_manager.h
tanmayv25: Sure

src/clients/c++/perf_client/perf_client.cc
tanmayv25: Yes. This is because we are restricted with our blocking synchronous API. I have modified the default mode of operation to asynchronous for sequence models.

PR #825, opened 2019-11-06
Add optional end-sequence control to sequence batcher

deadeyegoodwin:


src/core/sequence_batch_scheduler.cc
```diff
@@ -762,12 +826,22 @@ SequenceBatchScheduler::SequenceBatch::SchedulerThread(
...
-              // the appropriate sequence start indicator to the
-              // backend.
+              // Set the start, end, and ready control tensors
+              // appropriately...
               if ((request_header.flags() &
+                   (InferRequestHeader::FLAG_SEQUENCE_START |
+                    InferRequestHeader::FLAG_SEQUENCE_END)) ==
+                  (InferRequestHeader::FLAG_SEQUENCE_START |
+                   InferRequestHeader::FLAG_SEQUENCE_END)) {
+                request_provider->SetInputOverride(startend_input_overrides_);
```
GuanLuo: Is this for the case where the "sequence length" is 1?

deadeyegoodwin: Yes. Both start and end flag set on the request.

PR #830, opened 2019-11-06
Extract CopyBuffer() as a common utility. Replace other copy logics with it.

GuanLuo:


deadeyegoodwin:
@GuanLuo You need to manually rebase. I think @CoderHam had some changes that I just merged that are no longer necessary with your changes.

GuanLuo:
Rebased. Re-ran CI to make sure my changes do the same as what Hemant was fixing.

CoderHam:
Yes the changes I made to the onnx backend should be the same as when using the CopyBuffer utility

PR #836, opened 2019-11-07
Add libtorch back to L0_sequence_batcher

CoderHam:
libtorch was removed from trials earlier. Adding it back.


deadeyegoodwin:
@CoderHam This is still causing a segfault so we will not take it for 19.11. You should file a ticket for 19.12 so we can figure out what is going wrong here.

CoderHam:
Sure, I'll do that

PR #845, opened 2019-11-08
Change verbose messages so they don't seem like errors

deadeyegoodwin:


src/servers/grpc_server.cc
```diff
@@ -679,26 +679,26 @@ InferResponseAlloc(
         *actual_memory_type_id = pr->second.device_id_;
         use_shm = true;
 
-        LOG_VERBOSE(1) << "GRPC shared-memory: " << tensor_name << ", size "
-                       << byte_size << ", addr " << *buffer;
+        LOG_VERBOSE(1) << "GRPC: using shared-memory for " << tensor_name
```
CoderHam: How about wrapping tensor name in single quotes and using a ":" after size and addr ?

src/servers/grpc_server.cc
```diff
@@ -679,26 +679,26 @@ InferResponseAlloc(
...
     if (!use_shm) {
       // Can't allocate for any memory type other than CPU. If asked to
       // allocate on GPU memory then force allocation on CPU instead.
       if (*actual_memory_type != TRTSERVER_MEMORY_CPU) {
-        LOG_VERBOSE(1) << "GRPC allocation failed for type "
-                       << *actual_memory_type << " for " << tensor_name
-                       << ", will use type " << TRTSERVER_MEMORY_CPU;
+        LOG_VERBOSE(1)
+            << "GRPC: unable to provide " << tensor_name
+            << " in requested memory type, will use type TRTSERVER_MEMORY_CPU";
```
CoderHam: Since this is only triggered for GPU, it means we can replace  requested memory type with TRTSERVER_MEMORY_GPU

src/servers/grpc_server.cc
```diff
@@ -679,26 +679,26 @@ InferResponseAlloc(
...
         *actual_memory_type = TRTSERVER_MEMORY_CPU;
         *actual_memory_type_id = 0;
       }
 
       raw_output->resize(byte_size);
       *buffer = static_cast<void*>(&((*raw_output)[0]));
 
-      LOG_VERBOSE(1) << "GRPC allocation: " << tensor_name << ", size "
+      LOG_VERBOSE(1) << "GRPC: using buffer for " << tensor_name << ", size "
```
CoderHam: Use ":" after size and addr

src/servers/http_server.cc
```diff
@@ -363,9 +363,9 @@ HTTPAPIServer::ResponseAlloc(
     } else {
       // Can't allocate for any memory type other than CPU.
       if (preferred_memory_type != TRTSERVER_MEMORY_CPU) {
-        LOG_VERBOSE(1) << "HTTP allocation failed for type "
-                       << preferred_memory_type << " for " << tensor_name
-                       << ", will use type " << TRTSERVER_MEMORY_CPU;
+        LOG_VERBOSE(1)
+            << "HTTP: unable to provide " << tensor_name
+            << " in requested memory type, will use type TRTSERVER_MEMORY_CPU";
```
CoderHam: Same as GRPC, replace requested memory type with TRTSERVER_MEMORY_GPU

PR #847, opened 2019-11-08
Add L0_sequence_batcher test for cuda shared memory

CoderHam:


qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -104,6 +106,59 @@ def check_deferred_exception(self):
             if len(_deferred_exceptions) > 0:
                 raise _deferred_exceptions[0]
 
+    def precreate_register_regions(self, value_list, dtype, i, batch_size=1):
```
deadeyegoodwin: Are these the same as the ones in infer-utils? Can we just reuse those?

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -247,11 +303,23 @@ def check_sequence(self, trial, model_name, input_dtype, correlation_id,
                                         "sequence expected greater than " + str(gt_ms) +
                                         "ms response time, got " + str(seq_end_ms - seq_start_ms) + " ms")
             except Exception as ex:
+                print("zz", str(ex))
```
deadeyegoodwin: remove

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -399,6 +426,7 @@ def check_sequence_async(self, trial, model_name, input_dtype, correlation_id,
...
```
deadeyegoodwin: remove

qa/L0_sequence_batcher/sequence_batcher_test.py
CoderHam: The ones in the infer utils are specific to that case. In the sequence batcher the number of values may be variable and the function above takes care of the same to create, set and unregister all shared memory regions together to avoid any thread specific failures and make it cleaner.

PR #849, opened 2019-11-11
Add CentOS client Dockerfile

aleksa2808:
Add a Dockerfile which builds the client libraries and examples for a CentOS 7 system.
Also includes a few minor changes to CMake files in order to make them more portable.

### DIFF
--- a/Dockerfile.centos_client
+++ b/Dockerfile.centos_client
@@ -1,0 +1,89 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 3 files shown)


src/clients/c++/examples/CMakeLists.example
```diff
@@ -26,6 +26,8 @@
 
 cmake_minimum_required (VERSION 3.5)
 
+set(CMAKE_CXX_STANDARD 11)
```
aleksa2808: Unlike nvidia/cuda:10.1-devel-ubuntu-18, nvidia/cuda:10.1-devel-centos7 image contains a version of GCC (4.8.5) which doesn't default to a >=c++11 standard. This makes some files (eg. mutex implementations used by protobuf) to throw an error specifying that the standard of at least c++11 needs to be declared.

build/CMakeLists.txt
```diff
@@ -116,11 +116,7 @@ ExternalProject_Add(protobuf
 
 # Location where protobuf-config.cmake will be installed varies by
 # platform
-if (WIN32)
-  set(_FINDPACKAGE_PROTOBUF_CONFIG_DIR "${CMAKE_CURRENT_BINARY_DIR}/protobuf/cmake")
-else()
-  set(_FINDPACKAGE_PROTOBUF_CONFIG_DIR "${CMAKE_CURRENT_BINARY_DIR}/protobuf/lib/cmake/protobuf")
-endif()
+set(_FINDPACKAGE_PROTOBUF_CONFIG_DIR "${CMAKE_CURRENT_BINARY_DIR}/protobuf/${CMAKE_INSTALL_LIBDIR}/cmake/protobuf")
```
aleksa2808: This is almost the same change made in the PR #742, which was reverted as part of PR #763, possibly by rebasing.

Dockerfile.centos_client
```diff
@@ -0,0 +1,89 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Default setting is building on nvidia/cuda:10.1-devel-centos7
+ARG BASE_IMAGE=nvidia/cuda:10.1-devel-centos7
```
GuanLuo: This Dockerfile seems to be identical with `Dokerfile.client` except for the OS part in image name. Perhaps it can just be extension on `Dockerfile.client`, which takes OS as optional argument?

deadeyegoodwin:
@aleksa2808 The windows build did fail. You still need the conditional to check for WIN32. Notice how the path is not the same even when $CMAKE_INSTALL_LIBDIR expands to empty string.

Dockerfile.centos_client
aleksa2808: I thought about, as I tried to make the script as similar as possible. But, there are a few differences which, I think, could make a unified Dockerfile a bit messy:
- DEBIAN_FRONTEND environment variable (although having this variable at all seems [discouraged](https://github.com/moby/moby/issues/4032#issuecomment-34597177))
- apt-get / yum difference and the packages that they install
- cmake3 -> cmake rename
- the `.` at the end of the cmake command (this just removes a warning in the centos version)
- the `--ignore-installed` flag in the pip install command, which prevents an error in the centos version caused by numpy already being installed

The apt-get / yum difference seems like the biggest issue here. Do you have a suggestion for how this could be done in a unified Dockerfile?

aleksa2808:
@deadeyegoodwin yes, you're right. I changed the if block as per @tanmayv25's suggestion, by leaving the WIN32 part as it was and putting the current line in the else branch, which will cover Ubuntu and CentOS.

PR #853, opened 2019-11-11
Add correalation-ID control to sequence batcher

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo Yes, some use cases the backends want to lookup state via the correlation ID.

PR #854, opened 2019-11-11
Fix race in grpc streaming writes

deadeyegoodwin:


src/servers/grpc_server.cc
```diff
@@ -1825,15 +1809,15 @@ GRPCServer::Start()
       "InferHandler", server_, server_id_, trace_manager_, smb_manager_,
       &service_, infer_cq_.get(),
       infer_allocation_pool_size_ /* max_state_bucket_count */);
-  hinfer->Start(infer_thread_cnt_);
+  hinfer->Start(/* infer_thread_cnt_ FIXME */ 1);
```
tanmayv25: It looks like you missed reverting these.

src/servers/grpc_server.cc
```diff
@@ -1825,15 +1809,15 @@ GRPCServer::Start()
...
   infer_handler_.reset(hinfer);
 
   // Handler for streaming inference requests.
   StreamInferHandler* hstreaminfer = new StreamInferHandler(
       "StreamInferHandler", server_, server_id_, trace_manager_, smb_manager_,
       &service_, stream_infer_cq_.get(),
       infer_allocation_pool_size_ /* max_state_bucket_count */);
-  hstreaminfer->Start(stream_infer_thread_cnt_);
+  hstreaminfer->Start(/* stream_infer_thread_cnt_ FIXME */ 1);
```
tanmayv25: same as above.

src/servers/grpc_server.cc
deadeyegoodwin: Yes, I need to file a separate ticket for these...

PR #861, opened 2019-11-12
Add python shared memory string client examples

CoderHam:
- [x] Add `simple_cuda_shm_string_client.py`
- [x] Add `simple_cuda_shm_string_client.py`
- [x] Update tests for _L0_simple_shared_memory_example_ and _L0_simple_cuda_shared_memory_example_


src/clients/python/cuda_shared_memory/__init__.py
```diff
@@ -73,6 +73,54 @@ def _raise_error(msg):
     ex = CudaSharedMemoryException(msg)
     raise ex
 
+def serialize_string_tensor(input_tensor):
+    """Creates a shared memory region with the specified name and size.
```
deadeyegoodwin: This comment is wrong

src/clients/python/cuda_shared_memory/__init__.py
```diff
@@ -73,6 +73,54 @@ def _raise_error(msg):
...
+
+    Parameters
+    ----------
+    input_tensor : np.array
+        The string tensor to serialize.
+
+    Returns
+    -------
+    serialized_string_tensor : np.array
+        The serialized string tensor.
```
deadeyegoodwin: Say something about how the serialization is a tensor of type uint8 (or char)

src/clients/python/shared_memory/__init__.py
```diff
@@ -73,6 +73,55 @@ def _raise_error(msg):
     ex = SharedMemoryException(msg)
     raise ex
 
+def serialize_string_tensor(input_tensor):
+    """Creates a shared memory region with the specified name and size.
```
deadeyegoodwin: Fix

src/clients/python/shared_memory/__init__.py
```diff
@@ -73,6 +73,55 @@ def _raise_error(msg):
...
+
+    Parameters
+    ----------
+    input_tensor : np.array
+        The string tensor to serialize.
+
+    Returns
+    -------
+    serialized_string_tensor : np.array
+        The serialized string tensor.
```
deadeyegoodwin: Same fix as above

src/clients/python/cuda_shared_memory/__init__.py
CoderHam: Done

src/clients/python/cuda_shared_memory/__init__.py
CoderHam: Fixed

src/clients/python/shared_memory/__init__.py
CoderHam: Done

src/clients/python/shared_memory/__init__.py
CoderHam: Done

CoderHam:
> Looks ok but CI failed for both new examples so something is wrong somewhere

The failure was caused because I missed adding these shared memory clients in the cmake file

CoderHam:
> Looks ok but CI failed for both new examples so something is wrong somewhere

I had missed adding these in the cmake and they did not get copied successfully. Tested locally, re-ran CI for the same

PR #864, opened 2019-11-13
Add model control mode option to be more informative

GuanLuo:


src/servers/main.cc
```diff
@@ -265,18 +266,32 @@ std::vector<Option> options_{
...
+     "'repository-poll-secs'. For \"explicit\", the model load and unload is "
+     "initiated by using the model control APIs, and the models in the model "
+     "repository will not be loaded at startup, unless the model is specified "
+     "by --load-model."},
     {OPTION_ALLOW_POLL_REPO, "allow-poll-model-repository",
-     "Poll the model repository to detect changes. The poll rate is "
-     "controlled by 'repository-poll-secs'."},
+     "[DEPRECATED] Poll the model repository to detect changes. The poll rate "
+     "is controlled by 'repository-poll-secs'. This option is deprecated by "
+     "--model-control-mode, this option will be overridden if "
```
deadeyegoodwin: Here and elsewhere make the --allow options cause a failure if they are specified when --model-control-mode is specified.

src/servers/main.cc
```diff
@@ -265,18 +266,32 @@ std::vector<Option> options_{
     {OPTION_TRACE_RATE, "trace-rate",
      "Set the trace sampling rate. Default is 1000."},
 #endif  // TRTIS_ENABLE_TRACING
+    {OPTION_MODEL_CONTROL_MODE, "model-control-mode",
+     "Specify the mode for model management. Options are \"none\", \"poll\" "
+     "and \"explicit\". "
```
deadeyegoodwin: "The default is \"poll\". "

PR #873, opened 2019-11-13
Remove non-determinism from L0_shared_memory

deadeyegoodwin:


qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -80,40 +79,23 @@ def basic_inference(shm_ip0_handle, shm_ip1_handle, shm_op0_handle, shm_op1_hand
...
-for t in threads:
-    t.join()
+# Destroy shared memory then infer
+shm.destroy_shared_memory_region(shm_op0_handle)
+basic_inference(shm_ip0_handle, shm_ip1_handle, shm_op0_handle, shm_op1_handle, error_msg)
 if len(error_msg) > 1:
     raise Exception(error_msg[-1])
 
-# Register during inference - Should work fine
+# Register and inference - Should work fine
```
CoderHam: You can remove - should work fine

deadeyegoodwin:
@CoderHam In the ticket also note that the test should test both HTTP and GRPC APIs

PR #876, opened 2019-11-14
Supporting multiple model files in version directory

tanmayv25:
UPDATE : Actually we can load all files one be one and select the first passing. Visit all the files until we find the compatible files.

### DIFF
--- a/src/backends/caffe2/autofill.cc
+++ b/src/backends/caffe2/autofill.cc
@@ -71,42 +71,24 @@ AutoFillNetDef::Create(
 
   const auto version_path = JoinPath({model_path, *(version_dirs.begin())});
 
-  // There must be a single netdef model (which is spread across two
-  // files) within the version directory...
   std::set<std::string> netdef_files;
-  RETURN_IF_ERROR(GetDirectoryFiles(version_path, &netdef_files));
+  RETURN_IF_ERROR(GetDirectoryFiles(
+      version_path, true /* skip_hidden_files */, &netdef_files));
   const std::string expected_init_filename =
       std::string(kCaffe2NetDefInitFilenamePrefix) +
       std::string(kCaffe2NetDefFilename);
-  if (netdef_files.size() < 2) {
-    return Status(
-        RequestStatusCode::INTERNAL, "unable to autofill for '" + model_name +
-                                         "', unable to find netdef files: '" +
-                                         kCaffe2NetDefFilename + "' and '" +
-                                         expected_init_filename + "'");
-  } else if (netdef_files.size() > 2) {
-    return Status(
-        RequestStatusCode::INTERNAL,
-        "expects exactly 2 files in the version directory. Instead found: " +
-            std::to_string(netdef_files.size()) + " files.");
-  }
-
-  const std::string netdef0_file = *(netdef_files.begin());
-  const std::string netdef1_file = *(std::next(netdef_files.begin()));
 
   // If find both files named with the default netdef names then
   // assume it is a netdef. In the future we can be smarter here and
   // try to parse to see if it really is a netdef, and then try to
   // derive more of the configuration...
-  if (!(((netdef0_file == kCaffe2NetDefFilename) &&
-         (netdef1_file == expected_init_filename)) ||
-        ((netdef1_file == kCaffe2NetDefFilename) &&
-         (netdef0_file == expected_init_filename)))) {
+  if ((netdef_files.find(expected_init_filename) == netdef_files.end()) ||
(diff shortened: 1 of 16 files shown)


src/core/autofill.cc
```diff
@@ -219,8 +221,17 @@ AutoFill::Create(
   // Unable to determine the platform so just use the simple autofill,
   // or null if that fails.
   {
+    if (platform == Platform::PLATFORM_UNKNOWN) {
+      LOG_WARNING << "Autofiller failed to detect the platform (verify "
+                     "contents of version directory or use --log-verbose=1 for "
```
CoderHam: Shouldn't this be model directory instead of version directory? Also what happens when there is an extra file in the same directory as the version folders and config.pbtxt (say a .gitkeep)

tanmayv25: Fixed. Ignoring hidden files as well.

src/core/filesystem.cc
```diff
@@ -991,7 +991,15 @@ GetDirectoryFiles(const std::string& path, std::set<std::string>* files)
 {
   FileSystem* fs;
   RETURN_IF_ERROR(GetFileSystem(path, &fs));
-  return fs->GetDirectoryFiles(path, files);
+  std::set<std::string> all_files;
```
deadeyegoodwin: Don't do this unconditionally. Add a "skip_hidden_files = false" parameter to the function.

src/backends/caffe2/autofill.cc
```diff
@@ -78,12 +78,17 @@ AutoFillNetDef::Create(
...
       std::string(kCaffe2NetDefInitFilenamePrefix) +
       std::string(kCaffe2NetDefFilename);
-  if (netdef_files.size() != 2) {
+  if (netdef_files.size() < 2) {
     return Status(
         RequestStatusCode::INTERNAL, "unable to autofill for '" + model_name +
                                          "', unable to find netdef files: '" +
                                          kCaffe2NetDefFilename + "' and '" +
                                          expected_init_filename + "'");
+  } else if (netdef_files.size() > 2) {
```
deadeyegoodwin: Here and in the other autofillers you can't add this check. If you do then it is no longer possible to have multiple "models" in a version directory. This is generally supported and especially required for TensorRT (see model_control.proto cc_model_filenames). I think you don't want any of these changes in the autofill.cc files.

src/core/autofill.cc
```diff
@@ -219,8 +221,17 @@ AutoFill::Create(
   // Unable to determine the platform so just use the simple autofill,
   // or null if that fails.
   {
+    if (platform == Platform::PLATFORM_UNKNOWN) {
+      LOG_WARNING << "Autofiller failed to detect the platform (verify "
+                     "contents of model directory or use --log-verbose=1 for "
+                     "more details)";
+    } else {
+      LOG_WARNING << "Autofiller failed to retrieve model. Error Details: "
+                  << status.AsString();
```
GuanLuo: Isn't this duplicated with the verbose message above? i.e. line 212

src/backends/caffe2/autofill.cc
tanmayv25: This check was already implied by (netdef_files.size() != 2) or (plan_files.size() != 1). So, this doesn't change the current behavior of the autofiller. However, as stated that we can have multiple "models" in the version directory, the solution would be to make autofiller parse the instance_group for gpu device(if none then select device id =0), get the compute capability, parse the cc_model_filenames and get the specific model filename to autofill against. Do we want to add this support in the autofiller?

src/core/autofill.cc
tanmayv25: L213 prints the message in verbose mode only. This will let the user know that autofiller failed as a warning.

src/core/filesystem.cc
tanmayv25: Done

src/core/autofill.cc
CoderHam: You should print the TRT Autofill status only if there is an error. I don't think it is necessary in the case of success

tanmayv25: Please note that control will not reach here if there was a success.

CoderHam: What @GuanLuo and I are saying is that In 213 you are printing the Status already. When there is a failure it will be logged twice.

src/backends/caffe2/autofill.cc
deadeyegoodwin: That means the autofiller wouldn't work with cc_model_filenames. I think the fix is to just change chose checks to make sure directory is not empty (or for netdef that there are at least 2 files). I would not get any fancier than that.

PR #878, opened 2019-11-14
trtserver returns error code 1 if unable to gracefully unload all models

deadeyegoodwin:


src/servers/main.cc
```diff
@@ -1150,4 +1151,12 @@ main(int argc, char** argv)
   if (keep_alive == nullptr) {
     return 1;
   }
+
+  // Return error code if failed to stop gracefully.
+  if (stop_err != nullptr) {
+    TRTSERVER_ErrorDelete(stop_err);
```
GuanLuo: macro `LOG_IF_ERR` already invokes `TRTSERVER_ErrorDelete`, exiting directly should be fine.

deadeyegoodwin: Yes, shouldn't use stop_err after that point since it is deleted. Fixed.

PR #884, opened 2019-11-15
Add ModelReadystateInfo in server status to elaborate the current readiness state

GuanLuo:


src/core/model_repository_manager.cc
```diff
@@ -869,8 +893,8 @@ ModelRepositoryManager::Create(
       tf_allow_soft_placement, tf_memory_limit_mb, &backend_config_map);
 
   std::unique_ptr<BackendLifeCycle> life_cycle;
-  RETURN_IF_ERROR(
-      BackendLifeCycle::Create(server, backend_config_map, &life_cycle));
+  RETURN_IF_ERROR(BackendLifeCycle::Create(
+      server, status_manager.get(), backend_config_map, &life_cycle));
```
deadeyegoodwin: Why extract the stastus_manager pointer out of the shared_ptr? Since you are capturing it in BackendLifeCycle you should keep it as a shared_ptr. It gets confusing (and dangerous) to have the same pointer be both managed and unmanaged.

src/core/server_status.h
```diff
@@ -190,18 +189,20 @@ class ServerStatusManager {
   Status UpdateConfigForModel(
       const std::string& model_name, const ModelConfig& model_config);
 
+  Status SetModelVersionReadyState(
```
deadeyegoodwin: Comment (description)

src/core/server_status.proto
```diff
@@ -192,19 +192,40 @@ enum ModelReadyState {
...
   //@@
   //@@     Current readiness state for the model.
   //@@
   ModelReadyState ready_state = 1;
 
+  //@@  .. cpp:var:: ModelReadyStateInfo ready_state_info
+  //@@
+  //@@     Supplemental information regarding the current readiness state.
+  //@@
+  ModelReadyStateInfo ready_state_info = 5;
```
deadeyegoodwin: ModelReadyStateReason

GuanLuo:
Also updated L0_lifecycle to accommodate the change

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -444,6 +446,8 @@ def test_dynamic_model_load_unload(self):
 
                 version_status = ss.model_status[savedmodel_name].version_status[3]
                 self.assertEqual(version_status.ready_state, server_status.MODEL_UNAVAILABLE)
+                self.assertEqual(version_status.ready_state_info.message, "unloaded",
```
deadeyegoodwin: Isn't this going for fail, should be ready_state_reason?

GuanLuo: Yes it should have been, I missed the change in tests... Will relaunch CI.

PR #885, opened 2019-11-15
Custom backend and test for CORRID and END controls

deadeyegoodwin:


src/core/sequence_batch_scheduler.cc
```diff
@@ -673,8 +673,12 @@ SequenceBatchScheduler::SequenceBatch::SequenceBatch(
...
 
   if (!correlation_id_tensor_.empty()) {
-    if (correlation_id_datatype != TYPE_UINT64) {
-      LOG_ERROR << "unexpected control datatype, expected TYPE_UINT64 for "
+    if ((correlation_id_datatype != TYPE_UINT64) &&
+        (correlation_id_datatype != TYPE_INT64) &&
+        (correlation_id_datatype != TYPE_UINT32) &&
+        (correlation_id_datatype != TYPE_INT32)) {
+      LOG_ERROR << "unexpected control datatype, expected TYPE_UINT64, "
+                   "TYPE_INT64, TYPE_UINT32 or TYPE_INT32 for "
```
GuanLuo: The type has been checked in [`ValidateModelConfig()`](https://github.com/NVIDIA/tensorrt-inference-server/pull/885/files#diff-015674a2d5ebc8ce219556d349b36e57R537) in model_config_utils.cc

deadeyegoodwin: Yes, I'm just being extra careful here and it is on a non-performance-critical path.

PR #887, opened 2019-11-16
Add libtorch back to L0_sequence_batcher

CoderHam:
- [x] Add libtorch ensemble models in test runner script
- [x] Fix sequence utility for libtorch ensembles.


CoderHam:
Something seems to be failing, will investigate

PR #888, opened 2019-11-16
Fix CORRID control to be correct for each slot

deadeyegoodwin:


deadeyegoodwin:
The override maps are shared across requests, but originally I was mutating them to add the CORRID not realizing they were shared. That meant that the same CORRID was being used by every request (because every request was sharing the same overrides). I extended the overrides to allow multiple maps, so that requests can still share an override map, but also have individual override maps for CORRID.

GuanLuo:
Understood

PR #890, opened 2019-11-17
Improve CUDA feature check in build_wheel.sh

aleksa2808:
In the python client build_wheel.sh script, make the CUDA shared memory feature condition check if the needed files exist instead of looking for the CUDA_VERSION environment variable.  
The current check breaks builds that have the TRTIS_ENABLE_GPU flag disabled on systems that have CUDA installed.


deadeyegoodwin:
@CoderHam Can you test and review this PR.

GuanLuo:
@deadeyegoodwin should we also add build_variant test for client-side?

CoderHam:
> @CoderHam Can you test and review this PR.
On it

PR #893, opened 2019-11-18
Refactor repetitive code in sequence batcher tests

CoderHam:


qa/common/sequence_util.py
```diff
@@ -0,0 +1,420 @@
+from builtins import range
```
deadeyegoodwin: Needs a copyright

CoderHam: Added

PR #897, opened 2019-11-19
Fixing the computation of ByteSizes for TRT dynamic shapes

tanmayv25:


src/backends/tensorrt/plan_utils.cc
```diff
@@ -208,7 +209,19 @@ MaximumDims(
   }
 
   if (support_batching) {
-    max_dims->emplace_back(max_profile_dims.d[0]);
+    if (*max_allowed_batch_size > max_profile_dims.d[0]) {
```
GuanLuo: Isn't `max_allowed_batch_size` an output parameter? Why don't you just set it directly.

PR #904, opened 2019-11-21
Adding support for shared memory in perf-client

tanmayv25:


src/clients/c++/perf_client/request_rate_manager.cc
```diff
@@ -26,6 +26,24 @@
...
+  for (auto& thread : threads_) {
+    thread.join();
+    if (!threads_stat_[cnt]->status_.IsOk()) {
+      std::cerr << "Thread [" << cnt
+                << "] had error: " << (threads_stat_[cnt]->status_)
+                << std::endl;
+    }
+    cnt++;
+  }
+}
```
GuanLuo: The code in both derived class destructors are the same, why move it out of the base class destructor?

src/clients/c++/perf_client/perf_utils.h
```diff
@@ -112,3 +117,32 @@ std::string GetRandomString(const int string_length);
...
+    int shm_fd, size_t offset, size_t byte_size, void** shm_addr);
+
+// Destory the shared memory region with the given name.
+// \return error Returns if an error if unable to unlink shared memory region.
+nic::Error UnlinkSharedMemoryRegion(std::string shm_key);
+
+// Munmap the shared memory region from the base address with the given
+// byte_size.
+// \return error Returns if an error if unable to unmap shared memory region.
+nic::Error UnmapSharedMemory(void* shm_addr, size_t byte_size);
```
GuanLuo: I think @CoderHam has extracted these functions as common utilities for client library, you probably can just use that directly.

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -27,21 +27,97 @@
...
-
-  size_t cnt = 0;
-  for (auto& thread : threads_) {
-    thread.join();
-    if (!threads_stat_[cnt]->status_.IsOk()) {
-      std::cerr << "Thread [" << cnt
-                << "] had error: " << (threads_stat_[cnt]->status_)
+  nic::Error err;
+  if (shared_memory_type_ == SharedMemoryType::SYSTEM_SHARED_MEMORY) {
+    err = shared_memory_ctx_->UnregisterAllSharedMemory();
```
GuanLuo: Just to be safe, you should condition on whether `shared_memory_ctx_ == nullptr`

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -213,33 +306,297 @@ LoadManager::InitManagerInputs(
...
+      }
+
+      SerializeStringTensor(*string_data, &serialized_data);
+      batch1_bytesize = serialized_data.size();
+      data = (uint8_t*)&serialized_data[0];
+
+      if (used_new_allocation) {
+        delete string_data;
+      }
+    }
```
GuanLuo: The above is similar to what we do for preparing inputs for none shared memory case, can be extracted as a utility function?

src/clients/c++/perf_client/perf_utils.h
CoderHam: Yes, refer to `examples/shm_utils.h` (Only system shared memory)

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -27,21 +27,97 @@
...
+      }
+      err = UnlinkSharedMemoryRegion(region.first);
+      if (!err.IsOk()) {
+        std::cerr << "Unable to unlink shared memory with key: " << region.first
+                  << std::endl;
+      }
+    }
+  } else if (shared_memory_type_ == SharedMemoryType::CUDA_SHARED_MEMORY) {
+#if TRTIS_ENABLE_GPU
+    err = shared_memory_ctx_->UnregisterAllSharedMemory();
```
CoderHam: This should be for GPU and CPU both. Shift the ENABLE GPU flag to after this

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -213,33 +306,297 @@ LoadManager::InitManagerInputs(
...
+    size_t alloc_size = batch1_bytesize * batch_size_;
+    if (shared_memory_type_ == SharedMemoryType::SYSTEM_SHARED_MEMORY) {
+      std::string shm_key("/" + output->Name());
+      int shm_fd_op;
+      RETURN_IF_ERROR(
+          CreateSharedMemoryRegion(shm_key, alloc_size, &shm_fd_op));
+      RETURN_IF_ERROR(
+          MapSharedMemory(shm_fd_op, 0, alloc_size, (void**)&output_shm_ptr));
+
+      io_shm_size_[output->Name()] = alloc_size;
```
CoderHam: You can use a single map for size and address

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -27,21 +27,97 @@
...
+                  << "): Starting: " << static_cast<void*>(region.second)
+                  << ", size: " << io_shm_size_[region.first] << std::endl;
+      }
+      err = UnlinkSharedMemoryRegion(region.first);
+      if (!err.IsOk()) {
+        std::cerr << "Unable to unlink shared memory with key: " << region.first
+                  << std::endl;
+      }
+    }
+  } else if (shared_memory_type_ == SharedMemoryType::CUDA_SHARED_MEMORY) {
```
CoderHam: What if shared memory type is CUDA shared memory when TRTIS_ENABLE_GPU=OFF. Should return an error message for this case

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -213,33 +306,297 @@ LoadManager::InitManagerInputs(
...
+    if (input->Shape().empty()) {
+      auto it = input_shapes_.find(input->Name());
+      if (it != input_shapes_.end()) {
+        input->SetShape(it->second);
+      }
+    }
+    std::vector<char> serialized_data;
+    const uint8_t* data;
+    size_t batch1_bytesize;
+    if (input->DType() != ni::DataType::TYPE_STRING) {
```
CoderHam: What about when using variable dims inputs?

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -448,6 +450,25 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "Default is none.",
+                   18)
+            << std::endl;
+
+  std::cerr
+      << FormatMessage(
+             " --output-shm-size: Is used to compute the size of shared "
+             "memory to allocate for output tensors when output tensors "
+             "are string type or variable. This value should be large "
+             "enough to hold largest of the output tensors. The client "
```
CoderHam: larger than the size of the largest output tensor the model is expected to return.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -448,6 +450,25 @@ Usage(char** argv, const std::string& msg = std::string())
                    "order. Default is \"random\".",
                    18)
             << std::endl;
+  std::cerr << FormatMessage(
+                   " --shared-memory <\"system\"|\"cuda\"|\"none\">: Selects "
```
CoderHam: Specifies instead of selects

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -448,6 +450,25 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "the type of the shared memory to use for input data. "
```
CoderHam: input and output

src/clients/c++/perf_client/request_rate_manager.h
```diff
@@ -68,7 +68,14 @@ class RequestRateManager : public LoadManager {
   /// maintained on the server.
   /// \param sequence_length The base length of each sequence.
   /// \param zero_input Whether to fill the input tensors with zero.
-  /// \param factory The ContextFactory object used to create InferContext.
+  /// \param input_shapes The shape of the input tensors.
+  /// \param data_directory The path to the directory containing the data to
+  /// use for input tensors.
+  /// \param shared_memory_type The type of shared memory to use for inputs.
+  /// \param output_shm_size The size of the shared memory to allocate for the
```
CoderHam: size in bytes

src/clients/c++/perf_client/perf_utils.cc
```diff
@@ -166,3 +172,68 @@ ScheduleDistribution<Distribution::CONSTANT>(const double request_rate)
           std::chrono::duration<double>(1.0 / request_rate));
   return [period](std::mt19937& /*gen*/) { return period; };
 }
+
+nic::Error
+CreateSharedMemoryRegion(std::string shm_key, size_t byte_size, int* shm_fd)
```
CoderHam: Use these fucntions from shm utils instead of redefining here. Instead move the CUDA shared memory functions here if needed

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -448,6 +450,25 @@ Usage(char** argv, const std::string& msg = std::string())
...
+                   "Default is none.",
+                   18)
+            << std::endl;
+
+  std::cerr
+      << FormatMessage(
+             " --output-shm-size: Is used to compute the size of shared "
```
CoderHam: The size in bytes of the shared memory region to allocate per output tensor. Only needed when one or more of the outputs are of string type or variable shape. The value should be at least the maximum possible size (in bytes) of any of the output tensors the model can return. The client will use the following formula to calculate the total shared memory to allocate: output_shm_size * number_of_outputs * batch_size. Defaults to 100KB.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -770,14 +795,34 @@ main(int argc, char** argv)
...
         }
         break;
       }
       case 20:
         using_custom_intervals = true;
         request_intervals_file = optarg;
         break;
+      case 21: {
+        std::string arg = optarg;
+        if (arg.compare("system") == 0) {
```
CoderHam: In windows, shared memory is not supported. This means that build for the perf client should change accordingly and an error should be thrown if the client tries to run the perf client with shared memory.

src/clients/c++/perf_client/request_rate_manager.cc
tanmayv25: This is done because of the order in which destructors are called. The base class destructor will be called after the derived class. The worker threads depends on certain data members of derived class to operate and exit gracefully. So, the destruction of derived class should wait for the threads to finish before exiting. I don't want entire code to be replicated in all the derived classes hence, I moved the final clean-up to the base class.

src/clients/c++/perf_client/perf_utils.h
tanmayv25: Done

src/clients/c++/perf_client/perf_utils.cc
tanmayv25: Done.

src/clients/c++/perf_client/perf_client.cc
tanmayv25: We currently don't build our examples and perf_client on windows. Probably, will file a ticket to add the support.

src/clients/c++/perf_client/perf_client.cc
tanmayv25: Done

src/clients/c++/perf_client/perf_client.cc
tanmayv25: Done

src/clients/c++/perf_client/perf_client.cc
tanmayv25: Done

src/clients/c++/perf_client/perf_client.cc
tanmayv25: Done

src/clients/c++/perf_client/load_manager.cc
tanmayv25: Done

src/clients/c++/perf_client/load_manager.cc
tanmayv25: We have already used SetShape() with the user given shapes, hence will receive a correct ByteSize()

src/clients/c++/perf_client/load_manager.cc
tanmayv25: Done

src/clients/c++/perf_client/load_manager.cc
tanmayv25: The validation is performed in the perf_client.cc. When the user specifies shared-memory as cuda when TRTIS_ENABLE_GPU=0, then it fails.

src/clients/c++/perf_client/load_manager.cc
tanmayv25: Modified the structure.

src/clients/c++/perf_client/load_manager.cc
tanmayv25: Done

src/clients/c++/perf_client/request_rate_manager.cc
GuanLuo: Can you add inline comment about this?

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -301,98 +544,167 @@ LoadManager::PrepareInfer(
...
-          data = new std::vector<std::string>(
-              input_string_buf_.begin(),
-              input_string_buf_.begin() + batch1_num_strings);
-        }
-      } else {
-        return nic::Error(
-            ni::RequestStatusCode::INVALID_ARG,
-            "unable to find data for input '" + input->Name() + "'.");
-      }
+      std::shared_ptr<std::vector<std::string>> data;
```
GuanLuo: You can just make this a local variable instead of a smart pointer, I don't think you need to `new` the vector in `GetInputData()`.

src/clients/c++/perf_client/request_rate_manager.cc
deadeyegoodwin: Or better can you create a "StopWorkerThreads" or similar utility function in the base class that has this code and then call it from both derived destructors.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -243,6 +243,8 @@ Usage(char** argv, const std::string& msg = std::string())
   std::cerr << "II. INPUT DATA OPTIONS: " << std::endl;
   std::cerr << "\t-b <batch size>" << std::endl;
   std::cerr << "\t--input-data <\"zero\"|\"random\"|<path>>" << std::endl;
+  std::cerr << "\t--shared-memory <\"system\"|\"cuda\"|\"none\">" << std::endl;
+  std::cerr << "\t--output-shm-size <size in bytes>" << std::endl;
```
deadeyegoodwin: We've avoided using shm in external "controls". How about --output-shared-memory-size.

src/clients/c++/perf_client/perf_client.cc
```diff
@@ -448,6 +450,26 @@ Usage(char** argv, const std::string& msg = std::string())
...
+  std::cerr << FormatMessage(
+                   " --shared-memory <\"system\"|\"cuda\"|\"none\">: Specifies "
+                   "the type of the shared memory to use for input and output "
+                   "data. Default is none.",
+                   18)
+            << std::endl;
+
+  std::cerr
+      << FormatMessage(
+             " --output-shm-size: The size in bytes of the shared memory "
```
deadeyegoodwin: name fix

src/clients/c++/perf_client/perf_client.cc
CoderHam: Yes you can do that

src/clients/c++/perf_client/load_manager.cc
CoderHam: Ah, I see

src/clients/c++/perf_client/load_manager.cc
tanmayv25: This was needed because the SetFromString() function reads data from a string vector. So for the inputs of different size I had to prepare the tensor. It would have caused a considerable slow down while opening contexts, because the API serializes the string input data. Hence, to avoid the extraneous string copies, I am now serializing the string data in initialization phase itself. Also, I am maintaining different copies for different string inputs in the model to facilitate easy retrieval of their serialized byte sizes.

PR #908, opened 2019-11-21
HTTP server with V1 Dataplane support (KFServing/MLserving)

CoderHam:
- inference - POST http://host:port/v1/models/${MODEL_NAME}:predict
- model health - GET http://host:port/v1/models/${MODEL_NAME}


build/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ option(TRTIS_ENABLE_CUSTOM "Include Custom backend in server" OFF)
 # Endpoints
 option(TRTIS_ENABLE_HTTP "Include HTTP API in server" ON)
 option(TRTIS_ENABLE_GRPC "Include GRPC API in server" ON)
+option(TRTIS_ENABLE_KFSERVING "Include KFServing HTTP API in server" ON)
```
deadeyegoodwin: TRTIS_ENABLE_HTTP_V2  "Enable version 2 of the HTTP API in server"
Add a check before line 68 below that raises an error is this is set and TRTIS_ENABLE_HTTP is not enabled. So to get V2 you need to enable both flags.

Also, for now just disallow metrics when V2 is enabled. We could break out a metrics_server.cc but for now don't need to do that.

build/CMakeLists.txt
```diff
@@ -399,7 +400,7 @@ endif() # TRTIS_ENABLE_GCS
 if(${TRTIS_ENABLE_S3})
   set(TRTIS_DEPENDS ${TRTIS_DEPENDS} aws-sdk-cpp)
 endif() # TRTIS_ENABLE_S3
-if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
+if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS} OR ${TRTIS_ENABLE_KFSERVING})
```
deadeyegoodwin: Don't need this change

src/servers/CMakeLists.txt
```diff
@@ -115,6 +115,55 @@ if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
   )
 endif() # TRTIS_ENABLE_HTTP || TRTIS_ENABLE_METRICS
 
+#
+# kfserving http endpoint
```
deadeyegoodwin: Will need to disable the V1 endpoint above if V2 is enabled.

src/servers/CMakeLists.txt
```diff
@@ -115,6 +115,55 @@ if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
...
+set(KFSERVING_HTTP_ENDPOINT_OBJECTS "")
+set(KFSERVING_HTTP_ENDPOINT_LIBRARIES "")
+
+if(${TRTIS_ENABLE_KFSERVING})
+  find_package(libevhtp CONFIG REQUIRED)
+  message(STATUS "Using libevhtp ${libevhtp_VERSION}")
+
+  set(
+    KFSERVING_HTTP_ENDPOINT_SRCS
+    kfserving_http_server.cc
```
deadeyegoodwin: http_v2_server.cc

src/servers/CMakeLists.txt
```diff
@@ -115,6 +115,55 @@ if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
...
+# kfserving http endpoint
+#
+set(KFSERVING_HTTP_ENDPOINT_OBJECTS "")
+set(KFSERVING_HTTP_ENDPOINT_LIBRARIES "")
+
+if(${TRTIS_ENABLE_KFSERVING})
+  find_package(libevhtp CONFIG REQUIRED)
+  message(STATUS "Using libevhtp ${libevhtp_VERSION}")
+
+  set(
```
deadeyegoodwin: I think you can just set HTTP_ENDPOINT_SRCS and HTTP_ENDPOINT_HDRS depending on TRTIS_ENABLE_HTTP and TRTIS_ENABLE_HTTP_V2 and everything else is the same

src/servers/kfserving_http_server.h
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Just use this same header for both V1 and V2 and since using the TRTIS_ENABLE_HTTP_V2 flag to build one source file or the other will get the correct implementation.

src/servers/main.cc
```diff
@@ -56,6 +56,10 @@ static_assert(
 #include "src/servers/grpc_server.h"
 #endif  // TRTIS_ENABLE_GRPC
 
+#ifdef TRTIS_ENABLE_KFSERVING
```
deadeyegoodwin: I don't think we need any changes in this file.

src/servers/kfserving_http_server.cc
```diff
@@ -0,0 +1,949 @@
...
+    evhtp_send_reply(req, EVHTP_RES_METHNALLOWED);
+    return;
+  }
+
+  if (model_name.empty()) {
+    evhtp_send_reply(req, EVHTP_RES_BADREQ);
+    return;
+  }
+
+  bool ready = true;
```
deadeyegoodwin: Shouldn't you start with 'ready' false and then set it to true below only if everything works correctly and the status is MODEL_READY?

src/servers/kfserving_http_server.cc
```diff
@@ -0,0 +1,949 @@
...
+          kStatusHTTPHeader, request_status.ShortDebugString().c_str(), 1, 1));
+
+  evhtp_send_reply(
+      req, (ready && (err == nullptr)) ? EVHTP_RES_OK : EVHTP_RES_BADREQ);
+
+  TRTSERVER_ErrorDelete(err);
+}
+
+void
+KFServingHTTPAPIServer::HandleStatus(
```
deadeyegoodwin: I would not implement the "metadata" endpoint. Our status is not in a standard format and we don't want to create that standard format at this point.

src/servers/kfserving_http_server.cc
```diff
@@ -0,0 +1,949 @@
...
+        output_shm_map)
+{
+  // Extract input data from HTTP body and register in
+  // 'request_provider'.
+  size_t buffer_length = evbuffer_get_length(input_buffer);
+  char* request_buffer = (char*)malloc(sizeof(char) * buffer_length);
+  memset(request_buffer, 0, buffer_length);
+  evbuffer_copyout(input_buffer, request_buffer, buffer_length);
+
+  // TODO MessageToJsonString to convert message to string (bytes encoded in
```
deadeyegoodwin: Why is TODO still here? Is there something else needed?

build/CMakeLists.txt
CoderHam: Will do

src/servers/CMakeLists.txt
CoderHam: Yes, I will do that

src/servers/kfserving_http_server.cc
CoderHam: Removed

src/servers/kfserving_http_server.cc
CoderHam: Removing that, added for the client side changes

src/servers/kfserving_http_server.cc
CoderHam: Yes, fixed. READY is returned if all versions of the model have the MODEL_READY status

src/servers/kfserving_http_server.h
CoderHam: Done

src/servers/main.cc
CoderHam: Fixed

src/servers/CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ endif() # TRTIS_ENABLE_GRPC
 set(HTTP_ENDPOINT_OBJECTS "")
 set(HTTP_ENDPOINT_LIBRARIES "")
 
-if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
+if((${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS}) AND (NOT ${TRTIS_ENABLE_HTTP_V2}))
   find_package(libevhtp CONFIG REQUIRED)
   message(STATUS "Using libevhtp ${libevhtp_VERSION}")
```
deadeyegoodwin: Within here can just conditionalize HTTP_ENDPOINT_SRCS  on TRTIS_ENABLE_HTTP_V2. Don't need to duplicate everything.

src/servers/CMakeLists.txt
```diff
@@ -171,6 +217,7 @@ add_executable(
   ${TRTSERVER_HDRS}
   ${HTTP_ENDPOINT_OBJECTS}
   ${GRPC_ENDPOINT_OBJECTS}
+  ${KFSERVING_HTTP_ENDPOINT_OBJECTS}
```
deadeyegoodwin: After above change can remove here and below since HTTP_ENDPOINT_OBJECTS will have the correct objects for both V1 and V2 cases.

src/servers/http_v2_server.cc
```diff
@@ -0,0 +1,877 @@
...
+    response_header.set_id(request_header.id());
+    evhtp_headers_add_header(
+        req->headers_out,
+        evhtp_header_new(
+            kInferResponseHTTPHeader,
+            response_header.ShortDebugString().c_str(), 1, 1));
+    LOG_VERBOSE(1) << "Infer failed: " << request_status.msg();
+
+    evhtp_headers_add_header(
+        req->headers_out, evhtp_header_new(
```
deadeyegoodwin: Did you run the formatter... this looks strange.

src/servers/CMakeLists.txt
CoderHam: Fixed

src/servers/http_v2_server.cc
CoderHam: I hadn't resolved now

build/CMakeLists.txt
```diff
@@ -68,6 +69,15 @@ if(TRTIS_ENABLE_TENSORRT AND NOT TRTIS_ENABLE_GPU)
   message(FATAL_ERROR "TRTIS_ENABLE_TENSORRT=ON requires TRTIS_ENABLE_GPU=ON")
 endif()
 
+if(TRTIS_ENABLE_HTTP_V2 AND NOT TRTIS_ENABLE_HTTP)
+  message(FATAL_ERROR "TRTIS_ENABLE_HTTP_V2=ON requires TRTIS_ENABLE_HTTP=ON")
+endif()
+
+if (TRTIS_ENABLE_HTTP_V2 AND TRTIS_ENABLE_METRICS)
+  message("TRTIS_ENABLE_HTTP_V2=ON turns TRTIS_ENABLE_METRICS OFF")
```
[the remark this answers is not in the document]
CoderHam: Done

src/servers/CMakeLists.txt
CoderHam: Done

PR #909, opened 2019-11-25
Fixing the TRT backend logs in default case

tanmayv25:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -387,7 +388,7 @@ PlanBackend::CreateExecutionContext(
   if (context->is_dynamic_) {
     LOG_INFO << "Created instance " << instance_name << " on GPU " << gpu_device
              << " (" << cc << ") with stream priority " << cuda_stream_priority
-             << " and optimization profile " << profile_name;
+             << " and optimization profile " << std::to_string(profile_index);
```
GuanLuo: Maybe printing both name and index? The utility function `GetProfileIndex()` suggests that `profile_name` can be more than the string form of the index in the future?

tanmayv25: Done

GuanLuo: Also here?

PR #919, opened 2019-11-27
Support specifying multiple TensorRT optimization profiles for inference

GuanLuo:


GuanLuo:
Will rebase and rerun CI once #930 is merged

qa/common/gen_qa_models.py
```diff
@@ -728,8 +728,15 @@ def create_plan_dynamic_modelfile(models_dir, max_batch, model_version,
             profile[i].set_shape("INPUT1", min_shape, opt_shape, max_shape)
         else:
             bs = [ max_batch - i if max_batch > i else 1]
-            profile[i].set_shape("INPUT0", [1] + min_shape, bs + opt_shape, bs + max_shape)
-            profile[i].set_shape("INPUT1", [1] + min_shape, bs + opt_shape, bs + max_shape)
```
GuanLuo: @tanmayv25 I changed the max dims here to be `[max_batch, max_shape]` so that all profiles are aligned with the model config, is that okay? I think `bs + max_shape` is typo.

src/core/model_config.proto
```diff
@@ -155,15 +155,15 @@ message ModelInstanceGroup
   //@@
   repeated int32 gpus = 3;
 
-  //@@  .. cpp:var:: string profile
+  //@@  .. cpp:var:: string profile (repeated)
   //@@
   //@@     For TensorRT models, using inputs with dynamic shape, this
```
deadeyegoodwin: Description needs update. I think it needs to say the set of optimization profiles available to each instance and server will choose from those the optimal profile based on the shape of the input tensors.

src/core/model_config_utils.cc
```diff
@@ -679,15 +679,17 @@ ValidateModelConfig(
...
+        for (const auto& profile : group.profile()) {
+          char* end_ptr;
+          long profile_index = strtol(profile.c_str(), &end_ptr, 10);
+          if (*end_ptr != '\0' || profile_index < 0) {
+            return Status(
+                RequestStatusCode::INVALID_ARG,
+                "instance group " + group.name() + " of model " +
+                    config.name() + " and platform " + config.platform() +
+                    " specifies invalid profile " + profile +
+                    ". The field should contain a non-negative integer.");
```
deadeyegoodwin: "... should contain the string representation of a non-..."

src/core/model_config_utils.cc
```diff
@@ -679,15 +679,17 @@ ValidateModelConfig(
...
-        if (*end_ptr != '\0' || profile_index < 0) {
-          return Status(
-              RequestStatusCode::INVALID_ARG,
-              "instance group " + group.name() + " of model " + config.name() +
-                  " and platform " + config.platform() +
-                  " specifies invalid profile " + group.profile() +
-                  ". The field should contain a non-negative integer.");
+        for (const auto& profile : group.profile()) {
+          char* end_ptr;
+          long profile_index = strtol(profile.c_str(), &end_ptr, 10);
```
deadeyegoodwin: Can you move GetProfileIndex into model_config_utils.cc and have it return proper status on failure and then use it everywhere.

qa/common/gen_qa_models.py
tanmayv25: It was done purposely to catch our shape validation logic.

GuanLuo: Yeah, noticed and reverted the change on max_dims

PR #927, opened 2019-12-02
Add HTTP V2 test

CoderHam:


qa/L0_http_v2/test.sh
```diff
@@ -0,0 +1,128 @@
...
+    echo -e "\n***\n*** HTTP V2 Client Build Failed\n***"
+    RET=1
+fi
+
+DATADIR=`pwd`/models
+SERVER=/opt/tensorrtserver/bin/trtserver
+SERVER_ARGS="--model-repository=$DATADIR"
+source ../common/util.sh
+
+run_server_nowait
```
deadeyegoodwin: Just use run_server and then you don't need the sleep

CoderHam: **run_server** calls **wait_for_server_ready** which repeatedly curls the old HTTP url ->  `localhost:8000/api/health/ready` till it gets a success code 200 and hence would get stuck here when using HTTP V2

deadeyegoodwin: Can you put that comment in the script so we know why you are using it

CoderHam: Sure

PR #937, opened 2019-12-06
Add build option to disable infer stats collection

GuanLuo:


build/CMakeLists.txt
```diff
@@ -60,12 +60,17 @@ set(TRTIS_EXTRA_LIB_PATHS "" CACHE PATH "Extra library paths for TRTIS build")
 set(TRTIS_CLIENT_CMAKE_DIR "" CACHE PATH "Path to TRTIS client library cmake configuration")
 
 option(TRTIS_ENABLE_LOGGING "Include logging support in server" ON)
+option(TRTIS_ENABLE_INFER_STATS "Include inference statistics support in server" ON)
```
deadeyegoodwin: Thinking about this some more I think it will be better to have this be TRTIS_ENABLE_STATS and have it disable all stats collection, not just the inference ones.

GuanLuo: ok

GuanLuo: I think then the `TRTIS_ENABLE_METRICS` will also need to be disabled if `TRTIS_ENABLE_STATS=OFF`?

deadeyegoodwin: There are some metrics that don't depend on stats. Can you just show those metrics when stats are disabled? If that is too difficult it is ok to just require metrics to be disabled when stats are disabled.

GuanLuo: It shouldn't be hard, just need to condition the inference related metrics on `TRTIS_ENABLE_STATS`

PR #938, opened 2019-12-06
WIP: Fix tftrt optimization test

CoderHam:


qa/L0_tftrt_optimization/test.sh
```diff
@@ -178,6 +178,47 @@ for MODEL in \
         RET=1
     fi
 
+    set -e
+
+    kill $SERVER_PID
+    wait $SERVER_PID
+done
+
+for MODEL in \
```
deadeyegoodwin: So it was failing because the TensorRT optimization was not actually enabled?

CoderHam: Just confirmed with Guan, this was not the case. We see segfaults even with the current case.

PR #940, opened 2019-12-09
Add JSON Format Reporting for Server Status

CoderHam:
Option to use `?format=json` in request url or Accept: `application/json` to get server status as JSON

### DIFF
--- a/qa/L0_server_status/server_status_test.py
+++ b/qa/L0_server_status/server_status_test.py
@@ -32,6 +32,8 @@ from future.utils import iteritems
 import os
 import infer_util as iu
 import unittest
+from google.protobuf import json_format
+import json
 from tensorrtserver.api import *
 import tensorrtserver.api.server_status_pb2 as server_status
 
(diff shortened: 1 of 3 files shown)


CoderHam:
The L0_server_status checks the status using the client library and the client library runs only with format=binary. Do we extend the client library to return the server status as json/text if requested?

deadeyegoodwin:
You can just use python request directly and then parse the response into json and then check some fields to make sure it is ok.

CoderHam:
Added a test to convert to json, validate a field and convert back to protobuf to ensure it is identical to original.

ryanleary:
Could you also request the JSON payload via an `Accept: application/json` header too?

CoderHam:
> Could you also request the JSON payload via an `Accept: application/json` header too?

Added support for this.

qa/L0_server_status/server_status_test.py
```diff
@@ -217,6 +219,17 @@ def test_model_specific_infer(self):
             except InferenceServerException as ex:
                 self.assertTrue(False, "unexpected error {}".format(ex))
 
+    def test_json_server_status(self):
```
GuanLuo: The test case content doesn't make sense. You are still receiving protobuf message from server instead of JSON. Here you just testing the correctness of protobuf's JSON convertor...

CoderHam: Because of the way our client library is written, it defaults to using binary as the format. The same converter is what we are using inside the server to return the server status as a JSON.

GuanLuo: So what is the point of having this test case? It's not hitting the code change in this PR, right?

CoderHam: Fixed test coverage in following commit

PR #948, opened 2019-12-12
Add support to pass numpy array of bytes for string inputs

CoderHam:
- Using the python client API

### DIFF
--- a/qa/L0_string_io/string_client_test.py
+++ b/qa/L0_string_io/string_client_test.py
@@ -35,9 +35,8 @@ import unittest
 class ClientStringTest(unittest.TestCase):
     def test_tf_unicode_bytes(self):
         # We use a simple model that takes an input tensor of 6000 strings
-        # and returns an output tensors of 6000 strings. The input
-        # strings must represent integers. The output tensor is the
-        # same as the input tensor.
+        # and returns an output tensors of 6000 strings. The output tensor
+        # is the same as the input tensor.
         model_name = "graphdef_nobatch_zero_1_object"
         model_version = -1
         batch_size = 1
--- a/qa/L0_string_io/string_client_test.py
+++ b/qa/L0_string_io/string_client_test.py
@@ -46,7 +46,8 @@ class ClientStringTest(unittest.TestCase):
 
         # Create the data for the input tensor. Initialize the tensor to the
         # byte representation of the first 6000 unicode characters.
-        in0 = np.array([chr(i) for i in range(6000)])
+        # TODO: For now pass non-utf chars -> change if needed later
+        in0 = np.array([b'\xf4\x9a\xb6\xb1' for i in range(6000)])
         in0n = np.array([bytes(x, encoding='utf8') for x in in0])
 
         # Send inference request to the inference server. Get results for
--- a/qa/L0_string_io/string_client_test.py
+++ b/qa/L0_string_io/string_client_test.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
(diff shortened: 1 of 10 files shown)


qa/common/infer_util.py
```diff
@@ -65,7 +65,10 @@ def _prepend_string_size(input_values):
             # don't convert it to str as Python will encode the
             # bytes which may distort the meaning
             if obj.dtype.type == np.bytes_:
-                s = bytes(obj)
+                if type(obj.item()) == bytes:
+                    s = obj.item()
+                else:
+                    s = bytes(obj)
```
GuanLuo: Can we abstract all the processing code into a common functionality?

qa/L0_string_io/string_client_test.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+from builtins import range
+from tensorrtserver.api import *
+import unittest
+
+class ClientStringTest(unittest.TestCase):
+    def test_tf_unicode_bytes(self):
+        # We use a simple model that takes an input tensor of 6000 strings
+        # and returns an output tensors of 6000 strings. The input
+        # strings must represent integers. The output tensor is the
```
deadeyegoodwin: The strings don't need to represent integers do they?

qa/L0_string_io/string_client_test.py
```diff
@@ -0,0 +1,64 @@
...
+        model_version = -1
+        batch_size = 1
+
+        # Create the inference context for the model.
+        ctx = InferContext("localhost:8000", ProtocolType.HTTP, model_name, model_version, verbose=True)
+
+        # Create the data for the input tensor. Initialize the tensor to the
+        # byte representation of the first 6000 unicode characters.
+        in0 = np.array([chr(i) for i in range(6000)])
+        in0n = np.array([bytes(x, encoding='utf8') for x in in0])
```
deadeyegoodwin: This fails before you change due to some unicode not be representable as rtf-8? How are you able to encode with utf-8 here? Can you just make a "bytes" of the values 0 - 6000 (or at least one set of bytes that can convert to utf-8).

qa/common/infer_util.py
```diff
@@ -65,7 +65,10 @@ def _prepend_string_size(input_values):
             # don't convert it to str as Python will encode the
             # bytes which may distort the meaning
             if obj.dtype.type == np.bytes_:
```
deadeyegoodwin: Isn't the numpy dtype "object"?

qa/common/infer_util.py
CoderHam: Yes, I will do that

qa/L0_string_io/string_client_test.py
CoderHam: Nope, removed.

qa/L0_string_io/string_client_test.py
CoderHam: The resultant numpy array is not of type np.object. It fails because the eariler code gets the bytes representation of the byte string in numpy (which is not the actual byte string). My change checks for the case when the numpy array contains elements that are byte strings and appends them directly to the flattened bytes object.
It is not necessary to use np.object as the dtype and this will allow for easy backward compatibiltiy. The user now can send the string as either a numpy array of bytes string or np array of dtype string (unicode string) type or np.array of string with np.object dtype

qa/common/infer_util.py
CoderHam: Nope, the dtype for a numpy array of bytes need not be np.object. The bytes() representation is a sequence of bytes

qa/common/infer_util.py
CoderHam: For now I have kept them in isolated functions. Do we want to have a separate string utility file that does all this?

GuanLuo: After another look, this function and the `serialize_string_tensor()` function that calls it don't really belong to both the (CUDA) shared memory Python library. You should just remove them, and make the serialize function be an utility function provided by the `tensorrtserver.api` package, as all "client" needs this functionality will have to import `tensorrtserver.api` anyway

src/clients/python/__init__.py
```diff
@@ -266,6 +266,55 @@ def _raise_error(msg):
     raise ex
 
 
+def serialize_string_tensor(input_tensor):
+    """Serializes a string tensor into a flat numpy array of length prepend strings.
```
GuanLuo: Add description about the case of passing bytes directly?

qa/common/shm_util.py
```diff
@@ -55,7 +55,10 @@ def _prepend_string_size(input_values):
             # don't convert it to str as Python will encode the
             # bytes which may distort the meaning
             if obj.dtype.type == np.bytes_:
-                s = bytes(obj)
+                if type(obj.item()) == bytes:
+                    s = obj.item()
+                else:
+                    s = bytes(obj)
```
GuanLuo: you forgot to clean up here and above in infer_util.py ...

CoderHam: Fixed

PR #955, opened 2019-12-13
Improve sequence control messages

CoderHam:


src/backends/tensorflow/tf_utils.cc
```diff
@@ -37,22 +37,34 @@ CompareDimsExact(
   // then the tensorflow shape first dimension must be -1.
   if (supports_batching) {
     if ((model_shape->rank_ == 0) || (model_shape->dims_[0] != -1)) {
-      return false;
+      return Status(
+          RequestStatusCode::INVALID_ARG,
+          "for the model to support batching, the shape should have at least 1 "
+          "dimension and the first dimension should not be -1");
```
deadeyegoodwin: "should not be" -> "must be" ?

src/backends/tensorflow/tf_utils.cc
```diff
@@ -37,22 +37,34 @@ CompareDimsExact(
...
+            std::to_string(dims.size() + (supports_batching ? 1 : 0)) +
+            " dimensions");
   }
 
   for (int i = 0; i < dims.size(); ++i) {
     if (model_shape->dims_[i + (supports_batching ? 1 : 0)] != dims[i]) {
-      return false;
+      return Status(
+          RequestStatusCode::INVALID_ARG,
+          "the model expects dims " + ShapeToString(model_shape) +
```
deadeyegoodwin: "expects shape ... "
"... but the model ... specified shape ... "

src/backends/tensorflow/tf_utils.h
```diff
@@ -40,7 +40,7 @@ namespace nvidia { namespace inferenceserver {
 /// \param supports_batching If True then the configuration expects
 /// the model to support batching and so the shape must have the
 /// appropriate batch dimension.
-bool CompareDimsExact(
+Status CompareDimsExact(
```
deadeyegoodwin: Fix the \return description above

CoderHam:
> I think we can similarly improve these messages for other backends as well?

This is specifically for sequence control messages but yes I will look into the other backends as well

tanmayv25:
Look here for plan_backend: https://github.com/NVIDIA/tensorrt-inference-server/blob/18724376b845bfcc62344cf4a99e17696bb9e031/src/backends/tensorrt/plan_backend.cc#L520
It will be good to have a common format.

PR #959, opened 2019-12-16
Provide TRTServer API to set request header procedurally to avoid (de)serializing protobuf

GuanLuo:


src/core/trtserver.h
```diff
@@ -465,32 +466,127 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
-/// \param model_version The version of the model that the inference
-/// request is for, or -1 to select the latest (highest numbered)
-/// version.
-/// \param request_header_base Pointer to the serialized request
-/// header protobuf.
-/// \param request_header_byte_size The size of the serialized request
-/// header in bytes.
+/// \param request_header The request header object for this inference request.
 /// \return a TRTSERVER_Error indicating success or failure.
 TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestProviderNew(
```
deadeyegoodwin: Can't remove this API yet as that would break backwards compatibility too severely. So just keep this one as it is but mark it as deprecated and point to a new TRTSERVER_InferenceRequestProviderNewV2(TRTSERVER_InferenceRequestProvider** request_provider, TRTSERVER_Server* server);

src/core/trtserver.h
```diff
@@ -465,32 +466,127 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
 TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerDelete(
     TRTSERVER_TraceManager* trace_manager);
 
+/// TRTSERVER_InferenceRequestHeader
```
deadeyegoodwin: I don't think we need a whole new "class" here after all. Just add methods to Provider. In those methods return an error if it is not a V2 created provider.

src/core/trtserver.h
```diff
@@ -465,32 +466,127 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+/// Parse the request header object from a serialized request header protobuf.
+/// The request header protobuf must be serialized and provided as a base
+/// address and a size, in bytes.
+/// \param request_header The request header object.
+/// \param request_header_base Pointer to the serialized request
+/// header protobuf.
+/// \param request_header_byte_size The size of the serialized request
+/// header in bytes.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error*
```
deadeyegoodwin: Delete... don't add anything related to protobufs.

src/core/trtserver.h
GuanLuo: The idea is that, for benchmarking, the same request header object can be reused for all requests that are consistent. But provider can not be reused because it maintains internal states.

deadeyegoodwin: Good point, but I don't think "header" should be in the name since that is just an artifact from HTTP/GRPC. How about InferenceRequestOptions to match what we have for server.

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+/// will contain the same ID. The request sender can use the ID to correlate
+/// the response to corresponding request if needed.
+/// \param request_options The request options object.
+/// \param id The ID.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsSetId(
+    TRTSERVER_InferenceRequestOptions* request_options, uint64_t id);
+
+/// Set the flag associated with the request in a request options. 'flags'
+/// should holds a bitwise-or of all flag values. \param request_options The
```
deadeyegoodwin: Formatting

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+/// \param model_version The version of the model that the inference
+/// request is for, or -1 to select the latest (highest numbered) version.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsNew(
+    TRTSERVER_InferenceRequestOptions** request_options, const char* model_name,
+    int64_t model_version);
+
+/// Set the ID for the request in a request options. The response of the request
+/// will contain the same ID. The request sender can use the ID to correlate
+/// the response to corresponding request if needed.
```
deadeyegoodwin: Add a sentence about what happens if this API is not called. I think the ID will be 0 in that case.

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsSetId(
+    TRTSERVER_InferenceRequestOptions* request_options, uint64_t id);
+
+/// Set the flag associated with the request in a request options. 'flags'
+/// should holds a bitwise-or of all flag values. \param request_options The
+/// request options object. \param flags The flags. \return a TRTSERVER_Error
+/// indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsSetFlags(
+    TRTSERVER_InferenceRequestOptions* request_options, uint32_t flags);
```
deadeyegoodwin: Need to define an enumeration for the flags (see api.proto). Similar naming, etc. as we have with TRTSERVER_Model_Control_Mode except that must use explicit power-of-2 enum values.

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+
+/// Set the correlation ID for the request in a request options.
```
deadeyegoodwin: Use the description from api.proto.

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+    uint64_t correlation_id);
+
+/// Set the batch size for the request in a request options.
+/// \param request_options The request options object.
+/// \param batch_size The batch size.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsSetBatchSize(
+    TRTSERVER_InferenceRequestOptions* request_options, uint32_t batch_size);
+
+
```
deadeyegoodwin: Remove extra line

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsAddInput(
+    TRTSERVER_InferenceRequestOptions* request_options, const char* input_name,
+    const int64_t* dims, uint64_t dim_count, uint64_t batch_byte_size);
+
+/// Request the raw data of the specified output to be returned for requests
+/// associated with the request options.
+/// \param request_options The request options object.
+/// \param output_name The name of the output.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsAddOutputRaw(
```
deadeyegoodwin: To align better with the upcoming V2 API make this be "TRTSERVER_InferenceRequestOptionsAddOutput".

src/core/trtserver.h
```diff
@@ -465,13 +466,103 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+    const char* output_name);
+
+/// Request the classification values of the specified output to be returned
+/// for requests associated with the request options.
+/// \param request_options The request options object.
+/// \param output_name The name of the output.
+/// \param count Indicates how many classification values should be returned
+/// for the output. The 'count' highest priority values are returned.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestOptionsAddOutputCls(
```
deadeyegoodwin: TRTSERVER_InferenceRequestOptionsAddClassificationOutput

src/core/trtserver.h
```diff
@@ -492,6 +583,20 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_InferenceRequestProviderNew(
     TRTSERVER_Server* server, const char* model_name, int64_t model_version,
     const char* request_header_base, size_t request_header_byte_size);
 
+/// Create a new inference request provider object. The caller retains ownership
+/// of 'request_options' but may release it by calling
+/// TRTSERVER_InferenceRequestOptionsDelete once this function returns.
+/// However, the values in 'request_options' must not be modified until the
```
deadeyegoodwin: What does it mean that the options object can be delete but not its contents? For efficiency I think it is likely that we want to say that the provider does not take ownership but that the caller must extend the options object's lifetime to be longer than the lifetime of any provider created with the options. That way in the implementation we don't have to copy anything from options. But if we are copying everything from options anyway for other reasons then we can just say that the options object can be deleted immediately on return.

src/servers/memory_alloc.cc
```diff
@@ -598,16 +598,21 @@ main(int argc, char** argv)
...
-  request_header.SerializeToString(&request_header_serialized);
-
   // Create the inference request provider which provides all the
   // input information needed for an inference.
+  TRTSERVER_InferenceRequestOptions* request_options = nullptr;
+  FAIL_IF_ERR(
+      TRTSERVER_InferenceRequestOptionsNew(
+          &request_options, model_name.c_str(), model_version),
+      "creating inference request options");
+  FAIL_IF_ERR(
```
deadeyegoodwin: Don't use the common.h utility function or protobuf. Use the new API directly.

src/servers/simple.cc
```diff
@@ -436,30 +436,35 @@ main(int argc, char** argv)
...
-  request_header.SerializeToString(&request_header_serialized);
-
   // Create the inference request provider which provides all the
   // input information needed for an inference.
+  TRTSERVER_InferenceRequestOptions* request_options = nullptr;
+  FAIL_IF_ERR(
+      TRTSERVER_InferenceRequestOptionsNew(
+          &request_options, model_name.c_str(), model_version),
+      "creating inference request options");
+  FAIL_IF_ERR(
```
deadeyegoodwin: Use new API directly...

src/core/trtserver.h
```diff
@@ -465,13 +466,116 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER_TraceManagerNew(
...
+///
+/// Object representing the request meta-data needed for an inference.
+///
+
+/// Inference request option flags. The enum values must be power-of-2 values.
+typedef enum trtserver_requestoptionsflag_enum {
+  TRTSERVER_REQUEST_FLAG_NONE = 0,
+  TRTSERVER_REQUEST_FLAG_SEQUENCE_START = 1,
+  TRTSERVER_REQUEST__FLAG_SEQUENCE_END = 2
+} TRTSERVER_REQUEST_OPTIONS_FLAG;
```
deadeyegoodwin: Follow naming convention in this file:  "TRTSERVER_Request_Options_Flag"

PR #962, opened 2019-12-17
Upgrade prometheus to v0.7

CoderHam:


CoderHam:
> Was there any change to how the endpoint prints the comments along with the metric values?

Nope, it still prints it as a line with # as the first non-whitespace character. I looked through the docs but there is no way to disable to modify the way the comments are printed with the metrics.

PR #966, opened 2019-12-17
Java/Kotlin client for TRTIS

seovchinnikov:
Implemented in Kotlin (Java interop).
Supports generating constant inputs based on shapes from metadata.
Does not support float16 for now.
Uint8 is represented as byte (because byte-form is preserved, representation is different - to represent it right need to manually do  & 0xFF and convert it to int).


deadeyegoodwin:
Thanks for the PR. We are planning on setting up a contrib area in the repo for client contributions like this one. We likely won't get to that until next year but once we have it we will take a closer look at this PR.

seovchinnikov:
Have you got any dates for introduction of a contrib area?
Thanks!

deadeyegoodwin:
We are considering a couple of options so we don't have an exact date yet but it should be within the next couple of months. I will update here when we know more.

cprice404-aws:
This was incredibly helpful, thanks for sharing it.

deadeyegoodwin:
We've (finally) added the contribution area to Triton. https://github.com/triton-inference-server/contrib
If you are still interested in making this contribution, please update it for the latest V2 version of Triton and submit it in the contrib repo.

PR #972, opened 2019-12-18
Add option to preserve ordering for dynamic batcher

GuanLuo:


src/core/dynamic_batch_scheduler.cc
```diff
@@ -102,6 +103,10 @@ DynamicBatchScheduler::Create(
     return Status(
         RequestStatusCode::INTERNAL,
         "Initialization failed for all dynamic-batch scheduler threads");
+  } else {
```
deadeyegoodwin: Don't need else here since if clause is a return.

src/core/dynamic_batch_scheduler.h
```diff
@@ -116,6 +117,15 @@ class DynamicBatchScheduler : public Scheduler {
 
   const bool enforce_equal_shape_batch_;
   std::unordered_map<std::string, DimsList> pending_batch_shapes_;
+
+
```
deadeyegoodwin: Just one blank line

src/core/model_config.proto
```diff
@@ -579,6 +579,14 @@ message ModelDynamicBatching
   //@@     batching. Default is 0.
   //@@
   uint64 max_queue_delay_microseconds = 2;
+
+  //@@  .. cpp:var:: bool preserve_ordering
+  //@@
+  //@@     If true, the responses will be returned in the same order as the
```
deadeyegoodwin: Should the dynamic batcher preserve the ordering of responses to match the order of requests received by the scheduler. Default is false. If true, ...

src/core/model_config.proto
```diff
@@ -579,6 +579,14 @@ message ModelDynamicBatching
...
+  //@@     order of requests sent to the scheduler. Otherwise the responses will
```
deadeyegoodwin: "If false the responses [may] be returned in an arbitrary order. This option is specifically needed when a sequence of related inference requests (i.e. inference requests with the same correlation ID) are sent to the dynamic batcher to ensure that the sequence responses are in the correct order."

src/core/sequence_batch_scheduler.cc
```diff
@@ -1196,7 +1196,8 @@ OldestSequenceBatch::OldestSequenceBatch(
   Status status = DynamicBatchScheduler::Create(
       batcher_idx_, 1 /* runner_cnt */, GetCpuNiceLevel(config), OnInit,
       OnWarmup, OnSchedule, true /* dynamic_batching_enabled */,
-      false /* enforce_equal_shape_batch */, preferred_batch_sizes,
+      false /* enforce_equal_shape_batch */, false /* preserve_ordering */,
```
deadeyegoodwin: The dynamic sequence scheduler always needs to preserve the order since it is sending in sequence requests. So change to "true".

PR #974, opened 2019-12-19
WIP: Make sure the dynamic batcher doesn't combine requests with incompati…

deadeyegoodwin:
…ble shape tensors

### DIFF
--- a/src/backends/tensorrt/plan_backend.cc
+++ b/src/backends/tensorrt/plan_backend.cc
@@ -154,11 +154,34 @@ PlanBackend::CreateExecutionContexts(
           uint32_t runner_idx, std::vector<Scheduler::Payload>* payloads,
           std::function<void(Status)> func) {
         Run(runner_idx, payloads, func);
+      },
+      [this](
+          uint32_t runner_idx, const InferRequestHeader::Input& input,
+          const Scheduler::Payload& payload,
+          std::vector<int64_t>* shape) -> Status {
+        return PeekShapeTensor(runner_idx, input, payload, shape);
       }));
 
   LOG_VERBOSE(1) << "plan backend for " << Name() << std::endl << *this;
 
   return Status::Success;
+}
+
+Status
+PlanBackend::PeekShapeTensor(
+    uint32_t runner_idx, const InferRequestHeader::Input& input,
+    const Scheduler::Payload& payload, std::vector<int64_t>* shape)
+{
+  // Each runner performs the peek using the corresponding since it
+  // may require a CUDA stream to get the tensor contents.
+  if (runner_idx >= contexts_.size()) {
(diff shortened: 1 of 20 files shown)


src/core/dynamic_batch_scheduler.cc
```diff
@@ -349,32 +353,63 @@ DynamicBatchScheduler::SchedulerThread(
...
+
+      // For shape tensors must compare the contents of the tensor in
+      // addition to the tensor shape itself.
+      if (itr->second) {
+        //        RETURN_IF_ERROR(
+        //            request_provider->PeekShapeTensor(input, &shapes.second));
+      }
+
+      pending_batch_shapes_.emplace(
+          std::make_pair(input.name(), std::move(shapes)));
```
tanmayv25: So as per out discussion, for protecting all these PeekShapeTensor() calls shape tensors must set their respective enforce_equal_shape_tensors_ to true. And this information has to come from the model configuration. Specifically, the IsShapeTensor() must access the model configuration to return the correct property?

src/backends/tensorrt/plan_backend.cc
```diff
@@ -801,6 +824,56 @@ PlanBackend::Context::InitializeConfigOutputBindings(
...
+
+  // If have an provider override for this tensor already (because
+  // peek was already called, then mark it as not being consumed so
+  // that the next peek or usage will be able to access it. If don't
+  // already have an override, add it...
+  if (payload.request_provider_->HasInputOverride(input.name())) {
+    payload.request_provider_->SetInputOverrideConsumed(input.name(), false);
+  } else {
+    InferRequestProvider::InputOverride override;
+    override.content_.assign(dims, dims + element_cnt);
```
tanmayv25: So basically, we are using InputOverride() like this to reset the data in the buffer?

src/core/dynamic_batch_scheduler.cc
deadeyegoodwin: Yes, once the is_shape_tensor is available from the model configuration can re-implement/remove the existing IsShapeTensor() call. enforce_equal_stape_tensors_ isn't a bool, but yes it will be set based upon IsShapeTensor() and if there are not shape tensors the PeekShapeTensor will not be called.

src/backends/tensorrt/plan_backend.cc
deadeyegoodwin: InputOverride is the way you tell the provider to use the "override" data instead of using the data provided by the inference request.  We don't want to have to move the data from the GPU everytime it is peeked so we use the override to cache it.

deadeyegoodwin:
@tanmayv25 Have you incorporated all these changes? Can I close this PR?

PR #984, opened 2019-12-20
Cleanup infer_util

CoderHam:
Fix infer_zero function for variable shape tensors


CoderHam:
Sure, I can make these changes.

GuanLuo:
Yes, what @deadeyegoodwin mentioned is what I intended to mean. You can check if your changes works by removing the workaround I made in L0_batcher without changing the actual test case...

PR #993, opened 2019-12-30
Adding batching support and static batch testing for trt shape tensors

tanmayv25:
Adds the batching support for the shape tensor I/O. 

Tests for dynamic batching will be added in a separate PR

### DIFF
--- a/qa/common/infer_util.py
+++ b/qa/common/infer_util.py
@@ -416,3 +416,167 @@ def infer_zero(tester, pf, batch_size, tensor_dtype, input_shapes, output_shapes
                 shm.destroy_shared_memory_region(shm_op_handles[io_num])
 
     return results
+
+# Perform inference on a model that takes a shape and a dummy tensors as inputs,
+# resize the dummy tensor with the provided values in the shape tensor and finally
+# return the shape of the resized tensor.
+def infer_shape_tensor(tester, pf, batch_size, tensor_dtype, input_shape_values, dummy_input_shapes,
+               model_version=None, use_http=True, use_grpc=True,
+               use_streaming=True):
+    tester.assertTrue(use_http or use_grpc or use_streaming)
+    configs = []
+    if use_http:
+        configs.append(("localhost:8000", ProtocolType.HTTP, False))
+    if use_grpc:
+        configs.append(("localhost:8001", ProtocolType.GRPC, False))
+    if use_streaming:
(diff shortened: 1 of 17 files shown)


qa/common/infer_util.py
```diff
@@ -416,3 +416,167 @@ def infer_zero(tester, pf, batch_size, tensor_dtype, input_shapes, output_shapes
...
+    if use_http:
+        configs.append(("localhost:8000", ProtocolType.HTTP, False))
+    if use_grpc:
+        configs.append(("localhost:8001", ProtocolType.GRPC, False))
+    if use_streaming:
+        configs.append(("localhost:8001", ProtocolType.GRPC, True))
+    tester.assertEqual(len(input_shape_values), len(dummy_input_shapes))
+    io_cnt = len(input_shape_values)
+
+    if TEST_CUDA_SHARED_MEMORY and TEST_SYSTEM_SHARED_MEMORY:
```
deadeyegoodwin: Need to update to match @CoderHam recent change to pass the shared-memory settings in as parameters: https://github.com/NVIDIA/tensorrt-inference-server/pull/984

src/backends/tensorrt/plan_backend.cc
```diff
@@ -442,7 +462,8 @@ PlanBackend::CreateExecutionContext(
 
 Status
 PlanBackend::Context::ValidateInputs(
-    const ::google::protobuf::RepeatedPtrField<ModelInput>& ios)
+    const ::google::protobuf::RepeatedPtrField<ModelInput>& ios,
+    const std::set<std::string> shape_tensors)
```
deadeyegoodwin: const&

src/backends/tensorrt/plan_backend.cc
```diff
@@ -459,7 +499,8 @@ PlanBackend::Context::ValidateInputs(
 
 Status
 PlanBackend::Context::ValidateOutputs(
-    const ::google::protobuf::RepeatedPtrField<ModelOutput>& ios)
+    const ::google::protobuf::RepeatedPtrField<ModelOutput>& ios,
+    const std::set<std::string> shape_tensors)
```
deadeyegoodwin: const&

qa/common/infer_util.py
tanmayv25: I will remember to make these changes once we move to the master. This is to avoid redundant changes along with this PR. Currently, this PR is based on 20.01 release branch.

PR #994, opened 2019-12-31
Rework PlanBackend to use one dedicated thread per device for invoking CUDA API

GuanLuo:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -364,10 +453,62 @@ PlanBackend::CreateExecutionContext(
...
   }
 
-  // Create CUDA stream associated with the execution context
-  const int cuda_stream_priority =
-      GetCudaStreamPriority(Config().optimization().priority());
-  RETURN_IF_ERROR(context->CreateCudaStream(cuda_stream_priority));
+  // Passing the queue for available contexts here so that completion thread
+  // knows where to inform that the context is ready for inputs.
+  std::shared_ptr<SyncQueue<size_t>> lqueue = context_queue;
+  context->completion_thread_ = std::thread([context, context_idx, lqueue]() {
```
deadeyegoodwin: Completion thread functionality is a little big for a lambda. Can you put it in a static function and use that.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -364,10 +453,62 @@ PlanBackend::CreateExecutionContext(
...
+      for (auto& payload : *payloads) {
+        if (payload.stats_ != nullptr) {
+          payload.stats_->CaptureTimestamp(
+              ModelInferStats::TimestampKind::kComputeInputEnd);
+        }
+      }
+#endif  // TRTIS_ENABLE_STATS
+
+      // The model execution associated with the OnCompletePair
+      // has consumed the inputs
```
deadeyegoodwin: "... inputs. Put the context back into the available queue so that it can begin enqueuing new memcpys into the input buffers."

src/backends/tensorrt/plan_backend.cc
```diff
@@ -364,10 +453,62 @@ PlanBackend::CreateExecutionContext(
...
+      cudaEventSynchronize(context->ready_for_input_);
+      lqueue->Put(context_idx);
```
deadeyegoodwin: Potentially tricky race here where you should be calling cudaEventSync(ready_for_output_) before this.... we should discuss.

deadeyegoodwin:
@GuanLuo copyrights also need updating on all the files

src/backends/tensorrt/plan_backend.cc
```diff
@@ -134,22 +178,50 @@ PlanBackend::CreateExecutionContexts(
...
+          if (it == device_to_runner_map.end()) {
+            it = device_to_runner_map
+                     .emplace(gpu_device, available_context_queue_.size())
+                     .first;
+            available_context_queue_.emplace_back(new SyncQueue<size_t>());
+            next_context_.emplace_back(-1);
+          }
+          runner_idx = it->second;
+        }
+        // Initialize all key-value pair in advance so that the map
```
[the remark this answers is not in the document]
GuanLuo: The comment is out-dated... Will remove

PR #995, opened 2020-01-02
Add h2o build

CoderHam:


Dockerfile
```diff
@@ -288,6 +288,32 @@ COPY --from=trtserver_onnx /data/dldt/openvino_2019.1.144/deployment_tools/infer
      /opt/tensorrtserver/lib/
 RUN cd /opt/tensorrtserver/lib && ln -sf libtbb.so.2 libtbb.so
 
+# Install openssl (Needed for h2o)
+RUN apt-get update && apt-get install build-essential checkinstall zlib1g-dev -y
```
deadeyegoodwin: Can we just apt-get for openssl and libuv?
I actually see that 18.04 has an h2o package. Can we just use that?

CoderHam: We need openssl and libuv pkgconfigs to build the h2o library. If we use apt-get to install h2o, it installs an older version (2.2.4) instead and the libh2o.a file is missing in that case

deadeyegoodwin: did you try libh2o-dev package? It has a libh2o.so... can we use that shared library?

CoderHam: I did, it installed an even older version:
`411K /usr/lib/x86_64-linux-gnu/libh2o.so.0.13.4`

> Source: https://github.com/h2o/h2o/wiki/%5Bobsolete%5D-Install-H2O-on-Ubuntu-14.04
... please refer to h2o.github.io/install.html for installing H2O version 1.2.0 or above.

deadeyegoodwin: I'm not sure why the deb is using that versioning. It appears that it is h2o version 2.2.4. Have you tried using the shared library?

CoderHam: Works with the -dev shared library. I had to install h2o-evloop, wslay and numa as well

CoderHam:
> You'll likely have to install non-dev versions of the packages in the final container below so that libh2o is available when trtserver is run.

Done, re-ran CI with new changes.

CoderHam:
Fixed typo, re-ran CI and it passed as well

PR #1002, opened 2020-01-06
Test CORRID control in direct sequence batcher

deadeyegoodwin:


GuanLuo:
The PR only contains new tests and comment update, is this expected?

deadeyegoodwin:
@GuanLuo Yes, it is just a new test to make sure CORRID is working correctly. I fixed the title to be more accurate.

Issue #1001, opened 2020-01-06
CPU memory grows up while using CUDA shared memory


Prasanna19971124:
**Description**
I have been working with TRTIS CUDA shared memory, I am testing YOLOv3 model with three outputs. The client allocates a shared memory during class initialization and uses that memory to read and write tensors. I noticed that cuda shared memory results in increasingly growing main memory, maybe there is some pointer which is not being freed or a new chunk of memory is being allocated every time. 

Here is the code snippet:
```python
class TensorRT_SHM_Client :

  #parameters = dict, Inferecnce parameters
  #shm_parameters = dict, GPU Related shared memory parameters

  def __init__(self, parameters = {}, shm_parameters = {}, batch_size = 15):
    self.parameters = parameters
    model_name = parameters['modelname']
    protocol = ProtocolType.from_str(parameters['protocol'])
    self.healthCTX = ServerHealthContext(parameters['url'], protocol)
    if not self.healthCTX.is_live() :
      print('Health status :', "FAILED")
    status_ctx = ServerStatusContext(parameters['url'], protocol, model_name)
    print('Server status : ', status_ctx.get_server_status())

    self.shm_parameters = shm_parameters
    self.batch_size = batch_size

    #self.labels = load_labels()
    #
    self.inference_ctx = InferContext(parameters['url'], protocol, model_name, parameters["model_version"])
    self.shared_memory_ctx = SharedMemoryControlContext(parameters['url'], protocol)

    self.create_shared_memory(self.batch_size, 416, 416, 3, np.dtype('float32'))


  def create_shared_memory(self, batch_size, w, h, c, dtype):

    print(self.parameters)

    shm_size = batch_size * w * h * c * dtype.itemsize
    shm_op_size = batch_size * 255 * 13 * 13 * dtype.itemsize
    shm_op1_size = batch_size * 255 * 26 * 26 * dtype.itemsize
    shm_op2_size = batch_size * 255 * 52 * 52 * dtype.itemsize

    print('Allocating shared memory with size : ', shm_size)

    self.shm_handle = cudashm.create_shared_memory_region("resnet_50_ip", shm_size, 0)
    self.shm_handle_op = cudashm.create_shared_memory_region("resnet_50_op", shm_op_size, 0)
    self.shm_handle_op1 = cudashm.create_shared_memory_region("resnet_50_op1", shm_op1_size, 0)
[...]

PR #1004, opened 2020-01-07
Ouptut cuda shared memory memory cleanup

CoderHam:
Solves #1001


GuanLuo:
How does this PR solve #1001

CoderHam:
The temp buffer (tmp_addr) used to copy the data from cuda shared memory to host is not needed and is never cleared. The change I made above removes the need for this temporary buffer.

src/clients/python/__init__.py
```diff
@@ -1416,53 +1417,62 @@ def _get_results(self, outputs, batch_size, request_id=None):
...
-                        cval = shm_addr
-                        for b in range(batch_size):
-                            cval_len = start_pos + element_byte_size
-                            if cval_len == 0:
-                                val = np.empty(shape, dtype=result_dtype)
-                                results[output_name].append(val)
-                            else:
-                                val_buf = cast(cval, POINTER(c_byte * cval_len))[0]
-                                val = np.frombuffer(val_buf, dtype=result_dtype, offset=start_pos)
+                    try:
```
GuanLuo: Why `try` block here? I don't see there are early returns in the `try` block

GuanLuo: Or you should move `try` up to cover `_crequest_get_shared_memory_handle_info`. Note that exception may be raised from `_crequest_get_shared_memory_handle_info` after the internal buffer is allocated

CoderHam: We allocate a new memory buffer (in the case of GPU shared memory) in _crequest_get_shared_memory_handle_info. If for some reason there is an exception while processing that into a numpy array, we still want this buffer to be destroyed.

GuanLuo: > Or you should move `try` up to cover `_crequest_get_shared_memory_handle_info`. Note that exception may be raised from `_crequest_get_shared_memory_handle_info` after the internal buffer is allocated

Did you do this?

PR #1005, opened 2020-01-07
Additional Shape Tensor tests

tanmayv25:
Includes the testing of SequenceBatcher and DynamicSequence batcher in L0_trt_shape_tensors

### DIFF
--- a/src/core/sequence_batch_scheduler.cc
+++ b/src/core/sequence_batch_scheduler.cc
@@ -1207,6 +1207,9 @@ OldestSequenceBatch::OldestSequenceBatch(
     if (input.is_shape_tensor()) {
       enforce_equal_shape_tensors.insert({input.name(), true});
     } else if (GetElementCount(input) == -1) {
+      // This should be "true" but some existing custom backend
+      // implementations rely on ragged batches being allowed.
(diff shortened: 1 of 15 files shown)


src/core/sequence_batch_scheduler.cc
```diff
@@ -1197,11 +1197,25 @@ OldestSequenceBatch::OldestSequenceBatch(
...
+  // if the model allows one or more variable-size input tensors or
+  // has shape-tensor inputs. This is not needed if all input shapes
+  // are non-variable and if there are no shape tensors... so we
+  // don't enable it in that case for efficiency reasons.
+  std::unordered_map<std::string, bool> enforce_equal_shape_tensors;
+  for (const auto input : config.input()) {
+    if (input.is_shape_tensor()) {
+      enforce_equal_shape_tensors.insert({input.name(), true});
+    } else if (GetElementCount(input) == -1) {
+      enforce_equal_shape_tensors.insert({input.name(), false});
```
deadeyegoodwin: Put a comment here that this should be "true" but some existing custom backend implementations rely on ragged batches being allowed so need "false" until DLIS-904 is fixed.

PR #1011, opened 2020-01-08
Share CUDA engine when possible

GuanLuo:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -197,6 +193,64 @@ PlanBackend::CreateExecutionContexts(
...
+
+          const auto& mn_itr = models.find(cc_model_filename);
+          if (mn_itr == models.end()) {
+            return Status(
+                RequestStatusCode::INTERNAL, "unable to find PLAN model '" +
+                                                 cc_model_filename + "' for " +
+                                                 Name());
+          }
+
+
```
deadeyegoodwin: delete extra empty line

src/backends/tensorrt/plan_backend.cc
```diff
@@ -197,6 +193,64 @@ PlanBackend::CreateExecutionContexts(
...
+          // Create a CUDA engine shared by all contexts
+          RETURN_IF_ERROR(LoadPlan(
+              mn_itr->second, &eit->second.first, &eit->second.second));
+          // Validate whether the engine can be shared
+          bool is_dynamic = false;
+          auto binding_count = eit->second.second->getNbBindings();
```
deadeyegoodwin: Why is this a separate variable instead of just inline in the for?

src/backends/tensorrt/plan_backend.h
```diff
@@ -202,6 +207,11 @@ class PlanBackend : public InferenceBackend {
     std::vector<void*> buffer_bindings_;
   };
 
+  // [TODO] can an engine be shared across devices?
```
deadeyegoodwin: I don't think you can share an engine across devices...

src/backends/tensorrt/plan_backend.h
```diff
@@ -202,6 +207,11 @@ class PlanBackend : public InferenceBackend {
...
+  // CUDA engine shared across all model instances on the same device.
+  std::map<int, std::pair<nvinfer1::IRuntime*, nvinfer1::ICudaEngine*>>
+      device_engine_;
```
deadeyegoodwin: device_engines_

src/backends/tensorrt/plan_backend.cc
```diff
@@ -197,6 +193,64 @@ PlanBackend::CreateExecutionContexts(
...
+          for (int idx = 0; idx < binding_count; idx++) {
+            auto dims = eit->second.second->getBindingDimensions(idx);
+            // Detect whether dynamic or not
+            if (ContainsWildcard(dims)) {
+              is_dynamic = true;
+              break;
+            }
+          }
+          // Model with dynamic shapes can't share engine, set to engine to
+          // 'nullptr' as hint, but keeping runtime as it can be used repeatly
```
tanmayv25: repeatedly

src/backends/tensorrt/plan_backend.cc
```diff
@@ -197,6 +193,64 @@ PlanBackend::CreateExecutionContexts(
...
+          auto binding_count = eit->second.second->getNbBindings();
+          for (int idx = 0; idx < binding_count; idx++) {
+            auto dims = eit->second.second->getBindingDimensions(idx);
+            // Detect whether dynamic or not
+            if (ContainsWildcard(dims)) {
+              is_dynamic = true;
+              break;
+            }
+          }
+          // Model with dynamic shapes can't share engine, set to engine to
```
tanmayv25: Are you sure dynamic shapes can't share engine? As per what I know, the engine can still be shared but multiple contexts can not use a same optimization profile simultaneously.

GuanLuo: "multiple contexts can not use a same optimization profile simultaneously" is the issue, if we share engine in dynamic shape case, then we need to make sure the instances are aware of which profiles are in used.

PR #1012, opened 2020-01-08
Fix custom backend to set input tensor shape correctly for each payload.

deadeyegoodwin:
Required to support ragged batches in custom backends. Each payload's
input tensors can have a different shape.


deadeyegoodwin:
@GuanLuo There is a follow up change(s) that will make ragged-batch support explicit and also add more testing. ragged-batch will only be allowed for custom backends that enable it as other frameworks don't support it AFAIK.

PR #1013, opened 2020-01-09
Limit CUDA graph to fixed shape model. Fix CUDA graph for dynamic sequence batcher

GuanLuo:


qa/L0_cuda_graph/test.sh
```diff
@@ -0,0 +1,154 @@
...
+fi
+
+set +e
+python $TRT_CUDA_GRAPH_TEST TrtCudaGraphTest.test_fixed_shape>>$CLIENT_LOG 2>&1
+if [ $? -ne 0 ]; then
+    echo -e "\n***\n*** Test Failed\n***"
+    cat $CLIENT_LOG
+    RET=1
+fi
+set -e
```
tanmayv25: you can reduce the unnecessary toggling of -e flag.

qa/L0_cuda_graph/test.sh
```diff
@@ -0,0 +1,154 @@
...
+
+set +e
```
tanmayv25: The redundant toggling of -e setting can be reduced.

PR #1014, opened 2020-01-09
Support inference with different shapes in perf_client

tanmayv25:


docs/perf_client.rst
```diff
@@ -235,20 +235,44 @@ type INT32: ::
     "data" :
      [
         {
-          "INPUT0" : [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
-          "INPUT1" : [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
+          "INPUT0" :
+                {
+                    "content": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
```
deadeyegoodwin: This is not a backwards-compat change. Do you still handle the old format correctly?

tanmayv25: I removed the support for old format. Should I add the support for maintaining backwards compatibility?

deadeyegoodwin: Yes, and keep 1 test that uses the old format.

PR #1016, opened 2020-01-10
Status and Health endpoints for HTTP h2o server

CoderHam:
- [x] Health endpoint **/api/health/live|ready**
- [x] Status endpoint **/api/status?format=text|binary|json**

**hemantj-h2o** will server as the intermediate branch for the upcoming h2o support.


deadeyegoodwin:
@CoderHam Do you have CI running for this?

CoderHam:
> @CoderHam Do you have CI running for this?

Launched one now. I have tested this locally but not added any test cases yet. I will add the HTTP h2o build to L0_build_variants as well.

build/CMakeLists.txt
```diff
@@ -45,6 +45,7 @@ option(TRTIS_ENABLE_CUSTOM "Include Custom backend in server" OFF)
 option(TRTIS_ENABLE_HTTP "Include HTTP API in server" ON)
 option(TRTIS_ENABLE_GRPC "Include GRPC API in server" ON)
 option(TRTIS_ENABLE_HTTP_V2 "Enable version 2 of the HTTP API in server" OFF)
+option(TRTIS_ENABLE_HTTP_H2O "Enable h2o version of the HTTP API in server" OFF)
```
deadeyegoodwin: Since you are doing this on a separate branch you should just make all the changes to replace libevhtp with h2o. So don't add a build flag, remove libevhtp from make, etc.

docs/perf_client.rst
```diff
@@ -332,6 +332,52 @@ multiple streams for a sequence model with a single input named INPUT, shape [1]
 The above example describes three data streams with lengths 4, 3 and 2 respectively.
 The perf_client will hence produce sequences of length 4, 3 and 2 in this case.
 
+Users can also provide an optional "shape" field to the tensors. This is especially
```
deadeyegoodwin: These aren't your changes, you need to rebase.

src/servers/http_h2o_server.cc
```diff
@@ -0,0 +1,690 @@
...
+#include <pthread.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/resource.h>
+#include <sys/signalfd.h>
+#include <sys/time.h>
+#include <unistd.h>
+
```
deadeyegoodwin: Remove blank line. Put all includes together and let formatter order them

src/servers/http_h2o_server.cc
```diff
@@ -0,0 +1,690 @@
...
+// Handle HTTP requests to inference server APIs
+class HTTPAPIServer : public HTTPServer {
+ public:
+  explicit HTTPAPIServer(
+      const std::shared_ptr<TRTSERVER_Server>& server,
+      const std::shared_ptr<nvidia::inferenceserver::TraceManager>&
+          trace_manager,
+      const std::shared_ptr<SharedMemoryBlockManager>& smb_manager,
+      const std::vector<std::string>& endpoints, const int32_t port,
```
deadeyegoodwin: endpoints not used?

src/servers/http_h2o_server.cc
```diff
@@ -0,0 +1,690 @@
...
+      int64_t preferred_memory_type_id, void* userp, void** buffer,
+      void** buffer_userp, TRTSERVER_Memory_Type* actual_memory_type,
+      int64_t* actual_memory_type_id);
+  static TRTSERVER_Error* ResponseRelease(
+      TRTSERVER_ResponseAllocator* allocator, void* buffer, void* buffer_userp,
+      size_t byte_size, TRTSERVER_Memory_Type memory_type,
+      int64_t memory_type_id);
+
+  int create_listener(int32_t port);
+#if H2O_USE_LIBUV
```
deadeyegoodwin: Where is this defined? And is there a reason we need to have 2 choices? If libuv is better we should just use it otherwise we double our testing requirements.

src/servers/http_h2o_server.cc
```diff
@@ -0,0 +1,690 @@
...
+  h2o_custom_req_handler_t* handler =
+      reinterpret_cast<h2o_custom_req_handler_t*>(
+          h2o_create_handler(pathconf, sizeof(*handler)));
+  handler->http_server = this;
+  handler->super.on_req = on_req;
+  return pathconf;
+}
+
+int
+HTTPAPIServer::health(h2o_handler_t* _self, h2o_req_t* req)
```
deadeyegoodwin: _self is not our naming convention.

deadeyegoodwin: ::Health

src/servers/http_h2o_server.cc
```diff
@@ -0,0 +1,690 @@
...
+    return TRTSERVER_ErrorNew(
+        TRTSERVER_ERROR_INVALID_ARG,
+        "HTTP is enabled but none of the service endpoints have a valid "
+        "port assignment");
+  }
+  http_servers->clear();
+  for (auto const& ep_map : port_map) {
+    std::string addr = "0.0.0.0:" + std::to_string(ep_map.first);
+    LOG_INFO << "Starting HTTPService at " << addr;
+    http_servers->emplace_back(new HTTPAPIServer(
```
deadeyegoodwin: Should reconsider class hierarchy based on what h2o supports. Can we support multiple ports within same object? For handlers can we just register the regexs with h2o and it do all the routing to our health, status, etc. endpoints.

src/servers/http_h2o_server.cc
CoderHam: Sure, I can remove this. Libuv should be better but I have yet to test this.

build/CMakeLists.txt
CoderHam: Since support for h2o is not complete I have kept the older libevhtp server for now. I will phase it out once all h2o functionalities are added. Yes, I will also remove the build flag in the future and remove libevhtp from make.

src/servers/http_h2o_server.cc
CoderHam: Fixed

build/CMakeLists.txt
CoderHam: Done

src/servers/http_h2o_server.cc
CoderHam: We cannot support multiple ports with the same object. Yes we just register the endpoints in h2o and let it do the routing to the different endpoints

src/servers/http_server.cc
```diff
@@ -203,17 +74,9 @@ class HTTPAPIServer : public HTTPServerImpl {
...
-        health_regex_(R"(/(live|ready))"),
-        infer_regex_(R"(/([^/]+)(?:/(\d+))?)"), status_regex_(R"(/(.*))"),
-        modelcontrol_regex_(R"(/(load|unload)/([^/]+))"),
-        sharedmemorycontrol_regex_(
-            R"(/(register|cudaregister|unregister|status)(.*))"),
-        repository_regex_(R"(/(index))")
+      : server_(server), trace_manager_(trace_manager),
+        smb_manager_(smb_manager), endpoint_names_(endpoints),
+        allocator_(nullptr), port_(port), thread_cnt_(thread_cnt)
   {
```
deadeyegoodwin: Pull body of construtor and destructor outside of the class like the other functions.

src/servers/http_server.cc
```diff
@@ -283,32 +113,23 @@ class HTTPAPIServer : public HTTPServerImpl {
...
-      evbuffer* input_buffer,
-      TRTSERVER_InferenceRequestProvider* request_provider,
-      std::unordered_map<
-          std::string,
-          std::tuple<const void*, size_t, TRTSERVER_Memory_Type, int64_t>>&
-          output_shm_map);
-
-  static void OKReplyCallback(evthr_t* thr, void* arg, void* shared);
-  static void BADReplyCallback(evthr_t* thr, void* arg, void* shared);
+  int create_listener(int32_t port);
```
deadeyegoodwin: Follow naming convention for all functions

src/servers/http_server.cc
```diff
@@ -283,32 +113,23 @@ class HTTPAPIServer : public HTTPServerImpl {
...
+  int setup_ssl(
+      const char* cert_file, const char* key_file, const char* ciphers);
+  h2o_pathconf_t* register_handler(
+      h2o_hostconf_t* hostconf, const char* path,
+      int (*on_req)(h2o_handler_t*, h2o_req_t*));
+
+  static int Health(h2o_handler_t* handler_self, h2o_req_t* req);
+  static int Status(h2o_handler_t* handler_self, h2o_req_t* req);
+
+  h2o_globalconf_t config;
```
deadeyegoodwin: Member variables end with "_"

src/servers/http_server.cc
```diff
@@ -321,280 +142,291 @@ class HTTPAPIServer : public HTTPServerImpl {
...
 
-  re2::RE2 api_regex_;
-  re2::RE2 health_regex_;
-  re2::RE2 infer_regex_;
-  re2::RE2 status_regex_;
-  re2::RE2 modelcontrol_regex_;
-  re2::RE2 sharedmemorycontrol_regex_;
-  re2::RE2 repository_regex_;
+  int32_t port_;
+  int thread_cnt_;
```
deadeyegoodwin: thread_cnt_ unused

src/servers/http_server.cc
```diff
@@ -283,32 +113,23 @@ class HTTPAPIServer : public HTTPServerImpl {
...
-          std::string,
-          std::tuple<const void*, size_t, TRTSERVER_Memory_Type, int64_t>>&
-          output_shm_map);
-
-  static void OKReplyCallback(evthr_t* thr, void* arg, void* shared);
-  static void BADReplyCallback(evthr_t* thr, void* arg, void* shared);
+  int create_listener(int32_t port);
+  static void on_accept(uv_stream_t* listener, int status);
+  int setup_ssl(
+      const char* cert_file, const char* key_file, const char* ciphers);
```
deadeyegoodwin: Use const std::string&, not char*

src/servers/http_server.cc
```diff
@@ -321,280 +142,291 @@ class HTTPAPIServer : public HTTPServerImpl {
...
+  sock = h2o_uv_socket_create((uv_stream_t*)conn, (uv_close_cb)free);
+  h2o_accept(&http_server->accept_ctx, sock);
 }
 
-TRTSERVER_Error*
-HTTPAPIServer::ResponseRelease(
-    TRTSERVER_ResponseAllocator* allocator, void* buffer, void* buffer_userp,
-    size_t byte_size, TRTSERVER_Memory_Type memory_type, int64_t memory_type_id)
+int
+HTTPAPIServer::create_listener(int32_t port)
```
deadeyegoodwin: listening happends before accept so more this function ahead of on_accept

src/servers/http_server.cc
```diff
@@ -321,280 +142,291 @@ class HTTPAPIServer : public HTTPServerImpl {
...
+  return 0;
+Error:
+  uv_close((uv_handle_t*)&listener, NULL);
+  return r;
 }
 
-void
-HTTPAPIServer::Handle(evhtp_request_t* req)
+int
+HTTPAPIServer::setup_ssl(
```
deadeyegoodwin: Can we delete ssl for now (or #ifdef this out if it will be hard to find this same information later)

src/servers/http_server.cc
```diff
@@ -321,280 +142,291 @@ class HTTPAPIServer : public HTTPServerImpl {
...
+  if (!h2o_memis(req->method.base, req->method.len, H2O_STRLIT("GET"))) {
+    req->res.status = 400;
+    req->res.reason = "Only GET method is allowed";
+    h2o_add_header(
+        &req->pool, &req->res.headers, H2O_TOKEN_CONTENT_TYPE, NULL,
+        H2O_STRLIT("text/plain"));
+    h2o_generator_t generator = {NULL, NULL};
+    h2o_iovec_t body = h2o_strdup(&req->pool, "", SIZE_MAX);
+    h2o_start_response(req, &generator);
+    h2o_send(req, &body, 1, H2O_SEND_STATE_FINAL);
```
deadeyegoodwin: What does this magic "1" mean?

src/servers/http_server.cc
```diff
@@ -321,280 +142,291 @@ class HTTPAPIServer : public HTTPServerImpl {
...
+    return 0;
+  }
+
+  re2::RE2 health_regex_(R"(/api/health/(live|ready))");
```
deadeyegoodwin: Instead of this "second level" regex can we just register both paths and leave it to h2o to do the correct routing

src/servers/http_server.cc
```diff
@@ -321,280 +142,291 @@ class HTTPAPIServer : public HTTPServerImpl {
...
-void
-HTTPAPIServer::HandleStatus(evhtp_request_t* req, const std::string& status_uri)
+int
+HTTPAPIServer::Status(h2o_handler_t* handler_self, h2o_req_t* req)
 {
-  if (req->method != htp_method_GET) {
-    evhtp_send_reply(req, EVHTP_RES_METHNALLOWED);
-    return;
+  h2o_custom_req_handler_t* self = (h2o_custom_req_handler_t*)handler_self;
+  re2::RE2 status_regex(R"(/api/status(/(.*))?)");
```
deadeyegoodwin: I would think there is a way to register a regex and have h2o do the match for the correct handler and then deliver the matched groups to the handler

src/servers/http_server.cc
```diff
@@ -624,791 +458,140 @@ HTTPAPIServer::HandleStatus(evhtp_request_t* req, const std::string& status_uri)
...
 
-  if (byte_size != v[0].iov_len) {
-    return TRTSERVER_ErrorNew(
-        TRTSERVER_ERROR_INVALID_ARG,
-        std::string(
-            "unexpected size for CUDA shared-memory handle, expecting " +
-            std::to_string(byte_size) + " bytes")
-            .c_str());
-  }
+  signal(SIGPIPE, SIG_IGN);
```
deadeyegoodwin: Why ignore sigpipe?

src/servers/http_server.cc
```diff
@@ -624,791 +458,140 @@ HTTPAPIServer::HandleStatus(evhtp_request_t* req, const std::string& status_uri)
...
 
-  size_t offset = std::atoll(offset_str.c_str());
-  size_t byte_size = std::atoll(byte_size_str.c_str());
+  uv_loop_t loop;
+  uv_loop_init(&loop);
+  h2o_context_init(&ctx, &loop, &config);
 
-  TRTSERVER_Error* err = nullptr;
-  TRTSERVER_SharedMemoryBlock* smb = nullptr;
+  if (USE_MEMCACHED)
```
deadeyegoodwin: Delete, why would be use memcached?

src/servers/http_server.cc
```diff
@@ -624,791 +458,140 @@ HTTPAPIServer::HandleStatus(evhtp_request_t* req, const std::string& status_uri)
...
-    err = smb_manager_->CpuCreate(
-        &smb, name.c_str(), shm_key.c_str(), offset, byte_size);
-    if (err == nullptr) {
-      err = TRTSERVER_ServerRegisterSharedMemory(server_.get(), smb);
-    }
-  } else if (action_type_str == "cudaregister") {
-#ifdef TRTIS_ENABLE_GPU
-    int device_id = std::atoll(device_id_str.c_str());
+  // TODO Add server.crt and key files
+  if (USE_HTTPS && setup_ssl(
```
deadeyegoodwin: ifdef
need parens around conditional bodies.

src/servers/http_server.cc
```diff
@@ -624,791 +458,140 @@ HTTPAPIServer::HandleStatus(evhtp_request_t* req, const std::string& status_uri)
...
-        }
-      }
-    }
-  }
+  if (!exit_loop) {
+    uv_close((uv_handle_t*)&listener, NULL);
 
-  if (v_idx != n) {
+    // this will break the event loop
+    exit_loop = true;
```
deadeyegoodwin: Do you need to set this before you uv_close?

src/servers/http_server.cc
```diff
@@ -624,791 +458,140 @@ HTTPAPIServer::HandleStatus(evhtp_request_t* req, const std::string& status_uri)
...
-              std::string(
-                  "unexpected size for input '" + io.name() + "', expecting " +
-                  std::to_string(byte_size) + " bytes for model '" +
-                  model_name + "'")
-                  .c_str());
-        }
-      }
-    }
-  }
+  if (!exit_loop) {
```
deadeyegoodwin: Don't you need a separate thread to run the event loop? Otherwise won't you be looping here until the server is stopped?

src/servers/http_server.cc
CoderHam: Number of buffers (bufcnt)

src/servers/http_server.cc
CoderHam: Put it in an ifdef

src/servers/http_server.cc
CoderHam: Added inline comment as is convention

src/servers/http_server.cc
CoderHam: I will verify but to the best of my understanding h2o just accepts string paths and uses them to check if the url starts with that path. I will read some more and see if it is possible to use a regex instead.

src/servers/http_h2o_server.cc
CoderHam: @deadeyegoodwin unfortunately the does not support regex like input while registering the handler. We will have to use the above approach for now.

src/servers/http_server.cc
CoderHam: "You generally want to ignore the SIGPIPE and handle the error directly in your code. This is because signal handlers in C have many restrictions on what they can do. The most portable way to do this is to set the SIGPIPE handler to SIG_IGN ."

This is the recommended way suggested by h2o. But I can read further and remove it if not needed.

src/servers/http_server.cc
CoderHam: Yes. I will move this to a separate thread.

CoderHam:
Re-working the server to allow handlers to be run on multiple threads as set by **thread_cnt_**.

src/servers/http_server.cc
CoderHam: It is needed so h2o can capture and proceed to kill+cleanup the server accordingly.

CoderHam:
> Re-working the server to allow handlers to be run on multiple threads as set by **thread_cnt_**.

Since h2o does async management + scheduling internally, it is not necessary to allocate multiple threads to it explicitly.  Hence I am removing this argument.

src/servers/main.cc
```diff
@@ -240,8 +234,6 @@ std::vector<Option> options_{
      "The port for the server to listen on for HTTP requests."},
     {OPTION_HTTP_HEALTH_PORT, "http-health-port",
      "The port for the server to listen on for HTTP Health requests."},
-    {OPTION_HTTP_THREAD_COUNT, "http-thread-count",
```
deadeyegoodwin: You can't remove this flag as that would break backwards compatibility, but you can put a note where it is parsed below and just say that it is currently unused.

CoderHam: Sure, I'll do that

CoderHam:
> Looks ok, can you start a CI to ensure that build is ok.

Disabled metrics and launched the CI build

PR #1017, opened 2020-01-11
Add allow_ragged_batch option for input tensors

deadeyegoodwin:


deadeyegoodwin:
@tanmayv25 No. A static-batch request provides a single tensor for each input that is the input required for the entire batch. The tensor has only a single shape so all batch-1's composing it must be that shape.

PR #1027, opened 2020-01-14
Don't allow direct sequence batcher to form ragged batches

deadeyegoodwin:


qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -67,6 +67,17 @@
 if "custom" in _trials:
     _ragged_batch_supported_trials = ("custom",)
 
+# Not all models can be tested for ragged handling because the models
+# don't deal well with non-size-1 shapes
```
GuanLuo: What is non-size-1 shapes?

deadeyegoodwin: shape [ 1 ]. Some of the models don't work if shape is [ 2 ], [ 3 ], etc.

PR #1039, opened 2020-01-17
Use pinned memory as mid-layer for input tensor copies between non-pinned and device memory

GuanLuo:


src/core/provider.h
```diff
@@ -178,16 +178,23 @@ class InferRequestProvider {
...
 
-  // Retrieve the data buffer of input 'name'.
+  // Retrieve the data buffer of input 'name'. This function will not check
+  // input override.
   Status GetMemory(
       const std::string& name, std::shared_ptr<Memory>* input_buffer);
 
+  // Similar to above, but the function caller does not own the Memory object,
+  // nor extend its lifetime. This function will check input override.
+  Status GetMemory(const std::string& name, const Memory** input_buffer);
```
deadeyegoodwin: Can we name GetMemoryWithOverride or similar since otherwise it is somewhat confusing with the other GetMemory

PR #1048, opened 2020-01-22
Validating invalid output name across all backends

tanmayv25:
Currently only TF backends were being tested for a wrong output name (requested name not in model configuration) Moreover, the error was caught by TF when executing the model. 
This PR adds explicit checks and tests across all the frameworks for invalid output names in request payload.

### DIFF
--- a/src/backends/tensorrt/plan_backend.cc
+++ b/src/backends/tensorrt/plan_backend.cc
@@ -1062,6 +1062,12 @@ PlanBackend::Context::Run(
               name_ + "'");
     }
 
+    // Check if all the requested output names in the request header are valid
+    for (const auto& output :
+         payload.request_provider_->RequestHeader().output()) {
+      RETURN_IF_ERROR(base->ValidateOutputName(output.name()));
+    }
+
     total_batch_size += payload.request_provider_->RequestHeader().batch_size();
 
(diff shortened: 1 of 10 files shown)


src/backends/tensorrt/plan_backend.cc
```diff
@@ -1062,6 +1062,12 @@ PlanBackend::Context::Run(
               name_ + "'");
     }
 
+    // Check if all the requested output names in the request header are valid
+    for (const auto& output :
+         payload.request_provider_->RequestHeader().output()) {
+      RETURN_IF_ERROR(base->ValidateOutputName(output.name()));
```
GuanLuo: You can move this check to `NormalizeRequestHeader()` as this can be detected earlier and ease the load on scheduler and model instances.

src/core/backend.cc
```diff
@@ -40,6 +40,20 @@
 
 namespace nvidia { namespace inferenceserver {
 
+Status
+InferenceBackend::ValidateOutputName(const std::string& name) const
```
GuanLuo: After going through `NormalizeRequestHeader()`, I found an existing [`GetOutput()`](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/core/backend.cc#L58-L71) function which may come in handy.

PR #1049, opened 2020-01-22
Fix build break due to deprecated Python2

deadeyegoodwin:


CoderHam:
These changes aim to change our CI tests to use only Python3 but we still want to support a Python2 client build as well? If so shouldn't we be testing the python2 client in some sections of the CI just for coverage?

deadeyegoodwin:
@GuanLuo Python2 is deprecated so we don't need to test it.

tanmayv25:
As we no longer test python2 we must update [here](https://github.com/NVIDIA/tensorrt-inference-server/blob/aa8b7f7fc8cc115dea1900be169c7afe745b44b4/src/clients/python/library/setup.py#L52) as well. py2 should not be used in the naming.

PR #1051, opened 2020-01-23
Add basic Jetson build 

CoderHam:
Supports only TF for now.


jetson_build.sh
```diff
@@ -0,0 +1,109 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+TRTIS_VERSION=1.10.0jetson-dev
+
+# Git clone repo from github
+mkdir $HOME/trtis && cd ${HOME}/trtis && \
```
deadeyegoodwin: Be consistent with { } around variable

jetson_build.sh
```diff
@@ -0,0 +1,109 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin: Add "#!/bin/bash" as first line and set executable mode on file

jetson_build.sh
```diff
@@ -0,0 +1,109 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+TRTIS_VERSION=1.10.0jetson-dev
```
deadeyegoodwin: Initialize this from the VERSION file. So move after cloning the repo. Version will be same as on x86, so don't put "jetson" in it. Something like:

TRTIS_VERSION=`cat VERSION`

jetson_build.sh
```diff
@@ -0,0 +1,109 @@
...
+
+# Git clone repo from github
+mkdir $HOME/trtis && cd ${HOME}/trtis && \
+  git clone --recursive \
```
deadeyegoodwin: --recursive is not needed, but use --single-branch --depth=1

jetson_build.sh
```diff
@@ -0,0 +1,109 @@
...
+            libre2-dev \
+            libssl-dev \
+            libtool \
+            libboost-dev \
+            libh2o-dev \
+            libh2o-evloop-dev \
+            libnuma-dev \
+            libwslay-dev \
+            libuv1-dev && \
+    if [ $(cat /etc/os-release | grep 'VERSION_ID="16.04"' | wc -l) -ne 0 ]; then \
```
deadeyegoodwin: We should just support what we are actually testing, so just 18.04 I would think

jetson_build.sh
```diff
@@ -0,0 +1,109 @@
...
+                  -DTRTIS_EXTRA_LIB_PATHS="/opt/tensorrtserver/lib" \
+                  ../build && \
+            make -j16 trtis && \
+            mkdir -p /opt/tensorrtserver/include && \
+            cp -r trtis/install/bin /opt/tensorrtserver/. && \
+            cp -r trtis/install/lib /opt/tensorrtserver/. && \
+            cp -r trtis/install/include /opt/tensorrtserver/include/trtserver)
+
+# TODO publish this to a fixed data repository 
+tar -zcvf tensorrtserver${TRTIS_VERSION}.tgz /opt/tensorrtserver
```
deadeyegoodwin: Should have "jetpack-x.y.z" or similar in the filename. Can you find the version of jetpack in some envvar on the build system?

jetson_build.sh
CoderHam: Yes, 18.04

jetson_build.sh
CoderHam: They do not store the jetpack version in an environment variable
`echo $(dpkg-query --showformat='${Version}' --show nvidia-l4t-core)`
> 32.3.1-20191209230245

From this script [here](https://github.com/dusty-nv/jetson-inference/blob/7e81381a96c1ac5f57f1728afbfdec7f1bfeffc2/tools/install-pytorch.sh#L321) I was able to get the JetPack version

CoderHam: Followed the following [script](https://github.com/jetsonhacks/jetsonUtilities/blob/7623cb3b37aca11ff8f5e3bd0f63d4f595b4d21e/scripts/jetson_variables#L91) for now.

PR #1066, opened 2020-01-28
Status and Model Control API

tanmayv25:


src/clients/python/experimental_api_v2/examples/simple_client_v2.py
```diff
@@ -43,26 +43,55 @@
...
+    ### Model Control API ###
+    TRTISClient.load(model_name)
+    if TRTISClient.is_model_ready(model_name, 1):
+        print("PASS: load")
+
+    TRTISClient.load(model_name_2)
+    if TRTISClient.is_model_ready(model_name_2, 1):
+        print("PASS: load_2")
+
+    ### Health API ###
```
deadeyegoodwin: You should check server live and ready first.

src/clients/python/experimental_api_v2/examples/simple_client_v2.py
```diff
@@ -43,26 +43,55 @@
...
     if TRTISClient.is_server_live():
-        print("SUCCESS: is_server_live")
+        print("PASS: is_server_live")
 
     if TRTISClient.is_server_ready():
-        print("SUCCESS: is_server_ready")
+        print("PASS: is_server_ready")
+
+    if TRTISClient.is_model_ready(model_name, 1):
```
deadeyegoodwin: This is redundant with the checks in model control part

src/clients/python/experimental_api_v2/library/__init__.py
```diff
@@ -105,5 +209,126 @@ def is_server_ready(self):
             If unable to get readiness.
 
         """
-        self.response = self.client_stub.get("/api/health/ready")
+        self.response = self._client_stub.get("/api/health/ready")
+        self._last_request_id = _raise_if_error(self.response['NV-Status'])
         return self.response.status_code == 200
+
+    def is_model_ready(self, model_name, model_version):
```
deadeyegoodwin: model_version should be optional and default to -1, which means that the server will choose the version based on its version policy for the model.  Update docs and simple example appropriately.

src/clients/python/experimental_api_v2/library/__init__.py
```diff
@@ -105,5 +209,126 @@ def is_server_ready(self):
...
+            The version of the model
+
+        Returns
+        -------
+        bool
+            True if the model is ready to be served, False if otherwise.
+
+        Raises
+        ------
+        Exception
```
deadeyegoodwin: InferenceServerException

src/clients/python/experimental_api_v2/library/__init__.py
```diff
@@ -105,5 +209,126 @@ def is_server_ready(self):
...
+        """
+
+        self.response = self._client_stub.get("/api/status/" + model_name +
+                                              "?format=json")
+        self._last_request_id = _raise_if_error(self.response['NV-Status'])
+
+        status = json.loads(self.response.read())
+        return status
+
+    def load(self, model_name):
```
deadeyegoodwin: load_model

src/clients/python/experimental_api_v2/library/__init__.py
```diff
@@ -105,5 +209,126 @@ def is_server_ready(self):
...
+        """Request the inference server to load specified model.
```
deadeyegoodwin: or reload

src/clients/python/experimental_api_v2/library/__init__.py
```diff
@@ -105,5 +209,126 @@ def is_server_ready(self):
...
+        ------
+        Exception
+            If unable to load the model.
+
+        """
+        self.response = self._client_stub.post("/api/modelcontrol/load/" +
+                                               model_name)
+        self._last_request_id = _raise_if_error(self.response['NV-Status'])
+
+    def unload(self, model_name):
```
deadeyegoodwin: unload_model

src/clients/python/experimental_api_v2/library/__init__.py
tanmayv25: The api to retrieve model readiness with or without model version is a part of V2 right? Here as a workaround I am retrieving the model readiness by getting model_status first and then looking for the MODEL_READY value. I have made a note to update when integrating with HTTP V2.

src/clients/python/experimental_api_v2/library/__init__.py
tanmayv25: Even though I have modified Exception->InferenceServerException, all other kinds of exceptions can be raised by the underlying python modules. Would it be appropriate to document InferenceServerException here? Should we be catching these exceptions and raising them as an InferenceServerException, at the expense of visibility of the stack?

PR #1068, opened 2020-01-28
Update ONNX Runtime to v1.1.1

deadeyegoodwin:


Dockerfile
```diff
@@ -288,11 +281,15 @@ RUN cd /opt/tensorrtserver/lib/onnx && \
 
 # Minimum OpenVINO libraries required by ONNX Runtime to link and to run
 # with OpenVINO Execution Provider
-COPY --from=trtserver_onnx /data/dldt/openvino_2019.1.144/deployment_tools/inference_engine/lib/intel64/libinference_engine.so \
+COPY --from=trtserver_onnx /data/dldt/openvino_2019.3.376/deployment_tools/inference_engine/lib/intel64/libinference_engine.so \
+     /opt/tensorrtserver/lib/onnx/
+COPY --from=trtserver_onnx /data/dldt/openvino_2019.3.376/deployment_tools/inference_engine/lib/intel64/libcpu_extension.so \
+     /opt/tensorrtserver/lib/onnx/
+COPY --from=trtserver_onnx /data/dldt/openvino_2019.3.376/deployment_tools/inference_engine/lib/intel64/plugins.xml \
```
clockfly: The plugins.xml will trigger a build error 
```
patchelf --set-rpath $ORIGIN ./libtbb.so.2
patchelf --set-rpath $ORIGIN ./libcpu_extension.so
patchelf --set-rpath $ORIGIN ./libMKLDNNPlugin.so
patchelf --set-rpath $ORIGIN ./libonnxruntime.so.1.1.1
patchelf --set-rpath $ORIGIN ./libinference_engine.so
patchelf --set-rpath $ORIGIN ./plugins.xml
not an ELF executable
```

deadeyegoodwin: Yes, but doesn't seem to cause the build to stop. But we will look into fixing.

deadeyegoodwin: Fixed with https://github.com/NVIDIA/tensorrt-inference-server/pull/1156

PR #1075, opened 2020-01-30
Fix logging for GRPC and HTTP endpoints

deadeyegoodwin:


src/servers/main.cc
```diff
@@ -986,6 +987,16 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
     return false;
   }
 
+#ifdef TRTIS_ENABLE_LOGGING
+  // Initialize our own logging instance since it is used by GRPC and
+  // HTTP endpoints. This logging instance is separate from the one in
```
GuanLuo: Don't we only have one global logger `gLogger_`? How is this different from calling `TRTSERVER_ServerOptionsSetLogXXX`

deadeyegoodwin: libtrtserver.so doesn't expose its gLogger_. TRTSERVER_ServerOptionsSetLogXXX gets passed into libtrtserver.so and that sets the gLogger_ (in trtserver.cc I think). But we compile logging.cc into trtserver which gives us another gLogger_ that needs to be initialized in main.cc.

PR #1076, opened 2020-01-31
Jetson compatibility changes for Infer test

CoderHam:


qa/L0_infer/test.sh
```diff
@@ -130,7 +132,11 @@ for TARGET in cpu gpu; do
     for FW in $BACKENDS; do
       if [ "$FW" != "plan" ]; then
         for MC in `ls models/${FW}*/config.pbtxt`; do
-            echo "instance_group [ { kind: ${KIND} }]" >> $MC
+            if [ "$FW" == "custom" ]; then
```
deadeyegoodwin: Doesn't L0_infer use the addsub custom backend? That backend does support both CPU and GPU.

CoderHam: We built the custom backends (Dockerfile.custombackend) with TRTIS_ENABLE_GPU=OFF.
When we do this, the GPU instance is not supported.

CoderHam: Fixed.

PR #1084, opened 2020-02-04
Inference support h2o

CoderHam:


deadeyegoodwin:
@CoderHam can you rebase this to master TOT

CoderHam:
> @CoderHam can you rebase this to master TOT

Done. Rebased `hemantj-h2o` as well

src/servers/http_server.cc
```diff
@@ -69,6 +71,49 @@ class HTTPAPIServer : public HTTPServer {
       const std::vector<std::string>& endpoints, const int32_t port);
   ~HTTPAPIServer();
 
+  using ResponseMetaData = std::pair<
```
deadeyegoodwin: Use a proper class for this. See grpc_server.cc for how the response meta-data is handled and use an equivalent class/struct here. In general, parallel as much as possible what is in grpc_server.cc.

src/servers/http_server.cc
```diff
@@ -462,6 +711,250 @@ HTTPAPIServer::Status(h2o_handler_t* handler_self, h2o_req_t* req)
...
+    return 0;
+  }
+
+  int64_t model_version = -1;
+  if (!model_version_str.empty()) {
+    model_version = std::atoll(model_version_str.c_str());
+  }
+
+#ifdef TRTIS_ENABLE_TRACING
+  // Timestamps from evhtp are capture in 'req'. We record here since
```
deadeyegoodwin: Fix comment

src/servers/http_server.cc
```diff
@@ -462,6 +711,250 @@ HTTPAPIServer::Status(h2o_handler_t* handler_self, h2o_req_t* req)
...
+  if (self->http_server->trace_manager_ != nullptr) {
+    trace_meta_data.reset(self->http_server->trace_manager_->SampleTrace());
+    if (trace_meta_data != nullptr) {
+      trace_meta_data->tracer_->SetModel(model_name, model_version);
+      trace_meta_data->tracer_->CaptureTimestamp(
+          TRTSERVER_TRACE_LEVEL_MIN, "http recv start",
+          TIMEVAL_TO_NANOS(req->timestamps.request_begin_at));
+      trace_meta_data->tracer_->CaptureTimestamp(
+          TRTSERVER_TRACE_LEVEL_MIN, "http recv end",
+          TIMEVAL_TO_NANOS(req->timestamps.request_body_begin_at));
```
deadeyegoodwin: The beginning of the body is not the "http recv end". If there a timestamp or a callback where we can capture the time after the entire http request, including body, has been received?

src/servers/http_server.cc
```diff
@@ -462,6 +711,250 @@ HTTPAPIServer::Status(h2o_handler_t* handler_self, h2o_req_t* req)
...
+    }
+  }
+#endif  // TRTIS_ENABLE_TRACING
+
+  // Work around for h2o bug forces headers to lower case
```
deadeyegoodwin: Since this is a constant just define it in constants.h as kInferRequestHTTPHeaderLowerCase.

src/servers/http_server.cc
```diff
@@ -462,6 +711,250 @@ HTTPAPIServer::Status(h2o_handler_t* handler_self, h2o_req_t* req)
...
+      h2o_find_header_by_str(&req->headers, H2O_STRLIT(kInferHeaderLower), -1);
+
+  InferRequestHeader request_header_protobuf;
+  if (infer_request_cursor != -1) {
+    h2o_iovec_t* slot = &req->headers.entries[infer_request_cursor].value;
+    std::string infer_request_header = std::string(slot->base, slot->len);
+    if (!google::protobuf::TextFormat::ParseFromString(
+            infer_request_header, &request_header_protobuf)) {
+      req->res.status = 400;
+      req->res.reason = "Bad Request";
```
deadeyegoodwin: "Unable to parse inference request"

src/servers/http_server.cc
```diff
@@ -69,6 +71,49 @@ class HTTPAPIServer : public HTTPServer {
       const std::vector<std::string>& endpoints, const int32_t port);
   ~HTTPAPIServer();
 
+  using ResponseMetaData = std::pair<
+      std::vector<h2o_iovec_t*>,
+      std::unordered_map<
+          std::string,
+          std::tuple<const void*, size_t, TRTSERVER_Memory_Type, int64_t>>>;
+
+  // Class object associated to evhtp thread, requests received are bounded
```
deadeyegoodwin: Update comment

src/servers/http_server.cc
CoderHam: There is no timestamp for capturing the entire request end. There is however a timestamp for when it is done processing the received request. 
https://github.com/h2o/h2o/blob/07459687b819ece4f47ce2c4b5bfd2fae5e287e4/include/h2o.h#L1067
Would this be a more suitable timestamp?

deadeyegoodwin: Maybe. Or perhaps we should just capture it ourselves when h2o calls use with a complete request.

CoderHam: That would be only after the request is _processed_ and the respective handler is 'hit'

PR #1086, opened 2020-02-04
Fix L0_model_config

GuanLuo:


GuanLuo:
L0_model_config test has passed.

PR #1087, opened 2020-02-04
Fixing the dynamic batcher to work with TRT backend

tanmayv25:


GuanLuo:
Why did the abstraction cause hanging?

src/core/dynamic_batch_scheduler.cc
```diff
@@ -113,8 +113,8 @@ DynamicBatchScheduler::Create(
         "Initialization failed for all dynamic-batch scheduler threads");
   }
 
-  sched->completion_promises_ =
-      std::vector<std::shared_ptr<std::promise<void>>>(
+  sched->completion_queue_ =
```
deadeyegoodwin: completion_queues_ would be clearer since there are multiple queues.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -284,14 +283,7 @@ DynamicBatchScheduler::SchedulerThread(
...
           if (preserve_ordering_) {
-            // There is runner processing payloads before current runner
-            if (last_processing_runner_id_ != -1) {
-              completion_promise =
-                  completion_promises_[last_processing_runner_id_];
-            }
-            last_processing_runner_id_ = runner_id;
-            completion_promises_[runner_id] =
-                std::make_shared<std::promise<void>>();
+            runner_queue_.push(runner_id);
```
deadeyegoodwin: This should be moved to line 318 since we need to preserve ordering across instances even when not using the dynamic batcher

runner_queue is not protected here by completion_queue_mtx

src/core/dynamic_batch_scheduler.cc
tanmayv25: Done

src/core/dynamic_batch_scheduler.cc
tanmayv25: Fixed

tanmayv25:
@GuanLuo  Actually, the commit reverses **some** of the changes introduced by a previous PR #1085. If we are already pausing the workers, then it is not needed to hold the lock till the actual issue of the request. Each worker is responsible for an exclusive set of sequences. The issue previous PR tried to solve was to avoid a case where one thread is in the process of sending a request for nth sequence, when change in concurrency is requested. In this case workers re-distribute the sequences among themselves and another worker may race ahead of the previous worker and issue its request. Hence, waiting for all threads to complete all the requests before changing concurrency solves the problem.

To answer your question, it was not just the abstraction. There was a bug in the way failing cases were being handled. For now I have just reverted the change back to original. The abstraction will be reintroduced with another PR.

tanmayv25:
Skipping test_dynaseq_different_shape_values_parallel as it is still hanging some times.

PR #1089, opened 2020-02-05
Fix test breaks related to new upstreams

deadeyegoodwin:


src/clients/c++/examples/CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ install(
 # Install cmake files that builds the samples within the client install.
 install(
   FILES
-    ${CMAKE_BINARY_DIR}/../CMakeLists.txt
```
CoderHam: Add 2020 to the copyright

qa/L0_http_v2/test.sh
```diff
@@ -27,13 +27,38 @@
 
 set +e
 
-SIMPLE_V2_CLIENT=/workspace/install/bin/simple_v2_client
+SIMPLE_V2_CLIENT=/workspace/builddir/trtis-clients/install/bin/simple_v2_client
 
 RET=0
```
CoderHam: Move this down to just before the client is run (line 107)

PR #1090, opened 2020-02-06
Python2 is EOL, remove limited support that was present

deadeyegoodwin:


CoderHam:
There are some remnants of Python2 code like that is no longer needed (in python clients and utils)
```
from builtins import range
from future.utils import iteritems
from builtins import str
```
And possibly others

deadeyegoodwin:
@CoderHam Agreed but instead of trying to remove those now I think we should just clean them up as we work on the version 2 clients and examples.

PR #1103, opened 2020-02-12
Python client for GRPC V2: Build, Health, Metadata and basic Infer

tanmayv25:
Please ignore the contents of httpclient.py for now.


src/clients/python/experimental_api_v2/examples/simple_grpc_pyclient_v2.py
```diff
@@ -0,0 +1,87 @@
...
+        print("PASS: is_server_live")
+
+    if TRTISClient.is_server_ready():
+        print("PASS: is_server_ready")
+
+    if TRTISClient.is_model_ready(model_name):
+        print("PASS: is_model_ready")
+
+    metadata = TRTISClient.get_server_metadata()
+    if (metadata['name'] == 'inference:0'):
```
deadeyegoodwin: If metadata here just to protobuf object? Can you use "metadata.name"?

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,273 @@
...
+        except grpc.RpcError as rpc_error:
+            raise_error_grpc(rpc_error)
+
+    def get_server_metadata(self):
+        """Contact the inference server and get its metadata.
+
+        Returns
+        -------
+        dict
+            The JSON object holding the metadata
```
deadeyegoodwin: Since this is the GRPC client API there should be a way to get the protobuf objects directly. Perhaps have an option to request JSON return value that defaults to false.

Same for other methods below.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+        Parameters
+        ----------
+        inputs : list
+            A list of InferInput objects, each describing data for a input
+            tensor required by the model.
+        outputs : list
+            A list of InferOutput objects, each describing how the output
+            data must be returned. Only the output tensors present in the
+            list will be requested from the server.
+        batch_size : int
```
deadeyegoodwin: We shouldn't need the batch-size to be explicitly specified.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+            The batch size of the inference. Each input must provide
+            an appropriately sized batch of inputs.
+        request_id: string
+            Optional identifier for the request which will be
+            returned in the response. Default value is '0'.
```
deadeyegoodwin: Default value should be None, which means that no request ID will be used.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+
+        Parameters
+        ----------
+        inputs : list
+            A list of InferInput objects, each describing data for a input
+            tensor required by the model.
+        outputs : list
+            A list of InferOutput objects, each describing how the output
+            data must be returned. Only the output tensors present in the
+            list will be requested from the server.
```
deadeyegoodwin: Describe model_name and model_version parameters.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+        batch_size : int
+            The batch size of the inference. Each input must provide
+            an appropriately sized batch of inputs.
+        flags : int
```
deadeyegoodwin: Flags here but not above, and no request_id here. The flags should be the same on infer and async_infer (except callback).

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+
+        Parameters
+        ----------
+        numpy_array : numpy array
+            The tensor data in numpy array format
+        """
+        self._input.datatype = np_to_trtis_dtype(numpy_array.dtype)
+        self._input.shape.extend(numpy_array.shape)
+        self._input_contents = grpc_service_v2_pb2.InferTensorContents()
+        self._input_contents.raw_contents = numpy_array.tobytes()
```
deadeyegoodwin: Do you have to make sure it is flattened in row-major order? Or is that the default?

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+        self._input.contents.CopyFrom(self._input_contents)
```
deadeyegoodwin: This copy can be very expensive... can't we just assign directly to input.contents.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+    Parameters
+    ----------
+    result : Protobuf Message
+        The ModelInferResponse returned by the server
+    """
+
+    def __init__(self, result):
+        self._result = result
+
+    def get_output_in_numpy(self, name):
```
deadeyegoodwin: How about just "as_numpy" since this class is already known to be an output/result.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+        Protobuf Message or dict
+            The ModelInferRequest protobuf message or dict for the request
+            associated  with this response.
+        """
+        if as_json:
+            return json.loads(MessageToJson(self._result.request))
+        else:
+            return self._result.request
+
+    def get_statistics(self, as_json=False):
```
deadeyegoodwin: We should use consistent naming for the json option throughput.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -0,0 +1,674 @@
...
+        -------
+        Protobuf Message or dict
+            The InferStatistics protobuf message or dict for this response.
+        """
+        if as_json:
+            return json.loads(MessageToJson(self._result.statistics))
+        else:
+            return self._result.statistics
+
+    def get_complete_response(self, as_json=False):
```
deadeyegoodwin: "get_response"

PR #1104, opened 2020-02-13
Insert NVTX range for capturing non-CUDA events in Nsys

GuanLuo:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -1908,6 +1909,9 @@ PlanBackend::Context::ProcessResponse(
     // so that it can begin enqueuing new memcpys into the input buffers
     cudaEventSynchronize(event_set.ready_for_input_);
     context_queue->Put(context_idx);
+    {
+      NVTX_RANGE(nvtx_, "plan_input_available");
```
deadeyegoodwin: NVTX has the concept of a marker that is just a single instance in time. https://docs.nvidia.com/cuda/profiler-users-guide/index.html#nvtx-marker
You should add a NVTX_MARKER macro and use that.

PR #1109, opened 2020-02-14
Add queue policy settings in dynamic batch scheduler. Add priority level for policy queues

GuanLuo:


src/core/api.proto
```diff
@@ -223,6 +223,25 @@ message InferRequestHeader
   //@@     request.
   //@@
   repeated Output output = 3;
+
+  //@@  .. cpp:var:: uint32 priority [DLIS-1083]
```
deadeyegoodwin: Remove DLIS since not meaningful externally and these comments generate the docs.

src/core/api.proto
```diff
@@ -223,6 +223,25 @@ message InferRequestHeader
...
+  //@@
+  //@@     The priority value of this request. If the priority value is not
```
deadeyegoodwin: "If priority handling is not enabled for the model, then this value is ignored."

src/core/api.proto
```diff
@@ -223,6 +223,25 @@ message InferRequestHeader
...
+
+  //@@  .. cpp:var:: uint64 timeout_limit_microseconds [DLIS-997]
+  //@@
+  //@@     The timeout limit specified for this request. This value will
+  //@@     override timeout limit set by the model if the model allows timeout
+  //@@     limit override. When timeout expired, the model will act based on its
+  //@@     timeout policy. The default value is 0 which indicates that the
+  //@@     request will not have timeout limit.
+  //@@
+  uint64 timeout_limit_microseconds = 7;
```
deadeyegoodwin: timeout_microseconds

src/core/api.proto
```diff
@@ -223,6 +223,25 @@ message InferRequestHeader
...
+  //@@     The priority value of this request. If the priority value is not
+  //@@     enable by the model, then the request will not be prioritized.
+  //@@     The default value is 0 which indicates that the request will not be
+  //@@     prioritized.
+  //@@
+  uint32 priority = 6;
+
+  //@@  .. cpp:var:: uint64 timeout_limit_microseconds [DLIS-997]
+  //@@
+  //@@     The timeout limit specified for this request. This value will
```
deadeyegoodwin: "The timeout for this request. This value overrides the timeout specified by the model, if the model allows timeout override. If the request cannot be processed within this timeout, the request will be handled based on the model's timeout policy."

src/core/model_config.proto
```diff
@@ -623,6 +623,75 @@ message ModelOptimizationPolicy
   PinnedMemoryBuffer output_pinned_memory = 6;
 }
 
+//@@
+//@@.. cpp:var:: message RequestTimeoutPolicy
+//@@
+//@@   Timeout policy for inference requests. If specified, the scheduler will
+//@@   handle pending requests that exceed timeout accordingly.
+//@@
+message RequestTimeoutPolicy
```
deadeyegoodwin: ModelTimeoutPolicy

src/core/model_config.proto
```diff
@@ -623,6 +623,75 @@ message ModelOptimizationPolicy
...
+  enum Action {
+    //@@    .. cpp:enumerator:: Action::REJECT = 0
+    //@@
+    //@@       Reject the request and return error message accordingly.
+    //@@
+    REJECT = 0;
+
+    //@@    .. cpp:enumerator:: Action::DELAY = 1
+    //@@
+    //@@       Delay the request until all other requests within the timeout
```
deadeyegoodwin: "Delay the request until all other requests at the same (or higher) priority levels that have not reached their timeouts are processed. A delayed request will eventually be processed, but may be delayed indefinitely due to newly arriving requests."

Remove "The priority...". We need to allow a timeout policy for each priority level.

src/core/model_config.proto
```diff
@@ -623,6 +623,75 @@ message ModelOptimizationPolicy
...
+  //@@
+  Action action = 1;
+
+  //@@
+  //@@  .. cpp:var:: uint64 timeout_limit_microseconds
+  //@@
+  //@@     The timeout limit for every request, in microseconds.
+  //@@     The default value is 0 which indicates that no timeout limit is set.
+  //@@
+  uint64 timeout_limit_microseconds = 2;
```
deadeyegoodwin: default_timeout_microseconds

src/core/model_config.proto
```diff
@@ -623,6 +623,75 @@ message ModelOptimizationPolicy
...
+
+  //@@
+  //@@  .. cpp:var:: bool allow_override [DLIS-997]
+  //@@
+  //@@     Whether individual request can override the timeout limit.
+  //@@     The default value is false.
+  //@@
+  bool allow_override = 3;
```
deadeyegoodwin: allow_timeout_override

src/core/model_config.proto
```diff
@@ -661,6 +730,22 @@ message ModelDynamicBatching
   //@@     order.
   //@@
   bool preserve_ordering = 3;
+
```
deadeyegoodwin: We need to talk some more about the implementation. I think we will need to have the config specify the number of priority levels (since we will need a queue for each we will encourage the use of as few as possible). And then for each there can be a timeout policy. The default will handle the case where priority is disabled and for priority levels where there is no specific policy given.

src/core/model_config.proto
GuanLuo: Maybe we should also make it more general like "ModelQueuePolicy"? If we are going to add max queue size, I think it can be added here.

deadeyegoodwin: Sounds good. And then have 1 ModelQueuePolicy per priority level (plus a default one).

src/core/model_config.proto
GuanLuo: I agree. But I don't think we need a separate field to specify the number of priority levels, if the user wants to specify the priority level as they need, they just specify it in the map below. The empty map simply means priority handling is disabled, as all request will fallback to default policy.

deadeyegoodwin: That works but introduces a couple of cases where the user has to do something they don't want to. For example, what if I want 3 priority levels (0, 1, 2) that all use a default policy. I need to have a default policy but also a policy for priority level 2 so that I can add 2 to the map with that policy (I assume that we determine the total number of priority levels from the highest value in the map).

Similarly user always has to have a non-default policy for the highest priority level since that needs to go in the map, even if that policy is the same as the default.

GuanLuo:
Updated the protobuf files based on the comments.

src/core/model_config.proto
```diff
@@ -623,6 +623,70 @@ message ModelOptimizationPolicy
...
+  //@@
+  //@@     Whether individual request can override the timeout value.
+  //@@     The default value is false.
+  //@@
+  bool allow_timeout_override = 3;
+
+  //@@
+  //@@  .. cpp:var:: uint32 max_queue_size
+  //@@
+  //@@     The maximum queue size for holding requests. The request enqueued
```
deadeyegoodwin: "The maximum ... . A request will be rejected immediately if it can't be enqueued because the queue is full. The default value is 0 ..."

Right now we always reject. In the future we may want to add a policy setting for queue size as well. For example, we might want to allow the option to have the request enqueue in a lower-priority queue if its queue is full.

src/core/model_config.proto
```diff
@@ -661,6 +725,31 @@ message ModelDynamicBatching
...
+  //@@
+  ModelQueuePolicy default_queue_policy = 4;
+
+  //@@  .. cpp:var:: uint32 number_of_priority_level
+  //@@
+  //@@     The number of priority levels to be enabled for the model,
+  //@@     the priority level starts from 1 and the requests with higher
+  //@@     priority level will be prioritized than other requests.
+  //@@
+  uint32 number_of_priority_level = 5;
```
deadeyegoodwin: priority_levels

src/core/model_config.proto
```diff
@@ -661,6 +725,31 @@ message ModelDynamicBatching
...
+  //@@     The default queue policy used for requests that don't require
+  //@@     priority handling and requests that specify priority levels where
+  //@@     there is no specific policy given. If not specified, a policy with
+  //@@     default field values will be used.
+  //@@
+  ModelQueuePolicy default_queue_policy = 4;
+
+  //@@  .. cpp:var:: uint32 number_of_priority_level
+  //@@
+  //@@     The number of priority levels to be enabled for the model,
```
deadeyegoodwin: Any reason why we don't start priority level at 0?
You should make the highest priority be level 0 and higher priority values be lower priority.

GuanLuo: There is a priority level field in request header, and the default value is 0 if user doesn't specify a priority level. Given that, 0 is reserved for "no priority".

deadeyegoodwin: I see. If the request doesn't specify a priority what queue are you going to put it in? You have a default_queue_policy but you still need to know what queue to use. Seems like you also need a default_priority_level.

Also for priority_levels should also have a value of 0 (the default) mean that there is no prioritization at all... I think that it is equivalent to priority_levels == 1.

GuanLuo:
There are some implementation details that I need to fix for performance, but the PR is ready for review as that shouldn't affect the overall structure

src/core/api.proto
```diff
@@ -223,6 +223,26 @@ message InferRequestHeader
   //@@     request.
   //@@
   repeated Output output = 3;
+
+  //@@  .. cpp:var:: uint32 priority
+  //@@
+  //@@     The priority value of this request. If priority handling is not
+  //@@     enable for the model, then this value is ignored. The default value
+  //@@     is 0 which indicates that the request will not be prioritized.
```
deadeyegoodwin: ".. the request will be assigned the default priority associated with the model."

src/core/api.proto
```diff
@@ -223,6 +223,26 @@ message InferRequestHeader
...
+
+  //@@  .. cpp:var:: uint64 timeout_microseconds
+  //@@
+  //@@     The timeout for this request. This value overrides the timeout
+  //@@     specified by the model, if the model allows timeout override and if
+  //@@     the value is less than the default timeout specified by the model.
+  //@@     If the request cannot be processed within this timeout, the request
+  //@@     will be handled based on the model's timeout policy.
+  //@@     The default value is 0 which indicates that the request does not
+  //@@     override timeout value.
```
deadeyegoodwin: "... does not override the model's timeout value."

src/core/model_config.proto
```diff
@@ -623,6 +623,69 @@ message ModelOptimizationPolicy
...
+  //@@
+  //@@     The default timeout for every request, in microseconds.
+  //@@     The default value is 0 which indicates that no timeout is set.
+  //@@
+  uint64 default_timeout_microseconds = 2;
+
+  //@@
+  //@@  .. cpp:var:: bool allow_timeout_override
+  //@@
+  //@@     Whether individual request can override the default timeout value.
```
deadeyegoodwin: Add sentence after the first. "When true, individual requests can set a timeout that is less than the default timeout value but may not increase the timeout."

src/core/model_config.proto
```diff
@@ -661,6 +724,39 @@ message ModelDynamicBatching
...
+  //@@     priority handling and requests that specify priority levels where
+  //@@     there is no specific policy given. If not specified, a policy with
+  //@@     default field values will be used.
+  //@@
+  ModelQueuePolicy default_queue_policy = 4;
+
+  //@@  .. cpp:var:: uint32 priority_levels
+  //@@
+  //@@     The number of priority levels to be enabled for the model,
+  //@@     the priority level starts from 1 and the requests with lower
```
deadeyegoodwin: ".. the priority levels start from 1.  Requests with lower priority level will be handled before requests with higher priority levels. Requests with the same priority level will be handled in the order that they are received."

src/core/model_config.proto
```diff
@@ -661,6 +724,39 @@ message ModelDynamicBatching
...
+  //@@     priority level will be prioritized than other requests.
+  //@@
+  uint32 priority_levels = 5;
+
+  //@@  .. cpp:var:: uint32 default_priority_level
+  //@@
+  //@@     The priority level used for requests that don't specify their
+  //@@     priority. The value must be in the range yielded by
```
deadeyegoodwin: ".. must be in the range [ 1, 'priority_levels' ]."

src/core/model_config.proto
```diff
@@ -661,6 +724,39 @@ message ModelDynamicBatching
...
   bool preserve_ordering = 3;
+
+  //@@  .. cpp:var:: ModelQueuePolicy default_queue_policy
+  //@@
+  //@@     The default queue policy used for requests that don't require
+  //@@     priority handling and requests that specify priority levels where
+  //@@     there is no specific policy given. If not specified, a policy with
+  //@@     default field values will be used.
+  //@@
+  ModelQueuePolicy default_queue_policy = 4;
```
deadeyegoodwin: Move this down to just before priority_queue_policy

src/core/model_config_utils.cc
```diff
@@ -503,6 +503,28 @@ ValidateModelConfig(
...
+          (config.dynamic_batching().default_priority_level() >
+           priority_levels)) {
+        return Status(
+            RequestStatusCode::INVALID_ARG,
+            "default priority level must be in range [1, " +
+                std::to_string(priority_levels) + "]");
+      }
+      for (const auto& queue_policy :
+           config.dynamic_batching().priority_queue_policy()) {
+        if (queue_policy.first == 0 || queue_policy.first > priority_levels) {
```
deadeyegoodwin: parens around sub-expressions

src/core/dynamic_batch_scheduler.h
```diff
@@ -115,9 +134,10 @@ class DynamicBatchScheduler : public Scheduler {
   std::mutex mu_;
   std::condition_variable cv_;
 
-  // Queue holding inference requests for the model represented by
-  // this scheduler.
-  std::deque<Scheduler::Payload> queue_;
+  // Map from priority level to queue holding inference requests for the model
+  // represented by this scheduler. Priority level 0 is reserved for default
```
deadeyegoodwin: "Priority level 0...". Do you mean that if priority queues are not supported by the scheduler than priority zero entry is used as the single queue?

src/core/dynamic_batch_scheduler.cc
```diff
@@ -122,6 +122,66 @@ DynamicBatchScheduler::Create(
...
+    const uint32_t priority_levels, const ModelQueuePolicyMap& queue_policy_map,
+    std::unique_ptr<Scheduler>* scheduler)
+{
+  DynamicBatchScheduler* dyna_sched = new DynamicBatchScheduler(
+      runner_id_start, runner_cnt, OnInit, OnWarmup, OnSchedule, OnPeek,
+      dynamic_batching_enabled, enforce_equal_shape_tensors, preserve_ordering,
+      preferred_batch_sizes, max_queue_delay_microseconds);
+  std::unique_ptr<DynamicBatchScheduler> sched(dyna_sched);
+
+  sched->queue_ =
```
deadeyegoodwin: Don't you need to define queue_ in the other Create function? Also, there is a lot of duplicate code here. Can you instead just have a single Create that has default values for the new fields and then when priority_levels == 0 just create queues_ appropriately.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -264,22 +336,27 @@ DynamicBatchScheduler::SchedulerThread(
...
+                 << " queued payloads, current total = " << queue_.Size();
+      } else if (queue_.Empty()) {
         wait_microseconds = default_wait_microseconds;
       } else if (dynamic_batching_enabled_) {
         // Use dynamic batching to get request payload(s) to execute.
         wait_microseconds = GetDynamicBatch(runner_id);
-        if (wait_microseconds == 0) {
+
+        // Get payloads that are rejected from searching dynamic batch.
+        rejected_payloads = queue_.ReleaseRejectedPayloads();
```
deadeyegoodwin: This is copying a vector of payloads... should instead use a shared_ptr or unique_ptr for efficiency.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -264,22 +336,27 @@ DynamicBatchScheduler::SchedulerThread(
...
+
+        // Extract batch only if there is pending batch
+        auto pending_batch_queue_cnt = queue_.PendingBatchCount();
+        if (wait_microseconds == 0 && (pending_batch_queue_cnt != 0)) {
```
deadeyegoodwin: parens

src/core/dynamic_batch_scheduler.h
GuanLuo: Yes

deadeyegoodwin:
@GuanLuo You should also rebase onto master since it has moved to 20.02 upstreams.

src/core/dynamic_batch_scheduler.h
```diff
@@ -115,9 +137,10 @@ class DynamicBatchScheduler : public Scheduler {
   std::mutex mu_;
   std::condition_variable cv_;
 
-  // Queue holding inference requests for the model represented by
-  // this scheduler.
-  std::deque<Scheduler::Payload> queue_;
+  // Map from priority level to queue holding inference requests for the model
+  // represented by this scheduler. If priority queues are not supported by the
+  // scheduler, then priority zero entry is used as the single queue?
```
deadeyegoodwin: Remove "?"

src/core/infer_request.h
```diff
@@ -223,6 +223,12 @@ class InferenceRequest {
   uint32_t BatchSize() const { return batch_size_; }
   void SetBatchSize(uint32_t b) { batch_size_ = b; }
 
+  uint32_t Priority() const { return priority_; }
+  void SetPriority(uint32_t p) { priority_ = p; }
+
+  uint64_t TimeoutMs() const { return timeout_ms_; }
```
deadeyegoodwin: TimeoutMicroseconds and change variable to timeout_microseconds_ or timeout_us_

src/core/model_config.proto
```diff
@@ -623,6 +623,71 @@ message ModelOptimizationPolicy
...
+//@@.. cpp:var:: message ModelQueuePolicy
+//@@
+//@@   Queue policy for inference requests.
+//@@
+message ModelQueuePolicy
+{
+  //@@
+  //@@  .. cpp:enum:: TimeoutAction
+  //@@
+  //@@     The action applied on timeout request.
```
deadeyegoodwin: ".. applied to timed-out requests."

src/core/model_config.proto
```diff
@@ -623,6 +623,71 @@ message ModelOptimizationPolicy
...
+    //@@       are processed. A delayed request will eventually be processed,
+    //@@       but may be delayed indefinitely due to newly arriving requests.
+    //@@
+    DELAY = 1;
+  }
+
+  //@@
+  //@@  .. cpp:var:: TimeoutAction timeout_action
+  //@@
+  //@@     The action applied on timeout request. The default action is REJECT.
```
deadeyegoodwin: "... applied to timed-out requests."

src/core/model_config.proto
```diff
@@ -661,6 +726,40 @@ message ModelDynamicBatching
   //@@     order.
   //@@
   bool preserve_ordering = 3;
+
+  //@@  .. cpp:var:: uint32 priority_levels
+  //@@
+  //@@     The number of priority levels to be enabled for the model,
+  //@@     the priority level starts from 1. Requests with lower priority level
```
deadeyegoodwin: "... start from 1 and 1 is the highest priority. Requests are handled in priority order with all priority 1 requests processed before priority 2, all priority 2 processed before priority 3, etc. Requests with the same..."

src/core/scheduler_utils.cc
```diff
@@ -97,4 +99,282 @@ CompareWithPendingShape(
...
+
+Status
+PriorityQueue::Enqueue(uint32_t priority_level, Scheduler::Payload&& payload)
+{
+  auto status = queues_[priority_level].Enqueue(std::move(payload));
+  if (status.IsOk()) {
+    size_++;
+    front_priority_level_ = std::min(front_priority_level_, priority_level);
+    // Invalidate the pending batch cursor if the enqueued item is placed
+    // in within the pending batch. At the same priority level, the payload is
```
deadeyegoodwin: Remove "in"

PR #1110, opened 2020-02-14
Add L0_batcher and L0_sequence_batcher test to jetson

CoderHam:


qa/L0_batcher/test.sh
```diff
@@ -39,64 +39,75 @@ CLIENT_LOG="./client.log"
...
-                echo "dynamic_batching { preferred_batch_size: [ 2, 6 ], max_queue_delay_microseconds: 10000000 }" >> config.pbtxt)
+for BACKEND in $BACKENDS; do
+    if [ "$BACKEND" != "custom" ]; then
+      cp -r $DATADIR/qa_model_repository/${BACKEND}_float32_float32_float32 models/. &&
+        (cd models/${BACKEND}_float32_float32_float32 && \
+            sed -i "s/^max_batch_size:.*/max_batch_size: 8/" config.pbtxt && \
+            sed -i "s/^version_policy:.*/version_policy: { specific { versions: [1] }}/" config.pbtxt && \
+            echo "dynamic_batching { preferred_batch_size: [ 2, 6 ], max_queue_delay_microseconds: 10000000 }" >> config.pbtxt)
+    else
+      cp -r ../custom_models/custom_float32_float32_float32 models/. &&
```
GuanLuo: The only different is the copy path, you can just condition on the `cp` command, the rest is the same.

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -40,20 +40,32 @@
...
 else:
-    _trials = ("custom", "savedmodel", "graphdef", "netdef", "plan", "onnx")
+    _trials = BACKENDS.split(' ')
+
 # Add ensemble to the _trials
-ENSEMBLE_PREFIXES = ["simple_", "sequence_", "fan_"]
+if ENSEMBLES:
+    ENSEMBLE_PREFIXES = ["simple_", "sequence_", "fan_"]
+else:
+    ENSEMBLE_PREFIXES = []
```
GuanLuo: I am under the impression that capitalized variables are constant or environment variables, so you should keep `ENSEMBLE_PREFIXES` as it is, and condition on whether testing ensemble model at line 74, i.e. `if ENSEMBLES: append the ensemble trials to _trails`.

qa/L0_sequence_batcher/test.sh
```diff
@@ -56,76 +62,146 @@ export CUDA_VISIBLE_DEVICES=0
...
-        $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_graphdef_nobatch_sequence_int32 \
-        $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_savedmodel_nobatch_sequence_float32 \
-        $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_onnx_nobatch_sequence_int32 \
-        $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_libtorch_nobatch_sequence_int32 ; do
-    cp -r $m models0/. && \
-        (cd models0/$(basename $m) && \
+MODELS=""
+for BACKEND in $BACKENDS; do
+  if [[ $BACKEND == "plan" ]]; then
+    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/plan_nobatch_sequence_float32"
```
GuanLuo: I think you can avoid the ifs by changing to `MODELS="$MODELS $DATADIR/qa_sequence_model_repository/$BACKEND_nobatch_sequence_float32"`, then you don't need to handle every backends.

qa/L0_sequence_batcher/test.sh
```diff
@@ -56,76 +62,146 @@ export CUDA_VISIBLE_DEVICES=0
...
+    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/onnx_nobatch_sequence_int32"
+  elif [[ $BACKEND == "libtorch" ]]; then
+    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/libtorch_nobatch_sequence_int32"
+  elif [[ $BACKEND == "custom" ]]; then
+    MODELS="$MODELS ../custom_models/custom_sequence_int32"
+  fi
+
+  if [ "$ENSEMBLES" == "1" ]; then
+    if [[ $BACKEND == "plan" ]]; then
+      MODELS="$MODELS $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_plan_nobatch_sequence_float32"
```
GuanLuo: Similar to above.

qa/L0_sequence_batcher/test.sh
```diff
@@ -56,76 +62,146 @@ export CUDA_VISIBLE_DEVICES=0
...
-        $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_onnx_sequence_int32 \
-        $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_libtorch_sequence_int32 \
-        ../custom_models/custom_sequence_int32 ; do
-    cp -r $m models1/. && \
-        (cd models1/$(basename $m) && \
+
+MODELS=""
+for BACKEND in $BACKENDS; do
+  if [[ $BACKEND == "plan" ]]; then
+    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/plan_sequence_float32"
```
GuanLuo: See below

qa/L0_sequence_batcher/test.sh
```diff
@@ -56,76 +62,146 @@ export CUDA_VISIBLE_DEVICES=0
...
+    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/onnx_sequence_int32"
+  elif [[ $BACKEND == "libtorch" ]]; then
+    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/libtorch_sequence_int32"
+  elif [[ $BACKEND == "custom" ]]; then
+    MODELS="$MODELS ../custom_models/custom_sequence_int32"
+  fi
+
+  if [ "$ENSEMBLES" == "1" ]; then
+    if [[ $BACKEND == "plan" ]]; then
+      MODELS="$MODELS $DATADIR/qa_ensemble_model_repository/qa_sequence_model_repository/*_plan_sequence_float32"
```
GuanLuo: See below

qa/L0_sequence_batcher/test.sh
```diff
@@ -56,76 +62,146 @@ export CUDA_VISIBLE_DEVICES=0
...
-        $DATADIR/qa_ensemble_model_repository/qa_variable_sequence_model_repository/*_savedmodel_sequence_float32 \
-        $DATADIR/qa_ensemble_model_repository/qa_variable_sequence_model_repository/*_onnx_sequence_int32 \
-        $DATADIR/qa_ensemble_model_repository/qa_variable_sequence_model_repository/*_libtorch_sequence_int32 ; do
-    cp -r $m modelsv/. && \
-        (cd modelsv/$(basename $m) && \
+
+MODELS=""
+for BACKEND in $BACKENDS; do
+  if [[ $BACKEND == "plan" ]]; then
+    MODELS="$MODELS $DATADIR/qa_variable_sequence_model_repository/plan_sequence_float32"
```
GuanLuo: same

qa/L0_sequence_batcher/test.sh
```diff
@@ -56,76 +62,146 @@ export CUDA_VISIBLE_DEVICES=0
...
+    MODELS="$MODELS $DATADIR/qa_variable_sequence_model_repository/onnx_sequence_int32"
+  elif [[ $BACKEND == "libtorch" ]]; then
+    MODELS="$MODELS $DATADIR/qa_variable_sequence_model_repository/libtorch_sequence_int32"
+  elif [[ $BACKEND == "custom" ]]; then
+    MODELS="$MODELS ../custom_models/custom_sequence_int32"
+  fi
+
+  if [ "$ENSEMBLES" == "1" ]; then
+    if [[ $BACKEND == "plan" ]]; then
+      MODELS="$MODELS $DATADIR/qa_ensemble_model_repository/qa_variable_sequence_model_repository/*_plan_nobatch_sequence_float32"
```
GuanLuo: same

src/clients/python/api_v1/library/crequest.cc
```diff
@@ -247,13 +247,9 @@ ServerStatusContextGetServerStatus(
...
   nic::Error err = ctx->ctx->GetServerStatus(&server_status);
   if (err.IsOk()) {
-    if (server_status.SerializeToString(&ctx->status_buf)) {
-      *status = &ctx->status_buf[0];
-      *status_len = ctx->status_buf.size();
-    } else {
-      err = nic::Error(
-          ni::RequestStatusCode::INTERNAL, "failed to parse server status");
-    }
+    ctx->status_buf = server_status.ShortDebugString();
```
GuanLuo: `ShortDebugString()` and `SerializeToString()` are not the same, the prior is stored in terms of text while the later is stored in terms of byte. Why do you have to make this change?

qa/L0_sequence_batcher/sequence_batcher_test.py
CoderHam: In this case ENSEMBLE_PREFIXES is not an environment variable. I can do what you said above but I will also have to modify all the other places ENSEMBLE_PREFIXES is used to check if ENSEMBLES is True.

qa/L0_sequence_batcher/test.sh
CoderHam: The datatypes for the backends are different. FP32 for savedmodel, int32 for netdef graphdef onnx libtorch and so on

src/clients/python/api_v1/library/crequest.cc
CoderHam: This was breaking due to an issue in protobuf on Jetson. I spoke to @deadeyegoodwin about this workaround. Since it does not break backward compatibility it should be fine.

qa/L0_sequence_batcher/sequence_batcher_test.py
GuanLuo: You don't need to check other places, if you take a closer look on how other places use `ENSEMBLE_PREFIXES`

qa/L0_sequence_batcher/test.sh
GuanLuo: True. You can still simplify it though. If you look at `get_datatype` in the Python test, only three backends are the exceptions.

CoderHam: Done

CoderHam:
Caused because libtorch was added back. Removed and re-ran CI.

PR #1111, opened 2020-02-15
Fix dynamic sequence batcher and preserve_order option

deadeyegoodwin:


src/core/dynamic_batch_scheduler.cc
```diff
@@ -94,9 +94,9 @@ DynamicBatchScheduler::Create(
     auto thread_exit = std::make_shared<std::atomic<bool>>(false);
     sched->scheduler_threads_exit_.emplace_back(thread_exit);
     sched->scheduler_threads_.emplace_back(new std::thread(
-        [dyna_sched, runner_id, nice, thread_exit, &init_state]() {
+        [dyna_sched, runner_id, c, nice, thread_exit, &init_state]() {
```
tanmayv25: Instead of providing completion_id in the constructor, wouldn't it make more sense to generate it internally inside dyna_sched (Incrementing whenever SchedulerThread is called)? It will ensure the id always starts from 0 without depending on the user of the class.

deadeyegoodwin: DynamicBatchScheduler::Create assigns the completion IDs to each thread and also allocates the completion_queues_ vector which must be large enough for those IDs. So it seems best to keep that logic together.

PR #1118, opened 2020-02-18
Refactor to replace protobuf InferRequestHeader with class InferenceRequest

deadeyegoodwin:


src/core/memory.h
```diff
@@ -0,0 +1,145 @@
...
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#pragma once
+
+#if defined(TRTIS_ENABLE_HTTP) || defined(TRTIS_ENABLE_METRICS)
+#include <event2/buffer.h>
```
GuanLuo: Shouldn't need this here

src/core/infer_request.cc
```diff
@@ -0,0 +1,456 @@
...
+
+Status
+InferenceRequest::SetInputData(
+    const std::string& name, const std::shared_ptr<Memory>& data)
+{
+  auto pr = input_map_.insert({name, data});
+  if (!pr.second) {
+    return Status(
+        RequestStatusCode::INVALID_ARG,
+        "input '" + name + "' already has data, can't overwrite");
```
GuanLuo: We may want some override method in the future to reuse a request object

src/core/trtserver.cc
```diff
@@ -389,24 +392,29 @@ class TrtServerRequestOptions {
...
 
  private:
   const std::string model_name_;
   const int64_t model_version_;
+
+#ifdef TRTIS_ENABLE_GRPC_V2
+  std::string id_str_;
+#endif  // TRTIS_ENABLE_GRPC_V2
+
   std::shared_ptr<ni::InferRequestHeader> request_header_;
```
GuanLuo: Also replace this with `InferenceRequest`? Then it becomes the server front-end's responsibility to set the options according to the `InferRequestHeader` string it received.

src/backends/tensorrt/plan_utils.cc
```diff
@@ -348,6 +348,39 @@ ValidateDimension(
   return Status::Success;
 }
 
+Status
```
tanmayv25: Can we have a template for the two ValidateDimension function? As the code looks identical.

src/core/infer_request.cc
deadeyegoodwin: Yes, this change doesn't allow reuse. That will be a new TRTSERVER API that exposes InferenceRequest with more methods to modify it in place.

src/core/trtserver.cc
deadeyegoodwin: Yes, also part of future changes. Want to replace the entire options and provider objects with a single TRTSERVER_InferenceRequest object that is used for both request and response and that can be mutated and used across multiple requests.

src/backends/tensorrt/plan_utils.cc
deadeyegoodwin: Done

PR #1121, opened 2020-02-19
DataType support in V2 output tensors

tanmayv25:


src/core/common.proto
```diff
@@ -0,0 +1,72 @@
...
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+syntax = "proto3";
+
+package nvidia.inferenceserver;
+
+//@@.. cpp:namespace:: nvidia::inferenceserver
```
deadeyegoodwin: Is this new proto being included in the whl and in the client SDK includes?

tanmayv25: It is getting picked by the script. But having the DataType nested was not exposed to model_config_pb2 module directly. That change would have required a non-backwards compatible change on our clients. The current change imports the model_config.proto in api.proto to access the DataType enum.

PR #1127, opened 2020-02-24
Support for explicit tensors for GRPC V2

tanmayv25:


src/clients/python/experimental_api_v2/examples/grpc_v2_explicit_tensor_client.py
```diff
@@ -0,0 +1,116 @@
...
+
+    request.inputs.extend([input0, input1])
+
+    # Populate the outputs in the inference request
+    output0 = grpc_service_v2_pb2.ModelInferRequest(
+    ).InferRequestedOutputTensor()
+    output0.name = "OUTPUT0"
+    output0.data_format = "explicit"
+
+    output1 = grpc_service_v2_pb2.ModelInferRequest(
```
deadeyegoodwin: This is strange formatting

src/clients/python/experimental_api_v2/examples/grpc_v2_explicit_tensor_client.py
```diff
@@ -0,0 +1,116 @@
...
+    output1.name = "OUTPUT1"
+    output1.data_format = "explicit"
+    request.outputs.extend([output0, output1])
+
+    #print(request)
+
+    response = grpc_stub.ModelInfer(request)
+    print("model infer:\n{}".format(response))
+
+    #    for i in range(16):
```
deadeyegoodwin: Do you not want to check for correctness?

src/core/trtserver.h
```diff
@@ -640,6 +640,15 @@ TRTSERVER_InferenceRequestProviderInputBatchByteSize(
     TRTSERVER_InferenceRequestProvider* request_provider, const char* name,
     uint64_t* byte_size);
 
+/// Get datatype of the specified output tensor
```
deadeyegoodwin: I don't think we need to add this. You should be able to get the datatype from the response header.

src/servers/grpc_server_v2.cc
```diff
@@ -1209,8 +1231,63 @@ InferResponseAlloc(
         *actual_memory_type_id = 0;
       }
 
-      raw_output->resize(byte_size);
-      *buffer = static_cast<void*>(&((*raw_output)[0]));
+      if (data_format == DataFormat::EXPLICIT) {
```
deadeyegoodwin: Check the updated V2 spec. There is no "data_format" anymore (I'm also updating our grpc_version_v2.proto appropriately as part of the shared-memory updates). So on the response side we should just always use the "raw_contents" field to return the response.

src/servers/grpc_server_v2.cc
```diff
@@ -1315,10 +1448,71 @@ InferGRPCToInput(
                 .c_str());
       }
 
-      // FIXMEV2 handle non-raw content types
+      // Try to read the raw contents if available
```
deadeyegoodwin: It is an error if more than one of these is specified...

src/servers/grpc_server_v2.cc
```diff
@@ -1315,17 +1361,94 @@ InferGRPCToInput(
...
+      if (io.contents().bool_contents_size() != 0) {
+        RETURN_IF_ERR(InferGRPCToInputHelper(
+            io.name(), request.model_name(), "BOOL", io.datatype(), byte_size));
+        base = (const void*)io.contents().bool_contents().data();
+        byte_size = io.contents().bool_contents_size();
+      }
+
+      if (io.contents().int_contents_size() != 0) {
+        RETURN_IF_ERR(InferGRPCToInputHelper(
+            io.name(), request.model_name(), "INT32", io.datatype(),
```
deadeyegoodwin: INT32 is used for int8, int16 and int32 and same for UINT32. protobuf doesn't have types of 16-bit and 8-bit integers. So for those types you will also need to reserialize like you do for strings.

src/servers/grpc_server_v2.cc
```diff
@@ -1315,17 +1361,94 @@ InferGRPCToInput(
...
+        byte_size = serialized->size();
+      }
     }
 
     uint64_t expected_byte_size = 0;
     RETURN_IF_ERR(TRTSERVER_InferenceRequestProviderInputBatchByteSize(
         request_provider, io.name().c_str(), &expected_byte_size));
 
-    if (byte_size != expected_byte_size) {
+    if ((expected_byte_size != 0) && (byte_size != expected_byte_size)) {
```
deadeyegoodwin: Why? Is this because for STRING expected_byte_size is returned as 0? If so then you should more explicitly skip this check for string.

src/clients/python/experimental_api_v2/examples/grpc_v2_explicit_byte_content_client.py
```diff
@@ -0,0 +1,121 @@
...
+
+    # Generate the request
+    request = grpc_service_v2_pb2.ModelInferRequest()
+    request.model_name = model_name
+    request.model_version = -1
+
+    # Populate the inputs in inference request
+    input0 = grpc_service_v2_pb2.ModelInferRequest().InferInputTensor()
+    input0.name = "INPUT0"
+    input0.datatype = "STRING"
```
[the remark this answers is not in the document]
tanmayv25: Updated

src/servers/grpc_server_v2.cc
tanmayv25: Done

src/servers/grpc_server_v2.cc
```diff
@@ -1315,17 +1361,94 @@ InferGRPCToInput(
...
+            io.name(), request.model_name(), "FP64", io.datatype(), byte_size));
+        base = (const void*)io.contents().fp64_contents().data();
+        byte_size = io.contents().fp64_contents_size() * 8;
+      }
+
+      if (io.contents().byte_contents_size() != 0) {
+        RETURN_IF_ERR(InferGRPCToInputHelper(
+            io.name(), request.model_name(), "STRING", io.datatype(),
+            byte_size));
+        std::shared_ptr<std::string> serialized(new std::string(""));
```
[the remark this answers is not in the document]
tanmayv25: Fixed

src/servers/grpc_server_v2.cc
```diff
@@ -1315,17 +1361,94 @@ InferGRPCToInput(
...
+        serialized_data_map->emplace(io.name(), serialized);
+
+        // Serialize the output tensor strings. Each string is
+        // serialized as a 4-byte length followed by the string itself
+        // with no null-terminator.
+        for (int e = 0; e < io.contents().byte_contents_size(); ++e) {
```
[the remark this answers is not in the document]
tanmayv25: Done

src/servers/grpc_server_v2.cc
tanmayv25: fixed

src/servers/grpc_server_v2.cc
```diff
@@ -1315,24 +1362,163 @@ InferGRPCToInput(
...
+        base = (const void*)io.contents().bool_contents().data();
+        byte_size = io.contents().bool_contents_size();
+      }
+
+      if (io.contents().int_contents_size() != 0) {
+        if (io.datatype().compare("INT8") == 0) {
+          RETURN_IF_ERR(InferGRPCToInputHelper(
+              io.name(), request.model_name(), "INT8", io.datatype(),
+              byte_size));
+          std::shared_ptr<std::string> serialized(new std::string());
```
deadeyegoodwin: You should use string::reserve() to size the string to what is expected for the string so that it doesn't need to realloc.

src/servers/grpc_server_v2.cc
```diff
@@ -1315,24 +1362,163 @@ InferGRPCToInput(
...
+          serialized_data_map->emplace(io.name(), serialized);
+          for (const auto& element : io.contents().int_contents()) {
+            // Assuming the system is little-endian, picking the
+            // least significant byte of 32-bit integer as a
+            // int8 element
+            serialized->append(reinterpret_cast<const char*>(&element), 1);
```
deadeyegoodwin: Instead of magic numbers you should get the size of the datatype once above and use it as needed. We should have a utility to get the size from the V2 datatypes... you might need to add that.

src/servers/grpc_server_v2.cc
```diff
@@ -1315,24 +1362,163 @@ InferGRPCToInput(
...
+              reinterpret_cast<const char*>(&len), sizeof(uint32_t));
+          if (element.size() > 0) {
+            serialized->append(element.c_str(), len);
+          }
+        }
+        base = serialized->c_str();
+        byte_size = serialized->size();
+      }
     }
+    if (!has_byte_contents) {
```
deadeyegoodwin: Insert blank line

src/servers/grpc_server_v2.cc
```diff
@@ -1315,24 +1362,163 @@ InferGRPCToInput(
...
-
-    if (byte_size != expected_byte_size) {
-      return TRTSERVER_ErrorNew(
-          TRTSERVER_ERROR_INVALID_ARG,
-          std::string(
-              "unexpected size " + std::to_string(byte_size) + " for input '" +
-              io.name() + "', expecting " + std::to_string(expected_byte_size) +
-              " for model '" + request.model_name() + "'")
-              .c_str());
+      if ((byte_size != expected_byte_size)) {
```
deadeyegoodwin: remove double parens

src/servers/grpc_server_v2.cc
deadeyegoodwin: Or for performance reasons you can just keep it like it is but put comments explaining each magic number. Like "1 /* sizeof INT8 */"

src/servers/grpc_server_v2.cc
tanmayv25: Done

src/servers/grpc_server_v2.cc
tanmayv25: I have added the utility in src/servers/common.cc

src/servers/grpc_server_v2.cc
tanmayv25: Done

src/servers/grpc_server_v2.cc
tanmayv25: Done

PR #1128, opened 2020-02-24
Supporting classification results in GRPC V2

tanmayv25:


src/servers/common.cc
```diff
@@ -158,11 +158,12 @@ SetInferenceRequestOptions(
...
   for (const auto& output : request.outputs()) {
     // FIXMEV2 parameters
-    // if (output.has_cls()) {
-    //  RETURN_IF_ERR(TRTSERVER_InferenceRequestOptionsAddClassificationOutput(
-    //      request_options, output.name().c_str(), output.cls().count()));
-    //} else
-    {
+    if (output.parameters().find("classification") !=
+        output.parameters().end()) {
+      RETURN_IF_ERR(TRTSERVER_InferenceRequestOptionsAddClassificationOutput(
```
deadeyegoodwin: Need to make sure has_int64_param is true and return an error if not

tanmayv25: Unfortunately, the one_of fields have has_* functions private. Added the validation logic exposed for the one_of fields.

PR #1130, opened 2020-02-25
New HTTP v2 server

CoderHam:
- [x] Status
- [x] Metadata
- [x] Infer (Without Shared Memory, Only Flat Json Array I/O)


build/CMakeLists.txt
```diff
@@ -81,20 +81,20 @@ if(TRTIS_ENABLE_TENSORRT AND NOT TRTIS_ENABLE_GPU)
   message(FATAL_ERROR "TRTIS_ENABLE_TENSORRT=ON requires TRTIS_ENABLE_GPU=ON")
 endif()
 
-if(TRTIS_ENABLE_HTTP_V2 AND NOT TRTIS_ENABLE_HTTP)
-  message(FATAL_ERROR "TRTIS_ENABLE_HTTP_V2=ON requires TRTIS_ENABLE_HTTP=ON")
+if(TRTIS_ENABLE_HTTP_V2 AND NOT TRTIS_ENABLE_GRPC_V2)
+  message(FATAL_ERROR "TRTIS_ENABLE_HTTP_V2=ON requires TRTIS_ENABLE_GRPC_V2=ON")
```
GuanLuo: As you said, HTTP should not depend on GRPC

src/servers/CMakeLists.txt
```diff
@@ -107,27 +107,31 @@ endif() # TRTIS_ENABLE_GRPC || TRTIS_ENABLE_GRPC_V2
...
   message(STATUS "Using libevhtp ${libevhtp_VERSION}")
 
-  if(${TRTIS_ENABLE_HTTP_V2})
-    set(
-      HTTP_ENDPOINT_SRCS
-      http_v2_server.cc
-    )
-  else()
-    set(
+  if(${TRTIS_ENABLE_GRPC})
```
GuanLuo: `TRTIS_ENABLE_HTTP`?

src/servers/CMakeLists.txt
```diff
@@ -138,6 +142,13 @@ if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
   endif() # TRTIS_ENABLE_GPU
 
   add_dependencies(http-endpoint-library proto-library)
+  if(${TRTIS_ENABLE_HTTP})
+    add_dependencies(http-endpoint-library grpc-library)
```
GuanLuo: drop this, HTTP v1 does not depend on `grpc-library`

src/servers/CMakeLists.txt
```diff
@@ -148,29 +159,27 @@ if(${TRTIS_ENABLE_HTTP} OR ${TRTIS_ENABLE_METRICS})
     $<TARGET_OBJECTS:http-endpoint-library>
     $<TARGET_OBJECTS:proto-library>
   )
-  if(${TRTIS_ENABLE_HTTP_V2})
-    set(
+  if(${TRTIS_ENABLE_HTTP})
+    list(APPEND
       HTTP_ENDPOINT_OBJECTS
-      ${HTTP_ENDPOINT_OBJECTS}
       $<TARGET_OBJECTS:grpc-library>
```
GuanLuo: Same

src/servers/main.cc
```diff
@@ -154,15 +161,15 @@ enum OptionId {
...
   OPTION_STRICT_READINESS,
-#ifdef TRTIS_ENABLE_HTTP
+  OPTION_API_VERSION,
+#if defined(TRTIS_ENABLE_HTTP) || defined(TRTIS_ENABLE_HTTP_V2)
   OPTION_ALLOW_HTTP,
   OPTION_HTTP_PORT,
   OPTION_HTTP_HEALTH_PORT,
   OPTION_HTTP_THREAD_COUNT,
-#endif  // TRTIS_ENABLE_HTTP
+  #endif  // TRTIS_ENABLE_HTTP || TRTIS_ENABLE_HTTP_V2
```
GuanLuo: remove space before `#endif`

src/servers/main.cc
```diff
@@ -1107,17 +1182,18 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
   }
 
 
-#ifdef TRTIS_ENABLE_HTTP
+#if defined(TRTIS_ENABLE_HTTP) || defined(TRTIS_ENABLE_HTTP_V2)
+  api_version_ = api_version;
```
GuanLuo: Move out of ifdef macro, the same for below

src/servers/http_server_v2.h
```diff
@@ -0,0 +1,59 @@
...
+  static TRTSERVER_Error* CreateAPIServer(
+      const std::shared_ptr<TRTSERVER_Server>& server,
+      const std::shared_ptr<nvidia::inferenceserver::TraceManager>&
+          trace_manager,
+      const std::shared_ptr<SharedMemoryBlockManager>& smb_manager,
+      const std::map<int32_t, std::vector<std::string>>& port_map,
+      const int thread_cnt,
+      std::vector<std::unique_ptr<HTTPServerV2>>* http_servers);
+
+  static TRTSERVER_Error* CreateMetricsServer(
```
deadeyegoodwin: Please keep metrics server out of this PR. It can stand alone it its own PR.

src/servers/main.cc
```diff
@@ -79,21 +82,26 @@ bool allow_model_control_ = false;
 // The HTTP, GRPC and metrics service/s and ports. Initialized to
 // default values and modifyied based on command-line args. Set to -1
 // to indicate the protocol is disabled.
+// Default to using the V1 protocol.
```
deadeyegoodwin: You put this between the comment above that goes with the code below. Move this up above the comment.

src/servers/main.cc
```diff
@@ -491,6 +499,39 @@ StartHttpService(
 }
 #endif  // TRTIS_ENABLE_HTTP
 
+#ifdef TRTIS_ENABLE_HTTP_V2
+TRTSERVER_Error*
+StartHttpService(
```
deadeyegoodwin: StartHttpServiceV2

src/servers/main.cc
```diff
@@ -1106,21 +1183,21 @@ Parse(TRTSERVER_ServerOptions* server_options, int argc, char** argv)
...
   }
 
+  api_version_ = api_version;
 
-#ifdef TRTIS_ENABLE_HTTP
+#if defined(TRTIS_ENABLE_HTTP) || defined(TRTIS_ENABLE_HTTP_V2)
   http_port_ = http_port;
   http_health_port_ = http_health_port;
+  http_thread_cnt_ = http_thread_cnt;
 #ifdef TRTIS_ENABLE_HTTP_V2
```
deadeyegoodwin: The ports you use depends on api_version, not how built.

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1566 @@
...
+      const int thread_cnt)
+      : HTTPServerV2Impl(port, thread_cnt), server_(server),
+        trace_manager_(trace_manager), shm_manager_(shm_manager),
+        allocator_(nullptr), api_regex_(R"(/v2/(health|models)(.*))"),
+        health_regex_(R"(/(live|ready))"),
+        model_regex_(R"(/([^/]+)(?:/version/([0-9]+))?(/infer|/ready)?)")
+  {
+    TRTSERVER_Error* err = TRTSERVER_ServerId(server_.get(), &server_id_);
+    if (err != nullptr) {
+      server_id_ = "unknown:0";
```
deadeyegoodwin: Does this match what we do on the GRPC side if ServerId call fails?

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1566 @@
...
+  {
+    LOG_TRTSERVER_ERROR(
+        TRTSERVER_ResponseAllocatorDelete(allocator_),
+        "deleting response allocator");
+  }
+
+  //
+  // AllocPayload
+  //
+  // Simple structure that carries the userp payload needed for
```
deadeyegoodwin: AllocPayload and InferResponseAlloc should be move to a common set of classes/functions that can be used by both HTTP and GRPC servers. They are mostly identical except for the little code needed for difference between protobuf message and evbuffer. You should break that out into a virtual function or similar so that most of this code can be shared. Please work with @tanmayv25 on this. It can be done as a separate PR to this one.

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1566 @@
...
+      format = "text";
+    }
+
+    // The description of the raw outputs needs to go in the
+    // kInferResponseHTTPHeader since it is needed to interpret the
+    // body. The entire response (including classifications) is
+    // serialized at the end of the body.
+    response_header.set_id(request_id_);
+
+    std::string rstr;
```
deadeyegoodwin: There are no query parameters in the new API, and this seems to be returning protobuf response header?

src/servers/http_server_v2.cc
CoderHam: Yes it does. As seem [here](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/servers/grpc_server_v2.cc#L2610).

src/servers/http_server_v2.cc
CoderHam: Will do

deadeyegoodwin: Did this get filed?

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1539 @@
...
+      // allocate on GPU memory then force allocation on CPU instead.
+      if (*actual_memory_type != TRTSERVER_MEMORY_CPU) {
+        LOG_VERBOSE(1) << "HTTP: unable to provide '" << tensor_name << "' in "
+                       << MemoryTypeString(*actual_memory_type) << ", will use "
+                       << MemoryTypeString(TRTSERVER_MEMORY_CPU);
+        *actual_memory_type = TRTSERVER_MEMORY_CPU;
+        *actual_memory_type_id = 0;
+      }
+
+#if 0
```
deadeyegoodwin: How can this be ifdeffed out? Did you start from the current version of http_server.cc? This code seems old.

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1539 @@
...
+    if (element_cnt == 0) {
+      element_cnt = shape[i].GetInt();
+    } else {
+      element_cnt *= shape[i].GetInt();
+    }
+  }
+
+  // If flat
+  if (!tensor_data[0].IsArray()) {
+    // BOOL vector is broken
```
deadeyegoodwin: Why? Also, within Triton bool is an uint8 with value 0 or 1 so you should create a uint8_t vector and initialize it that way.

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1539 @@
...
+}
+
+TRTSERVER_Error*
+HTTPAPIServerV2::EVBufferToInput(
+    const std::string& model_name, const InferRequestHeader& request_header,
+    evbuffer* input_buffer,
+    TRTSERVER_InferenceRequestProvider* request_provider,
+    AllocPayload* alloc_payload)
+{
+  // Extract individual input data from HTTP body and register in
```
deadeyegoodwin: Why commented out? Again you likely want to base this file off http_server.cc as that is the up-to-date code.

src/servers/http_server_v2.cc
CoderHam: The support for shared memory is not added yet. It was ifdeffed out similar to the grpc_server_v2.cc [here](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/servers/grpc_server_v2.cc#L1342)

src/servers/http_server_v2.cc
CoderHam: Fixed.

src/servers/http_server_v2.cc
CoderHam: Just filed it

src/servers/http_server_v2.cc
deadeyegoodwin: Isn't it inside the !use_shm conditional? But I think this code is old relative to some changes @GuanLuo made.

GuanLuo: I think @CoderHam `#if 0` the wrong place.. The code here is for the case that does NOT use shared memory. The code here seems to be up to date actually, each ResponseAlloc invocation creates and reserves space on separate evhttp buffer while previously it uses the one buffer for all.

deadeyegoodwin:
@CoderHam This needs to be manually rebased.

src/servers/http_server_v2.cc
CoderHam: Fixed

src/servers/http_server_v2.cc
```diff
@@ -0,0 +1,1517 @@
...
+        "deleting response allocator");
+  }
+
+  //
+  // AllocPayload
+  //
+  // Simple structure that carries the userp payload needed for
+  // allocation. These are just pointers into a HandlerState
+  // object. HandlerState lifetime is always longer than what is
+  // required for allocation callback so HandlerState manages the
```
GuanLuo: What is `HandlerState`?

CoderHam: Cleaned up

PR #1136, opened 2020-02-26
Add command-line option to override minimum supported CUDA compute ca…

deadeyegoodwin:
…pability


deadeyegoodwin:
@GuanLuo I could be, but in general I don't want to introduce global state when not necessary.

PR #1139, opened 2020-02-27
Fix default onnx optimization level and make it configurable

CoderHam:


src/backends/onnx/onnx_backend.cc
```diff
@@ -87,9 +87,20 @@ OnnxBackend::CreateExecutionContexts(
...
   RETURN_IF_ORT_ERROR(ort_api->SetIntraOpNumThreads(session_options, 1));
-  // disable graph optimization
+
+  // set graph optimization level
+  GraphOptimizationLevel optimization_level =
+      GraphOptimizationLevel::ORT_ENABLE_ALL;
+  if (Config().optimization().has_graph()) {
+    int graph_level = Config().optimization().graph().level();
+    if (graph_level == -1) {
+      optimization_level = GraphOptimizationLevel::ORT_ENABLE_BASIC;
```
GuanLuo: -1 should be disable according to `model_config.proto`

CoderHam: When I discussed with David he suggested to enable just the basic ones so that all 'basic' ops are run on GPU at least.
If the user runs with DISABLE_ALL some ops will be run on CPU and cause an unwanted slowdown

PR #1143, opened 2020-02-28
Removing shared memory handling from core

tanmayv25:


CoderHam:
You need to rebase this onto master

src/servers/grpc_server.cc
```diff
@@ -1729,20 +1722,15 @@ SharedMemoryControlHandler::Process(Handler::State* state, bool rpc_ok)
...
-            &smb, request.register_().name(),
+        RETURN_IF_ERR(shm_manager_->AddSystemSharedMemory(
+            request.register_().name(),
             request.register_().system_shared_memory().shared_memory_key(),
             request.register_().system_shared_memory().offset(),
-            request.register_().byte_size());
-        if (err == nullptr) {
-          err = TRTSERVER_ServerRegisterSharedMemory(trtserver_.get(), smb);
-        }
+            request.register_().byte_size()));
```
GuanLuo: Just notice that why there is underscore for `register_()`?

src/servers/shared_memory_manager.h
```diff
@@ -0,0 +1,160 @@
...
+  /// (CPU) memory to the manager. Return TRTSERVER_ERROR_ALREADY_EXISTS
+  /// if a shared memory block of the same name already exists in the manager.
+  /// \param name The name of the memory block.
+  /// \param shm_key The name of the posix shared memory object
+  /// containing the block of memory.
+  /// \param offset The offset within the shared memory object to the
+  /// start of the block.
+  /// \param byte_size The size, in bytes of the block.
+  /// \return a TRTSERVER_Error indicating success or failure.
+  TRTSERVER_Error* AddSystemSharedMemory(
```
GuanLuo: How about changing to register / unregister terminology? The GRPC service is defined that way.

src/servers/shared_memory_manager.cc
```diff
@@ -67,39 +61,41 @@ MapSharedMemory(
...
-  return Status::Success;
+  return nullptr;
 }
 
-Status
+TRTSERVER_Error*
 UnmapSharedMemory(void* mapped_addr, size_t byte_size)
 {
   int tmp_fd = munmap(mapped_addr, byte_size);
   if (tmp_fd == -1) {
```
GuanLuo: Change the `tmp`s to more descriptive name.. I don't think unmap call will return a fd...

src/servers/grpc_server.cc
tanmayv25: It seems to be a protoc behavior. I am not sure why does it adds an under-score after the field name.

CoderHam: register is probably a keyword or reserved. It therefore adds '_' to the end of the variable name

src/servers/shared_memory_manager.cc
tanmayv25: Done

src/servers/shared_memory_manager.h
tanmayv25: Done

src/servers/shared_memory_manager.cc
CoderHam: You're right, the returned value is an indicator is success (0 is success, -1 for failure)

src/backends/onnx/loader.cc
```diff
@@ -26,9 +26,9 @@
 
 #include "src/backends/onnx/loader.h"
 
+#include <src/core/logging.h>
```
deadeyegoodwin: Looks like you need to run the formatter

src/core/server_status.proto
```diff
@@ -97,21 +97,6 @@ message ModelControlRequestStats
   StatDuration success = 1;
 }
 
-//@@
-//@@.. cpp:var:: message SharedMemoryControlRequestStats
-//@@
-//@@   Statistics collected for SharedMemoryControl requests.
-//@@
-message SharedMemoryControlRequestStats
```
deadeyegoodwin: I understand your earlier question  now. To avoid breaking existing code just mark this as deprecated with message indicating that it will always report 0.

src/core/server_status.proto
```diff
@@ -463,12 +448,6 @@ message ServerStatus
   //@@
   ModelControlRequestStats model_control_stats = 9;
 
-  //@@  .. cpp:var:: SharedMemoryControlRequestStats shm_control_stats
-  //@@
-  //@@     Statistics for SharedMemoryControl requests.
-  //@@
-  SharedMemoryControlRequestStats shm_control_stats = 10;
```
deadeyegoodwin: To avoid breaking existing code just mark this as deprecated.

src/core/server_status.proto
```diff
@@ -488,7 +467,7 @@ message SharedMemoryStatus
   //@@
   //@@     The list of active/registered shared memory regions.
   //@@
-  repeated SharedMemoryRegion shared_memory_region = 2;
+  repeated SharedMemoryRegion shared_memory_region = 1;
```
deadeyegoodwin: Revert

src/servers/CMakeLists.txt
```diff
@@ -248,6 +248,7 @@ target_link_libraries(
   ${HTTP_ENDPOINT_LIBRARIES}
   ${GRPC_ENDPOINT_LIBRARIES}
   ${TRACING_LIBRARIES}
+  PRIVATE ${CUDA_LIBRARIES}
```
deadeyegoodwin: Only when TRTIS_ENABLE_GPU

src/backends/onnx/loader.cc
tanmayv25: Actually, the formatter is moving the include up for some reason. The another PR fixed it though.

src/core/server_status.proto
tanmayv25: Done

src/core/server_status.proto
tanmayv25: Done

src/core/server_status.proto
tanmayv25: Fixed.

src/servers/CMakeLists.txt
tanmayv25: Fixed

src/backends/onnx/loader.cc
GuanLuo: I guess that's the `<>`?

tanmayv25: Ahh. I see..

PR #1145, opened 2020-02-28
Add TRTServer options for setting priority and request timeout

GuanLuo:
This PR is branch-off from `gluo-timeout` and targeted to `gluo-timeout` to show the change for adding TRTServer options on top of the changes in `gluo-timeout`

### DIFF
--- a/src/core/ensemble_scheduler.cc
+++ b/src/core/ensemble_scheduler.cc
@@ -167,6 +167,7 @@ class EnsembleContext {
   uint32_t flags_;
   uint64_t correlation_id_;
   uint32_t batch_size_;
+  uint32_t priority_;
 
   // Objects related to the ensemble infer request
   Status ensemble_status_;
@@ -272,6 +273,7 @@ EnsembleContext::EnsembleContext(
     batch_size_ = irequest->BatchSize();
     correlation_id_ = irequest->CorrelationId();
     flags_ = irequest->Flags();
+    priority_ = irequest->Priority();
 
     for (const auto& pr : irequest->Inputs()) {
       const auto& input = pr.second;
@@ -555,6 +557,9 @@ EnsembleContext::InitStep(size_t step_idx, std::shared_ptr<Step>* step)
   irequest->SetCorrelationId(correlation_id_);
   irequest->SetFlags(flags_);
   irequest->SetBatchSize((batch_size == 0 ? 1 : batch_size));
+  irequest->SetPriority(priority_);
+  // FIXME: need to think about how to interpret timeout in ensemble.
+  irequest->SetTimeoutMs(0);
   RETURN_IF_ERROR(irequest->Normalize(*backend));
 
   step->reset(new Step(step_idx));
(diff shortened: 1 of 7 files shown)


src/core/trtserver.cc
```diff
@@ -484,6 +486,22 @@ TrtServerRequestOptions::SetBatchSize(uint64_t batch_size)
...
+  std::lock_guard<std::mutex> lk(mtx_);
+  request_header_->set_priority(priority);
+  return nullptr;  // Success
+}
+
+TRTSERVER_Error*
+TrtServerRequestOptions::SetTimeoutMs(uint64_t timeout_ms)
+{
+  std::lock_guard<std::mutex> lk(mtx_);
+  request_header_->set_timeout_microseconds(timeout_ms);
```
nieksand: With `ms` I'd assume milliseconds.  Consider using `us` if it's really microseconds.

src/core/ensemble_scheduler.cc
```diff
@@ -555,6 +557,9 @@ EnsembleContext::InitStep(size_t step_idx, std::shared_ptr<Step>* step)
   irequest->SetCorrelationId(correlation_id_);
   irequest->SetFlags(flags_);
   irequest->SetBatchSize((batch_size == 0 ? 1 : batch_size));
+  irequest->SetPriority(priority_);
+  // FIXME: need to think about how to interpret timeout in ensemble.
```
deadeyegoodwin: I would say "Request for ensemble model cannot override the timeout values for the composing models."
To support an overall ensemble timeout we would have to add a timeout to the ModelEnsembling section of the model config and then check that timeout between each pass of the ensemble. Can you file a ticket for that.

src/core/trtserver.cc
```diff
@@ -484,6 +486,22 @@ TrtServerRequestOptions::SetBatchSize(uint64_t batch_size)
...
+TRTSERVER_Error*
+TrtServerRequestOptions::SetPriority(uint32_t priority)
+{
+  std::lock_guard<std::mutex> lk(mtx_);
+  request_header_->set_priority(priority);
+  return nullptr;  // Success
+}
+
+TRTSERVER_Error*
+TrtServerRequestOptions::SetTimeoutMs(uint64_t timeout_ms)
```
deadeyegoodwin: SetTimeoutMicroseconds

src/core/trtserver.cc
```diff
@@ -980,6 +998,26 @@ TRTSERVER_InferenceRequestOptionsSetBatchSize(
...
+    TRTSERVER_InferenceRequestOptions* request_options, uint32_t priority)
+{
+  TrtServerRequestOptions* loptions =
+      reinterpret_cast<TrtServerRequestOptions*>(request_options);
+  loptions->SetPriority(priority);
+  return nullptr;  // Success
+}
+
+TRTSERVER_Error*
+TRTSERVER_InferenceRequestOptionsSetTimeout(
```
deadeyegoodwin: SetTimeoutMicroseconds

src/core/ensemble_scheduler.cc
GuanLuo: The comment is added to #1109 where the request timeout is defined (`api.proto` )

PR #1146, opened 2020-02-28
Additional error logging for CUDA calls in perf_client

tanmayv25:


src/backends/onnx/loader.cc
```diff
@@ -28,10 +28,10 @@
 
 #include <future>
 #include <thread>
-#include "src/core/logging.h"
```
CoderHam: Remove from this PR, already fixed in #1139

src/clients/c++/perf_client/load_manager.cc
```diff
@@ -333,7 +333,8 @@ LoadManager::InitSharedMemory()
         return nic::Error(
             ni::RequestStatusCode::INTERNAL,
             "unable to allocate memory of " + std::to_string(alloc_size) +
-                "bytes on gpu for output " + output->Name());
```
CoderHam: Also add after line 101

tanmayv25: Done

PR #1152, opened 2020-03-03
Remove use of legacy TensorRT API

deadeyegoodwin:


qa/common/gen_qa_dyna_sequence_models.py
```diff
@@ -611,35 +611,36 @@ def create_plan_shape_tensor_modelfile(models_dir, model_version, max_batch, dty
     with open(model_version_dir + "/model.plan", "wb") as f:
         f.write(engine.serialize())
 
-    engine.destroy()
-    builder.destroy()
-    
+    del engine
```
tanmayv25: I remember reading somewhere in TRT docs that they advice using the builder,engine and network inside the with.. as.. construct. In case there is an exception above the engine and builder might not be successfully cleared.

deadeyegoodwin: Yes, and I don't think I need the del's as the objects will be deleted when the scope ends... but I did it for clarity. I didn't use "with" because it would have required a lot more code changing and error-prone indenting... and given what the script does any exceptions will cause abort and we don't care if things are cleaned up or not. But, if you think it could be a problem I can fix.

tanmayv25: Have a look at "Create and Destroy Functions" [here](https://docs.nvidia.com/deeplearning/sdk/tensorrt-api/python_api/migrationGuide.html).

tanmayv25: Yes, indentation change is an issue. Let's leave it as it is for now. Mostly these scripts are stable.

PR #1163, opened 2020-03-04
Add gen script for TensorRT models on JetPack 4.4

CoderHam:
This must be done since TRT models must be generated again on Jetson and stored in the **\<version\>_jetson** directory.

### DIFF
--- a/qa/L0_sequence_batcher/test.sh
+++ b/qa/L0_sequence_batcher/test.sh
@@ -41,6 +41,7 @@ BATCHER_TEST=sequence_batcher_test.py
 DATADIR=${DATADIR:="/data/inferenceserver/${REPO_VERSION}"}
 OPTDIR=${OPTDIR:="/opt"}
 SERVER=${OPTDIR}/tensorrtserver/bin/trtserver
+
 source ../common/util.sh
 
 RET=0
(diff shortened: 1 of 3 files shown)


qa/L0_batcher/test.sh
```diff
@@ -42,6 +42,7 @@ VERIFY_TIMESTAMPS=verify_timestamps.py
 DATADIR=${DATADIR:="/data/inferenceserver/${REPO_VERSION}"}
 OPTDIR=${OPTDIR:="/opt"}
 SERVER=${OPTDIR}/tensorrtserver/bin/trtserver
+JETSON=${JETSON:="0"}
```
deadeyegoodwin: Can you make this more generic. Perhaps at line 37 add something like:
if [ -n "$REPO_ARCH" ]; then
  REPO_VERSION=${REPO_VERSION}_${REPO_ARCH}
fi

And then you shouldn't need any other changes. That same change can be made to all test.sh files.

CoderHam: This needs to be done for only TRT and non-ensembles. I will still have to check in the future. But yes I could make it more generic by using REPO_ARCH instead

PR #1165, opened 2020-03-05
Changing data type of model_version to string in gRPC V2

tanmayv25:


src/servers/grpc_server_v2.cc
```diff
@@ -760,10 +760,12 @@ ModelReadyHandler::Process(Handler::State* state, bool rpc_ok)
...
         const ModelStatus& model_status = nitr->second;
 
-        // If requested version is -1 then find the highest valued
+        int64_t requested_version = -1;
+        // If request doesn't specify the version then find the highest valued
         // version.
-        int64_t requested_version = request.version();
-        if (requested_version == -1) {
+        if (!request.version().empty()) {
+          requested_version = std::stol(request.version());
```
deadeyegoodwin: stol will throw and exception if no conversion. So you should probably create a utility function to do the conversion and catch the error and return an appropriate error that you can assign to "err". All stol's below should use the utility function.

tanmayv25: Done

PR #1169, opened 2020-03-05
Close shared memory descriptor after mmap is done

CoderHam:


src/servers/shared_memory_manager.cc
```diff
@@ -172,6 +172,16 @@ SharedMemoryManager::RegisterSystemSharedMemory(
             .c_str());
   }
 
+  err = CloseSharedMemoryRegion(shm_fd);
```
deadeyegoodwin: What if MapSharedMemory above fails... then shm_fd is not closed.

CoderHam: Done

Issue #1173, opened 2020-03-09
S3 file system should support remote minio server as well


24hours:
**Is your feature request related to a problem? Please describe.**
I have setup a remote minio server at my organization, I tried to execute server with 
trtserver --model-repository s3://10.0.0.0:9000/model-bucket 
it just crash with "Could not find bucket with name 10.0.0.0"

**Describe the solution you'd like**
trtis should support s3 server with remote address

**Describe alternatives you've considered**
n/a

I find that the feature can be added by modifying 

https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/core/filesystem.cc#L608

s3://([0-9a-z-.]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-z.-_]+)*)?)")

CoderHam:
@24hours Sure, I will make these changes and test support for remote minio s3 buckets as well.

PR #1175, opened 2020-03-09
Fix regex for remote minio server

CoderHam:
Solve for #1173
Now supporting remote minio servers as well

### DIFF
--- a/qa/L0_s3_local/test.sh
+++ b/qa/L0_s3_local/test.sh
@@ -90,6 +90,8 @@ if [ "$SERVER_PID" == "0" ]; then
     exit 1
 fi
 
+sleep 5 
+
 set +e
 for MODEL_NAME in resnet_v1_50_graphdef resnet_v1_50_savedmodel; do
   $PERF_CLIENT -m $MODEL_NAME -p 3000 -t 1 >$CLIENT_LOG 2>&1
@@ -117,6 +119,8 @@ if [ "$SERVER_PID" == "0" ]; then
     wait $MINIO_PID
     exit 1
 fi
+
+sleep 5 
 
 set +e
 for MODEL_NAME in resnet_v1_50_graphdef resnet_v1_50_savedmodel; do
--- a/qa/L0_s3_local/test.sh
+++ b/qa/L0_s3_local/test.sh
@@ -43,7 +43,7 @@ PERF_CLIENT=../clients/perf_client
 DATADIR="/data/inferenceserver/${REPO_VERSION}/tf_model_store"
 
 SERVER=/opt/tensorrtserver/bin/trtserver
-SERVER_ARGS="--log-verbose=1 --model-repository=s3://localhost:4572/demo-bucket"
+SERVER_ARGS="--log-verbose=1 --model-repository=s3://127.0.0.1:4572/demo-bucket"
 SERVER_LOG="./inference_server.log"
 source ../common/util.sh
 
--- a/qa/L0_s3_local/test.sh
+++ b/qa/L0_s3_local/test.sh
@@ -43,8 +43,6 @@ PERF_CLIENT=../clients/perf_client
 DATADIR="/data/inferenceserver/${REPO_VERSION}/tf_model_store"
 
 SERVER=/opt/tensorrtserver/bin/trtserver
-SERVER_ARGS="--log-verbose=1 --model-repository=s3://127.0.0.1:4572/demo-bucket"
-SERVER_LOG="./inference_server.log"
 source ../common/util.sh
 
 rm -f *.log*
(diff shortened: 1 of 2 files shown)


src/core/filesystem.cc
```diff
@@ -605,7 +605,7 @@ S3FileSystem::ParsePath(
 {
   // Get the bucket name and the object path. Return error if input is malformed
   re2::RE2 s3_regex(
-      "s3://([a-z]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-z.-_]+)*)?)");
+      "s3://([0-9a-z-.]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z.-_]+)*)?)");
```
GuanLuo: Should we include capital characters for all parts that allows characters?

src/core/filesystem.cc
```diff
@@ -1042,7 +1042,7 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
         "-DTRTIS_ENABLE_S3=ON.");
 #else
     re2::RE2 s3_regex(
```
GuanLuo: You can make the regex to be a class member, seems to be identical as the one above.

qa/L0_s3_local/test.sh
```diff
@@ -43,7 +43,7 @@ PERF_CLIENT=../clients/perf_client
 DATADIR="/data/inferenceserver/${REPO_VERSION}/tf_model_store"
 
 SERVER=/opt/tensorrtserver/bin/trtserver
-SERVER_ARGS="--log-verbose=1 --model-repository=s3://localhost:4572/demo-bucket"
+SERVER_ARGS="--log-verbose=1 --model-repository=s3://127.0.0.1:4572/demo-bucket"
```
GuanLuo: Should just test both 127.0.0.1 (ip address) and localhost (domain name)

src/core/filesystem.cc
CoderHam: Bucket names should not contain upper-case letters. 
From what I understand the convention is to keep the hostname in lower case as well. I could support capital letters in the place of a hostname.

src/core/filesystem.cc
CoderHam: Done

src/core/filesystem.cc
deadeyegoodwin: S3 doesn't allow bucket names to have capitals? Or just a convention? We need to allow whatever S3 supports.
You should allow capitals in the hostname. Name resolution is case-insensitive but caps are allowed.

CoderHam: Already fixed to allow caps in hostname. The s3 bucket name can contain only lower-case characters, numbers, periods, and dashes. Its is a rule/requirement and not just a convention.

src/core/filesystem.cc
```diff
@@ -1042,7 +1044,7 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
...
-        "s3://([a-z]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z.-_]+)*)?)");
+        "s3://([0-9a-zA-Z-.]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z.-_]+)*)?)");
     std::string host_name, host_port, bucket, object;
     if (!RE2::FullMatch(
             path, s3_regex, &host_name, &host_port, &bucket, &object)) {
```
deadeyegoodwin: Why is CheckClient below commented out?

src/core/filesystem.cc
```diff
@@ -1042,7 +1044,7 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
         "-DTRTIS_ENABLE_S3=ON.");
 #else
     re2::RE2 s3_regex(
-        "s3://([a-z]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z.-_]+)*)?)");
+        "s3://([0-9a-zA-Z-.]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z.-_]+)*)?)");
```
deadeyegoodwin: You still have this s3 regex repeated here, when you should be able to have it only in the s3 filesystem object. The following logic should be moved into the S3Filesystem constructor.

src/core/filesystem.cc
```diff
@@ -1086,7 +1088,8 @@ JoinPath(std::initializer_list<std::string> segments)
       // Check if S3 path (s3://<bucket> or s3://<host>:<port>/<bucket>)
       if (!seg.empty() && !seg.rfind("s3://", 0)) {
         re2::RE2 s3_regex(
```
deadeyegoodwin: Why do we need all this extra logic for bucket? Segments are anything that is just joined by a '/'.

src/core/filesystem.cc
CoderHam: This is residual code from the initial s3 work. The CheckClient was not implemented for s3, only for GCS.

src/core/filesystem.cc
CoderHam: The s3 filesystem object is created once we know the host and port number since the configuration is done inside the constructor. 
Are you suggesting I pull this outside the constructor into a different function?

src/core/filesystem.cc
CoderHam: The path in case of minio s3 will contain the s3://host:port/bucket_name/rest_of_the_path
This code extracts the host and port number from the path and returns the true path in the s3 bucket.

deadeyegoodwin: Why is JoinPath doing all that. It should just be joining segments. If one of the segments is S3:// whatever then why do we need to do anything different. Is there some caller of JoinPath that expects this behavior?

src/core/filesystem.cc
deadeyegoodwin: I'm suggesting we don't want to have to create an s3_regex outside of the class. We know this is an s3 path so just pass the path to the constructor and let it do what it wants (find host and port, etc.).

src/core/filesystem.cc
CoderHam: Yes, one of the callers of JoinPath expects the s3 path i.e. the Tensorflow savedmodel and graphdef model functions. They expect the s3 path without the `host:port`

src/core/filesystem.cc
CoderHam: Done

src/core/filesystem.cc
CoderHam: Removed

src/core/filesystem.cc
deadeyegoodwin: Then can we have a separate filesystem function (like "Localize") that they can use when necessary to strip that part of the path. And put it in the filesystem objects so that the S3 one can use the common s3_regex. Let's not overload JoinPath to do additional work.

CoderHam: Fixed with a similar work around directly in DownloadFileFolder

qa/L0_s3_local/test.sh
```diff
@@ -118,6 +120,8 @@ if [ "$SERVER_PID" == "0" ]; then
     exit 1
 fi
 
+sleep 5
```
GuanLuo: Why the server reports ready while it shouldn't?

CoderHam: Filled a ticket to re-visit this issue.

PR #1179, opened 2020-03-11
System Shared Memory support for GRPC V2

tanmayv25:
The following enhancements are included in this PR:
1. New handler for handling multiple RPCs in grpc_server_v2
2. System and CUDA shared memory support in server
3. Client side support for system shared memory
4. Simple example using system shared memory

I plan to make the following changes as future set of PRs:
1. Client side support for cuda shared memory
2. Simple example using cuda shared memory
3. L0_shared_memory_v2 and L0_cuda_shared_memory_v2 using the new clients.
4. Package the shared_memory and cuda_shared_memory modules in separate packages outside tritongrpcclient and tritonhttpclient.

### DIFF
--- a/src/core/server.cc
+++ b/src/core/server.cc
@@ -83,6 +83,7 @@ InferenceServer::InferenceServer()
 
   id_ = "inference:0";
   protocol_version_ = 1;
+  extensions_.push_back("classification");
   extensions_.push_back("model_configuration");
   extensions_.push_back("system_shared_memory");
   extensions_.push_back("cuda_shared_memory");
--- a/src/core/server.cc
+++ b/src/core/server.cc
@@ -83,13 +83,9 @@ InferenceServer::InferenceServer()
 
   id_ = "inference:0";
   protocol_version_ = 1;
-  extensions_.push_back("classification");
-  extensions_.push_back("flat_tensor_format");
   extensions_.push_back("model_configuration");
-  extensions_.push_back("model_control");
-  extensions_.push_back("sequence");
-  extensions_.push_back("shared_memory_tensor_format");
-  extensions_.push_back("statistics");
+  extensions_.push_back("system_shared_memory");
+  extensions_.push_back("cuda_shared_memory");
 
   strict_model_config_ = true;
   strict_readiness_ = true;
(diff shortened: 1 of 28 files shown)


src/core/server.cc
```diff
@@ -83,13 +83,9 @@ InferenceServer::InferenceServer()
 
   id_ = "inference:0";
   protocol_version_ = 1;
-  extensions_.push_back("classification");
```
deadeyegoodwin: Classification and model-control are implemented, right? So those extensions should be added as well.

src/servers/grpc_server_v2.cc
```diff
@@ -1246,6 +1635,94 @@ InferResponseRelease(
...
 
+template <typename TensorType>
+TRTSERVER_Error*
+ParseSharedMemoryParams(
+    const TensorType& tensor, bool* has_shared_memory, std::string* region_name,
+    int64_t* offset, size_t* byte_size)
+{
+  *has_shared_memory = false;
+  *offset = 0 /* default value */;
+  if (tensor.parameters().find("shared_memory_region") !=
```
deadeyegoodwin: For perf reasons you should find each parameter only once and use the iterator thereafter (instead of using at() which causes another lookup in the map).

src/servers/grpc_server_v2.cc
```diff
@@ -1336,20 +1818,21 @@ InferGRPCToInput(
     const void* base;
     size_t byte_size;
     TRTSERVER_Memory_Type memory_type = TRTSERVER_MEMORY_CPU;
-    int64_t memory_type_id = 0;
+    int memory_type_id = 0;
```
deadeyegoodwin: In the TRTSERVER API type_id is an int64_t

src/servers/grpc_server_v2.cc
```diff
@@ -514,6 +515,394 @@ Handler<ServiceType, ServerResponderType, RequestType, ResponseType>::Stop()
   LOG_VERBOSE(1) << "Thread exited for " << Name();
 }
 
+template <typename ResponderType, typename RequestType, typename ResponseType>
```
deadeyegoodwin: This seems like much more code than the existing solution. It does use a single thread but it seems like there is likely something less complex? No need to change now but we should revisit when we put all the other low-overhead APIs onto this single thread.

PR #1180, opened 2020-03-11
Add CUDA memory manger to avoid raw cudaMalloc / cudaFree

GuanLuo:
@deadeyegoodwin Unfortunately the management library that I am using doesn't support growing the buffer until threshold, we can only preallocate enough CUDA memory on initialization. The other option is the [caching allocator](https://github.com/NVlabs/cub/blob/1.8.0/cub/util_allocator.cuh#L59) where it will allocate new memory if there is no appropriate cached bin, but we can't control the behavior when max_cached_size is reach (i.e. it `cudaFree()` the allocation if max_cached_size is reach). I will keep looking into other alternatives and replace the library if I find one that fits.

One concern about the growing approach is that we can not completely avoid `cudaMalloc` as it is needed for growing the pool, so some requests will be affected if they happen to trigger the growth.

### DIFF
--- a/src/core/server.cc
+++ b/src/core/server.cc
@@ -50,6 +50,10 @@
 #include "src/core/server.h"
 #include "src/core/server_status.pb.h"
 
+#ifdef TRTIS_ENABLE_GPU
+#include "src/core/cuda_memory_manager.h"
+#endif  // TRTIS_ENABLE_GPU
+
 namespace nvidia { namespace inferenceserver {
 
 namespace {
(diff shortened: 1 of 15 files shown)


src/core/server.cc
```diff
@@ -95,6 +99,7 @@ InferenceServer::InferenceServer()
   strict_readiness_ = true;
   exit_timeout_secs_ = 30;
   pinned_memory_pool_size_ = 1 << 28;
+  cuda_memory_pool_size_ = 1 << 28;
```
deadeyegoodwin: I think 256MB might be too big for GPU memory default. How about 64MB.

PR #1182, opened 2020-03-11
Support for nested arrays in Inference Input/Outputs

CoderHam:
Support for nested JSON input and output in HTTP V2 Inference Request/Response


src/servers/http_server_v2.cc
```diff
@@ -413,13 +413,55 @@ HTTPAPIServerV2::ResponseRelease(
...
+template <typename T>
+void
+ReadDataFromJsonHelper(
+    std::vector<T>* data_vec, const char* dtype,
+    const rapidjson::Value& payload_data, int* counter)
+{
+  for (size_t i = 0; i < payload_data.Size(); i++) {
+    // If last dimension
+    if (!payload_data[i].IsArray()) {
+      if (strcmp(dtype, "BOOL")) {
```
deadeyegoodwin: This comparison tree and others like it are very inefficient. You should convert the dtype string once into DataType and then pass that around and use as necessary with a switch. In common.cc you can rename GetDataTypeProtocolString to DataTypeToProtocolString and add ProtocolStringToDataType. In ProtocolStringToDataType instead comparing the entire string over and over break it into a search tree where you look at just the first letter, etc. so that you only have to do 1 full string compare at the end to verify.

src/servers/http_server_v2.cc
```diff
@@ -428,6 +470,70 @@ ReadDataArrayFromJson(
         "failed to parse request buffer, tensor data must be an array");
   }
 
+  // BOOL vector is broken
```
deadeyegoodwin: Old comment?

CoderHam: Removed

src/servers/http_server_v2.cc
CoderHam: Done.

src/servers/http_server_v2.cc
```diff
@@ -423,214 +481,224 @@ ReadDataArrayFromJson(
...
+  for (rapidjson::SizeType i = 0; i < shape.Size(); i++) {
+    if (element_cnt == 0) {
+      element_cnt = shape[i].GetInt();
+    } else {
+      element_cnt *= shape[i].GetInt();
+    }
   }
 
+  *byte_size =
+      element_cnt * GetDataTypeByteSize(DataTypeToProtocolString(dtype));
```
deadeyegoodwin: There's a GetByteSize that takes DataType directly... in model_config_utils I think.

src/servers/common.cc
```diff
@@ -316,6 +316,35 @@ GetDataTypeByteSize(const std::string& protocol_dtype)
   }
 }
 
+size_t
```
deadeyegoodwin: We should have this code duplication. Did you prepend the call with the namespace? Namespace "ni" should already be defined in http_server.cc, so ni::GetDataTypeByteSize should work.

CoderHam: Fixed. Removed code duplication

src/servers/CMakeLists.txt
```diff
@@ -203,13 +203,15 @@ set(
   common.cc
   shared_memory_manager.cc
   ../core/logging.cc
+  ../core/model_config.cc
```
deadeyegoodwin: To HTTP_ENDPOINT_OBJECTS above can you add model-config-library instead of doing this.

CoderHam: Done

PR #1185, opened 2020-03-12
Link rt to prevent build failure without GRPC

CoderHam:


GuanLuo:
Can you also remove the librt dependency [here](https://github.com/NVIDIA/tensorrt-inference-server/blob/master/src/servers/CMakeLists.txt#L403), I think librt is for shared memory support, and since we removed the handler out of Triton core, it should not be needed there.

CoderHam:
You're right. I removed it

PR #1187, opened 2020-03-12
Fixing perf_client to include correct shared memory size

tanmayv25:


deadeyegoodwin:
@tanmayv25 You need to manually rebase this.

PR #1190, opened 2020-03-13
Support for Binary Data in Inference I/O

CoderHam:
Add Binary Data support for HTTP V2 Server

Merge nested JSON support prior to this. Will need to rebase manually after that.


deadeyegoodwin:
@CoderHam Make a branch off nested branch with just the changes needed for this PR and then show then merge into the nested branch, that was can review just the new changes without seeing changes from other PRs.

CoderHam:
> @CoderHam Make a branch off nested branch with just the changes needed for this PR and then show then merge into the nested branch, that was can review just the new changes without seeing changes from other PRs.
Done

src/servers/http_server_v2.cc
```diff
@@ -1110,23 +1110,32 @@ HTTPAPIServerV2::EVBufferToInput(
   //
   // Get the addr and size of from the evbuffer.
   int buffer_len = evbuffer_get_length(input_buffer);
-  char json_buffer[buffer_len];
+
+  std::vector<char> complete_buffer(buffer_len);
```
deadeyegoodwin: You can copy into complete_buffer.data() below without reserving space in the vector. But copying the entire body into a buffer is something we can't do because of performance. Note how the original code was careful to use the data directly from the evbuffer, even if that meant it had to cross multiple blocks. We need to continue to do the same thing. Even for json parsing we only want to copy if the json part is not in a contiguous buffer.

src/servers/http_server_v2.cc
```diff
@@ -1101,32 +1101,72 @@ HTTPAPIServerV2::EVBufferToInput(
...
+    }
+  }
+
+  if (header_length == 0) {
+    return TRTSERVER_ErrorNew(
+        TRTSERVER_ERROR_INTERNAL, "unexpected error header length of size 0");
+  }
+
+  // Extract just the json header from the complete buffer
+  char json_buffer[header_length];
```
deadeyegoodwin: Use std::vector<char>, not char[]. But you don't need to make a copy at all if the first evbuffer is large enough to contain the entire thing. You likely should move the header extraction logic into a different function. On return you just have to know the updated v_idx (and a v_offset which tells you have many bytes in that block are already consumed from the header).

src/servers/http_server_v2.cc
```diff
@@ -1177,28 +1217,110 @@ HTTPAPIServerV2::EVBufferToInput(
...
-      if (byte_size != 0) {
-        return TRTSERVER_ErrorNew(
-            TRTSERVER_ERROR_INVALID_ARG,
-            std::string(
-                "unexpected size for input '" +
-                std::string(request_input["name"].GetString()) +
-                "', expecting " + std::to_string(byte_size) +
-                " bytes for model '" + model_name + "'")
-                .c_str());
+        // Copy one block at a time
```
deadeyegoodwin: Not copying... don't say copy.

src/servers/http_server_v2.cc
```diff
@@ -1177,28 +1217,110 @@ HTTPAPIServerV2::EVBufferToInput(
       // FIXMEV2 handle non-raw content types
       char* base;
       size_t byte_size;
-      RETURN_IF_ERR(ReadDataFromJson(request_input, &base, &byte_size));
+      rapidjson::Value::ConstMemberIterator itr =
+          request_input.FindMember("parameters");
+      if (itr != request_input.MemberEnd()) {
+        if (header_length == 0) {
```
deadeyegoodwin: Why is this check here?

src/servers/http_server_v2.cc
```diff
@@ -1177,28 +1217,110 @@ HTTPAPIServerV2::EVBufferToInput(
...
+              element_cnt *= shape[i].GetInt();
+            }
+          }
+          byte_size = element_cnt * element_size;
+        }
 
-      RETURN_IF_ERR(TRTSERVER_InferenceRequestProviderSetInputData(
-          request_provider, request_input["name"].GetString(), base, byte_size,
-          TRTSERVER_MEMORY_CPU, 0 /* memory_type_id */));
+        int v_idx_tmp = v_idx;
```
deadeyegoodwin: v_idx_tmp? There is only one v_idx which walks over all the data.

CoderHam: This is needed to manage the case where the binary inputs in the request buffer may not in the order of the inputs in the header i.e. INPUT0 could be after INPUT1 in the binary data section after the json header but the INPUT0 may be before INPUT1 in the json section

src/servers/http_server_v2.cc
CoderHam: Removed

src/servers/http_server_v2.cc
```diff
@@ -1101,32 +1101,72 @@ HTTPAPIServerV2::EVBufferToInput(
...
+        std::string(
+            "unexpected size for request header, expecting " +
+            std::to_string(header_length_tmp) +
+            " more bytes").c_str());
   }
 
   rapidjson::Document document;
   rapidjson::Document::AllocatorType& allocator = document.GetAllocator();
-  document.Parse(json_buffer);
+  document.Parse(&json_buffer[0]);
```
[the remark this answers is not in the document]
CoderHam: You are right. I need to null-terminate the string. Nope there isn't a parse method that accepts an explicit length.

src/servers/http_server_v2.cc
```diff
@@ -1101,32 +1101,72 @@ HTTPAPIServerV2::EVBufferToInput(
...
+  if (n > 0) {
+    v = static_cast<struct evbuffer_iovec*>(
+        alloca(sizeof(struct evbuffer_iovec) * n));
+    if (evbuffer_peek(input_buffer, -1, NULL, v, n) != n) {
+      return TRTSERVER_ErrorNew(
+          TRTSERVER_ERROR_INTERNAL, "unexpected error getting input buffers");
+    }
+  }
+
+  if (header_length == 0) {
```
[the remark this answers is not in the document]
CoderHam: The check is done only when one of the inputs have a member "parameters" i.e. the case when binary data is used.

src/servers/http_server_v2.cc
CoderHam: As you specified, I need to null-terminate the string. This means that even if the first evbuffer is large enough to contain the entire thing, if it contains binary data, I will have to create a copy and null terminate that copy at the point when the json content ends.

src/servers/http_server_v2.cc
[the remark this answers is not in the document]
CoderHam: In the case that header_length is 0 we must assume the entire buffer is the json header

src/servers/http_server_v2.cc
CoderHam: The check is done only when one of the inputs have a member "parameters" i.e. the case when binary data is used.

src/servers/http_server_v2.cc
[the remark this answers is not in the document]
deadeyegoodwin: https://github.com/Tencent/rapidjson/blob/master/include/rapidjson/document.h#L2433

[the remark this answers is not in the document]

src/servers/http_server_v2.cc
```diff
@@ -1096,41 +1099,109 @@ HTTPAPIServerV2::HandleServerMetadata(evhtp_request_t* req)
   TRTSERVER_ErrorDelete(err);
 }
 
+TRTSERVER_Error*
+HTTPAPIServerV2::EVBufferToJson(
+    rapidjson::Document* document, evbuffer_iovec* v, int* v_idx,
+    const size_t length, int n)
+{
+  size_t offset = 0, length_tmp = length;
```
deadeyegoodwin: "remaining_length" or similar name would convey more meaning than "length_tmp"

src/servers/http_server_v2.cc
```diff
@@ -1096,41 +1099,109 @@ HTTPAPIServerV2::HandleServerMetadata(evhtp_request_t* req)
...
+  char *json_base;
+  std::vector<char> json_buffer;
+
+  // No need to memcpy when number of iovecs is 1
+  if (n == 1) {
```
deadeyegoodwin: You can use this simple no copy case when (n > 0) and (iov_len >= length)

src/servers/http_server_v2.cc
deadeyegoodwin: Parameters are not just to indicate binary data, they are/will-be used for other purposes.
Can you write a utility routine like bool IsBinaryInputData(..) that returns true/false and also set parameters for the offset/size. Perhaps having offset just makes the implementation harder? We can just have 'size' and define that the binary data is in same order as inputs in the json. In fact I prefer that since it should be more performant.

src/servers/http_server_v2.cc
deadeyegoodwin: This can be simplified by dropping "offset" so we should do that.

src/servers/http_server_v2.cc
```diff
@@ -1297,9 +1450,17 @@ HTTPAPIServerV2::HandleInfer(
   if (err == nullptr) {
     std::unique_ptr<InferRequestClass> infer_request(new InferRequestClass(
         req, request_header_protobuf.id(), server_id_, unique_id));
+
+    // Find Inference-Header-Content-Length in header. If missing set to 0
```
deadeyegoodwin: This is not given when not using binary data... in that case does EVBufferToInput do the right thing?

src/servers/http_server_v2.cc
```diff
@@ -1096,41 +1099,109 @@ HTTPAPIServerV2::HandleServerMetadata(evhtp_request_t* req)
...
-  if (document.HasParseError()) {
-    return TRTSERVER_ErrorNew(
-        TRTSERVER_ERROR_INVALID_ARG,
-        std::string(
-            "failed to parse the request buffer: " +
-            std::string(GetParseError_En(document.GetParseError())) + " at " +
-            std::to_string(document.GetErrorOffset()))
-            .c_str());
+  int buffer_len = 0;
+  if (header_length == 0) {
```
deadeyegoodwin: Then clause doesn't do anything... and also if the kInferHeaderContentLengthHTTPHeader is not given then you don't read any json?

src/servers/http_server_v2.cc
CoderHam: I agree. Having just size and enforcing the order to be the same would be more performant

src/servers/http_server_v2.cc
CoderHam: Assuming that the offset is removed, we can simply rely on the byte size and will only go through the vecs once

src/servers/http_server_v2.cc
deadeyegoodwin: ok, I will update the extension doc to not include the offset.

src/servers/http_server_v2.cc
CoderHam: In the case that the header is missing header_length is set to 0. Then in EVBuffer we pick up the evbuffer length and make the entire content the json header.

src/servers/http_server_v2.cc
CoderHam: Fixed this check

PR #1191, opened 2020-03-13
Python client library support for cuda shared memory in gRPC V2

tanmayv25:


CoderHam:
libb64-dev needs to be installed on the jetson devices as well. Can you add this to the jetson-setup file? I will ssh into the machines and install them manually for now.

tanmayv25:
Sure

src/core/grpc_service_v2.proto
```diff
@@ -1123,7 +1123,7 @@ message CudaSharedMemoryRegisterRequest
 
   //@@  .. cpp:var:: bytes raw_handle
   //@@
-  //@@     The raw serialized cudaIPC handle.
+  //@@     The raw serialized cudaIPC handle in base64 encoding.
```
deadeyegoodwin: For protobuf we don't need to base64 endode this. For your python layer you can use base64... to make easier to visualize/debug the handle. But for here you should just encode the raw bytes, which means you will have to base64 decode in the library when constructing the message.

src/clients/python/experimental_api_v2/library/CMakeLists.txt
```diff
@@ -38,12 +38,14 @@ if(${TRTIS_ENABLE_HTTP_V2} OR ${TRTIS_ENABLE_GRPC_V2})
     target_link_libraries(cshmv2 rt)
 
     #
-    # libccudashm.so
+    # libccudashmv2.so
```
GuanLuo: Is it necessary to have v2 of the utility libraries? I think both v1 and v2 clients can use it?

src/servers/grpc_server_v2.cc
```diff
@@ -840,12 +841,16 @@ CommonHandler::SetUpAllRequests()
         TRTSERVER_Error* err = nullptr;
 #ifdef TRTIS_ENABLE_GPU
         const std::string& raw_handle = request.raw_handle();
-        char* handle_base = const_cast<char*>(raw_handle.c_str());
-        cudaIpcMemHandle_t* cuda_shm_handle =
-            reinterpret_cast<cudaIpcMemHandle_t*>(handle_base);
+        // Decode the handle from base64
+        void* cuda_shm_handle = malloc(raw_handle.size());
```
GuanLuo: There are a few `malloc()` calls for encoding / decoding purpose, are the resources being cleaned up properly?

src/clients/python/experimental_api_v2/library/CMakeLists.txt
tanmayv25: We have renamed every 'trtis' occurence to 'triton' in the v2 libraries.

src/core/grpc_service_v2.proto
tanmayv25: Done

src/clients/python/experimental_api_v2/library/cuda_shared_memory/__init__.py
```diff
@@ -114,14 +116,15 @@ def get_raw_handle(cuda_shm_handle):
...
+    bytes 
+        The raw serialized cudaIPC handle of underlying cuda shared memory
+        in base64 encoding
 
     """
     craw_handle = c_char_p()
     _raise_if_error(c_int(_ccudashm_get_raw_handle(cuda_shm_handle, byref(craw_handle))))
 
-    return craw_handle.value
+    return base64.b64decode(craw_handle.value.decode())
```
deadeyegoodwin: Why are you encoding it on the C side just to immediately decode it here? Why not just leave it as raw bytes. The return description also seems wrong.

tanmayv25: Ahh, I misinterpreted your earlier suggestion. Using base64 was a workaround for ctypes. The datatypes were null-terminated when passed from c to python.

Fixed as per your original suggestion.

PR #1193, opened 2020-03-16
Inference for experimental trtserver2 API

deadeyegoodwin:


src/core/infer_request.cc
```diff
@@ -181,49 +275,50 @@ InferenceRequest::Normalize(const InferenceBackend& backend)
   for (auto& pr : inputs_) {
     const ModelInput* input_config;
     RETURN_IF_ERROR(backend.GetInput(pr.first, &input_config));
+    auto& input = pr.second;
 
-    auto& shape = *pr.second.MutableShape();
+    auto new_shape = input.MutableShape();
+    *new_shape = input.OriginalShape();
```
GuanLuo: This should be moved to the if clause in line 332? Otherwise that if predicate will always be false.

src/core/infer_request.cc
```diff
@@ -181,49 +275,50 @@ InferenceRequest::Normalize(const InferenceBackend& backend)
...
       if (input_config->has_reshape()) {
         std::deque<int64_t> variable_size_values;
         for (int64_t idx = 0; idx < input_config->dims_size(); idx++) {
           if (input_config->dims(idx) == -1) {
-            variable_size_values.push_back(shape[idx]);
+            variable_size_values.push_back((*new_shape)[idx]);
           }
         }
 
-        shape.clear();
```
GuanLuo: This is needed, below is appending the reshaped value to the `new_shape`, so it should be empty to start with.

src/core/trtserver.cc
```diff
@@ -1722,6 +1774,418 @@ TRTSERVER_ServerInferAsync(
...
+      ni::InferRequestProvider::Create(lrequest, &infer_request_provider));
+
+  std::shared_ptr<ni::InferResponseProvider> infer_response_provider;
+  {
+    std::shared_ptr<ni::InferResponseProvider> del_response_provider;
+    RETURN_IF_STATUS_ERROR(ni::InferResponseProvider::Create(
+        lrequest, lbackend->GetLabelProvider(), response_allocator,
+        lresponsealloc->AllocFn(), response_allocator_userp,
+        lresponsealloc->ReleaseFn(), &del_response_provider));
+    infer_response_provider = del_response_provider;
```
GuanLuo: `std::move()` to avoid copy assignment

src/servers/simplev2.cc
```diff
@@ -0,0 +1,694 @@
+// Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
```
GuanLuo: 2020

src/core/trtserver2.h
```diff
@@ -0,0 +1,340 @@
...
+    TRTSERVER2_InferenceRequest* inference_request, const char* name,
+    const void* base, size_t byte_size, TRTSERVER_Memory_Type memory_type,
+    int64_t memory_type_id);
+
+/// Clear all input data from an input, releasing ownership of the
+/// buffer(s) that were appended to the input with
+/// TRTSERVER2_InferenceRequestAppendInputData.
+/// \param inference_request The request object.
+/// \param name The name of the input.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestRemoveAllInputData(
```
GuanLuo: So to change the input data, user will call `RemoveAllInputData` first and then call the append above?

src/core/infer_request.cc
deadeyegoodwin: OriginalShape will be empty if the request didn't specify a shape.

src/core/infer_request.cc
deadeyegoodwin: Good catch!

src/core/trtserver2.h
deadeyegoodwin: If you want to use the same input buffers you can just change the data in place and then run inference again. If you want to set an input to use a different buffer(s) then you must remove-all and then append the new buffers.

PR #1195, opened 2020-03-17
[HTTP V2] Update V2 Infer API, Fix response error json

CoderHam:
Fix Infer - Update to use new V2 Infer API
Ensure error message responses are in json form

### DIFF
--- a/src/servers/http_server_v2.cc
+++ b/src/servers/http_server_v2.cc
@@ -918,7 +918,6 @@ HTTPAPIServerV2::HandleModelReady(
   }
 
   TRTSERVER_ProtobufDelete(model_status_protobuf);
-
   evhtp_send_reply(
       req, (ready && (err == nullptr)) ? EVHTP_RES_OK : EVHTP_RES_BADREQ);
 


src/servers/http_server_v2.cc
```diff
@@ -768,18 +768,22 @@ HTTPAPIServerV2::HandleServerReady(
...
-      server_id_);
-
-  evhtp_headers_add_header(
-      req->headers_out,
-      evhtp_header_new(
-          kStatusHTTPHeader, request_status.ShortDebugString().c_str(), 1, 1));
-
-  evhtp_send_reply(
-      req, (ready && (err == nullptr)) ? EVHTP_RES_OK : EVHTP_RES_BADREQ);
+  if (err == nullptr) {
```
deadeyegoodwin: Why is this function called "Ready" when it is both live and ready. Why is the model version called "Health" when it is just ready. Please take some time to think of those coming after you into this code and make it as easy as possible for them to understand what is going on.

src/servers/http_server_v2.cc
```diff
@@ -768,18 +768,22 @@ HTTPAPIServerV2::HandleServerReady(
...
+    evhtp_send_reply(req, ready ? EVHTP_RES_OK : EVHTP_RES_BADREQ);
+  } else {
+    rapidjson::Document document;
```
deadeyegoodwin: Look carefully at the design doc, live and ready endpoints do not return a json response.

src/servers/http_server_v2.cc
```diff
@@ -998,27 +1010,21 @@ HTTPAPIServerV2::HandleModelMetadata(
 
     evbuffer_add(
         req->buffer_out, model_metadata.c_str(), model_metadata.size());
+    evhtp_send_reply(req, EVHTP_RES_OK);
+  } else {
+    document.Clear();
```
deadeyegoodwin: Should make a common routine for this, don't copy-paste. Also this seems unnecessarily complex for such a simple json response. Just create it as a string... will be better performance also.

src/servers/http_server_v2.cc
CoderHam: Sorry. I will make the necessary changes

src/servers/http_server_v2.cc

src/servers/http_server_v2.cc
```diff
@@ -271,12 +271,13 @@ class HTTPAPIServerV2 : public HTTPServerV2Impl {
       }
     }
 
+    void SetRequestId(uint64_t request_id) { request_id_ = request_id; }
```
deadeyegoodwin: The id is available in the InferenceRequest object, does it need to be maintained here as well?

CoderHam:
Resolved and started one now

> I don't see a CI for this.

deadeyegoodwin:
@CoderHam needs manual rebase

CoderHam:
Changes already picked up the Binary Data PR.

PR #1197, opened 2020-03-17
[HTTP V2] Add support for Model Repository/Control

CoderHam:
PS: Repository name is not yet supported. Also applies for GRPC V2.

### DIFF
--- a/src/core/trtserver.cc
+++ b/src/core/trtserver.cc
@@ -1648,8 +1648,9 @@ TRTSERVER_ServerModelRepositoryIndex(
 
 // Overload to remove GRPC V2 dependency
 TRTSERVER_Error*
-TRTSERVER_ServerModelRepositoryIndexHTTP(
-    TRTSERVER_Server* server, const char*** models, uint64_t* models_count)
+TRTSERVER2_ServerModelRepositoryIndex(
+    TRTSERVER_Server* server, const char* const** models,
+    uint64_t* models_count)
 {
   ni::InferenceServer* lserver = reinterpret_cast<ni::InferenceServer*>(server);
 
--- a/src/core/trtserver.cc
+++ b/src/core/trtserver.cc
@@ -644,10 +644,11 @@ class TrtServerModelIndex {
 
 TrtServerModelIndex::TrtServerModelIndex(
     const ni::ModelRepositoryIndex& model_repository_index)
-    : model_repository_index_(model_repository_index) {
+    : model_repository_index_(model_repository_index)
 {
-  for (const auto& model : model_repository_index.models())
-  index_.push_back(model.name().c_str());
+  {
+    for (const auto& model : model_repository_index_.models())
+      index_.push_back(model.name().c_str());
   }
 }
 
--- a/src/core/trtserver.cc
+++ b/src/core/trtserver.cc
@@ -635,28 +635,35 @@ class TrtServerModelIndex {
  public:
   TrtServerModelIndex(ni::ModelRepositoryIndex model_repository_index);
   TRTSERVER_Error* GetModelNames(
-      const char* const** models, uint64_t* models_count);
+      const char*** models, uint64_t* models_count);
 
  private:
   ni::ModelRepositoryIndex model_repository_index_;
-  std::vector<const char*> index_;
+  const char** models_;
 };
 
 TRTSERVER_Error*
 TrtServerModelIndex::GetModelNames(
-    const char* const** models, uint64_t* models_count)
+    const char*** models, uint64_t* models_count)
 {
-  index_.clear();
+  // index_.clear();
+  if (model_repository_index_.models_size() > 0) {
+    models_ = (const char**)malloc(model_repository_index_.models_size());
+  }
+  *models_count = 0;
   for (const auto& model : model_repository_index_.models()) {
-    index_.push_back(model.name().c_str());
+    // index_.push_back(model.name().c_str());
+    models_[*models_count] = model.name().c_str();
+    *models_count+=1;
   }
 
-  if (index_.empty()) {
-    *models_count = 0;
+  if (*models_count == 0) {
+    // if (index_.empty()) {
+    // *models_count = 0;
     *models = nullptr;
   } else {
-    *models_count = index_.size();
-    *models = &(index_[0]);
+    // *models_count = index_.size();
+    *models = &(models_[0]);
   }
 
   return nullptr;
@@ -1712,7 +1719,7 @@ TRTSERVER2_ServerModelIndex(
 
 TRTSERVER_Error*
 TRTSERVER2_ModelIndexNames(
-    TRTSERVER2_ModelIndex* model_index, const char* const** models,
+    TRTSERVER2_ModelIndex* model_index, const char*** models,
     uint64_t* models_count)
 {
   TrtServerModelIndex* index =
@@ -1726,7 +1733,7 @@ TRTSERVER2_ModelIndexDelete(TRTSERVER2_ModelIndex* model_index)
 {
   TrtServerModelIndex* index =
       reinterpret_cast<TrtServerModelIndex*>(model_index);
-  delete indices;
+  delete index;
   return nullptr;  // Success
 }
 
--- a/src/core/trtserver.cc
+++ b/src/core/trtserver.cc
@@ -633,13 +633,23 @@ TrtServerResponse::OutputData(
 //
 class TrtServerModelIndex {
  public:
-  TrtServerModelIndex(ni::ModelRepositoryIndex model_repository_index);
+  TrtServerModelIndex(const ni::ModelRepositoryIndex& model_repository_index);
   TRTSERVER_Error* GetModelNames(
       const char* const** models, uint64_t* models_count);
 
  private:
+  ni::ModelRepositoryIndex model_repository_index_;
   std::vector<const char*> index_;
 };
+
+TrtServerModelIndex::TrtServerModelIndex(
+    const ni::ModelRepositoryIndex& model_repository_index)
+    : model_repository_index_(model_repository_index) {
+{
+  for (const auto& model : model_repository_index.models())
+  index_.push_back(model.name().c_str());
+  }
+}
 
 TRTSERVER_Error*
 TrtServerModelIndex::GetModelNames(
@@ -654,14 +664,6 @@ TrtServerModelIndex::GetModelNames(
   }
 
   return nullptr;
-}
-
-TrtServerModelIndex::TrtServerModelIndex(
-    ni::ModelRepositoryIndex model_repository_index)
-{
-  for (const auto& model : model_repository_index.models()) {
-    index_.push_back(model.name().c_str());
-  }
 }
 
 
--- a/src/core/trtserver.cc
+++ b/src/core/trtserver.cc
@@ -633,19 +633,20 @@ TrtServerResponse::OutputData(
 //
 class TrtServerModelIndex {
  public:
-  TrtServerModelIndex() = default;
-  TRTSERVER_Error* GetModels(
+  TrtServerModelIndex(ni::ModelRepositoryIndex model_repository_index);
+  TRTSERVER_Error* GetModelNames(
       const char* const** models, uint64_t* models_count);
-  ni::ModelRepositoryIndex model_repository_index_;
 
  private:
+  ni::ModelRepositoryIndex model_repository_index_;
   std::vector<const char*> indices_;
 };
 
 TRTSERVER_Error*
-TrtServerModelIndex::GetModels(
+TrtServerModelIndex::GetModelNames(
     const char* const** models, uint64_t* models_count)
 {
+  indices_.clear();
   for (const auto& model : model_repository_index_.models()) {
     indices_.push_back(model.name().c_str());
   }
@@ -660,6 +661,13 @@ TrtServerModelIndex::GetModels(
 
   return nullptr;
 }
+
+TrtServerModelIndex::TrtServerModelIndex(
+    ni::ModelRepositoryIndex model_repository_index)
+    : model_repository_index_(model_repository_index)
+{
+}
+
 
 }  // namespace
 
@@ -1682,7 +1690,7 @@ TRTSERVER_ServerModelRepositoryIndex(
 }
 
 TRTSERVER_Error*
-TRTSERVER2_ServerModelRepositoryIndexNew(
+TRTSERVER2_ServerModelIndex(
     TRTSERVER_Server* server, TRTSERVER2_ModelIndex** model_indices)
 {
   ni::InferenceServer* lserver = reinterpret_cast<ni::InferenceServer*>(server);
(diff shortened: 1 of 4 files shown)


src/core/trtserver.cc
```diff
@@ -1594,6 +1594,38 @@ TRTSERVER_ServerModelRepositoryIndex(
   return nullptr;  // success
 }
 
+// Overload to remove GRPC V2 dependency
+TRTSERVER_Error*
+TRTSERVER_ServerModelRepositoryIndexHTTP(
```
deadeyegoodwin: New TRTSERVER API should be added in trtserver2.h. Call this one TRTSERVER2_ServerModelIndex, and see TRTSERVER_ServerExtensions for how to type the arguments. But I think you are actually going to have to create a TRTSERVER2_ModelIndex opaque object and return that (second argument will be TRTSERVER2_ModelIndex*, because otherwise you will not be able to control the life-cycle of the strings. See my comment below.

src/servers/http_server_v2.cc
```diff
@@ -784,6 +804,109 @@ HTTPAPIServerV2::HandleServerReady(
...
+  if (action == "load") {
+    err = TRTSERVER_ServerLoadModel(server_.get(), model_name.c_str());
+  } else if (action == "unload") {
+    err = TRTSERVER_ServerUnloadModel(server_.get(), model_name.c_str());
+  }
+
+  if (err == nullptr) {
+    evhtp_send_reply(req, EVHTP_RES_OK);
+  } else {
+    rapidjson::Document document;
```
deadeyegoodwin: FIx to use error return utility

src/servers/http_server_v2.cc
```diff
@@ -784,6 +804,109 @@ HTTPAPIServerV2::HandleServerReady(
...
+    rapidjson::StringBuffer buffer;
+    buffer.Clear();
+    rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+    document.Accept(writer);
+    std::string model_metadata(buffer.GetString());
+    evbuffer_add(
+        req->buffer_out, model_metadata.c_str(), model_metadata.size());
+    evhtp_send_reply(req, EVHTP_RES_OK);
+  } else {
+    document.Clear();
```
deadeyegoodwin: Utility...

src/core/trtserver.cc
```diff
@@ -1594,6 +1594,38 @@ TRTSERVER_ServerModelRepositoryIndex(
...
+      lserver->StatusManager(), ni::ServerStatTimerScoped::Kind::REPOSITORY);
+#endif  // TRTIS_ENABLE_STATS
+
+  ni::ModelRepositoryIndex model_repository_index;
+  RETURN_IF_STATUS_ERROR(
+      lserver->GetModelRepositoryIndex(&model_repository_index));
+
+  std::vector<const char*> indices;
+  for (const auto& model : model_repository_index.models()) {
+    indices.push_back(model.name().c_str());
```
deadeyegoodwin: You are capturing a string here but you don't know or control the lifecycle of 'model'. You need to have a TrtModelndex class (which you expose as TRTSERVER2_ModelIndex opaque pointer) and in that capture your own copy of the strings so you can control lifecycle.

src/core/trtserver2.h
```diff
@@ -297,6 +298,25 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
     const void** base, size_t* byte_size, TRTSERVER_Memory_Type* memory_type,
     int64_t* memory_type_id);
 
+/// Get the indices of all unique models in the model repository.
```
deadeyegoodwin: Follow pattern from trtserver.h, each API object has section, so above have:
//
// TRTSERVER2_ModelIndex
//
with TRTSERVER2_ModelIndexDelete(). You also need a function to get the actual index strings out of the object. Like TRTSERVER2_ModelIndexNames to return pointers to all the names. The returned pointers will be owned by the object and so the caller cannot modify or delete. See description for TRTSERVER_ServerExtensions.
You do not need a New because the only way to create is by the server call below.

src/core/trtserver2.h
```diff
@@ -297,6 +298,25 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
+/// The caller does not own the returned strings and must not modify or
+/// delete them. The lifetime of the returned strings extends only as
+/// long as 'server' and must not be accessed once 'server' is deleted.
+/// \param server The inference server object.
+/// \param model_indices Returns the TRTSERVER2_ModelIndex object that
+/// can then be used to get the indices of all unique models in the model
+/// repository.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_ServerModelRepositoryIndexNew(
```
deadeyegoodwin: The only function you need here is:
TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_ServerModelIndex(
    TRTSERVER_Server* server, TRTSERVER_ModelIndex** index);

src/core/trtserver.cc
```diff
@@ -1646,6 +1681,46 @@ TRTSERVER_ServerModelRepositoryIndex(
...
+
+#ifdef TRTIS_ENABLE_STATS
+  ni::ServerStatTimerScoped timer(
+      lserver->StatusManager(), ni::ServerStatTimerScoped::Kind::REPOSITORY);
+#endif  // TRTIS_ENABLE_STATS
+
+  TrtServerModelIndex* indices = new TrtServerModelIndex();
+  *model_indices = reinterpret_cast<TRTSERVER2_ModelIndex*>(indices);
+  RETURN_IF_STATUS_ERROR(
+      lserver->GetModelRepositoryIndex(&indices->model_repository_index_));
```
deadeyegoodwin: Don't expose model_repository_index_ publically. Just pass GetModelRepositoryIndex() into the constructor of TrtServerModelIndex.

src/core/trtserver.cc
```diff
@@ -626,6 +626,41 @@ TrtServerResponse::OutputData(
...
+  TRTSERVER_Error* GetModels(
+      const char* const** models, uint64_t* models_count);
+  ni::ModelRepositoryIndex model_repository_index_;
+
+ private:
+  std::vector<const char*> indices_;
+};
+
+TRTSERVER_Error*
+TrtServerModelIndex::GetModels(
```
deadeyegoodwin: GetModelNames? This can just return pointers directly into model_repository_index_ strings. Don't need to create indices_.

src/core/trtserver.cc
```diff
@@ -626,6 +626,49 @@ TrtServerResponse::OutputData(
...
+ private:
+  ni::ModelRepositoryIndex model_repository_index_;
+  std::vector<const char*> indices_;
+};
+
+TRTSERVER_Error*
+TrtServerModelIndex::GetModelNames(
+    const char* const** models, uint64_t* models_count)
+{
+  indices_.clear();
```
deadeyegoodwin: Don't use indices_, get the pointers to return directly from model_repository_index_.
As it is now if I call GetModelNames a second time I will invalidate all the pointers that I got back the first time I called it.

src/core/trtserver2.h
```diff
@@ -304,6 +305,42 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
     TRTSERVER2_InferenceRequest* inference_request);
 
+///
+/// TRTSERVER2_ModelIndex
+///
+/// Object representing model indices.
+///
+
+/// Get the object representing the indices of all unique models in the
+/// model repository.
```
deadeyegoodwin: model repositories.

src/core/trtserver2.h
```diff
@@ -304,6 +305,42 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
+/// \param server The inference server object.
+/// \param model_indices Returns the TRTSERVER2_ModelIndex object that
+/// is used to manage the lifecycle of the returned strings.
```
deadeyegoodwin: Don't write a programmer description, write the user description. "Return the TRTSERVER2_ModelIndex object that holds the index of all models contained in the server's model repository(s)."

src/core/trtserver2.h
```diff
@@ -304,6 +305,42 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_ServerModelIndex(
+    TRTSERVER_Server* server, TRTSERVER2_ModelIndex** model_indices);
+
+/// Get the names of all unique models in the model repository.
+/// The caller does not own the returned strings and must not modify or
+/// delete them. The lifetime of the returned strings extends only as
+/// long as 'model_indices' and must not be accessed once 'model_indices'
+/// is deleted.
+/// \param model_indices The TRTSERVER2_ModelIndex object that
```
deadeyegoodwin: All the places you call this model_indices is confusing, because TRTSERVER2_ModelIndex is a single index. It is an index that contains all the models in the repository(s). Currently it only has the names of the models but in the future it might have more index information. So "\param model_index The index of models".

src/core/trtserver2.h
```diff
@@ -304,6 +305,42 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
+/// is used to manage the lifecycle of the returned strings.
+/// \param models Returns the names of all unique models as an array of
+/// pointers to the name of each extension.
```
deadeyegoodwin: extension?

src/core/trtserver2.h
```diff
@@ -304,6 +305,42 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
+/// \param models_count Returns the number of unique models.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_ModelIndexNames(
+    TRTSERVER2_ModelIndex* model_indices, const char* const** models,
+    uint64_t* models_count);
+
+/// Delete a model indices object.
```
deadeyegoodwin: index

src/servers/http_server_v2.cc
```diff
@@ -846,6 +868,97 @@ HTTPAPIServerV2::HandleServerHealth(
...
+{
+  if (req->method != htp_method_GET) {
+    evhtp_send_reply(req, EVHTP_RES_METHNALLOWED);
+    return;
+  }
+
+  TRTSERVER_Error* err = nullptr;
+  const char* const* models;
+  uint64_t models_count;
+  TRTSERVER2_ModelIndex* model_indices = nullptr;
```
deadeyegoodwin: model_index

src/servers/http_server_v2.cc
```diff
@@ -846,6 +868,97 @@ HTTPAPIServerV2::HandleServerHealth(
...
+        TRTSERVER_ERROR_UNSUPPORTED,
+        "'repository_name' specification is not yet supported");
+  }
+
+  rapidjson::Document document;
+  document.SetObject();
+  rapidjson::Document::AllocatorType& allocator = document.GetAllocator();
+  if (err == nullptr) {
+    rapidjson::Value models_array(rapidjson::kArrayType);
+    LOG_VERBOSE(1) << "models size: " << models_count;
```
deadeyegoodwin: Remove these verbose logs

src/servers/http_server_v2.cc
```diff
@@ -846,6 +868,97 @@ HTTPAPIServerV2::HandleServerHealth(
...
+    for (uint64_t i = 0; i < models_count; i++) {
+      rapidjson::Value model_index;
+      model_index.SetObject();
+      std::string model_str(models[i]);
```
deadeyegoodwin: It is a little strange that you convert to a std::string and then extract the c_str out of that. Why not just use model directly and strlen(model)?

src/servers/http_server_v2.cc
```diff
@@ -846,6 +868,97 @@ HTTPAPIServerV2::HandleServerHealth(
...
+      LOG_VERBOSE(1) << "added name: " << model_str;
+      models_array.PushBack(model_index, allocator);
+    }
+    document.AddMember("index", models_array, allocator);
+
+    rapidjson::StringBuffer buffer;
+    buffer.Clear();
+    rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+    document.Accept(writer);
+    std::string model_metadata(buffer.GetString());
```
deadeyegoodwin: You do it again here...

src/core/trtserver.cc
```diff
@@ -626,6 +626,45 @@ TrtServerResponse::OutputData(
...
 }
 
+//
+// TrtServerModelIndex
+//
+// Implementation for TRTSERVER2_ModelIndex.
+//
+class TrtServerModelIndex {
+ public:
+  TrtServerModelIndex(ni::ModelRepositoryIndex model_repository_index);
```
deadeyegoodwin: const & the argument...

src/core/trtserver.cc
```diff
@@ -626,6 +626,45 @@ TrtServerResponse::OutputData(
...
+    *models = nullptr;
+  } else {
+    *models_count = index_.size();
+    *models = &(index_[0]);
+  }
+
+  return nullptr;
+}
+
+TrtServerModelIndex::TrtServerModelIndex(
```
deadeyegoodwin: Move this constructor implementation above GetModelNames. Also, you still need to keep a copy of model_repository_index in the object, otherwise all your name pointers are going to become invalid once the called deletes that object. So just add " : model_repository_index_(model_repository_index)"

src/core/trtserver2.h
```diff
@@ -304,6 +305,42 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
...
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_ServerModelIndex(
+    TRTSERVER_Server* server, TRTSERVER2_ModelIndex** model_index);
+
+/// Get the names of all unique models in the model repository.
+/// The caller does not own the returned strings and must not modify or
+/// delete them. The lifetime of the returned strings extends only as
+/// long as 'model_index' and must not be accessed once 'model_index'
+/// is deleted.
+/// \param model_index The TRTSERVER2_ModelIndex object that
+/// is used to manage the lifecycle of the returned strings.
```
deadeyegoodwin: Don't talk about "live-cycle", just "The index of models" should be sufficient.

src/core/trtserver.cc
```diff
@@ -626,6 +626,47 @@ TrtServerResponse::OutputData(
...
+ private:
+  ni::ModelRepositoryIndex model_repository_index_;
+  std::vector<const char*> index_;
+};
+
+TrtServerModelIndex::TrtServerModelIndex(
+    const ni::ModelRepositoryIndex& model_repository_index)
+    : model_repository_index_(model_repository_index) {
+{
+  for (const auto& model : model_repository_index.models())
```
deadeyegoodwin: You need to use model_repository_index_ so that your pointer references are into your copy of ModelRepositoryIndex.

src/core/trtserver.cc
```diff
@@ -626,6 +626,47 @@ TrtServerResponse::OutputData(
...
+  index_.push_back(model.name().c_str());
```
deadeyegoodwin: I think you need to run the formatter.

PR #1198, opened 2020-03-17
Add unit tests for CUDAMemoryManager, AllocatedMemory class

GuanLuo:
Also reorganize the location of utility function, and behavior of CUDAMemoryManager.


src/test/CMakeLists.txt
```diff
@@ -43,3 +43,97 @@ install(
...
+)
+set_target_properties(
+  memory_test
+  PROPERTIES
+    SKIP_BUILD_RPATH TRUE
+    BUILD_WITH_INSTALL_RPATH TRUE
+    INSTALL_RPATH_USE_LINK_PATH FALSE
+    INSTALL_RPATH ""
+)
+target_include_directories(memory_test PRIVATE ${GTEST_INCLUDE_DIR} PRIVATE ${CUDA_INCLUDE_DIRS} PRIVATE ${CNMEM_PATH}/include)
```
deadeyegoodwin: break long line

src/test/memory_test.cc
```diff
@@ -0,0 +1,356 @@
...
+}
+
+}  // namespace
+
+int
+main(int argc, char** argv)
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
```
deadeyegoodwin: Put newline at end of file

src/core/server.cc
```diff
@@ -149,9 +149,10 @@ InferenceServer::Init()
   CudaMemoryManager::Options cuda_options(
       min_supported_compute_capability_, cuda_memory_pool_size_);
   status = CudaMemoryManager::Create(cuda_options);
+  // If CUDA memory manager can't be created, just log error as the
+  // server can still function properly
   if (!status.IsOk()) {
-    ready_state_ = ServerReadyState::SERVER_FAILED_TO_INITIALIZE;
-    return status;
+    LOG_ERROR << status.Message();
```
deadeyegoodwin: Don't you need to do something special to fail appropriately if load/unload is called and the manager didn't create correctly?

GuanLuo: The Alloc / Free functions are checking if the CUDA memory manager is created, if not, the functions will return error

PR #1199, opened 2020-03-18
Python HTTP V2 client library

tanmayv25:
Note: Though a test template is included, this PR doesn't activate testing. There are some bugs on the server side which should be fixed before enabling the testing. 

The code changes can be used to reproduce the  failing cases.

### DIFF
--- a/qa/L0_http_v2/test.sh
+++ b/qa/L0_http_v2/test.sh
@@ -68,11 +68,6 @@ if [ $? -ne 0 ]; then
     RET=1
 fi
 
-if [ $(cat ${CLIENT_LOG}.health | grep "PASS" | wc -l) -ne 7 ]; then
-    cat ${CLIENT_LOG}.health
-    RET=1
-fi
-
 kill $SERVER_PID
 wait $SERVER_PID
 
--- a/qa/L0_http_v2/test.sh
+++ b/qa/L0_http_v2/test.sh
@@ -1,4 +1,5 @@
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#!/bin/bash
+# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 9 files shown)


src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
+        request_body = json.dumps(infer_request)
+        if not model_version:
+            request_uri = "v2/models/{}/infer".format(model_name)
+        else:
+            request_uri = "v2/models/{}/versions/{}/infer".format(
+                model_name, model_version)
+        response = self._client_stub.post(request_uri=request_uri,
+                                          body=request_body,
+                                          headers=headers)
+        response = self._client_stub.get(request_uri, headers=headers)
```
CoderHam: Remove this additional GET request

qa/L0_http_v2/test.sh
```diff
@@ -0,0 +1,88 @@
...
+fi
+
+# Test health
+python $SIMPLE_HEALTH_CLIENT -v >> ${CLIENT_LOG}.health 2>&1
+if [ $? -ne 0 ]; then
+    cat ${CLIENT_LOG}.health
+    RET=1
+fi
+
+if [ $(cat ${CLIENT_LOG}.health | grep "PASS" | wc -l) -ne 7 ]; then
```
nieksand: You could have the `simple_http_v2_health_metadata` return a non-zero exit code (`sys.exit(1)`) if anything does not PASS.  

Then you don't need hanky panky like counting the number of PASS statements in this bash script.

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
         Raises
         ------
-        InferenceServerException
+        Exception
             If unable to get server metadata.
 
         """
+        response = self._client_stub.get("v2", headers=headers)
+        raise_if_error(response)
+        metadata = json.loads(response.read())
```
nieksand: If response is file-like, then `json.load(response)`

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
+        return metadata
+
+    def infer(self,
+              inputs,
+              model_name,
+              model_version="",
+              outputs=None,
+              request_id=None,
+              parameters=None,
+              headers={}):
```
nieksand: Using a mutable type to default `headers` is dangerous.  

See: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

(It basically behaves like a static function variable in C++)

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
+            this_input._get_tensor() for this_input in inputs
+        ]
+        if outputs:
+            infer_request['outputs'] = [
+                this_output._get_tensor() for this_output in outputs
+            ]
+
+        request_body = json.dumps(infer_request)
+        if not model_version:
+            request_uri = "v2/models/{}/infer".format(model_name)
```
nieksand: Are model names guaranteed to only have URL-safe characters?  

Otherwise something like https://docs.python.org/3/library/urllib.parse.html#urllib.parse.quote

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
+        
+        """
+        self._parameters.clear()
+
+    def _get_tensor(self):
+        """Retrieve the underlying input as json string.
+
+        Returns
+        -------
+        str
```
nieksand: Your code seems to return a dict, not a string.

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
+    Parameters
+    ----------
+    name : str
+        The name of output tensor to associate with this object
+    """
+
+    def __init__(self, name):
+        self._output = {}
+        self._parameters = {}
+        self._output['name'] = name
```
nieksand: Rather than shoving in a dict, consider keeping `name` a string:
```
self._name = name
self._parameters = {}
```

Your `_get_tensor()` method becomes:
```
    return {'name': self._name, 'parameters': self._parameters}
```

Right now `_output` is a bit mushy with state.  It always has a `name` key and it might have a `parameters` key, but only if `_get_tensor()` was called.  And those parameters are also accessible via another member variable too.

src/clients/python/experimental_api_v2/examples/simple_http_v2_health_metadata.py
```diff
@@ -0,0 +1,83 @@
...
+    if (metadata['name'] == model_name):
+        print("PASS: get_model_metadata")
+    print(metadata)
+
+    # Passing incorrect model name
+    try:
+        metadata = triton_client.get_model_metadata("wrong_model_name")
+    except InferenceServerException as ex:
+        if "no status available for unknown model" in ex.message():
+            print("PASS: detected wrong model")
```
deadeyegoodwin: newline at end of file

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -107,198 +104,447 @@ def close(self):
...
         return response.status_code == 200
 
-    def is_server_ready(self):
+    def is_server_ready(self, headers={}):
         """Contact the inference server and get readiness.
 
+        Parameters
+        ----------
+        headers: dict
+            Optional dictionary specifying the headers that must be
```
deadeyegoodwin: "... specifying additional HTTP headers to include in the request."

src/clients/python/experimental_api_v2/library/httpclient.py
deadeyegoodwin: I agree. This class represents an output, so _name can be a "first class" member in the object. Same is true for other classes. Don't make a map to store member values, just store them directly.

src/clients/python/experimental_api_v2/library/httpclient.py
tanmayv25: Fixed

src/clients/python/experimental_api_v2/examples/simple_http_v2_health_metadata.py
tanmayv25: Fixed

CoderHam:
You haven't set the HTTP_V2 build flag  for the python httpclient in the Dockerfile.client

CoderHam:
@tanmayv25 the health bug was fixed in `http-v2-fix` and the inference segfault was fixed in `http-v2-binary`
I have tested your changes with these branches. Please re-run the CI as I have just added the HTTP_V2 build flag for the client build.

tanmayv25:
I will be enabling the L0_http_v2 tests in a separate PR.

PR #1207, opened 2020-03-20
Add API to release output buffer explicitly

GuanLuo:
In the case of having pre-allocate buffer pool, we want to release the buffer back to the pool once it is done.

### DIFF
--- a/src/core/trtserver2.h
+++ b/src/core/trtserver2.h
@@ -297,11 +297,11 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
     const void** base, size_t* byte_size, TRTSERVER_Memory_Type* memory_type,
     int64_t* memory_type_id);
 
-/// Clear all the output tensors. The meta data of the output tensors will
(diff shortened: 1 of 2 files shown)


src/core/trtserver2.h
```diff
@@ -297,6 +297,13 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputData(
     const void** base, size_t* byte_size, TRTSERVER_Memory_Type* memory_type,
     int64_t* memory_type_id);
 
+/// Clear all the output tensors. The meta data of the output tensors will
+/// become unaccesible and the result data will be released.
+/// \param inference_request The request object.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestClearAllOutputs(
```
deadeyegoodwin: Call this "...RemoveAll..." for consistency.
It is ok to add this API but in TRTSERVER2_ServerInferAsync we should be doing this when we PrepareForInference. Can you add a call to SetResponse(nullptr) there as well.

PR #1210, opened 2020-03-20
Update README and versions for 20.03 release

dzier:
Don't push until release


README.rst
```diff
@@ -27,16 +27,285 @@
 
 |License|
 
-NVIDIA TensorRT Inference Server
-================================
+NVIDIA Triton Inference Server
+==============================
 
-    **NOTE: You are currently on the r20.03 branch which tracks
```
deadeyegoodwin: Include note about the new name.

dzier: You want a note at the top here?

deadeyegoodwin: Look at the master README, there is bold text that starts with "NEW NAME", we want that included in this README as well.

dzier: Hold, let me fix that then

Issue #1204, opened 2020-03-19
C++ Image client for object detection


mhs-achyut:
**Is your feature request related to a problem? Please describe.**
I am in desperate need of a C++ image client for the object detection model. I tried following the C++ API documentation. But as a beginner to the entire InferContext Class and Result class, it is not obvious to write such a client quickly.
**Describe the solution you'd like**
I need a C++ client that can query an object detection model from the server and get multiple outputs such as bounding boxes, scores, and classes. This should be similar to the existing image_client.cc from trtis_clients.
**Describe alternatives you've considered**
I have considered building a python client. But my application is built in C++ so I desperately need help in building a C++ client.

**Additional context**
I can work with authors in helping them build an object detection model client. This would benefit the entire community and act as a placeholder for any object detection models in the future. So, I feel this is a worthwhile investment of time. Please let me know.

I have requested this a while ago in #217 as well.
Thanks a lot in advance.

deadeyegoodwin:
The client examples are meant to illustrate how the Python and C++ libraries are used. The image_client is nice because there are a lot of image classification models and they produce results that are generally easy-to-understand and interesting. But it is beyond the scope of the inference server project to provide clients for different application domains.

The inference server will soon announce that it is moving to a community standard REST and GRPC API for which there can be standard clients that can talk to any inference server that implements the standard: https://github.com/kubeflow/kfserving/tree/master/docs/predict-api/v2

So I think you goal of building a set of standard clients for different application domains is an interesting one, and I would encourage those clients to use the standard inferencing API... but the creation and hosting of those clients should likely belong to projects in those application domains.

mhs-achyut:
https://github.com/NVIDIA/triton-inference-server/pull/1212 has a starting point.

PR #1212, opened 2020-03-20
Object Detection Image Client

mhs-achyut:
This is derived from the image client. It takes in an image and outputs object detection tensors. Todo: Add NMS and visualization for the object detection outputs.


deadeyegoodwin:
As noted in #1204, Triton doesn't attempt to provide clients for all possible DL/ML models.

PR #1217, opened 2020-03-23
[GRPC V2] Sequence + Streaming API support for server and client 

tanmayv25:


src/clients/python/experimental_api_v2/examples/simple_grpc_v2_sequence_infer_client.py
```diff
@@ -0,0 +1,192 @@
...
+# three parameters are reserved for InferResult, InferenceServerException
+# and the sequence ID for the response.
+def completion_callback(user_data, result, error, sequence_id):
+    user_data._completed_requests.put((result, error, sequence_id))
+
+
+def async_send(triton_client, response_pool, values, batch_size, sequence_id,
+               user_data, model_name, model_version):
+    # Prepare the sequence metadata object
+    sequence_metadata = grpcclient.InferSequenceMetadata(
```
deadeyegoodwin: Just call the class InferSequence

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_sequence_infer_client.py
```diff
@@ -0,0 +1,192 @@
...
+
+    result0_list = []
+    result1_list = []
+
+    user_data_0 = UserData()
+    user_data_1 = UserData()
+
+    # Create the thread pool with the size as expected
+    # number of concurrent sequences.
+    response_pool = get_stream_response_processor_pool(2)
```
deadeyegoodwin: So if I want to have 1000 simultaneous sequences I need to have 1000 GRPC threads? Is that because GRPC requires a dedicated thread for each stream?

src/servers/grpc_server_v2.cc
```diff
@@ -319,7 +319,12 @@ class HandlerState {
     // The step of the entire context.
     Steps step_;
 
-    // True if this context should finish with OK status, false if
+    // The status of the requests in the stream. If OK then all the requests
+    // in the stream were successfull, otherwise stores the status of the
+    // first failing request in the stream.
+    grpc::Status request_status_;
```
deadeyegoodwin: Why do we need this? Why does the overall status handling for the stream have to change from what we have now? Even if individual requests in the stream fail that doesn't necessarily impact the overall stream status.

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_sequence_infer_client.py
```diff
@@ -0,0 +1,192 @@
...
+        value_data = np.full(shape=[batch_size, 1],
+                             fill_value=value,
+                             dtype=np.int32)
+        inputs = []
+        inputs.append(grpcclient.InferInput('INPUT'))
+        # Initialize the data
+        inputs[0].set_data_from_numpy(value_data)
+        outputs = []
+        outputs.append(grpcclient.InferOutput('OUTPUT'))
+        sequence_metadata.add_request(inputs,
```
deadeyegoodwin: We don't want to have to put requests into a sequence before sending them. The 'sequence' object manages the lifetime of the GRPC stream (is stream is being used) but not the lifetime of requests. So async_sequence_infer should take same arguments as async_infer plus the sequence (required).

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -635,102 +661,93 @@ def wrapped_callback(call_future):
...
             self._call_future = self._client_stub.ModelInfer.future(request)
             self._call_future.add_done_callback(wrapped_callback)
         except grpc.RpcError as rpc_error:
             raise_error_grpc(rpc_error)
 
-    def _get_inference_request(self, inputs, outputs, model_name, model_version,
-                               request_id, parameters):
-        """Creates and initializes an inference request.
+    def async_sequence_infer(self,
+                             sequence_metadata,
```
deadeyegoodwin: Just "sequence". Should the response_pool just be associated with the sequence? I think use_streaming should be set in the sequence.

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_sequence_infer_client.py
deadeyegoodwin: I'd like to understand this more and how it compares to what our current GRPC client does. Seems like perhaps this pool should be associated with a sequence then and should instead be a "global" in the InferenceServerClient. I will set up a meeting tomorrow so we can discuss.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -635,102 +661,94 @@ def wrapped_callback(call_future):
...
             will be used.
-        parameters: dict
-            Optional inference parameters described as key-value pairs.
-
-        Returns
-        -------
-        ModelInferRequest
-            The protobuf message holding the inference request.
-        
+        end_of_sequence: bool
```
deadeyegoodwin: I'm thinking about how we know when to close the GRPC stream. Does reseting the sequence do that? We should discuss this as well.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1007,3 +1025,261 @@ def get_response(self, as_json=False):
...
+        by the object.
+
+        Returns
+        -------
+        int
+            The ID of the current sequence
+        """
+        return self._id
+
+    def sequence_delivered(self):
```
deadeyegoodwin: I don't think you can know this... new requests can be sent any time.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1174,3 +1239,188 @@ def get_response(self, as_json=False):
...
+        blocks till response of all currently enqueued requests are not
+        received.
+        """
+        if self._is_initialized():
+            self._request_queue.put(None)
+            if self._handler.is_alive():
+                self._handler.join()
+            self._handler = None
+
+    def set_headers(self, headers):
```
deadeyegoodwin: Is this really necessary? There is already a way to set headers for individual requests, why also on the stream?

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1174,3 +1239,188 @@ def get_response(self, as_json=False):
...
+        Parameters
+        ----------
+        headers: dict
+            Optional dictionary specifying additional HTTP
+            headers to include while establising gRPC stream.
+        """
+        self.close()
+        self._headers = headers
+
+    def set_callback(self, new_callback):
```
deadeyegoodwin: Why allow the callback to be changed? Seems like an unnecessary complication.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -812,92 +887,82 @@ def wrapped_callback(call_future):
         except grpc.RpcError as rpc_error:
             raise_error_grpc(rpc_error)
 
-    def _get_inference_request(self, inputs, outputs, model_name, model_version,
-                               request_id, parameters):
-        """Creates and initializes an inference request.
+    def async_stream_infer(self,
+                           stream,
```
deadeyegoodwin: model_name, inputs, stream, ...

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -741,43 +790,58 @@ def infer(self,
...
         ----------
         callback : function
-            Python function that is invoked once the request is completed.
-            The function must reserve the last two arguments (result, error)
-            to hold InferResult and InferenceServerException objects
-            respectively which will be provided to the function when executing
-            the callback. The ownership of these objects will be given to the
-            user. The 'error' would be None for a successful inference.
+            Python function that is invoked upon receiving response from
+            the underlying stream. The function must have last argument
```
deadeyegoodwin: There is no underlying stream, is there?

src/core/grpc_service_v2.proto
```diff
@@ -778,6 +788,59 @@ message ModelInferResponse
...
+  Status status = 1;
+
+  //@@
+  //@@  .. cpp:var:: ModelInferResponse infer_response
+  //@@
+  //@@     Holds the results of the request.
+  //@@
+  ModelInferResponse infer_response = 2;
+
+  //@@
```
deadeyegoodwin: sequence_id is independent of streaming so I don't think we want it here. I think it should be in the response parameters. Let me look at the sequence extension doc and see...

src/servers/grpc_server_v2.cc
```diff
@@ -336,12 +336,17 @@ class HandlerState {
     context_ = context;
     unique_id_ = RequestStatusUtil::NextUniqueRequestId();
     step_ = start_step;
+    sequence_id_ = 0;
```
deadeyegoodwin: I don't think the GRPC endpoint needs to be aware of the sequence ID. It will be in the request and response objects and completely managed by Triton core.

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_sequence_sync_infer_client.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python
```
deadeyegoodwin: Should there also be a simple_grpc_v2_sequence_async_infer_client.py?

src/core/grpc_service_v2.proto
```diff
@@ -778,6 +788,59 @@ message ModelInferResponse
...
+  //@@     The status of the request.
+  //@@
+  message Status
+  {
+    //@@
+    //@@    .. cpp:var:: int32 code
+    //@@
+    //@@       The status code
+    //@@
+    int32 code = 1;
```
deadeyegoodwin: Where are these codes defined? Do we really need codes?

src/core/grpc_service_v2.proto
```diff
@@ -778,6 +788,59 @@ message ModelInferResponse
...
+    //@@
+    string message = 2;
+  }
+
+  //@@
+  //@@  .. cpp:var:: Status status
+  //@@
+  //@@     Holds the status of the request.
+  //@@
+  Status status = 1;
```
deadeyegoodwin: How is success vs failure handled? Need to document that.

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_sequence_sync_infer_client.py
deadeyegoodwin: OK, don't bother then.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1174,3 +1231,158 @@ def get_response(self, as_json=False):
...
+        blocks till response of all currently enqueued requests are not
+        received.
+        """
+        if self._is_initialized():
+            self._request_queue.put(None)
+            if self._handler.is_alive():
+                self._handler.join()
+            self._handler = None
+
+    def set_headers(self, headers):
```
deadeyegoodwin: OK, I'll merge this as is, but it seems to me that we should just have headers in the stream constructor just live we do for the infer methods. What is the use case for having a separate function?

PR #1220, opened 2020-03-24
HTTP V2 url parameters in Python client

tanmayv25:


src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -106,6 +106,62 @@ def close(self):
         """
         self._client_stub.close()
 
+    def _get(self, request_uri, headers):
```
deadeyegoodwin: I like the breakout of get and post, but having the query parameters specified as part of the host:port is somewhat strange. Also, it binds the same set of query parameters to every transaction from the client object, which is too restrictive. Instead I think you want to treat it just like the "headers" arg and allow a "query_params" dict as an optional argument in all those same places.

tanmayv25: Done

PR #1223, opened 2020-03-24
Rename /opt/tensorrtserver to /opt/tritonserver.

deadeyegoodwin:


CoderHam:
Could you file a parallel PR for similar changes on GitLab? Or do we want those to use  `/opt/tensorrtserver` for now?

PR #1227, opened 2020-03-25
[HTTP V2] Enable simple testing with Python client

CoderHam:
- [x] server health (live/ready)
- [x] server metadata
- [x] model ready
- [x] model metadata
- [x] infer (only JSON, including nested)
- [x] async infer (only JSON, including nested)

Will add Shared Memory, Model Control, Infer Binary support once these features are added to the httpclient library

### DIFF
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -79,6 +79,7 @@ RUN mkdir -p qa/common && \
     cp -r /workspace/docs/examples/model_repository/simple qa/L0_grpc_v2/models && \
     cp -r /workspace/docs/examples/model_repository/simple_string qa/L0_grpc_v2/models && \
     cp -r /workspace/docs/examples/model_repository/inception_graphdef qa/L0_grpc_v2/models && \
+    mkdir qa/L0_http_v2/models && \
     cp -r /workspace/docs/examples/model_repository/simple qa/L0_http_v2/models && \
     cp /opt/tritonserver/bin/simple qa/L0_simple_lib/. && \
     cp /opt/tritonserver/bin/simplev2 qa/L0_simple_lib/. && \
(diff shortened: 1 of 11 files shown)


src/clients/python/experimental_api_v2/examples/simple_grpc_v2_health_metadata.py
```diff
@@ -55,32 +55,45 @@
     model_name = 'simple'
 
     # Health
-    if triton_client.is_server_live():
-        print("PASS: is_server_live")
+    if not triton_client.is_server_live():
+        print("FAILED : is_server_live")
+        sys.exit(1)
```
tanmayv25: You would have change L0_grpc_v2/test.sh to handle these changes. Currently, it looks for 7 "PASS" messages.

Dockerfile.QA
```diff
@@ -79,7 +79,6 @@ RUN mkdir -p qa/common && \
     cp -r /workspace/docs/examples/model_repository/simple qa/L0_grpc_v2/models && \
     cp -r /workspace/docs/examples/model_repository/simple_string qa/L0_grpc_v2/models && \
     cp -r /workspace/docs/examples/model_repository/inception_graphdef qa/L0_grpc_v2/models && \
-    mkdir qa/L0_http_v2/models && \
```
tanmayv25: Are you sure this is not needed?

src/clients/python/experimental_api_v2/examples/simple_http_v2_infer_client.py
```diff
@@ -80,20 +80,19 @@
 
     print(results.get_response())
 
-    # FIXME: Uncomment when DLIS 1162 is fixed.
-    # Get the output arrays from the
-    # output0_data = results.as_numpy('OUTPUT0')
-    # output1_data = results.as_numpy('OUTPUT1')
+    # Get the output arrays from the results
```
tanmayv25: Can you add a test which uses a wrong model_name to send out inference?

Dockerfile.QA
CoderHam: You're right. It is needed.

src/clients/python/experimental_api_v2/examples/simple_http_v2_infer_client.py
CoderHam: Sure

PR #1232, opened 2020-03-26
[HTTP V2] Classification Support for Inference

CoderHam:


src/core/trtserver2.h
```diff
@@ -268,6 +268,33 @@ TRTSERVER_EXPORT TRTSERVER_Error* TRTSERVER2_InferenceRequestOutputDataType(
     TRTSERVER2_InferenceRequest* inference_request, const char* name,
     const char** datatype);
 
+/// Get the batch_size of an output tensor.
+/// \param inference_request The request object.
+/// \param name The name of the output.
+/// \param batch_size Returns the number of batches of the output.
+/// \return a TRTSERVER_Error indicating success or failure.
+TRTSERVER_EXPORT TRTSERVER_Error*
+TRTSERVER2_InferenceRequestOutputClassBatchSize(
```
deadeyegoodwin: We don't want to add these methods. How is classification result being handled in the grpc_v2 server? Eventually we will fix triton server core to handle everything internally as per the V2 spec...

CoderHam: > We don't want to add these methods. How is classification result being handled in the grpc_v2 server? Eventually we will fix triton server core to handle everything internally as per the V2 spec...

In grpc_server_v2 we use the response_header object directly to extract this information. The same is not available inside the http_server based on the V2 API design.

GuanLuo: Why the `InferenceRequestOutputShape` is not sufficient?

PR #1237, opened 2020-03-27
[HTTP V2] Model Configuration Support

CoderHam:


src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -780,9 +780,9 @@ def unregister_cuda_shared_memory(self,
 
     def infer(self,
               inputs,
+              outputs,
```
tanmayv25: 'outputs' is an optional field. According to specification if the 'outputs' are not provided then results for all output tensor of the model should be returned in default setting.

src/clients/python/experimental_api_v2/examples/simple_http_v2_infer_client.py
```diff
@@ -44,8 +44,8 @@ def test_infer(model_name, input0_data, input1_data):
     outputs.append(httpclient.InferOutput('OUTPUT1'))
     query_params = {'test_1': 1, 'test_2': 2}
     results = triton_client.infer(inputs,
-                              model_name,
-                            outputs=outputs,
+                            outputs,
+                            model_name,
```
tanmayv25: 'outputs' should remain optional.

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
```diff
@@ -0,0 +1,275 @@
...
+    requests = []
+    results = []
+
+    # Send request
+    for inputs, outputs, FLAGS.model_name, FLAGS.model_version in requestGenerator(
+            input_name, output_name, c, h, w, format, dtype, FLAGS):
+        try:
+            results.append(
+                triton_client.infer(inputs,
+                                  outputs,
```
tanmayv25: Here as well.

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
```diff
@@ -0,0 +1,275 @@
...
+    parser.add_argument('-u',
+                        '--url',
+                        type=str,
+                        required=False,
+                        default='localhost:8000',
+                        help='Inference server URL. Default is localhost:8000.')
+    parser.add_argument('image_filename', type=str, help='Input image.')
+    FLAGS = parser.parse_args()
+
+    # Create gRPC client for communicating with the server
```
deadeyegoodwin: http

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
```diff
@@ -0,0 +1,275 @@
...
+
+    repeated_image_data = [image_data for _ in range(FLAGS.batch_size)]
+    batched_image_data = np.stack(repeated_image_data, axis=0)
+
+    # Set the input data
+    inputs[0].set_data_from_numpy(batched_image_data)
+
+    outputs = []
+    outputs.append(httpclient.InferOutput(output_name))
+    outputs[0].set_parameter("classification", 2)
```
deadeyegoodwin: Why don't you use the classes argument from the command-line flags?

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
```diff
@@ -0,0 +1,275 @@
...
+
+    yield inputs, outputs, FLAGS.model_name, FLAGS.model_version
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
```
deadeyegoodwin: This seems like the image_client example that has not really been simplified by also does not really use all the still-existing image client options? Why don't you make a full-fledged image-client example that uses the http V2 protocol? Shouldn't it mostly (completely) just be the grpc_v2_image_client.py with an option to use HTTP and a couple of minor changes? Or course rename that file to v2_image_client.py

src/clients/python/experimental_api_v2/library/httpclient.py
CoderHam: Will make these changes. Currently the server returns only outputs for the outputs mentioned in the request header

CoderHam: Adding a ticket to track this

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
CoderHam: Sure I can go ahead and do this. I was simply following the design for the simple_grpc_*_client(s). I agree that on the same basis the simple_grpc_v2_class_client can also be removed and we can have a single v2 image client as you said.

CoderHam: Solved by #1244

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_class_client.py
```diff
@@ -176,7 +177,8 @@ def requestGenerator(input_name, output_name, c, h, w, format, dtype, FLAGS):
     inputs[0].set_data_from_numpy(batched_image_data)
 
     outputs = []
-    outputs.append(grpcclient.InferOutput(output_name, class_count=2))
+    outputs.append(grpcclient.InferOutput(output_name))
+    outputs[0].set_parameter("classification", FLAGS.classes)
```
tanmayv25: There has been some changes in the library API. You must set the 'class_count' in the InferOutput constructor. See  PR #1242

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
```diff
@@ -0,0 +1,269 @@
...
+
+    repeated_image_data = [image_data for _ in range(FLAGS.batch_size)]
+    batched_image_data = np.stack(repeated_image_data, axis=0)
+
+    # Set the input data
+    inputs[0].set_data_from_numpy(batched_image_data)
+
+    outputs = []
+    outputs.append(httpclient.InferOutput(output_name))
+    outputs[0].set_parameter("classification", 2)
```
tanmayv25: Should be set in the constructor.

CoderHam: Done

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_class_client.py
CoderHam: Done

src/clients/python/experimental_api_v2/examples/simple_grpc_v2_class_client.py
```diff
@@ -32,6 +32,7 @@
 from builtins import range
```
deadeyegoodwin: Should just delete these two class clients... they seem to be just be the same as or a subset of the v2_image_client.

CoderHam: These two class clients verify the results of the classification are indeed correct. The v2_image_client.py simply runs the inference but does not verify the results.

deadeyegoodwin: But you can/should have a CI that runs v2_image_client and verifies it's result. When we shift from V1->V2 we have several existing image_client using CI tests that will provide this coverage.

CoderHam: I will make this change in the image_client PR. Including dropping the simple_grpc/http_v2_class_client.py

CoderHam:
Thanks. Already done in the image client PR

src/clients/python/experimental_api_v2/examples/simple_http_v2_class_client.py
```diff
@@ -0,0 +1,269 @@
...
+    repeated_image_data = [image_data for _ in range(FLAGS.batch_size)]
+    batched_image_data = np.stack(repeated_image_data, axis=0)
+
+    # Set the input data
+    inputs = []
+    inputs.append(httpclient.InferInput(input_name, batched_image_data.shape, dtype))
+    inputs[0].set_data_from_numpy(batched_image_data)
+
+    outputs = []
+    outputs.append(httpclient.InferOutput(output_name, class_count=FLAGS.classes))
```
tanmayv25, line 171: You would have to use "binary_data=False" here and any 'set_data_from_numpy' calls. Have a look at #1253. Once DLIS-1204 is fixed all the example clients can enable binary data.

CoderHam: Done

PR #1238, opened 2020-03-27
[GRPC V2] Model Statistics support

tanmayv25:


src/core/grpc_service_v2.proto
```diff
@@ -184,6 +184,16 @@ service GRPCInferenceService
...
   {
   }
+
+  //@@  .. cpp:var:: rpc ModelStatistics(
+  //@@                     ModelStatisticsRequest)
+  //@@                   returns (ModelStatisticsResponse)
+  //@@
+  //@@     Get the cumulative inference statistics for a model.
+  //@@
+  rpc ModelStatistics(ModelStatisticsRequest) returns (ModelStatisticsResponse)
```
deadeyegoodwin: Can you mode this up to be just after ModelConfig, and move the request and response messages similarly.

tanmayv25: Done

PR #1242, opened 2020-03-30
Hiding explicit setting of parameters in V2 clients

tanmayv25:


src/clients/python/experimental_api_v2/examples/simple_grpc_v2_cudashm_client.py
```diff
@@ -112,22 +113,20 @@
...
-    inputs[-1].set_parameter("shared_memory_byte_size", input_byte_size)
+    inputs.append(grpcclient.InferInput('INPUT1'))
+    inputs[-1].set_data_from_shared_memory("input1_data", input_byte_size,
+                                           [1, 16], "INT32")
 
     outputs = []
     outputs.append(grpcclient.InferOutput('OUTPUT0'))
-    # outputs[-1].set_parameter("shared_memory_region", "output0_data")
-    # outputs[-1].set_parameter("shared_memory_byte_size", output_byte_size)
+    # outputs[-1].use_shared_memory("output0_data", output_byte_size)
```
deadeyegoodwin: Why is this commented out?

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1010,22 +1010,13 @@ class InferInput:
...
-    shape : list
-        The shape of the associated input. Default value is None.
-    datatype : str
-        The datatype of the associated input. Default is None.
+        The name of input whose data will be described by this object.
 
     """
 
-    def __init__(self, name, shape=None, datatype=None):
+    def __init__(self, name):
```
deadeyegoodwin: Why do you remove shape and datatype. Every input must explicitly specify a shape and datatype and the constructor is a natural place to do this. We want to make these required so don't provide a default.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1077,36 +1068,39 @@ def set_data_from_numpy(self, input_tensor):
         else:
             self._input.contents.raw_contents = input_tensor.tobytes()
 
-    def set_parameter(self, key, value):
-        """Adds the specified key-value pair in the requested input parameters
+    def set_data_from_shared_memory(self, region_name, byte_size, shape,
```
deadeyegoodwin: set_shared_memory() ?
Remove shape and datatype as they should be in constructor.
Where is offset? It can default to 0 but it should be there and set the corresponding parameter.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1077,36 +1068,39 @@ def set_data_from_numpy(self, input_tensor):
...
+        self._input.parameters[
+            'shared_memory_region'].string_param = region_name
+        self._input.parameters[
+            'shared_memory_byte_size'].int64_param = byte_size
+
+        self._input.ClearField('shape')
+        self._input.shape.extend(shape)
+        self._input.datatype = datatype
+
+    def reset(self):
```
deadeyegoodwin: Why a reset function? If they want to reuse the request object can't they just call use_shared_memory again with different values? If they want bigger changes (like changing the shape), then they just have to create a new input from scratch.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1143,36 +1137,43 @@ def name(self):
         """
         return self._output.name
 
-    def set_parameter(self, key, value):
-        """Adds the specified key-value pair in the requested output parameters
+    def mark_classification(self, count=1):
```
deadeyegoodwin: I would put this in the constructor with a default value of 0.

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1143,36 +1137,43 @@ def name(self):
...
+        Parameters
+        ----------
+        count : int
+            The number of classifications to be returned in result.
+            Default value is 1.
+
+        """
+        self._output.parameters['classification'].int64_param = count
+
+    def use_shared_memory(self, region_name, byte_size):
```
deadeyegoodwin: set_shared_memory()
offset?

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -1143,36 +1137,43 @@ def name(self):
...
-            raise_error("unsupported value type for the parameter")
 
-    def clear_parameters(self):
-        """Clears all the parameters that have been added to the output request.
+        self._output.parameters[
+            'shared_memory_region'].string_param = region_name
+        self._output.parameters[
+            'shared_memory_byte_size'].int64_param = byte_size
+
+    def reset(self):
```
deadeyegoodwin: Remove

PR #1243, opened 2020-03-30
 [GRPC V2] Scheduling Policy Extension

tanmayv25:
This PR is just for collecting reviews. The base branch needs to changed to master when #1242 is merged to the master.

### DIFF
--- a/src/clients/python/experimental_api_v2/library/grpcclient.py
+++ b/src/clients/python/experimental_api_v2/library/grpcclient.py
@@ -61,23 +61,20 @@ def _get_inference_request(model_name,
     request = grpc_service_v2_pb2.ModelInferRequest()
     request.model_name = model_name
     request.model_version = model_version
-    if request_id != None:
+    if request_id != "":
         request.id = request_id
     for infer_input in inputs:
         request.inputs.extend([infer_input._get_tensor()])
     for infer_output in outputs:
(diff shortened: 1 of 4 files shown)


src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -53,9 +53,11 @@ def _get_inference_request(model_name,
...
-                           sequence_start=None,
-                           sequence_end=None):
+                           sequence_id,
+                           sequence_start,
+                           sequence_end,
+                           priority,
+                           timeout):
     request = grpc_service_v2_pb2.ModelInferRequest()
     request.model_name = model_name
     request.model_version = model_version
```
deadeyegoodwin: if request_id is not None

src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -66,14 +68,15 @@ def _get_inference_request(model_name,
     for infer_output in outputs:
         request.outputs.extend([infer_output._get_tensor()])
     if sequence_id:
```
deadeyegoodwin: Can you be explicit with these conditionals, use "is not None", != 0, etc. as appropriate.

deadeyegoodwin:
@tanmayv25 rebase onto master and we can merge

PR #1244, opened 2020-03-31
Add v2_image_client.py example

CoderHam:
Replicate image_client.py with V2 protocol
Follow up in upcoming PRs: streaming, folders as input instead of image, async


deadeyegoodwin:
@CoderHam Do you have a CI for this?

CoderHam:
> @CoderHam Do you have a CI for this?

Started one

src/clients/python/experimental_api_v2/examples/CMakeLists.txt
```diff
@@ -57,3 +57,9 @@ if(${TRTIS_ENABLE_GRPC_V2})
     DESTINATION python
   )
 endif() # TRTIS_ENABLE_GRPC_V2
+
+install(
```
deadeyegoodwin, line 61: Should only install if V2 GRPC || V2 HTTP

CoderHam:
> Does this need a rebase? I don't see the class clients being delete here.

Yes. Rebased now.

PR #1250, opened 2020-03-31
[GRPC V2] Include model_name in InferResponse

tanmayv25:


src/core/grpc_service_v2.proto
```diff
@@ -656,7 +656,8 @@ message ModelInferRequest
   //@@
   //@@  .. cpp:var:: InferRequestedOutputTensor outputs (repeated)
   //@@
-  //@@     The requested output tensors for the inference.
+  //@@     The requested output tensors for the inference. Optional, if not
+  //@@     specified all outputs produced by the model will be returned.
```
GuanLuo: "all outputs specified in model config will be returned"? User can expose only a subset of the model outputs in model config.

tanmayv25: Good point.
@deadeyegoodwin we must update the language on our KFserving page.

PR #1251, opened 2020-03-31
Fix crash when NV-InferRequest not in InferRequestHeader

CoderHam:
Error message: "The request header should contain 'NV-InferRequest'"
Error Code: 400


deadeyegoodwin:
@CoderHam Can you add a test to some appropriate CI for this...

PR #1253, opened 2020-04-01
[HTTP V2] Bianry data extension support from client side

tanmayv25:
The examples explicitly disables the use of binary data because of some issues on the server side (DLIS-1204). Once the issue is fixed these "binary_data=False" settings can be removed.

### DIFF
--- a/src/clients/python/experimental_api_v2/library/httpclient.py
+++ b/src/clients/python/experimental_api_v2/library/httpclient.py
@@ -31,6 +31,7 @@ from urllib.parse import quote, quote_plus
 import rapidjson as json
 import numpy as np
 import gevent.pool
+import sys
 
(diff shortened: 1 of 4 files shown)


src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -1079,21 +1100,26 @@ def datatype(self):
...
         input associated with this object.
 
         Parameters
         ----------
         input_tensor : numpy array
             The tensor data in numpy array format
+        binary_data : bool
+            Indicates whether to set data for the input in binary format
+            or explicit tensor within JSON. The default value is True,
+            which means the data will be delivered to the server in the
```
deadeyegoodwin, line 1114: ".. data will be delivered as binary data..."

PR #1257, opened 2020-04-01
[HTTP V2] String support for JSON I/O, Return error for FP16

CoderHam:


src/servers/http_server_v2.cc
```diff
@@ -501,54 +501,88 @@ HTTPAPIServerV2::ResponseRelease(
...
+        case TYPE_FP64: {
+          double* data_vec = reinterpret_cast<double*>(base->data());
           data_vec[*counter] = (double)payload_data[i].GetDouble();
           break;
+        }
+        case TYPE_STRING: {
+          const char* cstr = payload_data[i].GetString();
+          uint32_t len = strlen(cstr);
+          int base_len = base->size();
+          base->resize(base_len + sizeof(uint32_t) + len);
```
GuanLuo, line 570: `resize()` is potentially expensive.. i.e. filling new space, and copies due to reallocations. Maybe better to do two pass on `payload_data`, first for getting total byte size and the second for the copies.

src/servers/http_server_v2.cc
```diff
@@ -501,54 +501,88 @@ HTTPAPIServerV2::ResponseRelease(
...
-        case TYPE_FP64:
+        }
+        case TYPE_FP64: {
+          double* data_vec = reinterpret_cast<double*>(base->data());
           data_vec[*counter] = (double)payload_data[i].GetDouble();
           break;
+        }
+        case TYPE_STRING: {
+          const char* cstr = payload_data[i].GetString();
+          uint32_t len = strlen(cstr);
```
GuanLuo, line 568: Shouldn't we be aware of the case where the content is not a C string? Say the serialized protobuf scenario. And if the rapidjson library be able to handle such case properly?

src/servers/http_server_v2.cc
```diff
@@ -501,54 +501,88 @@ HTTPAPIServerV2::ResponseRelease(
 template <typename T>
```
GuanLuo, line 501: template is not needed

src/servers/http_server_v2.cc
```diff
@@ -653,24 +648,33 @@ void
...
-      response_output_val->PushBack(data_val, allocator);
-      *counter += 1;
+      if (dtype != TYPE_STRING) {
+        rapidjson::Value data_val((T)(base[*counter]));
+        response_output_val->PushBack(data_val, allocator);
+        *counter += 1;
+      } else {
+        uint32_t* len = reinterpret_cast<uint32_t*>(base + *counter);
+        char* cstr =
+            reinterpret_cast<char*>(base + *counter + sizeof(uint32_t));
```
GuanLuo, line 669: So the returning data will not contain the length bytes?

src/servers/http_server_v2.cc
```diff
@@ -653,24 +648,33 @@ void
 WriteDataToJsonHelper(
```
GuanLuo, line 648: Drop template

src/servers/http_server_v2.cc
```diff
@@ -2083,8 +2096,11 @@ HTTPAPIServerV2::InferRequestClass::FinalizeResponse(
         if (!CheckSharedMemoryData(
                 request_output, &shm_region, &offset, &byte_size)) {
           // Write outputs into json array (if not shared memory)
-          WriteDataToJson(
+          err = WriteDataToJson(
               output_metadata[i], allocator, const_cast<void*>(base));
+          if (err != nullptr) {
```
GuanLuo, line 2101: Delete error

src/servers/http_server_v2.cc
CoderHam: @deadeyegoodwin can confirm this but we do not intend on supporting this through JSON. The same shall only be supported via the binary format method.

src/servers/http_server_v2.cc
CoderHam: The returning data will be an array (nested) of strings without their length bytes.

src/servers/http_server_v2.cc
deadeyegoodwin: Looking at www.json.org it actually looks like any byte array can likely be encoded in a json string using appropriate escapes. I would think that the rapidjson encoder/decoder should handle it. We will need to test thoroughly

src/servers/http_server_v2.cc
CoderHam: Why? This template needed

src/servers/http_server_v2.cc
CoderHam: In the case that JSON encodes the byte array as a json string (assuming this is base64 encoded) it will still be null terminated as well right? 
Additionally if it is base64 encoded, we should have a flag to denote the same correct? So that we can decode it before passing to the backend.

deadeyegoodwin: Nothing to do with base64. From the spec that I pointed to it appears that you can escape characters in the json string, so you should be able to escape a '\0' as well. You can feed some arbitrary byte arrays to radpijson and see how it encodes them into json.

src/servers/http_server_v2.cc

CoderHam:
> fp16 error looks ok in input side but don't we need to fail if the model produces fp16 data and the output is not requested in binary format?

We do return a failure when the model produces fp16 data and the output is not requested in binary format: 

https://github.com/NVIDIA/triton-inference-server/blob/hemantj-v2-json-cleanup/src/servers/http_server_v2.cc#L764-L773

> Also the string/bytes handling still needs updating if we can just escape characters in the string.

Working on the string/bytes handling. There was an issue with the client library that needed fixing first. It is currently unable to send requests with string inputs successfully to the server.

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -897,8 +898,10 @@ def infer(self,
                                                priority=priority,
                                                timeout=timeout)
 
+        print("infer_request: " + str(infer_request))
         request_body = json.dumps(infer_request)
-        json_size = sys.getsizeof(request_body)
+        print("request_body: ", request_body)
```
deadeyegoodwin, line 903: Remove debug prints and other debugging stuff...

CoderHam:
> Can you add some tests to the v2 string client that send bytes that contain special characters, including the null characters. You can use an identify models to check that the values git passed in/out correctly.

The string support from the client side is broken. Filed a ticket for the same (DLIS-1211) I will circle back to finish this once @tanmayv25 fixes this JSON and binary support for strings. 
The issue appears to be with rapidjson not being able to send the post request successfully when the request header has a json string array.

PR #1259, opened 2020-04-01
[HTTP V2] Fix check for shared memory type in GetStatusV2

CoderHam:


GuanLuo:
Can you add some tests for using shared memory with HTTP V2... That is not the only problematic places, I just pointed it out as example.

PR #1263, opened 2020-04-02
Remove use of ServerStatus from tritonserver.cc

deadeyegoodwin:


deadeyegoodwin:
> I am adding ModelStatistics API which still uses `lserver->GetStatus` to collect stats, should I add an endpoint in `server.h` to return the stats directly?

I don't think you can just yet... the ServerStatus class needs to be streamlined to drop the non-stats parts (which is being done in this PR) and then you will have an lserver->GetStatistics method that you can use. For now it is ok to keep using GetStatus.

PR #1267, opened 2020-04-03
Update server frontends to use Triton server API when api_version is set to v2

GuanLuo:
See tritionserver.h for API changes for supporting classification and for fixing output shape constraint.

### DIFF
--- a/src/core/model_repository_manager.cc
+++ b/src/core/model_repository_manager.cc
@@ -296,10 +296,10 @@ class ModelRepositoryManager::BackendLifeCycle {
   // Get the VersionStateMap representation of the specified model.
   const VersionStateMap GetVersionStates(const std::string& model_name);
 
-  // The the state of a specific model version.
+  // Get the states of all recorded model version.
   Status GetModelState(
-      const std::string& model_name, const int64_t model_version,
-      ModelReadyState* state);
+      const std::string& model_name,
+      std::map<int64_t, ModelReadyState>* states);
 
  private:
   struct BackendInfo {
@@ -512,25 +512,21 @@ ModelRepositoryManager::BackendLifeCycle::GetVersionStates(
 
 Status
 ModelRepositoryManager::BackendLifeCycle::GetModelState(
-    const std::string& model_name, const int64_t model_version,
-    ModelReadyState* state)
+    const std::string& model_name, std::map<int64_t, ModelReadyState>* states)
 {
   std::lock_guard<std::mutex> map_lock(map_mtx_);
   auto mit = map_.find(model_name);
   if (mit != map_.end()) {
-    auto vit = mit->second.find(model_version);
-    if (vit != mit->second.end()) {
-      const auto& backend_info = vit->second;
+    for (const auto& version_state : mit->second) {
+      const auto& backend_info = version_state.second;
       std::lock_guard<std::recursive_mutex> lock(backend_info->mtx_);
-      *state = backend_info->state_;
-      return Status::Success;
+      (*states)[version_state.first] = backend_info->state_;
     }
+    return Status::Success;
   }
 
   return Status(
-      Status::Code::NOT_FOUND, "model '" + model_name + "', version " +
-                                   std::to_string(model_version) +
-                                   " is not found");
+      Status::Code::NOT_FOUND, "model '" + model_name + "' is not found");
 }
 
 Status
@@ -1377,10 +1373,9 @@ ModelRepositoryManager::GetVersionStates(const std::string& model_name)
 
 Status
 ModelRepositoryManager::GetModelState(
-    const std::string& model_name, const int64_t model_version,
-    ModelReadyState* state)
+    const std::string& model_name, std::map<int64_t, ModelReadyState>* states)
 {
-  return backend_life_cycle_->GetModelState(model_name, model_version, state);
+  return backend_life_cycle_->GetModelState(model_name, states);
 }
 
 Status
--- a/src/core/model_repository_manager.cc
+++ b/src/core/model_repository_manager.cc
@@ -1389,7 +1389,7 @@ ModelRepositoryManager::GetInferenceBackend(
     backend->reset();
     status = Status(
         Status::Code::UNAVAILABLE,
-        "Inference request for unknown model '" + model_name + "'");
+        "Request for unknown model '" + model_name + "'");
   }
   return status;
 }
--- a/src/core/model_repository_manager.cc
+++ b/src/core/model_repository_manager.cc
@@ -296,10 +296,10 @@ class ModelRepositoryManager::BackendLifeCycle {
   // Get the VersionStateMap representation of the specified model.
   const VersionStateMap GetVersionStates(const std::string& model_name);
 
-  // Get the states of all recorded model version.
+  // Get the state of a specific model version.
   Status GetModelState(
-      const std::string& model_name,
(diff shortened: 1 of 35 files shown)


src/core/tritonserver.cc
```diff
@@ -85,7 +98,6 @@ TritonServerError::Create(
     ni::RequestStatusCode status_code, const std::string& msg)
 {
   // If 'status_code' is success then return nullptr as that indicates
-  // success
```
deadeyegoodwin, line 88: You probably didn't mean to delete this

src/core/provider.cc
```diff
@@ -78,6 +78,14 @@ AddClassResults(
       }
 
       cls->set_value(static_cast<float>(probs[idx[k]]));
+
```
deadeyegoodwin, line 81: Don't we need to conditionalize how we generate the class results based on the protocol?

src/core/tritonserver.cc
```diff
@@ -1548,6 +1676,22 @@ TRITONSERVER_InferenceRequestSetRequestedOutputClassificationCount(
   return nullptr;  // Success
 }
 
+TRITONSERVER_Error*
+TRITONSERVER_InferenceRequestRequestedOutputClassificationCount(
```
deadeyegoodwin, line 1680: Why do we need these special classification methods. Triton should just produce the output tensor as a string tensor and then the endpoint code can access it like any other tensor.

src/core/provider.cc
GuanLuo: Will add condition on protocol version. Right now it prepares class results for both versions...

src/core/tritonserver.cc
GuanLuo: So the TRITONSERVER_OutputShape / Data / Datatype APIs should be used for both raw and class outputs? Then the checking on whether the output is requested as class results will be moved inside those TRITONSERVER APIs?

GuanLuo:
@CoderHam I fixed some HTTP v2 clients and the test script, you should take a look and check if that makes sense.

src/core/tritonserver.cc
deadeyegoodwin: Yes. Someone on the "outside" of the TRITONSERVER APIs should just use the same APIs to read the output tensors... but if they requested class results then they will get an output that is the one-dim string tensor instead of what the model configuration specifies as the actual output type/shape.

src/core/model_repository_manager.cc
```diff
@@ -296,10 +296,10 @@ class ModelRepositoryManager::BackendLifeCycle {
   // Get the VersionStateMap representation of the specified model.
   const VersionStateMap GetVersionStates(const std::string& model_name);
 
-  // The the state of a specific model version.
+  // Get the states of all recorded model version.
```
deadeyegoodwin, line 299: Does't this make it the same as "GetVersionStates", which is not what is needed for where it is used.

src/core/server.cc
```diff
@@ -340,13 +340,19 @@ InferenceServer::ModelIsReady(
...
   ScopedAtomicIncrement inflight(inflight_request_counter_);
 
-  std::shared_ptr<InferenceBackend> backend;
-  RETURN_IF_ERROR(GetInferenceBackend(model_name, model_version, &backend));
-
-  ModelReadyState state;
-  RETURN_IF_ERROR(model_repository_manager_->GetModelState(
-      model_name, backend->Version(), &state));
-  *ready = (state == ModelReadyState::MODEL_READY);
+  std::map<int64_t, ModelReadyState> states;
```
deadeyegoodwin, line 343: backend->Version() gives us the actual version, doesn't it? So we don't have to worry about -1. I think the original code is correct.

GuanLuo: `GetInferenceBackend()` returns error if the backend is unloaded / unavailable. I guess in that case we can just return success and set ready to false. Then we can use the original code.

src/servers/http_server_v2.cc
```diff
@@ -2145,131 +1924,73 @@ HTTPAPIServerV2::InferRequestClass::FinalizeResponse(
...
-          class_array.PushBack(class_str, allocator);
-          count++;
-        }
-        batch_class_array.PushBack(class_array, allocator);
+      uint64_t offset = 0, byte_size = 0;
+      const char* shm_region = nullptr;
+      if (!CheckSharedMemoryData(
+              request_output, &shm_region, &offset, &byte_size)) {
+        // Write outputs into json array (if not shared memory)
+        WriteDataToJson(output_metadata[i], allocator, const_cast<void*>(base));
```
CoderHam, line 1981: Where is the handling for classification output?

src/core/server.cc
deadeyegoodwin: Sounds good.

src/servers/http_server_v2.cc
GuanLuo: It's move inside the server core, the frontend server won't be aware of it and will see cls outputs the same way as regular output.
See discussion
https://github.com/NVIDIA/triton-inference-server/pull/1267#discussion_r403347170

src/core/tritonserver.h
```diff
@@ -633,6 +633,19 @@ TRITONSERVER_InferenceRequestSetRequestedOutputClassificationCount(
...
+/// Get the classification count of the requested output.
+/// \param inference_request The request object.
+/// \param name The name of the output.
+/// \param count Indicates how many classification values will be
+/// returned for the output. The 'count' highest priority values are
+/// returned. A value of 0 indicates that the output tensor
+/// will not be returned as a classification.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_EXPORT TRITONSERVER_Error*
+TRITONSERVER_InferenceRequestRequestedOutputClassificationCount(
```
deadeyegoodwin, line 645: Why is this needed? The output shape indicates how many classifications there are.

src/core/tritonserver.h
```diff
@@ -1004,6 +1016,20 @@ TRITONSERVER_EXPORT TRITONSERVER_Error* TRITONSERVER_ServerModelMetadata(
     TRITONSERVER_Server* server, const char* model_name,
     const char* model_version, TRITONSERVER_Message** model_metadata);
 
+/// Get the statistics of the model being served as a TRITONSERVER_Message
+/// object. The caller takes ownership of the object and must call
+/// TRITONSERVER_MessageDelete to release the object.
+/// \param server The inference server object.
+/// \param model_name The name of the model to get metadata for.
```
deadeyegoodwin, line 1023: Update param descriptions

src/core/tritonserver.h
```diff
@@ -1004,12 +1003,26 @@ TRITONSERVER_EXPORT TRITONSERVER_Error* TRITONSERVER_ServerModelMetadata(
     TRITONSERVER_Server* server, const char* model_name,
     const char* model_version, TRITONSERVER_Message** model_metadata);
 
+/// Get the statistics of the model being served as a TRITONSERVER_Message
+/// object. The caller takes ownership of the object and must call
+/// TRITONSERVER_MessageDelete to release the object.
+/// \param server The inference server object.
+/// \param model_name The name of the model to get statistics for.
+/// \param model_version The version of the model to get readiness for.
```
deadeyegoodwin, line 1011: This comment needs updating, no need to run CI just let me know when it is updated.

GuanLuo: Updated

PR #1268, opened 2020-04-03
[HTTP V2] Add System + CUDA shared memory clients and testing 

CoderHam:
Separated shared memory and cuda shared memory from grpc into combined sharedmemoryutils library. 
Fixes to client side.
@tanmayv25 please have a look at the 'data' field. The follow should be missing for inputs and outputs in the case of shared memory.

### DIFF
--- a/src/clients/python/experimental_api_v2/examples/CMakeLists.txt
+++ b/src/clients/python/experimental_api_v2/examples/CMakeLists.txt
@@ -32,6 +32,8 @@ if(${TRTIS_ENABLE_HTTP_V2})
       simple_http_v2_health_metadata.py
       simple_http_v2_infer_client.py
       simple_http_v2_async_infer_client.py
+      simple_http_v2_cudashm_client.py
+      simple_http_v2_shm_client.py
       simple_http_v2_model_control.py
     DESTINATION python
   )
--- a/src/clients/python/experimental_api_v2/examples/simple_http_v2_cudashm_client.py
+++ b/src/clients/python/experimental_api_v2/examples/simple_http_v2_cudashm_client.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import argparse
+import numpy as np
+import os
+import sys
+from builtins import range
+import tritonhttpclient.core as httpclient
+import tritongrpcclient.cuda_shared_memory as cudashm
+from ctypes import *
+
+FLAGS = None
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v',
+                        '--verbose',
+                        action="store_true",
+                        required=False,
+                        default=False,
+                        help='Enable verbose output')
+    parser.add_argument('-u',
+                        '--url',
+                        type=str,
+                        required=False,
+                        default='localhost:8000',
+                        help='Inference server URL. Default is localhost:8000.')
+
+    FLAGS = parser.parse_args()
+
+    try:
+        triton_client = httpclient.InferenceServerClient(FLAGS.url)
+    except Exception as e:
+        print("channel creation failed: " + str(e))
+        sys.exit(1)
+
+    # To make sure no shared memory regions are registered with the
+    # server.
+    triton_client.unregister_system_shared_memory()
+    triton_client.unregister_cuda_shared_memory()
+
+    # We use a simple model that takes 2 input tensors of 16 integers
+    # each and returns 2 output tensors of 16 integers each. One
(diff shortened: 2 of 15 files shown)


src/clients/python/experimental_api_v2/examples/simple_http_v2_cudashm_client.py
```diff
@@ -0,0 +1,161 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import argparse
+import numpy as np
+import sys
+from builtins import range
+import tritonhttpclient.core as httpclient
+import tritongrpcclient.cuda_shared_memory as cudashm
```
tanmayv25, line 33: Users may choose to only install tritonhttpclient and still want to use shared memory. This should be a separate common utility package.

src/clients/python/experimental_api_v2/examples/CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ if(${TRTIS_ENABLE_HTTP_V2})
       simple_http_v2_health_metadata.py
       simple_http_v2_infer_client.py
       simple_http_v2_async_infer_client.py
+      simple_grpc_v2_cudashm_client.py
+      simple_grpc_v2_shm_client.py
```
GuanLuo, line 36: grpc or http?

src/clients/python/experimental_api_v2/examples/simple_http_v2_shm_client.py
```diff
@@ -0,0 +1,163 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import argparse
+import numpy as np
+import sys
+from builtins import range
+import tritonhttpclient.core as httpclient
+import tritongrpcclient.shared_memory as shm
+from ctypes import *
```
GuanLuo, line 34: `ctypes` is not used?

src/clients/python/experimental_api_v2/examples/simple_http_v2_cudashm_client.py
```diff
@@ -0,0 +1,161 @@
...
+
+    outputs.append(httpclient.InferOutput('OUTPUT1'))
+    # Uncomment with DLIS-1202
+    # outputs[-1].set_shared_memory("output1_data", output_byte_size)
+
+    results = triton_client.infer(model_name=model_name,
+                                  inputs=inputs,
+                                  outputs=outputs)
+
+    # TODO [DLIS-1202] : Currently, this example doesn't use shared memory
```
GuanLuo, line 134: Check #1265 for how to get (CUDA) shared memory data as numpy array.

src/servers/http_server_v2.cc
```diff
@@ -2277,6 +2282,7 @@ HTTPAPIServerV2::InferRequestClass::FinalizeResponse(
     response_meta_data_.response_json_.Accept(writer);
     const char* response_metadata = buffer.GetString();
     size_t json_length = strlen(response_metadata);
+    LOG_VERBOSE(1) << response_metadata;
```
GuanLuo, line 2285: Remove debug message.

src/servers/http_server_v2.cc
```diff
@@ -1887,13 +1889,14 @@ HTTPAPIServerV2::EVBufferToInput(
...
-              "memory");
-        }
+        // Uncomment after client fix (Don't send data in case of shared memory)
+        // if (output.FindMember("data") == output.MemberEnd()) {
+        //   return TRTSERVER_ErrorNew(
+        //       TRTSERVER_ERROR_INVALID_ARG,
+        //       "must not specify 'data' field in request output when using "
+        //       "shared "
+        //       "memory");
+        // }
```
GuanLuo, line 1899: Can be simply removed, requested output doesn't have "data" field anyway

src/servers/http_server_v2.cc
```diff
@@ -1812,12 +1813,13 @@ HTTPAPIServerV2::EVBufferToInput(
       const char* shm_region = nullptr;
       if (CheckSharedMemoryData(
               request_input, &shm_region, &offset, &byte_size)) {
-        if (request_input.FindMember("data") == request_input.MemberEnd()) {
-          return TRTSERVER_ErrorNew(
-              TRTSERVER_ERROR_INVALID_ARG,
-              "must not specify 'data' field in request input when using "
-              "shared memory");
-        }
+        // Uncomment after client fix (Don't send data in case of shared memory)
```
GuanLuo, line 1816: Shouldn't the client fix be included in this PR?

src/servers/http_server_v2.cc
```diff
@@ -1812,12 +1813,13 @@ HTTPAPIServerV2::EVBufferToInput(
...
+        // if (request_input.FindMember("data") == request_input.MemberEnd()) {
+        //   return TRTSERVER_ErrorNew(
+        //       TRTSERVER_ERROR_INVALID_ARG,
+        //       "must not specify 'data' field in request input when using "
+        //       "shared memory");
```
GuanLuo, line 1821: The conditional conflicts with the error message. If `Find()` returns `End()`, then "data" is not specified.

src/clients/python/experimental_api_v2/library/CMakeLists.txt
```diff
@@ -141,3 +141,29 @@ if(${TRTIS_ENABLE_GRPC_V2})
...
+add_custom_target(
+  client-wheel-shared-memory ALL
+  DEPENDS
+    "${shared_memory_wheel_stamp_file}"
+)
+
+install(
+  CODE "file(GLOB _Wheel \"${CMAKE_CURRENT_BINARY_DIR}/shared_memory/tritonsharedmemoryutils-*.whl\")"
+  CODE "file(INSTALL \${_Wheel} DESTINATION \"${CMAKE_INSTALL_PREFIX}/python\")"
+)
```
GuanLuo, line 169: new line

src/servers/http_server_v2.cc
CoderHam: Tanmay is making some fixes to the client. This will be a part of those fixes.

src/servers/http_server_v2.cc
CoderHam: This is to verify that the infer request header does not contain 'data' fields. 
The docs specify that in the case of shared memory: the “data” field of request_input must not be set and the returned $response_output must not define the “data” field.

src/servers/http_server_v2.cc
CoderHam: This is to verify that the infer request header does not contain 'data' fields. 
The docs specify that in the case of shared memory: the “data” field of request_input must not be set and the returned $response_output must not define the “data” field.

src/servers/http_server_v2.cc
GuanLuo: You are checking the "output" in the request object which shouldn't have "data" field at all https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md#request-output

GuanLuo: For output, the doc is referring to the response JSON message. You can add the check there if you want a sanity check, or just be sure that the "data" field will not be added in shared memory case.

src/servers/http_server_v2.cc
GuanLuo: Check again when will the conditional be evaluated to true. Do you intend to return error in that case?

src/servers/http_server_v2.cc
CoderHam: Fixed the check

tanmayv25:
This [commit ](https://github.com/NVIDIA/triton-inference-server/pull/1268/commits/0ba7e82c9f248bba598a181cbc9e0c3ff4528008) fixed the build issue you were seeing. You must also fix the simple_grpc_v2 clients to use the new modules.

CoderHam:
> This [commit ](https://github.com/NVIDIA/triton-inference-server/pull/1268/commits/0ba7e82c9f248bba598a181cbc9e0c3ff4528008) fixed the build issue you were seeing. You must also fix the simple_grpc_v2 clients to use the new modules.

Done

src/servers/http_server_v2.cc
CoderHam: Fixed

src/servers/http_server_v2.cc
GuanLuo: I think the fix is merged #1278

PR #1279, opened 2020-04-07
[Don't Merge] HTTP V2 C++ clients: health and metadata

tanmayv25:


src/clients/c++/experimental_api_v2/library/grpc_client.h
```diff
@@ -0,0 +1,140 @@
...
+  /// \param model_name The name of the model to check for readiness.
+  /// \param headers Map specifying additional HTTP headers to include in the
+  /// metadata of gRPC request.
+  /// \param model_version The version of the model to check for readiness.
+  /// The default value is an empty string which means then the server will
+  /// choose a versionbased on the model and internal policy.
+  /// \return Error object indicating success or failure of the request.
+  Error IsModelReady(
+      bool* ready, const std::string& model_name,
+      std::map<std::string, std::string>& headers,
```
deadeyegoodwin, line 88: Following python APIs it would be better to have model_version after model_name, and default it to empty string. Also, headers is going to be rarely used to forcing create of an empty map to call the API is awkward. Likely want to have 2 IsModelReady functions, one that takes headers and one that doesn't. Similarly for other functions.

tanmayv25: I have made headers optional too. As the headers would be rarely used, I think we can have single signature. The only thing is that in order to specify headers, user must specify model_version too. Let me know what you think.

PR #1280, opened 2020-04-07
Adding string tests for HTTP V2

tanmayv25:


src/clients/python/experimental_api_v2/examples/simple_http_v2_model_control.py
```diff
@@ -56,7 +56,7 @@
     model_name = 'simple'
 
     # There are two models in the repository directory
```
GuanLuo, line 58: Update comment

src/clients/python/experimental_api_v2/examples/simple_http_v2_string_infer_client.py
```diff
@@ -74,10 +74,10 @@
     input1_data = in1n.reshape(in1.shape)
 
     # Initialize the data
-    inputs[0].set_data_from_numpy(input0_data, binary_data=False)
+    inputs[0].set_data_from_numpy(input0_data, binary_data=True)
```
deadeyegoodwin, line 77: Should use the identity model to verify that strings that contain special characters and '\0' are handled correctly.

tanmayv25: Done.

src/clients/python/experimental_api_v2/examples/simple_http_v2_string_infer_client.py
```diff
@@ -103,5 +103,23 @@
...
+    
+    # Test with an identity model
+    model_name = "savedmodel_zero_1_object"
+    inputs.clear()
+    outputs.clear()
+    inputs.append(httpclient.InferInput('INPUT0', [1, 16], "BYTES"))
+
+    null_chars_array = np.array([u'he\0llo' for i in range(16)], dtype=object)
+    null_char_data = null_chars_array.reshape([1, 16])
+    inputs[0].set_data_from_numpy(null_char_data, binary_data=False)
```
deadeyegoodwin, line 115: Test with both binary_data true and false.

src/clients/python/experimental_api_v2/examples/simple_http_v2_string_infer_client.py
```diff
@@ -103,5 +103,23 @@
...
             print("error: incorrect difference")
             sys.exit(1)
+    
+    # Test with an identity model
+    model_name = "savedmodel_zero_1_object"
+    inputs.clear()
+    outputs.clear()
+    inputs.append(httpclient.InferInput('INPUT0', [1, 16], "BYTES"))
+
+    null_chars_array = np.array([u'he\0llo' for i in range(16)], dtype=object)
```
deadeyegoodwin, line 113: Does this need to be \x00 in python 3? https://python-reference.readthedocs.io/en/latest/docs/str/escapes.html

Also should add some tests of just binary arrays being passed as BYTES type for both binary_data true and false. For example just make a byte array of [ 1 ,2, 3, 0, ... ] and put it into a numpy array... did we resolve how byte arrays were best represented in numpy, I recall that there were serveral options.

src/servers/http_server_v2.cc
```diff
@@ -626,7 +625,7 @@ ReadDataFromJsonHelper(
         }
         case TYPE_STRING: {
           const char* cstr = payload_data[i].GetString();
-          uint32_t len = strlen(cstr);
```
CoderHam, line 629: If you notice the server uses strlen to get the size of the infer response JSON header. In the case the output contains a nullchar, won't the response header returned be incomplete?

tanmayv25: It's the other way around.
strlen takes the null character to mark the termination of a string. So, the length of string "he\0llo" will be 2. GetStringLength() on the other hand will hold the entire filed length of JSON. Look here for more information under "Query String" header.
https://rapidjson.org/md_doc_tutorial.html

src/clients/python/experimental_api_v2/examples/simple_http_v2_string_infer_client.py
tanmayv25: There are several different encoding possible. Similar to API V1 we are exposing the users to specify numpy array of dtype.type=np.bytes_. This will internally traslate to type "|S<byte_size>". 
@CoderHam Is there any other way to simplify the process?

PR #1288, opened 2020-04-08
Add InferenceResponse class

deadeyegoodwin:
The InferenceResponse class represent the result of an inference
execution, including any output tensor produced by the execution.


deadeyegoodwin:
This is going to be a fairly large and disruptive change so merging into the 'response' branch instead of master until it gets stabilized.

PR #1293, opened 2020-04-09
[HTTP V2] Make Outputs specification optional in Inference Request

CoderHam:
- Defaults to returning all outputs from model config. 
@tanmayv25 I believe I fixed the same for GRPC V2 as well

### DIFF
--- a/src/core/infer_request.h
+++ b/src/core/infer_request.h
@@ -311,6 +311,7 @@ class InferenceRequest {
 
   // Remove a single requested output or all requested outputs.
   Status RemoveRequestedOutput(const std::string& name);
+  Status OutputNames(const char** name_vec, uint64_t* name_count);
   Status RemoveAllRequestedOutputs();
 
   // Prepare this request for inference. We pass backend here as
--- a/src/core/infer_request.h
+++ b/src/core/infer_request.h
@@ -311,7 +311,6 @@ class InferenceRequest {
 
   // Remove a single requested output or all requested outputs.
   Status RemoveRequestedOutput(const std::string& name);
-  Status OutputNames(const char** name_vec, uint64_t* name_count);
   Status RemoveAllRequestedOutputs();
 
   // Prepare this request for inference. We pass backend here as
--- a/src/core/tritonserver.cc
+++ b/src/core/tritonserver.cc
@@ -1614,14 +1614,27 @@ TRITONSERVER_InferenceRequestAddRequestedOutput(
 }
 
 TRITONSERVER_Error*
-TRITONSERVER_InferenceRequestGetAllOutputs(
-    TRITONSERVER_InferenceRequest* inference_request, const char** names,
-    uint64_t* name_count)
+TRITONSERVER_InferenceRequestOutputCount(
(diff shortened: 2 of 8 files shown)


src/core/tritonserver.h
```diff
@@ -667,6 +667,16 @@ TRITONSERVER_InferenceRequestOutputShape(
     TRITONSERVER_InferenceRequest* inference_request, const char* name,
     const int64_t** shape, uint64_t* dim_count);
 
+/// Get the names of all output tensors.
```
GuanLuo, line 670: "... in the result" to be more accurate. The function is actually returning the name of the requested outputs

src/core/tritonserver.h
```diff
@@ -667,6 +667,16 @@ TRITONSERVER_InferenceRequestOutputShape(
...
+/// \param inference_request The request object.
+/// \param names The names of the outputs.
```
GuanLuo, line 672: Specify that the names is owned by the function caller and must be pre-allocated with enough space.
However, that is unsafe, for instance, the HTTP frontend implementation below assumes output counts <= 10 and segfault will happen otherwise. Should either let infer request object own the names, or create it in place and let function caller to release it on finish.

src/core/tritonserver.cc
```diff
@@ -1613,6 +1613,18 @@ TRITONSERVER_InferenceRequestAddRequestedOutput(
   return nullptr;  // Success
 }
 
+TRITONSERVER_Error*
```
deadeyegoodwin, line 1616: So here we have a real deficiency in the current API. There is no way to discover what output tensors are available. But we need an API that keeps ownership of the names with the InferenceRequest object so we don't have to worry about managing that. I suggest we add a TRITONSERVER_InferenceRequestOutputCount to return the number of output tensors (notice how we don't use "Get" in the API naming to be consistent with other existing API), and a TRITONSERVER_InferenceRequestOutputName(uint3_t index, const char** name) that returns the name of the output at that index.

src/servers/http_server_v2.cc
```diff
@@ -1959,80 +1963,139 @@ HTTPAPIServerV2::InferRequestClass::FinalizeResponse(
...
+          WriteDataToJson(
+              output_metadata[i], allocator, const_cast<void*>(base));
+        }
       }
+
+      response_outputs.PushBack(output_metadata[i], allocator);
     }
+    response_json.AddMember("outputs", response_outputs, allocator);
+  } else {
+    // Get outputs from model config (Assume max is 10)
```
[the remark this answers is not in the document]
CoderHam: Fixed

src/core/tritonserver.cc
CoderHam: Done

src/core/infer_request.h
```diff
@@ -311,6 +311,7 @@ class InferenceRequest {
 
   // Remove a single requested output or all requested outputs.
   Status RemoveRequestedOutput(const std::string& name);
+  Status OutputNames(const char** name_vec, uint64_t* name_count);
```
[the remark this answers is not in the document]
CoderHam: Done

PR #1296, opened 2020-04-10
GRPC V2 C++ client Inference support

tanmayv25:
Support for Asynchronous and Streaming API will be added with other PR.


src/clients/c++/experimental_api_v2/library/CMakeLists.txt
```diff
@@ -37,12 +37,12 @@ if(${TRTIS_ENABLE_GRPC_V2})
...
   set(
     REQUEST_SRCS
-    grpc_client.cc utils.cc
+    grpc_client.cc common_utils.cc grpc_utils.cc
   )
 
   set(
     REQUEST_HDRS
-    grpc_client.h utils.h
+    grpc_client.h common_utils.h grpc_utils.h base_utils.h
```
deadeyegoodwin, line 45: common_utils and base_utils? Why split?

src/clients/c++/experimental_api_v2/library/base_utils.h
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin, line 1: So there are some common classes between HTTP and GRPC... but they are not really utils are they? Can they just be in a common.h/.cc?

src/clients/c++/experimental_api_v2/library/common_utils.h
```diff
@@ -72,4 +72,52 @@ class DECLSPEC Error {
   std::string msg_;
 };
 
+//==============================================================================
+/// Structure to hold options for Inference Request.
```
deadeyegoodwin, line 76: Seems like these can be combined into common.h

src/clients/c++/experimental_api_v2/library/grpc_client.cc
```diff
@@ -249,10 +281,53 @@ InferenceServerGrpcClient::GetModelConfig(
   return err;
 }
 
+Error
+InferenceServerGrpcClient::Infer(
+    std::shared_ptr<InferResultGrpc>* result, const InferOptions& options,
```
deadeyegoodwin, line 286: Our use of unique_ptr and shared_ptr in the interfaces isn't really needed is it? Are we enforcing some allocation policy that we don't need to? Do we keep copies of result, input or output? If not perhaps we should pass by const& (or const* in the vectors), and return via pointer. The caller can then wrap the raw pointers in shared/unique if they want. What do you think?

src/clients/c++/experimental_api_v2/library/CMakeLists.txt
tanmayv25: Fixed.

src/clients/c++/experimental_api_v2/library/base_utils.h
tanmayv25: Fixed.

src/clients/c++/experimental_api_v2/library/common_utils.h
tanmayv25: Fixed

src/clients/c++/experimental_api_v2/library/grpc_client.cc
tanmayv25: You are right. We must not enforce any policy in our API level. Fixed,

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,250 @@
...
+/// An interface for InferInput object to describe the model input for
+/// inference.
+///
+class InferInput {
+ public:
+  /// Get the name of input associated with this object
+  /// \param name Returns the name of input.
+  /// \return Error object indicating success or failure of the
+  /// request.
+  virtual Error GetName(std::string* name) const = 0;
```
deadeyegoodwin, line 134: Elsewhere we don't use "Get" style naming. We would just say Name() and SetName(). Can we follow that same convention in the C++ API.

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,250 @@
...
+
+  /// Set the tensor data from the specified buffer represented
+  /// as vector.
+  /// \param input The vector holding input data.
+  /// \return Error object indicating success or failure of the
+  /// request.
+  virtual Error SetRaw(const std::vector<uint8_t>& input) = 0;
+
+  /// Set the tensor data from the specified vector of strings, where
+  /// each string represents an lement of the "BYTES" tensor ordered
```
deadeyegoodwin, line 172: [e]lement

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,250 @@
...
+  virtual Error SetSharedMemory(
+      const std::string& region_name, const size_t byte_size,
+      const size_t offset = 0) = 0;
+};
+
+//==============================================================================
+/// An interface for InferOutput object to describe the requested model
+/// output for inference.
+///
+class InferOutput {
```
deadeyegoodwin, line 194: InferRequestedOutput to be more explicit?

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,250 @@
...
+  virtual Error SetSharedMemory(
+      const std::string& region_name, const size_t byte_size,
+      const size_t offset = 0) = 0;
+};
+
+//==============================================================================
+/// An interface for InferResult object which allows to access and
+/// interpret the response from inference request.
+///
+class InferResult {
```
deadeyegoodwin, line 218: InferOutput? And in the description make sure to emphasis that this represents an output tensor in the response. As it is now is sounds like it is the entire response.

src/clients/c++/experimental_api_v2/library/grpc_utils.h
```diff
@@ -0,0 +1,181 @@
...
+  Error SetFromString(const std::vector<std::string>& input) override;
+
+  /// See InferInput::SetSharedMemory(
+  ///   const std::string& region_name, const size_t byte_size,
+  ///   const size_t offset)
+  Error SetSharedMemory(
+      const std::string& region_name, const size_t byte_size,
+      const size_t offset = 0) override;
+
+  /// Returns the unferlying InferInputTensor message.
```
deadeyegoodwin, line 81: underlying

src/clients/c++/experimental_api_v2/library/common.h
tanmayv25: InferResult object does point to the entire response. The users while obtaining the data for a particular output has to provide the output name and the pointer to the buffer in the message will be returned. It is similar to our python side.

src/clients/c++/experimental_api_v2/library/common.cc
```diff
@@ -0,0 +1,316 @@
...
+// PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#define DLL_EXPORTING
```
deadeyegoodwin, line 27: What is this for?

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,561 @@
...
+  /// \return The name of the input
+  const std::string& Name() const { return name_; }
+
+  /// Gets the size of data set into this input in bytes.
+  /// \return The size of data in bytes.
+  int64_t ByteSize() const { return byte_size_; }
+
+  /// Gets the size of data which will be send on the wire.
+  /// \return The size of data to be sent on wire in bytes.
+  size_t TotalSendByteSize() const { return total_send_byte_size_; }
```
deadeyegoodwin, line 307: Why is this needed and why is it different than ByteSize()?

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,561 @@
...
+  /// Prepare this input to receive new tensor values. Forget any
+  /// existing values that were set by previous calls to SetSharedMemory()
+  /// or SetRaw().
+  /// \return Error object indicating success or failure.
+  Error Reset();
+
+  /// Set tensor values for this input from a byte vector. The vector
+  /// is not copied and so it must not be modified or destroyed
+  /// until this input is no longer needed (that is until the Run()
+  /// call(s) that use the input have completed). For batched inputs
```
deadeyegoodwin, line 338: For batched inputs it would be nice to not have to say that it must be called once for each batch... really we just want them to be able to provide input in any types of chunks they want. It there anything in the implementation that relies on it being called exactly batch-size times?

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,561 @@
...
+  Error SharedMemoryInfo(
+      std::string* name, size_t* batch_byte_size, size_t* offset) const;
+
+  // Prepare to send this input as part of a request.
+  Error PrepareForRequest();
+
+  // Copy into 'buf' up to 'size' bytes of this input's data. Return
+  // the actual amount copied in 'input_bytes' and if the end of input
+  // is reached in 'end_of_input'
+  Error GetNext(
```
deadeyegoodwin, line 406: Why do we need to provide public methods to get the data out of an input? The user put the data in so they should already have it.

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,561 @@
...
+  /// \param byte_size The size of data in bytes.
+  /// \param offset The offset in shared memory region. Default value is 0.
+  /// \return Error object indicating success or failure of the
+  /// request.
+  Error SetSharedMemory(
+      const std::string& region_name, const size_t byte_size,
+      const size_t offset = 0);
+
+  // Copy the shared memory key, offset and batch_byte_size
+  Error SharedMemoryInfo(
```
deadeyegoodwin, line 489: Is this the inverse of SetSharedMemory? Why do we need to provide it? In general think about making the API as minimal as possible.

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -0,0 +1,561 @@
...
+  virtual Error Id(std::string* id) const = 0;
+
+  /// Get the shape of output result returned in the response.
+  /// \param output_name The name of the ouput to get shape.
+  /// \param shape Returns the shape of result for specified output name.
+  /// \return Error object indicating success or failure.
+  virtual Error Shape(
+      const std::string& output_name, std::vector<int64_t>* shape) const = 0;
+
+
```
deadeyegoodwin, line 535: remove extra blank line

src/clients/c++/experimental_api_v2/library/common.h
tanmayv25: Removed. It was carry over from our current V1. It is different in sense total_send_byte_size_ will be 0 for shared_memory inputs. Will add later if the need arises.

src/clients/c++/experimental_api_v2/library/common.cc
tanmayv25: It was used for decltype specification for Error class in common.h. I have removed for now. Will add when the need arise when enabling L0_win10 with V2 API.

PR #1306, opened 2020-04-14
Catch improper JSON and throw appropriate error

CoderHam:


src/servers/http_server_v2.cc
```diff
@@ -1691,10 +1781,10 @@ HTTPAPIServerV2::EVBufferToInput(
             irequest, input_name, base, byte_size, memory_type,
             memory_type_id));
       } else {
-        const rapidjson::Value& shape = request_input["shape"];
-        const char* dtype_str = request_input["datatype"].GetString();
+        // const rapidjson::Value& shape = request_input["shape"];
```
tanmayv25, line 1784: These lines can be removed as we are getting shape and datatype above.

src/servers/http_server_v2.cc
```diff
@@ -1254,11 +1254,34 @@ HTTPAPIServerV2::HandleSystemSharedMemory(
...
-              region_name, shm_key, offset, byte_size);
+          const auto& key_itr = register_request.FindMember("key");
+          if (key_itr == register_request.MemberEnd()) {
+            err = TRITONSERVER_ErrorNew(
+                TRITONSERVER_ERROR_INVALID_ARG,
+                "Shared memory register request has no 'key' field");
+          } else {
+            const char* shm_key = key_itr->value.GetString();
+            uint64_t offset = 0, byte_size = 0;
+            const auto& offset_itr = register_request.FindMember("offset");
```
tanmayv25, line 1265: > The “shared_memory_offset” parameter is optional and defaults to zero. The other two parameters are required. If only one of the two is given the server should return an appropriate error. 

From the specification. It looks like the shared_memory_offset should be optional and server should not throw the error in case offset is absent.

CoderHam: Thanks for catching this. Fixed this

PR #1323, opened 2020-04-17
Fix L0_storage_S3 test

CoderHam:


GuanLuo:
Why moving the test to elsewhere, I don't think it uses models that is not public accessible.

CoderHam:
> Why moving the test to elsewhere, I don't think it uses models that is not public accessible.

Cannot expose S3 credentials here.

GuanLuo:
The secrets can be set as CI environment variables, right?

CoderHam:
> Does this test both ways of using S3?
> Also you need to update the title since it still refers to gitlab.

The second way of testing by passing S3 ENV vars will be added with the follow up ticket and support for ENV vars

PR #1325, opened 2020-04-17
[GRPC V2 ]  Update gRPC Streaming API more intuitive 

tanmayv25:
There is a 1:1 correspondence with stream and client. These changes avoid confusion.

### DIFF
--- a/src/clients/python/experimental_api_v2/library/grpcclient.py
+++ b/src/clients/python/experimental_api_v2/library/grpcclient.py
@@ -30,6 +30,7 @@ import grpc
 import rapidjson as json
 import threading
 import queue
+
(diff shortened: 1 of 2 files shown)


src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -967,7 +1015,9 @@ def async_stream_infer(self,
                            priority=0,
                            timeout=None):
         """Runs an asynchronous inference over gRPC bi-directional streaming
-        API.
+        API. A stream should be established with a call to start_stream()
```
deadeyegoodwin, line 1018: "...stream [must] be...

PR #1330, opened 2020-04-17
Minimize when request normalization is needed

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo This is on a side branch but I followed up with some simplifications to your change to make sure we are only doing normalization when absolutely necessary.

PR #1333, opened 2020-04-17
Restore CUDA device only in backends that require safety

GuanLuo:


GuanLuo:
It seems to be the case. the L0_io test is testing various I/O device combinations and it is passing, so I think other backends are not sensitive to current device.

The responsibility shift sounds good.

PR #1343, opened 2020-04-21
Fix warmup feature in general case

GuanLuo:
Can't run the corresponding L0_warmup test yet as the backends in the tests are not updated with the request / response changes.


deadeyegoodwin:
@GuanLuo Did you try running L0_warmup with just the custom backend? Is that possible?

deadeyegoodwin:
That would probably be good to cover all backends...

PR #1347, opened 2020-04-21
More performant management of serialized input data

deadeyegoodwin:


deadeyegoodwin:
@tanmayv25 Small cleanup on the side branch but thought you should take a look

PR #1355, opened 2020-04-22
L0_infer and infer utilities for V2 testing

CoderHam:


qa/L0_infer/infer_test.py
```diff
@@ -48,9 +48,9 @@ def _full_exact(self, input_dtype, output0_dtype, output1_dtype,
         def _infer_exact_helper(tester, pf, tensor_shape, batch_size,
                 input_dtype, output0_dtype, output1_dtype,
                 output0_raw=True, output1_raw=True,
-                model_version=None, swap=False,
+                model_version="", swap=False,
```
deadeyegoodwin, line 51: We know that for Triton the version has to be an integer. So it would probably be better to keep the interface here as an integer and just convert it to a string before using the client API.

qa/L0_infer/test.sh
```diff
@@ -70,15 +70,16 @@ BACKENDS=${BACKENDS:="graphdef savedmodel netdef onnx libtorch plan custom"}
 # If ENSEMBLES not specified, set to 1
 ENSEMBLES=${ENSEMBLES:="1"}
 
-for TARGET in cpu gpu; do
+# for TARGET in cpu gpu; do
+for TARGET in cpu; do
```
deadeyegoodwin, line 74: Is gpu not working at all?

qa/L0_infer/test.sh
```diff
@@ -157,7 +158,7 @@ for TARGET in cpu gpu; do
               sed -i "s/dims: \[ 1 \]/dims: \[ -1, -1 \]/" config.pbtxt)
     fi
 
-    run_server
+    run_server_v2
```
deadeyegoodwin, line 161: Change the utilities so there is only the v2 versions... and rename them to not have _v2 suffix.

qa/L0_infer/test.sh
```diff
@@ -176,13 +177,6 @@ for TARGET in cpu gpu; do
         RET=1
     fi
 
-    grep -c "HTTP/1.1 200 OK" $CLIENT_LOG
```
deadeyegoodwin, line 179: Keep this?

qa/common/infer_util.py
```diff
@@ -52,30 +61,27 @@ def _range_repr_dtype(dtype):
...
-    for input_value in input_values:
-        input_list.append(serialize_string_tensor(input_value))
-    return input_list
 
 # Perform inference using an "addsum" type verification backend.
 def infer_exact(tester, pf, tensor_shape, batch_size,
                 input_dtype, output0_dtype, output1_dtype,
                 output0_raw=True, output1_raw=True,
-                model_version=None, swap=False,
+                model_version="", swap=False,
```
deadeyegoodwin, line 69: Keep model_version as integer

qa/common/infer_util.py
```diff
@@ -52,30 +61,27 @@ def _range_repr_dtype(dtype):
...
                 outputs=("OUTPUT0", "OUTPUT1"), use_http=True, use_grpc=True,
-                skip_request_id_check=False, use_streaming=True,
+                use_json=True, skip_request_id_check=False, use_streaming=True,
```
deadeyegoodwin, line 71: use_http_json_tensors would be more descriptive

qa/common/infer_util.py
```diff
@@ -84,474 +90,309 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
-                                            for x in (op1.flatten())], dtype=object).reshape(op1.shape))
-        else:
-            expected1_list.append(op1.astype(output1_dtype))
-
-        if input_dtype == np.object:
-            in0n = np.array([str(x) for x in in0.reshape(in0.size)], dtype=object)
-            in0 = in0n.reshape(in0.shape)
-            in1n = np.array([str(x) for x in in1.reshape(in1.size)], dtype=object)
-            in1 = in1n.reshape(in1.shape)
+    if (pf == "batch_to_nobatch") or (pf == "nobatch_to_batch") or (pf == "mix_nobatch_batch"):
```
deadeyegoodwin, line 106: There seem to be a lot of changes here in input data preparation. Is this all due to how inputs are now a single numpy array for the entire batch?

qa/L0_infer/infer_test.py
CoderHam: Fixed

qa/L0_infer/test.sh
CoderHam: Typo. Fixed now.

qa/L0_infer/test.sh
CoderHam: Done

qa/L0_infer/test.sh
CoderHam: Added back

qa/common/infer_util.py
CoderHam: Done

qa/common/infer_util.py
GuanLuo: Why not making `infer_exact()` accept the actual shape as the parameter, then it is the caller's responsibility to provide proper shape, and it is more intuitive in respect to v2 spec.

qa/common/infer_util.py
```diff
@@ -84,474 +90,313 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
         else:
-            results = ctx.run(
-                    { INPUT0 : input0_list, INPUT1 : input1_list },
-                    output_req, batch_size,
-                    priority=priority, timeout_us=timeout_us)
+            results = triton_client.infer(model_name,
+                                        inputs,
+                                        model_version=model_version,
+                                        outputs=output_req,
+                                        request_id=str(_unique_request_id()))
```
deadeyegoodwin, line 304: Did the V1 library automatically assign a request ID? Is that why you added it here explicitly because V2 does not?

CoderHam: That is correct

PR #1356, opened 2020-04-23
Fix sequence batcher. Fix null request

GuanLuo:
Manually tested using `simple_grpc_v2_sequence_sync_infer_client.py` and `simple_grpc_v2_sequence_stream_infer_client.py`. Also use sequence_sync_client with different length sequences to test null request.

### DIFF
--- a/src/core/infer_request.cc
+++ b/src/core/infer_request.cc
@@ -35,60 +35,6 @@ namespace nvidia { namespace inferenceserver {
 
 namespace {
 
-class PersistentBufferManager {
- public:
-  class Buffer {
-   public:
-    Buffer(size_t byte_size)
-        : buffer_(malloc(byte_size)), byte_size_(byte_size), count_(1)
-    {
-    }
-    ~Buffer() { free(buffer_); }
-    std::mutex mtx_;
-    void* buffer_;
-    size_t byte_size_;
-    size_t count_;
-  };
-
-  // Note that the buffer is not empty so that we can return a valid
-  // pointer when byte_size == 0
-  PersistentBufferManager() : buffer_(new Buffer(1)) { buffer_->count_++; }
-
-  ~PersistentBufferManager() { ReleaseBuffer(buffer_); }
-
-  // Return a buffer that is on heap so that it can be transferred as userp
-  Buffer* GetBuffer(size_t byte_size)
-  {
-    std::lock_guard<std::mutex> lk(mtx_);
-
-    if (byte_size > buffer_->byte_size_) {
-      ReleaseBuffer(buffer_);
-      buffer_ = new Buffer(byte_size);
-    }
-
-    std::lock_guard<std::mutex> blk(buffer_->mtx_);
-    buffer_->count_++;
-
-    return buffer_;
-  }
-
-  void ReleaseBuffer(Buffer* buffer)
-  {
-    std::lock_guard<std::mutex> blk(buffer->mtx_);
-    buffer->count_--;
-    if (buffer->count_ == 0) {
-      delete buffer;
-    }
-  }
-
- private:
-  std::mutex mtx_;
-  Buffer* buffer_;
-};
-
-std::unique_ptr<PersistentBufferManager> persistent_buffer_manager(
-    new PersistentBufferManager());
-
 // Utilities for Null request feature.
 TRITONSERVER_Error*
 NullResponseAlloc(
@@ -98,24 +44,9 @@ NullResponseAlloc(
     void** buffer_userp, TRITONSERVER_MemoryType* actual_memory_type,
     int64_t* actual_memory_type_id)
 {
-  auto buffer_sp = persistent_buffer_manager->GetBuffer(byte_size);
-  if (buffer_sp != nullptr) {
-    *buffer = buffer_sp->buffer_;
-    if (*buffer != nullptr) {
-      *actual_memory_type = TRITONSERVER_MEMORY_CPU;
-      *actual_memory_type_id = 0;
-      *buffer_userp = buffer_sp;
-      return nullptr;
-    } else {
-      persistent_buffer_manager->ReleaseBuffer(
-          reinterpret_cast<PersistentBufferManager::Buffer*>(buffer_sp));
-    }
-  }
-
-  *buffer_userp = nullptr;
   return TRITONSERVER_ErrorNew(
       TRITONSERVER_ERROR_INTERNAL,
-      "failed to allocate output buffer for null request.");
+      "unexpected allocation for null request, no output should be requestd.");
 }
 
 TRITONSERVER_Error*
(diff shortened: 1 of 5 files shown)


src/core/infer_request.h
```diff
@@ -428,6 +446,8 @@ class InferenceRequest {
 
   // The response factory associated with this request.
   InferenceResponseFactory response_factory_;
+
+  std::vector<std::function<void()>> internal_callbacks_;
```
deadeyegoodwin, line 450: release_callbacks_

src/core/infer_request.h
```diff
@@ -337,6 +346,15 @@ class InferenceRequest {
     return Status::Success;
   }
 
+  // Add a callback to be invoked on destruction. Multile callbacks
+  // can be added by calling this function in order, and they will be
+  // invoked in reversed order.
+  Status AddInternalCallback(std::function<void()>&& callback)
```
deadeyegoodwin, line 352: AddReleaseCallback.
These should be invoked when the request is released by Triton. The client can reuse the request or even just hold onto it for a long time and we don't want to defer these callbacks until then.

src/core/sequence_batch_scheduler.cc
```diff
@@ -1403,7 +1343,7 @@ OldestSequenceBatch::Enqueue(
   }
```
deadeyegoodwin, line 1343: Remove the FIXME a couple of lines above.

src/core/infer_request.cc
```diff
@@ -116,9 +168,50 @@ InferenceRequest::Release(std::unique_ptr<InferenceRequest>&& request)
...
+    // it with byte size that matches 'from'
+    if (input.first == *max_input_name) {
+      new_input->SetData(data);
+    } else {
+      new_input->AppendData(
+          data_base, input.second.Data()->TotalByteSize(), mem_type, mem_id);
+    }
+  }
+
+  lrequest->SetResponseCallback(
```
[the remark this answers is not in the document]
GuanLuo: I think we can do that, say letting the `CreateResponse()` returns `nullptr` for response if no callback is set for the response, and the backends should skip outputs for that particular response accordingly.

But then it becomes an additional rule for implementing custom backend. Developers will need to be aware that sometimes responses may not be created.

src/core/infer_request.cc
```diff
@@ -33,6 +33,58 @@
...
+// Utilities for Null request feature.
+TRITONSERVER_Error*
+NullResponseAlloc(
+    TRITONSERVER_ResponseAllocator* allocator, const char* tensor_name,
+    size_t byte_size, TRITONSERVER_MemoryType preferred_memory_type,
+    int64_t preferred_memory_type_id, void* userp, void** buffer,
+    void** buffer_userp, TRITONSERVER_MemoryType* actual_memory_type,
+    int64_t* actual_memory_type_id)
+{
+  *buffer = malloc(byte_size);
```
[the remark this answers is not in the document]
GuanLuo: I added a simple implementation of the persistent buffer class, it still uses mutex but should be okay for now.

src/core/infer_request.cc
```diff
@@ -33,6 +33,127 @@
 
 namespace nvidia { namespace inferenceserver {
 
+namespace {
+
+class PersistentBufferManager {
```
deadeyegoodwin, line 38: This wasn't exactly what I had in mind. We still have a mutex on every alloc... I think you should just put this to the side for now and we can revisit once we have a better undertanding of the full flow.

src/core/infer_request.cc
```diff
@@ -33,6 +33,127 @@
...
+// Utilities for Null request feature.
+TRITONSERVER_Error*
+NullResponseAlloc(
+    TRITONSERVER_ResponseAllocator* allocator, const char* tensor_name,
+    size_t byte_size, TRITONSERVER_MemoryType preferred_memory_type,
+    int64_t preferred_memory_type_id, void* userp, void** buffer,
+    void** buffer_userp, TRITONSERVER_MemoryType* actual_memory_type,
+    int64_t* actual_memory_type_id)
+{
+  auto buffer_sp = persistent_buffer_manager->GetBuffer(byte_size);
```
deadeyegoodwin, line 101: Just return an error since should be no allocations since no output requested.

src/core/infer_request.cc
```diff
@@ -33,6 +33,127 @@
...
+      "failed to allocate output buffer for null request.");
+}
+
+TRITONSERVER_Error*
+NullResponseRelease(
+    TRITONSERVER_ResponseAllocator* allocator, void* buffer, void* buffer_userp,
+    size_t byte_size, TRITONSERVER_MemoryType memory_type,
+    int64_t memory_type_id)
+{
+  if (buffer_userp != nullptr) {
```
deadeyegoodwin, line 127: Just return an error since should be no allocations since no output requested.

src/core/infer_request.cc
```diff
@@ -116,9 +168,50 @@ InferenceRequest::Release(std::unique_ptr<InferenceRequest>&& request)
...
-  return nullptr;
+  // Create a copy of 'from' request with artifical inputs and no requested
+  // outputs. Maybe more efficient to share inputs and other metadata,
+  // but that binds the Null request with 'from' request's lifecycle.
+  std::unique_ptr<InferenceRequest> lrequest(
+      new InferenceRequest(from.backend_raw_, from.requested_model_version_));
+  lrequest->needs_normalization_ = from.needs_normalization_;
+  lrequest->batch_size_ = from.batch_size_;
+
+  // Two pass: first to obtain the max input byte size for allocating a large
```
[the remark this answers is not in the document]
deadeyegoodwin: Just put this back to the way you had it for now.

src/core/infer_request.cc
```diff
@@ -116,9 +243,43 @@ InferenceRequest::Release(std::unique_ptr<InferenceRequest>&& request)
...
+    Input* new_input;
+    lrequest->AddOriginalInput(
+        input.first, input.second.DType(), input.second.Shape(), &new_input);
+    new_input->AppendData(
+        data_sp->buffer_, input.second.Data()->TotalByteSize(),
+        TRITONSERVER_MEMORY_CPU, 0);
+  }
+
+  lrequest->SetResponseCallback(
+      &null_allocator, nullptr, NullResponseComplete, nullptr);
```
deadeyegoodwin, line 276: Add a comment saying that since no outputs were requested there should be no allocations.

src/core/infer_request.h
```diff
@@ -337,6 +338,15 @@ class InferenceRequest {
     return Status::Success;
   }
 
+  // Add a callback to be invoked on destruction. Multile callbacks
+  // can be added by calling this function in order, and they will be
+  // invoked in reversed order.
+  Status AddReleaseCallback(std::function<void()>&& callback)
```
deadeyegoodwin, line 344: AddInternalReleaseCallback and the comment above needs to be updated.

src/core/sequence_batch_scheduler.cc
```diff
@@ -1073,6 +1064,13 @@ DirectSequenceBatch::SchedulerThread(
           if (use_null_request) {
             std::unique_ptr<InferenceRequest> ni(
                 InferenceRequest::CopyAsNull(*null_irequest));
+            // Note that when the not-ready control input of the
+            // request is "true" the model can't assume that any
+            // other inputs are meaningful, including CORRID. So we
+            // just use zero for that and use whatever values the
```
deadeyegoodwin, line 1070: "... just use zero for that." Delete rest of sentence.

deadeyegoodwin:
> @GuanLuo Need to trigger CI

Nevermind... merging.

PR #1357, opened 2020-04-23
C++ image client using V2 API

tanmayv25:
The http protocol can be supported after http_server_v2 is modified to provide classification result as binary_data.


GuanLuo:
FYI, the server frontend should be able to return classification result as binary data already.

tanmayv25:
There is a bug on the server for class outputs as binary. It reports "binary_data_size" as 0 in response to the clients. 
@CoderHam  Came across the same thing while working on infer_utils for tests. He can share the ticket here.

PR #1359, opened 2020-04-24
Allow passing S3 credentials via Environment Variables

CoderHam:
Priority: ENV Variables -> AWS_PROFILE -> default

### DIFF
--- a/qa/L0_storage_S3/test.sh
+++ b/qa/L0_storage_S3/test.sh
@@ -122,7 +122,7 @@ for ENV_VAR in "env" "env_dummy" "config"; do
             aws configure set default.region "dummy_region" && \
                 aws configure set aws_access_key_id "dummy_id" && \
                 aws configure set aws_secret_access_key "dummy_key"
-        else
+        elif [ "$ENV_VAR" == "env" ]; then
             rm ~/.aws/credentials && rm ~/.aws/config
         fi
 
@@ -180,7 +180,7 @@ for ENV_VAR in "env" "env_dummy" "config"; do
                 aws configure set default.region "dummy_region" && \
                     aws configure set aws_access_key_id "dummy_id" && \
                     aws configure set aws_secret_access_key "dummy_key"
-            else
+            elif [ "$ENV_VAR" == "env" ]; then
                 rm ~/.aws/credentials && rm ~/.aws/config
             fi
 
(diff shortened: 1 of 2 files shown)


dcyoung:
Will passing the region as an environment variable `AWS_DEFAULT_REGION=...` work in conjunction with the account id + secret supported in this PR?

CoderHam:
> Will passing the region as an environment variable `AWS_DEFAULT_REGION=...` work in conjunction with the account id + secret supported in this PR?

Sure. I have added support for AWS_DEFAULT_REGION as well in this PR.

dcyoung:
👍 Thanks!

qa/L0_storage_S3/test.sh
```diff
@@ -41,7 +41,7 @@ CLIENT_LOG_BASE="./client"
 INFER_TEST=infer_test.py
 
 # S3 credentials are necessary for this test. Pass via ENV variables
-aws configure set default.region us-west-1 && \
+aws configure set default.region $AWS_DEFAULT_REGION && \
```
deadeyegoodwin, line 44: If aws is always configured this way then how are you testing with just env vars? It seems there are 3 cases. With just envvars, with just "aws configure" and with both, where you set the lower priority method to some bogus values to make sure that it is not used.

CoderHam: Fixed

deadeyegoodwin: I still don't see how this is working. You are unconditionally setting this outside the loop. Shouldn't you be doing this in the loop as the last tested case?

qa/L0_storage_S3/test.sh
```diff
@@ -61,132 +61,169 @@ SERVER_TIMEOUT=360
...
+    SERVER_LOG=$SERVER_LOG_BASE.$ENV_VAR.log
+    CLIENT_LOG=$CLIENT_LOG_BASE.$ENV_VAR.log
     
-rm -f $SERVER_LOG_BASE* $CLIENT_LOG_BASE*
+    if [ "$ENV_VAR" != "env" ]; then
+        unset AWS_ACCESS_KEY_ID
+        unset AWS_SECRET_ACCESS_KEY
+        unset AWS_DEFAULT_REGION
+    elif
+        aws configure set default.region "dummy_region" && \
```
deadeyegoodwin, line 74: So the first time through the loop you set the configuration to be invalid... but that is ok because the envvars are going to be used.
The second time the envvars are unset... so now you ony have the configuration but isn't it invalid? And then below you set config again?
Can you run 3 iterations of the loop and at the start of each iteration setup envvars and config for each different case and clearly document what each iteration is testing. And then for the rest of the loop don't change env or config.

CoderHam: The config needs to be set correctly so that the aws s3 cli command work. It is set to the correct creds and then set back to dummy creds after the s3 command is run.

PR #1363, opened 2020-04-24
Fix number of samples generated on warmup

GuanLuo:


src/core/backend.cc
```diff
@@ -87,8 +87,7 @@ WarmupResponseComplete(TRITONSERVER_InferenceResponse* iresponse, void* userp)
 void
 WarmupRequestComplete(TRITONSERVER_InferenceRequest* request, void* userp)
 {
-  // do nothing with request object as it is created on stack,
-  // see SetConfiguredScheduler()
+  TRITONSERVER_InferenceRequestDelete(request);
   auto warmup_promise = reinterpret_cast<std::promise<void>*>(userp);
   warmup_promise->set_value();
```
deadeyegoodwin, line 92: Don't you need to wait for the response to complete?

GuanLuo: I am assuming the request will be released after the response is completed?

GuanLuo:
The input shape are being handled in `GenerateWarmupData()` which is fixed in previous PR.

src/core/backend.cc
deadeyegoodwin: Actually the request can usually be released before the response is sent. But there is not requirement that they go in any particular order. But after seeing the response you can be sure that the inference has completed (and thus the warmup has completed).

PR #1364, opened 2020-04-24
Update statistics extension

deadeyegoodwin:


docs/protocol/extension_statistics.md
```diff
@@ -42,72 +42,125 @@ $object and $array refer to the fundamental JSON types. #optional
...
+  output tensor data produced by the model framework / backend. For
+  example, this duration should include the time to copy output tensor
+  data from the GPU.
+
+```
+$batch_stats =
+{
+  "batch_size" : $number,
+  "count" : $number,
+  "compute_input" : $duration_stat,
```
tanmayv25, line 139: Why don't we have "queue" field in $batch_stats?

docs/protocol/extension_statistics.md
```diff
@@ -217,4 +296,33 @@ message InferStatistics
...
+  // The size of the batch.
+  uint64 batch_size = 1;
+
+  // The number of times the batch size was executed on the
+  // model. A single model execution performs inferencing for the entire
+  // request batch and can perform inferencing for multiple requests if
+  // dynamic batching is enabled.
+  uint64 count = 2;
+
+  // The count and cumulative duration to prepare input tensor data as
```
tanmayv25, line 312: Should we have another message holding "StatisticDuration" messages, and use that in InferBatchStatistics and InferStatistics? or is it too much nesting?

docs/protocol/extension_statistics.md
deadeyegoodwin: queue time only makes since for individual requests. The batch stats are for an entire batch which may contain multiple requests. We could probably think of some way to include it but I think it would just be confusing. Better to just have very batch specific stats here.

docs/protocol/extension_statistics.md
deadeyegoodwin: Yes, trying to avoid the nesting, and inferstats has some stats we don't want for batch.

docs/protocol/extension_statistics.md
```diff
@@ -42,72 +42,125 @@ $object and $array refer to the fundamental JSON types. #optional
...
+- “compute_output” : The count and cumulative duration to extract
+  output tensor data produced by the model framework / backend. For
+  example, this duration should include the time to copy output tensor
+  data from the GPU.
+
+```
+$batch_stats =
+{
+  "batch_size" : $number,
+  "count" : $number,
```
GuanLuo, line 138: "count" can be obtained from `$duration_stat`, will it be redundant?

deadeyegoodwin: @GuanLuo  Yes, good point. Can you remove it as part of your stats change.

PR #1365, opened 2020-04-24
Update TF backend to new request/response flow

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo Mostly just an FYI for you. This doesn't have the input and output utilities we discussed. I will work on those in a separate PR. This does pass the simplev2 test.

PR #1373, opened 2020-04-27
Fixing Windows10 build for V2 client library

tanmayv25:


src/clients/c++/experimental_api_v2/library/common.cc
```diff
@@ -228,7 +228,7 @@ InferInput::GetNext(
 
   while ((bufs_idx_ < bufs_.size()) && (size > 0)) {
     const size_t buf_byte_size = buf_byte_sizes_[bufs_idx_];
-    const size_t csz = std::min(buf_byte_size - buf_pos_, size);
+    const size_t csz = (std::min)(buf_byte_size - buf_pos_, size);
```
deadeyegoodwin, line 231: Strange thing for windows to require... seems like a bug in their compiler

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -40,6 +40,9 @@
 #include <string>
 #include <thread>
 #include <vector>
+#ifdef _WIN32
+#include <algorithm>
```
deadeyegoodwin, line 44: I this file requires something from <algorithm> we should just include it for all platforms

src/clients/c++/experimental_api_v2/library/common.cc
tanmayv25: In one of the standard headers for windows, min and max are defined as MACROS that takes two arguments as well. This causes a conflict between the macro and std::min function. Enclosing the (std::min)/(std::max) ensures the function is selected and not the macro.

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -40,6 +40,7 @@
 #include <string>
 #include <thread>
 #include <vector>
+#include <algorithm>
```
deadeyegoodwin, line 43: Probably need to run formatter. No need to restart CI after doing that.

tanmayv25: Done

PR #1380, opened 2020-04-28
Remove the V1 clients in v2test branch

tanmayv25:


CoderHam:
You should default the server to use api_version = 2 as well

PR #1381, opened 2020-04-28
Fix shared memory utils for V2

CoderHam:
- Fix shared memory utils for V2
- Add simple http v2 shm string client for testing shared memory string support
- Fix bugs in HTTP server Json parsing and creation
**PS**: Varying number of failures seen in L0_infer_cudashm seen due to improper base64 serialization / json creation while trying to register cuda shared memory handle


qa/common/infer_util.py
```diff
@@ -62,7 +62,15 @@ def _range_repr_dtype(dtype):
     return dtype
 
 
+def _prepend_string_size(input_values):
```
tanmayv25, line 65: _serialize_byte_tensor_list ? Also, we are passing ouput0_values to this function in L152

qa/common/infer_util.py
```diff
@@ -137,21 +146,20 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
-        input1_array_tmp = serialize_byte_tensor(input1_array)
-    else:
-        input0_array_tmp = input0_array
-        input1_array_tmp = input1_array
-
+    # prepend size of string to output string data
     if output0_dtype == np.object:
-        output0_array_tmp = serialize_byte_tensor(output0_array)
+        if batch_size == 1:
+            output0_array_tmp = _prepend_string_size([output0_array])
```
tanmayv25, line 152: This is needed just to get the output_byte_size right? Why not get the size directly without serializing first?

qa/common/infer_util.py
```diff
@@ -400,3 +437,445 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+        # create and register shared memory region for inputs and outputs
+        shm_io_handles = su.create_register_set_either_shm_region([shm_region_name_prefix[0]+str(io_num),
+                                                shm_region_name_prefix[1]+str(io_num)], input_array,
+                                                input_byte_size, output_byte_size, shm_client,
+                                                use_system_shared_memory, use_cuda_shared_memory)
+        if len(shm_io_handles) != 0:
+            shm_ip_handles.append(shm_io_handles[0])
+            shm_op_handles.append(shm_io_handles[1])
+            input_dict[input_name] = (shm_ip_handles[io_num], input_shapes)
+            output_dict[output_name] = (InferContext.ResultFormat.RAW, shm_op_handles[io_num])
```
tanmayv25, line 521: Remove InferContext.

qa/common/infer_util.py
```diff
@@ -400,3 +437,445 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+        else:
+            input_dict[input_name] = input_list
+            output_dict[output_name] = InferContext.ResultFormat.RAW
+
+    # Run inference and check results for each config
+    for config in configs:
+        model_name = tu.get_zero_model_name(pf, io_cnt, tensor_dtype)
+
+        ctx = InferContext(config[0], config[1], model_name, model_version,
```
tanmayv25, line 530: Looks like you missed changing these lines to use V2 APIs.

qa/common/infer_util.py
```diff
@@ -400,3 +437,445 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+        raise ValueError(
+            "Cannot set both System and CUDA shared memory flags to 1")
+
+    input_dict = {}
+    # output_dict = {}
+    expected_dict = {}
+    shm_ip_handles = list()
+    shm_op_handles = list()
+
+    shm_client = httpclient.InferenceServerClient("localhost:8000")
```
tanmayv25, line 592: You are using the http endpoint only for registering the shared memory. Can you use something like:

>  if config[1] == "http":
>             triton_client = httpclient.InferenceServerClient(
>                 config[0], verbose=True)
>         else:
>             triton_client = grpcclient.InferenceServerClient(
>                 config[0], verbose=True)

and then use triton_client to set up the shared memory and send the request? No need of having separate triton_client and shm_client. This way we can improve the test coverage.

qa/common/shm_util.py
```diff
@@ -53,12 +53,19 @@ def create_register_shm_regions(input0_list, input1_list, output0_byte_size,
             "Cannot set both System and CUDA shared memory flags to 1")
 
     if not (use_system_shared_memory or use_cuda_shared_memory):
-        return []
+        return [], None, None
 
     triton_client = httpclient.InferenceServerClient("localhost:8000")
```
tanmayv25, line 58: Same here. Can you pass the protocol to "create_register_shm_regions" and then use appropriate client?

qa/common/infer_util.py
```diff
@@ -137,21 +146,20 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
-        input0_array_tmp = serialize_byte_tensor(input0_array)
-        input1_array_tmp = serialize_byte_tensor(input1_array)
-    else:
-        input0_array_tmp = input0_array
-        input1_array_tmp = input1_array
-
+    # prepend size of string to output string data
     if output0_dtype == np.object:
-        output0_array_tmp = serialize_byte_tensor(output0_array)
+        if batch_size == 1:
```
GuanLuo, line 151: Actually we don't need `batch_size` at all in `infer_exact` as there is not such notion in v2. Do we have to condition on `batch_size`?

qa/common/infer_util.py
CoderHam: Right. The variable names should be generic

qa/common/infer_util.py
CoderHam: Sure, I could do that

qa/common/infer_util.py
CoderHam: This is created once for all the configs

qa/common/shm_util.py
CoderHam: Similarly, created once for all configs and reused after that

src/clients/python/experimental_api_v2/examples/simple_http_v2_shm_string_client.py
```diff
@@ -0,0 +1,194 @@
...
+import numpy as np
+import sys
+from builtins import range
+import tritonhttpclient.core as httpclient
+import tritonsharedmemoryutils.shared_memory as shm
+import tritonhttpclient.utils as utils
+
+FLAGS = None
+
+if __name__ == '__main__':
```
GuanLuo, line 38: I think we are having too many v2 clients. We don't need a client for every combinations of different features, I think most of the clients were introduced as sanity check before v2 test transition and we should reducing them to minimal set of examples for showing how to use each feature as we have more and more test coverage on V2.

For example, what this client is testing should already be covered in `L0_infer_xxx_shared_memory`. And the features shown here is the combination of what are shown in `simple_http_v2_shm_client.py` and `simple_http_v2_string_infer_client.py`

qa/common/infer_util.py
GuanLuo: That will limit the test environment to always have HTTP client installed, while the declaration of the test util suggests that it can be used to test HTTP / GRPC independently. i.e. we can't just install GRPC client and use the test for GRPC only.

qa/common/infer_util.py
```diff
@@ -165,13 +173,20 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
         INPUT0 = "INPUT__0"
         INPUT1 = "INPUT__1"
 
-    output0_byte_size = output0_array_tmp.nbytes
-    output1_byte_size = output1_array_tmp.nbytes
+    output0_byte_size = sum([o0.nbytes for o0 in output0_array_tmp])
+    output1_byte_size = sum([o1.nbytes for o1 in output1_array_tmp])
```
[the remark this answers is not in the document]
CoderHam: That is correct. Reverted back.

qa/common/infer_util.py
```diff
@@ -72,7 +80,8 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
                 correlation_id=0, shm_region_names=None, precreated_shm_regions=None,
                 use_system_shared_memory=False, use_cuda_shared_memory=False,
                 priority=0, timeout_us=0):
-    tester.assertTrue(use_http or use_grpc or use_streaming)
+    tester.assertTrue(
+        use_http or use_http_json_tensors or use_grpc or use_streaming)
```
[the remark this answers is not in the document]
CoderHam: GRPC has only one way of passing data to the server i.e. via our specified protobuf. HTTP on the other hand has two options i.e. binary or json. It would not make sense to have a common flag for both GRPC and HTTP.

qa/common/infer_util.py
CoderHam: We do. Without batch_size we cannot know if the first dimension of the shape is the batch_size of if the model input does not have the batch dimension

src/clients/python/experimental_api_v2/examples/simple_http_v2_shm_string_client.py
CoderHam: The reason this client is necessary is that the serialization of string data for the shm case happens outside the client library and in that case for string I/O hits different parts of the codebase

qa/common/infer_util.py
GuanLuo: I am confused, I thought we already passing the full shape (i.e. include batch dimension if model support batching): https://github.com/NVIDIA/triton-inference-server/pull/1355#discussion_r414192810

qa/common/infer_util.py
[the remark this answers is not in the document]
GuanLuo: @tanmayv25 Is binary extension not supported in GRPC?

src/clients/python/experimental_api_v2/examples/simple_http_v2_shm_string_client.py
GuanLuo: Good point. And you should add a comment in the place you do the extra serialization / deserialization, about why the extra steps are needed.

qa/common/infer_util.py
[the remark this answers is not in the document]
tanmayv25: For gRPC, the server can handle explicit tensors in the input, but in our client library we always use raw_contents to upload to the data to server.

qa/common/infer_util.py
tanmayv25: Can you can create the shared memory region once, but inside the loop, register and unregister these precreated shared memory regions using appropriate client? Otherwise we are missing out on gRPC testing.

src/clients/python/experimental_api_v2/examples/simple_http_v2_shm_string_client.py
CoderHam: Done

qa/common/infer_util.py
CoderHam: Fixed

GuanLuo:
I have a question regarding the shared memory utils, what is the reason of not allowing it to process string tensor directly? Now users need to serialize / deserialize the string tensors before they can pass it to `set_shm_region()`, but why don't we move the (de)serialization inside the shared memory utils so users don't need to do so explicitly? Like what we did for binary extension, the (de)serialization is performed internally

CoderHam:
> I have a question regarding the shared memory utils, what is the reason of not allowing it to process string tensor directly? Now users need to serialize / deserialize the string tensors before they can pass it to `set_shm_region()`, but why don't we move the (de)serialization inside the shared memory utils so users don't need to do so explicitly? Like what we did for binary extension, the (de)serialization is performed internally

I went through the client libraries a while back and I noticed that indeed it is possible to do the same. I however would like to do that in a different PR since this PR contains vital fixes to the client and http server. I will create a ticket for the improvement you suggested.

tanmayv25:
Yes. I was thinking of the same. We can modify the shared_memory_utils to expose same interface as "set_data_from_numpy" and "as_numpy".

qa/common/infer_util.py
```diff
@@ -399,4 +455,4 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
     su.unregister_cleanup_shm_regions(shm_regions, precreated_shm_regions, outputs,
                                       use_system_shared_memory, use_cuda_shared_memory)
 
-    return results
+    return results
```
tanmayv25, line 458: A new line

CoderHam: Fixed

PR #1384, opened 2020-04-29
Simplify shape tensor handling

deadeyegoodwin:
Require input shape tensors to be in contiguous CPU memory. This
removes the need for peek functionality and simplifies request
handling. If we need more functionality in the future we can promote
shape tensors to CPU memory early in request handling instead of where
we were doing it.


deadeyegoodwin:
@tanmayv25  I'm simplifying the handling of shape tensors (specifically the peek requirement) in the new request/response flow so I thought you might want to take a look.

Issue #1377, opened 2020-04-28
Server fails to load tensorflow_savedmodel when using s3 model repository


dcyoung:
**Description**
I have exported the same model in both formats `tensorflow_graphdef` and `tensorflow_savedmodel`.

For the graphdef, the output directory structure looks like:
```bash
  model_repository/
    model_r50/
      config.pbtxt
      labels.txt
      1/
        model.graphdef
```
For the savedmodel, the output directory structure looks like:
```bash
  model_repository/
    model_r50/
      config.pbtxt
      labels.txt
      1/
        model.savedmodel/
          saved_model.pb
          variables/
            variables.data-00000-of-00001
            variables.index
```

If I point the server to the model repository on a local file system, it is able to load both formats just fine. However, if I place the same model exports on s3, the server is only able to load the graphdef version. When attempting to load the savedmodel version, I'm encountering  following error:

```bash
...
I0428 01:50:23.718664 1 model_repository_manager.cc:735] CreateInferenceBackend() 'model_r50' version 1
E0428 01:50:24.018262 1 model_repository_manager.cc:840] failed to load 'model_r50' version 1: Internal: Failed to create local folder: /tmp/fileIXqtns/variables
I0428 01:50:24.018320 1 model_repository_manager.cc:622] TriggerNextAction() 'model_r50' version 1: 0
I0428 01:50:24.018342 1 model_repository_manager.cc:637] no next action, trigger OnComplete()
I0428 01:50:24.018464 1 model_repository_manager.cc:492] GetVersionStates() 'model_r50'
error: creating server: INTERNAL - failed to load all models
```

**Triton Information**
20.03

Are you using the Triton container or did you build it yourself?
provided container

**Expected behavior**
Parity between graphdef and savedmodel variants, across both s3 and local file system model repositories.

CoderHam:
Can you double check the S3 directory structure and permissions? We use a random file name generator to create a unique file/folder in `/tmp` of the form `/tmp/fileXXXXXX` where the S3 model is stored. If there is a name clash with a file that is already present it will give an error similar to what you see above. 

Could you cleanup all S3 duplicate files / folders in the docker container using `rm -rf /tmp/file*` and try again?

dcyoung:
Thanks for the quick reply!

> Can you double check the S3 directory structure and permissions?

The directory structure on s3 matches what is expected. I have verified that i can pull the entirety of this structure using the the same credentials as in the container. 

[...]

PR #1387, opened 2020-04-30
Fix bugs in S3 model loading noticed in Tensorflow savedmodel

CoderHam:
Fix to bug described in #1377


CoderHam:
The tests themselves were correct. The server was not reporting errors correctly (that i fixed in this PR) and hence there was no error reported during model loading.
The subfolders did not impact model loading since the variables folder was empty and we did not test for sanity of outputs when we have files under variables/* and hence the tests were successful.

PR #1388, opened 2020-05-01
perf_client using V2 API

tanmayv25:
Filing a PR to start collecting review. Most of the functionalities have been tested for concurrency manager + grpc protocol. There are some failures related to json parsing at http side that needs to be fixed.

- [X] ConcurrencyManager + All combinations of client API + All combinations of input data 
- [x] RequestRateManager and CustomLoadManager
~Fix some HTTP failures while parsing json.~

As this has become a huge PR because of code replication following might help in reviewing the code changes:

1. Most of the logic is same as the V1 perf_client as long as collecting and reporting stats is concerned.
2. The new ModelParser class allows perf_client to interpret model metadata and model config to get input/output details.
3. The new TritonClientWrapper class acts as a glue to the HTTP and gRPC library and tries to limit the clutter in the main code.
4. There have been significant changes in ConcurrencyManager::Infer() to handle new streaming API.

Apart from these major changes, various minor supporting changes are widespread.

UPDATE: All the test cases are verified for gRPC in L0_perf_client. There are still some issues with http protocol which will be addressed with a different PR:

1.  Server doesn't return error as json for bad requests.
2. Occasionally, the cuda raw IPC handle is not serialized properly.


CoderHam:
@tanmayv25 can you confirm that the HTTP failures are resolved now? I have rebased v2test with all our recent fixes.

tanmayv25:
@CoderHam  I still see the HTTP failures: batch_size > max_batch size and the issue of cuda shm handle b64 encoding, I will include the fixes in separate PR. I believe for now  We can include the perf_client_v2.

src/clients/c++/experimental_api_v2/library/common.h
```diff
@@ -222,6 +222,18 @@ class InferInput {
       InferInput** infer_input, const std::string& name,
       const std::vector<int64_t>& dims, const std::string& datatype);
 
+  /// Gets name of the associated input tensor.
+  /// \return The name of the tensor.
+  std::string Name() const { return name_; }
```
deadeyegoodwin, line 227: Return type should be const& for all three of these methods.

src/clients/c++/experimental_api_v2/perf_client/inference_profiler.h
```diff
@@ -0,0 +1,403 @@
...
+      const TimestampVector& timestamps,
+      const std::map<ModelIdentifier, ModelStatistics>& start_status,
+      const std::map<ModelIdentifier, ModelStatistics>& end_status,
+      const nic::InferStat& start_stat, const nic::InferStat& end_stat,
+      PerfStatus& summary);
+
+  /// A helper function to get the start and end of a measurement window.
+  /// \param timestamps The timestamps collected for the measurement.
+  /// \return the start and end timestamp of the measurement window.
+  std::pair<uint64_t, uint64_t> MeasurementTimestamp(
```
deadeyegoodwin, line 313: const& or pass in a std::pair* to intialize, or if not performance critical say something in the comment ("returning by value because not performance critical")

src/clients/c++/experimental_api_v2/perf_client/inference_profiler.h
```diff
@@ -0,0 +1,403 @@
...
+      const TimestampVector& timestamps);
+
+  /// \param timestamps The timestamps collected for the measurement.
+  /// \param valid_range The start and end timestamp of the measurement window.
+  /// \param valid_sequence_count Returns the number of completed sequences
+  /// during the measurement. A sequence is a set of correlated requests sent to
+  /// sequence model.
+  /// \return the vector of request latencies where the requests are completed
+  /// within the measurement window.
+  std::vector<uint64_t> ValidLatencyMeasurement(
```
deadeyegoodwin, line 323: const&? Same comment as above.

src/clients/c++/experimental_api_v2/perf_client/model_parser.h
```diff
@@ -0,0 +1,152 @@
...
+  /// \return Error object indicating success or failure.
+  nic::Error Init(
+      const rapidjson::Document& metadata, const rapidjson::Document& config,
+      const std::string& model_version,
+      const std::unordered_map<std::string, std::vector<int64_t>>& input_shapes,
+      std::unique_ptr<TritonClientWrapper>& client_wrapper);
+
+  /// Get the name of the target model
+  /// \return Model name as string
+  std::string ModelName() const { return model_name_; }
```
deadeyegoodwin, line 99: const& here and below

src/clients/c++/experimental_api_v2/perf_client/model_parser.cc
```diff
@@ -0,0 +1,297 @@
...
+          model_config.config(), step_version, client_wrapper, is_sequential));
+    }
+  }
+
+  return nic::Error::Success;
+}
+
+nic::Error
+ModelParser::Init(
+    const rapidjson::Document& metadata, const rapidjson::Document& config,
```
GuanLuo, line 146: I think `config` can be coverted to corresponding protobuf object, and vice versa. Can just convert it and call the other Init() function?

tanmayv25: We use config just to detect scheduler type and detecting which tensors are shape tensor. I believe it would be wasteful to convert config DOM into protobuf then extracting those values instead of directly extracting those values from DOM. If we were to call the other Init() function we would have to convert metadata to protobuf as well.

PR #1389, opened 2020-05-01
Fix L0_priority to recognize new CUDA stream priority level

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo This is a partial cherry pick of a change from 20.05

PR #1393, opened 2020-05-04
Fix server / client error checking

GuanLuo:
@CoderHam Can you check if the `asyn_infer()` also needs to be fixed? The function declaration says exception will be raised but the error will be passed into the client callback, I think we just need to clarify on how the error will be reported.

### DIFF
--- a/src/clients/python/experimental_api_v2/library/httpclient.py
+++ b/src/clients/python/experimental_api_v2/library/httpclient.py
@@ -1033,6 +1033,7 @@ class InferenceServerClient:
                               request_body=request_body,
                               headers=headers,
                               query_params=query_params)
+        _raise_if_error(response)
 
         return InferResult(response, self._verbose)
 
--- a/src/clients/python/experimental_api_v2/examples/simple_http_v2_infer_client.py
+++ b/src/clients/python/experimental_api_v2/examples/simple_http_v2_infer_client.py
@@ -30,7 +30,7 @@ import numpy as np
 import sys
 
 import tritonhttpclient.core as httpclient
-
+from tritonhttpclient.utils import InferenceServerException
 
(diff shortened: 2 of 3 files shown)


src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -1033,6 +1033,7 @@ def infer(self,
                               request_body=request_body,
                               headers=headers,
                               query_params=query_params)
+        _raise_if_error(response)
```
tanmayv25, line 1036: This will not work for the case output is in binary-data. We must call _raise_if_error on the json part parsed by InferResult class.

GuanLuo: For HTTP, this checks on the status code returned. Are you saying that this is the case where the status code will returned OK but the response is malformed?

tanmayv25: You are right.

tanmayv25:
This must be fixed as well:
https://github.com/NVIDIA/triton-inference-server/blob/gluo-fix_http/src/clients/python/experimental_api_v2/examples/simple_http_v2_infer_client.py#L148-L157

src/clients/python/experimental_api_v2/library/httpclient.py
CoderHam: Similar changes necessary in some test scripts as well. https://github.com/NVIDIA/triton-inference-server/blob/v2test/qa/common/infer_util.py#L337
I will make them in a different PR once this is merged

CoderHam: Yes in the case of inference the server always accompanies BADREQ with and error json response

PR #1397, opened 2020-05-04
Port infer_zero util and L0_infer_zero

CoderHam:
~~@tanmayv25 gRPC support for empty input/output tensors needs fixing. I have disabled them for now. (For non-empty tensors gRPC works as expected)~~
PS: Shared memory support will be added shortly.

### DIFF
--- a/qa/common/infer_util.py
+++ b/qa/common/infer_util.py
@@ -113,6 +113,11 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
 
     num_classes = 3
 
+    if model_version is not None:
+        model_version = str(model_version)
+    else:
+        model_version = ""
+
     input0_array = np.random.randint(low=val_min, high=val_max,
                                      size=tensor_shape, dtype=rinput_dtype)
     input1_array = np.random.randint(low=val_min, high=val_max,
@@ -312,11 +317,6 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
                         output_req.append(grpcclient.InferRequestedOutput(
                             OUTPUT1, class_count=num_classes))
 
-        if model_version is not None:
-            model_version = str(model_version)
-        else:
-            model_version = ""
-
         if config[2]:
             # TODO fix for streaming case
             continue
@@ -456,6 +456,8 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
                                       use_system_shared_memory, use_cuda_shared_memory)
 
     return results
+
+
 # resize the dummy tensor with the provided values in the shape tensor and finally
 # return the shape of the resized tensor.
 def infer_shape_tensor(tester, pf, tensor_dtype, input_shape_values, dummy_input_shapes,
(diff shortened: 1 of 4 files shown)


GuanLuo:
Does #1396 fix the issue mentioned in the description?

tanmayv25:
Can you try cherry-picking #1402 and give it a try?

qa/common/infer_util.py
```diff
@@ -456,3 +456,216 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+        expected_dict[output_name] = expected_array
+
+        input_byte_size = tu.shape_element_count(input_shape) *\
+            np.dtype(tensor_dtype).itemsize
+        output_byte_size = tu.shape_element_count(output_shape) *\
+            np.dtype(tensor_dtype).itemsize
+
+        # create and register shared memory region for inputs and outputs
+        shm_io_handles = []
+        # shm_io_handles = su.create_register_set_either_shm_region([shm_region_name_prefix[0]+str(io_num),
```
GuanLuo, line 532: Add a FIXME or TODO?

qa/common/infer_util.py
```diff
@@ -456,3 +456,216 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+                                          inputs,
+                                          model_version=model_version,
+                                          outputs=output_req,
+                                          request_id=str(_unique_request_id()),
+                                          priority=priority, timeout=timeout_us)
+
+        last_response = results.get_response()
+        if config[1] == "http":
+            if 'error' in last_response:
+                raise InferenceServerException(msg=last_response['error'])
```
GuanLuo, line 597: Exception will be raised from `infer()`

qa/common/infer_util.py
```diff
@@ -456,3 +456,216 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+            if config[1] == "http":
+                response_model_version = last_response["model_version"]
+            else:
+                response_model_version = last_response.model_version
+            tester.assertEqual(response_model_version, model_version)
+
+        if config[1] == "http":
+            response_outputs = last_response["outputs"]
+        else:
+            response_outputs = last_response.outputs
```
GuanLuo, line 615: Merge the assignment into one conditional

qa/common/infer_util.py
```diff
@@ -456,3 +456,216 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
...
+
+        for idx, result in enumerate(response_outputs):
+            if config[1] == "http":
+                result_name = result["name"]
+            else:
+                result_name = result.name
+
+            tester.assertTrue(result_name in expected_dict)
+            if use_system_shared_memory or use_cuda_shared_memory:
+                shm_handle = shm_op_handles[idx]
```
GuanLuo, line 626: Not safe to assume output order in response.

qa/common/infer_util.py
CoderHam: Fixing in next commit

qa/common/infer_util.py
```diff
@@ -113,6 +113,11 @@ def infer_exact(tester, pf, tensor_shape, batch_size,
 
     num_classes = 3
 
+    if model_version is not None:
+        model_version = str(model_version)
+    else:
+        model_version = ""
```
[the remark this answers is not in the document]
CoderHam: Correct but the model_version is used for each config. It is unnecessary to perform this each time.

CoderHam:
> Does #1396 fix the issue mentioned in the description?

No it only fixed the issue for http

> Can you try cherry-picking #1402 and give it a try?

Yes it did fix support for gRPC empty tensors.

qa/common/infer_util.py
CoderHam: We do not run the shared memory variants of the infer_zero tests. This is probably because we cannot create shared memory regions of size 0.

GuanLuo: Then you can strip the shared memory handling from this. Or you will create a non-empty shared memory region and use it for zero-element tensor. I will suggest to do the latter since it is a plausible use case.

qa/common/infer_util.py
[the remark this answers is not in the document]
GuanLuo: I don't think this is critical, but if you want, moving it to the top of `infer_zero()` or right before L543 will be better, right now it is in the middle of unrelated context.

qa/common/infer_util.py
CoderHam: Fixed

qa/common/infer_util.py

qa/common/infer_util.py
CoderHam: Ok, will do that

qa/common/shm_util.py
```diff
@@ -215,3 +221,58 @@ def unregister_cleanup_shm_regions(shm_regions, shm_handles, precreated_shm_regi
                     shm_regions[2+i]+'_data')
                 shm.destroy_shared_memory_region(shm_handles[3])
 
+
+def create_set_either_shm_region(shm_region_names, input_list, input_byte_size,
```
GuanLuo, line 225: Will those helper functions be reused in place other than infer_zero? It seems to be too specific

CoderHam: No, they are only used in infer_zero for the L0_infer_zero test

GuanLuo: Then move them to infer_zero

CoderHam: `create_set_shm_regions` and `register_add_shm_regions` are also only used for L0_infer but these helper functions are kept in `shm_util` instead of `inter_util`. Are you suggesting that we get rid of shm_util entirely?

GuanLuo: Personally I will remove it because the functions in `shm_util` only work for `infer_util`, like what we do in sequence util. But we can leave it as it is and revisit it later, for now, you can merge it if it fixes L0_infer_zero.

CoderHam: Will resolve this in a future PR. For now it works and fixes L0_infer_zero for V2

PR #1398, opened 2020-05-04
fix missing check for BOOL in ProtocolStringToDataType

CoderHam:


src/core/model_config.cc
```diff
@@ -465,8 +465,14 @@ ProtocolStringToDataType(const char* dtype, size_t len)
       return TYPE_FP64;
     }
   } else if (*dtype == 'B') {
-    if (!strcmp(dtype + 1, "YTES")) {
-      return TYPE_STRING;
+    if (dtype[1] == 'Y') {
+      if (!strcmp(dtype + 2, "TES")) {
+        return TYPE_STRING;
+      }
+    } else if (dtype[1] == 'O') {
```
GuanLuo, line 472: equivalent to `else if (!strcmp(dtype + 1, "OOL")`

CoderHam: Fixed

PR #1399, opened 2020-05-04
Fail HTTP request is endpoint unrecognized

deadeyegoodwin:


src/servers/http_server_v2.cc
```diff
@@ -2296,4 +2298,4 @@ HTTPServerV2::CreateMetricsServer(
   return nullptr;
 #endif  // TRTIS_ENABLE_METRICS
 }
-}}  // namespace nvidia::inferenceserver
```
CoderHam, line 2299: new line

PR #1401, opened 2020-05-05
fix bugs in api_regex, reading from and writing to json

CoderHam:
These fixes were made in v2test. Bringing them over to master

### DIFF
--- a/src/servers/http_server_v2.cc
+++ b/src/servers/http_server_v2.cc
@@ -216,7 +216,7 @@ class HTTPAPIServerV2 : public HTTPServerV2Impl {
         trace_manager_(trace_manager), shm_manager_(shm_manager),
         allocator_(nullptr), server_regex_(R"(/v2(?:/health/(live|ready))?)"),
         model_regex_(
-            R"(/v2/models/([^/]+)(?:/version/([0-9]+))?(?:/(infer|ready|config|stats))?)"),
+            R"(/v2/models/([^/]+)(?:/versions/([0-9]+))?(?:/(infer|ready|config|stats))?)"),
         modelcontrol_regex_(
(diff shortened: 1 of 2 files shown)


src/servers/http_server_v2.cc
```diff
@@ -715,25 +715,39 @@ WriteDataToJsonHelper(
...
     const DataType dtype)
 {
-  for (int i = 0; i < shape[shape_index].GetInt(); i++) {
-    if ((shape_index + 1) != (int)shape.Size()) {
-      rapidjson::Value response_output_array(rapidjson::kArrayType);
-      WriteDataToJsonHelper(
-          &response_output_array, allocator, shape, shape_index + 1, base,
-          counter, dtype);
-      response_output_val->PushBack(response_output_array, allocator);
+  if (shape_index == (int)shape.Size()) {
```
deadeyegoodwin, line 718: Does shape.Size() return a size_t? Why is 'shape_index' type "int"? If it is an index size_t would be more appropriate.

src/servers/http_server_v2.cc
```diff
@@ -715,25 +715,39 @@ WriteDataToJsonHelper(
...
+    if (dtype != TYPE_STRING) {
+      rapidjson::Value data_val((T)(base[*counter]));
```
deadeyegoodwin, line 720: base is already a T*, cast (T) is unnecessary.

src/servers/http_server_v2.cc
CoderHam: That is right. Will fix this

src/servers/http_server_v2.cc
CoderHam: Correct. Will fix in a follow up PR

PR #1404, opened 2020-05-05
Fix L0_trt_shape_tensors

GuanLuo:
Manually tested regular, system shared memory, and cuda shared memory (w/o HTTP).
Note that the `check_status` is changed as infer stats is collected differently in V2:
- batch infer is reported based on total batch size of the execution instead of batch size of the request
- ~~infer count is incremented by 1 for each request, instead of by the batch size of the request.~~

@CoderHam Note that I touched some sequence utilities, please check if those changes make sense to you.

### DIFF
--- a/qa/L0_trt_shape_tensors/trt_shape_tensor_test.py
+++ b/qa/L0_trt_shape_tensors/trt_shape_tensor_test.py
@@ -39,8 +39,6 @@ import test_util as tu
 import sequence_util as su
 
 import tritongrpcclient.core as grpcclient
-import tritongrpcclient.utils as grpcutils
-import tritonhttpclient.utils as httputils
 import os
 
 TEST_SYSTEM_SHARED_MEMORY = bool(int(os.environ.get('TEST_SYSTEM_SHARED_MEMORY', 0)))
@@ -317,8 +315,6 @@ class SequenceBatcherShapeTensorTest(su.SequenceBatcherTestUtil):
         dtype = np.float32
         try:
             model_name = tu.get_sequence_model_name("plan", dtype)
-            protocol = "streaming"
-
             self.check_setup(model_name)
 
             # Need scheduler to wait for queue to contain all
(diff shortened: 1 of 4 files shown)


qa/L0_trt_shape_tensors/trt_shape_tensor_test.py
```diff
@@ -112,88 +119,83 @@ def check_response(self,
...
+        self.assertEqual(stats.model_stats[0].name, model_name,
+                         "expect model stats for model {}".format(model_name))
+        self.assertEqual(stats.model_stats[0].version, "1",
+                         "expect model stats for model {} version 1".format(model_name))
+        actual_infer_cnt = stats.model_stats[0].inference_stats.success.count
+        self.assertEqual(actual_infer_cnt, infer_cnt,
+                        "expected model-inference-count {}, got {}".format(
+                                infer_cnt, actual_infer_cnt))
+
+        # FIXME Uncomment below after syncing with 'response'.
```
deadeyegoodwin, line 142: "response" confusing here. Say "updated V2 statistics schema from 'response' branch"

qa/L0_trt_shape_tensors/trt_shape_tensor_test.py
```diff
@@ -204,13 +206,14 @@ def test_wrong_shape_values(self):
...
                 np.float32,
-                over_shape_values, [[4, 4]],
+                over_shape_values, [[8, 4, 4]],
                 use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
-                use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY)
-        except InferenceServerException as ex:
-            self.assertEqual("inference:0", ex.server_id())
+                use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY,
+                batch_size=8)
+        # InferenceServerException will be raised from different namespace,
```
deadeyegoodwin, line 214: Do you mean because InferenceServerException comes from grpcclient.utils? @tanmayv25 it is confusing that there is an identical utils.py in both httpclient.utils and grpcclient.utils. Especially when you are trying to use both in the same file. I think you asked this before and I misunderstood but it probably makes sense to just have a tritonclient.utils model that both can depend on. What do you think?

GuanLuo: Depending on the client, the exception will be either `grpcclient.utils.InferenceServerException` or `httpclient.utils.InferenceServerException`, so here uses their comment base class to avoid catching them separately.

I agree, and to make everything into one package, we may have `tritonclient` and have three components: `utils`, `grpc`, `http`. User can choose to build with both or one of grpc and http.

PR #1405, opened 2020-05-05
Fix infer count

GuanLuo:


src/core/infer_stats.cc
```diff
@@ -51,14 +52,14 @@ InferenceStatsAggregator::UpdateFailure(
...
-    const uint64_t request_end_ns)
+    const size_t batch_size, const uint64_t request_start_ns,
+    const uint64_t queue_start_ns, const uint64_t compute_start_ns,
+    const uint64_t compute_input_end_ns, const uint64_t compute_output_start_ns,
+    const uint64_t compute_end_ns, const uint64_t request_end_ns)
 {
   std::lock_guard<std::mutex> lock(mu_);
 
-  infer_stats_.success_count_++;
+  infer_stats_.success_count_ += batch_size;
```
deadeyegoodwin, line 62: Thinking about this more we actually need to keep this as +1 (and also above in failure) since that is the only place where we are recording the number of requests. To know the number of inferences performed the client can use the batch stats. Batch stats also provide the number executions of the model. So it seems like just the client needs to change to use the batch stats to determine both the model "inference count" and "execution count".
For usability we likely want to add inference_count and execution_count to the top-level model infer stats instead of requiring ever client to calculate those values from the batch stats. Thought?

PR #1414, opened 2020-05-06
Reorganizing triton client utilities

tanmayv25:
We will have three packages:
1. tritongrpcclient : [grpcclient]
2. tritonhttpclient: [httpclient]
3. tritonclientutils: [utils, shared_memory and cuda_shared_memory]

### DIFF
--- a/docs/experimental_grpc_python_api.rst
+++ b/docs/experimental_grpc_python_api.rst
@@ -35,7 +35,7 @@ This module contains most of the core functionality of the library
 including setting up a connection, sending requests and receiving
 response to/from an active Triton server.
 
-.. automodule:: tritongrpcclient.grpcclient
+.. automodule:: tritongrpcclient
    :members:
 
 Client Utils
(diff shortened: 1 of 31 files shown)


docs/experimental_grpc_python_api.rst
```diff
@@ -35,13 +35,13 @@ This module contains most of the core functionality of the library
 including setting up a connection, sending requests and receiving
 response to/from an active Triton server.
 
-.. automodule:: tritongrpcclient.core
+.. automodule:: tritongrpcclient.grpcclient
```
deadeyegoodwin, line 38: The naming seems unnecessarily redundant. Is this the pythonic way? What is we just use __init__ for the grpcclient.py and then we would just need:

import tritongrpcclient

docs/experimental_grpc_python_api.rst
```diff
@@ -35,13 +35,13 @@ This module contains most of the core functionality of the library
...
    :members:
 
 Client Utils
 ------------
 
 This module exposes additional supporting utilities.
 
-.. automodule:: tritongrpcclient.utils
+.. automodule:: tritonclientutils.utils
```
deadeyegoodwin, line 46: This is also a little redundant but I guess ok since there are other things in tritonclientutils other than the actual utils

PR #1418, opened 2020-05-06
Fix bug that causes versions to be overwritten in model metadata

CoderHam:


src/core/tritonserver.cc
```diff
@@ -1093,69 +1092,66 @@ TRITONSERVER_ServerModelMetadata(
...
+  metadata.AddMember("name", rapidjson::StringRef(model_name), allocator);
 
   rapidjson::Value versions(rapidjson::kArrayType);
   for (const auto v : ready_versions) {
     auto version_str = std::to_string(v);
-    versions.PushBack(
-        rapidjson::Value(version_str.c_str(), version_str.size()),
-        metadata.GetAllocator());
+    rapidjson::Value version_val(version_str.c_str(), allocator);
+    versions.PushBack(version_val, allocator);
```
GuanLuo, line 1105: @deadeyegoodwin according to rapidjson's doc, it is incorrect to use temporary value in `PushBack()`, or you need to invoke the move schema explicitly via `Move()`. The change here is okay and it implicitly uses the move schema.

src/core/tritonserver.cc
```diff
@@ -1093,69 +1092,66 @@ TRITONSERVER_ServerModelMetadata(
...
+  auto& allocator = metadata.GetAllocator();
   metadata.SetObject();
   // Just store string reference in JSON object since it will be serialized to
   // another buffer.
-  metadata.AddMember(
-      "name", rapidjson::StringRef(model_name), metadata.GetAllocator());
+  metadata.AddMember("name", rapidjson::StringRef(model_name), allocator);
 
   rapidjson::Value versions(rapidjson::kArrayType);
   for (const auto v : ready_versions) {
```
GuanLuo, line 1102: Don't you need to check and only return one version if `model_version` is specified?

CoderHam:
> Looks good but I'm not seeing why this fixes anything. Is it because the rapidjson::Value constructor was taking a reference to the string instead of copying?

Yes. The reference was being over written each time the version_str was reassigned a new value and hence all of the `versions` in the `version` object pointed to the latest string version. If you used to allocator to create a copy of the string variable in a rapidjson::Value object it retains the original value.

src/core/tritonserver.cc
CoderHam: Yes. I did not touch that logic but I can fix it in this PR itself if you think thats a good idea,

src/core/tritonserver.cc
```diff
@@ -1093,69 +1092,72 @@ TRITONSERVER_ServerModelMetadata(
...
+  metadata.AddMember("name", rapidjson::StringRef(model_name), allocator);
 
   rapidjson::Value versions(rapidjson::kArrayType);
-  for (const auto v : ready_versions) {
-    auto version_str = std::to_string(v);
-    versions.PushBack(
-        rapidjson::Value(version_str.c_str(), version_str.size()),
-        metadata.GetAllocator());
+  if (model_version_int != -1) {
+    auto version_str = std::to_string(model_version_int);
```
GuanLuo, line 1103: Need to check if it is one of the ready version...

GuanLuo: My bad, I think `GetInferenceBackend()` will return error in that case. The version should be ready if control reaches here.

PR #1420, opened 2020-05-07
Port L0_server_status and L0_model_config to V2

CoderHam:


GuanLuo:
Have you coordinated with @tanmayv25 on L0_server_status? I think he has a ticket on replacing it to other test due to the status changes in V2

CoderHam:
> Have you coordinated with @tanmayv25 on L0_server_status? I think he has a ticket on replacing it to other test due to the status changes in V2

I wasn't aware of this. I am discussing the same with him now

tanmayv25:
In V1 the server status included the inference statistics, but with V2 the clients report inference statistics in a separate API. More so at client side we don't have any concept of server status.  Hence, I recommended splitting L0_server_status separate tests (L0_infer_statistics, L0_metadata etc,). That said I see the changes in this PR offer the appropriate coverage(metadata). We can split the test in next release if required,

CoderHam:
> In V1 the server status included the inference statistics, but with V2 the clients report inference statistics in a separate API. More so at client side we don't have any concept of server status. Hence, I recommended splitting L0_server_status separate tests (L0_infer_statistics, L0_metadata etc,). That said I see the changes in this PR offer the appropriate coverage(metadata, health, live and inference_statisitcs). We can split the test in next release if required,

I agree. It would make sense to split these tests.

CoderHam:
> The changes looks good for the current test cases. There are however two cases that you can add to improve your coverage for get_inference_statistics:
> 
> 1. model_name without version: server should report the statistics of all the models
> 2. No paramaters ie. without model_name or model_version: The server should report statistics of all the models with all their versions.

Sure. I will add these two cases as well.

qa/L0_model_config/compare_status.py
```diff
@@ -26,49 +26,60 @@
 
 import argparse
 import os
-import os.path
 import sys
-from tensorrtserver.api import *
-import tensorrtserver.api.server_status_pb2 as server_status
-import tensorrtserver.api.model_config_pb2 as model_config
+import json
+import tritonhttpclient.core as httpclient
```
deadeyegoodwin, line 31: These imports don't exist anymore, make sure you rebase to the top of v2test and you will see that this test fails.

qa/L0_model_config/compare_status.py
```diff
@@ -26,49 +26,60 @@
...
-    status_config = status.model_status[FLAGS.model].config
+   parser = argparse.ArgumentParser()
+   parser.add_argument('--expected_dir', type=str, required=True,
+                     help='Directory containing expected output files')
+   parser.add_argument('--model', type=str, required=True,
+                     help='Model name')
+   FLAGS, unparsed = parser.parse_known_args()
 
-    nonmatch = list()
+   # TODO Test HTTP as well. Requires additional handling since it is reported as Json.
```
deadeyegoodwin, line 48: Use FIXME to tag issues. Also instead of this being a loop you probably just want to test them one after the other since as you note the testing code will be different. So you can have the GRPC code be complete and just have a FIXME for needing to add the HTTP part

qa/L0_server_status/server_status_test.py
```diff
@@ -29,80 +29,87 @@
...
+                    triton_client = httpclient.InferenceServerClient(
+                        url=pair[0], verbose=True)
+                else:
+                    triton_client = grpcclient.InferenceServerClient(
+                        url=pair[0], verbose=True)
+
+                self.assertTrue(triton_client.is_server_live())
+                self.assertTrue(triton_client.is_server_ready())
+                server_metadata = triton_client.get_server_metadata()
+                model_metadata = triton_client.get_model_metadata(model_name)
```
deadeyegoodwin, line 94: You expect this call to throw an exception so the only code after here is the "self.assertTrue(False, "expected unknown model failure")". From below you can move the code that tests server_metadata since you expect that call to be successful.

qa/L0_server_status/server_status_test.py
```diff
@@ -111,17 +118,30 @@ def test_model_latest_infer(self):
...
+                        triton_client = httpclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+                    else:
+                        triton_client = grpcclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+
+                    self.assertTrue(triton_client.is_server_live())
+                    self.assertTrue(triton_client.is_server_ready())
+                    model_metadata = triton_client.get_model_metadata(model_name)
+                    model_metadata = triton_client.get_model_metadata(
```
deadeyegoodwin, line 132: Put 132-140 in the loop below so that you check the metadata for all 3 versions. You should check that the metadata name and version are both what you expect.

qa/L0_server_status/server_status_test.py
```diff
@@ -111,17 +118,30 @@ def test_model_latest_infer(self):
...
+                    if pair[1] == "http":
+                        triton_client = httpclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+                    else:
+                        triton_client = grpcclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+
+                    self.assertTrue(triton_client.is_server_live())
+                    self.assertTrue(triton_client.is_server_ready())
+                    model_metadata = triton_client.get_model_metadata(model_name)
```
deadeyegoodwin, line 131: Since you didn't specify version here you should be the metadata for the latest version, correct? You should add some assertions to check that.

qa/L0_server_status/server_status_test.py
```diff
@@ -131,57 +151,94 @@ def test_model_latest_infer(self):
...
+                    if pair[1] == "http":
+                        triton_client = httpclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+                    else:
+                        triton_client = grpcclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+
+                    self.assertTrue(triton_client.is_server_live())
+                    self.assertTrue(triton_client.is_server_ready())
+                    model_metadata = triton_client.get_model_metadata(
```
deadeyegoodwin, line 164: same comments as above

qa/L0_server_status/server_status_test.py
```diff
@@ -131,57 +151,94 @@ def test_model_latest_infer(self):
...
-                                            "expected 1 infer stats for v" + str(v) + " model " + model_name)
-                            self.assertTrue(1 in version_status.infer_stats,
-                                            "expected batch 1 status for v" + str(v) + " model " + model_name)
-                            infer_stats = version_status.infer_stats[1]
-                            self.assertTrue(infer_stats.success.count, 1)
+                            if pair[1] == "http":
+                                self.assertEqual(len(infer_stats['model_stats']), 1,
+                                                "expected 1 infer stats for version " + str(v) +
+                                                 " of model " + model_name)
+                                stats = infer_stats['model_stats'][0]['inference_stats']
```
deadeyegoodwin, line 184: [0]  You can't rely on the model stats being in order by version. You should instead verify that you get back 3 sets of stats and then loop over those stats. If the stat is for version 3 then check for non-zero success count, if version 1 or 2 check for zero success count.

qa/L0_server_status/server_status_test.py
```diff
@@ -131,57 +151,94 @@ def test_model_latest_infer(self):
...
+                                self.assertEqual(len(stats), 6,
```
deadeyegoodwin, line 185: inference_stats is not repeated

qa/L0_server_status/server_status_test.py
```diff
@@ -131,57 +151,94 @@ def test_model_latest_infer(self):
...
+                    if pair[1] == "http":
+                        triton_client = httpclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+                    else:
+                        triton_client = grpcclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+
+                    self.assertTrue(triton_client.is_server_live())
+                    self.assertTrue(triton_client.is_server_ready())
+                    model_metadata = triton_client.get_model_metadata(model_name)
```
deadeyegoodwin, line 232: Why have this metadata read and partial check, seems unrelated to why you are trying to test.

qa/L0_server_status/server_status_test.py
```diff
@@ -131,57 +151,94 @@ def test_model_latest_infer(self):
...
+
+                    if pair[1] == "http":
+                        self.assertEqual(model_name, model_metadata['name'])
+                    else:
+                        self.assertEqual(model_name, model_metadata.name)
+    
                     for v in (1, 3):
```
deadeyegoodwin, line 239: Also test that version 2 is not ready. Don't need a loop just 3 calls to is_model_ready

qa/L0_server_status/server_status_test.py
```diff
@@ -191,71 +248,61 @@ def test_model_specific_infer(self):
...
+                    if pair[1] == "http":
+                        triton_client = httpclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+                    else:
+                        triton_client = grpcclient.InferenceServerClient(
+                            url=pair[0], verbose=True)
+
+                    self.assertTrue(triton_client.is_server_live())
+                    self.assertTrue(triton_client.is_server_ready())
+                    model_metadata = triton_client.get_model_metadata(model_name)
```
deadeyegoodwin, line 261: above comments

qa/L0_server_status/server_status_test.py
```diff
@@ -191,71 +248,61 @@ def test_model_specific_infer(self):
...
+
+                    if pair[1] == "http":
+                        self.assertEqual(model_name, model_metadata['name'])
+                    else:
+                        self.assertEqual(model_name, model_metadata.name)
+
                     for v in (1, 3):
-                        self.assertTrue(v in server_status1.model_status[model_name].version_status,
```
deadeyegoodwin, line 201: instead of loop just test that v1 has stastics and v3 doesn't. Look in lifecycle_test.py for examples of how to do this.

qa/L0_model_config/compare_status.py
CoderHam: Fixed by #1424. No additional handling is needed anymore

qa/L0_server_status/server_status_test.py
CoderHam: @deadeyegoodwin The metadata reports all the versions in this case. Should it not be reporting all the versions?

deadeyegoodwin: The spec here is a little strange. For triton our metadata is always the same no matter the version requested. So I would say that the version that is in the request we should just use to see if the entire request should pass/fail (that is, if version is given and that version isn't actually available then fail). But otherwise the actual content returned for the request should always be the same, no matter if a version if requested or not.

CoderHam: To clarify -> This would mean that the content for each valid version/if no version is specified should contain all the versions that the model supports (and not just the specified/latest version)?

deadeyegoodwin: The versions listed in the metadata response are just a list of the version numbers that are available for that model. It depends on the version policy of the model. I'm saying that the requested version (if any) does not change what we return in that version list. The requesed version (if any) should just be used to fail the request if that version is not valid for the model. So, in practice, with Triton it doesn't make much sense to use a version in the metadata call (but we should still test it).

qa/L0_server_status/server_status_test.py
```diff
@@ -111,17 +117,42 @@ def test_model_latest_infer(self):
...
+                            self.assertTrue(str(v) in model_metadata.versions)
+
                     for v in (1, 2, 3):
-                        self.assertTrue(v in server_status0.model_status[model_name].version_status,
-                                        "expected version " + str(v) + " status for model " + model_name)
-                        self.assertEqual(server_status0.model_status[model_name].version_status[v].ready_state,
-                                        server_status.MODEL_READY)
+                        self.assertTrue(triton_client.is_model_ready(
+                            model_name, model_version=str(v)))
+                        model_metadata = triton_client.get_model_metadata(
```
deadeyegoodwin, line 146: Based on comments above this isn't going to work. We likely don't need so much testing around model metadata since it really is fairly simple. Really just need to make sure it returns the expected versions, input and output information. And that if we do call it with an unknown version that is fails.

CoderHam: Add tests for model metadata contents and unknown model version. We might need a fix to the model_repository_manager to have a more meaningful error message when sending an incorrect version. I reports the error message as `"Request for unknown model <model_name>"` in the case the `model_version` is unknown even if the `model_name` is valid.

deadeyegoodwin: OK, please file a 20.06 ticket for that.

CoderHam: Done

PR #1422, opened 2020-05-07
Fix base64 encoding to null terminate the string

tanmayv25:
The encoded base64 string was not getting null terminated. Fixed the termination to let the libraries detect the end.
I tested with the perf_client and it worked fine,
@CoderHam Can you try this fix with L0_* cuda shared memory tests?

### DIFF
--- a/src/clients/c++/experimental_api_v2/library/cencode.c
+++ b/src/clients/c++/experimental_api_v2/library/cencode.c
@@ -102,7 +102,7 @@ int base64_encode_blockend(char* code_out, base64_encodestate* state_in)
 	case step_A:
 		break;
 	}
-	*codechar++ = '\0';
+	*codechar++ = '\n';
 	
 	return codechar - code_out;
 }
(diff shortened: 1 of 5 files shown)


src/clients/c++/experimental_api_v2/library/cencode.c
```diff
@@ -102,7 +102,7 @@ int base64_encode_blockend(char* code_out, base64_encodestate* state_in)
 	case step_A:
 		break;
 	}
-	*codechar++ = '\n';
+	*codechar++ = '\0';
```
deadeyegoodwin, line 105: It would be better not to change this file. It seems that we just need to be careful to always use the returned size and not rely on an end-of-string indicator. It seems you made those changes below so can we no do this change.

src/clients/c++/experimental_api_v2/library/http_client.cc
```diff
@@ -91,14 +91,17 @@ GetQueryString(const Headers& query_params)
 }
 
 void
-Base64Encode(char* raw_ptr, size_t raw_size, char** encoded_ptr)
+Base64Encode(
```
deadeyegoodwin, line 94: Make sure that documentation for this clearly states that the returned encoded_ptr is not null terminated and that encoded_size must be used to determine length of the string.

CoderHam:
I tried the fix and it does pass the L0_cuda_shared_memory test. Thanks!

PR #1424, opened 2020-05-07
Ensure naming of protobuf message fields is preserved when converting to JSON

CoderHam:
Enable preserve_proto_field_names

### DIFF
--- a/src/clients/python/experimental_api_v2/library/grpcclient.py
+++ b/src/clients/python/experimental_api_v2/library/grpcclient.py
@@ -270,7 +270,7 @@ class InferenceServerClient:
             if self._verbose:
                 print(response)
             if as_json:
-                return json.loads(MessageToJson(response), preserving_proto_field_nam=True)
+                return json.loads(MessageToJson(response, preserving_proto_field_name=True))
             else:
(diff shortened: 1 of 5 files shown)


src/clients/python/experimental_api_v2/library/grpcclient.py
```diff
@@ -270,7 +270,7 @@ def get_server_metadata(self, headers=None, as_json=False):
             if self._verbose:
                 print(response)
             if as_json:
-                return json.loads(MessageToJson(response))
+                return json.loads(MessageToJson(response), preserving_proto_field_nam=True)
```
GuanLuo, line 273: `preserving_proto_field_name` should be parameter for `MessageToJson()`?

deadeyegoodwin: nam -> name
@GuanLuo are you referring to that misspelling?

CoderHam: Fixed. That and the fact that `preserving_proto_field_name` is a param to `MessageToJson()` and not `json.loads()`

PR #1433, opened 2020-05-07
Fix V2 python image client to handle non-batching models correctly

deadeyegoodwin:


src/clients/python/experimental_api_v2/examples/v2_image_client.py
```diff
@@ -232,13 +238,17 @@ def postprocess(results, output_name, batch_size):
...
 
-def requestGenerator(input_name, output_name, c, h, w, format, dtype, FLAGS):
+def requestGenerator(supports_batching, input_name, output_name, c, h, w, format, dtype, FLAGS):
     # Preprocess image into input data according to model requirements
     image_data = None
     with Image.open(FLAGS.image_filename) as img:
         image_data = preprocess(img, format, dtype, c, h, w, FLAGS.scaling)
 
-    repeated_image_data = [image_data for _ in range(FLAGS.batch_size)]
+    if not supports_batching:
```
CoderHam, line 247: Why not use np.repeat along a new axis here instead?

deadeyegoodwin: Not sure I understand. With no batching we just want the input to be the image [ h, w, c ]. We don't want to repeat it even once since that was giving us [ 1, h, w, c ].

PR #1435, opened 2020-05-07
Allow HTTP server to report complex error messages

tanmayv25:
The error messages can be complex sometimes. Relying on the rapidjson to do correct encoding, 

### After this fix :
root@27ab6cfd09e0:/opt/tritonserver/qa/clients# ../clients/perf_client_v2 -v -i http -m graphdef_sequence_float32 --shape INPUT:2 --input-data=/opt/tritonserver/qa/L0_perf_client/json_input_data_files/float_data_with_shape.json --input-data=/opt/tritonserver/qa/L0_perf_client/json_input_data_files/float_data_with_shape.json -p2000 --shared-memory=none
 Successfully read data for 6 stream/streams.
*** Measurement Settings ***
  Batch size: 1
  Measurement window: 2000 msec
  Using asynchronous calls for inference

Request concurrency: 1
  Pass [1] throughput: 0.5 infer/sec. Avg latency: 9353 usec (std 0 usec)
**Failed to retrieve results from inference request.
Thread [0] had error: 2 root error(s) found.
  (0) Invalid argument: Inputs to operation Select of type Select must have the same size and shape.  Input 0: [1,1] != input 1: [1,2]
         [[{{node Select}}]]
         [[OUTPUT/_7]]
  (1) Invalid argument: Inputs to operation Select of type Select must have the same size and shape.  Input 0: [1,1] != input 1: [1,2]
         [[{{node Select}}]]
0 successful operations.
0 derived errors ignored.**

### DIFF
--- a/src/servers/http_server_v2.cc
+++ b/src/servers/http_server_v2.cc
@@ -868,9 +868,16 @@ WriteDataToJson(
 void
 EVBufferAddErrorJson(evbuffer* buffer, TRITONSERVER_Error* err)
 {
+  rapidjson::Document response;
+  response.SetObject();
   std::string message = std::string(TRITONSERVER_ErrorMessage(err));
-  std::string message_json = "{ \"error\" : \"" + message + "\" }";
-  evbuffer_add(buffer, message_json.c_str(), message_json.size());
+  rapidjson::Value message_json(
+      rapidjson::StringRef(message.c_str(), message.size()),
+      response.GetAllocator());
+  rapidjson::StringBuffer buffer_json;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer_json);
+  response.Accept(writer);
+  evbuffer_add(buffer, buffer_json.GetString(), buffer_json.GetSize());
 }
 
 void
--- a/src/servers/http_server_v2.cc
+++ b/src/servers/http_server_v2.cc
@@ -874,6 +874,7 @@ EVBufferAddErrorJson(evbuffer* buffer, TRITONSERVER_Error* err)
   rapidjson::Value message_json(
       rapidjson::StringRef(message.c_str(), message.size()),
       response.GetAllocator());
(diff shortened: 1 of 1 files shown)


src/servers/http_server_v2.cc
```diff
@@ -868,9 +868,16 @@ WriteDataToJson(
...
 EVBufferAddErrorJson(evbuffer* buffer, TRITONSERVER_Error* err)
 {
+  rapidjson::Document response;
+  response.SetObject();
   std::string message = std::string(TRITONSERVER_ErrorMessage(err));
-  std::string message_json = "{ \"error\" : \"" + message + "\" }";
-  evbuffer_add(buffer, message_json.c_str(), message_json.size());
+  rapidjson::Value message_json(
+      rapidjson::StringRef(message.c_str(), message.size()),
+      response.GetAllocator());
```
deadeyegoodwin, line 876: You don't have to explicitly put message_json in the 'response'? Just using the 'response' allocator causes it to be added?

CoderHam: Why did you remove "error" field from the json response? Doesn't the client expect
`{error: <error_message>}` and not `<error_message>`
Also to answer your question David, when an allocator is passed to rapidjson:Value() it creates a copy of the string. When the allocator is not passed it uses the address of the string itself and does not create a copy of it.

deadeyegoodwin: Protocol requires a json response that is { "error" : <msg> }, so that part is correct. I just don't see where message_json gets added to 'response'. Perhaps I'm just missing it.

CoderHam: Tanmay had missed a line that does exactly that -> `response.AddMember("error", message_json, response.GetAllocator());`
I fixed it

tanmayv25: My bad. The line got accidentally dropped while committing. Thanks @CoderHam for fixing it.

PR #1437, opened 2020-05-08
Fix value type requested for rapidjson value

GuanLuo:


GuanLuo:
After some thinking about it, we actually need to be more defensive on using rapidjson,  I will file a ticket regarding that.

Currently this ticket is a minimal fix for v2 L0_large_payload, I am okay with either merging it just for the fix or postponing it to be PR for the upcoming ticket mentioned above.

PR #1443, opened 2020-05-08
Use python3-opencv to install opencv

deadeyegoodwin:


CoderHam:
Is there a reason we want to move to the python3-opencv version?

Dockerfile.QA
```diff
@@ -45,8 +45,7 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 # image_preprocess custom backend requires opencv
```
GuanLuo, line 46: If we are building custom backend here, does `python3-opencv` satisfy the dependency on C++ side?

deadeyegoodwin:
> Is there a reason we want to move to the python3-opencv version?

There was but now not a priority. L1_contamer was failing because of python 2 issues and it turns out that python2 is getting into the container from opencv. But even using python3-opencv doesn't seem to fix... and I fixed the contamer issue in another way. I think python3-opencv is more future proof (hopefully in ubuntu 20.04 there will be no python2 anymore).

PR #1450, opened 2020-05-08
Fix test utils

GuanLuo:
This PR cherry-picked changes from #1449 to kick off a up-to-date CI, will remove the cherry-picked commit after CI passes build stage


GuanLuo:
@tanmayv25 I don't think we have tests to reuse InferInput / InferRequestedOutput objects with different set functions.

PR #1455, opened 2020-05-11
Update client build instructions

dzier:


docs/client.rst
```diff
@@ -174,6 +181,14 @@ build does not include the C++ examples.
 The MSBuild.exe may need to be invoked twice for a successfull
 build.
 
+A wheel file may fail to generate if the python environment is not set up 
+correctly. The windows build environment will not fail correctly if there is a 
+failure when running the python script. To verify that wheel file is executed
```
deadeyegoodwin, line 186: ".. is generated correctly..."

This is confusing. Is the command actually generating the wheel file? If so you should state more clearly that the command is generating the wheel (not just verifying it), and also say where the resulting wheel file can be found.

docs/client.rst
```diff
@@ -174,6 +181,17 @@ build does not include the C++ examples.
 The MSBuild.exe may need to be invoked twice for a successfull
 build.
 
+A wheel file may fail to generate if the python environment is not set up 
+correctly. The windows build environment will not fail correctly if there is a 
+failure when running the python script. To verify that wheel file is generated
```
deadeyegoodwin, line 186: There is no reason for users to run the below build command, correct? I'd prefer not to give them the below command which builds a wheel without any version. Can't we instead tell them that after the build completes they should look in install directory xyz for the wheel file. If it is not there then something went wrong in the build and they should check the build log? Did you look into why cmake is silently allowing a failed build to continue? Isn't there some way to fix that?

JulianOrteil: If I may also jump in here, by default, when Python is installed via the installer, it does not create a "python3.exe" file as noted in this pull request at the very top. Could that be fixed in the cmake files?

dzier: Cmake launches a bash window. That bash window does not exit with the return code of the command it executed. I can look into more and see if there is a better way. I was providing a debugging help here. You asked to include the comments I replied with in the Issue. I am OK with leaving this out all together. Or keeping it until we figure out how to get CMAKE to exit correctly on a bash script error.

dzier:
Removed the strange requirement on renaming python. Will create another PR that looks for the correct python executable.

PR #1462, opened 2020-05-11
Fix model repository error message for incorrect version

CoderHam:
**Old Error Message:**
Missing/Incorrect model_name/model_version: `"Request for unknown model 'graphdef_float32_float32_float32'"`

**New Error Message:**
Missing/Incorrect model_name: `"Request for unknown model: 'graphdef_float32_float32_float32' is not found"`
Model name correct but version=-1 and no version is available anymore: `"Request for unknown model: 'graphdef_float32_float32_float32' has no available versions"`
Missing/Incorrect model_version: `"Request for unknown model: 'graphdef_float32_float32_float32' version 1 is not found"`

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -196,7 +196,7 @@ class LifeCycleTest(unittest.TestCase):
             self.assertEqual("inference:0", ex.server_id())
             self.assertTrue(
                 ex.message().startswith(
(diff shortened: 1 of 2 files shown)


CoderHam:
Fix test failure in cases where model version is not passed no model versions are present. Missing model version is represented by `version = -1`. The error message earlier returned was `'model_name' version -1 is not found` which did not make sense. Now it returns an error that the model has no available versions.

PR #1464, opened 2020-05-12
Fix HTTP async lib and support async and streaming in python image cl…

tanmayv25:
…ient

### DIFF
--- a/src/clients/python/experimental_api_v2/examples/v2_image_client.py
+++ b/src/clients/python/experimental_api_v2/examples/v2_image_client.py
@@ -45,7 +45,6 @@ else:
 
 
 class UserData:
-
     def __init__(self):
         self._completed_requests = queue.Queue()
 
--- a/src/clients/python/experimental_api_v2/examples/v2_image_client.py
+++ b/src/clients/python/experimental_api_v2/examples/v2_image_client.py
@@ -29,12 +29,32 @@ import argparse
 import numpy as np
 from PIL import Image
 import sys
+from functools import partial
+import gevent
 
 import tritongrpcclient
 import tritongrpcclient.model_config_pb2 as mc
 import tritonhttpclient
 from tritonclientutils.utils import triton_to_np_dtype
 from tritonclientutils.utils import InferenceServerException
+
+if sys.version_info >= (3, 0):
+    import queue
+else:
+    import Queue as queue
+
+
+class UserData:
+
+    def __init__(self):
+        self._completed_requests = queue.Queue()
+
+
+# Callback function used for async_stream_infer()
+def completion_callback(user_data, result, error):
+    # passing error raise and handling out
+    user_data._completed_requests.put((result, error))
+
 
 FLAGS = None
 
@@ -100,9 +120,9 @@ def parse_model_grpc(model_metadata, model_config):
         h = input_metadata.shape[1]
         w = input_metadata.shape[2]
 
-    return (model_config.max_batch_size,
-            input_metadata.name, output_metadata.name, c, h, w,
-            input_config.format, input_metadata.datatype)
+    return (model_config.max_batch_size, input_metadata.name,
+            output_metadata.name, c, h, w, input_config.format,
+            input_metadata.datatype)
 
 
 def parse_model_http(model_metadata, model_config):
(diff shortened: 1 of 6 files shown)


src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -1120,6 +1120,11 @@ def async_infer(self,
             Optional url query parameters to use in network
             transaction.
 
+        Returns
+        -------
+        Greenlet object
```
deadeyegoodwin, line 1125: Period at end. Is there a link to what a Greenlet is that you could add here?

tanmayv25: Added the link.

src/clients/python/experimental_api_v2/examples/v2_image_client.py
```diff
@@ -29,13 +29,33 @@
...
 from tritonclientutils.utils import InferenceServerException
 
+if sys.version_info >= (3, 0):
+    import queue
+else:
+    import Queue as queue
+
+
+class UserData:
+
```
CoderHam, line 48: Remove extra line

src/clients/python/experimental_api_v2/examples/v2_image_client.py
```diff
@@ -373,20 +408,73 @@ def requestGenerator(supports_batching, input_name, output_name, c, h, w, format
...
+                                        model_version=FLAGS.model_version,
+                                        outputs=outputs))
 
     except InferenceServerException as e:
         print("inference failed: " + str(e))
-        sys.exit()
+        if FLAGS.streaming:
+            triton_client.stop_stream()
+        sys.exit(1)
+
```
CoderHam, line 462: I'm curious, what would the behavior be if the program exits before stop_stream is called?

tanmayv25: The stream will get stuck and system will hang. Python, unfortunately, is not very adept at cleanups(destructor logic). If you look at our library we use an iterator class to keep the stream up and running. The order of clean up is not guaranteed by the python's garbage collector. We can only clean the client object when the stream itself is closed, which happens when None object is passed into the stream. If the None object is not passed, then the stream is kept open preventing client from getting deleted.

GuanLuo: "We can only clean the client object when the stream itself is closed", so the client's __del__ function will never be called unless the stream is closed? Otherwise, can we call stop_stream() in the __del__()?

tanmayv25: We have kept stop_stream in the del(). But the issue is python garbage collector will not call the del() until client ref is zero. While sending the requests on the grpc stream, grpc library copies this ref hence incrementing the counter. This leads to deadlock. Client's del() will not be called unless the stream is closed, and stream will not be closed unless None is passed by the stop_stream(). 

To avoid this precarious situation, it is best for user to ensure the stop_stream() is explicitly called.

src/clients/python/experimental_api_v2/examples/v2_image_client.py
```diff
@@ -373,20 +407,73 @@ def requestGenerator(supports_batching, input_name, output_name, c, h, w, format
...
+            elif FLAGS.async_set:
+                if FLAGS.protocol.lower() == "grpc":
+                    triton_client.async_infer(FLAGS.model_name,
+                                              inputs,
+                                              partial(completion_callback,
+                                                      user_data),
+                                              model_version=FLAGS.model_version,
+                                              outputs=outputs)
+                else:
+                    greenlets.append(
```
deadeyegoodwin, line 442: Ideally we should not expose gevent/greenlet through our API since it ties us to that implementation. Instead we could wrap the greenlet in our own class. But why return the greenlet? We already have a callback mechanism to indicate completion. Can we just bind the callback to the greenlet completion (there appears to be an API for that)?

tanmayv25: We were calling the gevent version of sleep to relinquish control to underlying greenlets. These work with cooperative multi-tasking. We do have attached callback. But the callback will not be invoked if the current greenlet(the main execution stream) does not relinquish control. We were doing that with gevent.sleep(). This kind of exposes the underlying greenlet implementation but will lead to continuous polling. The join() call on the underlyinh greenlet is less cpu intensive. I can fallback to our previous gevent.sleep() implementation. The thing is the greenlet will not execute until and unless someone releases the control to it.

PR #1466, opened 2020-05-12
Fix server readiness

GuanLuo:


GuanLuo:
@deadeyegoodwin This is v2-only equivalent of #1465

PR #1469, opened 2020-05-12
Add Custom Backend coverage to S3

CoderHam:
Custom backend did not support loading of model from S3. This PR fixes support for it and adds test coverage for custom backend models. 

Fix bug that causes incomplete deletion of models.

### DIFF
--- a/src/backends/custom/custom_backend_factory.cc
+++ b/src/backends/custom/custom_backend_factory.cc
@@ -57,24 +57,25 @@ CustomBackendFactory::CreateBackend(
     const double min_compute_capability,
     std::unique_ptr<InferenceBackend>* backend)
 {
-  const auto path =
+  const auto nonlocal_path =
       JoinPath({model_repository_path, model_name, std::to_string(version)});
 
-  // Read all the files in 'path'.
+  // Localize 'nonlocal_path' so that the entire model version
+  // directory is available locally.
+  std::shared_ptr<LocalizedDirectory> local_dir;
+  RETURN_IF_ERROR(LocalizeDirectory(nonlocal_path, &local_dir));
+
+  // Read all the files in 'local_dir'.
   std::set<std::string> custom_files;
-  RETURN_IF_ERROR(
-      GetDirectoryFiles(path, true /* skip_hidden_files */, &custom_files));
+  RETURN_IF_ERROR(GetDirectoryFiles(
+      local_dir->Path(), true /* skip_hidden_files */, &custom_files));
 
   std::unordered_map<std::string, std::string> custom_paths;
-  std::vector<std::shared_ptr<LocalizedDirectory>> local_custom_path;
-
   for (const auto& filename : custom_files) {
-    const auto custom_path = JoinPath({path, filename});
-    local_custom_path.push_back(std::shared_ptr<LocalizedDirectory>(nullptr));
-    RETURN_IF_ERROR(LocalizeDirectory(custom_path, &local_custom_path.back()));
+    const auto custom_path = JoinPath({local_dir->Path(), filename});
     custom_paths.emplace(
         std::piecewise_construct, std::make_tuple(filename),
-        std::make_tuple(local_custom_path.back()->Path()));
+        std::make_tuple(custom_path));
   }
 
   // Create the vector of server parameter values, indexed by the
@@ -89,7 +90,8 @@ CustomBackendFactory::CreateBackend(
   // requested for this model.
   std::unique_ptr<CustomBackend> local_backend(
       new CustomBackend(min_compute_capability));
-  RETURN_IF_ERROR(local_backend->Init(path, server_params, model_config));
+  RETURN_IF_ERROR(
+      local_backend->Init(nonlocal_path, server_params, model_config));
   RETURN_IF_ERROR(local_backend->CreateExecutionContexts(custom_paths));
 
   *backend = std::move(local_backend);
--- a/src/backends/custom/custom_backend_factory.cc
+++ b/src/backends/custom/custom_backend_factory.cc
@@ -68,10 +68,13 @@ CustomBackendFactory::CreateBackend(
   std::unordered_map<std::string, std::string> custom_paths;
   for (const auto& filename : custom_files) {
     const auto custom_path = JoinPath({path, filename});
+    std::string local_custom_path;
 
+    RETURN_IF_ERROR(
+        DownloadFileFolder(custom_path, &local_custom_path));
     custom_paths.emplace(
         std::piecewise_construct, std::make_tuple(filename),
-        std::make_tuple(custom_path));
+        std::make_tuple(local_custom_path));
   }
 
   // Create the vector of server parameter values, indexed by the
@@ -88,6 +91,11 @@ CustomBackendFactory::CreateBackend(
       new CustomBackend(min_compute_capability));
   RETURN_IF_ERROR(local_backend->Init(path, server_params, model_config));
   RETURN_IF_ERROR(local_backend->CreateExecutionContexts(custom_paths));
+
+  // Destroy local copy if exists
+  for (const auto& custom_path : custom_paths) {
+    RETURN_IF_ERROR(DestroyFileFolder(custom_path.second));
+  }
 
   *backend = std::move(local_backend);
   return Status::Success;
--- a/src/backends/custom/custom_backend_factory.cc
+++ b/src/backends/custom/custom_backend_factory.cc
@@ -71,10 +71,10 @@ CustomBackendFactory::CreateBackend(
   for (const auto& filename : custom_files) {
     const auto custom_path = JoinPath({path, filename});
     local_custom_path.push_back(std::shared_ptr<LocalizedDirectory>(nullptr));
-    RETURN_IF_ERROR(LocalizeFileFolder(custom_path, &local_custom_path.back()));
+    RETURN_IF_ERROR(LocalizeDirectory(custom_path, &local_custom_path.back()));
     custom_paths.emplace(
         std::piecewise_construct, std::make_tuple(filename),
-        std::make_tuple(local_custom_path.back()->local_path_));
+        std::make_tuple(local_custom_path.back()->Path()));
   }
 
   // Create the vector of server parameter values, indexed by the
--- a/src/backends/custom/custom_backend_factory.cc
+++ b/src/backends/custom/custom_backend_factory.cc
@@ -66,20 +66,15 @@ CustomBackendFactory::CreateBackend(
       GetDirectoryFiles(path, true /* skip_hidden_files */, &custom_files));
 
   std::unordered_map<std::string, std::string> custom_paths;
-  std::vector<std::shared_ptr<TemporaryDirectory>> local_custom(
-      custom_files.size());
-  for (size_t s = 0; s < custom_files.size(); s++) {
-    local_custom[s] = std::make_shared<TemporaryDirectory>("");
-  }
+  std::vector<std::shared_ptr<LocalizedDirectory>> local_custom_path;
 
-  int i = 0;
   for (const auto& filename : custom_files) {
     const auto custom_path = JoinPath({path, filename});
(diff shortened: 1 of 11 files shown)


src/backends/custom/loader.cc
```diff
@@ -72,17 +74,28 @@ LoadCustom(
   *ExecuteFn = nullptr;
   *ExecuteV2Fn = nullptr;
   *custom_version = 0;
+  Status status;
+
+  // Add handling for S3 path
+  // FIXME Add GCS Support
+  std::string local_path = "";
+  if (path.rfind("s3://", 0) == 0) {
```
GuanLuo, line 82: You can just call `DownloadFileFolder()` without checking the "path type"

src/backends/custom/loader.cc
```diff
@@ -72,17 +74,28 @@ LoadCustom(
...
+    status = DownloadFileFolder(path, &local_path);
```
GuanLuo, line 83: Why are you downloading files to local here instead of in the backend factory like for TF? Does the file need to be presented even after dlopen()? If so, remember to destroy the local folder in unload custom. If not, perhaps just follow the pattern for TF

CoderHam: Fixed. Yes the file can be deleted after dlopen is called.

src/core/filesystem.cc
```diff
@@ -1269,7 +1319,16 @@ Status
 DestroyFileFolder(const std::string& path)
 {
   FileSystem* fs;
-  RETURN_IF_ERROR(GetFileSystem(path, &fs));
+  std::cerr << "path: " << path << "\n";
```
GuanLuo: Remove debug message

src/core/filesystem.cc
```diff
@@ -1269,7 +1319,19 @@ Status
...
 {
   FileSystem* fs;
-  RETURN_IF_ERROR(GetFileSystem(path, &fs));
+  // If path represents local temporary file then must be S3
+  if (path.rfind("/tmp/file", 0) == 0) {
+#ifdef TRTIS_ENABLE_S3
+    Aws::SDKOptions options;
+    Aws::InitAPI(options);
+    static S3FileSystem s3_fs(options);
+    fs = &s3_fs;
```
GuanLuo, line 1328: In fact we actually don't care about the file system type, we will delete the folder if and only if the path is to our designated tmp path. We just need to check if the path is the designated path (we know it must be in local file system), then use the `LocalFileSystem::DestroyFileFolder()` to do the right thing.

GuanLuo: You just need to move what is in the `S3FileSystem::DestroyFileFolder()` over to `LocalFileSystem`. Currently it is a little bit strange because "S3FileSystem" is deleting local files...

src/core/filesystem.cc
```diff
@@ -1029,10 +1035,54 @@ S3FileSystem::DownloadFileFolder(
...
+  struct stat s_buf;
+
+  if (stat(path, &s_buf))
+    return 0;
+
+  return S_ISDIR(s_buf.st_mode);
+}
+
+int
+DeleteFolderRecursive(const std::string& path)
```
GuanLuo, line 1050: Move these two functions to be `XXFileSystem` class members. And only implement in LocalFileSystem

src/core/filesystem.cc
CoderHam: Fixed

src/core/filesystem.cc
CoderHam: Fixed

src/core/filesystem.cc
```diff
@@ -1269,8 +1348,14 @@ Status
 DestroyFileFolder(const std::string& path)
 {
   FileSystem* fs;
-  RETURN_IF_ERROR(GetFileSystem(path, &fs));
-  return fs->DestroyFileFolder(path);
+  // If path represents local temporary file then must be deleted
+  if (path.rfind("/tmp/file", 0) == 0) {
+    static LocalFileSystem lfs;
+    fs = &lfs;
+    return fs->DestroyFileFolder(path);
```
GuanLuo, line 1355: remove `fs` and directly call `return lfs.DestroyFileFolder(path);`. Otherwise looks good

CoderHam: Done

src/backends/custom/loader.cc
```diff
@@ -27,6 +27,7 @@
 #include "src/backends/custom/loader.h"
 
 #include <dlfcn.h>
+
```
deadeyegoodwin, line 30: Remove blank line. We don't separate system and local includes...

src/core/filesystem.cc
```diff
@@ -224,10 +228,54 @@ LocalFileSystem::DownloadFileFolder(
...
+
+  closedir(dp);
+  return rmdir(path.c_str());
+}
+
 Status
 LocalFileSystem::DestroyFileFolder(const std::string& path)
 {
-  // Do nothing
+  if (IsPathDirectory(path.c_str())) {
```
deadeyegoodwin, line 269: You are deleting the local copy of the model?!? Isn't this going to delete the model repos? Including our repos on dldata? This method should be named something like ReleaseDownloadFileFolder to make it clear that it is not (always) deleting/destroying.

src/backends/custom/loader.cc
CoderHam: Removed

src/core/filesystem.cc
CoderHam: This is only deleted in the case of the path matching the format of `"/tmp/file*"` i.e. in the case the file is a local copy of a file in a cloud repository.

CoderHam: Fixed naming to ReleaseDownloadFileFolder

deadeyegoodwin: Why make a copy of a local filesystem model? Why is the custom backend any different than other cases?

CoderHam: The behavior for Custom Backend is the same as what it was for other backends and has not changed. What has changed (fixed) is the cleanup of temporary files that are created when downloading from S3. 

The ReleaseDownloadFileFolder of the local filesystem clears models that were created in by calling DownloadFileFolder in S3. It is confusing as first but as @Guan mentioned the file being released resides in the local filesystem and not S3.

deadeyegoodwin: I think it was confusing because of the naming of the functions. Even DownloadFileFolder should probably change to LocalizeFileFolder or similar. You can have LocalizeFileFolder return a shared_ptr object which represents the temp local copy and then have the destructor of that object remove the directory... then you wouldn't need ReleaseDownloadFileFolder and the ownership and functionality would be clearer.

deadeyegoodwin: The new class/object can be called TemporaryDirectory or similar. This will also fix some bugs we have now where if errors occur the tmpdir is not cleaned up because DestroyFileFolder is not called.

CoderHam: Sure. I will make these changes

src/core/filesystem.h
```diff
@@ -25,12 +25,65 @@
...
 #include <string>
 #include "google/protobuf/message.h"
 #include "src/core/status.h"
 
 namespace nvidia { namespace inferenceserver {
 
+struct TemporaryDirectory {
+  std::string model_path;
+
+  TemporaryDirectory(std::string path) : model_path(path) {}
```
deadeyegoodwin, line 40: I don't understand this class. It should represent a directory... despite what I said earlier I would call it LocalizedDirectory. It should have a method where you can get the path to the directory (not a model_path... just a path). It should have a way to be constructed "in place" by giving it the path... LocalizedDirectory constructed this way doesn't need to do any cleanup on destruction. Another constructor takes no argument and so it creates a temporary directory and on destruction deletes that directory.

src/core/filesystem.h
```diff
@@ -25,12 +25,65 @@
...
+
+  ~TemporaryDirectory()
+  {
+    // Only delete file/folder if it is a local copy created from a Cloud
+    // repository
+    if (model_path.rfind("/tmp/file", 0) == 0) {
```
deadeyegoodwin, line 46: Hacky, don't depend on file path...

src/core/filesystem.cc
```diff
@@ -541,19 +535,13 @@ GCSFileSystem::ReadTextFile(const std::string& path, std::string* contents)
 }
 
 Status
-GCSFileSystem::DownloadFileFolder(
-    const std::string& path, std::string* local_path)
-{
-  // For GCS we don't actually need to download the folder. We use tensorflow's
-  // built in support for loading models directly from GCS.
```
deadeyegoodwin, line 548: Not sure how I missed this before... we need full support for GCS for all models... we don't use TF's built in support anymore, do we... and that wouldn't work for other frameworks anyway. I'm guessing other use-cases don't work either, like ONNX models that are split across separate files.

src/backends/custom/custom_backend_factory.cc
```diff
@@ -66,12 +66,20 @@ CustomBackendFactory::CreateBackend(
       GetDirectoryFiles(path, true /* skip_hidden_files */, &custom_files));
 
   std::unordered_map<std::string, std::string> custom_paths;
+  std::vector<std::shared_ptr<TemporaryDirectory>> local_custom(
+      custom_files.size());
+  for (size_t s = 0; s < custom_files.size(); s++) {
```
deadeyegoodwin, line 71: You don't need a separate loop to initialize local_custom. Just above like 78 you can local_custom.push_back() and then pass &local_custom.back() into LocalIzeFileFolder() to initialize (see my comments below for LocalizeFileFolder.

src/backends/custom/custom_backend_factory.cc
```diff
@@ -66,12 +66,20 @@ CustomBackendFactory::CreateBackend(
...
+  int i = 0;
   for (const auto& filename : custom_files) {
     const auto custom_path = JoinPath({path, filename});
-
+    RETURN_IF_ERROR(LocalizeFileFolder(custom_path, local_custom[i]));
     custom_paths.emplace(
         std::piecewise_construct, std::make_tuple(filename),
-        std::make_tuple(custom_path));
+        std::make_tuple(local_custom[i]->model_path));
+    i++;
```
deadeyegoodwin, line 82: Below why are we using custom_paths instead of local_custom (local_custom_paths would be a better name)? We localized the model, why are we not using the localized version?

src/core/filesystem.cc
CoderHam: I will file a ticket for this.

src/core/filesystem.h
CoderHam: The subdirectories are created and managed in the LocalizeFileFolder function. It may not be just a directory but a set of subdirectories as well.

deadeyegoodwin: That's fine.. LocalizedDirectory just represents the top-level directory

CoderHam: The logic for the same is already there in  `S3FileSystem::LocalizeFileFolder()`

deadeyegoodwin: The LocalizedDirectory object only represents the top-level directory and is used solely to provide that top-level directory path and to know/control when it needs to be deleted. Filling the contents if the directory should stay with the different FileSystem classes.

src/backends/custom/custom_backend_factory.cc
CoderHam: Fixed naming. custom_paths are updated with the local_custom_paths corresponding to them. Therefore it contains the localized paths and not the true cloud repository path.

src/core/filesystem.cc
```diff
@@ -71,9 +71,9 @@ class FileSystem {
       const std::string& path, std::set<std::string>* files) = 0;
   virtual Status ReadTextFile(
       const std::string& path, std::string* contents) = 0;
-  virtual Status DownloadFileFolder(
-      const std::string& path, std::string* local_path) = 0;
-  virtual Status DestroyFileFolder(const std::string& path) = 0;
+  virtual Status LocalizeFileFolder(
```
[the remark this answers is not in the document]
CoderHam: Fixed

PR #1473, opened 2020-05-12
Allow InferenceRequest batch-size to have value 0

deadeyegoodwin:


src/core/infer_request.cc
```diff
@@ -471,9 +471,9 @@ InferenceRequest::Normalize()
 
   // Determine the batch size and shape of each input.
   if (model_config.max_batch_size() == 0) {
-    // Model does not support Triton-style batching so treat as
-    // batch-size 1 and leave the tensor shapes as they are.
-    batch_size_ = 1;
+    // Model does not support Triton-style batching so set as
+    // batch-size 0 and leave the tensor shapes as they are.
+    batch_size_ = 0;
```
CoderHam, line 476: If this logic is changed here don't you need to fix the `batch_size`  as `std::max(1U, request_->BatchSize());` for all backends including libtorch, plan, onnx, caffe2 etc?

deadeyegoodwin: Yes, but this is on the "response" branch so those backends need to be revisit in whole and this is one change that needs to be made as part of that.

GuanLuo: Why the change is needed? Internally Triton can ask the backend on whether it supports batching

deadeyegoodwin: Generally there is no need to throw away the information that we learned during normalize that the request doesn't use batching (which is what we were doing by setting batch-size to 1 in cases where there was no batching). Right now it probably doesn't make a difference but it the future that may be useful information to have in the request. The specific reason I made the change now is because it is easier to do now then after a lot of code has been ported over assuming the old way, and because for classification I specifically needed to know for an individual request whether it represented a batch or now.

PR #1480, opened 2020-05-13
Fix classification for v2

deadeyegoodwin:


src/servers/grpc_server_v2.cc
```diff
@@ -2419,46 +2439,69 @@ InferResponseCompleteCommon(
...
+      for (uint32_t bs = 0; bs < std::max((uint32_t)1, batch_size); ++bs) {
+        std::vector<std::string> class_strs;
+        RETURN_IF_ERR(TopkClassifications(
+            iresponse, output_idx,
+            reinterpret_cast<const char*>(base) + class_offset,
+            ((class_offset + batch1_byte_size) > byte_size) ? 0
+                                                            : batch1_byte_size,
+            datatype, classification_count, &class_strs));
+
+        // Serialize for binary representation...
```
GuanLuo, line 2480: Always return in binary form?

src/core/infer_response.cc
```diff
@@ -94,6 +97,22 @@ InferenceResponse::AddOutput(
   return Status::Success;
 }
 
+Status
+InferenceResponse::ClassificationLabel(
+    const InferenceResponse::Output& output, const uint32_t class_index,
+    const char** label) const
+{
+  const auto& label_provider = backend_->GetLabelProvider();
```
GuanLuo, line 105: So in the case of ensemble, it will need to store the label provider from other backends internally so that here it can retrieve the one corresponding to the output?

src/servers/grpc_server_v2.cc
deadeyegoodwin: For GRPC Triton always returns in raw_contents format.

src/core/infer_response.cc
deadeyegoodwin: When initializing the LabelProvider for an ensemble backend, you should look at the ensemble outputs and find the corresponding labels attached to that output (from other models) and then load those into the ensemble's label provider.

PR #1482, opened 2020-05-14
Update ONNX backend to request / response flow

GuanLuo:
Manually tested with `simple_grpc_v2_string_infer_client.py` and `simple_grpc_v2_infer_client.py`

### DIFF
--- a/src/backends/onnx/onnx_backend.cc
+++ b/src/backends/onnx/onnx_backend.cc
@@ -540,50 +540,6 @@ OnnxBackend::Context::ValidateOutputs(
   return Status::Success;
 }
 
-namespace {
-
-#ifdef TRTIS_ENABLE_STATS
-#define FAIL_ALL_AND_RETURN_IF_ERROR(REQUESTS, RESPONSES, S)        \
-  do {                                                              \
-    const auto& status__ = (S);                                     \
-    if (!status__.IsOk()) {                                         \
-      for (auto& response : (RESPONSES)) {                          \
-        if (response != nullptr) {                                  \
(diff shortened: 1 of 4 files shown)


src/backends/onnx/onnx_backend.cc
```diff
@@ -545,49 +540,125 @@ OnnxBackend::Context::ValidateOutputs(
...
+          true /* release_requests */);
+      return;
     }
 
-    total_batch_size += payload.request_->BatchSize();
-
-    // All payloads must have equally-sized input tensors so use any
-    // payload as the representative for the input tensors.
-    repr_input_request = payload.request_.get();
+    total_batch_size += request->BatchSize();
```
deadeyegoodwin, line 615: std::max(1U, ... with batch size

Make sure to fix in other places as well.

src/backends/onnx/onnx_backend.cc
```diff
@@ -545,49 +540,125 @@ OnnxBackend::Context::ValidateOutputs(
   return Status::Success;
 }
 
-Status
+namespace {
+
+#ifdef TRTIS_ENABLE_STATS
+#define FAIL_ALL_AND_RETURN_IF_ERROR(REQUESTS, RESPONSES, S)        \
```
deadeyegoodwin, line 546: I think these macros would be good to but into backend.h... and I think TF backend is doing the wrong thing WRT updating the stats when it does this equivalent thing. Can you update TF backend to use this macro as well.

GuanLuo:
Manually tested with L0_infer

PR #1488, opened 2020-05-14
Fix grpc_v2_image_client to be like grpc_image_client

tanmayv25:
Adds the async, streaming and image directory support to the grpc_v2_image_client.
Tested locally. Will add the test once #1464 gets merged.

### DIFF
--- a/src/clients/python/experimental_api_v2/examples/grpc_v2_image_client.py
+++ b/src/clients/python/experimental_api_v2/examples/grpc_v2_image_client.py
@@ -29,6 +29,7 @@ import argparse
 import numpy as np
 from PIL import Image
 import os
+import sys
 
 import grpc
 from tritongrpcclient import grpc_service_v2_pb2
@@ -196,7 +197,7 @@ def postprocess(results, filenames, batch_size):
 
 
 def requestGenerator(input_name, output_name, c, h, w, format, dtype, FLAGS,
-                     input_filenames):
+                     result_filenames):
     request = grpc_service_v2_pb2.ModelInferRequest()
     request.model_name = FLAGS.model_name
     request.model_version = FLAGS.model_version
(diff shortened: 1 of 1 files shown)


src/clients/python/experimental_api_v2/examples/grpc_v2_image_client.py
```diff
@@ -215,7 +216,8 @@ def requestGenerator(input_name, output_name, c, h, w, format, dtype, FLAGS,
 
     filenames.sort()
 
-    output = grpc_service_v2_pb2.ModelInferRequest().InferRequestedOutputTensor()
+    output = grpc_service_v2_pb2.ModelInferRequest().InferRequestedOutputTensor(
```
CoderHam, line 219: Fix formatting

tanmayv25: Done

src/clients/python/experimental_api_v2/examples/grpc_v2_image_client.py
```diff
@@ -237,23 +239,32 @@ def requestGenerator(input_name, output_name, c, h, w, format, dtype, FLAGS,
...
-    input.contents.CopyFrom(input_contents)
-    request.inputs.extend([input])
-
-    yield request
+    # Send requests of FLAGS.batch_size images. If the number of
+    # images isn't an exact multiple of FLAGS.batch_size then just
+    # start over with the first images until the batch is filled.
+    image_idx = 0
+    last_request = False
+    while not last_request:
```
CoderHam, line 246: If the number of images is not equal to batch size shouldn't the images be repeated as described in the comments?
The check should remain `for idx in range(FLAGS.batch_size):`
inside the loop having `image_idx = image_idx % len(filenames)` retains this behavior. 

Additionally why do you need a `result_filenames`? The name here refers to the name of the input file.

tanmayv25: The point of having `for idx in range(FLAGS.batch_size):` is to ensure the inner loop runs for `batch_size` times. The `idx` must not be confused with `image_idx`. The `image_idx` will loop around the available images to create the inputs.
For instance, if the batch_size is 2 and the num of images in the directory are 3(say A, B and C), then the logic selects A and B for the first batch and C and A for the second batch. Hence, starting over the first image as described in the comment.

result_filenames is the list of list of input filenames. From the above example, it will be [[A,B], [C,A]]. The client needs to record the filename to print along with the inferred label.

PR #1493, opened 2020-05-15
Add API to unset shared memory in client library

tanmayv25:


GuanLuo:
Are you merging into the correct target branch? Or you need to rebase your working branch

CoderHam:
@GuanLuo is right. I believe this needs to merged into master and then v2test needs to be rebased onto master.

src/clients/python/experimental_api_v2/library/httpclient.py
```diff
@@ -1469,6 +1469,17 @@ def set_shared_memory(self, region_name, byte_size, offset=0):
         if offset != 0:
             self._parameters['shared_memory_offset'] = offset
 
+    def unset_shared_memory(self):
+        """Clears the shared memory option set by the last call to
+        InferRequestedOutput.set_shared_memory(). After call to this
+        function requested output will no longer be returned in a
+        shared memory region.
+        """
+
```
GuanLuo, line 1478: Should also revert the binary_data setting if necessary, see `set_shared_memory()` https://github.com/NVIDIA/triton-inference-server/blob/7e51d1a319d001708f3286bae75b9fc8c18ccec9/src/clients/python/experimental_api_v2/library/httpclient.py#L1464-L1465

tanmayv25: Done.

PR #1496, opened 2020-05-15
Rename files and reference to remove v2 from naming

deadeyegoodwin:


qa/L0_trt_shape_tensors/trt_shape_tensor_test.py
```diff
@@ -139,7 +139,7 @@ def check_status(self, model_name, batch_exec, request_cnt, infer_cnt):
                         "expected model-request-count {}, got {}".format(
                                 request_cnt, actual_request_cnt))
 
-        # FIXME Uncomment below after updated V2 statistics schema from
+        # FIXME Uncomment below after updated statistics schema from
```
tanmayv25, line 142: Can these lines be uncommented now the tests are running on 'response'? Here and other tests with similar messages.

deadeyegoodwin: There are some more statistics changes coming that add instance_count and execution_count. I need to update the extension and then as part of updating the implementation we should be able to fix the several places like this.

PR #1508, opened 2020-05-19
Enhance model index API to show state and reason

deadeyegoodwin:


src/servers/http_server.cc
```diff
@@ -1231,26 +1231,66 @@ void
 HTTPAPIServer::HandleRepositoryIndex(
     evhtp_request_t* req, const std::string& repository_name)
 {
-  if (req->method != htp_method_GET) {
+  if (req->method != htp_method_POST) {
```
CoderHam, line 1234: Don't the clients need to modified as well?

deadeyegoodwin: Yes, updated

PR #1511, opened 2020-05-19
Benchmark L0_perf_nomodel

CoderHam:


qa/L0_perf_nomodel/run_test.sh
```diff
@@ -25,15 +25,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-REPO_VERSION=${NVIDIA_TRITON_SERVER_VERSION}
```
deadeyegoodwin, line 28: Why remove this? It is in all our CI tests.

CoderHam: I believe this check already happens in our test.sh file and so I believe it was no longer in run_test.sh file. Should we still have it in the run_test.sh as it is always run from the test file itself?

deadeyegoodwin: You are right, keep it like it is.

CoderHam: @deadeyegoodwin should I make this change across the other perf run_test files?

deadeyegoodwin: I don't think they use the REPO_VERSION directly for this others... but you can check.

qa/L0_perf_nomodel/test.sh
```diff
@@ -79,26 +81,34 @@ TEST_NAMES=(
     "${UNDERTEST_NAME} 16MB I/O Latency HTTP"
     "${UNDERTEST_NAME} Maximum Throughput GRPC"
     "${UNDERTEST_NAME} Maximum Throughput HTTP")
```
deadeyegoodwin, line 83: Here an other places below you left the closing paren here instead of moving it to the end

CoderHam: Fixed

PR #1523, opened 2020-05-20
Extend the lifetime of InferenceRequest objects for TRT backend

tanmayv25:
TRT backend supports the pipelined execution of multiple requests. This PR will fix the issue arising when Backend::Run() returns but the ProcessResponse is not done finalizing the request.

### DIFF
--- a/src/backends/tensorrt/plan_backend.cc
+++ b/src/backends/tensorrt/plan_backend.cc
@@ -139,8 +139,7 @@ PlanBackend::Context::~Context()
   DestroyEventSet();
 
   // Notify the completion thread to exit
-  completion_queue_.Put(
-      std::make_tuple(nullptr, 0, 0, 0, 0, 0, nullptr, nullptr));
+  completion_queue_.Put(nullptr);
   if (completion_thread_.joinable()) {
     completion_thread_.join();
   }
@@ -1385,13 +1384,16 @@ PlanBackend::Run(
       static_cast<Context*>(contexts_[next_context_[runner_idx]].get());
 
   bool run_failed = true;
-  for (const auto& request : *context->requests_) {
+  for (const auto& request : *context->payload_->requests_) {
     if (request != nullptr) {
       run_failed = false;
       break;
     }
   }
 
+  // On error, handle the response here instead of delegating to the completion
+  // thread as the completion thread will wait on CUDA events unconditionally,
+  // which can be ignored on error.
   if (run_failed) {
     // On inference error, place the context back to the queue immediately
     // as all works for the context should be ignored.
@@ -1401,11 +1403,7 @@ PlanBackend::Run(
     auto event_set_idx = context->next_set_;
     context->next_set_ = (event_set_idx + 1) % context->EVENT_SET_COUNT;
     // Put the details needed by the ProcessResponse thread on the queue
-    context->completion_queue_.Put(std::make_tuple(
-        this, event_set_idx, context->compute_start_ns_,
-        context->compute_input_end_ns_, context->compute_output_start_ns_,
-        context->total_batch_size_, std::move(context->requests_),
-        std::move(context->responses_)));
+    context->completion_queue_.Put(context->payload_);
   }
 
   // Set the next context to be executed on this runner, will block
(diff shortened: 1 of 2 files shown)


src/backends/tensorrt/plan_backend.h
```diff
@@ -268,7 +269,10 @@ class PlanBackend : public InferenceBackend {
     std::vector<void*> buffer_bindings_;
 
     size_t total_batch_size_;
-    std::vector<std::unique_ptr<InferenceRequest>>* requests_;
+    uint64_t compute_start_ns_;
+    uint64_t compute_input_end_ns_;
+    uint64_t compute_output_start_ns_;
+    std::shared_ptr<std::vector<std::unique_ptr<InferenceRequest>>> requests_;
```
deadeyegoodwin, line 275: Use in-place or unique_ptr for requests and responses

src/backends/tensorrt/plan_backend.h
```diff
@@ -232,8 +232,9 @@ class PlanBackend : public InferenceBackend {
     // Assume that the lifetime of the payload is extended until the completion
     // callback is called
     SyncQueue<std::tuple<
```
deadeyegoodwin, line 234: This giant magic tuple is impossible to understand. Can you make it an actual struct.

src/backends/tensorrt/plan_backend.h
```diff
@@ -232,8 +232,9 @@ class PlanBackend : public InferenceBackend {
...
-        InferenceBackend*, size_t, const uint64_t, const size_t,
-        std::vector<std::unique_ptr<InferenceRequest>>*,
+        InferenceBackend*, size_t, const uint64_t, const uint64_t,
+        const uint64_t, const size_t,
+        std::shared_ptr<std::vector<std::unique_ptr<InferenceRequest>>>,
```
deadeyegoodwin, line 237: use in-place or unique_ptr, not shared_ptr.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -1535,7 +1536,15 @@ PlanBackend::Context::Run(
...
   NVTX_RANGE(nvtx_, "Run " + name_);
 
-  requests_ = &requests;
+  INFER_STATS_SET_TIMESTAMP(compute_start_ns_);
+
+  // Need to move the InferenceRequest objects as the lifetime must
+  // be extended till ProcessResponse completes. TensorRT backend
+  // supports the pipelined execution of multiple InferenceRequest
+  // objects.
+  requests_.reset(new std::vector<std::unique_ptr<InferenceRequest>>());
```
deadeyegoodwin, line 1545: Just std::move the entire vector

src/backends/tensorrt/plan_backend.h
tanmayv25: Done

src/backends/tensorrt/plan_backend.cc
tanmayv25: Done

src/backends/tensorrt/plan_backend.h
tanmayv25: Done

src/backends/tensorrt/plan_backend.h
tanmayv25: I am using a shared pointer to the new Payload struct as it will be placed on the SyncQueue structure.

src/backends/tensorrt/plan_backend.h
```diff
@@ -229,13 +229,46 @@ class PlanBackend : public InferenceBackend {
...
+      // The total batch size for the request
+      size_t total_batch_size_;
+
+      // The timestamps for reporting stats
+      uint64_t compute_start_ns_;
+      uint64_t compute_input_end_ns_;
+      uint64_t compute_output_start_ns_;
+
+      // All the composing InferenceRequest objects
+      std::unique_ptr<std::vector<std::unique_ptr<InferenceRequest>>> requests_;
```
deadeyegoodwin, line 263: I don't think you even need unique_ptr here. You can use std::move to efficiently move to/from the in-place vectors.

src/backends/tensorrt/plan_backend.h
```diff
@@ -267,9 +300,8 @@ class PlanBackend : public InferenceBackend {
     // The array size is equal to Context::total_bindings_
     std::vector<void*> buffer_bindings_;
 
-    size_t total_batch_size_;
-    std::vector<std::unique_ptr<InferenceRequest>>* requests_;
-    std::shared_ptr<std::vector<std::unique_ptr<InferenceResponse>>> responses_;
+    // The request details of the ongoing model execution
+    std::shared_ptr<Payload> payload_;
```
deadeyegoodwin, line 304: There is only one owner of a payload at any given time, so you should use a unique_ptr here and in the completion_queue_ above. You can then std::move the payload to transfer ownership.

tanmayv25:
Sure. I will include these changes in fix to  shape tensors.

PR #1532, opened 2020-05-22
Expose the newly added statistics to perf_client output

tanmayv25:

### DIFF
--- a/src/clients/c++/perf_client/perf_client.cc
+++ b/src/clients/c++/perf_client/perf_client.cc
@@ -1304,16 +1304,16 @@ main(int argc, char** argv)
 
       for (PerfStatus& status : summary) {
         uint64_t avg_queue_ns = status.server_stats.queue_time_ns /
-                                status.server_stats.execution_count;
+                                status.server_stats.success_count;
         uint64_t avg_compute_input_ns =
             status.server_stats.compute_input_time_ns /
-            status.server_stats.execution_count;
+            status.server_stats.success_count;
         uint64_t avg_compute_infer_ns =
             status.server_stats.compute_infer_time_ns /
-            status.server_stats.execution_count;
+            status.server_stats.success_count;
         uint64_t avg_compute_output_ns =
             status.server_stats.compute_output_time_ns /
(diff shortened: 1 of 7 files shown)


src/clients/c++/perf_client/perf_client.cc
```diff
@@ -1303,9 +1304,20 @@ main(int argc, char** argv)
 
       for (PerfStatus& status : summary) {
         uint64_t avg_queue_ns = status.server_stats.queue_time_ns /
-                                status.server_stats.request_count;
-        uint64_t avg_compute_ns = status.server_stats.compute_time_ns /
-                                  status.server_stats.request_count;
+                                status.server_stats.execution_count;
```
deadeyegoodwin, line 1307: Have to be careful with counts. The success count tells you how many successful requests there were. inference_count tells you how many inferences were performed where a batch size n request is counted as n inferences. execution_count tells you how many times the model ran... which with dynamic batching can be significantly less than the number of requests or inference_count. I'm not sure you are using execution_count correctly here and maybe elsewhere.

src/clients/c++/perf_client/triton_client_wrapper.cc
```diff
@@ -286,12 +286,15 @@ TritonClientWrapper::ParseStatistics(
                       std::make_pair(this_stat.name(), this_stat.version()),
                       ModelStatistics())
                   .first;
-    it->second.request_count_ = this_stat.inference_stats().success().count();
```
deadeyegoodwin, line 289: I think you still want request_count

src/clients/c++/perf_client/perf_client.cc
tanmayv25: Is it safe to assume the success metric count to be same as other metric counts(compute_*, queue*) as we were doing earlier? Or should we just record the average using there own individual metric counts?

tanmayv25: I was wrong assuming the execution count to be equal to the number of times these metrics were added. Fixed.

Issue #1534, opened 2020-05-22
regex incorrect defined for s3 path.


24hours:
**Description**
If s3 path model contain `-` then triton will report fail to obtain metadata information. 

**Triton Information**
20.05

Are you using the Triton container or did you build it yourself?
build myself

**To Reproduce**
`trtserver --model-repository=s3://minio:9000/bucket_name/model-repo`
this path will fail 

`trtserver --model-repository=s3://minio:9000/bucket_name/model_repo`
this path will work fine

the following line is the culprit 
https://github.com/NVIDIA/triton-inference-server/blob/master/src/core/filesystem.cc#L638

should be defined as `s3://([0-9a-zA-Z-.]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z._-]+)*)?)` instead of `s3://([0-9a-zA-Z-.]+):([0-9]+)/([0-9a-z.-]+)(((/[0-9a-zA-Z.-_]+)*)?)` note the sequence of `-` in last regex group. 

**Expected behavior**
triton should work as long as path is valid.

CoderHam:
Thank you for bringing this to our notice. We will fix this for the next release

CoderHam:
@24hours the bucket name itself cannot contain `_` as mentioned [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html). I can test the regex you gave me but as I understand the order of the `-` should not matter.

24hours:
Hi, 

You are right, bucket name cannot contain `_` it my oversight on bug reporting. 

`.-_` imply range between `.` to `_` , `._-` will match character between `.`, `-`, `_`

you may refer this answer on this https://stackoverflow.com/questions/3028642/regular-expression-for-letters-numbers-and?rq=1

PR #1538, opened 2020-05-22
Fix S3 regex to include '-'

CoderHam:
Solved #1534

### DIFF
--- a/qa/L0_s3_local/test.sh
+++ b/qa/L0_s3_local/test.sh
@@ -42,13 +42,13 @@ PERF_CLIENT=../clients/perf_client
 
 DATADIR="/data/inferenceserver/${REPO_VERSION}/qa_model_repository"
 
-TMPDIR="qa_model_repository"
-cp -r $DATADIR $TMPDIR
+MODELDIR="qa_model_repository"
+cp -r $DATADIR $MODELDIR
 
 # Create model with name that has all types of allowed characters
 DUMMY_MODEL="Model_repo-1.0"
-cp -r $TMPDIR/libtorch_float32_float32_float32 $TMPDIR/$DUMMY_MODEL
-sed -i 's/libtorch_float32_float32_float32/Model_repo-1.0/g' $TMPDIR/$DUMMY_MODEL/config.pbtxt
+cp -r $MODELDIR/libtorch_float32_float32_float32 $MODELDIR/$DUMMY_MODEL
+sed -i 's/libtorch_float32_float32_float32/Model_repo-1.0/g' $MODELDIR/$DUMMY_MODEL/config.pbtxt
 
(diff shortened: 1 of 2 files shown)


src/core/filesystem.cc
```diff
@@ -634,8 +634,8 @@ S3FileSystem::ParsePath(
 S3FileSystem::S3FileSystem(
     const Aws::SDKOptions& options, const std::string& s3_path)
     : options_(options), s3_regex_(
-                             "s3://([0-9a-zA-Z-.]+):([0-9]+)/([0-9a-z.-]+)(((/"
-                             "[0-9a-zA-Z.-_]+)*)?)")
+                             "s3://([0-9a-zA-Z.-]+):([0-9]+)/([0-9a-z.-]+)(((/"
```
deadeyegoodwin, line 637: Why does just moving it in the group change anything. If it is a special character shouldn't wee escape it somehow?

CoderHam: As described by this [comment](https://github.com/NVIDIA/triton-inference-server/issues/1534#issuecomment-632499547).

`.-_` imply range between `.` to `_` , `._-` will match character between `.`, `-`, `_`.

This is because `-` also acts as a range char in regex. Yes I believe since it is a special char escaping it would be the appropriate way to go about it. Fixing it now.

CoderHam:
> Can you update the CI test to make sure we use a name that includes all the allowed characters. 

Expanded test coverage for `L0_s3_local`. The test however will still fail since the model control API appears to be broken.

qa/L0_s3_local/test.sh
```diff
@@ -71,13 +79,13 @@ export AWS_ACCESS_KEY_ID=minio && \
 
 # create and add data to bucket
 python -m pip install awscli-local && \
-    awslocal --endpoint-url=http://localhost:4572 s3 mb s3://demo-bucket && \
-    awslocal s3 sync $DATADIR s3://demo-bucket
+    awslocal --endpoint-url=http://localhost:4572 mb s3://demo-bucket1.0 && \
```
deadeyegoodwin, line 82: Need an underscore in the bucket name and an upper case.

qa/L0_s3_local/test.sh
```diff
@@ -42,6 +42,14 @@ PERF_CLIENT=../clients/perf_client
 
 DATADIR="/data/inferenceserver/${REPO_VERSION}/qa_model_repository"
 
+TMPDIR="qa_model_repository"
```
deadeyegoodwin, line 45: MODELDIR would be a more meaningful name then TMPDIR.

qa/L0_s3_local/test.sh
CoderHam: The bucket name can contain only lower-case characters, numbers, periods, and dashes

PR #1540, opened 2020-05-22
Fix preserve ordering

GuanLuo:


deadeyegoodwin:
@GuanLuo  I see why this is more complex than I originally thought... I'm also thinking that there is something more fundamental that needs to be added to request/response to more easily support this case. Let me think about this some more and I might take over this PR next week.

PR #1542, opened 2020-05-23
Add L0_perf_kaldi Benchmark

CoderHam:


qa/L0_perf_kaldi/test.sh
```diff
@@ -0,0 +1,90 @@
...
+git clone ssh://git@gitlab-master.nvidia.com:12051/dl/JoC/asr_kaldi.git && \
+    cd asr_kaldi && \
+    ./scripts/docker/build.sh && \
+    ./scripts/docker/launch_download.sh && \
+    ./scripts/docker/launch_server.sh
+
+CONCURRENCY=2000
+
+# Client only supports GRPC (5 iterations on the dataset)
+./scripts/docker/launch_client.sh -i 5 -c ${CONCURRENCY} >> client_1.log 2>&1
```
tanmayv25, line 40: Is the client in the docker image something similar to our perf_client?

CoderHam: Similar yes but they claim that the cannot use perf_client instead. There is a related discussion in the ticket.

tanmayv25: Thanks for the pointer!

PR #1547, opened 2020-05-26
Make JSON class to encapsulate rapidjson

deadeyegoodwin:


src/servers/http_server.cc
```diff
@@ -394,287 +433,318 @@ ReadDataFromJson(
...
-        uint8_t* bool_base = reinterpret_cast<uint8_t*>(base);
-        WriteDataToJsonHelper(
-            &data_val, allocator, shape, 1, bool_base, &counter, dtype);
-        break;
+    case TRITONSERVER_TYPE_UINT8: {
+      RETURN_IF_ERR(WriteDataToJsonCheck(
+          output_name, byte_size, sizeof(uint8_t) * element_count));
+      const uint8_t* cbase = reinterpret_cast<const uint8_t*>(base);
+      for (size_t e = 0; e < element_count; ++e) {
+        data_json->AppendUInt(cbase[e]);
```
CoderHam, line 488: Here and for all Appending of output data, don't we want to preserve the shape?

deadeyegoodwin:
> Changes look good. Should I file a ticket for C++ Http client to use and expose TritonJson objects instead of direct RapidJson objects?

@tanmayv25 Do you mean so that we don't expose rapidjson in our C++ client library? I think that would be a good change since it will allow us (or users) to substitute other json implementation at build time if they have something else they prefer. Please file that for 20.06.

src/servers/http_server.cc
deadeyegoodwin: So that the returned json array is *not* flattened like it is in this code? I'm ok returning flattened for now. We would need to add an option to allow users to select "non-flattened" if we wanted to support that.

CoderHam: So for the 20.06 release do we intend on adding support for non-flat output tensors back (along with the option for users to ask for shaped outputs)?

deadeyegoodwin: Not for 20.06. We needed non-flat input because that was part of the spec. But for outputs the spec doesn't really say or have a way to control how they are returned. If we get requests then we will add it.

PR #1549, opened 2020-05-26
Report 'triton' as the server name by default

deadeyegoodwin:


src/core/tritonserver.cc
```diff
@@ -289,7 +289,7 @@ class TritonServerOptions {
 };
 
 TritonServerOptions::TritonServerOptions()
-    : server_id_("inference:0"),
+    : server_id_("triton"),
```
CoderHam, line 292: Could you also propagate these changes to L0_lifecycle, L0_server_status, server.cc and simple_grpc/http_health_metadata.cc/py?

deadeyegoodwin: Yes, I will fix those/

PR #1550, opened 2020-05-26
Fixing FAIL_ALL_AND_RETURN_IF_ERROR macro

tanmayv25:
### Before
The server segfaults when the status is not ok

### After
../clients/perf_client -m plan_float32_float32_float32-4-32 --shape INPUT0:16 --shape INPUT1:16 -b 10*** Measurement Settings ***  Batch size: 10  Measurement window: 5000 msec  Using synchronous calls for inference

Request concurrency: 1
Failed to maintain requested inference load. Worker thread(s) failed to generate concurrent requests.
Thread [0] had error: inference request batch-size must be <= 8 for 'plan_float32_float32_float32-4-32'

### DIFF
--- a/src/core/backend.h
+++ b/src/core/backend.h
@@ -43,7 +43,7 @@ namespace nvidia { namespace inferenceserver {
         if (response != nullptr) {                                        \
           LOG_STATUS_ERROR(                                               \
               InferenceResponse::SendWithStatus(                          \
-                  std::move(response), status__),                         \
+                  std::move(response), (S)),                              \
               (LOG_MSG));                                                 \
         }                                                                 \
       }                                                                   \
--- a/src/core/backend.h
+++ b/src/core/backend.h
@@ -35,43 +35,41 @@
 namespace nvidia { namespace inferenceserver {
 
 #ifdef TRITON_ENABLE_STATS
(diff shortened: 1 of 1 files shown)


src/core/backend.h
```diff
@@ -43,7 +43,7 @@ namespace nvidia { namespace inferenceserver {
         if (response != nullptr) {                                        \
           LOG_STATUS_ERROR(                                               \
               InferenceResponse::SendWithStatus(                          \
-                  std::move(response), status__),                         \
+                  std::move(response), (S)),                              \
```
GuanLuo, line 46: Don't want to evaluate `S` twice. Will declaring `status__` to `const auto` solve the issue?

tanmayv25: Right. We pass the function call in S as well. Let me see.

tanmayv25: Done

GuanLuo:
So the actual issue is the conflicting naming when expanding `LOG_STATUS_ERROR()`?

deadeyegoodwin:
> So the actual issue is the conflicting naming when expanding `LOG_STATUS_ERROR()`?

Name conflict with status__? Tricky... probably should use naming based on the macro name or something.

tanmayv25:
This should fix L0_trt_dynamic_shapes and some other tests. Yes, we must be very careful with using macros.

PR #1551, opened 2020-05-26
Update README and versions for 20.05 release

dzier:
Note that there is another commit here that updates the URL links in the documentation to point to the master-v1 branch


deadeyegoodwin:
@dzier  Take a look at the commit I just added. If it is ok with you then you can go ahead and generate the doc.

PR #1552, opened 2020-05-26
Fixing L0_http and L0_grpc

tanmayv25:
@CoderHam  This should fix the model control. It broke because we missed modifying the regex in http in PR #1509.

### DIFF
--- a/src/clients/python/examples/grpc_image_client.py
+++ b/src/clients/python/examples/grpc_image_client.py
@@ -33,8 +33,7 @@ import sys
 import struct
 
 import grpc
-from tritongrpcclient import grpc_service_pb2
-from tritongrpcclient import grpc_service_pb2_grpc
+from tritongrpcclient import grpc_service_pb2, grpc_service_pb2_grpc
 import tritongrpcclient.model_config_pb2 as mc
 
 FLAGS = None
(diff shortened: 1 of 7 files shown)


src/clients/python/examples/grpc_image_client.py
```diff
@@ -64,6 +65,17 @@ def model_dtype_to_np(model_dtype):
         return np.dtype(object)
     return None
 
+def deserialize_bytes_tensor(encoded_tensor):
```
CoderHam, line 68: This is already available in the utils package

tanmayv25: This is to be served as an example if users want to bypass our client package. Hence, I included the deserialization logic here itself. I would avoid using the client packages here.

src/clients/python/examples/grpc_image_client.py
```diff
@@ -30,6 +30,7 @@
 from PIL import Image
 import os
 import sys
+import struct
 
 import grpc
 from tritongrpcclient import grpc_service_pb2
```
CoderHam, line 36: the imports can be comma separated such as `from a import x, y`

tanmayv25: Done

PR #1562, opened 2020-05-27
Remove deprecated cmdline options and change defaults

deadeyegoodwin:
Remove
--grpc-infer-thread-count
--grpc-stream-infer-thread-count
--allow-poll-model-repository
--allow-model-control

--model-control-mode default is "none"


CoderHam:
Since we no longer support multiple ports should we also remove OPTION_HTTP_HEALTH_PORT and the related code?

deadeyegoodwin:
> Since we no longer support multiple ports should we also remove OPTION_HTTP_HEALTH_PORT and the related code?

Yes, that is a good point, I will add it to this PR.

CoderHam:
I think it would be more appropriate to rename to the `L0_multiple_ports` test to `L0_ports` or something like that

deadeyegoodwin:
> I think it would be more appropriate to rename to the `L0_multiple_ports` test to `L0_ports` or something like that

Well, there are still multiple ports being tested, they are just not all http ports...

PR #1569, opened 2020-05-28
Fix windows client build/test

CoderHam:


src/clients/python/library/utils_build_wheel.sh
```diff
@@ -45,27 +45,11 @@ function main() {
...
-    fi
-  fi
-
   cp utils.py \
-    "${WHLDIR}/tritonclientutils/."
+    "${WHLDIR}/tritonclientutils/__init__.py"
 
   cp utils_setup.py "${WHLDIR}"
-  touch ${WHLDIR}/tritonclientutils/__init__.py
+  # touch ${WHLDIR}/tritonclientutils/__init__.py
```
deadeyegoodwin, line 53: Is touch needed or not?

src/clients/python/library/CMakeLists.txt
```diff
@@ -147,41 +147,81 @@ endif() # TRITON_ENABLE_GRPC
...
+
+if(${TRITON_ENABLE_GPU})
+  add_custom_command(
+      OUTPUT "${utils_wheel_stamp_file}"
+      COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/utils_build_wheel.sh"
+      ARGS "${CMAKE_CURRENT_BINARY_DIR}/utils/"
+      DEPENDS
+        ${CMAKE_CURRENT_BINARY_DIR}/VERSION
+        ${CMAKE_CURRENT_BINARY_DIR}/utils_setup.py
+        cshm
```
deadeyegoodwin, line 161: Utils shouldn't depend on csgm and ccudashm? Don't need the GPU if-else.

src/clients/python/library/CMakeLists.txt
```diff
@@ -147,41 +147,81 @@ endif() # TRITON_ENABLE_GRPC
...
+#
+# triton client shared memory utils wheel file
+#
 if(NOT WIN32)
-  set(utils_wheel_stamp_file "utils_stamp.whl")
-  configure_file(utils_setup.py utils_setup.py COPYONLY)
+  set(shm_utils_wheel_stamp_file "shm_utils_stamp.whl")
+  configure_file(shm_utils_setup.py shm_utils_setup.py COPYONLY)
 
   if(${TRITON_ENABLE_GPU})
```
deadeyegoodwin, line 194: Since most of the it-def is identical it would be better to define a SHM_UTILS_DEPENDS that is either "cshm" or "cshm ccudashm"

src/clients/python/library/utils_build_wheel.sh
CoderHam: It is not needed. Removed

CoderHam:
> What are the full names of the produced whl files? Are we handling arch and noarch correctly?

Here are the full names. Yes we handle arch and noarch correctly. 'any' symbolizes noarch while linux symbolizes all linux distros. 
```
tritonhttpclient-1.14.0.dev0-py3-none-any.whl
tritongrpcclient-1.14.0.dev0-py3-none-any.whl
tritonclientutils-1.14.0.dev0-py3-none-any.whl
tritonshmutils-1.14.0.dev0-py3-none-linux.whl
```

PR #1573, opened 2020-05-28
Fix L0_shared_memory

tanmayv25:


qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -256,7 +256,7 @@ def test_too_big_shm(self):
             error_msg, "input2_data", 128)
         if len(error_msg) > 0:
             self.assertTrue(
-                "unexpected size 128 for inference input 'INPUT1', expecting 64" in error_msg[-1])
+                "unexpected total batch size 128 for input 'INPUT1', expecting 64" in error_msg[-1])
```
CoderHam, line 259: Good catch. Needs to be fixed in `cuda_shared_memory_test.py` as I understand.

tanmayv25: L0_infer_cudashm and other cuda shared memory tests are still failing. I will send out the PR to fix them.

PR #1576, opened 2020-05-29
Fix L0_build_variants

CoderHam:
- Fix No GPU server build
- Fix build without stats collection

@GuanLuo can you have a look at the onnx change I made? I appears the string cuda copy check was always false.

### DIFF
--- a/src/backends/onnx/onnx_backend.cc
+++ b/src/backends/onnx/onnx_backend.cc
@@ -809,6 +809,7 @@ OnnxBackend::Context::SetInputTensors(
       TRITONSERVER_MemoryType mem_type;
       auto input_buffer = input_buffers->back()->MutableBuffer(&mem_type);
       size_t buffer_offset = 0;
+      bool string_cuda_copy = false;
       for (size_t ridx = 0; ridx < requests.size(); ++ridx) {
         const InferenceRequest::Input* in;
         auto status = requests[ridx]->ImmutableInput(name, &in);
@@ -836,7 +837,7 @@ OnnxBackend::Context::SetInputTensors(
                     src_byte_size, src_buffer,
                     input_buffer + buffer_offset + input_offset, stream_,
                     &cuda_used);
-                *cuda_copy |= cuda_used;
+                string_cuda_copy |= cuda_used;
               }
             }
             if (status.IsOk()) {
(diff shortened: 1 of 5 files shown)


tanmayv25:
@CoderHam  Did you check L0_build_variants with these changes?

CoderHam:
> @CoderHam Did you check L0_build_variants with these changes?

Yes, all tests except _Build without stats collection_ pass. I have a fix for that on the gitlab side by turning TRITON_ENABLE_METRICS=OFF as it fails only because of this check:
https://github.com/NVIDIA/triton-inference-server/blob/master/build/CMakeLists.txt#L77-L79

src/backends/onnx/onnx_backend.cc
```diff
@@ -854,13 +853,6 @@ OnnxBackend::Context::SetInputTensors(
         buffer_offset += expected_byte_sizes[ridx];
       }
 
-#ifdef TRITON_ENABLE_GPU
-      // Synchronize to ensure the buffer is ready to be modified
-      if (string_cuda_copy) {
-        cudaStreamSynchronize(stream_);
-      }
-#endif  // TRITON_ENABLE_GPU
```
GuanLuo, line 862: Why dropping this?

src/core/cuda_utils.h
```diff
@@ -47,7 +47,8 @@ namespace nvidia { namespace inferenceserver {
 
 #ifndef TRITON_ENABLE_GPU
 using cudaStream_t = void*;
-#endif  // TRITON_ENABLE_GPU
+using cudaEvent_t = void*;
```
GuanLuo, line 50: I don't think `cudaEvent_t` is mentioned in cuda_utils?

src/backends/onnx/onnx_backend.cc
CoderHam: If you notice the string_cuda_copy is never used after initializing. It is always false. Thus this is never run

GuanLuo: Keep `string_cuda_copy` and change line 840 to `string_cuda_copy |= cuda_used;`. FYI, @tanmayv25 this should fix the L0_infer-cudashm

src/core/cuda_utils.h
CoderHam: You're right. Removed.

src/backends/onnx/onnx_backend.cc
CoderHam: Sure. Made this fix.

tanmayv25: @GuanLuo , I did try that yesterday evening but the backend  still read weird values from the memory. Did you check this change fixed L0_infer_cudashm?

GuanLuo: I didn't check, just noticed that in this PR. It's weird that the issue is not due to synchronization as it only fails on cudashm variation. I will go over this part more closely once I get to it.

PR #1581, opened 2020-05-29
Fix warmup for TensorRT backend. Fix warmup batch handling for sequence model

GuanLuo:


src/core/infer_request.h
```diff
@@ -379,8 +380,15 @@ class InferenceRequest {
   // Create a copy of 'from' suitable for use as a "null" request as
   // required for the direct sequence batcher. The returned copy will
   // contain only the minimum content required for a null request.
+  // The statistics of the copy will not be collected.
   static InferenceRequest* CopyAsNull(const InferenceRequest& from);
 
+  // Create a copy of 'from' suitable for use. The caller must be aware that
```
deadeyegoodwin, line 386: I don't like the semantics of this copy. Can you just regenerate the request again when it is needed multiple times for warmup. Also there is no need to suppress stats collection for warmup.

src/core/infer_request.h
```diff
@@ -379,8 +380,15 @@ class InferenceRequest {
...
+  // the callbacks are copied and input data are shared. 'from' must have longer
+  // lifecycle than returned copy.
+  // The statistics of the copy will not be collected.
+  static InferenceRequest* Copy(const InferenceRequest& from);
```
deadeyegoodwin, line 390: Remove this decl

GuanLuo: done

PR #1587, opened 2020-05-29
Fix L0_tf_gpu_io

tanmayv25:


src/backends/tensorflow/base_backend.cc
```diff
@@ -787,6 +787,9 @@ BaseBackend::Context::Run(
                                                 : TRITONSERVER_MEMORY_CPU,
             (TRTISTF_TensorIsGPUTensor(tensor)) ? gpu_device_ : 0);
       }
+
+      LOG_VERBOSE(1) << "input '" << input_name << "' is GPU tensor: "
```
CoderHam, line 791: If we now have independent error messages for input and outputs that are on the GPU maybe expand L0_tf_gpu_io to include  the number of "is GPU tensor: 1" that are expected instead of passing if even 1 is found?

tanmayv25: I do not understand. Isn't the number of "is GPU tensor: 1" depends on how many requests the perf_client got to run( which is not a fixed number)? We had independent input/output messages earlier too. Do you what me to verify the ratio of input and output messages in L0_tf_gpu_io?

CoderHam: I see what you mean. No this is good for now.

PR #1592, opened 2020-06-02
Update C++ client libraries to use TritonJson

deadeyegoodwin:
The C++ API is changes to return JSON as a string, not as a rapidjson object model.  The examples themselves continue to use radpijson to parse that json when necessary (but any json library could be used).


deadeyegoodwin:
@tanmayv25  This remove rapidjson from C++ API. I'll hold off on merging this until the rest of the CI is fixed.

PR #1594, opened 2020-06-02
Fix image clients for non-batching models (like example onnx models)

CoderHam:
Fixes image model tests for onnx.

### DIFF
--- a/src/clients/c++/examples/image_client.cc
+++ b/src/clients/c++/examples/image_client.cc
@@ -215,13 +215,13 @@ Postprocess(
   // Validate shape. Special handling for non-batch model
   if (!batching) {
     if ((shape.size() != 1) || (shape[0] != (int)topk)) {
-      std::cerr << "received incorrect shapes for " << output_name << std::endl;
+      std::cerr << "received incorrect shape for " << output_name << std::endl;
       exit(1);
     }
   } else {
     if ((shape.size() != 2) || (shape[0] != (int)batch_size) ||
         (shape[1] != (int)topk)) {
-      std::cerr << "received incorrect shapes for " << output_name << std::endl;
+      std::cerr << "received incorrect shape for " << output_name << std::endl;
       exit(1);
     }
   }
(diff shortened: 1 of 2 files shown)


CoderHam:
@deadeyegoodwin I saw your PR for the client library rework to use TritonJson. I have not incorporated those changes here yet.

src/clients/c++/examples/image_client.cc
```diff
@@ -930,11 +940,7 @@ main(int argc, char** argv)
   std::vector<int64_t> shape;
   // Include the batch dimension if required
   if (model_info.max_batch_size_ != 0) {
-    if (batch_size > 0) {
```
deadeyegoodwin, line 933: There is a test above for batch_size > 0 so don't need this.

src/clients/c++/examples/image_client.cc
```diff
@@ -211,12 +211,21 @@ Postprocess(
...
-  // Validate shape
-  if ((shape.size() != 2) || (shape[0] != (int)batch_size) ||
-      (shape[1] != (int)topk)) {
-    std::cerr << "received incorrect shapes for " << output_name << std::endl;
-    exit(1);
+
+  // Validate shape. Special handling for non-batch model
+  if (!batching) {
+    if ((shape.size() != 1) || (shape[0] != (int)topk)) {
+      std::cerr << "received incorrect shapes for " << output_name << std::endl;
```
deadeyegoodwin, line 218: "incorrect shape"

src/clients/c++/examples/image_client.cc
```diff
@@ -211,12 +211,21 @@ Postprocess(
...
+      exit(1);
+    }
+  } else {
+    if ((shape.size() != 2) || (shape[0] != (int)batch_size) ||
+        (shape[1] != (int)topk)) {
+      std::cerr << "received incorrect shapes for " << output_name << std::endl;
```
deadeyegoodwin, line 224: "incorrect shape"

src/clients/c++/examples/image_client.cc
CoderHam: Yes I removed this in this PR itself.

deadeyegoodwin: Ha, yes. I looked at it backwards.

PR #1596, opened 2020-06-02
Fix L0_client_tar

deadeyegoodwin:


qa/L0_client_tar/test.sh
```diff
@@ -32,75 +32,42 @@ rm -fr triton_client
...
 fi
 
-grep -c "failed to connect" $CLIENT_LOG
-if [ $? -ne 0 ]; then
-    cat $CLIENT_LOG
-    echo -e "\n***\n*** Test Failed\n***"
-    exit 1
-fi
+# Check static libraries
+for l in libgrpcclient.so libgrpcclient_static.a libhttpclient.so libhttpclient_static.a; do
```
tanmayv25, line 48: At some point, I think we must add the tests for these libraries than just checking there presence,

PR #1597, opened 2020-06-03
Fix data synchronization and data validity

GuanLuo:


src/backends/tensorflow/base_backend.cc
```diff
@@ -866,11 +865,11 @@ BaseBackend::Context::Run(
   // into each. For tensors with string data type we must handle
   // ourselves since we must use TF-specific string tensor APIs.
   cuda_copy = false;
-
+  // The serialized string buffer must be valid until output copies are done
+  std::vector<std::unique_ptr<std::string>> string_buffer;
```
tanmayv25, line 869: So, we hold onto the serialized string and extend the lifetime before the event is synchronized?
Not that it matters much, but I am curious as to why only the first 6 values of the OUTPUT1 were different and that too  [-134, -1, 1, 0, 1, 0].

GuanLuo: I still don't know what the cause is, the issue disappears when I ensure the responder is in proper scope, even though in the failing case responder is not holding any buffer. But at least the code change here is doing the right thing, string_buffer, collector and responder must be valid until the corresponding copies are completed

PR #1601, opened 2020-06-03
Davidg gen

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo  This change makes it possible to generated the ONNX models using specific onnx version and opset. But even with this change I am seeing "failed to load 'onnx_object_int32_object' version 1: Internal: onnx runtime error 2: Failed to load model with error: Unknown model file format version." which seems to come from a model being loaded that requires an opset higher than what is in the server. But I generated models with opset 11 so should be ok. Perhaps ORT 1.3.0 will fix but for now 20.06 Triton can't load generated onnx models.

PR #1602, opened 2020-06-03
Fix model policy

GuanLuo:


src/core/infer_request.cc
```diff
@@ -425,6 +425,9 @@ InferenceRequest::PrepareForInference()
 Status
 InferenceRequest::Normalize()
 {
+  if (priority_ == 0) {
```
deadeyegoodwin, line 428: This is taken care of in SetPriority

GuanLuo: SetPriority will not be called if the request doesn't specify a priority

deadeyegoodwin: Then I think it would be better to call SetPriority(0) from the InferRequest constructor so that the logic stays in 1 place.

PR #1604, opened 2020-06-03
Add V2 Go client and enable test

CoderHam:


Dockerfile.QA
```diff
@@ -44,19 +44,34 @@ FROM ${BUILD_IMAGE} AS build
...
 # image_preprocess custom backend requires opencv
+# go client requires golang
 RUN apt-get update && apt-get install -y --no-install-recommends \
             libopencv-dev \
-            libopencv-core-dev
+            libopencv-core-dev \
+            golang-go \
+            protobuf-compiler
+
+apt-get update && apt-get install -y --no-install-recommends \
```
deadeyegoodwin, line 54: This can't be in a dockerfile.... doesn't it give you an error? And it is redundant with above.

Dockerfile.QA
```diff
@@ -44,19 +44,34 @@ FROM ${BUILD_IMAGE} AS build
...
+            golang-go \
+            protobuf-compiler
 
 WORKDIR /workspace
 RUN cd builddir && \
     make -j16 test-util custom-backend
 
+# need protoc-gen-go to generate go specific gRPC modules
+RUN go get github.com/golang/protobuf/protoc-gen-go && \
```
deadeyegoodwin, line 63: Put this at line 53 after the apt-get run. This is part of setting up packages so it should stay together with similar commands.

Dockerfile.QA
```diff
@@ -44,19 +44,34 @@ FROM ${BUILD_IMAGE} AS build
...
+    go get google.golang.org/grpc
+
+ENV PATH "$PATH:$(go env GOPATH)/bin"
```
deadeyegoodwin, line 66: By putting the env and the above install command here they are not going to be included in the final CI/devel container? Notice that this is a multipart docker file. I think you want to put this all around line 264 and the ENV should be set at the end with the other ENV

Dockerfile.QA
CoderHam: That was doubly added by mistake

Dockerfile.QA
CoderHam: I believe the apt-get should also be moved to the CI container. Moved that to the CI container section as well

src/clients/go/grpc_simple_client.go
```diff
@@ -0,0 +1,258 @@
...
+	outputs := Postprocess(inferResponse)
+	outputData0 := outputs[0]
+	outputData1 := outputs[1]
+
+	fmt.Println("\nChecking Inference Outputs\n--------------------------")
+	for i := 0; i < outputSize; i++ {
+		fmt.Printf("%d + %d = %d\n", inputData0[i], inputData1[i], outputData0[i])
+		fmt.Printf("%d - %d = %d\n", inputData0[i], inputData1[i], outputData1[i])
+	}
+}
```
tanmayv25, line 258: Add newline

src/clients/go/grpc_simple_client.go
```diff
@@ -0,0 +1,258 @@
...
+	inferResponse := ModelInferRequest(client, rawInput, FLAGS.ModelName, FLAGS.ModelVersion)
+
+	/* We expect there to be 2 results (each with batch-size 1). Walk
+	over all 16 result elements and print the sum and difference
+	calculated by the model. */
+	outputs := Postprocess(inferResponse)
+	outputData0 := outputs[0]
+	outputData1 := outputs[1]
+
+	fmt.Println("\nChecking Inference Outputs\n--------------------------")
```
tanmayv25, line 253: Instead of just checking for this statement on our CI, can you validate the output data as well?

CoderHam: The output data is being validated as a part of the test script

tanmayv25: I see. The PostProcess is comparing the values.

qa/L0_simple_go_client/test.sh
```diff
@@ -0,0 +1,77 @@
...
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+RET=0
+
+git clone --single-branch --depth=1 -b ${UPSTREAM_BRANCH_NAME} \
```
deadeyegoodwin, line 48: This doesn't follow the pattern of how we do the CI tests. If we need something from the repo we copy it into the test directory as part of Dockerfile.QA. You should do that here instead of using git.

CoderHam: Done.

PR #1605, opened 2020-06-04
Fix L0_sequence_batcher non-ensemble tests

deadeyegoodwin:


qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -716,12 +723,15 @@ def test_ragged_batch(self):
...
+                threads[0].start()
+                threads[1].start()
+                threads[2].start()
+                time.sleep(3)
+                threads[3].start()
                 for t in threads:
                     t.join()
                 self.check_deferred_exception()
-                self.check_status(model_name, {2: 3, 1: 6}, 12, 12)
+                self.check_status(model_name, {4:9}, 9, 12)
```
GuanLuo, line 734: The description of the test (line 642) says the different shape requests should not be batched together, there shouldn't be batch 4 executions?

deadeyegoodwin: Notice that 2 of the sequences have the same shape. So 3 runs for those 2 + 3 runs for each of the 2 others. Each of these 9 runs is batch-size 4 even though some of those slots have null requests because of the different shapes. The batch stats count the full batch size even if some of the slots are null... by design.

GuanLuo: I see, so the other two different shape sequences will still occupy slots even when they can not yet be executed.

PR #1606, opened 2020-06-04
Fix reshape for output and ensemble

GuanLuo:


src/core/infer_response.cc
```diff
@@ -73,6 +73,12 @@ InferenceResponse::AddOutput(
 
   LOG_VERBOSE(1) << "add response output: " << outputs_.back();
 
+  if (backend_ != nullptr) {
```
deadeyegoodwin, line 76: backend_raw_ is guaranteed to be non-null so you should use that.

src/core/infer_response.h
```diff
@@ -137,6 +137,9 @@ class InferenceResponse {
     // The shape of the output tensor.
     const std::vector<int64_t>& Shape() const { return shape_; }
 
+    // Reshape the output tensor if specified in output config.
+    void Reshape(const ModelOutput* output_config);
```
deadeyegoodwin, line 141: This can be private

src/core/infer_response.cc
GuanLuo: InferenceResponse class only has `backend_`

src/core/infer_response.h
GuanLuo: `Rehape()` is member function of `InferenceResponse::Output` and called in `InferenceResponse`

PR #1607, opened 2020-06-04
Update ONNX Runtime version to 1.3.0

GuanLuo:


Dockerfile
```diff
@@ -373,36 +384,18 @@ RUN apt-get update && \
...
-COPY --chown=1000:1000 --from=tritonserver_onnx /data/dldt/openvino_2019.3.376/deployment_tools/model_optimizer \
-     /opt/openvino_scripts/openvino_2019.3.376/deployment_tools/model_optimizer/
-COPY --chown=1000:1000 --from=tritonserver_onnx /data/dldt/openvino_2019.3.376/tools \
-     /opt/openvino_scripts/openvino_2019.3.376/tools
-ENV INTEL_CVSDK_DIR /opt/openvino_scripts/openvino_2019.3.376
-ENV PYTHONPATH /opt/openvino_scripts:$INTEL_CVSDK_DIR:$INTEL_CVSDK_DIR/deployment_tools/model_optimizer:$INTEL_CVSDK_DIR/tools:$PYTHONPATH
-
-# ONNX Runtime requires Python3 and additional packages to
-# convert ONNX models to OpenVINO models in its OpenVINO execution accelerator.
-RUN apt-get update && \
```
deadeyegoodwin, line 399: Does openvino no longer require python?

GuanLuo: It seems like it

deadeyegoodwin: That's good news.

PR #1608, opened 2020-06-04
Fix ensemble_scheduler to return inference error

tanmayv25:


src/core/ensemble_scheduler.cc
```diff
@@ -769,8 +772,15 @@ EnsembleContext::ScheduleSteps(
 {
   for (auto& step : steps) {
     step->ctx_ = context;
-    context->is_->InferAsync(step->request_);
-    step.release();
+    {
+      std::lock_guard<std::mutex> lock(context->mutex_);
+      context->ensemble_status_ = context->is_->InferAsync(step->request_);
```
GuanLuo, line 777: check if `ensemble_status_` is okay before executing this part to ensure `FinishEnsemble()` will be called only once, or modify `FinishEnsemble()` to allow repeated access (but only the first access will take effect)

tanmayv25: Nice catch!

PR #1610, opened 2020-06-04
Fix TRT backend to handle dynamically batched requests

tanmayv25:
This PR in itself doesn't fix the L0_trt_shape_tensor tests. There is only one failing test left:test_sequence_different_shape_values. 
We can not use the dummy value as content for shape tensors in null requests for sequence scheduler . The shape values must be in range of optimization profile. I am working out on a solution.


deadeyegoodwin:
@tanmayv25  Sounds good but please discuss your solution with me before implementing

PR #1613, opened 2020-06-04
Split L0_perf_nomodel into 3 parts

CoderHam:
- for each type of shared memory


CoderHam:
> Nice change. Cleans things up well.

The CI passes but for some reason the BENCHMARK_URL is not available even after exporting. Fixing in a follow up PR

PR #1617, opened 2020-06-04
Fix shape tensor handling in sequence batcher

tanmayv25:
This should fix all the L0_trt_shape_tensor tests

### DIFF
--- a/src/core/infer_request.h
+++ b/src/core/infer_request.h
@@ -70,6 +70,9 @@ class InferenceRequest {
     // mutable method would allow it to get out-of-sync.
     const std::string& Name() const { return name_; }
 
+    // Whether or not the input is a tensorrt shape tensor
+    bool IsShapeTensor() const { return is_shape_tensor_; }
+
     // Data type of the input tensor.
     DataType DType() const { return datatype_; }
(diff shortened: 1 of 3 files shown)


src/core/infer_request.h
```diff
@@ -92,6 +95,9 @@ class InferenceRequest {
     // data.
     Status SetData(const std::shared_ptr<Memory>& data);
 
+    // Set the input to be treated as a shape tensor.
+    Status SetAsShapeTensor();
```
deadeyegoodwin, line 99: SetShapeTensor(const bool is_shape_tensor), no need to rerun CI

src/core/infer_request.h
```diff
@@ -70,6 +70,9 @@ class InferenceRequest {
     // mutable method would allow it to get out-of-sync.
     const std::string& Name() const { return name_; }
 
+    // Whether or not the input is a tensorrt shape tensor
+    bool IsShapeTensor() const { return is_shape_tensor_; }
```
deadeyegoodwin, line 74: Put this and SetShapeTensor at line 90 just before Data().

PR #1623, opened 2020-06-05
Cleaning shared memory manager

tanmayv25:
Removes the V1 functions from the manager


src/servers/shared_memory_manager.h
```diff
@@ -26,9 +26,9 @@
 #pragma once
 
 #include <cstring>
+#include <map>
+#include <memory>
 #include <mutex>
-#include <unordered_map>
-#include "src/core/server_status.pb.h"
```
deadeyegoodwin, line 31: You should also be able to delete src/core/server_status.proto

PR #1629, opened 2020-06-05
Handle no InferRequestedOutput case in c++ http client library

tanmayv25:


src/clients/c++/library/http_client.cc
```diff
@@ -330,6 +330,9 @@ HttpInferRequest::PrepareRequestJson(
     }
 
     request_json->Add("outputs", std::move(outputs_json));
+  } else {
+    return Error(
+        "request should include atleast one InferRequestedOutput object");
```
deadeyegoodwin, line 335: "at least"

src/clients/c++/library/http_client.h
```diff
@@ -319,9 +319,9 @@ class InferenceServerHttpClient : public InferenceServerClient {
   /// \param result Returns the result of inference.
   /// \param options The options for inference request.
   /// \param inputs The vector of InferInput describing the model inputs.
-  /// \param outputs Optional vector of InferRequestedOutput describing how the
-  /// output must be returned. If not provided then all the outputs in the model
-  /// config will be returned as default settings.
+  /// \param outputs The vector of InferRequestedOutput describing how the
```
GuanLuo, line 322: Why it can't be optional?

src/clients/c++/library/http_client.cc
GuanLuo, line 335: "at least"

tanmayv25: Done

src/clients/c++/library/http_client.h
tanmayv25: If we don't specify any InferRequestedOutput, the client fails to attach the parameters to use binary extension to the output. So, by default the server will return the output as json in response. The http client library currently doesn't support output content returned as json.

PR #1634, opened 2020-06-08
Install C++ client headers

deadeyegoodwin:


tanmayv25:
Changes look good. We must enhance the current L0_client_tar test to actually compile the small client example. This would help in catching other issues. Do you want me to take look?

deadeyegoodwin:
> Changes look good. We must enhance the current L0_client_tar test to actually compile the small client example. This would help in catching other issues. Do you want me to take a look?

No, it is not a priority right now.

PR #1651, opened 2020-06-10
Add basic integration for Triton generic backend

deadeyegoodwin:


src/backends/backend/tritonbackend.h
```diff
@@ -0,0 +1,566 @@
...
+///
+
+/// Create the response factory associated with a request. This
+/// function can be called only once for a given request. Subsequent
+/// calls with result in a TRITONSERVER_ERROR_ALREADY_EXISTS error.
+///
+/// \param factory Returns the new response factory.
+/// \param request The inference request.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_EXPORT TRITONSERVER_Error* TRITONBACKEND_ResponseFactoryNew(
```
GuanLuo, line 261: So the response factory is going to be decoupled from the request?

deadeyegoodwin: Yes, the response factory allows you to send requests even after you have released the request itself. Releasing the request  also releases all the buffers holding the input tensors so you want to release the request as soon as possible. You can hold onto the response factory for much longer so that you can generate as many responses as needed.

PR #1656, opened 2020-06-11
Improving C++ gRPC client performance for successive requests

tanmayv25:
This PR will improve the performance of C++ grpc client library when successive requests carry identical number of inputs(preferably content byte size).

UPDATE: Enabled the original request rates in  CI.

### DIFF
--- a/src/clients/c++/library/grpc_client.cc
+++ b/src/clients/c++/library/grpc_client.cc
@@ -1053,7 +1053,7 @@ InferenceServerGrpcClient::PreRunProcessing(
 
   // Remove extra InferInputTensor submessages, that are not required for
   // this request.
-  if (index < infer_request_.inputs().size()) {
+  while (index < infer_request_.inputs().size()) {
     infer_request_.mutable_inputs()->RemoveLast();
   }
 
@@ -1087,6 +1087,7 @@ InferenceServerGrpcClient::PreRunProcessing(
     }
     index++;
   }
+
   // Remove extra InferRequestedOutputTensor submessages, that are not required
   // for this request.
   while (index < infer_request_.outputs().size()) {
--- a/src/clients/c++/library/grpc_client.cc
+++ b/src/clients/c++/library/grpc_client.cc
@@ -1008,6 +1008,12 @@ InferenceServerGrpcClient::PreRunProcessing(
     auto grpc_input = (infer_request_.inputs().size() <= index)
                           ? infer_request_.add_inputs()
                           : infer_request_.mutable_inputs()->Mutable(index);
+
+    if (input->IsSharedMemory()) {
+      // The input contents must be cleared when using shared memory.
+      grpc_input->Clear();
+    }
+
     grpc_input->set_name(input->Name());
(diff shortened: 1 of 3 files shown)


src/clients/c++/library/grpc_client.cc
```diff
@@ -1046,10 +1048,24 @@ InferenceServerGrpcClient::PreRunProcessing(
...
       }
     }
+    index++;
   }
 
+  // Remove extra InferInputTensor submessages, that are not required for
+  // this request.
+  if (index < infer_request_.inputs().size()) {
+    infer_request_.mutable_inputs()->DeleteSubrange(
+        index - 1, (infer_request_.inputs().size() - index));
```
deadeyegoodwin, line 1058: If, for example, if there is one input, aren't you deleting the subrange that starts at 0? Also, I think just calling RemoveLast the appropriate number of times will be more efficient.

tanmayv25: My bad. 
> RemoveLast the appropriate number of times will be more efficient.
Is it because the ownership is retained by the array?

src/clients/c++/library/grpc_client.cc
```diff
@@ -1046,10 +1048,23 @@ InferenceServerGrpcClient::PreRunProcessing(
         }
       }
     }
+    index++;
   }
 
+  // Remove extra InferInputTensor submessages, that are not required for
+  // this request.
+  if (index < infer_request_.inputs().size()) {
```
deadeyegoodwin, line 1056: This needs to be "while" with index++ in the loop. The reason it is more efficient is that the other way may have shifted array elements... depending on how it was implemented.

src/clients/c++/library/grpc_client.cc
```diff
@@ -1070,6 +1085,12 @@ InferenceServerGrpcClient::PreRunProcessing(
             .set_int64_param(offset);
       }
     }
+    index++;
+  }
+  // Remove extra InferRequestedOutputTensor submessages, that are not required
```
deadeyegoodwin, line 1090: Blank line here, and in the loop you need an index++

src/clients/c++/library/grpc_client.cc
tanmayv25: Ahh.. I am making many mistakes today!
About index++, after the RemoveLast call, infer_request_.inputs().size() will be decremented by 1, so we must be fine.

tanmayv25:
With this fix our request-rate-manager is working fine,

deadeyegoodwin:
> With this fix our request-rate-manager is working fine,

Do you mean that perf is now the same as V1? Or perhaps a little better?

tanmayv25:
> Do you mean that perf is now the same as V1? Or perhaps a little better?

I had to reduce the request-rate-range from the CI earlier because the V2 perf_client was unable to follow the schedule. I suspected the perf_client_v2 back then. But it looks like this change has fixed the issue. 

That said the performance of V1 and V2 for large batch sizes appears to be identical, atleast on my machine. We will see the benchmarks results soon.

PR #1658, opened 2020-06-12
Add support and test coverage for ONNX custom ops

CoderHam:
Grabbing the generated custom ops library from the onnxruntime unit-tests for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -277,6 +277,12 @@ WORKDIR /workspace
 RUN rm -fr *
 COPY . .
 
+# Copy ONNX custom op library and model (Needed for testing)
+COPY --from=tritonserver_onnx /workspace/build/Release/libcustom_op_library.so \
+    /workspace/qa/L0_custom_ops/
+COPY --from=tritonserver_onnx /workspace/build/Release/testdata/custom_op_library/custom_op_test.onnx \
+    /workspace/qa/L0_custom_ops/
+
 # Build the server.
 #
 # - Need to find CUDA stubs if they are available since some backends
--- a/qa/L0_custom_ops/test.sh
+++ b/qa/L0_custom_ops/test.sh
@@ -133,11 +133,11 @@ wait $SERVER_PID
 mkdir -p onnx_custom_ops/custom_op/1 && \
     cp custom_op_test.onnx onnx_custom_ops/custom_op/1/model.onnx
 
-touch config.pbtxt
-echo "name: \"custom_op\"" >> config.pbtxt && \
-echo "platform: \"onnxruntime_onnx\"" >> config.pbtxt && \
-echo "max_batch_size: 0" >> config.pbtxt && \
-echo "custom_library_path: \"./libcustom_op_library.so\"" >> config.pbtxt
+touch onnx_custom_ops/custom_op/config.pbtxt
+echo "name: \"custom_op\"" >> onnx_custom_ops/custom_op/config.pbtxt && \
+echo "platform: \"onnxruntime_onnx\"" >> onnx_custom_ops/custom_op/config.pbtxt && \
+echo "max_batch_size: 0" >> onnx_custom_ops/custom_op/config.pbtxt && \
+echo "op_library_filename: \"./libcustom_op_library.so\"" >> onnx_custom_ops/custom_op/config.pbtxt
 
 SERVER_ARGS="--model-repository=onnx_custom_ops --strict-model-config=false"
 SERVER_LD_PRELOAD=""
--- a/qa/L0_custom_ops/test.sh
+++ b/qa/L0_custom_ops/test.sh
@@ -41,6 +41,7 @@ CLIENT_LOG="./client.log"
 ZERO_OUT_TEST=zero_out_test.py
 CUDA_OP_TEST=cuda_op_test.py
 MOD_OP_TEST=mod_op_test.py
+ONNX_OP_TEST=onnx_op_test.py
 
 SERVER=/opt/tritonserver/bin/tritonserver
 SERVER_LOG="./inference_server.log"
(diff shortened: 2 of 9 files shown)


src/core/model_config.proto
```diff
@@ -1290,4 +1290,13 @@ message ModelConfig
   //@@     model.
   //@@
   repeated ModelWarmup model_warmup = 16;
+
+  //@@  .. cpp:var:: string custom_library_path
```
deadeyegoodwin, line 1294: op_library_filename

src/core/model_config.proto
```diff
@@ -1290,4 +1290,13 @@ message ModelConfig
...
+  //@@
+  //@@     Optional path of the custom library that contains custom
```
deadeyegoodwin, line 1296: "Optional paths of the libraries providing custom operations for this model. Valid only for ONNX models."

src/core/model_config.proto
```diff
@@ -1290,4 +1290,13 @@ message ModelConfig
...
+  //@@     operations that are used in this specific model.
+  //@@     Necssary only for ONNXRuntime models. Please use LD_PRELOAD for other
+  //@@     types of models.
+  //@@
+  string custom_library_path = 18;
```
deadeyegoodwin, line 1301: This needs to be a repeated string in general. Does ONNXRT API allow loading multiple op libraries?

Dockerfile
```diff
@@ -277,6 +277,12 @@ WORKDIR /workspace
 RUN rm -fr *
 COPY . .
 
+# Copy ONNX custom op library and model (Needed for testing)
```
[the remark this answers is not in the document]
CoderHam: The qa directories are not copied till line 278

src/core/model_config.proto
CoderHam: The API docs don't explicitly mention if multiple op libraries can be loaded in the same session. I have asked regarding the same. Do we still want it to be a repeated string if multiple op libraries are not supported?

deadeyegoodwin: Yes, because in the future it may be supported in onnxrt or some other framework. I should have also said that instead of having the repeated directly in the top-level ModelConfig you should have a ModelOperations message. For now that message just has the op library member but we may have other things in the future.

qa/L0_custom_ops/test.sh
```diff
@@ -128,4 +129,41 @@ fi
 kill $SERVER_PID
 wait $SERVER_PID
 
+# ONNX
+mkdir -p onnx_custom_ops/custom_op/1 && \
+    cp custom_op_test.onnx onnx_custom_ops/custom_op/1/model.onnx
+
+touch config.pbtxt
```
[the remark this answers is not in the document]
CoderHam: fixed.

src/core/model_config.proto
CoderHam: Sure, I agree that having a ModelOperations message would be neater as well. I'll make this change

CoderHam: Their reply was: https://github.com/microsoft/onnxruntime/issues/4192#issuecomment-643447705
I have restructure the message and made op_library_filename a repeated field.

src/backends/onnx/autofill.cc
```diff
@@ -311,6 +312,13 @@ AutoFillOnnx::Create(
 
   RETURN_IF_ORT_ERROR(ort_api->CreateSessionOptions(&session_options));
 
+  // Does nothing if op_libraries is empty
+  for (const auto& lib_filename : op_libraries) {
+    void* library_handle = nullptr;  // leak this, no harm.
```
deadeyegoodwin, line 317: Isn't there a way to unload or close a handle?

src/backends/onnx/onnx_backend.cc
```diff
@@ -301,6 +301,16 @@ OnnxBackend::CreateExecutionContext(
     glock.lock();
   }
 
+  // Register all op libraries that contains custom operations.
+  if (Config().has_model_operations()) {
+    auto model_ops = Config().model_operations();
+    for (const auto& lib_filename : model_ops.op_library_filename()) {
+      void* library_handle = nullptr;  // leak this, no harm.
```
deadeyegoodwin, line 308: Is something actually leaking? What are you supposed to do with this handle?

src/core/autofill.cc
```diff
@@ -124,6 +124,9 @@ AutoFill::Create(
   const Platform platform = GetPlatform(config.platform());
 #endif
 
+  // If model operations is specified, use it to set the session options for ONNX.
+  auto model_ops = config.model_operations();
```
deadeyegoodwin, line 128: This is only needed in the ONNX case so define it only where needed below.

src/core/model_config.proto
```diff
@@ -1136,6 +1136,21 @@ message ModelWarmup
 //@@
 message ModelConfig
 {
+  //@@
+  //@@  .. cpp:var:: message ModelOperations
+  //@@
+  //@@     The metadata of libraries providing custom operations for this model.
+  //@@
+  message ModelOperations
```
deadeyegoodwin, line 1144: Please follow the established conventions. We do not use nested messages in ModelConfig. Create message at top-level as is done for other members of ModelConfig

src/core/autofill.cc
CoderHam: Will do

src/backends/onnx/onnx_backend.cc
CoderHam: This is not documented. It is from the onnxruntime example [here](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/test/shared_lib/test_inference.cc). From what I understand the op libraries should be unregistered as a part of the cleanup after the session is closed.

src/backends/onnx/autofill.cc
CoderHam: No such function exists. The use of the handle has not been described. From what I understand it is only exposed to keep the ownership with the user.

deadeyegoodwin: It would be much more helpful if your comment said something about the use of the handle not being defined... saying it is leaking when we don't really know what is happening is confusing. It would also be good to check with ONNXRT to understand what the use of the handle is for and if any cleanup is needed to unload the libraries when the model is unloaded.

deadeyegoodwin:
@CoderHam You also need to update the documentation custom ops section.

CoderHam:
Yes, will check with ONNXRT and will update the docs in a follow up PR.

PR #1671, opened 2020-06-16
Update docs for onnx custom ops

CoderHam:


docs/custom_operation.rst
```diff
@@ -146,3 +146,30 @@ to use the `NGC PyTorch container
...
+field. For example, if you follow `this example
+<https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/test/shared_lib/test_inference.cc>`_
+in the `microsoft/onnxruntime<https://github.com/microsoft/onnxruntime>`_
+repository and your ONNXRuntime custom operations are compiled into libonnxcustom.so,
+adding the following to the model configuraion of your model makes those 
+operations available to that specific ONNX model::
+
+  $ model_operations { op_library_filename: "/path/to/libonnxcustom.so" }
+
+A limitation of this approach is that the custom operations must be
```
deadeyegoodwin, line 166: The first part of this paragraph is not true given that you can have the onnx shared library inside the model repo. Is the second part true? Does ONNXRT have a shared namespace for all custom operations? Even though you are loading them per-model? That would be strange.

CoderHam: You are right. This paragraph is not relevant for ONNXRT

PR #1673, opened 2020-06-17
Add option on CUDA event synchronization

GuanLuo:
Default will be blocking sync


src/core/model_config.proto
```diff
@@ -474,6 +474,15 @@ message ModelOptimizationPolicy
     //@@       backend.
     //@@
     bool graphs = 1;
+
```
deadeyegoodwin, line 477: For this and for 'graphs' above can you put that the default is "false". That is the protobuf default but still useful to document it explicitly.

PR #1675, opened 2020-06-17
Expose the decoupled nature of request and responses to the server

tanmayv25:


src/core/model_config.proto
```diff
@@ -1144,6 +1144,25 @@ message ModelOperations
   repeated string op_library_filename = 1;
 }
 
+//@@
+//@@ .. cpp:var:: message ModelTransactionType
+//@@
+//@@    The specification that describes the nature of transactions
+//@@    to be expected from the model.
+//@@
+message ModelTransactionType
```
deadeyegoodwin, line 1153: ModelTransactionPolicy

src/core/model_config.proto
```diff
@@ -1144,6 +1144,25 @@ message ModelOperations
...
+{
+  //@@  .. cpp:var:: bool decoupled
+  //@@
+  //@@     Indicates whether responses generated by the model are decoupled with
+  //@@     the requests issued to it, which means the number of responses
+  //@@     generated by model may differ from number of requests issued. If
```
deadeyegoodwin, line 1159: ".. of requests issued, and that the responses may be out of order relative to the order of requests."

src/core/model_config.proto
```diff
@@ -1144,6 +1144,25 @@ message ModelOperations
...
+  //@@     False, the model is expected to generated exactly one response per
```
deadeyegoodwin, line 1160: "The default is false, which means the model will generate exactly one response for each request."

src/core/infer_request.cc
```diff
@@ -487,6 +487,8 @@ InferenceRequest::Normalize()
 {
   const ModelConfig& model_config = backend_raw_->Config();
 
+  is_decoupled_ = model_config.model_transaction_type().decoupled();
```
deadeyegoodwin, line 490: Decoupled is a property of the model. Both request and response have a handle to the backend which in turn has the config object from which they can get the decoupled flag. If we want to have an "IsDecoupled()" utility method we should put that in the InferenceBackend class (like we do for DefaultPriorityLevel).

src/core/tritonserver.h
```diff
@@ -834,6 +834,17 @@ TRITONSERVER_EXPORT TRITONSERVER_Error* TRITONSERVER_InferenceResponseId(
...
 
+/// Check whether or not the response is from a model with decoupled
+/// requests and responses.
+///
+/// \param inference_response The response object.
+/// \param is_decoupled Returns the truth value of whether the response
+/// is decoupled or not.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_EXPORT TRITONSERVER_Error*
+TRITONSERVER_InferenceResponseIsDecoupled(
```
deadeyegoodwin, line 845: I think we will likely need some function like this to help us understand some properties of the response, but to future-proof I think we will want to make it as general as possible. Can you hold off on this part of the change and include it with your grpc_server changes... we might have more insight into exactly what we need at that point.

PR #1678, opened 2020-06-17
Reorganize the grpc server to bring related code together

tanmayv25:
Just moves the code specific to Infer RPC implementation towards the actual RPC implementation. 
Renames Handler ->InferHandler as it is used only as a base class for ModelInferHandler and ModelStreamInferHandler. Similar renaming.
These changes make the implementation more comprehensible and easier to maintain.

### DIFF
--- a/src/servers/grpc_server.cc
+++ b/src/servers/grpc_server.cc
@@ -190,345 +190,15 @@ operator<<(std::ostream& out, const Steps& step)
 }
 
 //
-// AllocPayload
+// The server has separate handling mechanisms for inference RPCs
+// and non-inference RPCs.
 //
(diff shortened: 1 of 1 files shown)


tanmayv25:
We are deleting the CommonCallData here:
https://github.com/NVIDIA/triton-inference-server/blob/2d8cde09a9ac092bb55a36daefeb5876727e1fab/src/servers/grpc_server.cc#L346

tanmayv25:
When shutting down the server, it is expected the grpc will flush out all the data object from the completion queue and all these objects will be deleted. It's a fair assumption for grpc and there should not be memory leak.

PR #1694, opened 2020-06-22
Remove explicit batch-size from response

deadeyegoodwin:


src/core/infer_response.cc
```diff
@@ -188,29 +191,34 @@ InferenceResponse::Output::~Output()
 }
 
 void
-InferenceResponse::Output::Reshape(const ModelOutput* output_config)
+InferenceResponse::Output::Reshape(
+    const bool has_batch_dim, const ModelOutput* output_config)
 {
-  if (output_config->has_reshape()) {
```
tanmayv25, line 193: We should still check this condition, right?

src/core/infer_response.h
```diff
@@ -138,12 +138,7 @@ class InferenceResponse {
     const std::vector<int64_t>& Shape() const { return shape_; }
 
     // Reshape the output tensor if specified in output config.
```
tanmayv25, line 140: Should elaborate the comment saying the function should be called only when output_config has reshape field. And keep this check in the function implementation? :
` if (output_config->has_reshape()) {`

deadeyegoodwin: I'll update the comment. This function is only used locally so I'd rather avoid the additional check.

PR #1697, opened 2020-06-22
Fix batch size usage in ensemble scheduler. Fix build

GuanLuo:
PR against `davidg-bs` branch


GuanLuo:
@deadeyegoodwin  done

PR #1701, opened 2020-06-24
Align grpc server with decoupled API

tanmayv25:
~There is some bug with state machine which manifests when issuing multiple requests and large number of responses. I am working on fixing it.~

UPDATE: Fixed the state machine to conform with the contract of the GRPC. 
Important aspects of these contracts are:

1.  There should be at max a single outstanding write on both side of the stream.
2. The multiple writes of the decoupled response must be separated with notification on completion queue.

I have used the Alarm workaround to place the task on the completion queue as described [here](https://github.com/G-Research/grpc_async_examples/blob/1fe69d1d025916f15fa1dc4a48704b57b49e3cd9/async_streaming_server_alarm.cc#L138). The server seems to be pretty stable with these changes. I have tested with multiple simultaneous inflight requests and delay times. The order of the responses are also interleaved now.

```
# python /workspace/src/clients/python/examples/simple_grpc_custom_repeat.py 
I0624 02:35:50.149131 4774 repeat.cc:539] TRITONBACKEND_ModelExecute: model repeat_int32 with 1 requests
I0624 02:35:50.149320 4774 repeat.cc:577] waiting 1000 ms before releasing requests
I0624 02:35:50.149389 4774 repeat.cc:397] waiting 1000 ms, then sending response 1 of 10
I0624 02:35:51.149497 4774 repeat.cc:592] TRITONBACKEND_ModelExecute: model repeat_int32 released 1 requests
I0624 02:35:51.149922 4774 repeat.cc:447] sent response 1 of 10
I0624 02:35:51.149992 4774 repeat.cc:397] waiting 1000 ms, then sending response 2 of 10
I0624 02:35:52.150434 4774 repeat.cc:447] sent response 2 of 10
I0624 02:35:52.150494 4774 repeat.cc:397] waiting 1000 ms, then sending response 3 of 10
I0624 02:35:53.150948 4774 repeat.cc:447] sent response 3 of 10
I0624 02:35:53.151062 4774 repeat.cc:397] waiting 1000 ms, then sending response 4 of 10
I0624 02:35:54.151447 4774 repeat.cc:447] sent response 4 of 10
I0624 02:35:54.151527 4774 repeat.cc:397] waiting 1000 ms, then sending response 5 of 10
I0624 02:35:55.151909 4774 repeat.cc:447] sent response 5 of 10
I0624 02:35:55.151968 4774 repeat.cc:397] waiting 1000 ms, then sending response 6 of 10
I0624 02:35:56.152405 4774 repeat.cc:447] sent response 6 of 10
I0624 02:35:56.152466 4774 repeat.cc:397] waiting 1000 ms, then sending response 7 of 10
I0624 02:35:57.152849 4774 repeat.cc:447] sent response 7 of 10
I0624 02:35:57.152911 4774 repeat.cc:397] waiting 1000 ms, then sending response 8 of 10
I0624 02:35:58.153486 4774 repeat.cc:447] sent response 8 of 10
I0624 02:35:58.153546 4774 repeat.cc:397] waiting 1000 ms, then sending response 9 of 10
I0624 02:35:59.153993 4774 repeat.cc:447] sent response 9 of 10
I0624 02:35:59.154055 4774 repeat.cc:397] waiting 1000 ms, then sending response 10 of 10
I0624 02:36:00.154513 4774 repeat.cc:447] sent response 10 of 10
100 : 100
101 : 101
102 : 102
103 : 103
104 : 104
105 : 105
106 : 106
107 : 107
108 : 108
109 : 109
PASS: Decoupled API
```

### DIFF
--- a/src/core/grpc_service.proto
+++ b/src/core/grpc_service.proto
@@ -742,6 +742,23 @@ message ModelInferResponse
 }
 
 //@@
+//@@.. cpp:var:: message ControlResponse
+//@@
+//@@   The control signal for a stream inference response
+//@@
+message ControlResponse
+{
+  //@@
+  //@@  .. cpp:var:: bool final
+  //@@
+  //@@     Whether or not the response is last to be generated for
+  //@@     a particular request. If false, then further responses
+  //@@     must be expected.
+  //@@
+  bool final = 1;
+}
+
+//@@
 //@@.. cpp:var:: message ModelStreamInferResponse
 //@@
 //@@   Response message for ModelStreamInfer.
@@ -762,6 +779,14 @@ message ModelStreamInferResponse
   //@@     Holds the results of the request.
   //@@
   ModelInferResponse infer_response = 2;
+
+  //@@
+  //@@  .. cpp:var:: ControlResponse control_response
+  //@@
+  //@@     Returns the control flags associated with
+  //@@     the response.
+  //@@
+  ControlResponse control_response = 3;
 }
 
 //@@
--- a/src/core/grpc_service.proto
+++ b/src/core/grpc_service.proto
@@ -742,23 +742,6 @@ message ModelInferResponse
 }
 
 //@@
-//@@.. cpp:var:: message ControlResponse
-//@@
-//@@   The control signal for a stream inference response
-//@@
-message ControlResponse
-{
-  //@@
-  //@@  .. cpp:var:: bool final
-  //@@
-  //@@     Whether or not the response is last to be generated for
-  //@@     a particular request. If false, then further responses
-  //@@     must be expected.
-  //@@
-  bool final = 1;
-}
-
-//@@
 //@@.. cpp:var:: message ModelStreamInferResponse
 //@@
 //@@   Response message for ModelStreamInfer.
@@ -779,14 +762,6 @@ message ModelStreamInferResponse
   //@@     Holds the results of the request.
   //@@
   ModelInferResponse infer_response = 2;
-
-  //@@
-  //@@  .. cpp:var:: ControlResponse control_response
-  //@@
-  //@@     Returns the control flags associated with
-  //@@     the response.
-  //@@
-  ControlResponse control_response = 3;
 }
 
 //@@
--- a/src/core/tritonserver.h
+++ b/src/core/tritonserver.h
@@ -1243,10 +1243,10 @@ typedef enum tritonserver_modelindexflag_enum {
 
 /// Model transaction policy flags. The enum values must be
 /// power-of-2 values.
-typedef enum tritonserver_txn_policy_flag_enum {
-  TRITONSERVER_TXN_POLICY_ONE_TO_ONE = 0,
-  TRITONSERVER_TXN_POLICY_DECOUPLED = 1
-} TRITONSERVER_ModelTxnPolicyFlag;
+typedef enum tritonserver_txn_property_flag_enum {
+  TRITONSERVER_TXN_ONE_TO_ONE = 0,
+  TRITONSERVER_TXN_DECOUPLED = 1
+} TRITONSERVER_ModelTxnPropertyFlag;
 
 /// Create a new server object. The caller takes ownership of the
 /// TRITONSERVER_Server object and must call TRITONSERVER_ServerDelete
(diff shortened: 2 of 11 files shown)


src/core/tritonserver.h
```diff
@@ -1288,6 +1295,27 @@ TRITONSERVER_EXPORT TRITONSERVER_Error* TRITONSERVER_ServerModelBatch(
...
+///
+/// \param server The inference server object.
+/// \param model_name The name of the model.
+/// \param model_version The version of the model.  If -1 then the
+/// server will choose a version based on the model's policy.
+/// \param txn_flags Returns flags indicating the transaction policy of the
+/// model.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_EXPORT TRITONSERVER_Error*
+TRITONSERVER_ServerModelTransactionPolicy(
```
deadeyegoodwin, line 1315: Let's call this ...TransactionProperties(). I'm going to change the batch properties name to TRITONSERVER_ServerModelBatchProperties() as well. Can you also add a voidp argument as in ModelBatch to provide some future proofing.

src/core/tritonserver.h
```diff
@@ -1288,6 +1295,27 @@ TRITONSERVER_EXPORT TRITONSERVER_Error* TRITONSERVER_ServerModelBatch(
     TRITONSERVER_Server* server, const char* model_name,
     const int64_t model_version, uint32_t* flags, void** voidp);
 
+/// Get the transaction policy of the model. The policy is communicated
+/// by a flags value.
+///
+///   - TRITONSERVER_TXN_POLICY_ONE_TO_ONE: The model generates exactly
+///     one request per response.
```
deadeyegoodwin, line 1302: one response per request.

src/core/tritonserver.h
```diff
@@ -1288,6 +1295,27 @@ TRITONSERVER_EXPORT TRITONSERVER_Error* TRITONSERVER_ServerModelBatch(
...
+///
+///   - TRITONSERVER_TXN_POLICY_DECOUPLED: The model may generate multiple
+///     responses per request.
```
deadeyegoodwin, line 1305: may generate zero to many responses per request.

src/core/tritonserver.h
```diff
@@ -1198,6 +1198,13 @@ typedef enum tritonserver_modelindexflag_enum {
   TRITONSERVER_INDEX_FLAG_READY = 1
 } TRITONSERVER_ModelIndexFlag;
 
+/// Model transaction policy flags. The enum values must be
+/// power-of-2 values.
+typedef enum tritonserver_txn_policy_flag_enum {
+  TRITONSERVER_TXN_POLICY_ONE_TO_ONE = 0,
```
deadeyegoodwin, line 1204: TXN_PROPERTY_.. or just TRITONSERVER_TXN_ONE_TO_ONE

src/core/tritonserver.h
```diff
@@ -1198,6 +1198,13 @@ typedef enum tritonserver_modelindexflag_enum {
...
+  TRITONSERVER_TXN_POLICY_DECOUPLED = 1
+} TRITONSERVER_ModelTxnPolicyFlag;
```
deadeyegoodwin, line 1206: TxnProperyFlag

src/servers/grpc_server.cc
```diff
@@ -1515,12 +1515,19 @@ class InferHandlerState {
...
     explicit Context(const uint64_t unique_id = 0)
-        : unique_id_(unique_id), step_(Steps::START), finish_ok_(true)
+        : unique_id_(unique_id), ongoing_requests_(0), step_(Steps::START),
+          finish_ok_(true)
     {
       ctx_.reset(new grpc::ServerContext());
       responder_.reset(new ServerResponderType(ctx_.get()));
     }
 
+    // Increments the ongoing reques counter
```
deadeyegoodwin, line 1525: request

src/servers/grpc_server.cc
```diff
@@ -2806,20 +2858,34 @@ ModelInferHandler::InferResponseComplete(
...
   State* state = reinterpret_cast<State*>(userp);
 
+  // Increment the callback index
+  state->cb_counter_++;
+
   LOG_VERBOSE(1) << "ModelInferHandler::InferResponseComplete, "
                  << state->unique_id_ << " step " << state->step_;
 
+  // Defer to the callback with the final response
+  if ((flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) == 0) {
```
deadeyegoodwin, line 2870: For ModelInfer we expect there to be a single response and it must have the FINAL flag. So if the final flag is not given we should log an error. Or maybe the below code handle it?

src/core/grpc_service.proto
```diff
@@ -741,6 +741,23 @@ message ModelInferResponse
   repeated InferOutputTensor outputs = 5;
 }
 
+//@@
+//@@.. cpp:var:: message ControlResponse
+//@@
+//@@   The control signal for a stream inference response
+//@@
+message ControlResponse
```
deadeyegoodwin, line 749: Does the client library need to know the final response? I don't think so. If we need to add something like this we would need to add it for HTTP as well... I'd rather not have it at all.

src/servers/grpc_server.cc
tanmayv25: The below code will correctly handle the cases where model produced multiple responses for ModelInfer if ModelConfig didn't specify decoupled policy. It just keeps on deferring the return of actual response. Nevertheless, it;'s better to log an error because may be the single response doesn't carry the specify the FINAL flag.

src/core/grpc_service.proto
tanmayv25: Python library doesn't the end signal but I believe we might need something like this for C++ client library to handle statistics. I have removed it from this PR.

PR #1711, opened 2020-06-25
Update ensemble to handle decoupled request/response

GuanLuo:


src/core/ensemble_scheduler.cc
```diff
@@ -41,22 +41,57 @@ namespace {
 
 class EnsembleContext;
 
+using IterationCount = size_t;
```
CoderHam, line 44: Why is this necessary?

GuanLuo: For clarity, too many `size_t` for different kinds of variables.

GuanLuo:
@CoderHam @deadeyegoodwin @tanmayv25 The PR is ready for review

PR #1712, opened 2020-06-25
For repeat backend, clarify handling of zero-size input 

deadeyegoodwin:


deadeyegoodwin:
@tanmayv25 Note that the repeat backend can be used to test the 1 request -> 0 responses case

tanmayv25:
Right! I have tested the operation with repeat_count = 0 in 'simple_grpc_custom_repeat.py'. 
requests
```
I0625 20:44:42.586483 13237 repeat.cc:539] TRITONBACKEND_ModelExecute: model repeat_int32 with 1 requests
I0625 20:44:42.586652 13237 repeat.cc:577] waiting 1000 ms before releasing requests
I0625 20:44:43.586850 13237 repeat.cc:592] TRITONBACKEND_ModelExecute: model repeat_int32 released 1 requests
```
With this PR we can see a message on the server too.

PR #1716, opened 2020-06-26
Add L0_http_fuzz test

CoderHam:
Remaining: Fuzz testing JSON contents to ensure our parser and server is robust to erogenous JSONs.

### DIFF
--- a/qa/L0_http_fuzz/fuzztest.py
+++ b/qa/L0_http_fuzz/fuzztest.py
@@ -32,10 +32,10 @@ import os
 
 
 class FuzzTest(unittest.TestCase):
-    def _run_fuzz(self, url):
+    def _run_fuzz(self, url, logger):
         session = Session(target=Target(
             connection=TCPSocketConnection("127.0.0.1", 8000)),
-            keep_web_open=False)
+            fuzz_loggers=logger, keep_web_open=False)
 
         s_initialize(name="Request"+url)
         with s_block("Request-Line"):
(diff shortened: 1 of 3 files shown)


qa/L0_http_fuzz/fuzztest.py
```diff
@@ -0,0 +1,49 @@
...
+        session = Session(target=Target(
+            connection=TCPSocketConnection("127.0.0.1", 8000)),
+            keep_web_open=False)
+
+        s_initialize(name="Request")
+        with s_block("Request-Line"):
+            s_group("Method", ["GET", "HEAD", "POST", "PUT",
+                               "DELETE", "CONNECT", "OPTIONS", "TRACE"])
+            s_delim(" ", name="space-1")
+            s_group("Request-URI", ["/v2/models/simple", "/v2/models/simple/infer",
```
deadeyegoodwin, line 19: We have a lot more endpoints then what is listed here. I would have thought that we could give a grammer of what are paths that should be accepted and it would try variants of those and also known exploits.

CoderHam: It does try known exploits and variants of the paths. It allows specification of a string but the mutations of the string are not controlled by us. Hence I specify endpoints in this fashion. I will expand to include all the other endpoints as well

qa/L0_http_fuzz/fuzztest.py
```diff
@@ -0,0 +1,82 @@
...
+                                    "/v2/models/simple/versions/v1", "/v2/models/simple/versions/v1",
+                                    "/v2/models/simple/config", "/v2/models/simple/stats",
+                                    "/v2/health/ready", "/v2/health/live",
```
deadeyegoodwin, line 48: I think server metadata (/v2) and model ready are still missing.

CoderHam: Fixed

PR #1717, opened 2020-06-26
Fix the python client library to handle async requests correctly

tanmayv25:


src/clients/python/library/httpclient.py
```diff
@@ -1186,8 +1192,14 @@ def wrapped_post(request_uri, request_body, headers, query_params):
         g = self._pool.apply_async(
             wrapped_post, (request_uri, request_body, headers, query_params))
 
+        # Schedule the greenlet to run in this loop iteration
         g.start()
 
+        # Relinquish control to greenlet loop. Using non-zero
+        # value to ensure the control is transferred to the
+        # event loop.
+        gevent.sleep(0.01)
```
deadeyegoodwin, line 1201: Are you sure that this is needed? What is the impact of waiting 0.01 seconds... if this is needed could we just as well sleep only 0.001 seconds?

PR #1729, opened 2020-06-30
Adding end-to-end tests for decoupled API

tanmayv25:


tanmayv25:
@GuanLuo I think you can add the ensemble decoupled models to this testing framework.

tanmayv25:
I will add further tests for decoupled model.

PR #1744, opened 2020-07-03
Fix error handling in grpc server for decoupled model

tanmayv25:


qa/L0_decoupled/decoupled_test.py
```diff
@@ -277,6 +290,42 @@ def test_no_streaming(self):
...
+        self.inputs_[1].set_shape([repeat_count + 1])
+        self.inputs_[1].set_data_from_numpy(delay_data)
+
+        # Initialize data for WAIT
+        self.inputs_[2].set_data_from_numpy(wait_data)
+
+        user_data = UserData()
+        result_dict = {}
+
+        try:
```
nieksand, line 323: The unittest library has `assertRaises()` for enforcing exceptions:
```
with self.assertRaises(InferenceServerException) as cm:
    self._stream_infer(1, repeat_count, user_data, result_dict)

self.assertIn("expected IN and DELAY shape to match, got [1] and [2]", str(cm.exception))
``` 

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

Issue #1415, opened 2020-05-06
fasterrcnn_resnet50_fpn TorchScript model cannot be loaded


adamm123:
**Description**
Cannot load FasterRCNN model exported to TorchScript using Triton/TensorRTServer.
Follow-up #1313

**Triton Information**
What version of Triton are you using?
`nvcr.io/nvidia/tritonserver:20.03-py3`
`nvcr.io/nvidia/tensorrtserver:20.02-py3`

Are you using the Triton container or did you build it yourself?
`container`
**To Reproduce**

Pytorch container was used for exporting a model to TorchScript, then 
corresponding Triton/TensorRTServer container was used to load exported model.
Tried software versions:
| Pytorch Container | Triton Container | Result |
|--------------------------|-----------------------|------------------------------------------- |
| nvcr.io/nvidia/pytorch:20.03-py3 | nvcr.io/nvidia/tritonserver:20.03-py3 |  Failed to load the model |
| nvcr.io/nvidia/pytorch:20.02-py3 | nvcr.io/nvidia/tensorrtserver:20.02-py3 | Failed to load the model |
| nvcr.io/nvidia/pytorch:20.01-py3 | N/A | Failed to export the model |
| nvcr.io/nvidia/pytorch:19.12-py3 | N/A | Failed to export the model |

1. Export FasterRCNN model (in Pytorch container):
```python
import torch
import torchvision

model = torchvision.models.detection.fasterrcnn_resnet50_fpn()
model = model.eval()
script = torch.jit.script(model)
script.save('model.pt')
```
2. Prepare repository and copy model into it:
```
mkdir -p models/object_detection/1
cp model.pt models/object_detection/1
```
3. Run Triton/TensorRTServer(replace <path_to_model_repository> with correct path):
```
nvidia-docker run --rm --shm-size=1g --ulimit memlock=-1 --ulimit stack=67108864 -p8000:8000 -p8001:8001 -p8002:8002 -v/<path_to_model_repository>/models:/models nvcr.io/nvidia/tritonserver:20.03-py3 trtserver --model-repository=/models --log-verbose=1 --strict-model-config=false
```
Running server results in following error:
```
inference_server_1  | E0506 06:54:57.221317 1 model_repository_manager.cc:840] failed to load 'od' version 1: Internal: load failed for libtorch model -> 'od': 
inference_server_1  | Unknown builtin op: torchvision::_new_empty_tensor_op.
[...]

PR #1750, opened 2020-07-07
Add support for Torchvision ops

CoderHam:
Partial resolution to #1415. Although Torchvision ops will be supported with the changes from this PR, not all models will work with the Triton.
We may need to expand the libtorch backend to support inputs as List of Tensors apart from regular Tensors. Outputs in some cases can be in different form than Tenors. For example fasterrcnn expects Inputs as [List of Tensors](https://github.com/pytorch/vision/blob/master/torchvision/models/detection/faster_rcnn.py#L28-L29) and Outputs as [Dict of Tensors](https://github.com/pytorch/vision/blob/master/torchvision/models/detection/faster_rcnn.py#L39-L40)

Included test in L0_custom_ops

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -161,6 +161,7 @@ ARG TRITON_CONTAINER_VERSION=20.07dev
 
 # libgoogle-glog0v5 is needed by caffe2 libraries.
 # libcurl4-openSSL-dev is needed for GCS
+# python3-dev is needed by Torchvision
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
             autoconf \
@@ -176,6 +177,7 @@ RUN apt-get update && \
             rapidjson-dev \
             libb64-dev \
             patchelf \
+            python3-dev \
             software-properties-common && \
     if [ $(cat /etc/os-release | grep 'VERSION_ID="16.04"' | wc -l) -ne 0 ]; then \
         apt-get install -y --no-install-recommends \
@@ -224,7 +226,7 @@ COPY --from=tritonserver_pytorch /opt/conda/lib/libmkl_intel_lp64.so /opt/triton
 COPY --from=tritonserver_pytorch /opt/conda/lib/libmkl_rt.so /opt/tritonserver/lib/pytorch/
 COPY --from=tritonserver_pytorch /opt/conda/lib/libmkl_vml_def.so /opt/tritonserver/lib/pytorch/
 
-# LibTorch headers and libraries
+# LibTorch and Torchvision headers and libraries
 COPY --from=tritonserver_pytorch /opt/conda/lib/python3.6/site-packages/torch/include \
      /opt/tritonserver/include/torch
 COPY --from=tritonserver_pytorch /opt/conda/lib/python3.6/site-packages/torch/lib/libtorch.so \
@@ -235,10 +237,17 @@ COPY --from=tritonserver_pytorch /opt/conda/lib/python3.6/site-packages/torch/li
       /opt/tritonserver/lib/pytorch/
 COPY --from=tritonserver_pytorch /opt/conda/lib/python3.6/site-packages/torch/lib/libcaffe2_nvrtc.so \
      /opt/tritonserver/lib/pytorch/
+COPY --from=tritonserver_pytorch /opt/pytorch/vision/torchvision/csrc \
+    /opt/tritonserver/include/torchvision/torchvision/
+COPY --from=tritonserver_pytorch /opt/pytorch/vision/build/libtorchvision.so \
+    /opt/tritonserver/lib/pytorch/
 RUN cd /opt/tritonserver/lib/pytorch && \
     for i in `find . -mindepth 1 -maxdepth 1 -type f -name '*\.so*'`; do \
         patchelf --set-rpath '$ORIGIN' $i; \
     done
+
+# Need to include pytorch in LD_LIBRARY_PATH since Torchvision load custom ops
+ENV LD_LIBRARY_PATH /opt/tritonserver/lib/pytorch/:$LD_LIBRARY_PATH
 
 # Onnx Runtime headers and library
 # Put include files to same directory as ONNX Runtime changed the include path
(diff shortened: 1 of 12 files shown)


build/server/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release)
 endif()
 
-set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wno-unused-parameter -Werror")
+# Need to turn unused-but-set-variable off due to Torchvision
+set(CMAKE_CXX_FLAGS "-Wall -Wextra -Wno-unused-parameter -Wno-unused-but-set-variable -Werror")
```
deadeyegoodwin, line 35: Use set_target_properties in cmake file to add this flag only to the pytorch parts of the build that need it.

Dockerfile
```diff
@@ -234,11 +236,18 @@ COPY --from=tritonserver_pytorch /opt/conda/lib/python3.6/site-packages/torch/li
...
+    /opt/tritonserver/include/torchvision/torchvision/
+COPY --from=tritonserver_pytorch /opt/pytorch/vision/build/libtorchvision.so \
+    /opt/tritonserver/lib/pytorch/
 RUN cd /opt/tritonserver/lib/pytorch && \
     for i in `find . -mindepth 1 -maxdepth 1 -type f -name '*\.so*'`; do \
         patchelf --set-rpath '$ORIGIN' $i; \
     done
 
+# Need to include pytorch in LD_LIBRARY_PATH since Torchvision load custom ops
+ENV LD_LIBRARY_PATH /opt/tritonserver/lib/pytorch/:$LD_LIBRARY_PATH
```
deadeyegoodwin, line 249: This is needed at build time?

Dockerfile
```diff
@@ -329,7 +338,7 @@ RUN LIBCUDA_FOUND=$(ldconfig -p | grep -v compat | awk '{print $1}' | grep libcu
                   -DTRITON_ENABLE_PYTORCH=ON \
                   -DTRITON_ENABLE_ENSEMBLE=ON \
                   -DTRITON_ONNXRUNTIME_INCLUDE_PATHS="/opt/tritonserver/include/onnxruntime" \
-                  -DTRITON_PYTORCH_INCLUDE_PATHS="/opt/tritonserver/include/torch" \
+                  -DTRITON_PYTORCH_INCLUDE_PATHS="/opt/tritonserver/include/torch;/opt/tritonserver/include/torch/torch/csrc/api/include;/opt/tritonserver/include/torchvision;/usr/include/python3.6" \
```
deadeyegoodwin, line 341: Why the python? Are the torchscript ops written in python? So they require that python be installed on the server?

Dockerfile
CoderHam: No, this is needed only at runtime, I'll move this to the devel container

Dockerfile
CoderHam: Torchvision has a dependency on Python.h since it links against certain python modules. This is why when adding Torchvision support we require to include python.

Dockerfile.QA
```diff
@@ -308,5 +308,9 @@ RUN rm -fr qa/L0_copyrights qa/L0_build_variants && \
     pip3 install --upgrade qa/pkgs/triton*.whl
 
 ENV LD_LIBRARY_PATH /opt/tritonserver/qa/clients:${LD_LIBRARY_PATH}
+
+# Need to include pytorch in LD_LIBRARY_PATH since Torchvision loads custom ops
+ENV LD_LIBRARY_PATH /opt/tritonserver/lib/pytorch/:$LD_LIBRARY_PATH
```
deadeyegoodwin, line 313: For other models that use custom ops usage we require the user to update the path to point to any required libraries for the custom ops. I'd rather not do something different in this case and even if we did we should put the path in the server container. For the QA container the individual tests set the LD_LIBRARY_PATH appropriately, see L0_custom_ops/test.sh

CoderHam:
> We will need a test for this, don't we have some model that uses torchvision that was given to us to demonstrate this issue?

Yes, I prepared a fasterrccn model that I used to test if the model will be loaded successfully. I will be adding it to the libtorch_model_store for 20.07.

Dockerfile.QA
CoderHam: > For other models that use custom ops usage we require the user to update the path to point to any required libraries for the custom ops. I'd rather not do something different in this case and even if we did we should put the path in the server container. For the QA container the individual tests set the LD_LIBRARY_PATH appropriately, see L0_custom_ops/test.sh

Since we want torchvision ops to always be available we will need the LD_LIBRARY_PATH set permanently. Since without it the server will not run. I can update the docs to remove the section that asks the user to set the LD_LIBRARY_PATH for pytorch custom ops.

Based on your earlier [comment](https://github.com/NVIDIA/triton-inference-server/pull/1750#discussion_r451002073) I thought you wanted it moved from the server (where it was before this). I understand now that you meant it to be in the final base image and not the build image

deadeyegoodwin: Why will the server not run? Do the core pytorch libraries have a dependency on the torchvision library?

CoderHam: The torchvision library has a dependency on the pytorch library since it automatically loads the custom ops and thus the server fails to run with the error that torch.so was not found is the LD_LIBRARY_PATH is not set correctly.

Dockerfile.QA
```diff
@@ -311,5 +311,9 @@ RUN rm -fr qa/L0_copyrights qa/L0_build_variants && \
...
```
deadeyegoodwin, line 316: This needs to be in the final tritonserver container image, doesn't it? Not here.

Dockerfile
```diff
@@ -435,6 +441,9 @@ RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
  && ldconfig \
  && rm -f ${_CUDA_COMPAT_PATH}/lib
 
+# Need to include pytorch in LD_LIBRARY_PATH since Torchvision loads custom ops
+ENV LD_LIBRARY_PATH /opt/tritonserver/lib/pytorch/:$LD_LIBRARY_PATH
```
deadeyegoodwin, line 445: Move this up right after "ENV PATH" since they are related envvars. You don't need to restart CI.

"... since Torchvision loads custom ops from that path"

PR #1758, opened 2020-07-08
Fix intermittent hang in L0_sequence_stress

deadeyegoodwin:


qa/L0_sequence_stress/sequence_stress.py
```diff
@@ -397,6 +397,13 @@ def stress_thread(name, seed, pass_cnt, correlation_id_base, trial, model_name,
...
 
+        # We need to explicitly close each client so that streams get
+        # cleaned up and closed correctly, otherwise the application
+        # can hang when exiting.
+        for c, i in client_metadata_list:
+            print("thread {} closing client {}".format(name, i))
+            c.close()
+
     except Exception as ex:
         _thread_exceptions_mutex.acquire()
```
tanmayv25, line 408: Can you also close the client in the exception handler?

deadeyegoodwin: Done

PR #1760, opened 2020-07-08
Adding the end-to-end timeout on the client side

tanmayv25:
Timeout specification has been added for the inference request. Have intentionally left streaming timeout from the python grpc because of a bug in grpc. For streaming RPC, most likely it treats the timeout in milliseconds instead of seconds(even though it is documented as seconds).  Will explore the issue and add it later.

### DIFF
--- a/src/clients/c++/library/grpc_client.cc
+++ b/src/clients/c++/library/grpc_client.cc
@@ -31,6 +31,7 @@
 #include "src/clients/c++/library/grpc_client.h"
 
 #include <grpcpp/grpcpp.h>
+#include <chrono>
 #include <cstdint>
 #include <future>
 #include <iostream>
@@ -44,9 +45,12 @@ namespace {
 // Use map to keep track of GRPC channels. <key, value> : <url, Channel*>
 // If context is created on url that has established Channel, then reuse it.
 std::map<std::string, std::shared_ptr<grpc::Channel>> grpc_channel_map_;
+std::mutex grpc_channel_map_mtx_;
 std::shared_ptr<grpc::Channel>
 GetChannel(const std::string& url)
 {
+  std::lock_guard<std::mutex> lock(grpc_channel_map_mtx_);
+
   const auto& channel_itr = grpc_channel_map_.find(url);
   if (channel_itr != grpc_channel_map_.end()) {
     return channel_itr->second;
@@ -795,6 +799,13 @@ InferenceServerGrpcClient::Infer(
   for (const auto& it : headers) {
     context.AddMetadata(it.first, it.second);
   }
+
+  if (options.client_timeout_ != 0) {
+    auto deadline = std::chrono::system_clock::now() +
+                    std::chrono::microseconds(options.client_timeout_);
+    context.set_deadline(deadline);
+  }
+
   err = PreRunProcessing(options, inputs, outputs);
   sync_request->Timer().CaptureTimestamp(RequestTimers::Kind::SEND_END);
   if (!err.IsOk()) {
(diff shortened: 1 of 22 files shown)


src/clients/c++/library/grpc_client.cc
```diff
@@ -851,6 +862,13 @@ InferenceServerGrpcClient::AsyncInfer(
   for (const auto& it : headers) {
     async_request->grpc_context_.AddMetadata(it.first, it.second);
   }
+
+  if (options.client_timeout_ != 0) {
```
deadeyegoodwin, line 866: What does the client see when a timeout happens? Do they get an exception? A particular error code returned in the result object?

src/clients/c++/library/http_client.cc
```diff
@@ -1080,8 +1080,12 @@ InferenceServerHttpClient::Infer(
 
   // During this call SEND_END (except in above case), RECV_START, and
   // RECV_END will be set.
-  if (curl_easy_perform(easy_handle_) != CURLE_OK) {
+  auto curl_status = curl_easy_perform(easy_handle_);
+  if (curl_status != CURLE_OK) {
     sync_request->http_code_ = 400;
```
deadeyegoodwin, line 1085: So in a timeout the client just sees a 400 code. Seems like they need to be able to differentiate a timeout from other 400 failures. There isn't a good HTTP code for this so we may just have to use 499 or something.

src/clients/c++/library/http_client.cc
```diff
@@ -1080,8 +1080,12 @@ InferenceServerHttpClient::Infer(
...
+    if (curl_status == CURLE_OPERATION_TIMEDOUT) {
```
deadeyegoodwin, line 1086: I would use cout and conditionalize with the verbose flag since it is not necessarily an error when a timeout occurs.

src/clients/python/library/grpcclient.py
```diff
@@ -893,6 +894,11 @@ def infer(self,
             model-specific action such as terminating the request. If not
             provided, the server will handle the request using default setting
             for the model.
+        client_timeout : int
```
deadeyegoodwin, line 897: How is this timeout indicated to the client? A special exception?

src/clients/python/library/grpcclient.py
```diff
@@ -893,6 +894,11 @@ def infer(self,
...
+            The maximum end-to-end time, in seconds, the request is allowed
```
deadeyegoodwin, line 898: Is this really seconds? That seems very coarse.

tanmayv25:
> What about the HTTP python client?
The geventhttplib doesn't offer separate request level timeouts. 
We already have the logic of specifying the timeouts while creating clients here:
https://github.com/NVIDIA/triton-inference-server/blob/master/src/clients/python/library/httpclient.py#L135-L138

src/clients/c++/library/grpc_client.cc
tanmayv25: They will see an InferenceServerException with message "Deadline Exceeded" message.

src/clients/python/library/grpcclient.py
tanmayv25: They will see an InferenceServerException with message "Deadline Exceeded" message.

src/clients/python/library/grpcclient.py
tanmayv25: > Parameters
> request – The request value for the RPC.
> timeout – An optional duration of time in seconds to allow for the RPC.
> metadata – Optional metadata to be transmitted to the service-side of the RPC.
Link: [here](https://grpc.github.io/grpc/python/_modules/grpc.html#UnaryUnaryMultiCallable.__call__)
Looks like it.. The library also doesn't accept the fractional value for timeout. So, yes the minimum timeout that we can provide is 1s.

src/clients/c++/library/grpc_client.cc
deadeyegoodwin: Add this to the documentation for "client_timeout".

Issue #1724, opened 2020-06-29
Triton container not loading the latest version of model when S3 location is updated


yasirofficial:
**Description**
When using models at AWS S3 by giving the S3 path in --model-repository, the triton container fails to see new version of model uploaded in the bucket 

**Triton Information**
tritonserver:20.03.1-py3

Are you using the Triton container or did you build it yourself?
Triton Container

**To Reproduce**
Start a triton container with model-repository link to AWS S3 and the version policy : latest. The latest version of model will be loaded. After some time, update the AWS S3 bucket with new version of model. It fails to load the new model (or in other words, it skips the update) if the model has served any inference requests in past.

Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).

- Tensorflow saved model (MaskRcnn)

**Expected behavior**
Whenever a new model is added or a new version of model is added in AWS S3, it should be loaded by the triton container

CoderHam:
We could run some tests to confirm is this still exists in our latest master. Is this an issue you experience with all backends or just Tensorflow?

CoderHam:
Can you try to use the latest release 20.06? It appears to work fine for S3 model polling. We are adding a test to prevent regressions in the future. Please remember to use the `--model-control-mode=poll` flag

CoderHam:
Unable to reproduce in master. Please re-open if you see the issue in 20.06

PR #1763, opened 2020-07-08
Include polling test for S3 model loading

CoderHam:
Expand test coverage for L0_storage_s3 to include polling model control mode.


CoderHam:
It appears the issue #1724 is not reproducible in master. I have added a test as well to prevent regressions in the future.

PR #1779, opened 2020-07-09
Document thread-safefy of allocation callbacks

deadeyegoodwin:


src/core/tritonserver.h
```diff
@@ -314,6 +314,17 @@ typedef TRITONSERVER_Error* (*TRITONSERVER_ResponseAllocatorStartFn_t)(
 ///   For each response, TRITONSERVER_InferenceResponseDelete called
 ///    - release_fn: called once for each output tensor in the response
 ///
+/// In all cases the start_fn, alloc_fn and release_fn callback
+/// functions must be thread-safe. Typically making these functions
+/// thread-safe does not require explicit locking. The recommended way
+/// to implement these functions is to have each inference request
+/// initiated by TRITONSERVER_ServerInferAsync provide a 'user_p'
```
GuanLuo, line 321: `TRITONSERVER_ServerInferAsync` doesn't takes `user_p`, I think now it is set in `TRITONSERVER_InferenceRequestSetReleaseCallback`? Also, comments for some other APIs are outdated in the same way.

deadeyegoodwin: Good catch. Updated.

PR #1780, opened 2020-07-09
Include ONNX and Libtorch backends in DeepRecommender perf test

CoderHam:
@deadeyegoodwin I removed the Dynamic Batch case and include coverage for HTTP perf as well. I'll update the dashboard as well shortly. 

Limit to batch size 256 to prevent OOM for libtorch. Remove onnx_trt for now due to segfault in server with batch size 256.

### DIFF
--- a/qa/L0_perf_deeprecommender/test.sh
+++ b/qa/L0_perf_deeprecommender/test.sh
@@ -100,9 +100,9 @@ for FRAMEWORK in graphdef plan graphdef_trt onnx onnx_trt libtorch; do
 done
 
 #
-# Test large static batch = 256 w/ 2 instances
+# Test large static batch = 192 w/ 2 instances
 #
-STATIC_BATCH=256
+STATIC_BATCH=192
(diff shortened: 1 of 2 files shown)


qa/L0_perf_deeprecommender/test.sh
```diff
@@ -48,9 +48,9 @@ STATIC_BATCH=1
 DYNAMIC_BATCH=1
 INSTANCE_CNT=1
 
-# Copy TF Model
-rm -fr tfmodels && mkdir -p tfmodels && \
-    cp -r $REPODIR/perf_model_store/deeprecommender_graphdef tfmodels/.
+# Copy TF, ONNX and PyTorch Models
```
tanmayv25, line 51: Even though the comment says to copy the onnx and pytorch models. I don't see that happening,

CoderHam: Fixed

PR #1789, opened 2020-07-13
Track TF AMP performance in L0_perf_resnset

GuanLuo:
@deadeyegoodwin @dzier merging into 20.07 branch as the change in TF is available since 20.07


CoderHam:
> @CoderHam We'll need some way to add this TF-AMP result to the corresponding jester visualization.

The way the jester board is currently structured allows it to automatically pick up new frameworks. In this case the TF AMP model will be tagged with framework `savedmodel_amp` based on the model name.

PR #1797, opened 2020-07-15
Support secure channel in gRPC endpoint

tanmayv25:


src/servers/main.cc
```diff
@@ -239,6 +245,14 @@ std::vector<Option> options_
        "allocated for reuse. As long as the number of in-flight requests "
        "doesn't exceed this value there will be no allocation/deallocation of "
        "request/response objects."},
+      {OPTION_GRPC_USE_SSL, "grpc-use-ssl", Option::ArgBool,
+       "Use SSL authentication for GRPC requests."},
+      {OPTION_GRPC_SERVER_CERT, "grpc-server-cert", Option::ArgStr,
```
deadeyegoodwin, line 250: Document the defaults if flags not specified.

src/clients/c++/library/grpc_client.h
```diff
@@ -40,6 +40,16 @@ namespace nvidia { namespace inferenceserver { namespace client {
 /// metadata
 typedef std::map<std::string, std::string> Headers;
 
+struct SslOptions {
+  explicit SslOptions() {}
+  // File holding PEM-encoded root certificates
```
deadeyegoodwin, line 45: Document default behavior when not specified.

src/clients/c++/library/grpc_client.cc
```diff
@@ -46,8 +48,22 @@ namespace {
 // If context is created on url that has established Channel, then reuse it.
 std::map<std::string, std::shared_ptr<grpc::Channel>> grpc_channel_map_;
 std::mutex grpc_channel_map_mtx_;
+
+void
+ReadFile(const std::string& filename, std::string& data)
+{
+  std::ifstream file(filename.c_str(), std::ios::in);
```
deadeyegoodwin, line 55: Clear 'data' to make sure return empty if no file.
Better checking if filename is empty?

src/clients/c++/library/grpc_client.cc
```diff
@@ -46,8 +48,22 @@ namespace {
...
+  if (file.is_open()) {
+    std::stringstream ss;
+    ss << file.rdbuf();
+    file.close();
+    data = ss.str();
+  }
+  return;
```
deadeyegoodwin, line 62: Remove

PR #1799, opened 2020-07-15
Increase the max allowed size for grpc messages in python client

tanmayv25:


deadeyegoodwin:
@tanmayv25  Was there a related issue for this? Make sure to reference and close it.

PR #1802, opened 2020-07-16
Fixes some of the memory leaks

tanmayv25:
Fixes most of the 'definitely lost' memory leaks reported by valgrind. This effort covers only custom models and does not fixes leaks in other backends. There is a 32 bytes leak within cnmem library and in dlopen calls. I read that valgrind sometimes report false positive leaks in dlopen. 

See the attached file for detail report after the fixes : [mem_custom.txt](https://github.com/NVIDIA/triton-inference-server/files/4929019/mem_custom.txt)

### DIFF
--- a/src/core/tritonserver.cc
+++ b/src/core/tritonserver.cc
@@ -552,7 +552,7 @@ TRITONSERVER_LogIsEnabled(TRITONSERVER_LogLevel level)
   return false;
 }
 
-TRITONSERVER_Error*
+void
 TRITONSERVER_LogMessage(
     TRITONSERVER_LogLevel level, const char* filename, const int line,
     const char* msg)
@@ -570,12 +570,10 @@ TRITONSERVER_LogMessage(
     case TRITONSERVER_LOG_VERBOSE:
       LOG_VERBOSE_FL(1, filename, line) << msg;
       break;
+    default:
+      LOG_ERROR_FL(__FILE__, __LINE__)
+          << "unknown logging level '" << level << "'";
   }
-
-  return TRITONSERVER_ErrorNew(
-      TRITONSERVER_ERROR_INVALID_ARG,
-      std::string("unknown logging level '" + std::to_string(level) + "'")
-          .c_str());
 }
 
 //
--- a/src/core/tritonserver.cc
+++ b/src/core/tritonserver.cc
@@ -552,7 +552,7 @@ TRITONSERVER_LogIsEnabled(TRITONSERVER_LogLevel level)
   return false;
 }
 
-void
+TRITONSERVER_Error*
 TRITONSERVER_LogMessage(
     TRITONSERVER_LogLevel level, const char* filename, const int line,
     const char* msg)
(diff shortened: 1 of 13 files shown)


src/core/tritonserver.cc
```diff
@@ -570,12 +570,10 @@ TRITONSERVER_LogMessage(
     case TRITONSERVER_LOG_VERBOSE:
       LOG_VERBOSE_FL(1, filename, line) << msg;
       break;
+    default:
+      LOG_ERROR_FL(__FILE__, __LINE__)
+          << "unknown logging level '" << level << "'";
   }
-
-  return TRITONSERVER_ErrorNew(
```
GuanLuo, line 575: Why dropping this?

src/core/tritonserver.cc
```diff
@@ -552,7 +552,7 @@ TRITONSERVER_LogIsEnabled(TRITONSERVER_LogLevel level)
   return false;
 }
 
-TRITONSERVER_Error*
+void
```
deadeyegoodwin, line 555: I can see why you want to do this. But I would instead just like to make it easier to ignore the error. So can you add IGNORE_ERROR to backend_utils.h and use it in LOG_IF_ERROR. Also add a LOG macro that uses LOG_IF_ERROR with success error so that it always logs and in the backends replace TRITONSERVER_LogMessage with the LOG macro.

src/core/tritonserver.cc
```diff
@@ -570,12 +570,10 @@ TRITONSERVER_LogMessage(
     case TRITONSERVER_LOG_VERBOSE:
       LOG_VERBOSE_FL(1, filename, line) << msg;
       break;
+    default:
```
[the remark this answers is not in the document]
tanmayv25: Fixed

src/core/tritonserver.cc
tanmayv25: Dropped earlier because we were ignoring the error any way. The new allocation was not getting removed. Hence, logged the error in default above. I ahve incorporated David's suggestion above which takes care of the leak.

src/servers/main.cc
```diff
@@ -1184,6 +1184,7 @@ main(int argc, char** argv)
   if (keep_alive == nullptr) {
     return 1;
   }
+  delete keep_alive;
```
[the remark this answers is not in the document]
tanmayv25: In some of my local test I didn't observe the segfaults. I have started off the CI. If it fails then I will add this back.

src/core/tritonserver.h
```diff
@@ -161,7 +161,6 @@ TRITONSERVER_EXPORT bool TRITONSERVER_LogIsEnabled(TRITONSERVER_LogLevel level);
 /// \param filename The file name of the location of the log message.
 /// \param line The line number of the log message.
 /// \param msg The log message.
-/// \return a TRITONSERVER_Error indicating success or failure.
```
deadeyegoodwin, line 164: Need to put this line back in.

PR #1816, opened 2020-07-22
Port TensorFlow backend with Triton backend API

GuanLuo:
`model_instance.h/.cc` is the backend API version of `backend_context.h/.cc`. `backend_utils.h/.cc` added utilities from `model_config_utils.h/.cc`

### DIFF
--- a/src/backends/backend/examples/backend_utils.cc
+++ b/src/backends/backend/examples/backend_utils.cc
@@ -428,7 +428,14 @@ GetTypedSequenceControlProperties(
 
             seen_control = true;
 
-            if (c.Find("int32_false_true") || c.Find("fp32_false_true")) {
+            TritonJson::Value int32_false_true, fp32_false_true;
+            bool found_int32 =
+                (c.Find("int32_false_true", &int32_false_true) &&
+                 (int32_false_true.ArraySize() > 0));
+            bool found_fp32 =
+                (c.Find("fp32_false_true", &fp32_false_true) &&
+                 (fp32_false_true.ArraySize() > 0));
+            if (found_int32 || found_fp32) {
               return TRITONSERVER_ErrorNew(
                   TRITONSERVER_ERROR_INVALID_ARG,
                   (std::string(
--- a/src/backends/backend/examples/backend_utils.cc
+++ b/src/backends/backend/examples/backend_utils.cc
@@ -270,8 +270,12 @@ GetBooleanSequenceControlProperties(
             seen_control = true;
 
             TritonJson::Value int32_false_true, fp32_false_true;
-            bool found_int32 = (c.Find("int32_false_true", &int32_false_true) && (int32_false_true.ArraySize() > 0));
-            bool found_fp32 = (c.Find("fp32_false_true", &fp32_false_true) && (fp32_false_true.ArraySize() > 0));
+            bool found_int32 =
+                (c.Find("int32_false_true", &int32_false_true) &&
+                 (int32_false_true.ArraySize() > 0));
+            bool found_fp32 =
+                (c.Find("fp32_false_true", &fp32_false_true) &&
+                 (fp32_false_true.ArraySize() > 0));
             if (found_fp32 && found_int32) {
               return TRITONSERVER_ErrorNew(
                   TRITONSERVER_ERROR_INVALID_ARG,
--- a/src/backends/backend/examples/backend_utils.cc
+++ b/src/backends/backend/examples/backend_utils.cc
@@ -68,6 +68,126 @@ ShapeToString(const std::vector<int64_t>& shape)
   return ShapeToString(shape.data(), shape.size());
 }
 
+std::string
+DataTypeToString(TRITONSERVER_DataType dtype)
+{
+  switch (dtype) {
+    case TRITONSERVER_TYPE_INVALID:
+      return "INVALID";
+    case TRITONSERVER_TYPE_BOOL:
+      return "BOOL";
+    case TRITONSERVER_TYPE_UINT8:
+      return "UINT8";
+    case TRITONSERVER_TYPE_UINT16:
+      return "UINT16";
+    case TRITONSERVER_TYPE_UINT32:
+      return "UINT32";
+    case TRITONSERVER_TYPE_UINT64:
+      return "UINT64";
+    case TRITONSERVER_TYPE_INT8:
+      return "INT8";
+    case TRITONSERVER_TYPE_INT16:
+      return "INT16";
+    case TRITONSERVER_TYPE_INT32:
+      return "INT32";
+    case TRITONSERVER_TYPE_INT64:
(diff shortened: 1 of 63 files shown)


src/backends/backend/examples/backend_utils.cc
```diff
@@ -35,10 +38,18 @@ std::string
...
+    case Kind::CPU:
+      str += "CPU";
+      break;
+    case Kind::GPU:
+      str += "GPU (" + std::to_string(device_id_) + ")";
+      break;
+    case Kind::MODEL:
+      str += "MODEL";
+      break;
+    default:
```
deadeyegoodwin, line 51: Don't add default, so that we get a compiler error if we add an enum and don't add it to this switch.

src/backends/backend/examples/backend_utils.cc
```diff
@@ -135,6 +151,126 @@ ShapeToString(const std::vector<int64_t>& shape)
   return ShapeToString(shape.data(), shape.size());
 }
 
+std::string
+DataTypeToString(TRITONSERVER_DataType dtype)
```
deadeyegoodwin, line 155: Instead of this should use TRITONSERVER_DataTypeString which is now available in TRITONSERVER API.

src/backends/backend/examples/backend_utils.cc
```diff
@@ -135,6 +151,126 @@ ShapeToString(const std::vector<int64_t>& shape)
...
+  int64_t cnt = GetElementCount(dims);
+  if (cnt == -1) {
+    return -1;
+  }
+
+  return cnt * dt_size;
+}
+
+size_t
+GetDataTypeByteSize(const TRITONSERVER_DataType& dtype)
```
deadeyegoodwin, line 238: Use TRITONSERVER_DataTypeByteSize

src/backends/backend/tensorflow/tensorflow.cc
```diff
@@ -0,0 +1,1923 @@
...
+ModelState::ProcessRequest(
+    TRITONBACKEND_Request** requests, const uint32_t request_count,
+    const uint64_t exec_start_ns)
+{
+  inflight_thread_count_++;
+  auto instance = available_instances_.Pop();
+
+  // Currently launch thread for each batch, but we may launch long-running
+  // thread for each instance.
+  std::thread process_thread(
```
deadeyegoodwin, line 1346: I think performance-wise this will be an issue, but we can revisit that later. Also will likely need to run using a specific instance due to rate-limiter changes.

src/backends/backend/tensorflow/tensorflow.cc
```diff
@@ -0,0 +1,1923 @@
...
+  // *for a given model*. But since this backend could be used by
+  // multiple models the implementation needs to handle multiple
+  // models executing at the same time. Good practice for this is to
+  // use only function-local and model-specific state (obtained from
+  // 'model'), which is what we do here.
+  ModelState* state;
+  RETURN_IF_ERROR(
+      TRITONBACKEND_ModelState(model, reinterpret_cast<void**>(&state)));
+
+  // At this point we accept ownership of 'requests', which means that
```
deadeyegoodwin, line 1906: There is another "ownership accepting" note above... I think this one is correct but just need to make sure ownership is handled correctly.

src/core/model_config_utils.cc
```diff
@@ -603,17 +603,6 @@ GetNormalizedModelConfig(
     RETURN_IF_ERROR(ReadTextProto(config_path, config));
   }
```
deadeyegoodwin, line 605: Here we should normalize platform and backend based on the documentation link I sent. And in validate we should make sure that platform and backend are valid values.

src/core/server_message.h
```diff
@@ -46,13 +46,29 @@ class TritonServerMessage {
...
 
   TritonServerMessage(std::string&& msg)
   {
     str_buffer_ = std::move(msg);
     base_ = str_buffer_.data();
     byte_size_ = str_buffer_.size();
+    from_json_ = false;
+  }
+
+  TritonServerMessage(const TritonServerMessage& rhs)
```
deadeyegoodwin, line 60: Where are we copying a message?

GuanLuo: When instantiating the TritonBackend object https://github.com/NVIDIA/triton-inference-server/blob/master/src/backends/backend/triton_backend_manager.cc#L119

deadeyegoodwin: OK, I guess it was copying incorrectly before...

src/backends/backend/tensorflow/tensorflow.cc
GuanLuo: Are you referring to the comment [here](https://github.com/NVIDIA/triton-inference-server/blob/9d30c5a7be030f53cb83fafd9d980fa5ddf60196/src/backends/backend/tensorflow/tensorflow.cc#L1431-L1441)? I think both are correct, the other note is actually saying that the request will not be released until the end even if error occurs during the execution.

deadeyegoodwin: Yes, you are right.

src/backends/backend/tensorflow/tensorflow.cc
GuanLuo: I ran perf_client (on `graphdef_float32_float32_float32`) and it did show performance-wise issue, changed to long-running instance thread helps but it still shows that is less performant than not using backend API.

src/core/model_config_utils.cc
```diff
@@ -603,81 +603,82 @@ GetNormalizedModelConfig(
...
-
-      LOG_VERBOSE(1) << "autofilled config: " << config->DebugString();
-    }
+    LOG_VERBOSE(1) << "autofilled config: " << config->DebugString();
+  }
 
-    if (config->platform().empty()) {
-      return Status(
-          Status::Code::INVALID_ARG,
-          "must specify platform for model '" + config->name() + "'");
```
GuanLuo, line 632: @deadeyegoodwin I changed the logic here and removed this check. The idea is that current autofill will simply do nothing for backend API (except for supported platform), so autofill path doesn't need to condition on `config->backend().empty()`. Then platform may still be empty after autofill.

src/core/infer_request.h
```diff
@@ -304,9 +304,13 @@ class InferenceRequest {
 
   // Add an override input to the request. If 'input' is non-null
   // return a pointer to the newly added input.
+  // FIXME passing batch size is special handling for backend API.
```
deadeyegoodwin, line 307: So once all backends move to new API we can get rid of the batch_size parameter?

GuanLuo: Yes, the shape will change to contain batch size if support batching, since all backend can only access input properties via backend API which only provides shape with batch dim

Issue #1472, opened 2020-05-12
Enable  AMP without any  TensorRT optimization for Tensorflow  in the inference server.


whatdhack:
**Is your feature request related to a problem? Please describe.**
Would be nice to be able to just enable FP16  (e.g. TF_ENABLE_AUTO_MIXED_PRECISION ) without any TensorRT optimization in the inference server. 

**Describe the solution you'd like**
Just FP16 , no other optimization

**Describe alternatives you've considered**
Running Tensorflow directly 

**Additional context**
Using TensorRT optmization appears to cause adverse issues, such as OOM,  in large networks such as MaskRCNN.  However Tensorflow with just FP16 does not have this issue.

deadeyegoodwin:
TF_ENABLE_AUTO_MIXED_PRECISION refers to AMP which is targeted at training, not inference. Not sure if/how it would work with inference. Being an environment variable doesn't really work for a multi-model server like Triton. It would be great if TF introduced a session configuration option to select mixed precision for a model. They are working on new runtime optimizations so perhaps that is something they will add and then Triton could adopt it and expose it via the existing optimization settings in the model configuration.

whatdhack:
I just ran some tests on a MaskRCNN Saved Model  in nvcr.io/nvidia/tensorflow:20.03-tf1-py3.  Seems to work very well for inference - requires less memory and speeds up significantly.  The following are the numbers. 

**TF_ENABLE_AUTO_MIXED_PRECISION =1, memory = 4.2GB , inference speed  0.25 sec .**
vs53 
**memory = 7.1 GB , inference speed  0.53 sec .**

whatdhack:
@deadeyegoodwin , I took a quick look at https://github.com/NVIDIA/triton-inference-server/blob/master/src/backends/tensorflow/tensorflow_backend_tf.cc , if there is place where I can add something similar to TF_ENABLE_AUTO_MIXED_PRECISION.  What TF_ENABLE_AUTO_MIXED_PRECISION  would be equivalent to inside the tensorflow_backend code ?

deadeyegoodwin:
I don't know of a TF configuration setting that you could use... you can look though the config protobuf to see if there is something.

whatdhack:
@deadeyegoodwin ,  The following option is available in plain TF , works in 1.15.2 .  Is there a way to set this option in the TF library used by Triton  ?

config.graph_options.rewrite_options.auto_mixed_precision = 1

[...]

PR #1817, opened 2020-07-22
Update docs with TF mixed-precision optimization

deadeyegoodwin:


deadeyegoodwin:
Related to #1472

docs/optimization.rst
```diff
@@ -308,6 +306,27 @@ cutting latency in half. The benefit provided by TensorRT will vary
 based on the model, but in general it can provide significant
 performance improvement.
 
+TensorFlow Automatic FP16 Optimization
+......................................
+
+TensorFlow has another option to provide FP16 optimization that can be
```
CoderHam, line 312: provide speedup by using the FP16

deadeyegoodwin: I don't want to say "speedup" explicitly. Already calling this an optimization.

CoderHam: Makes sense. Resolving

PR #1819, opened 2020-07-23
Fix S3 support for arbitrary model directory depth

CoderHam:


src/core/filesystem.cc
```diff
@@ -978,57 +978,62 @@ S3FileSystem::LocalizeDirectory(
...
+        RETURN_IF_ERROR(ParsePath(s3_fpath, &file_bucket, &file_object));
+
+        s3::Model::GetObjectRequest object_request;
+        object_request.SetBucket(file_bucket.c_str());
+        object_request.SetKey(file_object.c_str());
+
+        auto get_object_outcome = client_.GetObject(object_request);
+        if (get_object_outcome.IsSuccess()) {
+          auto& retrieved_file =
+              get_object_outcome.GetResultWithOwnership().GetBody();
```
deadeyegoodwin, line 1028: Is there no API to just save an object directly to disk?

CoderHam: No such API via the c++ sdk

PR #1823, opened 2020-07-23
Make sure that all unittests are executed

Tabrizian:
I added a line at the end of `$CLIENT_LOG` which contains a simple JSON showing the number of tests failed and executed. If this looks good to you, I will add this to all other unittests. The test result report looks like this:

```JSON
{"total": 42, "errors": 0, "failures": 0, "skipped": 0}
```


qa/L0_infer/test.sh
```diff
@@ -168,23 +168,33 @@ for TARGET in cpu gpu; do
...
-    grep -c "HTTPSocketPoolResponse status=200" $CLIENT_LOG
+    # At the end of $CLIENT_LOG there is a single line JSON containing the
+    # result of unittests.
+    test_result_json=`tail -n 1 $CLIENT_LOG`
+
+    echo $test_result_json | jq
     if [ $? -ne 0 ]; then
         cat $CLIENT_LOG
-        echo -e "\n***\n*** Test Failed To Run\n***"
+        echo -e "\n***\n*** Test Failed\n***"
```
deadeyegoodwin, line 185: "Test Failed: unable to parse test results"

qa/L0_infer/test.sh
```diff
@@ -168,23 +168,33 @@ for TARGET in cpu gpu; do
...
         RET=1
+    else
+        num_failures=`echo $test_result_json | jq .failures`
+        num_tests=`echo $test_result_json | jq .total`
+        if [ $num_failures != "0" ] || [ $num_tests -le 0 ]; then
```
deadeyegoodwin, line 190: $num_tests should be compared against the exact number of tests that are expected so that we can detect if fewer were run for some reason. I would add an EXPECTED_NUM_TESTS at the start of the file which the expected value. When someone changes the tests they will need to also update that.

qa/L0_infer/test.sh
```diff
@@ -168,23 +168,33 @@ for TARGET in cpu gpu; do
...
+            cat $CLIENT_LOG
+            echo -e "\n***\n*** Test Failed To Run\n***"
```
deadeyegoodwin, line 192: Include num_failures, num_tests and expected number of tests in this output.

qa/L0_infer/test.sh
```diff
@@ -168,23 +168,33 @@ for TARGET in cpu gpu; do
...
+            RET=1
+        fi
     fi
 
+
```
deadeyegoodwin, line 197: Don't have multiple blank lines.

qa/common/test_util.py
```diff
@@ -230,3 +232,40 @@ def get_dyna_sequence_model_name(pf, dtype):
...
+
+    current_result = None
+
+    @classmethod
+    def setResult(cls, amount, errors, failures, skipped):
+        cls.amount, cls.errors, cls.failures, cls.skipped = \
+            amount, errors, failures, skipped
+
+    def tearDown(self):
+        # this is called immediately after the test method
```
deadeyegoodwin, line 249: So current_result contains the aggregate result not only of the test that just finished but also of all previously run tests? Otherwise I don't see where you are accumulating the counts.

Tabrizian: `current_result` contains the aggregate count after all the test methods finish execution. We only need to pass the same result object to `run` so that it accumulates the results. As far as I have checked, each `run` is called once for every test method in the UnitTest class (e.g. InferTest).

qa/L0_infer/test.sh
Tabrizian: Done.

qa/L0_infer/test.sh
```diff
@@ -72,6 +72,9 @@ export BACKENDS
 ENSEMBLES=${ENSEMBLES:="1"}
 export ENSEMBLES
 
+# Update if the total number of tests changed
+EXPECTED_NUM_TESTS="42"
```
deadeyegoodwin, line 76: Move this up to line 42. Don't need the comment on it... it will become a standard thing in all test.sh that use python unittest

qa/L0_infer/test.sh
```diff
@@ -39,6 +39,7 @@ export CUDA_VISIBLE_DEVICES=0
 
 CLIENT_LOG_BASE="./client"
 INFER_TEST=infer_test.py
+EXPECTED_NUM_TESTS="42"
```
CoderHam, line 42: This might be different for the shared memory variants of L0_infer. You can run the system shared memory variant by setting TEST_SYSTEM_SHARED_MEMORY=1 and confirm the same. Specifically the tests that use class output instead of raw are not run for the shared memory. We can discuss this offline if you have any questions.

qa/L0_infer/test.sh
```diff
@@ -72,6 +73,7 @@ export BACKENDS
 ENSEMBLES=${ENSEMBLES:="1"}
 export ENSEMBLES
 
+
```
CoderHam, line 76: remove extra line

qa/L0_infer/infer_test.py
```diff
@@ -47,7 +47,7 @@
 np_dtype_string = np.dtype(object)
 
 
-class InferTest(unittest.TestCase):
+class InferTest(tu.TestResCollector):
```
CoderHam, line 50: Shouldn't the unittest import no longer be needed after this change?

Tabrizian: Sorry I missed this comment. You're right, I will remove those imports.

Tabrizian: It is required in the end of files for running the unittest.

qa/L0_s3_local/test.sh
```diff
@@ -90,7 +88,7 @@ RET=0
 SERVER_ARGS="--model-repository=s3://localhost:4572/demo-bucket1.0 --model-control-mode=explicit"
 SERVER_LOG="./inference_server_hostname.log"
 
-BACKENDS="graphdef libtorch netdef onnx plan savedmodel custom"
+BACKENDS="graphdef libtorch netdef onnx plan savedmodel"
```
CoderHam, line 91: `git rebase master` to pick up latest changes to this file.

qa/L0_storage_S3/test.sh
```diff
@@ -152,14 +153,11 @@ for ENV_VAR in "env" "env_dummy" "config"; do
...
 
-        # Copy custom model 
-        cp -r /opt/tritonserver/qa/custom_models/custom_float32_float32_float32/ models/
-
         # Copy models with string inputs and remove nobatch (bs=1) models
         cp -r /data/inferenceserver/${REPO_VERSION}/qa_model_repository/*_object_object_object/ models/
         rm -rf models/*nobatch*
 
-        for FW in graphdef savedmodel netdef onnx libtorch plan custom; do
+        for FW in graphdef savedmodel netdef onnx libtorch plan; do
```
CoderHam, line 160: Same. Rebase onto master and confirm the changes.

qa/common/util.sh
```diff
@@ -268,3 +268,27 @@ function create_nop_modelfile () {
...
+        num_tests=`echo $test_result_json | jq .total`
+        if [ $num_failures != "0" ] || [ $num_tests -ne $expected_num_tests ]; then
+            cat $log_file
+            echo -e "\n***\n*** Test Failed: Expected $expected_num_tests test(s), $num_tests test(s) executed, and $num_failures test(s) failed. \n***"
+            return 1
+        fi
+    fi
+
+    return 0
+}
```
CoderHam, line 299: add back deleted newline at end of file

.gitignore
```diff
@@ -1,2 +1,3 @@
 /builddir
 /.vscode
+*.so
```
CoderHam, line 3: do not delete the newline at the end of files

qa/L0_batcher/batcher_test.py
```diff
@@ -59,7 +59,7 @@
 _deferred_exceptions = []
 
 
-class BatcherTest(unittest.TestCase):
+class BatcherTest(tu.TestResCollector):
```
deadeyegoodwin, line 62: Don't use non-standard abbreviations in names when possible, so use TestResultCollector

qa/L0_batcher/batcher_test.py
```diff
@@ -77,6 +77,7 @@ def tearDown(self):
             destroy_fun = cudashm.destroy_shared_memory_region
         for precreated_shm_region in self.precreated_shm_regions_:
             destroy_fun(precreated_shm_region)
+        super().tearDown()
```
deadeyegoodwin, line 80: Why do you have to explicitly call the super here, but not for setUp?

Tabrizian: The class that I have written for collecting the test results doesn't need to use `setUp`. Result collection happens in the `run` method which is not explicitly called in subclasses. However, I need to ensure that `tearDown` is always called since storing the test result happens there.

Tabrizian:
@deadeyegoodwin @CoderHam I fixed all the problems. There are 6 tests failing which is not related to these changes.

deadeyegoodwin:
> @deadeyegoodwin @CoderHam I fixed all the problems. There are 6 tests failing which are not related to these changes.

Great! Can you rebase to master then and re-run CI. None of the test should fail if you base off of master.

Tabrizian:
@deadeyegoodwin None of the tests is failing now. There is only a performance test failing which I saw that happens in other MRs as well.

PR #1824, opened 2020-07-23
GCS localize models similar to S3

CoderHam:
- [x]  Supporting arbitrary directory depth

Do we want to merge this in or wait for the GCS test to be added to the CI?


awhillas:
GCS please!

PR #1828, opened 2020-07-24
Use 'inference' package for GRPC protobuf

deadeyegoodwin:
The GRPC protobuf implements the KFserving open inferencing protocols so we don't want to tie it to Triton with a package. Also use same package for the model configuration protobuf since that is used in a GRPC protocol extension.

### DIFF
--- a/src/backends/backend/backend_factory.h
+++ b/src/backends/backend/backend_factory.h
@@ -48,7 +48,7 @@ class TritonBackendFactory {
 
   Status CreateBackend(
       const std::string& model_repository_path, const std::string& model_name,
-      const int64_t version, const ModelConfig& model_config,
+      const int64_t version, const inference::ModelConfig& model_config,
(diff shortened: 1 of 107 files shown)


CoderHam:
The build failed. You will need to remove the namespace ni (nvidia::inferenceserver) from the clients where it is used as a prefix to the protobuf messages.

CoderHam:
These changes look good to me. It will likely break backward compatibility and I think it would be useful to state this in the release notes. We see some cases where the clients use an older version of our clientsdk container with a newer version of server (Although we advise them to use the same release of both generally)

deadeyegoodwin:
This is in the V2 protocols. But we do mention that V2 is still beta and so this kind of breakage can happen. I'm hoping that in 20.09 we will exit beta and then we will make a stronger statement of compatibility.

PR #1830, opened 2020-07-24
Fix L0_storage_GCS to work in CI

CoderHam:
- Add coverage for GCS polling
- Moved yo gitlab to secure sensitive information


qa/L0_storage_GCS/test.sh
```diff
@@ -180,6 +205,47 @@ for MAYBE_SLASH in "" "/"; do
...
+# exit code. So need to explicitly check CLIENT_LOG to make sure
+# we see some running tests
+python $INFER_TEST >$CLIENT_LOG 2>&1
+if [ $? -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+grep -c "HTTPSocketPoolResponse status=200" $CLIENT_LOG
```
GuanLuo, line 234: Is this still necessary after the change in unit test checking?

GuanLuo: I think the corresponding Python script need to inherit from `tu.TestResultCollector`

CoderHam: Fixed. The script already does inherit from `tu.TestResultCollector`

PR #1831, opened 2020-07-24
Add TRITONBACKEND_ModelInstance to backend API

deadeyegoodwin:


src/backends/backend/triton_model.cc
```diff
@@ -208,16 +212,22 @@ TritonModel::Create(
...
-        // exec_fn as the last reference to the model. So we hold a
+        // exec_fn as the last reference to the model). So we hold a
         // copy of the backend here... This convoluted flow will be
         // cleaned up once legacy InferenceBackend is replaced with
         // TritonModel.
         std::shared_ptr<TritonBackend> backendx = backend;
 
+        TRITONBACKEND_ModelInstance* triton_model_instance =
+            reinterpret_cast<TRITONBACKEND_ModelInstance*>(
+                raw_local_model->instances_[runner_idx].get());
```
GuanLuo, line 223: In the porting TF 1 PR, there was an issue of using thread_local vector to store requests for multiple instance cases, and the WAR is having each instance to have a placeholder for requests to the instance like [this](https://github.com/NVIDIA/triton-inference-server/pull/1816/files#diff-189b3e56f7ed6b4bcf51a526919da4c5R1373-R1380). Since now there is instance abstraction, maybe this can be done internally? Or it will not be a concern with this change, `runner_cnt` is the same as the number of instances so maybe there is always one thread_local vector for each of the instances?

deadeyegoodwin: thread_local is still what we want, since we don't want to realloc the vector every time but we do need to have a separate copy of that vector for each thread. I will fix the comment since it still refers to a "single" thread and that is no longer the case.

GuanLuo: I guess the question is whether 2 concurrent executions will have access to the same vector. The TF 1 backend was like a "non-blocking" model because there was one scheduler thread and multiple instances, so the vector became a critical section and the WAR was needed. I don't know if this is still true after your model instance changes.

PR #1836, opened 2020-07-28
Future-proof location of model artifacts in backend API

deadeyegoodwin:


src/backends/backend/examples/identity.cc
```diff
@@ -404,14 +404,18 @@ TRITONBACKEND_ModelInitialize(TRITONBACKEND_Model* model)
...
-  const char* cdir;
-  RETURN_IF_ERROR(TRITONBACKEND_ModelRepositoryPath(model, &cdir));
-  std::string dir(cdir);
-
+  // Can get location of the model artifacts. Normally we would need
+  // to check the artifact type to make sure it was something we can
+  // handle... but we are just going to log the location so we don't
+  // need the check. We would use the location if we wanted to load
+  // something from the model's repo.
+  TRITONBACKEND_ModelArtifactType artifact_type;
```
CoderHam, line 412: Where would the TRITONBACKEND_ModelArtifactType be consumed? I see only one ModelArtifactType i.e. FILESYSTEM, do we plan on adding others?

deadeyegoodwin: Yes, we will need to add an in-memory way of getting the model artifacts at some point and for that there will be a new type and the 'location' value will be a key/handle that will be used with a (not yet added) API to callback into user code to get the in-memory copy of a model configuration file or other artifact.

PR #1841, opened 2020-07-28
Set device before destroying TRT / CUDA resources in plan backend

GuanLuo:


GuanLuo:
@tanmayv25 This is an issue raised for other use case where an plugin's destructor asserts on the current device to be the same as the current device during its creation. We can argue that the plugin should really set device for itself, but it is just good practice for Triton to do it explicitly, as the cudaFree will cause issue if peer access / unified memory is not enabled, although that is less likely to happen.

PR #1850, opened 2020-07-30
Add Triton backend API for model config completion

GuanLuo:
@deadeyegoodwin This is how the APIs will look like


GuanLuo:
One of the L0_model_config test is updated because the new auto-complete relies on the platform / backend to be specified (or auto-completed based on default file name) before loading the model, whereas the test case was trying to loading the model and used the load result as hint for platform / backend.

src/backends/backend/tensorflow/tensorflow.cc
```diff
@@ -1849,6 +2149,60 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
         "triton backend API version does not support this backend");
   }
 
+  // The backend configuration may contain information needed by the
```
deadeyegoodwin, line 2152: This message needs updating

src/backends/backend/tensorflow/tensorflow.cc
```diff
@@ -1849,6 +2149,60 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
...
+      TRITONSERVER_LOG_INFO,
+      (std::string("backend configuration:\n") + buffer).c_str());
+
+  ni::TritonJson::Value backend_config;
+  if (byte_size != 0) {
+    RETURN_IF_ERROR(backend_config.Parse(buffer, byte_size));
+  }
+
+  // Set default backend values. Note that those values should be set during
+  // TRITONBACKEND_Initialize, but for TensorFlow they are set while creating
```
deadeyegoodwin, line 2173: I don't understand this comment... we are setting the values and we are in TRITONBACKEND_Initialize

src/backends/backend/triton_model.cc
```diff
@@ -171,9 +197,17 @@ TritonModel::Create(
...
   // Create and initialize the model.
   std::unique_ptr<TritonModel> local_model(new TritonModel(
-      server, localized_model_dir, backend, min_compute_capability));
-  RETURN_IF_ERROR(
-      local_model->Init(version_path, model_config, "" /* platform */));
+      server, localized_model_dir, backend, min_compute_capability,
+      auto_complete_config));
+  // If request for auto completion, Init() will be postponed until
+  // UpdateModelConfig() is called as Init() assumes the model config
+  // is well-formed.
```
deadeyegoodwin, line 204: This isn't ideal because if the backend never calls SetModelConfig then Init will not be called. I think it it ok for now as this complicated flow will go away once all backends are moved over and we can get rid of the InferenceBackend class.

src/backends/backend/tritonbackend.h
```diff
@@ -593,6 +593,37 @@ TRITONBACKEND_EXPORT TRITONSERVER_Error* TRITONBACKEND_ModelConfig(
...
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONBACKEND_EXPORT TRITONSERVER_Error* TRITONBACKEND_ModelAutoCompleteConfig(
+    TRITONBACKEND_Model* model, bool* auto_complete_config);
+
+/// Set the model configuration in Triton server. Only the inputs, outputs,
+/// and max batch size can be changed. Any other changes to the model
+/// configuration will be ignored by Triton. This function can only be called
+/// from TRITONBACKEND_ModelInitialize, calling in any other context will result
+/// in an error being returned. The function does not take ownership of the
+/// message object and the caller can call TRITONSERVER_MessageDelete to release
```
deadeyegoodwin, line 614: ".. and so the caller should call ..."

src/core/model_config_utils.cc
```diff
@@ -1659,4 +1668,29 @@ ModelConfigToJson(
   return Status::Success;
 }
 
+Status
+JsonToModelConfig(
+    const std::string& json_config, const uint32_t config_version,
+    inference::ModelConfig* protobuf_config)
+{
+  // Currently only support 'config_version' 1, which is the json
+  // representation of the ModelConfig protobuf with the int64 fields
```
deadeyegoodwin, line 1677: Fix comment... we are not doing anything with int64 fields here

src/backends/backend/triton_model.cc
GuanLuo: Okay. I added a FIXME comment as reminder.

PR #1852, opened 2020-07-31
Update NGC versions post-20.07 release

dzier:
This also contains the update to the README and docs post-20.07 release.

Do Not Squash

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,9 +28,9 @@
 # Multistage build.
 #
 
-ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:20.06-py3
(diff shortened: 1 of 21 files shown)


src/core/dynamic_batch_scheduler.cc
```diff
@@ -372,6 +373,7 @@ DynamicBatchScheduler::SchedulerThread(
             for (auto& request : requests) {
               completion_queue_.emplace_back();
               auto queue_slot = &completion_queue_.back();
+              queue_slot->second = 0;
```
GuanLuo, line 376: I think this two lines are not needed

dzier: Yes, just saw this. It failed to build. I have removed the line and restarted the CI.

PR #1867, opened 2020-08-05
Remove TF lib dependency from server core. Add command line option for selecting TF libary version

GuanLuo:
Note that `libtriton_tensorflow2.so` is still linking against Triton-TF 1 library, can either postpone this change until Triton-TF 2 library is available or apply followup change by that time.


src/backends/backend/triton_model.cc
```diff
@@ -150,8 +150,25 @@ TritonModel::Create(
   RETURN_IF_ERROR(LocalizeDirectory(
       JoinPath({model_repository_path, model_name}), &localized_model_dir));
 
-  const std::string backend_libname =
-      "libtriton_" + model_config.backend() + ".so";
+  std::string backend_libname = "libtriton_" + model_config.backend() + ".so";
+  if (model_config.backend() == "tensorflow") {
```
deadeyegoodwin, line 154: You'll have to adjust the backend name used for the global path below as well. So that with /opt/tritonserver/backends/tensorflow1 or /opt/tritonserver/backends/tensorflow2 is used.

PR #1874, opened 2020-08-06
Add python backend

Tabrizian:


src/backends/backend/python/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+
+#
+# libpython_interpreter.so
+#
+configure_file(libtriton_python.ldscript libtriton_python.ldscript COPYONLY)
+
+find_package(gRPC CONFIG REQUIRED)
+find_package(Protobuf CONFIG REQUIRED)
+
+
```
CoderHam, line 37: Extra line. Remove.

GuanLuo:
Is this PR ready for review?

Tabrizian:
@GuanLuo It is not completely ready yet. There are a couple of issues I need to fix. Specifically, I need to fix these:

1. Batching of the inference requests.
2. Add config for Python runtime
3. Fix the model inference statistics.
4. Fix installation of python protobuf files.

But I created this so that I can apply the comments that you may have along the way.

deadeyegoodwin:
@Tabrizian please don't try to do all that in one giant change. It will be very difficult to review such a large change. Try to get the minimum working and then incrementally improve with subsequent PRs.

Tabrizian:
@deadeyegoodwin Sorry about creating this large PR. This example currently works. I have compiled and ran it with the test clients that I have created. It is mostly cleaning up the codebase received from the other team + migration to the new API + fixing the build to work under the new tree structure. If we can get this merged, I will create a separate PR for each of the fixes I mentioned in https://github.com/NVIDIA/triton-inference-server/pull/1874#issuecomment-671557238 and they will be small incremental PRs.

deadeyegoodwin:
@Tabrizian OK, please remove the [WIP] if you are ready for review.

Tabrizian:
@deadeyegoodwin It should be ready now.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+    RETURN_IF_ERROR(TRITONBACKEND_ResponseNew(&response, request));
+    responses.push_back(response);
+  }
+
+  ModelState* model_state = instance_state->StateForModel();
+
+  bool supports_batching_initialized = false;
+  bool supports_batching = false;
+
+  if (!supports_batching_initialized) {
```
deadeyegoodwin, line 734: You just set 'supports_batching_initialized' to false, above. Also none of the support_batching variables are even used... I don't think this backend needs to understand anything about batching... it is just passing the requests onto the python.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+    grpc::ClientContext context;
+    ni::InferenceBatch result_batch;
+
+    const auto err = instance_state->stub->InferenceRequest(
+        &context, input_batch, &result_batch);
+
+    if (!err.ok()) {
+      LOG_ERROR << "something went wrong with interfere request\n"
+                << err.error_message();
+    }
```
deadeyegoodwin, line 779: You cannot just log an error and continue. You need to send an error response for each request using the response objects you already created. See identity.cc for details on how error handling must be implemented.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+#include <string>
+#include <thread>
+#include <unordered_map>
+#include <vector>
+
+// Fix include path for protobuf
+#include "python_host.grpc.pb.h"
+#include "src/backends/backend/tritonbackend.h"
+#include "src/core/json.h"
+#include "src/core/model_config.h"
```
deadeyegoodwin, line 65: You cannot use model_config.h. I don't think you should really be looking at the model configuration. Just pass it on to the python as part of Init as JSON. In general the model configuration is only used at initialization time to determine if the model configuration is compatible with this backend. This code itself cannot do that since it is just a communication wrapper with the python.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+
+namespace nic = nvidia::inferenceserver::custom;
```
deadeyegoodwin, line 67: Remove these. The only way you can talk to Triton is via the backend API. You can't use any headers from Triton except for json.h (this will be more obvious when you move to the new github repo).

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+    const auto err =
+        stub->InterpreterInit(&context, *initialization_params, &result);
+    if (err.ok() && result.code() == 0) {
+      return nic::ErrorCodes::Success;
+    } else {
+      std::this_thread::sleep_for(std::chrono::seconds(1));
+    }
+  }
+
+  std::this_thread::sleep_for(std::chrono::seconds(5));
```
deadeyegoodwin, line 402: Why? And why is there an InterpreterInit in the loop and also another InterpreterInit below.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+  uint64_t buffer_byte_size = 0;
+  TRITONSERVER_MemoryType input_memory_type = TRITONSERVER_MEMORY_CPU;
+  int64_t input_memory_type_id = 0;
+  for (size_t j = 0; j < input_buffer_count; ++j) {
+    GUARDED_RESPOND_IF_ERROR(
+        responses, r,
+        TRITONBACKEND_InputBuffer(
+            in, j, &input_buffer, &buffer_byte_size, &input_memory_type,
+            &input_memory_type_id));
+    data_buffer->append((const char*)input_buffer, buffer_byte_size);
```
deadeyegoodwin, line 538: What if buffer that comes back is not in TRITONSERVER_MEMORY_CPU?

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+  }
+
+  return nic::ErrorCodes::Success;
+}
+
+TRITONSERVER_Error*
+ModelState::Create(TRITONBACKEND_Model* triton_model, ModelState** state)
+{
+  TRITONSERVER_Message* config_message;
+  // TODO: Make this work with every version
```
deadeyegoodwin, line 548: Remove TODO. You are requesting "version 1" format of the configuration, and that is what Triton will give you.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+  uint32_t input_dims_count;
+  uint64_t input_byte_size;
+  uint32_t input_buffer_count;
+  GUARDED_RESPOND_IF_ERROR(
+      responses, r,
+      TRITONBACKEND_InputProperties(
+          in, &input_name, &input_dtype, &input_shape, &input_dims_count,
+          &input_byte_size, &input_buffer_count));
+
+  if (iidx == 0) {
```
deadeyegoodwin, line 514: No, you can't do this in general to find batch size. Also you don't need to know the batch size.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+    batch_size = input_shape[0];
+  }
+
+  // Update input_tensor.
+  input_tensor->set_name(input_name);
+  input_tensor->set_dtype(static_cast<inference::DataType>(input_dtype));
```
deadeyegoodwin, line 520: Don't use Types from model_config.proto. You should just send the TRITONSERVER_DataType directly. We should provide a utility on the python side to "decode" it.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+    const char* model_name, const uint64_t model_version,
+    ni::TritonJson::Value&& model_config)
+    : triton_server_(triton_server), triton_model_(triton_model),
+      model_name_(model_name), model_version_(model_version),
+      model_config_(std::move(model_config))
+{
+}
+
+TRITONSERVER_Error*
+ModelState::SupportsFirstDimBatching(bool* supports)
```
deadeyegoodwin, line 587: remove

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+
+  TRITONBACKEND_Backend* backend;
+  RETURN_IF_ERROR(TRITONBACKEND_ModelBackend(model, &backend));
+
+  ModelState* model_state;
+  RETURN_IF_ERROR(ModelState::Create(model, &model_state));
+  RETURN_IF_ERROR(
+      TRITONBACKEND_ModelSetState(model, reinterpret_cast<void*>(model_state)));
+
+
```
deadeyegoodwin, line 654: Remove extra blank lines

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,902 @@
...
+    // Time to output.
+    ni::TritonJson::Value outputs;
+    LOG_IF_ERROR(
+        instance_state->model_config.MemberAsArray("output", &outputs),
+        "can't get model outputs");
+
+    for (size_t j = 0; j < requested_output_count; ++j) {
+      // Prepare output buffers.
+      TRITONBACKEND_Output* output;
```
deadeyegoodwin, line 789: The request itself will indicate the name of the outputs it expects, you don't need to look at the model configuration. Look carefully at identity.cc to see how outputs are handled. We are going to have to send over to the python which outputs we want produced so it knows what to return. But whatever it returns we should just return those outputs are they are given to from the python.

src/backends/backend/python/python_host.proto
```diff
@@ -0,0 +1,70 @@
...
+  message ValuePair
+  {
+    string key = 1;
+    string value = 2;
+  }
+
+  repeated ValuePair model_command = 1;
+}
+
+message InferenceData
```
deadeyegoodwin, line 44: message Tensor
This describes a request or response tensor.

src/backends/backend/python/python_host.proto
```diff
@@ -0,0 +1,70 @@
...
+message StatusCode
+{
+  int64 code = 1;
+}
+
+message Empty {}
+
+service PythonInterpreter
+{
+  rpc InterpreterInit(InitializationCommand) returns (StatusCode) {}
```
deadeyegoodwin, line 67: Make these verbs. Init(), Fini(), Execute().

src/backends/backend/python/python_host.proto
```diff
@@ -0,0 +1,70 @@
...
+  rpc InterpreterShutdown(Empty) returns (StatusCode) {}
+  rpc InferenceRequest(InferenceBatch) returns (InferenceBatch) {}
```
deadeyegoodwin, line 69: Execute(ExecuteRequest) returns (ExecuteResponse) { }

message InferenceResponse
{
  string name;
  repeated Tensor inputs;
}

message ExecuteResponse
{
  repeated InferenceResponse responses;
}

src/backends/backend/python/python_host.proto
```diff
@@ -0,0 +1,70 @@
...
+
+message InferenceData
+{
+  string name = 1;
+  inference.DataType dtype = 2;
+  repeated int64 dims = 3;
+  bytes raw_data = 5;
+}
+
+message InferenceBatch
```
deadeyegoodwin, line 52: This describes an inference request. batch_size is not needed.
message InferenceRequest
{
  string id;
  uint64 correlation_id;
  repeated Tensor inputs;
  repeated string requested_output_names;
}

message ExecuteRequest
{
  repeated InferenceRequest requests;
}

src/backends/backend/python/python_interpreter.cc
Tabrizian: Should I replicate the TRITONSERVER_DataType in the `python_host.proto` in order to remove the `model_config` dependency? I can convert the enum to int and use int in the protobuf but I'm not sure whether it is a good choice.

src/backends/backend/python/python_interpreter.cc
Tabrizian: Yes, they were not necessary. I removed them.

src/backends/backend/python/python_interpreter.cc
Tabrizian: Added an if statement to return error.

src/backends/backend/python/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+FILE(GLOB PROTOBUF_SRCS ${CMAKE_CURRENT_BINARY_DIR}/*.h)
+
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
+add_library(
+  triton_python SHARED
+  python_interpreter.cc
+  ${PROTOBUF_HEADERS}
+  ${PROTOBUF_SRCS}
+  $<TARGET_OBJECTS:model-config-library>
```
deadeyegoodwin, line 84: Should not use model-config-library

src/backends/backend/python/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+        triton_python
+        PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
+
+target_link_libraries(
+  triton_python
+  PRIVATE gRPC::grpc++
+)
+
+if(${TRTIS_ENABLE_GPU})
+target_include_directories(
```
deadeyegoodwin, line 115: Isn't this already given above?

src/backends/backend/python/python_host.proto
```diff
@@ -0,0 +1,81 @@
...
+{
+  string id = 1;
+  uint64 correlation_id = 2;
+  repeated Tensor inputs = 3;
+  repeated string requested_output_names = 4;
+}
+
+message InferenceResponse
+{
+  string name = 1;
```
deadeyegoodwin, line 60: This needs to be fixed as we discussed

src/backends/backend/python/python_utils.h
```diff
@@ -0,0 +1,56 @@
...
+// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+// OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#pragma once
+
+namespace nvidia { namespace inferenceserver { namespace backend {
+
+static const std::unordered_map<std::string, TRITONSERVER_DataType>
```
deadeyegoodwin, line 30: Remove these as we discussed

src/backends/backend/python/resources/startup.py
```diff
@@ -0,0 +1,163 @@
...
+    13: np.bytes_
+}
+
+
+def protobuf_to_numpy_type(data_type):
+    return _TYPE_MAPPING_DICTIONARY[data_type]
+
+
+def numpy_to_protobuf_type(data_type):
+    return {v: k for k, v in _TYPE_MAPPING_DICTIONARY.items()}[data_type]
```
deadeyegoodwin, line 48: This is very inefficient. Implement like protobuf_to_numpy_type, or generate the table once at startup.

src/backends/backend/python/resources/startup.py
```diff
@@ -0,0 +1,163 @@
...
+            for response_name, response_value in response.items():
+                name = response_name
+                tensor = Tensor(name=name,
+                                dtype=numpy_to_protobuf_type(
+                                    response_value.dtype.type),
+                                dims=response_value.shape,
+                                raw_data=response_value.tobytes())
+                tensors.append(tensor)
+            exec_responses.append(InferenceResponse(inputs=tensors))
+        execute_response = ExecuteResponse(responses=exec_responses)
```
deadeyegoodwin, line 147: Is this copying the entire exec_responses into responses? We want to construct ExecuteResponse in place with no copying.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -0,0 +1,841 @@
...
+  RETURN_IF_ERROR(TRITONBACKEND_ModelVersion(model, &version));
+
+  TRITONSERVER_LogMessage(
+      TRITONSERVER_LOG_VERBOSE, __FILE__, __LINE__,
+      (std::string("TRITONBACKEND_ModelInitialize: ") + name + " (version " +
+       std::to_string(version) + ")")
+          .c_str());
+
+  TRITONBACKEND_Backend* backend;
+  RETURN_IF_ERROR(TRITONBACKEND_ModelBackend(model, &backend));
```
deadeyegoodwin, line 572: 'backend' is not used...

src/backends/backend/python/resources/startup.py
Tabrizian: I don't think it is copying the responses. It is just the constructor argument.

src/backends/backend/python/python_utils.h
Tabrizian: Not sure why I added these sorry.

src/backends/backend/python/CMakeLists.txt
Tabrizian: It is not. The build failed in the CI complaining about `cuda_runtime_api` header.

src/backends/backend/python/resources/startup.py
deadeyegoodwin: Are you sure? So if after line 147 you were to modify exec_responses, those modifications would also appear in execute_response?

src/backends/backend/python/CMakeLists.txt
Tabrizian: Yes, it was given. Sorry.

src/backends/backend/python/python_host.proto
```diff
@@ -0,0 +1,80 @@
...
+{
+  string id = 1;
+  uint64 correlation_id = 2;
+  repeated Tensor inputs = 3;
+  repeated string requested_output_names = 4;
+}
+
+message InferenceResponse
+{
+  repeated Tensor inputs = 2;
```
deadeyegoodwin, line 60: Response should be "outputs" and the value should be 1

Tabrizian: Sure, I will fix this. Thanks.

PR #1875, opened 2020-08-06
Fix formatting of examples and unittests

Tabrizian:
These files pop up in the git changes whenever I run `python tools/format.py src qa`. This PR fixes this problem.


tanmayv25:
That is strange. It formats fine for me. Are you using python 3.x? What is the version for yapf on your machine?

Tabrizian:
Oh, sorry. You're right I was using Python 2 for formatting which seems to have caused the problem.

Issue #1428, opened 2020-05-07
Don't require curl CMake config


kpedro88:
**Is your feature request related to a problem? Please describe.**
The client CMakeLists only search for a CMake-based installation of curl: https://github.com/NVIDIA/triton-inference-server/blob/466dbdf96518e3a16fb4e39a97e29254affbe25b/src/clients/c%2B%2B/api_v1/library/CMakeLists.txt#L32 (and elsewhere).
However, the CMake installation of curl is described as "poorly maintained" and lacks some features:
https://github.com/curl/curl/blob/ace6ae4d0eb35eda85dbd076e76b1f34a52ebb06/CMakeLists.txt#L51
https://curl.haxx.se/docs/knownbugs.html#cmake_support_gaps
CMake is capable of finding curl in a variety of ways: https://github.com/Kitware/CMake/blob/master/Modules/FindCURL.cmake

**Describe the solution you'd like**
All lines `find_package(CURL CONFIG REQUIRED)` should be replaced with `find_package(CURL REQUIRED)`, to improve the flexibility of the build.

**Describe alternatives you've considered**
Currently using:
```
sed -i 's/find_package(CURL CONFIG REQUIRED)/find_package(CURL REQUIRED)/' CMakeLists.txt
```
but this is hacky.

**Additional context**
This will make building easier for users who require curl features that aren't supported in the CMake build.

PR #1879, opened 2020-08-06
flag to find curl package without using cmake config

kpedro88:
This PR addresses #1428.

The originally proposed solution, essentially `s/find_package(CURL CONFIG REQUIRED)/find_package(CURL REQUIRED)/`, did not retain the expected default behavior. When running the build according to https://docs.nvidia.com/deeplearning/triton-inference-server/master-user-guide/docs/build.html, CMake found the system curl, even though a newer curl was installed in the build directory.

Instead, the approach of #1846 is followed to introduce a configuration option `TRITON_CURL_WITHOUT_CONFIG`. This option is disabled by default, so the expected build behavior is preserved. This is safer, as the option should only be selected by users who know that they also need to point to the correct files/directories with `-DCURL_LIBRARY` and `-DCURL_INCLUDE_DIR`.

This change was tested in r20.07 and then rebased to master to avoid merge conflicts.

### DIFF
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -62,6 +62,7 @@ set(TRITON_EXTRA_LIB_PATHS "" CACHE PATH "Extra library paths for Triton Server 
 # Client
 set(TRITON_CLIENT_CMAKE_DIR "" CACHE PATH "Path to Triton client library cmake configuration")
 option(TRITON_CLIENT_SKIP_EXAMPLES "Skip building client examples" OFF)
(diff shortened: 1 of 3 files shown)


CoderHam:
@kpedro88  Thank you for your contribution. Could you complete the [CLA ](https://github.com/NVIDIA/triton-inference-server/blob/master/CONTRIBUTING.md)and send it out to triton-cla@nvidia.com, if you have not already done so before?

kpedro88:
@CoderHam I sent the completed CLA to @deadeyegoodwin a while ago.

PR #1883, opened 2020-08-07
Fix Typo in Dockerfile

CoderHam:


deadeyegoodwin:
@CoderHam What is this? There are a lot of changes in here other than Dockerfile.

CoderHam:
I think David Zier already fixed this. The force push he did caused the above diff to be created. In any case since the Tensorflow image name typo is fixed I am closing this.

PR #1887, opened 2020-08-08
Added valgrind CI that checks for unexpected memory leaks

xprotobeast2:
[DLIS-1485] Current acceptable status accounts for known memory leaks such as from NVIDIA/cnmem#12.


qa/L0_valgrind/test.sh
```diff
@@ -0,0 +1,72 @@
...
+
+if [ "$BYTES_LOST" = "$OKAY_LOST_BYTES" ]; then
+    echo -e "\n***\n*** Test PASSED\n***"
+else
+    sed -n '/are definitely lost/,/bytes/p' $TEST_LOG
+    echo -e "\n***\n*** Test FAILED\n***"
+    RET=1
+fi
+
+exit $RET
```
GuanLuo, line 72: Add newline

qa/common/util.sh
```diff
@@ -234,6 +234,46 @@ function run_server_nowait () {
     SERVER_PID=$!
 }
 
+# Run inference server inside a memory management tool like Valgrind/ASAN. 
+# Return once server's health endpoint shows ready or timeout expires. Sets 
+# SERVER_PID to pid of SERVER, or 0 if error (including expired timeout)
+function run_server_leakcheck () {
```
GuanLuo, line 240: You may want to make this utility for only running the server, like `run_server`, the caller should kill the server explicitly. Because we may want to collect valgrind result of different server states (i.e. running n inferences, or invoke all endpoints...)

Dockerfile.QA
```diff
@@ -304,6 +304,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
                               protobuf-compiler && \
     rm -rf /var/lib/apt/lists/*
 
+# CI/QA for memcheck requires valgrind
+RUN apt-get update && apt-get install -y valgrind
```
deadeyegoodwin, line 308: Just merge with the above apt-get. You can add your comment to the existing one.

qa/L0_valgrind/test.sh
```diff
@@ -0,0 +1,75 @@
...
+
+DATADIR=/data/inferenceserver/${REPO_VERSION}/tftrt_model_store
+
+SERVER=/opt/tritonserver/bin/tritonserver
+SERVER_ARGS="--model-repository=$DATADIR"
+SERVER_LOG="./inference_server.log"
+
+LEAKCHECK=/usr/bin/valgrind
+LEAKCHECK_ARGS="--leak-check=full --show-possibly-lost=no --log-file=$TEST_LOG" 
+OKAY_LOST_BYTES="80"
```
deadeyegoodwin, line 48: Can you add a comment and a cut-paste showing what the 80 expected lost bytes are.

CoderHam: Additionally this will vary based on the model (backend) you are using or if the instance is being run on CPU/GPU. You should run some experiments to confirm the same. I would prefer if you grep the logs for definitely lost bytes and ignore sections that have `cmem` and `Tensorflow::NewSession`.

PR #1892, opened 2020-08-10
Fix clientsdk to include protobuf and grpc headers

Tabrizian:


src/clients/c++/library/CMakeLists.txt
```diff
@@ -187,6 +187,14 @@ if(${TRITON_ENABLE_GRPC})
       ${CMAKE_CURRENT_BINARY_DIR}/../../../core/grpc_service.pb.h
       DESTINATION include
   )
+
+  install(
+      DIRECTORY
+      ${CMAKE_CURRENT_BINARY_DIR}/../../../../../protobuf/include/
```
deadeyegoodwin, line 193: Is there a single protobuf and grpc include file, or a lot of them? If there are many it would be better to install into a protobuf/ and grpc/ subdirectory respectively.

Tabrizian: They are already inside folders. Here is how the directory structure looks like:
```
root@bb4318d74453:/workspace/builddir/client/install/include# tree -L 1  
.
|-- benchmark
|-- common.h
|-- google
|-- grpc
|-- grpc++
|-- grpc_client.h
|-- grpc_service.grpc.pb.h
|-- grpc_service.pb.h
|-- grpcpp
|-- http_client.h
|-- ipc.h
`-- model_config.pb.h
```

PR #1897, opened 2020-08-11
Advance GRPC to 1.25.0

tanmayv25:
Need to advance the grpc version to 1.25.0 as [`rules_proto_grpc`](https://github.com/rules-proto-grpc/rules_proto_grpc), used for generating `*.pb.cc` and `*pb.h` files inside bazel build system doesn't have a release supporting GRPC 1.24.0.


deadeyegoodwin:
Why do we care about bazel?

tanmayv25:
> Why do we care about bazel?

It is related to DLIS-1508. Some users might want to compile their own grpc client library from our proto files, (may or may not be bazel). This change helps them to use one such popular tool with most likely not much cost on our part.

PR #1898, opened 2020-08-12
Add batch input to provide defined set of batch-wise inputs

GuanLuo:
Created PR to showcase what the model config will look like


src/core/model_config.proto
```diff
@@ -713,6 +713,85 @@ message ModelQueuePolicy
 //@@
 message ModelDynamicBatching
 {
+  //@@  .. cpp:var:: message BatchInput
+  //@@
+  //@@     A batch input is an additional input for the requests in the batch
```
deadeyegoodwin, line 718: "A batch input is an additional input that must be added by the backend based on all the requests in a batch."

Let's not tie this specifically to ragged input as we may find other uses for it in the future.

src/core/model_config.proto
```diff
@@ -713,6 +713,85 @@ message ModelQueuePolicy
...
+
+    //@@    .. cpp:var:: Kind kind
+    //@@
+    //@@       The kind of this batch input.
+    //@@
+    Kind kind = 1;
+
+    //@@    .. cpp:var:: string name
+    //@@
+    //@@       The name of the model input.
```
deadeyegoodwin, line 776: "... of the model input that the backend will create for this batch input."

src/core/model_config.proto
```diff
@@ -713,6 +713,85 @@ message ModelQueuePolicy
...
+    //@@    .. cpp:var:: DataType data_type
+    //@@
+    //@@       The input's datatype. The data type can be TYPE_INT32 or
+    //@@       TYPE_FP32.
+    //@@
+    DataType data_type = 3;
+
+    //@@    .. cpp:var:: string target_input
+    //@@
+    //@@       The name of the model input that are used to populate
```
deadeyegoodwin, line 789: "The backend derives the value for each batch input from one or more other inputs. 'target_inputs' gives the names of those inputs."

src/core/model_config.proto
```diff
@@ -713,6 +713,85 @@ message ModelQueuePolicy
...
+    //@@       batch input value.
+    //@@
+    string target_input = 4;
```
deadeyegoodwin, line 792: Change this to "repeated string" so that in the future we can support a batch input that must derive from more than one input.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -1794,92 +1847,317 @@ PlanBackend::Context::Run(
...
 
-    const InferenceRequest::Input* repr_input;
-    FAIL_ALL_AND_RETURN_IF_ERROR(
-        payload_->requests_, payload_->responses_, metric_reporter_.get(),
-        repr_input_request->ImmutableInput(name, &repr_input),
-        "failed to obtain the input '" + name + "'");
-    // Get the shape of the input. The request has already checked
-    // that the request shape is valid so don't need to do it here.
-    const auto& batch1_shape = repr_input->Shape();
+    if (buffer_is_ragged_[bindex]) {
```
deadeyegoodwin, line 1850: Can you pull out some of the large inline code blocks into separate functions so that this function doesn't become unreadable.

qa/L0_trt_batch_input/models/ragged_acc_shape/config.pbtxt
```diff
@@ -0,0 +1,75 @@
...
+
+name: "ragged_acc_shape"
+platform: "tensorrt_plan"
+max_batch_size: 4
+input [
+  {
+    name: "RAGGED_INPUT"
+    data_type: TYPE_FP32
+    dims: [ -1 ]
+    allow_ragged_batch: true
```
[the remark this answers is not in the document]
GuanLuo: I don't think that a input is 'target_input' in batch_input implies that the input needs to be ragged-batched, the input can still be batched regularly if desirable.

src/backends/tensorrt/plan_backend.cc
GuanLuo: I think the processing code can be moved into InputCollector and it returns the shape for the batch input and set the value into the provided buffer, because the processing for different batch input kind should be the same across all backends, the differences may just be how to pass the shape and data to the models.

src/core/model_config.proto
```diff
@@ -713,6 +713,85 @@ message ModelQueuePolicy
...
+      //@@
+      BATCH_ACCUMULATED_ELEMENT_COUNT_WITH_ZERO = 2;
+
+      //@@      .. cpp:enumerator:: Kind::BATCH_MAX_ELEMENT_COUNT_AS_SHAPE = 3
+      //@@
+      //@@         Among the requests in the batch, the max element count of the
+      //@@         'target_input' will be added as input with shape
+      //@@         [max_element_count] for the first request in the batch.
+      //@@         For other requests, such input will be with shape [0].
+      //@@         The data of the tensor will be dummy data.
```
deadeyegoodwin, line 763: When we create this tensor we must provide some data buffer. Is is zero-filled or uninitialized?

qa/L0_trt_batch_input/models/ragged_acc_shape/config.pbtxt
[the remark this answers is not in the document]
deadeyegoodwin: Good point. Just leave it as is.

src/backends/tensorrt/plan_backend.cc
deadeyegoodwin: If the code is general then moving most/all of it into the collector would be great.

src/core/model_config.proto
GuanLuo: It is uninitialized currently. Plan backend just allocated a large enough buffer for this input on model creation.

deadeyegoodwin: Can you change the comment to say "The data of the tensor will be uninitialized".

GuanLuo:
I moved BatchInput out from ModelDynamicBatching as now it is being handled within the model backend and not tied with any batcher.

PR #1903, opened 2020-08-14
Add utils.py for Python models

Tabrizian:
Add utils.py for Python models.

### DIFF
--- a/src/backends/backend/python/CMakeLists.txt
+++ b/src/backends/backend/python/CMakeLists.txt
@@ -125,3 +125,5 @@ string(REPLACE "../../../../" "../../../../../../../../../"  PROTO_FIX_PY ${PROT
 
 install(FILES ${protocol_core_py} ${PROTO_FIX_PY} ${protocol_grpc_py} resources/startup.py
        DESTINATION lib/python/runtime)
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/resources/utils.py
+       DESTINATION lib/python/runtime)
--- a/src/backends/backend/python/CMakeLists.txt
+++ b/src/backends/backend/python/CMakeLists.txt
@@ -39,7 +39,7 @@ set(gRPC_CPP_PLUGIN_EXECUTABLE $<TARGET_FILE:gRPC::grpc_cpp_plugin>)
 file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/python_host.proto DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
 set(Protobuf_SRC_ROOT_FOLDER ${CMAKE_CURRENT_BINARY_DIR})
 
-protobuf_generate_cpp(PROTO_SRCS PROTO_HOST_HDRS ${CMAKE_CURRENT_BINARY_DIR}/python_host.proto)
+protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS ${CMAKE_CURRENT_BINARY_DIR}/python_host.proto)
 
 protobuf_generate_python(PROTO_PY ${CMAKE_CURRENT_BINARY_DIR}/python_host.proto)
 
@@ -71,10 +71,14 @@ add_custom_target(
         python_libs ALL 
         DEPENDS ${PROTO_PY} ${protocol_grpc_py})
 
-FILE(GLOB PROTOBUF_HEADERS ${CMAKE_CURRENT_BINARY_DIR}/*.cc)
-FILE(GLOB PROTOBUF_SRCS ${CMAKE_CURRENT_BINARY_DIR}/*.h)
-
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
+# TODO: Fix Protobuf paths
+string(REPLACE "../../../../" "../../../../../../../../../"  PROTOBUF_HEADERS ${PROTO_HDRS})
+string(REPLACE "../../../../" "../../../../../../../../../"  PROTOBUF_SRCS ${PROTO_SRCS})
+
+set_source_files_properties(${PROTOBUF_HEADERS} PROPERTIES GENERATED TRUE)
+set_source_files_properties(${PROTOBUF_SRCS} PROPERTIES GENERATED TRUE)
 
 add_library(
   triton_python SHARED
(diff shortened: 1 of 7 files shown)


src/backends/backend/python/CMakeLists.txt
```diff
@@ -125,3 +129,5 @@ string(REPLACE "../../../../" "../../../../../../../../../"  PROTO_FIX_PY ${PROT
 
 install(FILES ${protocol_core_py} ${PROTO_FIX_PY} ${protocol_grpc_py} resources/startup.py
        DESTINATION lib/python/runtime)
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/resources/utils.py
```
deadeyegoodwin, line 132: utils.py is too generic. Call it triton_python_backend_utils.py.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -744,9 +749,22 @@ TRITONBACKEND_ModelInstanceExecute(
   ni::ExecuteResponse execute_response;
 
   // Perform inference on the Python side
-  const auto err = instance_state->stub->Execute(
+  const auto status = instance_state->stub->Execute(
```
deadeyegoodwin, line 752: What if there were errors in some of the requests and so you have already sent error responses for those? How does the python code know about that? The lifecycle of requests and responses is important and is documented in the Backend section of the docs and in tritonbackend.h.
We should discuss the overall handling of this lifecycle as what is here now seems to be incorrect.

src/backends/backend/python/python_interpreter.cc
```diff
@@ -744,9 +749,22 @@ TRITONBACKEND_ModelInstanceExecute(
...
       &context, execute_request, &execute_response);
 
+  if (!status.ok()) {
+    return TRITONSERVER_ErrorNew(
```
deadeyegoodwin, line 756: Lifecycle. In general once you have created and returned any response you can't return an error from this function. You must handle all response sending yourself.

src/backends/backend/python/resources/startup.py
```diff
@@ -118,49 +117,77 @@ def __init__(self, module_path, *args, **kwargs):
...
 
     def Init(self, request, context):
+        """Init is called on TRITONBACKEND_ModelInstanceInitialize. `request`
+        object contains an args which includes a `model_config` key
+        containing the model configuration. This paramter is passed by
+        default to every ModelInstance.
+        """
         # TODO: Add error handling and returning correct status codes
         args = {x.key: x.value for x in request.args}
         self.model = self.initializer_func(args)
```
deadeyegoodwin, line 127: self.model isn't a good name. self.backend would be better.

src/backends/backend/python/resources/utils.py
```diff
@@ -0,0 +1,160 @@
...
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import numpy as np
+
+
+class InferenceRequest:
+    """An InfrenceRequest object is used to represent Protobufs coming from
```
deadeyegoodwin, line 31: Don't say "protobuf", that is an implementation detail that is hidden from user.

deadeyegoodwin: InferenceRequest represents a request for inference for a model that executes using this backend.

src/backends/backend/python/resources/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    and simultaneously calling different methods with different threads is
+    not supported and will cause undefined behavior.
+
+    Parameters
+    ----------
+    inputs : list
+        A list of Tensor objects, each describing data for an input tensor
+        required by the model
+    request_id : str
+        Optional identifier for the request. If specified will be returned
```
deadeyegoodwin, line 42: ID assoiciated with this request, or empty string if no ID is associated with the request.

src/backends/backend/python/resources/utils.py
```diff
@@ -0,0 +1,160 @@
...
+        in the response. Default value is an empty string which means no
+        request_id will be used.
+    correlation_id : str
```
deadeyegoodwin, line 45: Correlation ID associated with this request, or empty string if no correlation ID is associated with the request.

src/backends/backend/python/resources/utils.py
```diff
@@ -0,0 +1,160 @@
...
+        Optional identifier for the request. If specified will be returned
+        in the response. Default value is an empty string which means no
+        request_id will be used.
+    requested_output_name : list
+        A list of strings, each describing the requested output.
```
deadeyegoodwin, line 50: The names of the output tensors that should be calculated and returned for this request.

src/backends/backend/python/resources/utils.py
```diff
@@ -0,0 +1,160 @@
...
+        list
+            A list of Tensor objects
+        """
+        return self._output_tensors
+
+
+class Tensor:
+    """A Tensor object is used to represent inputs and output data for an
+    InferenceRequest or InferenceResponse. The object is intended to be used
+    by a single thread and simultaneously calling different methods with
```
deadeyegoodwin, line 126: These thread comments are not really correct. These are trivial classes so I don't think  you need to say anything about threading.

src/backends/backend/python/python_interpreter.cc
Tabrizian: I release all the requests if an error happens during inference.

src/backends/backend/python/resources/triton_python_backend_utils.py
```diff
@@ -0,0 +1,155 @@
...
+        -------
+        list
+            A list of Tensor objects
+        """
+        return self._output_tensors
+
+
+class Tensor:
+    """A Tensor object is used to represent inputs and output data for an
+    InferenceRequest or InferenceResponse. The object is intended to be used
```
deadeyegoodwin, line 120: Remove the thread part of the comment

PR #1909, opened 2020-08-19
Use worker threads to run CopyBuffer in parallel 

CoderHam:
- [x] Use global `AsyncWorkQueue` to make parallel calls to `CopyBuffer` or any other such functions. 
- [x] Add command line argument (`--async-worker-count`) for number of worker threads in AsyncWorkQueue. When set to 1 or lower then do not create threads. Add checks manually in this case to remove `AddTask` calls. 
- [x] Limit parallel CopyBuffer calls to CPU and not GPU copies 
- [ ] Investigate effects on performance.


src/core/thread_pool.h
```diff
@@ -0,0 +1,110 @@
...
+  Status AddTask(
+      const std::string& msg, const TRITONSERVER_MemoryType src_memory_type,
+      const int64_t src_memory_type_id,
+      const TRITONSERVER_MemoryType dst_memory_type,
+      const int64_t dst_memory_type_id, const size_t byte_size, const void* src,
+      void* dst, cudaStream_t cuda_stream, bool* cuda_used);
+
+  // Run CopyBuffer on the worker threads. Must add to queue using AddTask
+  // before calling ProcessQueue.
+  Status ProcessQueue();
```
GuanLuo, line 68: Can you illustrate the intended workflow of using this thread pool class? I don't think the public functions except `AddTask()` are necessary.
Also who is responsible for calling `ProcessQueue()`? It simply tells worker threads to consume the tasks in the queue, why don't you just do the same when invoking `AddTask()`

src/core/thread_pool.h
```diff
@@ -0,0 +1,110 @@
...
+  // Wait for all pending worker threads to finish.
+  Status AwaitCompletion();
+
+  // Get Id of next available worker thread. If all workers are occupied then
+  // wait for them to finish and start from the beginning.
+  Status GetNextAvailableId(int* worker_id);
+
+  // Add CopyBuffer task to queue
+  Status AddTask(
+      const std::string& msg, const TRITONSERVER_MemoryType src_memory_type,
```
GuanLuo, line 60: You should either change the class name to be specific for data copy or, for generality, wrap all the arguments and the `CopyHandler()` below into a lambda that is ready to be called without arguments. You probably need this class to be template class for latter

src/core/thread_pool.cc
```diff
@@ -0,0 +1,115 @@
...
+Status
+WorkerThreadPool::ProcessQueue()
+{
+  while (!queue_.Empty()) {
+    auto task_data = std::move(queue_.Get());
+    int worker_id;
+    RETURN_IF_ERROR(GetNextAvailableId(&worker_id));
+    futures_[worker_id] = task_data->status_.get_future();
+    worker_threads_[worker_id] =
+        std::thread(CopyBufferHandler, task_data.get());
```
GuanLuo, line 107: Don't create new thread for each task

src/core/thread_pool.cc
```diff
@@ -0,0 +1,115 @@
...
+WorkerThreadPool::GetNextAvailableId(int* worker_id)
+{
+  *worker_id = -1;
+  for (size_t i = 0; i < worker_threads_.size(); i++) {
+    if (!worker_threads_[i].joinable()) {
+      *worker_id = i;
+    }
+  }
+  if (*worker_id == -1) {
+    RETURN_IF_ERROR(AwaitCompletion());
```
GuanLuo, line 69: If this condition is true, you are returning invalid index for `ProcessQueue()` to use.

src/core/thread_pool.h
```diff
@@ -0,0 +1,110 @@
...
+    std::promise<Status> status_;
+  };
+
+  // Helper around CopyBuffer
+  static void CopyBufferHandler(CopyBufferData* data);
+
+  int thread_count_;
+  std::vector<std::thread> worker_threads_;
+  std::vector<std::future<Status>> futures_;
+  std::condition_variable cv_;
```
GuanLuo, line 106: What is this `cv_` for?

src/core/thread_pool.h
GuanLuo: I think you can just re-purpose this function to be waiting for all tasks are completed? And then once this function returns, the state within the class object is reset for next set of tasks.

CoderHam: AddTask will add the CopyBuffer task to the queue. 
To execute the tasks we must call ProcessQueue. The ProcessQueue returns the status (success or failure + failure msg) once all the tasks are completed.
Are you suggested we start processing the task as soon as it is added to the queue instead of waiting for all the tasks to be added to the queue?

CoderHam: If the number of tasks > number of worker threads, we may have to wait at AddTask for a thread to be available so that it can trigger the next CopyBuffer op.

src/core/thread_pool.h
CoderHam: Yes I think it is better to maintain it as a template so that we can do this for other operations that can benefit from being run in parallel.

src/core/thread_pool.cc
CoderHam: Yes we should return 0 here after waiting for the threads to complete

src/core/thread_pool.h
CoderHam: Remnant of earlier changes. Removed since not using anymore

src/core/thread_pool.h
CoderHam: On second thought aggregating the status of CopyBuffer means that the class cannot be generic. I will fix the name to avoid confusion.

GuanLuo: Why it can't be generic? We can just formulate that the Status object is used to signal whether the task is performed successfully. This class doesn't care about the detail of the Status object anyway, it is left for the caller to interpret the Status object.

src/core/thread_pool.h
GuanLuo: > Are you suggested we start processing the task as soon as it is added to the queue instead of waiting for all the tasks to be added to the queue?

Yes, I don't see a reason for worker thread to wait until it is told to consume the queue, which introduce unnecessary wait, it can just actively wait for task inside the queue.

> If the number of tasks > number of worker threads, we may have to wait at AddTask for a thread to be available so that it can trigger the next CopyBuffer op.

Please take a look at dynamic batcher or plan backend for how sync_queue is being used. The producer doesn't need to worry about the consumer, it just places tasks into the queue.

src/core/backend_context.cc
```diff
@@ -731,26 +733,27 @@ BackendInputCollector::SetFixedSizeInputTensor(
...
-              status),
-          "error setting TensorFlow input tensor");
-      return cuda_copy;
-    }
+        &cuda_used[idx]);
 
     input_offset += src_byte_size;
   }
 
+  Status status = worker_pool_.ProcessQueue();
```
GuanLuo, line 745: If you re-purpose `ProcessQueue()` to be waiting for all tasks to be finished as suggested, you can defer the call to this as late as possible to reduce the number of "consume tasks -> wait for completion" cycles.

src/core/thread_pool.h
CoderHam: Fixed.

src/core/thread_pool.h
```diff
@@ -0,0 +1,126 @@
...
+      GetSingleton()->promises_.Put(std::move(promise));
+    } else {
+      GetSingleton()->futures_[worker_id] = promise.get_future();
+      GetSingleton()->worker_threads_[worker_id].reset(&task_data);
+    }
+    return Status::Success;
+  }
+
+  // Run task threads remaining in queue on the worker threads.
+  static Status CompleteQueue()
```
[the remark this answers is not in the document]
deadeyegoodwin: The thread pool itself doesn't know anything about cuda_copy or what is being executed. The completio queue (which is implemented entirely by the caller and within the lambda being executed) can return the cuda_copy value.

src/core/CMakeLists.txt
```diff
@@ -137,6 +137,7 @@ set(
   backend.cc
   backend_context.cc
   cuda_utils.cc
+  thread_pool.cc
```
deadeyegoodwin, line 140: Change filename to match class functionality. async_work_queue.cc

src/core/backend_context.h
```diff
@@ -34,6 +34,7 @@
 #include "src/core/memory.h"
 #include "src/core/model_config.h"
 #include "src/core/status.h"
+#include "src/core/thread_pool.h"
```
deadeyegoodwin, line 37: Don't need this?

src/core/cuda_utils.h
```diff
@@ -25,13 +25,16 @@
...
 
+#include <future>
 #include <set>
 #include "src/core/status.h"
 
 #ifdef TRITON_ENABLE_GPU
 #include <cuda_runtime_api.h>
 #endif  // TRITON_ENABLE_GPU
 
+#include "src/core/sync_queue.h"
```
deadeyegoodwin, line 36: Why is this include down here and not right after src/core/status.h?

src/core/cuda_utils.h
```diff
@@ -25,13 +25,16 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #pragma once
 
+#include <future>
```
deadeyegoodwin, line 28: What is this include for?

src/core/thread_pool.h
```diff
@@ -0,0 +1,65 @@
...
+#include <mutex>
+#include <thread>
+#include <vector>
+
+#include "src/core/status.h"
+#include "src/core/sync_queue.h"
+
+namespace nvidia { namespace inferenceserver {
+
+class AsyncWorkQueue {
```
deadeyegoodwin, line 40: Need a description for what this class is.

src/core/thread_pool.h
```diff
@@ -0,0 +1,65 @@
...
+ public:
+  // Should only be called once to set the number of worker threads.
+  static void SetWorkerCount(size_t thread_count);
```
deadeyegoodwin, line 43: Instead of set and get should just have a static Initialize(size_t thread_count) that is called once to start the threads. This will be called early when the server is initialized.

src/core/thread_pool.h
```diff
@@ -0,0 +1,65 @@
...
+  // Add a task to the queue.
+  static void AddTask(const std::function<void(void)> task);
+
+ private:
+  AsyncWorkQueue() = default;
+  ~AsyncWorkQueue();
+  static AsyncWorkQueue* GetSingleton();
+  static void Initialize();
+
+  size_t thread_count_;
```
deadeyegoodwin, line 57: Should need this, it is the length of the worker_thread_.

src/core/thread_pool.cc
```diff
@@ -0,0 +1,107 @@
...
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#include <condition_variable>
+#include <deque>
+#include <future>
+#include <mutex>
+#include <thread>
+#include <vector>
+
+#include "src/core/thread_pool.h"
```
deadeyegoodwin, line 33: Coding standard has this as first include, followed by blank line, followed by other includes. Please follow the standard used in other files.

src/core/thread_pool.cc
```diff
@@ -0,0 +1,107 @@
...
+size_t
+AsyncWorkQueue::GetWorkerCount()
+{
+  return GetSingleton()->thread_count_;
+}
+
+void
+AsyncWorkQueue::AddTask(const std::function<void(void)> task)
+{
+  std::lock_guard<std::mutex> lock(GetSingleton()->mutex_);
```
deadeyegoodwin, line 81: You don't need an additional lock, the sync queue takes care of that. Consult sync_queue.h, it is a very simple class.

src/core/thread_pool.cc
```diff
@@ -0,0 +1,107 @@
...
+  GetSingleton()->task_queue_.Put(task);
+  GetSingleton()->queue_pending.notify_one();
```
deadeyegoodwin, line 83: You don't need your own notification, the sync queue does that.

src/core/thread_pool.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+void
+AsyncWorkQueue::Initialize()
+{
+  std::function<void(void)> task;
+
+  while (true) {
```
deadeyegoodwin, line 91: This is unnecessarily complex. Just have do { task = Get(); if (task == nullptr) break; task(); }

src/core/thread_pool.cc
```diff
@@ -0,0 +1,107 @@
...
+
+AsyncWorkQueue::~AsyncWorkQueue()
+{
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    exit_ = true;
+    queue_pending.notify_one();
+  }
+
+  for (auto& worker_thread : worker_threads_) {
```
deadeyegoodwin, line 45: Just push 'num_threads' nullptrs onto the task queue and then the joins.

src/servers/main.cc
```diff
@@ -324,6 +325,9 @@ std::vector<Option> options_
       {OPTION_BACKEND_DIR, "backend-directory", Option::ArgStr,
        "The global directory searched for backend shared libraries. Default is "
        "'/opt/tritonserver/backends'."},
+      {OPTION_ASYNC_WORKER_COUNT, "async-worker-count", Option::ArgInt,
```
deadeyegoodwin, line 328: Right now we are using a singleton async work class just for buffer copy acceleration but in the future we may have more than one worker pool for different uses. So let's make the flag specific to the current use case. Use --buffer-manager-thread-count and a description like. "The number of threads used to accelerate copies and other operations required to manage input and output tensor contents. Default is 1."  We can correctly handle 0 value but it is confusing to make that the default since it implies no threads are doing copies.

Also adjust naming for related API additions.

src/core/backend_context.cc
```diff
@@ -291,24 +295,55 @@ BackendResponder::Finalize()
...
+                    status),
+                "error sending TensorFlow response");
+          }
+        }
+      }
+
+      offset += response_byte_size;
+    }
+
+    for (size_t i = 0; i < cuda_used_vec.size(); i++) {
```
deadeyegoodwin, line 329: You can't do it this way because you have no way of knowing that the task have yet set cuda_used_vec. Put cuda_used values in the completion queue with the status as we discussed.

src/core/backend_context.cc
```diff
@@ -270,7 +272,9 @@ BackendResponder::Finalize()
     char* pinned_buffer = def.pinned_memory_->MutableBuffer(
         &pinned_memory_type, &pinned_memory_id);
 
-    size_t offset = 0;
+    std::deque<bool> cuda_used_vec(def.responses_.size(), false);
+    SyncQueue<Status> completion_queue;
+    size_t offset = 0, count = 0;
```
deadeyegoodwin, line 277: async_task_count

src/core/backend_context.cc
```diff
@@ -291,24 +295,55 @@ BackendResponder::Finalize()
...
+
+      offset += response_byte_size;
+    }
+
+    for (size_t i = 0; i < cuda_used_vec.size(); i++) {
+      need_sync_ |= cuda_used_vec[i];
+    }
+
+    size_t index = 0;
+    for (auto& pr : def.responses_) {
```
deadeyegoodwin, line 334: You just need to Get from the completion queue 'async_task_count' times and aggregate the status and cuda_used. You can't send response inside the loop as that could result in sending multiple responses if there are multiple errors.

src/core/cuda_utils.h
CoderHam: Unnecessary. Removed.

src/core/cuda_utils.h
CoderHam: Fixed

src/core/backend_context.cc
CoderHam: Ok. I will rework to return status once. But we should send the response for the first case that fails and break of the loop out right?

src/core/backend_context.cc
CoderHam: Fixed

src/core/thread_pool.h
CoderHam: We need get to change the logic for the CopyBuffer when thread count is 1 since that would not correspond to a benefit in perf.

src/core/thread_pool.cc
CoderHam: Fixed

src/core/backend_context.cc
deadeyegoodwin: No, you can't break out of the loop because then you would leave completed task results in the queue. You need to remember on of the failed statuses and then outside the loop return that status.

src/core/backend_context.cc
CoderHam: Done

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,87 @@
...
+size_t
+AsyncWorkQueue::GetWorkerCount()
+{
+  return GetSingleton()->worker_threads_.size();
+}
+
+void
+AsyncWorkQueue::AddTask(const std::function<void(void)> task)
+{
+  GetSingleton()->task_queue_.Put(task);
```
deadeyegoodwin, line 72: std::move(task)

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,87 @@
...
+}
+
+size_t
+AsyncWorkQueue::GetWorkerCount()
+{
+  return GetSingleton()->worker_threads_.size();
+}
+
+void
+AsyncWorkQueue::AddTask(const std::function<void(void)> task)
```
deadeyegoodwin, line 70: std::function is a class and there is a cost to passing it by value. Use "std::function<void(void)>&& task" to indicate that the function is taking ownership of 'task' and that std::move should be used when calling AddTask().

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,87 @@
...
+{
+  GetSingleton()->task_queue_.Put(task);
+}
+
+void
+AsyncWorkQueue::InitializeThread()
+{
+  std::function<void(void)> task;
```
deadeyegoodwin, line 78: Don't declare this separately, just put on line 80.

src/core/tritonserver.cc
```diff
@@ -213,6 +215,9 @@ class TritonServerOptions {
   unsigned int ExitTimeout() const { return exit_timeout_; }
   void SetExitTimeout(unsigned int t) { exit_timeout_ = t; }
 
+  unsigned int BufferManagerThreadCount() const { return async_worker_count_; }
```
deadeyegoodwin, line 218: Why use async_worker_count_ when the value is the buffer manager thread count?

src/core/tritonserver.cc
```diff
@@ -1523,6 +1539,7 @@ TRITONSERVER_ServerNew(
   lserver->SetStrictReadinessEnabled(loptions->StrictReadiness());
   lserver->SetExitTimeoutSeconds(loptions->ExitTimeout());
   lserver->SetBackendCmdlineConfig(loptions->BackendCmdlineConfigMap());
+  ni::AsyncWorkQueue::Initialize(loptions->BufferManagerThreadCount());
```
deadeyegoodwin, line 1542: This should be in server.cc::Init function. At the beginning just before the pinned memory manager.

src/servers/main.cc
```diff
@@ -807,6 +812,7 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
   int32_t exit_timeout_secs = 30;
   int32_t repository_poll_secs = repository_poll_secs_;
   int64_t pinned_memory_pool_byte_size = 1 << 28;
+  int32_t buffer_manager_thread_count = 0;
```
deadeyegoodwin, line 815: Default is 1. Don't confuse the argument value with what you use to initialize the async work queue.

src/core/backend_context.cc
```diff
@@ -291,24 +295,56 @@ BackendResponder::Finalize()
...
       } else {
-        bool cuda_used = false;
-        status = CopyBuffer(
-            response_output->Name(), pinned_memory_type, pinned_memory_id,
-            response_memory_type, response_memory_type_id, response_byte_size,
-            pinned_buffer + offset, const_cast<void*>(response_buffer), stream_,
-            &cuda_used);
-        need_sync_ |= cuda_used;
+        // Only use parallel CopyBuffer for CPU-CPU copy when worker count > 1
+        if ((AsyncWorkQueue::GetWorkerCount() > 1) &&
```
deadeyegoodwin, line 299: You should have a "bool use_async_cpu_copy" that is set when the class is constructed.

src/core/backend_context.cc
```diff
@@ -291,24 +295,56 @@ BackendResponder::Finalize()
...
+            (response_memory_type == TRITONSERVER_MEMORY_CPU) &&
+            (pinned_memory_type == TRITONSERVER_MEMORY_CPU)) {
```
deadeyegoodwin, line 301: This function only does pinned->cpu and cpu->cpu copies, so how is this conditional ever going to be true?

src/core/backend_context.cc
```diff
@@ -270,7 +272,9 @@ BackendResponder::Finalize()
     char* pinned_buffer = def.pinned_memory_->MutableBuffer(
         &pinned_memory_type, &pinned_memory_id);
 
-    size_t offset = 0;
+    SyncQueue<std::pair<Status, bool>> completion_queue;
+    std::vector<size_t> async_task_ids;
```
deadeyegoodwin, line 276: I don't understand what async_task_ids is for. "count" should be named "async_copy_count" and I don't see why that isn't enough.

src/core/backend_context.cc
```diff
@@ -291,24 +295,56 @@ BackendResponder::Finalize()
...
+        }
+      }
+
+      offset += response_byte_size;
+      count++;
+    }
+
+    size_t index = 0;
+    for (auto& pr : def.responses_) {
+      if (std::find(async_task_ids.begin(), async_task_ids.end(), index) !=
```
deadeyegoodwin, line 333: Is this to determine if the response had an async copy sent. It is very confusingly named and implemented. Can you instead just send the "std::unique_ptr<InferenceResponse>* response" along with the task and include it in the completion queue. That will be a lot cleaner and simpler to understand.

src/core/backend_context.cc
CoderHam: Do you mean as a member of the BackendInputCollector and BackendResponder class?

src/core/backend_context.cc
CoderHam: Yes. I could do that instead. I can get rid of async_task_ids entirely then.

src/core/backend_context.cc
CoderHam: The count was for all counts and not just async_copy counts. This will be removed when response is sent as a param to the task

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,87 @@
...
+{
+  static AsyncWorkQueue singleton;
+  return &singleton;
+}
+
+void
+AsyncWorkQueue::Initialize(size_t thread_count)
+{
+  if (thread_count > 1) {
+    for (size_t id = 0; id < thread_count; id++) {
```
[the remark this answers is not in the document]
CoderHam: This simply iterates a range

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,87 @@
...
+      GetSingleton()->worker_threads_.push_back(std::unique_ptr<std::thread>(
+          new std::thread([] { InitializeThread(); })));
```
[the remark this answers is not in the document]
CoderHam: The lambda is still necessary.

src/core/backend_context.cc
CoderHam: This check needs to be fixed everywhere. The AsyncWorkQueue should work for both CPU_PINNED and CPU.

src/core/backend_context.cc
CoderHam: We can't generalize to add response in all cases the because there is a case a CopyBuffer call corresponds to reporting errors for multiple requests.

src/servers/main.cc
```diff
@@ -324,6 +325,10 @@ std::vector<Option> options_
       {OPTION_BACKEND_DIR, "backend-directory", Option::ArgStr,
        "The global directory searched for backend shared libraries. Default is "
        "'/opt/tritonserver/backends'."},
+      {OPTION_BUFFER_MANAGER_THREAD_COUNT, "buffer-manager-thread-count",
+       Option::ArgInt,
+       "The number of threads used to accelerate copies and other operations "
+       "required to manage input and output tensor contents. Default is 1."},
```
deadeyegoodwin, line 331: I was incorrect before. The default here should be zero so that we don't create any worker threads unless the user explicitly requests it.

CoderHam: That's right. Although CopyBuffer would not be triggered we should not have these threads created by default. Fixed!

src/core/backend_context.cc
deadeyegoodwin: This is still convoluted. It doesn't matter that you have to do something different elsewhere in the code. Here you should do the correct and clear thing and send the response along with the task and get it back in the completion queue. Also, it seems like you are assuming that the tasks complete in the same order as they are queued (and in the def.responses_ order). That is not necessarily true and will be fixed by passing the response.

src/core/backend_context.cc
```diff
@@ -478,7 +513,9 @@ BackendResponder::FlushPendingPinned(
   if (pinned_memory_type != TRITONSERVER_MEMORY_CPU_PINNED) {
```
deadeyegoodwin, line 513: In this case here we didn't allocate a pinned buffer so we are doing CPU->GPU or GPU->CPU direct, and so we don't want any of the async copy changes below.

src/core/backend_context.cc
```diff
@@ -553,7 +624,9 @@ BackendResponder::FlushPendingPinned(
     // response outputs so that we can do the pinned->CPU copies in
```
deadeyegoodwin, line 624: This is in the else clause where a pinned buffer was used. The CopyBuffer above is the one that may be CPU->PINNED and so is the one that needs async if doing CPU->PINNED. If the above did GPU->PINNED then the copy below will always involve GPU and so doesn't need async.

src/core/backend_context.cc
```diff
@@ -839,34 +945,78 @@ BackendInputCollector::Finalize()
 
   // After the above sync all the GPU->pinned copies are complete. Any
   // deferred copies of pinned->CPU can now be done.
+
```
deadeyegoodwin, line 948: InputCollector is (mostly) similar to the responder so similar comments to above apply

src/core/backend_context.cc
```diff
@@ -946,26 +1015,55 @@ BackendInputCollector::SetFixedSizeInputTensor(
     }
 
     // Direct copy without intermediate pinned memory.
```
deadeyegoodwin, line 1017: Isn't this case the same discussion we already had where we don't want to async (for now) when it is direct GPU<->CPU copy that is not using a pinned buffer?

deadeyegoodwin (changes requested):

src/core/backend_context.cc
```diff
@@ -839,30 +868,67 @@ BackendInputCollector::Finalize()
...
-        def.tensor_buffer_ + def.tensor_buffer_offset_, stream_, &cuda_used);
-    need_sync_ |= cuda_used;
+    // Only use parallel CopyBuffer for CPU-CPU copy when worker count > 1
+    if (use_async_cpu_copy_) {
+      AsyncWorkQueue::AddTask(std::move(std::bind(
+          CopyBufferHandler, "pinned buffer", pinned_memory_type,
+          pinned_memory_id, def.tensor_memory_type_, def.tensor_memory_id_,
+          def.pinned_memory_->TotalByteSize(), pinned_buffer,
+          def.tensor_buffer_ + def.tensor_buffer_offset_, stream_,
+          reinterpret_cast<void*>(&def.requests_), &completion_queue)));
```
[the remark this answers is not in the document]
CoderHam: Each CopyBuffer corresponds to a set of deferred requests. The scope of `def` is limited to `for (auto& def : deferred_pinned_)` whereas we need to iterate and report errors for all requests in the RequestsList.

src/core/backend_context.cc
```diff
@@ -839,30 +868,67 @@ BackendInputCollector::Finalize()
...
+      Status status = CopyBuffer(
+          "pinned buffer", pinned_memory_type, pinned_memory_id,
+          def.tensor_memory_type_, def.tensor_memory_id_,
+          def.pinned_memory_->TotalByteSize(), pinned_buffer,
+          def.tensor_buffer_ + def.tensor_buffer_offset_, stream_, &cuda_used);
+      need_sync_ |= cuda_used;
+
+      // If something goes wrong with the copy all the pending
+      // responses fail...
+      if (!status.IsOk()) {
```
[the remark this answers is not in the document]
CoderHam: To share this for both cases we would need to capture and store status and def.requests_ for the sequential CopyBuffer case as well

src/core/backend_context.cc
```diff
@@ -839,30 +868,67 @@ BackendInputCollector::Finalize()
...
+  for (size_t i = 0; i < async_task_count; i++) {
+    auto completed_item = completion_queue.Get();
+    auto completed_status = std::get<0>(completed_item);
+    need_sync_ |= std::get<1>(completed_item);
 
     // If something goes wrong with the copy all the pending
     // responses fail...
-    if (!status.IsOk()) {
-      for (auto& pr : def.requests_) {
+    if (!completed_status.IsOk()) {
```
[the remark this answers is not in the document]
CoderHam: This is not consistent with our earlier discussion of using async_task_count to count the number of AsyncTasks being run and then accumulate the cuda_used and status from them. Do we keep a vector of status from the regular CopyBuffer case and accumulate that as well?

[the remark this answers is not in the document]
deadeyegoodwin: I see, this code is ok then.

src/core/backend_context.cc
CoderHam: Fixed

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,84 @@
...
+#include <thread>
+#include <vector>
+
+namespace nvidia { namespace inferenceserver {
+
+AsyncWorkQueue::~AsyncWorkQueue()
+{
+  for (const auto& worker_thread : worker_threads_) {
+    GetSingleton()->task_queue_.Put(nullptr);
+    worker_thread->join();
```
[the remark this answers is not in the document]
CoderHam: Good catch. This might explain the thread related issue I saw earlier.

GuanLuo:
Closing this PR as it doesn't provide benefit as expected, the overhead of dispatching copy jobs is over the gain of parallel copies. We may keep the branch in case we want to revisit it later

PR #1911, opened 2020-08-19
Replace mktemp with mkdtemp

Tabrizian:
Fix build warning complaining about dangerous use of `mktemp`.


Tabrizian:
Fixed the merge conflict.

PR #1912, opened 2020-08-19
Expose sequence information to Backend APIs

tanmayv25:
Some of the backends might need this information about requests for their proper handling. 
Kaldi custom backend needs this information.

### DIFF
--- a/src/backends/backend/triton_model.cc
+++ b/src/backends/backend/triton_model.cc
@@ -528,6 +528,33 @@ TRITONBACKEND_RequestCorrelationId(TRITONBACKEND_Request* request, uint64_t* id)
 }
 
 TRITONSERVER_Error*
+TRITONBACKEND_RequestStartsSequence(
+    TRITONBACKEND_Request* request, bool* starts_seq)
+{
+  InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
+  *starts_seq = tr->Flags() & TRITONSERVER_REQUEST_FLAG_SEQUENCE_START;
+  return nullptr;  // success
+}
+
+TRITONSERVER_Error*
+TRITONBACKEND_RequestEndsSequence(
+    TRITONBACKEND_Request* request, bool* ends_seq)
+{
+  InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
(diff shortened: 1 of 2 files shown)


src/backends/backend/triton_model.cc
```diff
@@ -527,6 +527,33 @@ TRITONBACKEND_RequestCorrelationId(TRITONBACKEND_Request* request, uint64_t* id)
...
+TRITONBACKEND_RequestEndsSequence(
+    TRITONBACKEND_Request* request, bool* ends_seq)
+{
+  InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
+  *ends_seq = tr->Flags() & TRITONSERVER_REQUEST_FLAG_SEQUENCE_END;
+  return nullptr;  // success
+}
+
+TRITONSERVER_Error*
+TRITONBACKEND_RequestBatchSize(
```
deadeyegoodwin, line 549: We are not exposing batch-size explicitly anymore. InferenceRequest::BatchSize is going to be removed once all backends are moved over to the new API.

src/backends/backend/triton_model.cc
```diff
@@ -527,6 +527,33 @@ TRITONBACKEND_RequestCorrelationId(TRITONBACKEND_Request* request, uint64_t* id)
   return nullptr;  // success
 }
 
+TRITONSERVER_Error*
+TRITONBACKEND_RequestStartsSequence(
```
deadeyegoodwin, line 531: The way backends are supposed to know sequence start and end is to use the control inputs. How are they determining this in their current code? We could expose the request flags as a single TRITONBACKEND_RequestFlags function.

deadeyegoodwin:
@tanmayv25 be sure to increment the minor version when making any change to tritonbackend.h

src/backends/backend/triton_model.cc
tanmayv25: I see.. I will revert back to the control inputs...

tanmayv25:
Closing this PR as the changes are no longer required.

deadeyegoodwin:
@tanmayv25  I think it makes sense to add the Flags API. The control inputs are useful if you plan to feed them into a model or otherwise treat them as tensor inputs. If not it may be more convenient to have the flags directly if you are just using them for some control-flow, etc.

PR #1913, opened 2020-08-19
Fix undefined protobuf symbols in the client shared library

Tabrizian:
Fix the ldscript to make symbols starting with `inference::` global.

### DIFF
--- a/qa/L0_client/grpc_test.cc
+++ b/qa/L0_client/grpc_test.cc
@@ -32,6 +32,8 @@ int
 main(int argc, char* argv[])
 {
   std::unique_ptr<nvic::InferenceServerGrpcClient> client;
+  // Add a symbol from protobufs to verify correct linking
+  inference::ModelConfigResponse model_config;
   nvic::InferenceServerGrpcClient::Create(&client, "localhost:8001");
   bool live;
   client->IsServerLive(&live);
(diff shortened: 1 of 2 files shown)


qa/L0_client/grpc_test.cc
```diff
@@ -32,6 +32,8 @@ int
 main(int argc, char* argv[])
 {
   std::unique_ptr<nvic::InferenceServerGrpcClient> client;
+  // Add a symbol from protobufs to verify correct linking
+  inference::ModelConfigResponse model_config;
```
deadeyegoodwin, line 36: Don't you get a build failure for unused variable?

Tabrizian: The build process for these test clients is separate. I didn't use `-Werror` flag. I can create another test case and move the sample clients from here to there so that I can print some actual responses from Triton server. I can't start a server in this test case since it is placed in `/tmp` and I believe this is only for building clients.

GuanLuo: Will this test fail without the fix below? I am curious because we do have other C++ clients that uses `inference::ModelConfigResponse` but those clients can be compiled without this PR (except that those are built with CMake)

Tabrizian: Yes, the reason is that the other clients are built with the static library and it seems like CMake doesn't apply ldscript to those libraries. The static libraries currently do not have any problem and `inference::ModelConfigResponse` can be found there.

deadeyegoodwin: Since this is just a test it is ok to not use -Werror. I'll merge as is.

PR #1924, opened 2020-08-25
Add model config for output scattering

GuanLuo:


src/core/model_config.proto
```diff
@@ -327,6 +327,14 @@ message ModelOutput
...
   //@@
   bool is_shape_tensor = 6;
+
+  //@@  .. cpp:var:: string scatter_with_input_shape
+  //@@
+  //@@     This field can be specified if the output  should be scattered
+  //@@     according to a the input's shape. The dynamic dimension of the
+  //@@     output will be set to the value of the same dimension in the input.
+  //@@
+  string scatter_with_input_shape = 7;
```
deadeyegoodwin, line 337: I would rather have this handled in a manner similar to batch_input. So Add a batch_output with structure similar to batch_input.

GuanLuo: Changed the option to be a BatchOutput message

src/core/model_config.proto
```diff
@@ -327,6 +363,13 @@ message ModelOutput
   //@@     model.
   //@@
   bool is_shape_tensor = 6;
+
+  //@@  .. cpp:var:: BatchOutput batch_output
+  //@@
+  //@@     Instruction for output that requires special handling by the model
+  //@@     backend.
+  //@@
+  BatchOutput batch_output = 7;
```
deadeyegoodwin, line 372: Put at top-level as peer to batch_input.

PR #1926, opened 2020-08-25
Add CUDA graph specification in model config

GuanLuo:
@deadeyegoodwin Given the limitation of CUDA graph, this is the new config option to increase the usefulness of CUDA graph in the case of dynamic shape. User needs to specify the combinations explicitly so that the users can control what the graphs to be captured for different models and Triton doesn't need to populate the combinations.

### DIFF
--- a/src/core/model_config.proto
+++ b/src/core/model_config.proto
@@ -600,7 +600,8 @@ message ModelOptimizationPolicy
       //@@
       //@@         Specification of tensor dimension.
       //@@
-      message Shape {
+      message Shape
+      {
         //@@        .. cpp:var:: int64 dim (repeated)
         //@@
         //@@           The dimension.
--- a/src/core/model_config.proto
+++ b/src/core/model_config.proto
@@ -600,7 +600,7 @@ message ModelOptimizationPolicy
       //@@
       //@@         Specification of tensor dimension.
       //@@
-      message Dims {
+      message Shape {
         //@@        .. cpp:var:: int64 dim (repeated)
         //@@
         //@@           The dimension.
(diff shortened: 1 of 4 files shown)


src/core/model_config.proto
```diff
@@ -547,6 +547,37 @@ message ModelOptimizationPolicy
...
+
+      //@@      .. cpp:var:: int32 batch_size
+      //@@
+      //@@         The batch size of the CUDA graph.
+      //@@
+      int32 batch_size = 1;
+
+      //@@      .. cpp:var:: map<string, Dims> input
+      //@@
+      //@@         The specification of the inputs.
```
deadeyegoodwin, line 576: This is the shape captured for each input. Does the shape include the batch dimension? Is 'batch_size' handled like max_batch_size? Can you add some more comments to clarify. In general I'm ok with the format.

src/core/model_config.proto
```diff
@@ -547,6 +547,37 @@ message ModelOptimizationPolicy
...
+    //@@
+    //@@       Specification of the CUDA graph to be captured.
+    //@@
+    message GraphSpec
+    {
+      //@@      .. cpp:var:: message Dims
+      //@@
+      //@@         Specification of tensor dimension.
+      //@@
+      message Dims {
```
deadeyegoodwin, line 560: Call this "Shape"

src/core/model_config.proto
GuanLuo: I will update that. The `batch_size` is handled like `max_batch_size` and the shape doesn't include batch dimension to match the convention in model config.

PR #1928, opened 2020-08-25
Some minor backend cleanup

deadeyegoodwin:


src/backends/backend/tensorflow/tensorflow.cc
```diff
@@ -788,9 +783,6 @@ ModelState::ModelState(TRITONBACKEND_Model* triton_model)
                                          Name() + "'")
                                             .c_str()));
   }
-
-  THROW_IF_BACKEND_MODEL_ERROR(nib::ModelPaths(
```
GuanLuo, line 792: You may also remove `ModelPaths` class if it is not going to be used.

deadeyegoodwin: I thought about that but it seems like it my be useful in the future so I just kept it.

PR #1934, opened 2020-08-27
Quick fix for valgrind related test failures

xprotobeast2:
Filter out new memory leaks and set TRITON_ENABLE_ASAN flag during base build


Dockerfile
```diff
@@ -319,6 +319,7 @@ RUN LIBCUDA_FOUND=$(ldconfig -p | grep -v compat | awk '{print $1}' | grep libcu
                   -DTRITON_ENABLE_METRICS_GPU=ON \
                   -DTRITON_ENABLE_STATS=ON \
                   -DTRITON_ENABLE_TRACING=ON \
+                  -DTRITON_ENABLE_ASAN=ON \
```
deadeyegoodwin, line 322: Don't turn on ASAN. It is not meant for a production build.

PR #1941, opened 2020-08-28
Fix GRPC shutdown

Tabrizian:
- [X] Fix GRPC graceful shutdown
- [X] Update args initialization model parameters
- [x] Update Python API error handling

### DIFF
--- a/src/backends/backend/python/CMakeLists.txt
+++ b/src/backends/backend/python/CMakeLists.txt
@@ -73,7 +73,8 @@ add_custom_target(
 
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
-# TODO: Fix Protobuf paths
+# FIXME The relative paths created by the protobuf_generate_cpp and
+# protobuf_generate_python are incorrect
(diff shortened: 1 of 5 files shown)


deadeyegoodwin:
@Tabrizian  Is this PR meant to only have the GRPC shutdown fix? It has a 10 commits and a lot of other code. Perhaps you need to rebase it?

Tabrizian:
@deadeyegoodwin It doesn't only fix the grpc shutdown. There were some other minor issues that this PR is trying to fix. The reason for 10 commits is that I cherry-picked all the changes except the CI tests. The issues that this PR fixes:

1. Add `TritonError` and `PythonModelException` to triton_python_utils for error handling in Python models.
2. Change the protobuf to accommodate for error objects.
3. Add `python_runitme` flag to enable the user to change the path to `python`.
3. Fix GRPC shutdown.

Sorry, I should've created separate PRs for each of them.

deadeyegoodwin:
No need to do that now, but you should change the description to be more general since it is more than just the GRPC shutdown fix.

PR #1949, opened 2020-08-31
Add CI tests for Python backend

Tabrizian:
- [X]  Test for dynamic batching of the requests
- [X]  Test for returning correct error when Python model has an error during inference phase
- [X]  Test for using Pytorch
- [X]  Test for when one of the outputs has an error
- [X]  Count the number of initialization arguments


qa/python_models/execute_error/config.pbtxt
```diff
@@ -0,0 +1,53 @@
...
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "execute_error"
+backend: "python"
+platform: "custom"
```
deadeyegoodwin, line 29: No platform

Tabrizian:
@deadeyegoodwin I didn't add the test case that had an error during initialization because of the problem that #1953 is fixing. It resulted in non-zero return codes and failure of the unit test. I can add it either here or in a later PR along with tests for data types.

PR #1953, opened 2020-08-31
Gracefully stop server if error during initialization

deadeyegoodwin:


src/core/server.cc
```diff
@@ -188,9 +188,9 @@ InferenceServer::Init()
 }
 
 Status
-InferenceServer::Stop()
+InferenceServer::Stop(const bool force)
 {
-  if (ready_state_ != ServerReadyState::SERVER_READY) {
+  if (!force && (ready_state_ != ServerReadyState::SERVER_READY)) {
     return Status::Success;
```
Tabrizian, line 194: So returning early from here was the root cause, right?

deadeyegoodwin: The problem was returning early from the Init() function without waiting for any in-flight threads to complete (which is what Stop() does).

PR #1956, opened 2020-08-31
Moved TRITONSERVER_DISABLE_BACKEND_UNLOAD into test script

xprotobeast2:


qa/common/util.sh
```diff
@@ -240,6 +240,8 @@ function run_server_nowait () {
 function run_server_leakcheck () {
     SERVER_PID=0
 
+    export TRITONSERVER_DISABLE_BACKEND_UNLOAD=1
```
deadeyegoodwin, line 243: Instead of export just put it at the front of the cmdline below in line 264 before the LD_PRELOAD

qa/common/util.sh
```diff
@@ -261,6 +261,7 @@ function run_server_leakcheck () {
       echo "=== Running LD_PRELOAD=$SERVER_LD_PRELOAD $SERVER $SERVER_ARGS"
     fi
 
+    TRITONSERVER_DISABLE_BACKEND_UNLOAD=1
```
deadeyegoodwin, line 264: You have to put it on the same line as the executable and then the envvar setting applies only to that executable (like the LD_PRELOAD in the following line). That way this setting won't make it into the shell env and perhaps impact other runs of tritonserver. You just leave a space:

TRITONSERVER_DISABLE_BACKEND_UNLOAD=1 LD_PRELOAD=$SERVER_LD_PRELOAD $LEAKCHECK ...

xprotobeast2: Ah okay got it, I've been trying multiple ways of doing this, and some mess up the SERVER_PID=$! line but I forgot about this way. Thanks!

PR #1977, opened 2020-09-03
Move common files to triton-inference-server/common repo

deadeyegoodwin:
Note that this is merging to davidg-repo, not master


src/backends/backend/python/python.cc
```diff
@@ -1028,3 +1019,5 @@ TRITONBACKEND_ModelInstanceFinalize(TRITONBACKEND_ModelInstance* instance)
 }
 
 }  // extern "C"
```
GuanLuo, line 1021: Can `extern "C"` be used within a namespace?

GuanLuo (approved):

src/backends/backend/python/python.cc
Tabrizian: It seems like it is ok but the namespace boundaries are not respected. See below:

https://stackoverflow.com/questions/28996944/extern-c-linkage-inside-c-namespace

build/client/CMakeLists.txt
```diff
@@ -48,6 +48,24 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
...
 
+#
+# Dependencies
+#
+# FetchContent's composibility isn't very good. We must include the
+# transitive closure of all repos so that we can override the tag.
+#
+include(FetchContent)
+
+set(REPO_TAG main)
```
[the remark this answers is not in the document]
deadeyegoodwin: Yes, we will have some way of tagging the multiple repos as part of the release. We need to discuss.

src/backends/backend/python/python.cc
deadeyegoodwin: It should be ok in this case. The extern "C" causes the TRITONBACKEND_* functions to not be mangled. But because they are in the namespace the functions and classes they call don't need to explicitly have the namespace prefix(es).

GuanLuo: Okay

PR #1980, opened 2020-09-03
Improve Python backend utils

Tabrizian:


src/backends/backend/python/resources/triton_python_backend_utils.py
```diff
@@ -226,6 +257,90 @@ def message(self):
...
+    """
+    input_tensors = inference_request.inputs()
+    for input_tensor in input_tensors:
+        if input_tensor.name() == name:
+            return input_tensor
+
+    return None
+
+
+def get_input_properties_by_name(model_config, name):
```
deadeyegoodwin, line 289: Why is it get_input_properties... and get_output_config...?
Use config for both.

qa/python_models/execute_error/model.py
```diff
@@ -41,10 +41,9 @@ def execute(self, requests):
         """
         responses = []
         for request in requests:
-            input_tensors = request.inputs()
-            in_all = input_tensors[0].numpy_array()
```
deadeyegoodwin, line 45: Can we call the function as_numpy() to be consistent with our client APIs?

deadeyegoodwin (changes requested):

qa/python_models/execute_error/model.py
Tabrizian: Fixed.

deadeyegoodwin (approved):

PR #1987, opened 2020-09-04
Support TRT profiles with static shapes

tanmayv25:
These changes fixes from the fallout with following update from TensorRT:
https://github.com/NVIDIA/TensorRT/blob/master/include/NvInferRuntime.h#L1163-L1181

In summary, if the presence of WILD_CARD in the engine dimension is now exposed to the profile level. It will not be -1 if a profile specifies `max_shape==min_shape==opt_shape`. 

This PR includes the following change:

1.  Relaxes the requirement of all profiles to span from `1 - max_batch_size` in the batch dimension.
2.  Moves the property of being dynamic from engine level to individual bindings(each profile has their own independent set of bindings).
3. Ensure correct static profile is selected as per the shapes received in the request.
4. Attempts to minimize the `SetBindingDimension` call to only those bindings that are dynamic in nature which might lead to some compute improvements.

For instance, a user can specify max_batch_size=4 with the following four static profile shapes: [1, 128], [2,128], [3,128], [4, 128].
The only requirement is all the instance groups must explicitly include all the profiles.


src/core/model_config.proto
```diff
@@ -157,13 +157,14 @@ message ModelInstanceGroup
 
   //@@  .. cpp:var:: string profile (repeated)
   //@@
-  //@@     For TensorRT models, using inputs with dynamic shape, this
+  //@@     For TensorRT models, containing multiple optimization profile, this
```
deadeyegoodwin, line 160: Drop first comma. "For TensorRT models containing ..."

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #1991, opened 2020-09-08
Add CI tests to make Python backend work with L0_infer

Tabrizian:


qa/common/gen_qa_model_repository
```diff
@@ -190,8 +190,10 @@ cat >$HOST_SRCDIR/$C2SCRIPT <<EOF
 #!/bin/bash
 set -e
 python3 $SRCDIR/gen_qa_models.py --netdef --libtorch --models_dir=$DESTDIR
+python3 $SRCDIR/gen_qa_models.py --python --models_dir=$DESTDIR
```
deadeyegoodwin, line 193: As we discussed, I don't think we need to generate models this way.

qa/common/test_util.py
```diff
@@ -215,6 +215,27 @@ def validate_for_libtorch_model(input_dtype, output0_dtype, output1_dtype,
     return True
 
 
+def validate_for_python_model(input_dtype, output0_dtype, output1_dtype,
+                              input_shape, output0_shape, output1_shape):
+    """Return True if input and output dtypes are supported by a Python model."""
+
+    # STRING, FLOAT16 and UINT16 data types are not supported currently
```
deadeyegoodwin, line 222: Why are these not supported?

deadeyegoodwin (changes requested):

qa/common/test_util.py
Tabrizian: I thought it is a limitation with the HTTP api that cannot contain FP16 and UINT16. It seems to be working with the Python backend.

qa/L0_infer/test.sh
```diff
@@ -35,6 +35,9 @@ if [ -z "$REPO_VERSION" ]; then
     exit 1
 fi
 
+# Install pytorch
+pip3 install torch==1.6.0+cpu torchvision==0.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```
deadeyegoodwin, line 39: Why do we need pytorch to run L0_infer? We can't install pytorch here as it may interfere with the existing libtorch testing happening in this test.

qa/common/gen_qa_model_repository
```diff
@@ -170,7 +170,7 @@ if [ $? -ne 0 ]; then
 fi
 
 docker pull $ONNX_IMAGE
-docker run --gpus device=$CUDA_DEVICE --rm --entrypoint $SRCDIR/$ONNXSCRIPT \
+docker run --gpus device=$CUDA_DEVICE --rm --entrypoint $SRCDIR/$ONNXSCRIPT
```
deadeyegoodwin, line 173: I don't think you want this edit.

deadeyegoodwin (changes requested):

qa/L0_infer/test.sh
CoderHam: I'd have a separate L0_infer_python test for the python scripts that installs the necessary packages. Additionally why do you need torchvision for any of your tests? Don't think you use torchvision ops in these addsub models

qa/common/gen_qa_models.py
```diff
@@ -1280,6 +1280,229 @@ def create_onnx_modelconfig(models_dir, max_batch, model_version, input_shape,
...
+                out_0, out_1 = self.model(in_0.as_numpy().astype(np.int32), in_1.as_numpy().astype(np.int32))
+                out_tensor_0 = pb_utils.Tensor("OUTPUT0", out_0)
+                out_tensor_1 = pb_utils.Tensor("OUTPUT1", out_1)
+            else:
+                out_0, out_1 = self.model(in_0.as_numpy(), in_1.as_numpy())
+                out_tensor_0 = pb_utils.Tensor("OUTPUT0", out_0)
+                out_tensor_1 = pb_utils.Tensor("OUTPUT1", out_1)
+            responses.append(pb_utils.InferenceResponse([out_tensor_0, out_tensor_1]))
+        return responses
+'''.format(input_shape, torch_input_dtype, example_input, example_input)
```
CoderHam, line 1360: Where are these variables being used in the string placeholder?

qa/common/gen_qa_models.py
```diff
@@ -1672,6 +1913,10 @@ def create_fixed_models(models_dir,
                         required=False,
                         action='store_true',
                         help='Generate Pytorch LibTorch models')
+    parser.add_argument('--python',
+                        required=False,
+                        action='store_true',
+                        help='Generate Python models')
```
CoderHam, line 1919: Python (PyTorch) models? Since currently you are only generating pytorch models

qa/common/gen_qa_models.py
```diff
@@ -1697,6 +1942,9 @@ def create_fixed_models(models_dir,
     if FLAGS.libtorch:
         import torch
         from torch import nn
+    if FLAGS.python:
```
CoderHam, line 1945: Group with libtorch using **or**

CoderHam (changes requested):

qa/L0_infer/test.sh
Tabrizian: I've removed Pytorch from all the installations and simplified AddSub so that it doesn't require PyTorch and uses only numpy.

qa/common/gen_qa_models.py
Tabrizian: Fixed

qa/common/gen_qa_models.py
Tabrizian: In the new version, they are Python models.

qa/L0_infer/infer_test.py
```diff
@@ -232,6 +232,17 @@ def _infer_exact_helper(tester,
                                         output1_raw=output1_raw,
                                         swap=swap)
 
+        if prefix == "":
```
deadeyegoodwin, line 235: You should still use tu.validate_for_python_model here. It can just always return true.

qa/python_models/sub_add/model.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin, line 1: It strange to have both add_sub and sub_add when they are nearly identical. There is a better way to do this that we can discuss.

deadeyegoodwin (commented):
There are a couple of follow-on changes we can make that I commented on, but I'm going to merge it as is for now.

PR #1997, opened 2020-09-09
Document int64->string for as_json=true results in grpc client

deadeyegoodwin:


CoderHam:
LGTM. If this is a _bug_ in GRPC then it would be great to link in a related ticket (Or maybe I missed it in another discussion) we can use to track the progress of the fix.

CoderHam (approved):

deadeyegoodwin:
The issue has been reported many times... protobuf refuses to fix it. For example: https://github.com/protocolbuffers/protobuf/issues/2954

PR #2013, opened 2020-09-12
add support for customizing s3 endpoint url through ENV variable

wbzqe:
I want to use a private cloud OSS which compatible with s3 api ,  but triton unable to set the endpoint url

### DIFF
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -756,12 +756,16 @@ S3FileSystem::S3FileSystem(
   const char* secret_key = std::getenv("AWS_SECRET_ACCESS_KEY");
   const char* key_id = std::getenv("AWS_ACCESS_KEY_ID");
   const char* region = std::getenv("AWS_DEFAULT_REGION");
+  const char* endpoint = std::getenv("AWS_ENDPOINT_URL");
   if ((secret_key != NULL) && (key_id != NULL)) {
     credentials.SetAWSAccessKeyId(key_id);
     credentials.SetAWSSecretKey(secret_key);
     config = Aws::Client::ClientConfiguration();
     if (region != NULL) {
       config.region = region;
+    }
+   if (endpoint != NULL) {
+      config.endpointOverride = endpoint;
     }
   } else if (const char* profile_name = std::getenv("AWS_PROFILE")) {
     config = Aws::Client::ClientConfiguration(profile_name);


deadeyegoodwin:
@CoderHam  Why does host:port support not work for this case?

CoderHam:
It is odd. I am double checking the regex to confirm that it supports such cases. Based on the changes it appears that might be the cause of this failure. If I fix the regex this change will not be needed.

CoderHam:
@wbzqe The endpoint url does not specify any port. I tested `s3://s3-cn-east-1.qiniucs.com:8000/bucket/test` and it works fine. Please repeat your experiment with the `:<port>` included.

wbzqe:
> @wbzqe The endpoint url does not specify any port. I tested `s3://s3-cn-east-1.qiniucs.com:8000/bucket/test` and it works fine. Please repeat your experiment with the `:<port>` included.

The error output:

```
root@triton-server-78d479f96c-vh8l4:/opt/tritonserver# tritonserver --model-repository=s3://s3-cn-east-1.qiniucs.com:80/demo-bucket/triton-models
I0915 02:21:03.546184 101 metrics.cc:184] found 2 GPUs supporting NVML metrics
I0915 02:21:03.551739 101 metrics.cc:193]   GPU 0: P106-100
I0915 02:21:03.557510 101 metrics.cc:193]   GPU 1: P106-100
I0915 02:21:03.557702 101 server.cc:119] Initializing Triton Inference Server
W0915 02:21:03.938804 101 server.cc:160] failed to enable peer access for some device pairs
error: creating server: Invalid argument - repository path is not a valid directory
```

I read the code and tested "s3_regex " with "s3://s3-cn-east-1.qiniucs.com:80/bucket/test" , it works fine .  

But  ` !list_objects_outcome.GetResult().GetContents().empty()`  got  `false` , This is where the problem in my case should be . 

Then i tried `s3fs` with same endpoint ,  it works fine .

```
Status
S3FileSystem::IsDirectory(const std::string& path, bool* is_dir)
{
  *is_dir = false;
  std::string bucket, object_path;
  RETURN_IF_ERROR(ParsePath(path, &bucket, &object_path));

  // Check if the bucket exists
  s3::Model::HeadBucketRequest head_request;
  head_request.WithBucket(bucket.c_str());

  auto head_bucket_outcome = client_.HeadBucket(head_request);
  if (!head_bucket_outcome.IsSuccess()) {
    return Status(
        Status::Code::INTERNAL,
        "Could not get MetaData for bucket with name " + bucket);
  }

  // Root case - bucket exists and object path is empty
  if (object_path.empty()) {
    *is_dir = true;
    return Status::Success;
  }

  // List the objects in the bucket
  s3::Model::ListObjectsRequest list_objects_request;
  list_objects_request.SetBucket(bucket.c_str());
  list_objects_request.SetPrefix(AppendSlash(object_path).c_str());
  auto list_objects_outcome = client_.ListObjects(list_objects_request);

  if (list_objects_outcome.IsSuccess()) {
    *is_dir = !list_objects_outcome.GetResult().GetContents().empty();
  } else {
    return Status(
        Status::Code::INTERNAL, "Failed to list objects with prefix " + path);
  }
  return Status::Success;
}
```

PR #2029, opened 2020-09-15
Add cmake controls for repo tags

deadeyegoodwin:


GuanLuo (approved):

Dockerfile
```diff
@@ -325,6 +325,10 @@ COPY --from=tritonserver_onnx /workspace/build/Release/testdata/custom_op_librar
 # - Need to find CUDA stubs if they are available since some backends
 # may need to link against them. This is identical to the logic in TF
 # container nvbuild.sh
+ARG TRITON_COMMON_REPO_TAG=main
+ARG TRITON_CORE_REPO_TAG=main
+ARG TRITON_BACKEND_REPO_TAG=main
```
CoderHam, line 330: We would need to have these build args be overridden by the gitlab CI (to be the branch name?). Moreover we would want to set/overwrite these values using CI variables.

deadeyegoodwin: Typically you just want to use the main branch of the dependent repos. If your CI run requires a change form one of the repos then you edit Dockerfile and include that in CI run. And then need to remove it before merging. I agree that is awkward and we will have something better once the reorg is further along.

deadeyegoodwin: But we could enhance CI to have addition vars that could be passed into the build

CoderHam: I think enhancing the CI route would be cleaner. Nevertheless these changes look good and can be merged in

CoderHam (approved):

PR #2030, opened 2020-09-15
Return appropriate error when 'data' value is null

CoderHam:


src/servers/http_server.cc
```diff
@@ -2110,8 +2110,16 @@ HTTPAPIServer::EVBufferToInput(
               TRITONSERVER_MEMORY_CPU, 0 /* memory_type_id */));
         } else {
           // JSON... presence of "data" already validated but still
-          // checking here. Flow in this endpoint needs to be
-          // reworked...
+          // checking here. First check if Null then fetch array contents.
+          // Flow in this endpoint needs to be reworked...
+          bool is_null;
+          RETURN_MSG_IF_ERR(
+              request_input.IsMemberNull("data", &is_null),
```
deadeyegoodwin, line 2117: So there is a "data" field but its value is null, as in 
"data" : null
If so the client shouldn't be sending that. It should just not send a data field. If request has "data" : null then I am ok with the existing error message.

deadeyegoodwin (changes requested):

src/servers/http_server.cc
CoderHam: Yes that's correct. I can revert this change and fix the python client library instead.

CoderHam: After adding the check, if 'data' is not set, we get the following error:
```
tritonclient.utils.InferenceServerException: Input must set only one of the following fields: 'data', 'binary_data_size' in 'parameters', 'shared_memory_region' in 'parameters'. But no field is set
```

src/servers/http_server.cc
```diff
@@ -2112,7 +2112,6 @@ HTTPAPIServer::EVBufferToInput(
           // JSON... presence of "data" already validated but still
           // checking here. Flow in this endpoint needs to be
           // reworked...
-          triton::common::TritonJson::Value tensor_data;
```
deadeyegoodwin, line 2115: Doesn't this cause a compile error.

deadeyegoodwin (changes requested):

src/servers/http_server.cc
CoderHam: Typo. Fixed.

deadeyegoodwin (approved):

PR #2034, opened 2020-09-16
Remove Python baceknd

Tabrizian:
This should be merged after #2033.


deadeyegoodwin (commented):
This looks ok, but you need to base this change on davidg-install and make the appropriate changes to resolve the conflicts that will occur.

Dockerfile
```diff
@@ -366,7 +365,6 @@ RUN LIBCUDA_FOUND=$(ldconfig -p | grep -v compat | awk '{print $1}' | grep libcu
             mkdir -p /opt/tritonserver/include && \
             cp -r server/install/bin /opt/tritonserver/. && \
             cp -r server/install/lib /opt/tritonserver/. && \
-            cp -r server/install/backends /opt/tritonserver/. && \
```
CoderHam, line 369: Remember to make the necessary changes in build_variants test as well.

Tabrizian: Right, I should change those tests too.

deadeyegoodwin (approved):

PR #2039, opened 2020-09-17
Formalize reformat-free input. Minor plan_backend clean up

GuanLuo:
In server-client communication, the input shape will still be in linear format so that plan backend can set binding dimension properly, but it will expect the client to prepare the input data in the corresponding non-linear form.

Cleaned up the use of `binding_index` and `io_index`. `binding_index` is the index that takes profile index into account, and it is used to prepare the executing profile properly, `io_index` is used to refer to io information that is consistent across profiles.


deadeyegoodwin (approved):
Looks good so I'll merge. There are some FIXME in the code and we need some testing. Do you have separate tickets for those?

GuanLuo:
Added followup tickets

PR #2044, opened 2020-09-19
Fix graceful termination

Tabrizian:
If you press Ctrl-C during model loading, Triton does not trigger `SingleHandler` which is required for a graceful shutdown.

### DIFF
--- a/src/servers/main.cc
+++ b/src/servers/main.cc
@@ -1154,6 +1154,10 @@ main(int argc, char** argv)
   auto shm_manager =
       std::make_shared<nvidia::inferenceserver::SharedMemoryManager>();
 
+  // Trap SIGINT and SIGTERM to allow server to exit gracefully
+  signal(SIGINT, SignalHandler);
+  signal(SIGTERM, SignalHandler);
+
   // Create the server...
   TRITONSERVER_Server* server_ptr = nullptr;
   FAIL_IF_ERR(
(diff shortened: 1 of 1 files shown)


src/servers/main.cc
```diff
@@ -1154,6 +1154,10 @@ main(int argc, char** argv)
   auto shm_manager =
       std::make_shared<nvidia::inferenceserver::SharedMemoryManager>();
 
+  // Trap SIGINT and SIGTERM to allow server to exit gracefully
+  signal(SIGINT, SignalHandler);
```
deadeyegoodwin, line 1158: The problem with moving this here is that if you use Ctrl-C while loading is happening, then loading will continue until all the models are loaded... and then the endpoints will be started... and then triton will recognize the Ctrl-C and unload the models and exit. That means there can be a long delay between using Ctrl-C and triton actually exiting. Have you found a specific problem with the way things are currently handled.

Tabrizian: The problem is that if Ctrl-C is pressed before all the models are loaded, Triton will not trigger `ModelInstanceFinalize` and `ModelFinalize`. In the case of Python backend, this leads to not terminating GRPC servers for the models that have already been loaded.

deadeyegoodwin: So what happens when triton exits non-gracefully?  Non-graceful exit needs to be handled because the sig handler can't catch all exits.

Tabrizian:
I will create a separate PR to address this issue.

PR #2045, opened 2020-09-19
Fix GRPC server to maintain non decoupled response state properly

GuanLuo:


src/servers/grpc_server.cc
```diff
@@ -2205,6 +2205,19 @@ InferResponseFree(
...
 
+TRITONSERVER_Error*
+InferResponseStart(TRITONSERVER_ResponseAllocator* allocator, void* userp)
+{
+  AllocPayload<inference::ModelInferResponse>* payload =
+      reinterpret_cast<AllocPayload<inference::ModelInferResponse>*>(userp);
+
+  // ModelInfer RPC expects exactly one response per request. Hence, always call
+  // GetNonDecoupledResponse() to create one response object on response start.
+  payload->response_queue_->GetNonDecoupledResponse();
```
tanmayv25, line 2216: I don't understand how does it fix the gRPC. See Line 2186, it will allocate/get the response whenever an allocation is attempted. For non-decoupled case, the allocation will be made in the first call and later the reference to the same response will be return. This will just be an extra step.

tanmayv25 (approved):
I think it will handle the cases with no outputs. approving.. but do explain how it solves the linked problem.

PR #2047, opened 2020-09-20
support base url in httpclient

spacedragon:
Current the python HTTP client doesn't support a URL with a base path. 
So if I init the client with 

```
    tritonhttpclient.InferenceServerClient("http://a.b.c.d/base_path")   
```

The `base_path` is ignored

### DIFF
--- a/src/clients/python/library/tritonclient/http/__init__.py
+++ b/src/clients/python/library/tritonclient/http/__init__.py
@@ -189,8 +189,11 @@ class InferenceServerClient:
                  ssl_options=None,
                  ssl_context_factory=None,
                  insecure=False):
-        scheme = "https://" if ssl else "http://"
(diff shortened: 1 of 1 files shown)


DavidLangworthy (approved):
Looks good.  Thank you.

deadeyegoodwin:
Thanks for the PR and from it we understand your request. Any enhancement to client API like this must be made in both Python and C++ clients. We will look into how to best do that and let you know if it aligns with this change.

tanmayv25:
@deadeyegoodwin We can merge this PR. This issue was specific to `geventhttpclient` in our python side implementation. The libcurl based C++ implementation takes into account of the complete path.

tanmayv25:
The implementation from this PR is included in the PR #2062. This adds proper documentation and the tests cases over it.

PR #2055, opened 2020-09-22
Add backend state summary to Triton logs

Tabrizian:
A screenshot of how the logs will look like:

image

### DIFF
--- a/src/backends/backend/triton_backend_manager.cc
+++ b/src/backends/backend/triton_backend_manager.cc
@@ -329,13 +329,15 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 //
 
 TritonBackendManager&
-TritonBackendManager::Instance() {
+TritonBackendManager::Instance()
+{
   static TritonBackendManager triton_backend_manager{};
   return triton_backend_manager;
 }
 
 const std::unordered_map<std::string, std::weak_ptr<TritonBackend>>&
-TritonBackendManager::BackendMap() {
+TritonBackendManager::BackendMap()
+{
   return backend_map_;
 }
 
@@ -345,7 +347,7 @@ TritonBackendManager::CreateBackend(
     const BackendCmdlineConfig& backend_cmdline_config,
     std::shared_ptr<TritonBackend>* backend)
 {
-  TritonBackendManager &singleton_manager = TritonBackendManager::Instance();
+  TritonBackendManager& singleton_manager = TritonBackendManager::Instance();
 
   std::lock_guard<std::mutex> lock(singleton_manager.Mutex());
 
--- a/src/backends/backend/triton_backend_manager.cc
+++ b/src/backends/backend/triton_backend_manager.cc
@@ -327,15 +327,27 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 //
 // TritonBackendManager
 //
+
+TritonBackendManager&
+TritonBackendManager::Instance() {
+  static TritonBackendManager triton_backend_manager{};
+  return triton_backend_manager;
+}
+
+const std::unordered_map<std::string, std::weak_ptr<TritonBackend>>&
+TritonBackendManager::BackendMap() {
+  return backend_map_;
+}
+
 Status
 TritonBackendManager::CreateBackend(
     const std::string& name, const std::string& dir, const std::string& libpath,
     const BackendCmdlineConfig& backend_cmdline_config,
     std::shared_ptr<TritonBackend>* backend)
 {
-  static TritonBackendManager singleton_manager;
+  TritonBackendManager &singleton_manager = TritonBackendManager::Instance();
 
-  std::lock_guard<std::mutex> lock(singleton_manager.mu_);
+  std::lock_guard<std::mutex> lock(singleton_manager.Mutex());
 
   const auto& itr = singleton_manager.backend_map_.find(libpath);
   if (itr != singleton_manager.backend_map_.end()) {
--- a/src/backends/backend/triton_backend_manager.cc
+++ b/src/backends/backend/triton_backend_manager.cc
@@ -328,7 +328,7 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 // TritonBackendManager
 //
 TritonBackendManager&
-TritonBackendManager::Instance_()
+TritonBackendManager::Singleton()
 {
   static TritonBackendManager triton_backend_manager;
   return triton_backend_manager;
@@ -340,7 +340,7 @@ TritonBackendManager::CreateBackend(
     const BackendCmdlineConfig& backend_cmdline_config,
     std::shared_ptr<TritonBackend>* backend)
 {
-  TritonBackendManager& singleton_manager = Instance_();
+  TritonBackendManager& singleton_manager = Singleton();
   std::lock_guard<std::mutex> lock(singleton_manager.mu_);
 
   const auto& itr = singleton_manager.backend_map_.find(libpath);
@@ -365,16 +365,18 @@ TritonBackendManager::CreateBackend(
   return Status::Success;
 }
 
-std::unique_ptr<std::unordered_map<std::string, std::vector<std::string>>>
-TritonBackendManager::BackendState()
+Status
+TritonBackendManager::BackendState(
+    std::unique_ptr<std::unordered_map<std::string, std::vector<std::string>>>*
+        backend_state)
 {
-  TritonBackendManager& singleton_manager = Instance_();
+  TritonBackendManager& singleton_manager = Singleton();
   std::lock_guard<std::mutex> lock(singleton_manager.mu_);
 
   auto backend_map = singleton_manager.backend_map_;
 
   std::unique_ptr<std::unordered_map<std::string, std::vector<std::string>>>
-      backend_state(
+      backend_state_map(
           new std::unordered_map<std::string, std::vector<std::string>>);
   for (const auto& backend_pair : backend_map) {
     auto libpath = backend_pair.first;
@@ -384,12 +386,13 @@ TritonBackendManager::BackendState()
       const char* backend_config;
       size_t backend_config_size;
       backend->BackendConfig().Serialize(&backend_config, &backend_config_size);
-      backend_state->insert(
+      backend_state_map->insert(
           {backend->Name(), std::vector<std::string>{libpath, backend_config}});
     }
   }
+  *backend_state = std::move(backend_state_map);
 
-  return backend_state;
+  return Status::Success;
 }
 
 }}  // namespace nvidia::inferenceserver
--- a/src/backends/backend/triton_backend_manager.cc
+++ b/src/backends/backend/triton_backend_manager.cc
@@ -328,18 +328,8 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 // TritonBackendManager
 //
 
-TritonBackendManager&
-TritonBackendManager::Instance()
-{
-  static TritonBackendManager triton_backend_manager{};
-  return triton_backend_manager;
-}
-
-const std::unordered_map<std::string, std::weak_ptr<TritonBackend>>&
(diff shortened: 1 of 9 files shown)


CoderHam:
If you add `server_id`, `server_version` and `server_extension` to the table it can be removed from server.cc

src/core/server.cc
```diff
@@ -109,6 +113,90 @@ InferenceServer::InferenceServer()
...
+  auto backend_map = triton_backend_manager.BackendMap();
+
+  for (const auto& backend_pair : backend_map) {
+    std::vector<std::string> backend_record;
+    auto backend = backend_pair.second.lock();
+    if (backend != nullptr) {
+      backend_record.emplace_back(backend->Name());
+      const char* config_message;
+      size_t config_message_size;
+      backend->BackendConfig().Serialize(&config_message, &config_message_size);
```
CoderHam, line 185: Do we want to print out the entire config message? I think this is necessary only if the model config is autofilled.

CoderHam (changes requested):

src/core/server.cc
```diff
@@ -47,6 +47,10 @@
 #include "src/core/pinned_memory_manager.h"
 #include "src/core/server.h"
 
+#include "src/backends/backend/triton_backend_manager.h"
+
```
CoderHam, line 51: Remove extra lines after and before triton_backend_manager.h . 
Also if you check 
`#include "src/core/server.h"`
Is mistakenly added twice. Remove second occurrence.

deadeyegoodwin:
@Tabrizian In backend table "path" and "config" are swapped. 
Can you put the backend table before the model table.
Put version and extensions in the first table and then you can remove them from the other logging.

src/backends/backend/triton_backend_manager.cc
```diff
@@ -327,15 +327,29 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 //
 // TritonBackendManager
 //
+
+TritonBackendManager&
+TritonBackendManager::Instance()
```
deadeyegoodwin, line 332: Instead of breaking the encapsulation of the class by exposing these internals, just add a BackendState() method that fills a map from backend-name to a record of state information (path and config).

src/core/server.cc
```diff
@@ -109,6 +113,90 @@ InferenceServer::InferenceServer()
...
+            ModelReadyStateString(model_status_pair.first);
+
+        if (model_status_pair.second != "") {
+          model_status += ": " + model_status_pair.second;
+        }
+
+        model_record.emplace_back(model_name);
+        model_record.emplace_back(model_version);
+        model_record.emplace_back(model_status);
+        models_table->insert_row(model_record);
```
CoderHam, line 155: You are re-using the same model_record for all versions in the model_version_map. Shift declaration of model record inside for loop.

src/core/server.cc
```diff
@@ -109,6 +113,90 @@ InferenceServer::InferenceServer()
...
+  triton::common::TablePrinter::Create(&models_table, model_headers);
+
+  for (const auto& model_state : model_states) {
+    std::vector<std::string> model_record;
+    auto model_version_map = model_state.second;
+    std::string model_name = model_state.first;
+
+    // If model_version_map size is zero, no version is found for this model
+    if (model_version_map.size() == 0) {
+      model_record.emplace_back(model_name);
```
CoderHam, line 137: Shift declaration of model_record to here

src/backends/backend/triton_backend_manager.h
```diff
@@ -143,8 +143,24 @@ class TritonBackendManager {
...
+  static TritonBackendManager& Instance();
+  static std::mutex& Mutex()
+  {
+    static std::mutex m;
+    return m;
+  }
+
+  TritonBackendManager() {}
+
+  TritonBackendManager(TritonBackendManager const&) = delete;
```
deadeyegoodwin, line 155: Use DISALLOW_COPY_AND_ASSIGN macro from constants.h.
Put macro at the start of the "private:" section.

CoderHam (changes requested):

src/core/server.cc
```diff
@@ -189,12 +277,15 @@ InferenceServer::Init()
       // failure is due to a model not loading correctly so we just
       // continue if not exiting on error.
       ready_state_ = ServerReadyState::SERVER_READY;
+      PrintSummary();
     }
-    return status;
+  } else {
+    ready_state_ = ServerReadyState::SERVER_READY;
+    status = Status::Success;
```
deadeyegoodwin, line 284: Setting status here is redundant

src/core/server.h
```diff
@@ -97,6 +97,9 @@ class InferenceServer {
   Status ModelReadyVersions(
       std::map<std::string, std::vector<int64_t>>* model_versions);
 
+  // Print backends and models summary
+  void PrintSummary();
```
deadeyegoodwin, line 101: PrintBackendAndModelSummary()

deadeyegoodwin (changes requested):

src/core/server.cc
Tabrizian: Usually, backends do not have a lot of config flags. I can add an if statement to print it only if it is being auto-filled. 

/cc @deadeyegoodwin

deadeyegoodwin: I think @CoderHam  was confusing the backend config with model config. I think what you have it ok.

deadeyegoodwin:
@Tabrizian  There is also a cuda pool you should include in the options table

src/core/server.cc
CoderHam: That's right I was mistaken. Marking this as resolved.

Tabrizian:
@deadeyegoodwin CUDA pool will be shown if there is anything inside the map. Currently, the map is empty.

src/backends/backend/triton_backend_manager.h
Tabrizian: I removed exposing the Singleton instance altogether. Marking this as resolved.

Tabrizian:
My bad, it seems like I was using the wrong flag. I fixed the CUDA memory pool.

Tabrizian:
image
 Above, is the screenshot of the latest version of the backend. If the screen size is small, the table will look like below:

image

src/backends/backend/triton_backend_manager.cc
```diff
@@ -356,7 +360,19 @@ TritonBackendManager::CreateBackend(
       name, dir, libpath, backend_cmdline_config, backend));
   singleton_manager.backend_map_.insert({libpath, *backend});
 
+  const char* backend_config;
```
deadeyegoodwin, line 363: What about when a backend unloads? Don't try to update the BackedState incrementally, just calculate it each time BackendState() is called.

src/backends/backend/triton_backend_manager.h
```diff
@@ -143,9 +143,14 @@ class TritonBackendManager {
       const BackendCmdlineConfig& backend_cmdline_config,
       std::shared_ptr<TritonBackend>* backend);
 
+  static const std::unordered_map<std::string, std::vector<std::string>>&
+  BackendState();
+
  private:
-  std::mutex mu_;
```
deadeyegoodwin, line 147: Why did you change this mutex name? As a convention we typically use mu_. Can you put it back like it was?

src/backends/backend/triton_backend_manager.h
```diff
@@ -143,9 +143,14 @@ class TritonBackendManager {
...
+  static std::unordered_map<std::string, std::vector<std::string>>
+      backend_state_;
```
deadeyegoodwin, line 151: Don't need this.

src/backends/backend/triton_backend_manager.cc
```diff
@@ -356,7 +360,19 @@ TritonBackendManager::CreateBackend(
...
+  backend_state_.insert(
+      {name, std::vector<std::string>{libpath, backend_config}});
+
   return Status::Success;
 }
 
+const std::unordered_map<std::string, std::vector<std::string>>&
+TritonBackendManager::BackendState()
+{
+  return TritonBackendManager::backend_state_;
```
deadeyegoodwin, line 375: Create backend-state here using backend_map_. Be sure to lock mutex.

src/core/model_repository_manager.cc
```diff
@@ -699,7 +699,7 @@ ModelRepositoryManager::BackendLifeCycle::Load(
 
   switch (backend_info->state_) {
     case ModelReadyState::READY:
-      LOG_INFO << "re-loading: " << model_name << ":" << version;
+      LOG_VERBOSE(1) << "re-loading: " << model_name << ":" << version;
```
deadeyegoodwin, line 702: Revert these logging changes

deadeyegoodwin (changes requested):

Tabrizian:
@deadeyegoodwin I had to change TritonBackendManager singleton a little bit to make the instance available to the internal static methods (e.g. `BackendState`). I changed the visibility of all those methods to ensure that it doesn't break the encapsulations.

src/backends/backend/triton_backend_manager.cc
```diff
@@ -327,14 +327,20 @@ TRITONBACKEND_BackendSetState(TRITONBACKEND_Backend* backend, void* state)
 //
 // TritonBackendManager
 //
+TritonBackendManager&
+TritonBackendManager::Instance_()
```
deadeyegoodwin, line 331: Singleton(). Putting _ at the end of a function name is not conforming to our coding standards.

src/backends/backend/triton_backend_manager.cc
```diff
@@ -359,4 +365,31 @@ TritonBackendManager::CreateBackend(
   return Status::Success;
 }
 
+std::unique_ptr<std::unordered_map<std::string, std::vector<std::string>>>
+TritonBackendManager::BackendState()
```
deadeyegoodwin, line 369: Convention would be to return Status from function and add std::unique_ptr<...>* as argument which you then initialize with the map.

src/backends/backend/triton_backend_manager.h
```diff
@@ -143,9 +144,19 @@ class TritonBackendManager {
       const BackendCmdlineConfig& backend_cmdline_config,
       std::shared_ptr<TritonBackend>* backend);
 
+  static std::unique_ptr<
+      std::unordered_map<std::string, std::vector<std::string>>>
+  BackendState();
+
+  TritonBackendManager(){};
```
deadeyegoodwin, line 151: Why is this needed as it is empty? It it is needed add it in the private section as:
TritonBackendManager() = default

src/backends/backend/triton_backend_manager.h
```diff
@@ -143,9 +144,19 @@ class TritonBackendManager {
...
+
  private:
+  static std::unordered_map<std::string, std::vector<std::string>>
+      backend_state_;
+  static TritonBackendManager& Instance_();
   std::mutex mu_;
   std::unordered_map<std::string, std::weak_ptr<TritonBackend>> backend_map_;
+  DISALLOW_COPY_AND_ASSIGN(TritonBackendManager);
```
deadeyegoodwin, line 159: Put this right after "private:"

src/backends/backend/triton_backend_manager.h
```diff
@@ -143,9 +144,19 @@ class TritonBackendManager {
...
       std::shared_ptr<TritonBackend>* backend);
 
+  static std::unique_ptr<
+      std::unordered_map<std::string, std::vector<std::string>>>
+  BackendState();
+
+  TritonBackendManager(){};
+
  private:
+  static std::unordered_map<std::string, std::vector<std::string>>
```
deadeyegoodwin, line 154: Should not be needed. Is it even used?

deadeyegoodwin (changes requested):

src/backends/backend/triton_backend_manager.h
Tabrizian: If I don't add this, the deleted copy constructor will be picked up as the default constructor instead of this constructor.

deadeyegoodwin (approved):

PR #2056, opened 2020-09-23
Renaming perf_client to perf_analyzer

tanmayv25:
Creates a symbolic link named `perf_client` pointing to `perf_analyzer` executable. Have `L0_perf_analyzer` tests directly the perf_analyzer binary while others pass through the soft link `perf_client`.

PIP wheels does not support packaging symbolic links. It will dereference the link and store the copy. So, the package will install duplicate binaries called perf_analyzer and perf_client. See the relevant discussion here:  https://discuss.python.org/t/symbolic-links-in-wheels/1945
If it's ok to break the backwards compatibility, we can only include the `perf_analyzer`  binary within the package.


deadeyegoodwin:
We should not break backwards compat. It is ok to have perf_analyzer be a copy if pp whl cannot handle the symlink.

deadeyegoodwin (approved):

PR #2061, opened 2020-09-25
ONNX Runtime C API test template

CoderHam:


src/test/fw_c_test_templates/onnxruntime/onnxruntime_c_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  const OrtMemoryInfo* allocator_info;
+  CheckStatus(ort_api->AllocatorGetInfo(allocator, &allocator_info));
+
+  // This model has only 2 inputs and 1 output node.
+  // input__0: {4096, 13} FP16
+  // input__1: {4096, 26} INT64
+  // output__0: {4096, 1} FP16
+
+  // Use OrtSession to get metadata of inputs and outputs if not known
+  std::vector<const char*> input_names = {"input__0", "input__1"};
```
tanmayv25, line 120: Shouldn't we derive these parameters directly from the ONNX model itself?

src/test/fw_c_test_templates/onnxruntime/onnxruntime_c_test.cc
```diff
@@ -0,0 +1,171 @@
...
+  // Add in your own validation script for outputs.
+
+  ort_api->ReleaseValue(input_tensors[0]);
+  ort_api->ReleaseValue(input_tensors[1]);
+  ort_api->ReleaseSession(session);
+  ort_api->ReleaseSessionOptions(session_options);
+  ort_api->ReleaseEnv(env);
+  printf("Ran Successfully!\n");
+  return 0;
+}
```
tanmayv25, line 171: newline

tanmayv25 (changes requested):

src/test/fw_c_test_templates/onnxruntime/onnxruntime_c_test.cc
CoderHam: This is just a template. I could make it generic and allow it to run for any kind of model but for now I've only designed it for the DLRM model.

GuanLuo:
Why adding an ORT test on our end? If it is just for testing the model execution on ORT natively, why not using `onnxruntime_perf_test` executable? Is that not sufficient?

CoderHam:
> Why adding an ORT test on our end? If it is just for testing the model execution on ORT natively, why not using `onnxruntime_perf_test` executable? Is that not sufficient?

I wasn't aware that this executable exists. Do we document this somewhere?
I agree that if this is a generic perf maintained by ORT then we can remove our template and just add the documentation.

CoderHam:
@GuanLuo when I run the `onnxruntime_perf_test` I get the following error. The error is not descriptive enough to understand the cause of the failure.

@deadeyegoodwin I think we should still have a template of this sort to allow users to re-create the same scenario and see the same error they would from inside Tritonserver. What do you think?

```
2020-09-29 00:28:57.403046922 [I:onnxruntime:, inference_session.cc:174 ConstructorCommon] Creating and using per session threadpools since use_per_session_threads_ is true
2020-09-29 00:28:57.403811976 [I:onnxruntime:, inference_session.cc:840 Initialize] Initializing session.
2020-09-29 00:28:57.403828696 [I:onnxruntime:, inference_session.cc:865 Initialize] Adding default CPU execution provider.
2020-09-29 00:28:57.403837091 [I:onnxruntime:Default, bfc_arena.cc:18 BFCArena] Creating BFCArena for Cpu
2020-09-29 00:28:57.403846564 [V:onnxruntime:Default, bfc_arena.cc:35 BFCArena] Creating 21 bins of max chunk size 256 to 268435456
2020-09-29 00:28:57.405304630 [I:onnxruntime:, reshape_fusion.cc:33 ApplyImpl] Fused reshape node: 60
2020-09-29 00:28:57.405323173 [I:onnxruntime:, reshape_fusion.cc:37 ApplyImpl] Total fused reshape node count: 1
2020-09-29 00:28:57.405584709 [I:onnxruntime:, graph.cc:2712 CleanUnusedInitializers] Removing initializer '54'. It is no longer used by any node.
2020-09-29 00:28:57.405918606 [I:onnxruntime:, reshape_fusion.cc:37 ApplyImpl] Total fused reshape node count: 0
2020-09-29 00:28:57.406138529 [I:onnxruntime:, reshape_fusion.cc:37 ApplyImpl] Total fused reshape node count: 0
2020-09-29 00:28:57.407910763 [V:onnxruntime:, inference_session.cc:679 TransformGraph] Node placements
2020-09-29 00:28:57.407925278 [V:onnxruntime:, inference_session.cc:681 TransformGraph] All nodes have been placed on [CPUExecutionProvider].
2020-09-29 00:28:57.407949259 [V:onnxruntime:, session_state.cc:71 CreateGraphInfo] SaveMLValueNameIndexMapping
2020-09-29 00:28:57.407989631 [V:onnxruntime:, session_state.cc:116 CreateGraphInfo] Done saving OrtValue mappings.
2020-09-29 00:28:57.409158296 [I:onnxruntime:, finalize_session_state.cc:173 SaveInitializedTensors] Saving initialized tensors.
2020-09-29 00:28:57.409304786 [I:onnxruntime:, finalize_session_state.cc:225 SaveInitializedTensors] Done saving initialized tensors
2020-09-29 00:28:57.410017485 [I:onnxruntime:, inference_session.cc:954 Initialize] Session successfully initialized.
there is no test data for model model.onnx
Run failed:failed to initialize.
```

GuanLuo:
Does the model contain dynamic dimensions? Try add '-I' so that it will set dynamic dimensions if any.

I agree that the error is not informative, but I don't think Triton should create and maintain such template as it is closely tied to ORT, it would be great if we can use utility maintained by ORT, like the perf test above, and we would raise issue to them if it needs to be improved.

CoderHam:
> Does the model contain dynamic dimensions? Try add '-I' so that it will set dynamic dimensions if any.

No, the dimensions were not dynamic. The version of DLRM I used works on fixed size inputs. 

Tasking this to onnx may be a better idea and save us the time and effort of maintaining the template.

GuanLuo:
It's a non-batching model?

CoderHam:
> It's a non-batching model?

Yes it is a non-batching model.

GuanLuo:
Okay, I was thinking that the 4096 you used is for batch dimension

CoderHam:
Closing this ticket for now. I will communicate the limited logging for the onnxruntime_perf_test to ONNX Runtime

PR #2062, opened 2020-09-26
Allow optional base path in http url

tanmayv25:
Picks major implementation from PR #2047
Adds simple test cases for HTTP clients and documents the optional feature.

### DIFF
--- a/src/clients/python/library/tritonclient/http/__init__.py
+++ b/src/clients/python/library/tritonclient/http/__init__.py
@@ -138,7 +138,10 @@ class InferenceServerClient:
     Parameters
     ----------
     url : str
-        The inference server URL, e.g. 'localhost:8000'.
+        The inference server name, port and optional base path 
+        in the following format: host:port/<base-path>, e.g.
+        'localhost:8000'.
+
     verbose : bool
         If True generate verbose output. Default value is False.
     concurrency : int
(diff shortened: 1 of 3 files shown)


src/clients/python/library/tritonclient/http/__init__.py
```diff
@@ -189,8 +192,11 @@ def __init__(self,
                  ssl_options=None,
                  ssl_context_factory=None,
                  insecure=False):
-        scheme = "https://" if ssl else "http://"
-        self._parsed_url = URL(scheme + url)
+        if not url.startswith("http://") and not url.startswith("https://"):
```
deadeyegoodwin, line 195: Did you mean to add this? So putting the scheme in the URL overrides the 'ssl' setting? We don't document that override and we don't document that you can put the scheme in the url. Is the behavior the same as in C++? It seems like we should instead check to make sure that the URL does not have any *:// prefix and raise an error if it does.

tanmayv25: Good point. I have added a check to raise an error upon detecting the scheme in the url. 
I also found that we are missing the full-fledged SSL support in our C++ http client library. But when added would do something similar. I have added a ticket to support the certificate authentication in C++.

deadeyegoodwin (approved):

PR #2065, opened 2020-09-28
Add test for non-graceful Python backend exit

Tabrizian:


qa/L0_backend_python/test.sh
```diff
@@ -103,6 +103,29 @@ set -e
...
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+# Trigger non-graceful termination of Triton
+kill -9 $SERVER_PID
+
+# Wait 10 seconds so that Python gRPC server can detect non-graceful exit
```
CoderHam, line 118: Wouldn't this be dependent on `--exit-timeout-secs`? The default is 30 seconds I believe.

Tabrizian: It is different from that. I killed the Triton with `-9` so it will exit immediately. This is dependent on the lines below of the python_backend. It will be 6 seconds maximum, setting it to 10 to give it some margin.

https://github.com/triton-inference-server/python_backend/commit/57fa7abf497bef199071f9665cb6a879139a1a4a#diff-360c0054757ac9c0b4a774839a0b76f3R300-R333

deadeyegoodwin (commented):
Did you confirm that this test failed before you added the graceful exit change

CoderHam:
David is right. The CI running for branch does not include the L0_backend_python test. Could you re-run the test with the TRITON_NIGHTLY flag?

Tabrizian:
@deadeyegoodwin Sorry for the delay. I tried it locally and the test fails. @CoderHam Sure, I'll set the nightly flag so that we can see the test passing in the CI.

deadeyegoodwin (approved):

PR #2078, opened 2020-10-02
Update ONNX Runtime version to 1.5.1

GuanLuo:
Change build steps to use https://github.com/microsoft/onnxruntime/tree/v1.5.1/dockerfiles as reference instead of https://github.com/microsoft/onnxruntime/tree/master/tools/ci_build/github/linux/docker because the latter is configured for ORT's CI environment

### DIFF
--- a/build.py
+++ b/build.py
@@ -335,6 +335,9 @@ WORKDIR /workspace
 ENV PATH /usr/local/nvidia/bin:/usr/local/cuda/bin:/workspace/cmake-3.14.3-Linux-x86_64/bin:/opt/miniconda/bin:$PATH
 ENV LD_LIBRARY_PATH /opt/miniconda/lib:/usr/lib:/usr/lib/x86_64-linux-gnu:$LD_LIBRARY_PATH
 
+# The Onnx Runtime dockerfile is the collection of steps in 
+# https://github.com/microsoft/onnxruntime/tree/v1.5.1/dockerfiles
+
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -50,8 +50,8 @@
 # Map from container version to corresponding component versions
 # container-version -> (ort version, ort openvino version)
 CONTAINER_VERSION_MAP = {
-    '20.08': ('1.4.0', '2020.2'),
-    '20.09': ('1.4.0', '2020.2')
+    '20.08': ('1.5.1', '2020.4'),
+    '20.09': ('1.5.1', '2020.4')
```
deadeyegoodwin, line 54: We didn't change what was used in 20.08 and 20.09 so don't change these. Instead add an entry for 20.10 with the new onnx versions. Perhaps this is just for the CI run...

deadeyegoodwin (changes requested):

build.py
GuanLuo: The issue is that there is mismatch between the dev version and the container version, i..e. this change targets on 20.10 but we are using 20.09 container. I think for now, we should change it for CI run, but then correct it on the release branch. 

In addition to that, the change in 20.08 is because current master is still using 20.08 due to the late 20.09 release.. We can change it once the master is updated to use 20.09 container

deadeyegoodwin (approved):

PR #2079, opened 2020-10-02
Add reformat-free input model generation script. Add reformat-free input model test

GuanLuo:
@tanmayv25 Changing data byte size calculation for reformat-free tensor to be first round-up the vectorized dimension, and then calculate the byte size with the new shape. The reasons of not using `getStride()` are:
-  it doesn't work in vectorized dim == 0 case, round-up is needed for getting the actual size requirements.
- it may not return properly unless all input shapes are updated. i.e. for add-sub dynamic model, the element-wise op requires all input shapes are consistent for `getStride()` to return proper values. But currently the backend handles inputs one by one: `set input0 dimension -> get input0 strides -> set input1 dimension -> get input1 strides`, so `getStride()` doesn't work unless we change the backend to `set all input shapes -> get all input strides`.

### DIFF
--- a/qa/L0_trt_reformat_free/trt_reformat_free_test.py
+++ b/qa/L0_trt_reformat_free/trt_reformat_free_test.py
@@ -1,0 +1,153 @@
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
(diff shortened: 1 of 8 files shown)


qa/common/gen_qa_trt_format_models.py
```diff
@@ -0,0 +1,432 @@
...
+    # Fixed shape
+    create_plan_model(FLAGS.models_dir, 0, 1, (13, 2, 1), (13, 2, 1),
+                      (13, 2, 1), np.float16, np.float16, np.float16,
+                      trt.TensorFormat.CHW2, trt.TensorFormat.LINEAR)
+    create_plan_model(FLAGS.models_dir, 8, 1, (13, 2, 1), (13, 2, 1),
+                      (13, 2, 1), np.float16, np.float16, np.float16,
+                      trt.TensorFormat.CHW2, trt.TensorFormat.LINEAR)
+
+    # Dynamic shape
+    # FIXME The generated model inputs are LINEAR format instead of CHW32 format
```
tanmayv25, line 423: Is this comment still valid? I see L117 and L118 will get trt.TensorFormat.CHW32.

qa/common/gen_qa_trt_format_models.py
```diff
@@ -0,0 +1,432 @@
...
+    create_plan_model(FLAGS.models_dir, 0, 1, (-1, 2, 1), (-1, 2, 1),
+                      (-1, 2, 1), np.float32, np.float32, np.float32,
+                      trt.TensorFormat.CHW32, trt.TensorFormat.LINEAR)
```
tanmayv25, line 426: I remember you mentioning something about not using non-linear formats in our testing, but the reason is slipping my mind. Can you explain why we test only non-linear formats in input?

qa/L0_trt_reformat_free/trt_reformat_free_test.py
```diff
@@ -0,0 +1,271 @@
...
+    def test_chw32_input(self):
+        model_name = "plan_CHW32_LINEAR_float32_float32_float32"
+        for bs in [1, 8]:
+            input_np = np.arange(26 * bs, dtype=np.float32).reshape(
+                (bs, 13, 2, 1))
+            expected_output0_np = input_np + input_np
+            expected_output1_np = input_np - input_np
+            reformatted_input_np = reformat("CHW32", input_np)
+
+            # Use shared memory to bypass the shape check in client library,
```
tanmayv25, line 232: Do you think this warrants another API to set tensor data that skips data byte sizes checks? We must not force users to use shared memory if they want to use reformat-free models. We can do this as another PR.

tanmayv25 (commented):
> it may not return properly unless all input shapes are updated. i.e. for add-sub dynamic model, the element-wise op requires all input shapes are consistent for getStride() to return proper values. But currently the backend handles inputs one by one: set input0 dimension -> get input0 strides -> set input1 dimension -> get input1 strides, so getStride() doesn't work unless we change the backend to set all input shapes -> get all input strides.

I see... I think we should be fine with your changes.

qa/common/gen_qa_trt_format_models.py
GuanLuo: The output handling is not implemented, currently the responder is calculating the model output offset by output shape, but in the format case, the shape doesn't necessarily map to actual byte size of the output buffer

qa/common/gen_qa_trt_format_models.py
GuanLuo: Yes, the corresponding test doesn't manipulate the input as the input format is actually linear.

qa/L0_trt_reformat_free/trt_reformat_free_test.py
GuanLuo: Yes, I think that will be good. We can discuss about how to achieve it, in addition to it, perhaps we should extend the format field in model config to include the TRT formats, so the user may inquire the input format and preprocess the data accordingly if needed. Currently we assume that the user knows what the model is expecting beforehand.

qa/common/gen_qa_trt_format_models.py
GuanLuo: DLIS-1661

qa/L0_trt_reformat_free/trt_reformat_free_test.py
GuanLuo: DLIS-1662

deadeyegoodwin (approved):

PR #2087, opened 2020-10-05
Add flags to enable TF2 infer/batcher test

CoderHam:


qa/common/sequence_util.py
```diff
@@ -51,8 +51,7 @@
     int(os.environ.get('TEST_SYSTEM_SHARED_MEMORY', 0)))
 _test_cuda_shared_memory = bool(
     int(os.environ.get('TEST_CUDA_SHARED_MEMORY', 0)))
-_test_valgrind = bool(
-    int(os.environ.get('TEST_VALGRIND', 0)))
+_test_valgrind = bool(int(os.environ.get('TEST_VALGRIND', 0)))
```
deadeyegoodwin, line 54: Why this change?

deadeyegoodwin:
What about L0_sequence_batcher?

qa/common/sequence_util.py
CoderHam: Picked up by formatter.

CoderHam:
> What about L0_sequence_batcher?

The sequence batcher is still failing on Jetson. I can add support for enabling TF2 testing on x86 but I want to hold off enabling it on jetson till the timing related failures are resolved on Jetson.

qa/common/sequence_util.py
deadeyegoodwin: ok

deadeyegoodwin:
I would add it as part of this change. You don't need to add a jetson test that uses it until it is working. For x86 I filed a separate ticket and for x86 we just want to run L0_infer.

CoderHam:
> I would add it as part of this change. You don't need to add a jetson test that uses it until it is working. For x86 I filed a separate ticket and for x86 we just want to run L0_infer.

Done

deadeyegoodwin (approved):

PR #2089, opened 2020-10-06
Add shared libraries needed for ONNX Runtime TRT / OpenVINO accelerator

GuanLuo:
However, the tests that involves TRT accelerator will still be failing due to the program can't exit successfully, see https://github.com/microsoft/onnxruntime/issues/5384 for detail


GuanLuo:
Closing, will try ORT 1.5.1 again once it is fixed in upstream

PR #2093, opened 2020-10-06
Multi-output ensemble client

mvpel:
The existing ensemble_image_client.py code is only able to collect a single output from the model. This change allows the client to request, retrieve, and display multiple outputs from a model, and displays the outputs for each input file separately in an orderly list.

In addition, the "-m / --model-name" argument is added to allow the user to specify the model on the command line, rather than hardcoding it. It defaults to the original hardcoded model name of "preprocess_resnet50_ensemble" to insure backwards compatibility is maintained.


CoderHam:
The ensemble image client serves as an example. Though this PR makes sense the point of the example is to be as simple as possible. Thus I would prefer not complicating the ensemble_image_client.py further unless necessary.

PR #2096, opened 2020-10-06
Add common and core REPO_TAGs to Dockerfile.client and Dockerfile.QA

CoderHam:


Dockerfile.client
```diff
@@ -75,9 +75,13 @@ COPY build build
 COPY src/clients src/clients
 COPY src/core src/core
 
+TRITON_COMMON_REPO_TAG=main
```
deadeyegoodwin, line 78: You need to make these 

ARG TRITON_CORE_REPO_TAG=main

at the top of the file, and then here have:

ARG TRITON_COMMON_REPO_TAG
ARG TRITON_CORE_REPO_TAG

Then in .gitlabci-yml you can use --build-arg to set them appropriately on the docker build invocation so that they get set to whatever is configuration for the CI run.

deadeyegoodwin (changes requested):

Dockerfile.client
CoderHam: Done.

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #2114, opened 2020-10-13
Fixes for backend loading via dlopen

deadeyegoodwin:
Compile core to export server and backend API symbols
Use LOCAL for dlopen to avoid exposing symbols.

### DIFF
--- a/src/core/tritonserver.cc
+++ b/src/core/tritonserver.cc
@@ -24,8 +24,6 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#include "triton/core/tritonserver.h"
-
 #include <string>
 #include <vector>
 #include "src/core/backend.h"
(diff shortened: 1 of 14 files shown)


src/core/tritonserver.cc
```diff
@@ -43,6 +41,7 @@
 #include "src/core/server.h"
 #include "src/core/server_message.h"
 #include "src/core/status.h"
+#include "src/core/tritonserver_apis.h"
```
CoderHam, line 44: Why are we creating a new file instead of using tritonserver.h? Files that do not need tritonbackend.h end up including it as well.

deadeyegoodwin: tritonserver.h (and tritonbackend.h) are used with different declations (either import or export of the api). tritonserver_apis.h sets defines that cause the API to be exported from libtritonserver.so. Direct uses of tritonserver.h and tritonbackend.h are normal client uses that "import" the API. All uses to the headers should use tritonserver_api.h with "core" triton. I combined them into 1 header for convenience.

CoderHam (approved):
These changes look good. I see a win10 client build failure that is likely caused by changes outside this PR, could you have a look to see what's wrong?

GuanLuo (approved):

PR #2117, opened 2020-10-14
Avoid downtime by unloading previous version only after load. Fix bug

GuanLuo:


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -1425,6 +1469,121 @@ def test_model_control(self):
...
+            self.assertTrue(False, "unexpected error {}".format(ex))
+
+        # Make sure 3 versions of the model are loaded
+        try:
+            triton_client = httpclient.InferenceServerClient("localhost:8000",
+                                                             verbose=True)
+            self.assertTrue(triton_client.is_server_live())
+            self.assertTrue(triton_client.is_server_ready())
+            self.assertTrue(triton_client.is_model_ready(model_name, "1"))
+            self.assertTrue(triton_client.is_model_ready(model_name, "3"))
```
deadeyegoodwin, line 1513: Can you explictly check that the platform for the newly loaded model is correct

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -1425,6 +1469,121 @@ def test_model_control(self):
...
+        # Reload models, v1 should still be available until v2 is loaded
+        # The load is requested in other thread as it is blocking API,
+        # and the v1 availibility should be tested during the reload
+        thread = threading.Thread(target=self._async_load, args=(model_name,))
+        thread.start()
+        # wait for time < model creation delay to ensure load request is sent
+        time.sleep(3)
+        # DLIS-1698
+        # A dummy request to any endpoint is needed as the first request after
+        # the async loading above will be blocked until the load returns
```
deadeyegoodwin, line 1551: Put DLIS-1698 here and other places where you needed to workaround that issue.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -1425,6 +1469,121 @@ def test_model_control(self):
...
+            1,
+            3,
+        ), model_shape)
+
+    def test_model_availability_on_reload(self):
+        model_name = "identity_zero_1_int32"
+        model_base = "identity"
+        model_shape = (16,)
+
+        # Make sure version 1 of the model is loaded
```
deadeyegoodwin, line 1528: Can we have another test like this where model have both 1/ and 2/ directories but the version policy only exposes 1. Then change version policy to only expose 2 and make sure that works correctly.

src/core/model_repository_manager.cc
```diff
@@ -589,47 +596,120 @@ Status
...
+            std::lock_guard<std::recursive_mutex> lock(unload_backend->mtx_);
+            unload_backend->next_action_ = ActionType::UNLOAD;
+            unload_backend->OnComplete_ = [this, unload_backend]() {
+              std::lock_guard<std::recursive_mutex> map_lock(map_mtx_);
+              unloading_backends_.erase((uintptr_t)unload_backend);
+            };
+            TriggerNextAction(model_name, version, unload_backend);
+          }
+
+          // Check if all versions are completed and perform deferred unload
```
deadeyegoodwin, line 694: What happens if some or all of the loaded/loading model versions fail? Do we just cancel the unload? I think we should.

src/core/model_repository_manager.cc
```diff
@@ -256,12 +256,13 @@ class ModelRepositoryManager::BackendLifeCycle {
...
-  // If 'force_unload', all versions that are being served will
-  // be unloaded before loading the specified versions.
+  // If 'defer_unload' is false, all versions that are being served will
+  // be unloaded before loading the specified versions. Otherwise, the versions
+  // not specified in the load will be unloaded after the load is finished.
   Status AsyncLoad(
       const std::string& repository_path, const std::string& model_name,
       const std::set<int64_t>& versions,
-      const inference::ModelConfig& model_config, bool force_unload = true,
+      const inference::ModelConfig& model_config, bool defer_unload = false,
```
deadeyegoodwin, line 265: So when we call AsyncLoad we explicitly set defer_unload to true?

deadeyegoodwin (changes requested):

src/core/model_repository_manager.cc
GuanLuo: I agree, maybe the unload should be cancelled only if all loading versions fail as the rollback of the whole load request

src/core/model_repository_manager.cc
GuanLuo: Yes

src/core/model_repository_manager.cc
deadeyegoodwin: I think it is more intuitive to abort the load unless all versions load correctly. Can you file a ticket.

GuanLuo: Created DLIS-1707. But one thing about aborting the load is that the reason of failure is not going to be shown in model index, so user will need to look at the server log to figure out what is the issue with the model.

deadeyegoodwin: But the response to the load request will show the error, correct?

GuanLuo: True, but that just mentions what versions are not loaded, so we should add more detail to the error message?

deadeyegoodwin (approved):

PR #2119, opened 2020-10-14
Rate Limiter for cross-model prioritization and resource constraining

tanmayv25:
Note: This is just a rate limiter unit implementation and it is not connected to the server yet.

I will add the extensive testing with other PR.

The logs of the rate_limter_test covering very basic functionalities from my machine

```
[==========] Running 7 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 7 tests from RateLimiterTest
[ RUN      ] RateLimiterTest.SingleInstanceSingleInfer
[       OK ] RateLimiterTest.SingleInstanceSingleInfer (0 ms)
[ RUN      ] RateLimiterTest.SingleInstanceMultiInfer
[       OK ] RateLimiterTest.SingleInstanceMultiInfer (0 ms)
[ RUN      ] RateLimiterTest.MultiInstanceMultiInfer
[       OK ] RateLimiterTest.MultiInstanceMultiInfer (0 ms)
[ RUN      ] RateLimiterTest.SpecificInstanceMultiInfer
[       OK ] RateLimiterTest.SpecificInstanceMultiInfer (1 ms)
[ RUN      ] RateLimiterTest.SimplePriority
[       OK ] RateLimiterTest.SimplePriority (0 ms)
[ RUN      ] RateLimiterTest.SimpleResource
[       OK ] RateLimiterTest.SimpleResource (0 ms)
[ RUN      ] RateLimiterTest.NoLimiting
[       OK ] RateLimiterTest.NoLimiting (0 ms)
[----------] 7 tests from RateLimiterTest (1 ms total)

[----------] Global test environment tear-down
[==========] 7 tests from 1 test suite ran. (1 ms total)
[  PASSED  ] 7 tests.
```

### DIFF
--- a/src/core/model_config.proto
+++ b/src/core/model_config.proto
@@ -146,6 +146,65 @@ message ModelInstanceGroup
   //@@     of instances created. Default is 1.
   int32 count = 2;
 
+  //@@
+  //@@  .. cpp:var:: message RateLimiterSpec
+  //@@
+  //@@     The specifications required by the rate limiter to properly
+  //@@     schedule the inference requests across the different models
+  //@@     and their instances.
+  //@@
+  message RateLimiterSpec
+  {
+    //@@  .. cpp:var:: message Resource
+    //@@
+    //@@     The resoure constraints required to execute an inference
+    //@@     on the instance of the model.
+    //@@
+    message Resource
+    {
+      //@@  .. cpp:var:: message global
+      //@@
+      //@@     Whether or not the resource is global. If true then the resource
+      //@@     is assumed to be shared among the GPU devices otherwise specified
+      //@@     count of the resource is assumed for each GPU device associated
+      //@@     with the instance.
+      //@@
+      bool global = 1;
+
+      //@@  .. cpp:var:: message count
+      //@@
+      //@@     The number of resources required for the execution of the model
+      //@@     instance.
+      //@@
+      uint32 count = 2;
+    }
+
+    //@@  .. cpp:var:: map<string, ResourceCount> resources
+    //@@
+    //@@     The resources required to execute the request on a model instance.
+    //@@     Resources are just names with a corresponding count. The execution
+    //@@     of the instance will be blocked until the specificied resources are
+    //@@     available. By default an instance uses no rate-limiter resources.
+    //@@
+    map<string, Resource> resources = 1;
+
+    //@@  .. cpp:var:: uint32 priority
+    //@@
+    //@@     The weighting value to be used for prioritizing across instances.
+    //@@     An instance with priority 2 will be given 1/2 the number of
+    //@@     scheduling chances as an instance_group with priority 1. The
+    //@@     default priority is 1.
+    //@@
+    uint32 priority = 2;
+  }
+
+  //@@  .. cpp:var:: RateLimiterSpec rate_limiter_spec
+  //@@
+  //@@     The rate limiter specific settings to be associated with this
+  //@@     instance group.
+  //@@
+  RateLimiterSpec rate_limiter_spec = 6;
+
   //@@  .. cpp:var:: int32 gpus (repeated)
   //@@
   //@@     GPU(s) where instances should be available. For each GPU listed,
--- a/src/core/model_config.proto
+++ b/src/core/model_config.proto
@@ -74,6 +74,63 @@ enum DataType {
 }
 
 //@@
+//@@  .. cpp:var:: message ModelRateLimiter
+//@@
+//@@     The specifications required by the rate limiter to properly
+//@@     schedule the inference requests across the different models
+//@@     and their instances.
+//@@
+message ModelRateLimiter
+{
+  //@@  .. cpp:var:: message Resource
+  //@@
+  //@@     The resource property.
+  //@@
+  message Resource
+  {
+    //@@  .. cpp:var:: string name
+    //@@
+    //@@     The name associated with the resource.
+    //@@
+    string name = 1;
+
+    //@@  .. cpp:var:: bool global
+    //@@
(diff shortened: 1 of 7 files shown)


src/core/rate_limiter.h
```diff
@@ -0,0 +1,248 @@
...
+  void LoadModelHelper(
+      const std::string& model_name, const int64_t version, const int device_id,
+      const Specification& rate_limit_spec, ModelContext* model_context,
+      std::vector<std::shared_ptr<ModelInstance>>* model_instances);
+
+  void OnStage(ModelInstance* instance_ptr);
+  void OnRelease(ModelInstance* instance_ptr);
+  void AttemptAllocation();
+
+  class Comparator {
```
deadeyegoodwin, line 157: ScaledPriorityComparator

deadeyegoodwin (changes requested):

src/core/rate_limiter.h
```diff
@@ -0,0 +1,248 @@
...
+#include "src/core/status.h"
+
+namespace nvidia { namespace inferenceserver {
+
+// Limits the rate at which requests are dispatched from the scheduler
+class RateLimiter {
+ public:
+  class ModelInstance;
+
+ private:
```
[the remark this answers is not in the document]
tanmayv25: ModelContext(private) is required by the ModelInstance(public) below.

src/core/model_config.proto
```diff
@@ -146,6 +203,13 @@ message ModelInstanceGroup
   //@@     of instances created. Default is 1.
   int32 count = 2;
 
+  //@@  .. cpp:var:: ModelRateLimiter rate_limiter
+  //@@
+  //@@     The rate limiter specific settings to be associated with this
+  //@@     instance group.
```
deadeyegoodwin, line 209: "Optional, if not specified no rate limiting will be applied to this instance group."

src/core/rate_limiter.h
```diff
@@ -0,0 +1,263 @@
...
+  /// will be available -> allocated -> available.
+  /// \param ignore_resources_and_priority Whether or not to ignore resource
+  /// constraints and cross-model priority. An available instance is directly
+  /// allocated when true.
+  /// \return Status object indicating success or failure.
+  static Status Create(
+      const bool ignore_resources_and_priority,
+      std::unique_ptr<RateLimiter>* rate_limiter);
+
+  /// Loads the configified model to the RateLimiter
```
deadeyegoodwin, line 69: "Load"  might be confusing on first read since you are not really loading anything. Perhaps "AddModel"? And the description would be "Add model to the set of models being managed by the rate limiter."

src/core/rate_limiter.h
```diff
@@ -0,0 +1,263 @@
...
+  /// \return Status object indicating success or failure.
+  Status LoadModel(
+      const std::string& model_name, const int64_t version,
+      const inference::ModelConfig& model_config);
+
+  /// Unloads the configified model from the RateLimiter
+  /// \param model_name The name of the model to load.
+  /// \param version The version of the model to load.
+  /// \return Status object indicating success or failure.
+  Status UnloadModel(const std::string& model_name, const int64_t version);
```
deadeyegoodwin, line 82: "RemoveModel"?

src/core/rate_limiter.h
```diff
@@ -0,0 +1,263 @@
...
+
+  /// Enqueues the callback to the configified model. In future, when the
```
deadeyegoodwin, line 84: Not sure what "configified" means or if it is important. And isn't what is being requested is a model instance? So perhaps "RequestModelInstance"?

src/core/rate_limiter.h
```diff
@@ -0,0 +1,263 @@
...
+  /// as a signal to proceed with the execution.
+  /// \param OnSchedule The callback function to be called when scheduling.
+  /// \param model_name The name of the model.
+  /// \param version The version of the model.
+  /// \param instance_index The index to a configific instance of the model.
+  /// The default value is -1 which means that an instance with highest
+  /// priority will be selected for the execution.
+  /// \return Status object indicating success or failure.
+  Status EnqueueModelRequest(
+      const StandardScheduleFunc& OnSchedule, const std::string& model_name,
```
deadeyegoodwin, line 101: Using model name, version, instance index... we have adstractions for TritonModel and TritonModelInstance that we could probably be using directly. This is ok for now but we should revisit when integrating this into scheduler to see how we can use those abstractions.

src/core/rate_limiter.cc
```diff
@@ -0,0 +1,630 @@
...
+{
+  {
+    std::lock_guard<std::mutex> lk1(model_contexts_mtx_);
+    std::lock_guard<std::mutex> lk2(model_instances_mtx_);
+
+    auto& model_context = model_contexts_[std::make_pair(model_name, version)];
+    auto& model_instances =
+        model_instances_[std::make_pair(model_name, version)];
+    {
+      for (const auto& group : model_config.instance_group()) {
```
deadeyegoodwin, line 60: Repeating this code here is no good. We should find a way to use TritonModelInstance abstractions in a later PR.

deadeyegoodwin (changes requested):

src/core/rate_limiter.h
tanmayv25: That is right. I have implemented it like this for simplicity.
In future it will be more like passing raw pointers to TritonModel/TritonModelInstance to add and remove model instances. We can use the API to attain the name, version and instance index. The TritonModelInstance pointer and additional information(RateLimiterConfig) will be wrapped within RateLimiter::ModelInstance class. This way we would not have to disturb the backend API.

deadeyegoodwin (approved):

PR #2124, opened 2020-10-14
Remove Caffe2 backend and relevant tests

CoderHam:
Ensembles that used caffe2 models use graphdef instead. 
In server status and lifecycle use onnx instead of netdef. 
TODO: Follow up changes to doc to replace Caffe2 models in fetch_models.sh and documentation.


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -140,24 +140,23 @@ def test_parse_error_modelfail(self):
 
         # And other models should be loaded successfully
         try:
-            for base_name in ["savedmodel", 'netdef']:
```
deadeyegoodwin, line 143: For all these lifecycle tests it is important the we don't just remove netdef but that we replace it with something else. An onnx model would be best.

qa/L0_lifecycle/test.sh
```diff
@@ -192,9 +192,7 @@ mkdir models models_0
 for i in graphdef savedmodel ; do
     cp -r $DATADIR/qa_model_repository/${i}_float32_float32_float32 models/.
 done
-for i in netdef plan ; do
```
deadeyegoodwin, line 195: don't just remove netdef, replace with onnx

qa/L0_sequence_corrid_batcher/test.sh
```diff
@@ -59,7 +59,6 @@ for m in \
         $DATADIR/qa_dyna_sequence_model_repository/plan_dyna_sequence_int32 \
         $DATADIR/qa_dyna_sequence_model_repository/onnx_dyna_sequence_int32 \
         $DATADIR/qa_dyna_sequence_model_repository/libtorch_dyna_sequence_int32 \
-        $DATADIR/qa_dyna_sequence_model_repository/netdef_dyna_sequence_int32 ; do
```
deadeyegoodwin, line 62: Need to put "; do" on previous line

qa/L0_server_status/server_status_test.py
```diff
@@ -141,160 +141,155 @@ def test_unknown_model_version(self):
...
         tensor_shape = (1, input_size)
-        platform_name = {
-            'graphdef': 'tensorflow_graphdef',
-            'netdef': 'caffe2_netdef'
-        }
 
-        # There are 3 versions of *_int32_int32_int32 and all
+        # There are 3 versions of graphdef_int32_int32_int32 and all
         # should be available.
-        for platform in ('graphdef', 'netdef'):
```
deadeyegoodwin, line 151: For these tests I think it is ok to remove "netdef" without a replacement, but keep the for loop in case we want to add something back later.

deadeyegoodwin (changes requested):

deadeyegoodwin:
@CoderHam  Also change the description of the commit to explicitly say "Remove Caffe2 backend..."

build.py
```diff
@@ -54,7 +54,8 @@
     '20.09': ('20.09', '1.4.0', '2020.2'),  # 2.3.0
     '20.10dev': ('20.09', '1.4.0', '2020.2'),  # 2.4.0dev
     '20.10': ('20.10', '1.4.0', '2020.2'),  # 2.4.0
-    '20.11dev': ('20.09', '1.4.0', '2020.2')  # 2.5.0dev
+    '20.11dev': ('20.09', '1.4.0', '2020.2'),  # 2.5.0dev
+    '20.11dev_nocaffe2': ('20.09', '1.4.0', '2020.2')  # 2.5.0dev_nocaffe2
```
CoderHam, line 58: Note: Will remove `nocaffe2` from both branches are successfully CI

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #2125, opened 2020-10-15
Send larger batch than preferred batch size when queue delay is exceeded

GuanLuo:


qa/L0_batcher/batcher_test.py
```diff
@@ -950,6 +950,78 @@ def test_multi_batch_delayed_sum_gt_max_preferred(self):
...
+        precreated_shm1_regions = self.create_advance(['op10', 'op11'])
+        precreated_shm2_regions = self.create_advance(['op20', 'op21'])
+        for trial in _trials:
+            try:
+                model_name = tu.get_model_name(trial, np.float32, np.float32,
+                                               np.float32)
+
+                self.check_setup(model_name)
+
+                # Need scheduler to wait for queue to contain 2 requests
```
deadeyegoodwin, line 982: 3 requests?

qa/L0_batcher/batcher_test.py
```diff
@@ -950,6 +950,78 @@ def test_multi_batch_delayed_sum_gt_max_preferred(self):
             except Exception as ex:
                 self.assertTrue(False, "unexpected error {}".format(ex))
 
+    def test_multi_batch_delayed_use_max_batch(self):
```
deadeyegoodwin, line 953: So max_batch_size is 8, and preferred size is 2 and 6.
The first 2 requests are sent for total batch size of 7, since that doesn't match preferred they just sit
But then we wait 11 seconds which should cause timeout to expire... so why are not both request processed at that point?

deadeyegoodwin (changes requested):

qa/L0_batcher/batcher_test.py
GuanLuo: The sum of 2 requests is larger than any preferred batches so the first requests will be send separately if the second's timeout is not expired. But after we wait 11 seconds, some of the requests that can fit in the max batch have expired timeout and thus we send all requests fit in the max batch at once.

qa/L0_batcher/batcher_test.py
GuanLuo: yes, typo..

deadeyegoodwin (approved):

PR #2131, opened 2020-10-16
Monitor only devices visible to CUDA

Tabrizian:


src/core/metrics.cc
```diff
@@ -172,20 +172,42 @@ Metrics::InitializeNvmlMetrics()
...
+      LOG_ERROR << "failed to get PCI Bus ID for CUDA device " << didx
+                << ": " << cudaGetErrorString(cuerr);
+      return false;
+    }
+
     nvmlDevice_t gpu;
-    nvmlReturn_t nvmlerr = nvmlDeviceGetHandleByIndex(didx, &gpu);
+    nvmlReturn_t nvmlerr = nvmlDeviceGetHandleByPciBusId_v2(pcibusid_str, &gpu);
+    if (nvmlerr != NVML_SUCCESS) {
+      LOG_ERROR << "failed to get device from PCI Bus ID: NVML_ERROR " << nvmlerr;
```
CoderHam, line 207: You can remove the NVML_ERROR and use `nvmlErrorString(nvmlerr)` instead. Look [here](https://docs.nvidia.com/deploy/nvml-api/group__nvmlErrorReporting.html#group__nvmlErrorReporting_1g93eca32cce1cbd5961fbecce10d41750).

src/core/metrics.cc
```diff
@@ -172,20 +172,42 @@ Metrics::InitializeNvmlMetrics()
...
   // Create NVML metrics for each GPU
-  LOG_INFO << "found " << dcnt << " GPUs supporting NVML metrics";
-  for (unsigned int didx = 0; didx < dcnt; ++didx) {
+  LOG_INFO << "found " << dcnt << " CUDA visible devices.";
+  for (int didx = 0; didx < dcnt; ++didx) {
     // Get handle for the GPU
+    cudaDeviceProp gpu_properties;
+    cudaError_t cudaerr = cudaGetDeviceProperties(&gpu_properties, didx);
+    if (cudaerr != cudaSuccess) {
+      LOG_ERROR << "failed to get device properties for cuda device " << didx << "CUDA_ERROR "
```
CoderHam, line 190: Remove CUDA_ERROR and replace with cudaGetErrorString(cudaerr)

src/core/metrics.cc
```diff
@@ -365,7 +387,7 @@ Metrics::UUIDForCudaDevice(int cuda_device, std::string* uuid)
   }
 
   nvmlDevice_t device;
-  nvmlReturn_t nvmlerr = nvmlDeviceGetHandleByPciBusId(pcibusid_str, &device);
+  nvmlReturn_t nvmlerr = nvmlDeviceGetHandleByPciBusId_v2(pcibusid_str, &device);
   if (nvmlerr != NVML_SUCCESS) {
     LOG_ERROR << "failed to get device from PCI Bus ID: NVML_ERROR " << nvmlerr;
```
CoderHam, line 392: Modify to `nvmlErrorString(nvmlerr)` instead

CoderHam (changes requested):

src/core/metrics.cc
```diff
@@ -172,20 +172,42 @@ Metrics::InitializeNvmlMetrics()
...
-  for (unsigned int didx = 0; didx < dcnt; ++didx) {
+  LOG_INFO << "found " << dcnt << " CUDA visible devices.";
+  for (int didx = 0; didx < dcnt; ++didx) {
     // Get handle for the GPU
+    cudaDeviceProp gpu_properties;
+    cudaError_t cudaerr = cudaGetDeviceProperties(&gpu_properties, didx);
+    if (cudaerr != cudaSuccess) {
+      LOG_ERROR << "failed to get device properties for cuda device " << didx << "CUDA_ERROR "
+                << cudaerr;
+      return false;
```
deadeyegoodwin, line 192: These 3 "return false" should be continue so that we just skip GPUs that we have trouble with. But then dcnt gets out of sync with the number of entries in gpu_utilization_, etc. What we should instead do is at live 236 save the "nvmlDevice_t gpu" value into nvml_device_ to that it is parallel to gpu_utilization_, etc. Then at line 246 and later we don't use dcnt but instead use the nvml_device_ and other vectors that all have the same length.

deadeyegoodwin (changes requested):

src/core/metrics.cc
Tabrizian: Do you want me to change all the error handlings to this way? Since other API calls to nvml in this file handle the errors the same way.

/cc @deadeyegoodwin

deadeyegoodwin: Yes, it would be a good improvement to update all the error handling to give a better message.

deadeyegoodwin (approved):

deadeyegoodwin (dismissed):

qa/L0_metrics/test.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash  
+# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
```
deadeyegoodwin, line 2: 2020

Tabrizian: Good catch.

deadeyegoodwin (approved):

PR #2132, opened 2020-10-16
Document allowed model repository modifications

deadeyegoodwin:


docs/model_management.md
```diff
@@ -129,3 +135,34 @@ changes:
   *label_filename* property of the output it corresponds to in the
   [model configuration](model_configuration.md) must be performed at
   the same time.
+
+## Modifying the Model Repository
+
+Each model in a model repository [resides in its own
+sub-directory](model_repository.md#repository-layout). The activity
+allowed on the contents of a model's sub-directory varies depending on
+how Triton is using that model. The loaded or unloaded state of a
```
CoderHam, line 144: How about just "The state of a model ... " since loaded/unloaded are not the only two states, Additionally loaded is represented by READY and unloaded by UNAVAILABLE. 
Additionally I do not think model metadata returns model state anymore. There is only model ready right?

docs/model_management.md
```diff
@@ -46,9 +46,11 @@ ignored. Model load and unload requests using the [model control
 protocol](protocols/extension_model_repository.md) will have no affect
 and will return an error response.
 
-This model control mode is selected by specifing
+This model control mode is selected by specifying
 --model-control-mode=none when starting Triton. This is the default
-model control mode.
+model control mode. Changing the model repository while Triton is
+running must be carefully, as explained in [Modifying the Model
```
CoderHam, line 52: must be done carefully. In other places well.

CoderHam (changes requested):

docs/model_management.md
deadeyegoodwin: Model metadata will return error if the model isn't loaded.

CoderHam: Yes you're correct.

CoderHam (approved):

PR #2134, opened 2020-10-17
Fix non-decoupled custom backend to return empty responses

### LINKED ISSUE #2104: Triton 20.08 hangs on inference when Custom Backend (Legacy, either v1 or v2) has no output tensors specified
Hey, I have moved the Kaldi from NVIDIA DeepLearning Examples into the new triton version and I observed the following bug. Kaldi is built on Legacy custom backend r19.12 in which the client requests output tensor iff the last chunk of a sequence is passed (is_end=True). For intermediate (is_end=False) chunks infer() function takes an empty output tensor array and the model returns an empty array of results. However in 20.08 the client simply hangs.

It turns out that the engine expects that the tensor specified in the model's configuration be populated even though we don't request it in the client.

The solution was to populate the output tensor for each time step.

**Triton Information**
20.08
Pre-built container

**To Reproduce**
It's complicated

**Expected behavior**
I expected to receive a response with an empty array of results.

tanmayv25:
Fixes #2104 

### Before

`tritonclient.utils.InferenceServerException: Deadline Exceeded`

### After

```
POST /v2/models/custom_identity_int32/infer, headers {'Inference-Header-Content-Length': 164}
b'{"inputs":[{"name":"INPUT0","shape":[1,1],"datatype":"INT32","parameters":{"binary_data_size":4}}],"outputs":[{"name":"OUTPUT0","parameters":{"binary_data":true}}]}\n\x00\x00\x00'
<HTTPSocketPoolResponse status=200 headers={'content-type': 'application/json', 'content-length': '71'}>
bytearray(b'{"model_name":"custom_identity_int32","model_version":"1","outputs":[]}')
```

### DIFF
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -202,6 +202,8 @@ RUN cp /tmp/tritonbuild/tritonserver/build/test-util/install/lib/libidentity.so 
         qa/L0_batcher/. && \
     cp /tmp/tritonbuild/tritonserver/build/test-util/install/lib/libidentity.so \
         qa/L0_client_timeout/. && \
+    cp /tmp/tritonbuild/tritonserver/build/test-util/install/lib/libidentity.so \
+        qa/L0_custom_legacy/. && \
     mkdir -p qa/L0_infer_shm && \
     cp -r qa/L0_infer/. qa/L0_infer_shm && \
     mkdir -p qa/L0_infer_cudashm && \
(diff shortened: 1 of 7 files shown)


deadeyegoodwin (commented):
@tanmayv25 How can we test this. Can  you modify an existing legacy custom backend to have some option not to set an output and add an appropriate test.

deadeyegoodwin (approved):

PR #2139, opened 2020-10-19
Adding max_queue_delay related configs for direct sequence

GuanLuo:


src/core/model_config.proto
```diff
@@ -1104,7 +1104,36 @@ message ModelSequenceBatching
...
+    //@@    .. cpp:var:: float minimum_slot_utilization
+    //@@
+    //@@       The rate of slot utilization that the selected preferred batch
+    //@@       size must satisfy, otherwise, the largest possible batch will
+    //@@       be executed. The utilization is calculated based on the number of
+    //@@       ready requests in the selected preferred batch. The utilization
+    //@@       is range from 0 to 1 where 1 represents full utilization.
+    //@@       Default is 0.
+    //@@
+    float minimum_slot_utilization = 2;
```
GuanLuo, line 1127: @deadeyegoodwin I added this field because direct sequence batcher reserves batch slot for individual sequence, so there may be the case where: preferred batch size is 4 and the slots are assigned to: null, null, null, seq 1, seq 2, seq 3, user can use this field to let batcher send a larger batch while "null, null, null, seq 1" forms a preferred batch.

Do you mind to provide feedback on how you feel about the preferred batch size handling here.

deadeyegoodwin: I don't think we want preferred_batch_size at all because of the "null" slot issue that you mention. I think the delay along with minimum_slot_utilization is sufficient. 'minimum_slot_utilization' should be "The minimum slot utilization that must be satisfied to execute the batch before 'max_queue_delay_microseconds' expires. For example, a value of 0.5 indicates that the batch should be executed as soon as 50% or more of the slots are ready even if the max_queue_delay_microseconds timeout has not expired. The default is 1.0, indicating that a batch will not execute before max_queue_delay_microseconds timeout expires unless every batch slot is ready.

deadeyegoodwin (changes requested):

src/core/model_config.proto
GuanLuo: I mention the default to be 0.0 instead of 1.0 because in protobuf 0 will be the default value if not specified, we shouldn't replace 0 to 1 as 0 is also an acceptable value. But you can also argues that 0 doesn't make the delay meaningful as all batches will be executed immediately, so I am also okay with changing the description to say 0 will be overwritten to 1.0

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -2034,6 +2034,275 @@ def test_backlog_sequence_timeout(self):
...
+            except Exception as ex:
+                self.assertTrue(False, "unexpected error {}".format(ex))
+            finally:
+                if TEST_SYSTEM_SHARED_MEMORY or TEST_CUDA_SHARED_MEMORY:
+                    self.cleanup_shm_regions(precreated_shm0_handles)
+                    self.cleanup_shm_regions(precreated_shm1_handles)
+
+    def test_queue_delay_half_min_util(self):
+        # Test model that have set max queue delay but minimum slot utilization
+        # is 0. So that all batch will be sent immediately.
```
deadeyegoodwin, line 2128: Update comment

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -2034,6 +2034,275 @@ def test_backlog_sequence_timeout(self):
...
+            except Exception as ex:
+                self.assertTrue(False, "unexpected error {}".format(ex))
+            finally:
+                if TEST_SYSTEM_SHARED_MEMORY or TEST_CUDA_SHARED_MEMORY:
+                    self.cleanup_shm_regions(precreated_shm0_handles)
+                    self.cleanup_shm_regions(precreated_shm1_handles)
+
+    def test_queue_delay_full_min_util(self):
+        # Test model that have set max queue delay but minimum slot utilization
+        # is 0. So that all batch will be sent immediately.
```
deadeyegoodwin, line 2218: update comment

src/core/model_config.proto
deadeyegoodwin: Perhaps just add one final sentence. "max_queue_delay_microseconds will be ignored unless minimum_slot_utilization is set to a non-zero value".
Does that make sense?

src/core/sequence_batch_scheduler.cc
```diff
@@ -993,6 +1000,8 @@ DirectSequenceBatch::SchedulerThread(
         //      b) the required tensor shapes for the batch for the
         //      case where ragged batching is not allowed
```
deadeyegoodwin, line 1001: Add "3) Determine the earliest enqueue time and number of ready sequences if queue delay is enabled"

src/core/sequence_batch_scheduler.cc
```diff
@@ -1041,8 +1050,47 @@ DirectSequenceBatch::SchedulerThread(
...
+          // check again at the maximum allowed delay.
+          struct timespec now;
+          clock_gettime(CLOCK_MONOTONIC, &now);
+          uint64_t now_ns = TIMESPEC_TO_NANOS(now);
+          uint64_t current_batch_delay_ns = (now_ns - earliest_enqueue_time_ns);
+          if ((current_batch_delay_ns > pending_batch_delay_ns_) ||
+              (((float)ready_cnt) / max_batch_size_ >=
+               minimum_slot_utilization_)) {
             wait_microseconds = 0;
+            LOG_ERROR << "execute now";
```
deadeyegoodwin, line 1074: Seems like all this logging should be LOG_VERBOSE

src/core/sequence_batch_scheduler.cc
```diff
@@ -1041,8 +1050,47 @@ DirectSequenceBatch::SchedulerThread(
...
             }
 
+            earliest_enqueue_time_ns = std::min(
+                earliest_enqueue_time_ns, queue.front()->QueueStartNs());
+            ready_cnt++;
             max_seq_slot = seq_slot;
+          }
+        }
+
+        if (max_seq_slot != -1) {
```
deadeyegoodwin, line 1060: Can we also check pending_batch_delay and minimum_slot_util here and just skip the whole thing if either is zero.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #2174, opened 2020-10-26
Add unittest for Python backend environment

Tabrizian:


deadeyegoodwin:
@Tabrizian Can you manually rebase this.

qa/python_models/model_env/model.py
```diff
@@ -0,0 +1,46 @@
...
+import os
+
+sys.path.append('../../')
+import triton_python_backend_utils as pb_utils
+
+
+class TritonPythonModel:
+
+    def initialize(self, args):
+        print("My_ENV =", os.environ["MY_ENV"])
```
deadeyegoodwin, line 38: I think it would be a better test if you would just return an error if My_ENV doesn't have the expected value and then the server would fail to start and then test.sh would fail. Grepping the log is not the best thing to do unless there is no other option.

deadeyegoodwin (changes requested):

qa/python_models/model_env/model.py
Tabrizian: Fixed.

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #2177, opened 2020-10-27
Async Model Load/Unload for GRPC endpoint

tanmayv25:
Only RepositoryModelUnload and RepositoryModelLoad specify the async flag true. Other RPCs are quite direct and overhead of spawning an extra thread and an extra state transition can become more pronounced.

Note that this implementation frees up the handler to serve the new incoming requests. Whenever the execution is completed the **async_thread_** will add the CallData back to the completion queue to later write response to client.

### DIFF
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -988,9 +988,7 @@ for protocol in grpc http; do
     kill $SERVER_PID
     wait $SERVER_PID
 
-    if [[ $protocol == "grpc" ]]; then
-        unset TRITONSERVER_USE_GRPC
-    fi
+    unset TRITONSERVER_USE_GRPC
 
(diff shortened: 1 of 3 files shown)


qa/L0_lifecycle/test.sh
```diff
@@ -950,83 +950,101 @@ wait $SERVER_PID
...
+    set -e
 
-kill $SERVER_PID
-wait $SERVER_PID
+    kill $SERVER_PID
+    wait $SERVER_PID
 
-LOG_IDX=$((LOG_IDX+1))
+    if [[ $protocol == "grpc" ]]; then
+        unset TRITONSERVER_USE_GRPC
```
deadeyegoodwin, line 992: I would just unconditionally unset it.

deadeyegoodwin (changes requested):

PR #2189, opened 2020-10-28
Fix crash when creating multiple instance of Tritonserver (in the same process)

CoderHam:
When creating multiple instances of `TRITONSERVER_Server` using `TRITONSERVER_ServerNew()` use the pinned and cuda memory pool defined by the first server and log the same.
- Earlier error reported that PinnedMemoryManager and CudaMemoryManager were already created 

Now just log the follow message and return success.
```
cuda_memory_manager.cc:80] New CUDA memory pools could not be created since they already exists
pinned_memory_manager.cc:182] New pinned memory pool of size 268435456 could not be created since one already exists of size 268435456
```
Note: The user must use the Pinned and CUDA memory pools as specified by the first instance of Tritonserver.

### DIFF
--- a/src/core/pinned_memory_manager.cc
+++ b/src/core/pinned_memory_manager.cc
@@ -49,6 +49,7 @@ PointerToString(void* ptr)
 }  // namespace
 
 std::unique_ptr<PinnedMemoryManager> PinnedMemoryManager::instance_;
+uint64_t PinnedMemoryManager::pinned_memory_byte_size_;
 
 PinnedMemoryManager::PinnedMemoryManager(
     void* pinned_memory_buffer, uint64_t size)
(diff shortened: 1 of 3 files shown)


src/core/pinned_memory_manager.cc
```diff
@@ -179,8 +179,10 @@ Status
 PinnedMemoryManager::Create(const Options& options)
 {
   if (instance_ != nullptr) {
-    return Status(
-        Status::Code::ALREADY_EXISTS, "PinnedMemoryManager has been created");
+    LOG_INFO << "New pinned memory pool of size "
+             << options.pinned_memory_pool_byte_size_
+             << " could not be created since one already exists.";
```
GuanLuo, line 184: Maybe use `WARNING`, and add more info about the existing pool will be used and what is the pool's setting (if can be retrieved easily..)

CoderHam: Done

deadeyegoodwin (approved):

PR #2203, opened 2020-10-29
Add Model Analyzer to Triton SDK Container

Tabrizian:
It should be merged after 20.11 code freeze.

### DIFF
--- a/Dockerfile.client
+++ b/Dockerfile.client
@@ -37,7 +37,7 @@ FROM ${BASE_IMAGE}
 ENV DEBIAN_FRONTEND=noninteractive
 
 # DCGM version to install for Model Analyzer
-ENV DCGM_VERSION=2.0.13
+ARG DCGM_VERSION=2.0.13
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
@@ -118,7 +118,7 @@ RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86
 
 # Install Model Analyzer
 ARG TRITON_MODEL_ANALYZER_REPO_TAG
-ENV TRITON_MODEL_ANALYZER_REPO "https://github.com/triton-inference-server/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}"
+ARG TRITON_MODEL_ANALYZER_REPO "https://github.com/triton-inference-server/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}"
 RUN pip3 install "git+${TRITON_MODEL_ANALYZER_REPO}"
 
 ENV PATH /workspace/install/bin:${PATH}
--- a/Dockerfile.client
+++ b/Dockerfile.client
@@ -29,7 +29,7 @@ ARG BASE_IMAGE=nvidia/cuda:11.1-devel-ubuntu18.04
 
 ARG TRITON_COMMON_REPO_TAG=main
 ARG TRITON_CORE_REPO_TAG=main
-ARG TRITON_MODEL_ANALYZER_REPO_TAG=main-dev
+ARG TRITON_MODEL_ANALYZER_REPO_TAG=main
 
 FROM ${BASE_IMAGE}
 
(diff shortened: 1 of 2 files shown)


Dockerfile.sdk
```diff
@@ -29,12 +29,16 @@ ARG BASE_IMAGE=nvidia/cuda:10.1-devel-ubuntu18.04
 
 ARG TRITON_COMMON_REPO_TAG=main
 ARG TRITON_CORE_REPO_TAG=main
+ARG TRITON_MODEL_ANALYZER_REPO_TAG=main-dev
```
dzier, line 32: This will need to be `main` branch before we merge this change

Dockerfile.sdk
```diff
@@ -29,12 +29,16 @@ ARG BASE_IMAGE=nvidia/cuda:10.1-devel-ubuntu18.04
...
 
 FROM ${BASE_IMAGE}
 
 # Ensure apt-get won't prompt for selecting options
 ENV DEBIAN_FRONTEND=noninteractive
 
+# DCGM version to install for Model Analyzer
+ENV DCGM_VERSION=2.0.13
```
dzier, line 40: We may have to hold off on this for now. If we go with NVML, then all of this new installation is not needed.

Dockerfile.sdk
```diff
@@ -108,6 +112,19 @@ RUN pip3 install --upgrade numpy pillow && \
     "tritonclient-*-manylinux1_x86_64.whl" | xargs printf -- '%s[all]' | \
     xargs pip3 install --upgrade
 
+# Install DCGM
+RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb && \
+    dpkg -i datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb
+
+# Install Model Analyzer
+ARG TRITON_MODEL_ANALYZER_REPO_TAG
+RUN git clone --single-branch --depth=1 -b $TRITON_MODEL_ANALYZER_REPO_TAG \
```
dzier, line 121: If there is a top level setup.ini file, you don't need to clone the repo. You can just use:
```
ARG TRITON_MODEL_ANALYZER_REPO=https://github.com/triton-inference-server/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}

pip install "git+${TRITON_MODEL_ANALYZER_REPO}"
```

This is one way we will instruct people install the model analyzer.

Dockerfile.sdk
```diff
@@ -108,6 +112,19 @@ RUN pip3 install --upgrade numpy pillow && \
...
+    https://github.com/triton-inference-server/model_analyzer.git && \
+    (export VERSION=`cat /workspace/VERSION` && cd model_analyzer && python3 setup.py install)
 
-ENV PATH //workspace/install/bin:${PATH}
+# Remove Model Analyzer Git source
+RUN rm -rf model_analyzer
+
+ENV PATH /workspace/install/bin:${PATH}
```
dzier, line 128: With pip install, this is not needed

dzier (changes requested):
Note that this will be used for how we release Model Analyzer.

Also, this does not rename the Triton client containers, so the PR title should change. You should separate adding Model Analyzer from renaming the container in two different PRs / commits.

Dockerfile.sdk
Tabrizian: This was in the Dockerfile before. I think this is because of the `perf_analyzer`.

Dockerfile.sdk
CoderHam: Please add TRITON_MODEL_ANALYZER_REPO_TAG to the gitlab yml (gitlab PR)

Tabrizian: It is in the gitlab PR.

Dockerfile.client
```diff
@@ -29,12 +29,16 @@ ARG BASE_IMAGE=nvidia/cuda:10.1-devel-ubuntu18.04
...
 ARG TRITON_CORE_REPO_TAG=main
+ARG TRITON_MODEL_ANALYZER_REPO_TAG=main
 
 FROM ${BASE_IMAGE}
 
 # Ensure apt-get won't prompt for selecting options
 ENV DEBIAN_FRONTEND=noninteractive
 
+# DCGM version to install for Model Analyzer
+ENV DCGM_VERSION=2.0.13
```
deadeyegoodwin, line 40: Why not an ARG? Is this envvar needed when the container is running?

Dockerfile.client
```diff
@@ -108,6 +112,15 @@ RUN pip3 install --upgrade numpy pillow && \
     "tritonclient-*-manylinux1_x86_64.whl" | xargs printf -- '%s[all]' | \
     xargs pip3 install --upgrade
 
+# Install DCGM
+RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb && \
+    dpkg -i datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb
+
+# Install Model Analyzer
+ARG TRITON_MODEL_ANALYZER_REPO_TAG
+ENV TRITON_MODEL_ANALYZER_REPO "https://github.com/triton-inference-server/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}"
```
deadeyegoodwin, line 121: Why not an ARG? Is this envvar needed when the container is running?

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #2069, opened 2020-09-29
Torchscript-Resnet-50 model examples added.

snlpatel001213:
The Torchscript model example was missing and hence added the same.
This example also provides intuition on how to deploy the 20+ PyTorch model with the help of Triton.

PR #2217, opened 2020-11-03
TorchScript example added. 

snlpatel001213:
…ing the model with triton. This example was missing and pytorch being choice of framework we should include this example as well.


deadeyegoodwin:
I think you must have missed my comment that I left in your older PR. https://github.com/triton-inference-server/server/pull/2069#issuecomment-714784632

This PR should be made in the contrib repo. You can create an "model examples" directory and make a subdirectory for this example within it.

PR #2219, opened 2020-11-03
Create multi-server example

CoderHam:
- Allow 3 repos and create 2 instances of Tritonserver that use:
   - Repo 1 and Repo 2
   - Repo 1 and Repo 3
- Use Model Repository APIs to load models
- Use Inference APIs to run models
- Use Model Repository APIs to unload models

Repo 1 has model: simple1
Repo 2 has model: simple2
Repo 3 has model: simple3
All 3 models do the same thing but have different names to avoid conflict.

Created an L0_multiserver test. Since it is technically a server I have kept it in src/servers and not src/test
Test passes successfully!

### DIFF
--- a/qa/L0_multi_server/test.sh
+++ b/qa/L0_multi_server/test.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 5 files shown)


qa/L0_multi_server/test.sh
```diff
@@ -0,0 +1,96 @@
...
+    echo -e "\n***\n*** Test Failed\n***"
+    exit 1
+fi
+
+MODELSDIR=`pwd`/models
+DATADIR=/data/inferenceserver/20.09dev/qa_model_repository
+
+export CUDA_VISIBLE_DEVICES=0
+
+# Must explicitly set LD_LIBRARY_PATH so that clients can find
```
GuanLuo, line 43: clients?

src/servers/multi_server.cc
```diff
@@ -0,0 +1,973 @@
...
+      "failed to load model");
+
+  // Wait for the models to become available.
+  bool is_int1 = true, is_int2 = true, is_int3 = true;
+  bool is_torch_model1 = false, is_torch_model2 = false,
+       is_torch_model3 = false;
+  AwaitModelReady(server1, "simple1", &is_int1, &is_torch_model1);
+  AwaitModelReady(server1, "simple2", &is_int2, &is_torch_model2);
+  AwaitModelReady(server2, "simple1", &is_int1, &is_torch_model1);
+  AwaitModelReady(server2, "simple3", &is_int3, &is_torch_model3);
```
GuanLuo, line 928: As an example, it can be more specific so that it is simpler.. We know that the models in repo is in float so we can remove `is_int`, and I would say use other framework instead of libtorch to avoid additional handling on IO names

src/servers/multi_server.cc
```diff
@@ -0,0 +1,973 @@
...
+
+  // Inference
+  RunInferenceAndValidate(
+      server1, allocator, "simple1", is_int1, is_torch_model1);
+  RunInferenceAndValidate(
+      server1, allocator, "simple2", is_int2, is_torch_model2);
+  RunInferenceAndValidate(
+      server2, allocator, "simple1", is_int1, is_torch_model1);
+  RunInferenceAndValidate(
+      server2, allocator, "simple3", is_int3, is_torch_model3);
```
GuanLuo, line 946: Also add test on operating on the wrong model? i.e. load/infer/unload simple3 in server1, to make sure different servers are monitoring different repositories.

src/servers/multi_server.cc
CoderHam: As discussed with @deadeyegoodwin we thought it would be good to cover multiple backends (except TRT)

src/servers/multi_server.cc
CoderHam: Good catch. I will add this

qa/L0_multi_server/test.sh
CoderHam: Typo. Fixed

src/servers/multi_server.cc
GuanLuo: Why except TRT? Besides that, we do have at least 3 different backends even excluding TRT and Torch, right? If not, then we have to keep libtorch in the example, it is just annoying to follow the additional naming convention.

CoderHam: Replacing libtorch with plan for now. I agree removing libtorch makes it better.

src/servers/multi_server.cc
```diff
@@ -0,0 +1,947 @@
...
+  // Wait and until the servers are both live and ready.
+  CheckServerLiveAndReady(server1);
+  CheckServerLiveAndReady(server2);
+
+  // Print status of the servers.
+  PrintServerStatus(server1);
+  PrintServerStatus(server2);
+
+  // Load models in both servers.
+  FAIL_IF_ERR(
```
tanmayv25, line 882: `TRITONSERVER_ServerLoadModel` is currently blocking right?
Will it not be a better idea to have the load/unload calls to server1 and server2 concurrently for this testing purpose?
May be we catch any synchronizing issues with model_repository?

PR #2248, opened 2020-11-13
TF Serving Client Backend for perf_analyzer

tanmayv25:
There are two important things to note:
By default, perf_analyzer will use the "serving_default" signature. The model might name the predict engine differently. In this case, use "--model-signature-name" option with perf_client.
The model metadata didn't carry any information regarding whether the first variable dimension is batching or not. I have hence, simplified the logic. The batch dimension, if present, should be provided explicitly to the perf_analyzer using --shape option. For instance, above --shape x:3,1 is generating request of batch size 3 to the model.  One can generate, batch size 10 requests by specifying --shape x:10,1.

**Also, I understand the build is not very clean but it does what it supposed to do. I will address improving the build through another ticket after we are done implementing torchserve.**

perf_analyzer running on TF Serving instance: 

```
root@ffc7411518be:/# /tmp/host/perf_analyzer --service-kind tfserving -m half_plus_two --shape x:3,1 -i grpc -v
*** Measurement Settings ***
  Measurement window: 5000 msec
  Using synchronous calls for inference

Request concurrency: 1
  Pass [1] throughput: 3384.2 infer/sec. Avg latency: 294 usec (std 106 usec)
  Pass [2] throughput: 3415.2 infer/sec. Avg latency: 291 usec (std 99 usec)
  Pass [3] throughput: 3390.2 infer/sec. Avg latency: 294 usec (std 107 usec)
  Client: 
    Request count: 16951
    Throughput: 3390.2 infer/sec
    Avg latency: 294 usec (standard deviation 107 usec)
    p50 latency: 267 usec
    p90 latency: 377 usec
    p95 latency: 435 usec
    p99 latency: 633 usec
    Avg gRPC time: 283 usec (marshal 4 usec + response wait 279 usec + unmarshal 0 usec)
Concurrency: 1, throughput: 3390.2 infer/sec, latency 294 usec
```

### DIFF
--- a/src/clients/c++/perf_analyzer/client_backend/tensorflow_serving/CompileProto.cmake
+++ b/src/clients/c++/perf_analyzer/client_backend/tensorflow_serving/CompileProto.cmake
@@ -1,3 +1,29 @@
+# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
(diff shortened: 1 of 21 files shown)


src/clients/c++/perf_analyzer/client_backend/tensorflow_serving/CompileProto.cmake
```diff
@@ -0,0 +1,90 @@
+# A function that creates CPP sources from proto files.
+function(COMPILE_PROTO USE_GRPC PROTO_PATH OUT_PATH SRCS HDRS)
```
deadeyegoodwin, line 2: I thought cmake had built-in support for generating C from proto

deadeyegoodwin (dismissed):

src/clients/c++/perf_analyzer/client_backend/tensorflow_serving/CompileProto.cmake
tanmayv25: I believe you mean protobuf_generate_cpp. The function produces the .pb.h and .pb.cc in the same file of invocation. But for TFS client we need the protos to retain the highly nested directory nature. I remember in  triton we would cleverly modify the import statetments in the protos to allow monolithic design but these are a lot of files here. In short, this function allows us to control the OUT_PATH of the protos and retain the nested structure from an external repository. The OUT_PATH will only carry the compiled protos without any rest of the junk from TF and TFS repos.

deadeyegoodwin (approved):

PR #2249, opened 2020-11-13
Allow -b flag for TFServing backend in perf_analyzer

tanmayv25:


src/clients/c++/perf_analyzer/main.cc
```diff
@@ -1083,11 +1085,8 @@ main(int argc, char** argv)
           << "perf_analyzer does not support async API for TensorFlow Serving."
           << std::endl;
       return 1;
-    } else if (batch_size > 1) {
-      std::cerr
-          << "perf_analyzer does not support -b flag with TensorFlow Serving "
-             "backend. See --shape to specify explicit batch dimension."
-          << std::endl;
+    } else if (!using_batch_size) {
+      batch_size = 0;
```
deadeyegoodwin, line 1089: Shouldn't there be some kind of error message?

deadeyegoodwin (changes requested):

src/clients/c++/perf_analyzer/main.cc
tanmayv25: If -b is not provided to the perf_analyzer cli, then it will assume model soen't support batching.

deadeyegoodwin (approved):

PR #2252, opened 2020-11-17
Modify L0_multi_server to run multiple TRITONSERVER_Server instances in parallel

CoderHam:
Use locks during creation of Cuda memory pool and enabling metrics

### DIFF
--- a/src/core/cuda_memory_manager.cc
+++ b/src/core/cuda_memory_manager.cc
@@ -57,6 +57,7 @@ PointerToString(void* ptr)
 namespace nvidia { namespace inferenceserver {
 
 std::unique_ptr<CudaMemoryManager> CudaMemoryManager::instance_;
+std::mutex CudaMemoryManager::cuda_memory_creating_;
 
 CudaMemoryManager::~CudaMemoryManager()
 {
@@ -76,6 +77,8 @@ CudaMemoryManager::Reset()
 Status
 CudaMemoryManager::Create(const CudaMemoryManager::Options& options)
 {
+  // Ensure thread-safe creation of CUDA memory pool
+  std::lock_guard<std::mutex> lock(cuda_memory_creating_);
   if (instance_ != nullptr) {
     LOG_WARNING << "New CUDA memory pools could not be created since they "
                    "already exists";
--- a/src/core/cuda_memory_manager.cc
+++ b/src/core/cuda_memory_manager.cc
@@ -57,7 +57,7 @@ PointerToString(void* ptr)
 namespace nvidia { namespace inferenceserver {
 
 std::unique_ptr<CudaMemoryManager> CudaMemoryManager::instance_;
-std::mutex CudaMemoryManager::cuda_memory_creating_;
+std::mutex CudaMemoryManager::instance_mu_;
 
 CudaMemoryManager::~CudaMemoryManager()
 {
@@ -71,6 +71,7 @@ CudaMemoryManager::~CudaMemoryManager()
 void
 CudaMemoryManager::Reset()
 {
+  std::lock_guard<std::mutex> lock(instance_mu_);
   instance_.reset();
 }
 
@@ -78,7 +79,7 @@ Status
 CudaMemoryManager::Create(const CudaMemoryManager::Options& options)
 {
   // Ensure thread-safe creation of CUDA memory pool
-  std::lock_guard<std::mutex> lock(cuda_memory_creating_);
+  std::lock_guard<std::mutex> lock(instance_mu_);
   if (instance_ != nullptr) {
     LOG_WARNING << "New CUDA memory pools could not be created since they "
                    "already exists";
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -152,7 +152,6 @@ Metrics::EnableGPUMetrics()
   // Ensure thread-safe enabling of GPU Metrics
   std::lock_guard<std::mutex> lock(singleton->gpu_metrics_enabling_);
   if (singleton->gpu_metrics_enabled_) {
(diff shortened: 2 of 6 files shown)


src/core/cuda_memory_manager.cc
```diff
@@ -57,6 +57,7 @@ PointerToString(void* ptr)
 namespace nvidia { namespace inferenceserver {
 
 std::unique_ptr<CudaMemoryManager> CudaMemoryManager::instance_;
+std::mutex CudaMemoryManager::cuda_memory_creating_;
```
deadeyegoodwin, line 60: Call this 'instance_mu_' or similar because I think it should also be used in the Reset call to protect intantance_.reset()

src/core/metrics.cc
```diff
@@ -148,6 +148,9 @@ void
 Metrics::EnableGPUMetrics()
 {
   auto singleton = GetSingleton();
+
+  // Ensure thread-safe enabling of GPU Metrics
+  std::lock_guard<std::mutex> lock(singleton->gpu_metrics_enabling_);
   if (singleton->gpu_metrics_enabled_) {
     LOG_WARNING << "GPU metrics already enabled";
```
deadeyegoodwin, line 155: Remove this warning, just return

src/servers/multi_server.cc
```diff
@@ -776,15 +785,109 @@ PrintModelStats(
...
+void
+CreateAndRunTritonserverInstance(
+    std::vector<std::string> model_repository_paths, size_t thread_id,
+    bool verbose_level)
+{
+  std::unique_lock<std::mutex> lock(mutex);
+  counter++;
+  cv.wait(lock);
+
+  TRITONSERVER_ServerOptions* server_options = nullptr;
```
deadeyegoodwin, line 801: Running this just one time means that it is very unlikely that the 2 threads will actually be doing anything at the same time. First it would be better to run with more threads / server instances. The models are small? So we should be able to run many instances /threads. 128? 256? And then the body of this function should iterate a large number of times (as many as we can and still have a ~5 minute runtime). That will allow the threads to interleave in many ways. Will need to Stop and Delete the server at the end of each iteration.

deadeyegoodwin (changes requested):

src/servers/multi_server.cc
CoderHam: Yes I had expanded the multi_server to be able to run for N threads. Will modify test.sh to run for 128 threads.

CoderHam: Doesn't TRITONSERVER_ServerDelete call Stop anyway and would be called as soon as CreateAndRunTritonserverInstance ends?

deadeyegoodwin: Could be, then you don't need stop. But you are going to create, run, delete in a look within CreateAndRunTrionServerInstance. You can't just run that once for each thread... that doesn't provide much of a test.

CoderHam: Made the changes to run CreateAndRunTrionServerInstance 4 times and changed thread count to 64. The time to run the test is approximately 20 mins

deadeyegoodwin: 4 times is likely not enough, 100's would be better. Why does it take so long to run? Are we doing something slow during the iterations.

CoderHam: We use locks during the creation of Tritonserver to avoid failures during cuda memory pool creation and enabling of metrics. As well as when destroying the cuda memory pool.  Additionally loading/unloading the models may be adding time. However the _major slowdown_ would be loading these models and _waiting_ for the server to be ready to perform inference. 

Just to confirm we want to run CreateAndRunTritonserverInstance 100 times in 64/128 threads?

GuanLuo (changes requested):
I overlooked this before, I think you need to take a look at cnmem to check how to have multiple managed memory. `cnmemInit()` doesn't return a handle so either it is not supported or additional handling needs to be done.

GuanLuo:
I see. So we haven't move it to non-singleton yet? Are we planning to do so?

CoderHam:
Yes, the shift to a clear, non-singleton approach will happen later. These changes are meant to only unblock users from using multiple instances of Tritonserver in parallel. 
> I see. So we haven't move it to non-singleton yet? Are we planning to do so?

src/servers/multi_server.cc
```diff
@@ -776,15 +786,120 @@ PrintModelStats(
...
+void
+RepeatedlyCreateAndRunInstance(
+    std::vector<std::string> model_repository_paths, size_t thread_id,
+    bool verbose_level)
+{
+  std::unique_lock<std::mutex> lock(mutex);
+  counter++;
+  cv.wait(lock);
+
+  for (size_t i = 0; i < 10; i++) {
```
deadeyegoodwin, line 887: Number of iterations should be a cmdline arg.

src/servers/multi_server.cc
deadeyegoodwin: I'd still like to understand why we can only run this loop 32 * 10 = 320 times.

deadeyegoodwin (changes requested):

GuanLuo:
> Yes, the shift to a clear, non-singleton approach will happen later.

Cool. Can you add this concern as a reminder in the tracking ticket?

src/servers/multi_server.cc
CoderHam: Reduced the sleep I was using to 100ms. Now we can run 32 threads 32 times and it takes 20-22 mins. Reducing it further does not decrease runtime significantly. 
PS: Increasing to 2x threads (64) increases overall runtime by 2x as well.

deadeyegoodwin (dismissed):

PR #2259, opened 2020-11-18
Remove PyTorch from LD_LIBRARY_PATH; Fix PyTorch custom op docs

CoderHam:
User must add Pytorch libs to LD_LIBRARY_PATH to load models with custom ops. Not needed for torchvision default ops.

### DIFF
--- a/docs/custom_operations.md
+++ b/docs/custom_operations.md
@@ -148,7 +148,9 @@ is currently no way to handle it.
 Starting with the 20.07 release of Triton the [TorchVision
 operations](https://github.com/pytorch/vision) will be included with
 the PyTorch backend and hence they do not have to be explicitly added
-as custom operations.
+as custom operations. The Pytorch shared libraries must be available
+for the TorchVision custom operations to be loaded. To do this,
+we must add /opt/tritonserver/backends/pytorch to the LD_LIBRARY_PATH.
 
 When building the custom operations shared library it is important to
 use the same version of PyTorch as is being used in Triton. You can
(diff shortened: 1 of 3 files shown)


docs/custom_operations.md
```diff
@@ -148,7 +148,9 @@ is currently no way to handle it.
 Starting with the 20.07 release of Triton the [TorchVision
 operations](https://github.com/pytorch/vision) will be included with
 the PyTorch backend and hence they do not have to be explicitly added
-as custom operations.
+as custom operations. The Pytorch shared libraries must be available
```
deadeyegoodwin, line 151: If the user does this they will break their tritonserver in the way we are seeing now. Did you try use --no-as-needed and linking torchvision against libtriton_pytorch.so?

deadeyegoodwin (changes requested):

docs/custom_operations.md
CoderHam: We already do that in our current "main" [here](https://github.com/triton-inference-server/pytorch_backend/blob/main/CMakeLists.txt#L129)

CoderHam: Looking into alternative solutions as well

CoderHam: To clarify: We do not need torchvision.so to load and run the torchvision custom ops. Torchvision looks for torch.so (and other pytorch deps) to load the ops/models.

deadeyegoodwin (approved):

PR #2262, opened 2020-11-18
WAR for glibc rapid thread spawn issue, in model repository manager

deadeyegoodwin:


CoderHam:
Do we know 100ms will be enough?

CoderHam (approved):
I agree that using a thread pool instead of spawning new threads is a better solution. We can work on fixing this in the next release.

deadeyegoodwin:
100ms is an looong time. Probably could go much smaller but I erred on the cautious side given that loading 10 models / sec should be ok.

GuanLuo (approved):

PR #2271, opened 2020-11-19
Add ensemble CI test for Python backend

Tabrizian:


qa/python_models/add_sub/config.pbtxt
```diff
@@ -0,0 +1,59 @@
...
+]
+output [
+  {
+    name: "OUTPUT1"
+    data_type: TYPE_FP32
+    dims: [ 4 ]
+  }
+]
+
+instance_group [{ kind: KIND_CPU }]
```
CoderHam, line 59: Add newline

qa/python_models/ensemble/config.pbtxt
```diff
@@ -0,0 +1,72 @@
...
+        key: "OUTPUT0"
+        value: "OUTPUT0"
+      }
+      output_map {
+        key: "OUTPUT1"
+        value: "OUTPUT1"
+      }
+    }
+  ]
+}
```
CoderHam, line 72: Add newline

CoderHam (changes requested):
You can remove 
```
```

from `python_test.py` and also use httpclient directly instead of client_util since you are only running the test with the HTTP client library.

qa/python_models/add_sub/config.pbtxt
Tabrizian: It has a new line. It just doesn't show it.

qa/L0_backend_python/python_test.py
```diff
@@ -208,6 +201,29 @@ def test_init_args(self):
...
+                                      np_to_triton_dtype(input_data_0.dtype)),
+                httpclient.InferInput("INPUT1", input_data_1.shape,
+                                      np_to_triton_dtype(input_data_1.dtype))
+            ]
+            inputs[0].set_data_from_numpy(input_data_0)
+            inputs[1].set_data_from_numpy(input_data_1)
+            result = client.infer(model_name, inputs)
+            output0 = result.as_numpy('OUTPUT0')
+            output1 = result.as_numpy('OUTPUT1')
+            self.assertIsNotNone(output0)
```
deadeyegoodwin, line 221: Can you compare output0 and output1 against calculated values using input_data_0 and input_data_1 to make sure you get the expected values back.

deadeyegoodwin (changes requested):

qa/L0_backend_python/python_test.py
Tabrizian: I did compare it in the lines below.

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #2279, opened 2020-11-20
Fix memory leak in python client shared memory

CoderHam:
- must munmap before shm_unlink

### DIFF
--- a/src/clients/python/library/tritonclient/utils/shared_memory/__init__.py
+++ b/src/clients/python/library/tritonclient/utils/shared_memory/__init__.py
@@ -252,8 +252,8 @@ class SharedMemoryException(Exception):
             -2: "unable to get shared memory descriptor",
             -3: "unable to initialize the size",
             -4: "unable to read/mmap the shared memory region",
-            -5: "unable to munmap the shared memory region",
-            -6: "unable to unlink the shared memory region"
+            -5: "unable to unlink the shared memory region",
+            -6: "unable to munmap the shared memory region"
         }
         self._msg = None
         if type(err) == str:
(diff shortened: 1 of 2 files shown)


src/clients/python/library/tritonclient/utils/shared_memory/__init__.py
```diff
@@ -252,7 +252,8 @@ def __init__(self, err):
             -2: "unable to get shared memory descriptor",
             -3: "unable to initialize the size",
             -4: "unable to read/mmap the shared memory region",
-            -5: "unable to unlink the shared memory region"
+            -5: "unable to munmap the shared memory region",
```
GuanLuo, line 255: Don't change the defined error, just append the new type of error

CoderHam: Fixed :)

GuanLuo: Did you push your fix to remote?

CoderHam: Changes had not gotten pushed. Just did it.

src/clients/python/library/tritonclient/utils/shared_memory/shared_memory.cc
```diff
@@ -130,12 +130,21 @@ GetSharedMemoryHandleInfo(
...
+  SharedMemoryHandle* handle =
+      reinterpret_cast<SharedMemoryHandle*>(shm_handle);
+  char* shm_addr = reinterpret_cast<char*>(handle->base_addr_);
+  size_t byte_size = handle->byte_size_;
+  int status = munmap(shm_addr, byte_size);
+  if (status == -1) {
+    return -6;
+  }
+
+  std::string shm_key = handle->shm_key_.c_str();
```
tanmayv25, line 142: You can directly pass the `handle->shm_key_.c_str()` in the shm_unlink function below. No need to create an additional string.

CoderHam: Done

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #2286, opened 2020-11-24
Fix backend to successfully create KIND_MODEL instances

tanmayv25:


src/backends/backend/triton_model_instance.cc
```diff
@@ -79,10 +79,15 @@ TritonModelInstance::CreateInstances(
               model, group.name(), c, TRITONSERVER_INSTANCEGROUPKIND_GPU,
               device_id, instances));
         }
+      } else if (group.kind() == inference::ModelInstanceGroup::KIND_MODEL) {
+        RETURN_IF_ERROR(CreateInstance(
+            model, group.name(), c, TRITONSERVER_INSTANCEGROUPKIND_MODEL,
+            -2 /* device_id */, instances));
```
deadeyegoodwin, line 85: Why do you use -2 for device ID? It should be ignored for KIND_MODEL.

deadeyegoodwin (changes requested):

src/backends/backend/triton_model_instance.cc
tanmayv25: Just following the convention here: https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L1083
This indeed will be ignored for kind_model here: https://github.com/triton-inference-server/server/blob/master/src/backends/backend/triton_model_instance.cc#L43-L50

qa/L0_model_config/autofill_noplatform_success/tensorflow_savedmodel/kind_model_config/expected.3
```diff
@@ -0,0 +1,43 @@
+name: "kind_model_config"
```
Tabrizian, line 1: License

src/backends/backend/triton_model_instance.cc
deadeyegoodwin: The -2 is tensorflow specific. Here in this common code you should just use 0 (zero).

tanmayv25: Done

deadeyegoodwin (approved):

PR #2290, opened 2020-11-24
PR to include session token in env vars when using temp credentials f…

nsiddharth:
PR to include session token in env vars when using temp credentials for IAM roles
More information can be found here: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html


CoderHam (approved):
These changes look good to me. Thank you for your contribution @nsiddharth. If this is your first contribution to the Tritonserver repository could you complete the [CLA](https://github.com/NVIDIA/triton-inference-server/blob/master/CONTRIBUTING.md) and send it out to triton-cla@nvidia.com, if you have not already done so before?

nsiddharth:
I have signed the CLA and sent it. Is there any ETA on when we can expect the next release with these changes merged? It would be good to know so that we can use the docker image directly. Thanks!

deadeyegoodwin:
We got the CLA, thanks. We've passed the code freeze for the december release, so you should expect these changes in the 21.01 release (end of January).

PR #2295, opened 2020-11-25
Move custom backend out to separate repo

CoderHam:


src/core/persistent_backend_manager.cc
```diff
@@ -61,7 +61,8 @@ Status
 PersistentBackendManager::InitPersistentBackends(
     const BackendCmdlineConfigMap& config_map)
 {
-  for (const auto& be : {"pytorch", "tensorflow", "onnxruntime"}) {
+  // FIXME this should be customized based on which backend is enabled.
+  for (const auto& be : {"pytorch", "tensorflow", "onnxruntime", "custom"}) {
```
deadeyegoodwin, line 65: We shouldn't need to persistently load the "custom" backend so it doesn't need to be added here.

deadeyegoodwin (changes requested):

src/core/persistent_backend_manager.cc
CoderHam: Yes, removed.

PR #2298, opened 2020-11-29
Fix docker image tag in docs: py3-sdk -> py3-clientsdk

praveen-palanisamy:
As per the current instructions in [quickstart.md](https://github.com/triton-inference-server/server/blob/master/docs/quickstart.md), the suggested command for getting the client libraries and examples using docker is:
`$ docker pull nvcr.io/nvidia/tritonserver:<xx.yy>-py3-sdk`
 which will fail since there's no such tag on NVCR for any supported values for `xx.yy`

This PR adds the minor fix to update the docs to:
`$ docker pull nvcr.io/nvidia/tritonserver:<xx.yy>-py3-clientsdk`

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -128,19 +128,19 @@ Use docker pull to get the client libraries and examples image
 from NGC.
 
 ```
-$ docker pull nvcr.io/nvidia/tritonserver:<xx.yy>-py3-sdk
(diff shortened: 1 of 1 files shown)


Tabrizian:
This is an intentional rename of the `cliensdk` container which will be released in about one month. Please use the `r20.11` branch of the docs. https://github.com/triton-inference-server/server/blob/r20.11/docs/quickstart.md

PR #2309, opened 2020-12-02
Use TF2 in the benchmark comparison

tanmayv25:


CoderHam:
Will this compare **TFServing with TF2** to **Triton + TF2**?

deadeyegoodwin (approved):

PR #2314, opened 2020-12-04
Increasing the instance count in benchmark

tanmayv25:
The CI has passed here:
https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/1860341
We will use 4 instances for a fair comparison with TFS.


qa/L0_perf_resnet/test.sh
```diff
@@ -182,7 +182,7 @@ done
 MODEL_NAME=${TF_MODEL_NAME}
 REPO=$REPODIR/perf_model_store
 STATIC_BATCH=128
-INSTANCE_CNT=1
+INSTANCE_CNT=4
```
deadeyegoodwin, line 185: Having instance count 4 with only concurrency of 4 is not going to keep those instances busy. Also I'd rather not change what we are measuring now but instead add a new measurement if that is necessary for comparison with TFS. Perhaps we should just change how we run the TFS test so that it only runs 1 instance at a time?

tanmayv25: TFS calls session->Run() with each incoming request. There are other cli options like `tensorflow_session_parallelism` 
 and `tensorflow_inter/intra_op_parallelism`  which can be used to limit the activity. But as Triton by design, makes only a single session->Run call per instance at a time and does not share session across instances in a gpu as of now, performance comparison at a >1 concurrency  is unfair for Triton.  We can benchmark for these concurrencies once we allow concurrent calls to session->Run() within triton.
This test case i added is a new one just intended to collect numbers that can be compared with Triton at a large batch size.

deadeyegoodwin (approved):
This is ok but we only have min latency testing (concurrency 1) for TFS, which is ok for now but we may want throughput testing at some point.

PR #2319, opened 2020-12-08
Update nvidia-docker documentation

deadeyegoodwin:


nvidia_entrypoint.sh
```diff
@@ -77,13 +77,6 @@ if ! cat /proc/cpuinfo | grep flags | sort -u | grep avx >& /dev/null; then
   sleep 2
 fi
 
-if [[ "$(df -k /dev/shm |grep ^shm |awk '{print $2}') " == "65536 " ]]; then
```
CoderHam, line 80: I think we should keep this documentation and just change `nvidia-docker` to `docker`.

CoderHam (changes requested):

PR #2338, opened 2020-12-11
Add async work queue for parallel execution of tasks. Parallelize only the copy of pinned input for simplicity

GuanLuo:


src/core/async_work_queue.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void
+AsyncWorkQueue::AddTask(const std::function<void(void)>&& task)
+{
+  GetSingleton()->task_queue_.Put(std::move(task));
+}
+
+void
+AsyncWorkQueue::InitializeThread()
```
deadeyegoodwin, line 82: InitializeThread is a strange name.... how about WorkThread.

src/core/server.cc
```diff
@@ -130,6 +132,11 @@ InferenceServer::Init()
   // maintaining a reference to them).
   status = PersistentBackendManager::Create(
       backend_cmdline_config_map_, &persist_backend_manager_);
+  // Always initialize async work queue for now
+  if (status.IsOk()) {
+    status = AsyncWorkQueue::Initialize(
```
deadeyegoodwin, line 137: Seems like we don't need to do this, only initialize when thread_count > 0.

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+Status
+AsyncWorkQueue::Initialize(size_t worker_count)
+{
+  if (worker_count < 1) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "Async work queue must be initialized with positive 'worker_count'");
+  }
```
deadeyegoodwin, line 61: Add another check here for avoid double initialization.

src/core/async_work_queue.cc
```diff
@@ -0,0 +1,94 @@
...
+size_t
+AsyncWorkQueue::WorkerCount()
+{
+  return GetSingleton()->worker_threads_.size();
+}
+
+void
+AsyncWorkQueue::AddTask(const std::function<void(void)>&& task)
+{
+  GetSingleton()->task_queue_.Put(std::move(task));
```
deadeyegoodwin, line 78: Return Status, and add a check to make sure that there are some worker threads initialized.

deadeyegoodwin (changes requested):

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #2341, opened 2020-12-15
Description of correctly loaded models

Sn0flingan:
It was not clear that the output needed to display the models loaded in the table. I added this to the expected output and provided some explanations for the errors I ran into, which I presume could be common errors.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -89,10 +89,27 @@ see output like the following, Triton is ready to accept inference
 requests.
 
 ```
++----------------------+---------+--------+
+| Model                | Version | Status |
++----------------------+---------+--------+
+| densenet_onnx        | 1       | READY  |
+| inception_graphdef   | 1       | READY  |
+| simple               | 1       | READY  |
+| simple_dyna_sequence | 1       | READY  |
+| simple_identity      | 1       | READY  |
+| simple_int8          | 1       | READY  |
+| simple_sequence      | 1       | READY  |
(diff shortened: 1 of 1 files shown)


docs/quickstart.md
```diff
@@ -89,11 +89,28 @@ see output like the following, Triton is ready to accept inference
...
++----------------------+---------+--------+
+...
+...
+...
 I1002 21:58:57.891440 62 grpc_server.cc:3914] Started GRPCInferenceService at 0.0.0.0:8001
 I1002 21:58:57.893177 62 http_server.cc:2717] Started HTTPService at 0.0.0.0:8000
 I1002 21:58:57.935518 62 http_server.cc:2736] Started Metrics Service at 0.0.0.0:8002
 ```
 
+If the models are not loaded check your CUDA drivers and the path to the model repository.
```
deadeyegoodwin, line 112: I would rather say something more general here.... perhaps "All the models should show "READY" status to indicate that they loaded correctly. If a model fails to  load the status will report the failure and a reason for the failure."

deadeyegoodwin (changes requested):

docs/quickstart.md
Sn0flingan: Alright, sounds like a good idea. For me though the models did not show up in the table until I had an updated CUDA driver and a correct path. So I added that to the updated description, I hope that is ok.

deadeyegoodwin (approved):

PR #2382, opened 2021-01-06
Using curl_mime_data_cb to capture HTTP send/recv times for TorchServe

tanmayv25:
Using the callback based approach allows perf analyzer to capture and report the http send/receive times for TorchServe service.

### Before
```
root@58d21eb46ebc:/home/model-server# /tmp/host/perf_analyzer -m resnet50 -u localhost:8080 -i http --service-kind torchserve --input-data data.json
 Successfully read data for 1 stream/streams with 1 step/steps.
*** Measurement Settings ***
  Measurement window: 5000 msec
  Using synchronous calls for inference

Request concurrency: 1
  Client: 
    Request count: 849
    Throughput: 169.8 infer/sec
    Avg latency: 5876 usec (standard deviation 528 usec)
    p50 latency: 5809 usec
    p90 latency: 5920 usec
    p95 latency: 6004 usec
    p99 latency: 7310 usec
    Avg HTTP time: 5852 usec (send/recv 0 usec + response wait 5852 usec)
Concurrency: 1, throughput: 169.8 infer/sec, latency 5876 usec

```

### After
```
root@58d21eb46ebc:/home/model-server# /tmp/host/perf_analyzer -m resnet50 -u localhost:8080 -i http --service-kind torchserve --input-data data.json
 Successfully read data for 1 stream/streams with 1 step/steps.
*** Measurement Settings ***
  Measurement window: 5000 msec
  Using synchronous calls for inference

Request concurrency: 1
  Client: 
    Request count: 847
    Throughput: 169.4 infer/sec
    Avg latency: 5889 usec (standard deviation 505 usec)
    p50 latency: 5793 usec
    p90 latency: 5947 usec
    p95 latency: 6484 usec
    p99 latency: 7075 usec
    Avg HTTP time: 5872 usec (**send/recv 83 usec** + response wait 5789 usec)
Concurrency: 1, throughput: 169.4 infer/sec, latency 5889 usec
```

### DIFF
--- a/src/clients/c++/perf_analyzer/client_backend/torchserve/http_client.h
+++ b/src/clients/c++/perf_analyzer/client_backend/torchserve/http_client.h
@@ -30,6 +30,8 @@
 #include "src/clients/c++/perf_analyzer/client_backend/torchserve/torchserve_infer_input.h"
 
 #include <curl/curl.h>
+#include <stdio.h>
+#include <stdlib.h>
 
 namespace nic = nvidia::inferenceserver::client;
 
@@ -91,6 +93,10 @@ class HttpClient : public nic::InferenceServerClient {
       const std::vector<InferInput*>& inputs,
       const std::vector<const InferRequestedOutput*>& outputs,
       const Headers& headers, std::shared_ptr<HttpInferRequest>& request);
+
+  static size_t ReadCallback(
+      char* buffer, size_t size, size_t nitems, void* userp);
+  static int SeekCallback(void* userp, curl_off_t offset, int origin);
   static size_t InferResponseHeaderHandler(
       void* contents, size_t size, size_t nmemb, void* userp);
   static size_t InferResponseHandler(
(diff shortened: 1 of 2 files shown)


src/clients/c++/perf_analyzer/client_backend/torchserve/http_client.h
```diff
@@ -108,18 +114,30 @@ class HttpClient : public nic::InferenceServerClient {
...
   std::string& DebugString() { return *infer_response_buffer_; }
+  FILE* FilePtr() { return file_ptr_.get(); }
   friend HttpClient;
   friend InferResult;
 
  private:
   // Pointer to the list of the HTTP request header, keep it such that it will
   // be valid during the transfer and can be freed once transfer is completed.
   struct curl_slist* header_list_;
+  std::unique_ptr<FILE, Deleter> file_ptr_;
```
GuanLuo, line 140: Why you provide a deleter if it does nothing?

tanmayv25: The default deleter for unique_ptr will call delete on the FILE*. But, the close call is already clearing off the resources. If I don't provide an empty deleter a second attempt to delete the pointer will be made when the unique_ptr object is destructed. An alternative is to check and close the FILE* in the deleter instead of [here](https://github.com/triton-inference-server/server/blob/1398887ecfc686bb4703cdbb04c487d7fcef75e0/src/clients/c%2B%2B/perf_analyzer/client_backend/torchserve/http_client.cc#L182). But I would prefer to close the file as soon as we are done uploading the contents.

GuanLuo: If you are using an explicit function for closing the file, then I don't think you really need a smart pointer. Besides that, it is possible to have resource leak if `CloseFileData()` doesn't get called, smart pointer with empty deleter doesn't help that either. I would have moved the closing file logic inside the deleter and explicitly call `reset()` if you want to close it earlier (or let `CloseFileData()` be just calling reset() to make the intention more clear).

GuanLuo (dismissed):

PR #2390, opened 2021-01-07
Add test for Python backend when the input is in GPU

Tabrizian:
Tests for triton-inference-server/python_backend#30


deadeyegoodwin (approved):

CoderHam (approved):
I see you fixed the np.object dtype check in this PR as well. Good!

PR #2405, opened 2021-01-12
Remove pytorch custom build and move to pytorch repository

CoderHam:
- Similar to TF, PyTorch build now copies out files from the Pytorch docker directly.


deadeyegoodwin (dismissed):
On gitlab side need to remove the pull/push of the tritonserver-pytorch image as it is not longer needed.

deadeyegoodwin (approved):

PR #2420, opened 2021-01-14
Add test for unicode characters in Python backend

Tabrizian:


deadeyegoodwin (dismissed):

qa/python_models/string/model.py
```diff
@@ -0,0 +1,55 @@
...
+            input_tensors = request.inputs()
+            in_0 = pb_utils.get_input_tensor_by_name(request, "INPUT0")
+            unicode_character = '😀'
+            if str(in_0.as_numpy().tobytes(), encoding='utf-8') == unicode_character:
+                out_tensor_0 = pb_utils.Tensor("OUTPUT0", np.array(['success'], dtype=np.bytes_))
+            else:
+                out_tensor_0 = pb_utils.Tensor("OUTPUT0", np.array(['failure'], dtype=np.bytes_))
+            responses.append(
+                pb_utils.InferenceResponse([out_tensor_0]))
+        return responses
```
CoderHam, line 55: Add new line after

qa/L0_backend_python/python_test.py
```diff
@@ -245,6 +245,22 @@ def test_ensemble(self):
...
+            input_data = np.array([bytes(utf8, encoding='utf-8')], dtype=np.bytes_)
+            inputs = [
+                httpclient.InferInput("INPUT0", shape,
+                                      np_to_triton_dtype(input_data.dtype))
+            ]
+            inputs[0].set_data_from_numpy(input_data)
+            result = client.infer(model_name, inputs)
+            output0 = result.as_numpy('OUTPUT0')
+            self.assertTrue(output0 is not None)
+            self.assertTrue(str(output0[0], encoding='ascii') == 'success')
```
CoderHam, line 262: Instead of returning success I would pass unicode and have an identity model that returns the same so you can verify that the handling of unicode for outputs is also validated.

CoderHam (changes requested):

qa/L0_backend_python/python_test.py
Tabrizian: Fixed. Although, since it is just a blob of bytes I don't think it makes a difference.

qa/python_models/string/model.py
Tabrizian: It does have a new line. Whenever it doesn't, it shows a little sign at the end of the line.

deadeyegoodwin (approved):

PR #2422, opened 2021-01-15
Wrap <> pseudovariables in markdown, to avoid their interpretation as HTML

okdimok:
In the Web version of the current documentation, there existed many cases when \<pseudovariables\> were used, but < and > symbols weren't properly escaped. In these cases, they were interpreted by the Markdown engines as inline HTML, and were completely invisible on the page. In this PR I escaped most of the < and > symbols in the documentation. Note, that such symbols inside the code blocks are not interpreted as inline HTML and thus there is no need to escape them.

See example:
in the comment source:  `<var1> vs \<var2\>`
rendered: <var1> vs \<var2\>

### DIFF
--- a/docs/client_libraries.md
+++ b/docs/client_libraries.md
@@ -134,7 +134,7 @@ from NGC.
 $ docker pull nvcr.io/nvidia/tritonserver:<xx.yy>-py3-sdk
 ```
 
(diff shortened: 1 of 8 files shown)


deadeyegoodwin (approved):
Thanks for the contribution!

PR #2424, opened 2021-01-16
Improve BackendInputCollector to only synchronize with async task at Finalize()

GuanLuo:


src/core/cuda_utils.cc
```diff
@@ -90,26 +90,9 @@ CopyBuffer(
...
-              dst, dst_memory_type_id, src, src_memory_type_id, byte_size,
-              cuda_stream),
-          msg + ": failed to perform CUDA copy");
-    } else {
-      RETURN_IF_CUDA_ERR(
-          cudaMemcpyAsync(dst, src, byte_size, copy_kind, cuda_stream),
-          msg + ": failed to perform CUDA copy");
-    }
+    RETURN_IF_CUDA_ERR(
+        cudaMemcpyAsync(dst, src, byte_size, cudaMemcpyDefault, cuda_stream),
```
GuanLuo, line 94: Note that I change cudaMemcpy to syntax that utilize UVM. Because UVM is supported after compute capability 2.0 so I assume the device that Triton runs on will have UVM support.

deadeyegoodwin:
@GuanLuo  I think you need to rebase onto master to fix the CI

PR #2425, opened 2021-01-16
Remove VERSION file to TRITON_VERSION

deadeyegoodwin:
This is needed because on windows (case insensitive file system) the
VERSION file was being picked up as the C++ <version> header which
causes build failures.


tanmayv25 (approved):
You mean standard library header <version>? Just curious where the conflict occurs.

deadeyegoodwin:
Apparently there can be a <version> include (perhaps optional for C++). Boost looks for it and on windows it finds our VERSION file because how we setup the include paths.

PR #2441, opened 2021-01-22
Additional windows testing

deadeyegoodwin:


CoderHam (dismissed):
I did not see the non-windows CIs (Jetson and x86 linux) run. I don't foresee any issues with them so I am approving the same.

PR #2444, opened 2021-01-22
For persistent backends, don't ever unload the corresponding shared library

deadeyegoodwin:


src/backends/backend/triton_backend_manager.h
```diff
@@ -131,6 +134,10 @@ class TritonBackend {
 
   // Opaque state associated with the backend.
   void* state_;
+
+  // Should be backend shared library be unloaded (e.g. dlclose) when
```
CoderHam, line 138: Should "the"?

src/core/persistent_backend_manager.cc
```diff
@@ -100,6 +100,11 @@ PersistentBackendManager::InitPersistentBackend(
     RETURN_IF_ERROR(TritonBackendManager::CreateBackend(
         backend_name, backend_dir, backend_libpath, *config, &persist_backend));
     persist_backends_.push_back(persist_backend);
+
+    // Show that persistent backend shared libraries should never be
```
CoderHam, line 104: I think "Show That" is not necessary

src/backends/backend/triton_backend_manager.cc
```diff
@@ -224,6 +226,25 @@ TritonBackend::~TritonBackend()
   LOG_STATUS_ERROR(UnloadBackendLibrary(), "failed unloading backend");
 }
 
+void
+TritonBackend::SetUnloadEnabled(const bool enable)
+{
+  // For memory leak debugging do not allow shared libraries to be
```
CoderHam, line 232: I believe this comment should still be under UnloadBackendLibrary()

CoderHam (changes requested):
A few small fixes else LGTM

src/backends/backend/triton_backend_manager.cc
deadeyegoodwin: It refers to the immediately following code. Will update the comment to be more clear.

CoderHam (approved):

PR #2445, opened 2021-01-23
Upgrade Google-Cloud-Cpp SDK, GRPC

CoderHam:
-   gRPC "v1.25.0" ->"v1.29.1"
-   google-cloud-cpp"v0.10..0" ->"v1.22.0"


build.py
```diff
@@ -593,6 +593,7 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
             unzip \
             wget \
             zlib1g-dev \
+            doxygen \
```
GuanLuo, line 596: Isn't doxygen only used for generating documentation?

CoderHam: Yes, I think it may be an optional decency, I will remove this

build.py
```diff
@@ -570,6 +570,8 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
             unzip \
             wget \
             zlib1g-dev \
+            libc-ares-dev \
```
deadeyegoodwin, line 573: We build c-ares in cmake and we explicitly use the built version there... so why do we need it here?

CoderHam: I was trying to apt-get install this instead of adding it as an External Project (because thats how google-cloud-cpp does it) removing it since I reverted back to the external project build in cmake.

deadeyegoodwin (approved):

PR #2447, opened 2021-01-25
fix : security context + resources limit

sinaure:
To deploy to k8s the deployment template should use security context in the spec

### DIFF
--- a/deploy/single_server/templates/deployment.yaml
+++ b/deploy/single_server/templates/deployment.yaml
@@ -47,6 +47,11 @@ spec:
         release: {{ .Release.Name }}
 
     spec:
+      {{- if .Values.securityContext.enabled }}
+      securityContext:
+        fsGroup:    {{ .Values.securityContext.userId }}
+        runAsUser:  {{ .Values.securityContext.groupId }}
+      {{- end }}  
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.imageName }}"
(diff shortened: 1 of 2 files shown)


deploy/single_server/templates/deployment.yaml
```diff
@@ -47,6 +47,8 @@ spec:
         release: {{ .Release.Name }}
 
     spec:
+      securityContext:
```
deadeyegoodwin, line 50: What is the difference in having securityContext here compared to below. Also, why is runAsUser not moved as well?

deploy/single_server/templates/deployment.yaml
```diff
@@ -66,16 +68,19 @@ spec:
             - containerPort: 8002
               name: metrics
           livenessProbe:
+            failureThreshold: 30
```
deadeyegoodwin, line 71: Adding failureThreshold, initialDelaySeconds and periodSeconds is good but add {{ .Values.image.livenessFailureThreshold }}, etc. variables to values.yaml and leave the defaults at 5. For initialDelaySeconds use default of 0 (to match k8s default).

deadeyegoodwin (changes requested):

deploy/single_server/templates/deployment.yaml
sinaure: securityContext should be optional (running it as userId 1000 throws exceptions while container starts).
tested with v1.20.2 of k8s the position of securityContext should be placed at spec level

Issue #2403, opened 2021-01-12
Can't Load Models Using Distributed MinIO


dpsommer:
## Description  

Shortly after upload, Triton will fail to load models from distributed MinIO.  

While I believe the fundamental error lies with MinIO or the underlying S3 implementation, Triton could work around the issue by consistently parsing object paths so as not to make requests with leading `/` characters in the `prefix` parameter.  

Link to the MinIO issue: https://github.com/minio/minio/issues/11265

## Triton Information  

*What version of Triton are you using?*  
**20.12**  

*Are you using the Triton container or did you build it yourself?*  
Triton container - **tritonserver:20.12-py3**  

## Steps to Reproduce  

Set up a distributed MinIO cluster (detailed steps can be found [here](https://docs.min.io/docs/deploy-minio-on-docker-compose.html)):  

```
$ docker-compose pull
Pulling minio1 ... done
Pulling minio2 ... done
Pulling minio3 ... done
Pulling minio4 ... done
Pulling nginx  ... done
$ docker-compose up
Creating network "minio_default" with the default driver
Creating volume "minio_data1-1" with default driver
Creating volume "minio_data1-2" with default driver
Creating volume "minio_data2-1" with default driver
Creating volume "minio_data2-2" with default driver
Creating volume "minio_data3-1" with default driver
Creating volume "minio_data3-2" with default driver
Creating volume "minio_data4-1" with default driver
Creating volume "minio_data4-2" with default driver
Creating minio_minio3_1 ... done
Creating minio_minio4_1 ... done
Creating minio_minio2_1 ... done
Creating minio_minio1_1 ... done
Creating minio_nginx_1  ... done
Attaching to minio_minio2_1, minio_minio1_1, minio_minio4_1, minio_minio3_1, minio_nginx_1
```

Create a bucket (here using the MinIO `mc` CLI)  

```
$ mc alias set localhost http://localhost:9000 minio minio123
$ mc mb localhost/testbucket
```

Upload any model (used [GoogleNet](https://github.com/onnx/models/tree/master/vision/classification/inception_and_googlenet/googlenet) from the ONNX model repository in testing)  

```
testbucket
    /googlenet
        config.pbtxt
        /1
            model.onnx
```

`config.pbtxt`
```
name: "googlenet"
platform: "onnxruntime_onnx"
input [
  {
    name: "data_0"
    data_type: TYPE_FP32
    dims: [ 1, 3, 224, 224 ]
  }
]
output [
  {
[...]

PR #2454, opened 2021-01-26
Ignore extra slashes in S3 path

CoderHam:
Ignores both extra trailing and leading slashes in S3 path
Resolves https://github.com/triton-inference-server/server/issues/2403#issuecomment-764859864

### DIFF
--- a/qa/L0_storage_S3/test.sh
+++ b/qa/L0_storage_S3/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2018-2021, NVIDIA CORPORATION. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -93,10 +93,11 @@ for ENV_VAR in "env" "env_dummy" "config"; do
 
     KIND="KIND_GPU"
 
-    for MAYBE_SLASH in "" "/"; do
+    # Test coverage for extra slashes 
+    for MAYBE_SLASH in "" "/" "//"; do
 
         ROOT_REPO="$BUCKET_URL$MAYBE_SLASH"
-        MODEL_REPO="${BUCKET_URL_SLASH}models${MAYBE_SLASH}"
+        MODEL_REPO="${BUCKET_URL}/${MAYBE_SLASH}models${MAYBE_SLASH}"
 
         # copy models in model directory
         rm -rf models && mkdir -p models
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -1141,6 +1141,7 @@ class S3FileSystem : public FileSystem {
  private:
   Status ParsePath(
       const std::string& path, std::string* bucket, std::string* object);
+  Status CleanPath(const std::string& path, std::string* clean_path);
   Aws::SDKOptions options_;
   s3::S3Client client_;
   re2::RE2 s3_regex_;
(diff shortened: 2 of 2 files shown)


src/core/filesystem.cc
```diff
@@ -1175,6 +1180,40 @@ S3FileSystem::ParsePath(
   return Status::Success;
 }
 
+Status
+S3FileSystem::CleanPath(const std::string& path, std::string* clean_path)
+{
+  // Remove extra slashes
+  *clean_path = "s3://";
+  int true_length = path.size();
+  for (size_t i = path.size() - 1; i++) {
```
deadeyegoodwin, line 1189: I think you want i--instead of i++

src/core/filesystem.cc
```diff
@@ -1150,19 +1151,23 @@ Status
...
+  RETURN_IF_ERROR(CleanPath(path, &clean_path));
+
   // Get the bucket name and the object path. Return error if input is malformed
   std::string host_name, host_port;
   if (!RE2::FullMatch(
-          path, s3_regex_, &host_name, &host_port, bucket, object)) {
-    int bucket_start = path.find("s3://") + strlen("s3://");
-    int bucket_end = path.find("/", bucket_start);
+          clean_path, s3_regex_, &host_name, &host_port, bucket, object)) {
+    int bucket_start = clean_path.find("s3://") + strlen("s3://");
```
deadeyegoodwin, line 1162: Isn't the path required to start with exactly "s3://"? Can't you just check this once at the beginning of the function and return error if fails, and thereafter just work with the rest of the path. That should also make it much easier to remove the leading and trainling /.

src/core/filesystem.cc
```diff
@@ -1175,6 +1180,40 @@ S3FileSystem::ParsePath(
...
+{
+  // Remove extra slashes
+  *clean_path = "s3://";
+  int true_length = path.size();
+  for (size_t i = path.size() - 1; i++) {
+    if (path[i] == '/') {
+      true_length -= 1;
+    }
+  }
+  std::string true_path = path.substr(0, true_length);
```
deadeyegoodwin, line 1194: This seems unnecessarily complex, can you just use something like this: https://www.techiedelight.com/trim-string-cpp-remove-leading-trailing-spaces/ except for '/' instead of whitespace

src/core/filesystem.cc
```diff
@@ -1175,6 +1180,40 @@ S3FileSystem::ParsePath(
...
 
+Status
+S3FileSystem::CleanPath(const std::string& path, std::string* clean_path)
+{
+  // Remove extra slashes
+  *clean_path = "s3://";
+  int true_length = path.size();
+  for (size_t i = path.size() - 1; i++) {
+    if (path[i] == '/') {
+      true_length -= 1;
```
deadeyegoodwin, line 1191: This is going do reduce length for internal slashes, isn't it? See comment below for an easier solution.

deadeyegoodwin (changes requested):

src/core/filesystem.cc
CoderHam: Yes

src/core/filesystem.cc
CoderHam: The function needs to convert **abc.com:9000/testbucket//object//aa/** ->**abc.com:9000/testbucket/object/aa**

deadeyegoodwin (commented):
Ok, but need some test additions to cover the additional cases.

CoderHam:
> Ok, but need some test additions to cover the additional cases.

Added test coverage and started a CI

qa/L0_storage_S3/test.sh
```diff
@@ -93,10 +93,11 @@ for ENV_VAR in "env" "env_dummy" "config"; do
 
     KIND="KIND_GPU"
 
-    for MAYBE_SLASH in "" "/"; do
+    # Test coverage for extra slashes 
+    for MAYBE_SLASH in "" "/" "//"; do
```
deadeyegoodwin, line 97: Don't we also need to test with multiple internal slashes? I thought we had tests with the model repo at the top-level (just the bucket) and also more deeply nested...

deadeyegoodwin (changes requested):

qa/L0_storage_S3/test.sh
CoderHam: the addition of `// ` does add multiple internal slashes
The urls are now: 
```
s3://triton-bucket-1211
s3://triton-bucket-1211/models
s3://triton-bucket-1211/
s3://triton-bucket-1211//models/
s3://triton-bucket-1211//
s3://triton-bucket-1211///models//
```

CoderHam: No we did not have tests that are deeper nested but when navigating the repo with the above s3 paths we are cleaning deeper urls as well since we use CleanPath in ParsePath as well

src/core/filesystem.cc
```diff
@@ -1150,19 +1151,23 @@ Status
 S3FileSystem::ParsePath(
     const std::string& path, std::string* bucket, std::string* object)
 {
-  // Get the bucket name and the object path. Return error if input is malformed
+  // Cleanup extra slashes
+  std::string clean_path;
+  LOG_STATUS_ERROR(CleanPath(path, &clean_path), "failed to parse S3 path");
```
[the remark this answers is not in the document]
CoderHam: Yes, fixed!

deadeyegoodwin (approved):

PR #2459, opened 2021-01-26
Add ModelRepositoryAgents to model configuration. Implement TritonRepoAgent API

GuanLuo:
cherry-picked changes from #2273


src/core/triton_repo_agent.h
```diff
@@ -0,0 +1,131 @@
...
+  const std::string name_;
+
+  // dlopen / dlsym handles
+  void* dlhandle_;
+  TritonRepoAgentInitFn_t init_fn_;
+  TritonRepoAgentFiniFn_t fini_fn_;
+  TritonRepoAgentModelActionFn_t model_action_fn_;
+};
+
+class TritonRepoAgentModel {
```
deadeyegoodwin, line 76: This should represent a single agent and corresponds to TRITONREPOAGENT_AgentModel. It should hold the state associated with the AgentModel (and the TRITONREPOAGENT API needs to be enhanced with SetState and State methods. The idea is that this AgentModel object persists as long as the model is loaded... Instead of having Initialize/Finalize the TRITONREPOAGENT_ModelAction will indicate its life-cycle. LOAD when it is created and UNLOAD indicates it is about to be destroyed (we need to add the UNLOAD).

deadeyegoodwin: We'll need something in the model repository manager to hold the list of these associated with a model.

deadeyegoodwin (changes requested):

src/core/filesystem.cc
```diff
@@ -1619,6 +1678,25 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
   return Status::Success;
 }
 
+Status
+GetFileSystemByType(FileSystemType type, FileSystem** file_system)
```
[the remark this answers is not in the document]
GuanLuo: I agree that the one that takes a path should internally invoke GetFileSystem(type, ...), but it is blocked because the constructor of some remote filesystems requires a path to be passed. That is why I asked if there are tests for multiple remote filesystems earlier.

src/core/triton_repo_agent.h
GuanLuo: I was thinking that the implementation of `TRITONREPOAGENT_AgentModel` would be holding the list of agents associated with the model and the specific parameters to be passed to the agents. In this way the model repository manager would be holding one "agent model" for one model.

From your description, it seems like you are expecting a "agent model" to be associated with only one agent and the model repository manager will hold a list of "agent model"s for one model?

I think both ideas are the same, just being represented differently

src/core/model_repository_manager.cc
```diff
@@ -1164,21 +1167,44 @@ ModelRepositoryManager::LoadModelByDependency()
...
-
       model_states.emplace_back(new ModelState(valid_model));
       auto model_state = model_states.back().get();
-      std::set<int64_t> versions;
-      Status status;
-      status = VersionsToLoad(
-          repository_path, valid_model->model_name_, valid_model->model_config_,
-          &versions);
+      const auto itr = infos_.find(valid_model->model_name_);
+      // FIXME clean up version logic and move agent invocation into
```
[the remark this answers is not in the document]
GuanLuo: It was supposed to be a separate enhancement. But it wouldn't be true if we want the lifecycle of the AgentModel to tie with the model, then it will be part of the BackendLifeCycle

src/core/triton_repo_agent.h
deadeyegoodwin: Within Triton core it may not matter so much but for the person writing an agent the AgentModel abstraction represents just one model using that agent (with associated user state). Since it needs to be that way for person writing the agent I'd prefer the abstractions within Triton to be the same.

src/core/triton_repo_agent.cc
```diff
@@ -0,0 +1,479 @@
...
+    TRITONREPOAGENT_Agent* agent, TRITONREPOAGENT_AgentModel* model,
+    const char* location)
+{
+  TritonRepoAgentModel* tam = reinterpret_cast<TritonRepoAgentModel*>(model);
+  RETURN_TRITONSERVER_ERROR_IF_ERROR(tam->DeleteMutableLocation());
+  return nullptr;  // success
+}
+
+TRITONSERVER_Error*
+TRITONREPOAGENT_ModelRepositoryLocationUpdate(
```
deadeyegoodwin, line 393: *_ModelRepostoryUpdate(), no "location"

deadeyegoodwin (changes requested):

GuanLuo:
Added end-to-end testing using checksum repo agent

qa/L0_repoagent_checksum/test.sh
```diff
@@ -0,0 +1,93 @@
...
+    fi
+else
+    echo -e "\n***\n*** Expect fail to start $SERVER\n***"
+    cat $SERVER_LOG
+    kill $SERVER_PID
+    wait $SERVER_PID
+    exit 1
+fi
+
+# Set correct md5sum
```
deadeyegoodwin, line 59: Can you also checksum the config.pbtxt as a test when there are multiple checksums

deadeyegoodwin (changes requested):

qa/L0_repoagent_checksum/test.sh
GuanLuo: I added another file to be checked because the checksum is written in the config.pbtxt which has circular dependency to calculate its checksum

deadeyegoodwin (approved):

PR #2465, opened 2021-01-28
Use Data Center GPU Manager (DCGM) to get metrics instead of NVML

CoderHam:
- TODO fix windows build for the same

### DIFF
--- a/src/core/metrics.cc
+++ b/src/core/metrics.cc
@@ -35,7 +35,7 @@
 
 #ifdef TRITON_ENABLE_METRICS_GPU
 #include <cuda_runtime_api.h>
-#include <nvml.h>
+#include <dcgm_agent.h>
 #endif  // TRITON_ENABLE_METRICS_GPU
 
 namespace nvidia { namespace inferenceserver {
@@ -122,10 +122,10 @@ Metrics::Metrics()
 Metrics::~Metrics()
 {
 #ifdef TRITON_ENABLE_METRICS_GPU
-  // Signal the nvml thread to exit and then wait for it...
-  if (nvml_thread_ != nullptr) {
-    nvml_thread_exit_.store(true);
-    nvml_thread_->join();
+  // Signal the dcgm thread to exit and then wait for it...
+  if (dcgm_thread_ != nullptr) {
+    dcgm_thread_exit_.store(true);
+    dcgm_thread_->join();
   }
 #endif  // TRITON_ENABLE_METRICS_GPU
 }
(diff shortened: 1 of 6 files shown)


src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
 #else
-  nvmlReturn_t nvmlerr = nvmlInit();
-  if (nvmlerr != NVML_SUCCESS) {
-    LOG_WARNING << "failed to initialize, GPU metrics will not be available: "
-                << nvmlErrorString(nvmlerr);
+  dcgmReturn_t dcgmerr = dcgmInit();
+  if (dcgmerr != DCGM_ST_OK) {
+    LOG_WARNING
+        << "failed to initialize DCGM, GPU metrics will not be available: "
+        << errorString(dcgmerr);
```
deadeyegoodwin, line 175: Is this a dcgm-specific error string method?

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
-      LOG_WARNING << "failed to get device properties for device " << didx
-                  << ", GPU metrics will not be available for this device: "
-                  << cudaGetErrorString(cudaerr);
-      continue;
-    }
+  dcgmerr = dcgmUpdateAllFields(dcgm_handle_, 1);
+  if (dcgmerr != DCGM_ST_OK) {
+    LOG_WARNING << "DCGM failed to update all fields, GPU metrics will not "
+                   "be available: "
+                << errorString(dcgmerr);
```
deadeyegoodwin, line 190: Follow consistent pattern, report both message and number as above

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
 
-  int dcnt;
-  cudaError_t cudaerr = cudaGetDeviceCount(&dcnt);
-  if (cudaerr != cudaSuccess) {
-    LOG_WARNING
-        << "failed to get device count, GPU metrics will not be available: "
-        << cudaGetErrorString(cudaerr);
+  dcgmerr = dcgmStartEmbedded(DCGM_OPERATION_MODE_MANUAL, &dcgm_handle_);
+  if (dcgmerr != DCGM_ST_OK) {
+    LOG_WARNING << "Error: dcgmStartEmbedded returned \""
```
deadeyegoodwin, line 181: Follow lowercase pattern... "error, dcgm..."

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
-      LOG_WARNING << "failed to get Bus ID for device " << didx
-                  << ", GPU metrics will not be available for this device: "
-                  << cudaGetErrorString(cudaerr);
-      continue;
-    }
+  unsigned int all_gpu_ids[DCGM_MAX_NUM_DEVICES];
+  int gpu_count;
+  dcgmerr = dcgmGetAllDevices(dcgm_handle_, all_gpu_ids, &gpu_count);
+  if (dcgmerr != DCGM_ST_OK) {
+    LOG_WARNING << "failed to get device info and count, GPU metrics will not "
```
deadeyegoodwin, line 198: "error, unable to get ..."

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
-        cudaDeviceGetPCIBusId(pcibusid_str, sizeof(pcibusid_str) - 1, didx);
-    if (cudaerr != cudaSuccess) {
-      LOG_WARNING << "failed to get Bus ID for device " << didx
-                  << ", GPU metrics will not be available for this device: "
-                  << cudaGetErrorString(cudaerr);
-      continue;
-    }
+  unsigned int all_gpu_ids[DCGM_MAX_NUM_DEVICES];
+  int gpu_count;
+  dcgmerr = dcgmGetAllDevices(dcgm_handle_, all_gpu_ids, &gpu_count);
```
deadeyegoodwin, line 196: Does DCGM only return the cuda visible devices so the filtering is not needed?

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
+                << errorString(dcgmerr);
+    return false;
+  }
 
-    nvmlDevice_t gpu;
-    nvmlReturn_t nvmlerr = nvmlDeviceGetHandleByPciBusId_v2(pcibusid_str, &gpu);
-    if (nvmlerr != NVML_SUCCESS) {
-      LOG_WARNING << "failed to get device from Bus ID "
+  // Get DCGM metrics for each GPU. Some devices may have problems using DCGM
+  // API and thus device count/ids needs to be updated.
```
deadeyegoodwin, line 205: I don't think this comment is quite right... we are just collecting the gpus that support dcgm and can only report metrics for those gpus

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
+  std::vector<uint32_t> available_gpu_ids;
+  dcgmDeviceAttributes_t gpu_attributes[DCGM_MAX_NUM_DEVICES];
+  for (int i = 0; i < gpu_count; i++) {
+    gpu_attributes[i].version = dcgmDeviceAttributes_version;
+    dcgmerr = dcgmGetDeviceAttributes(
+        dcgm_handle_, all_gpu_ids[i], &gpu_attributes[i]);
+    if (dcgmerr != DCGM_ST_OK) {
+      LOG_WARNING << "failed to get device properties for device "
```
deadeyegoodwin, line 213: failed -> error

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
+          dcgmReturn_t dcgmerr = dcgmGetLatestValuesForFields(
+              handle, available_gpu_ids[didx], fields, field_count,
+              field_values);
+          if (dcgmerr != DCGM_ST_OK) {
+            power_limit_fail_cnt[didx]++;
+            power_usage_fail_cnt[didx]++;
+            energy_fail_cnt[didx]++;
+            util_fail_cnt[didx]++;
+            mem_fail_cnt[didx]++;
+            LOG_WARNING << "failed to get field values for device "
```
deadeyegoodwin, line 280: failed -> error

src/core/metrics.cc
```diff
@@ -156,205 +156,215 @@ Metrics::EnableGPUMetrics()
...
+            // Power limit
+            if (power_limit_fail_cnt[didx] < fail_threshold) {
+              unsigned int power_limit = field_values[0].value.i64;
+              if ((field_values[0].status == DCGM_ST_OK) &&
+                  (!DCGM_INT64_IS_BLANK(power_limit))) {
+                power_limit_fail_cnt[didx] = 0;
+              } else {
+                power_limit_fail_cnt[didx]++;
+                power_limit = 0;
+                LOG_WARNING << "failed to get power limit for GPU " << didx
```
deadeyegoodwin, line 293: failed -> error

src/test/custom/identity/CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ if(${TRITON_ENABLE_GPU})
 target_include_directories(identity PRIVATE ${CUDA_INCLUDE_DIRS})
 target_link_libraries(
   identity
-  PUBLIC ${CUDA_nvml_LIBRARY}
```
deadeyegoodwin, line 55: Does this really need to link against DCGM (or previously NVML)?

deadeyegoodwin (changes requested):

src/core/metrics.cc
CoderHam: Yes that is correct

src/core/metrics.cc
CoderHam: Fixed to -> "... thus these devices needs to be ignored."

src/core/metrics.cc
CoderHam: Yes, DCGM will report a list of all visible devices but it is possible dcgmGetDeviceAttributes is unable to get the properties for some of those devices

src/test/custom/identity/CMakeLists.txt
CoderHam: No, reverted to NVML

src/core/metrics.cc
deadeyegoodwin: To be clear, by visible devices I mean only those visisble by CUDA due to CUDA_VISIBLE_DEVICES (and NVIDIA_VISIBLE_DEVICES in containers)

src/test/custom/identity/CMakeLists.txt
deadeyegoodwin: You shouldn't need to link either. We no longer use nvml at all, and I don't know why the legacy custom backend was linking against it in the first place...

CoderHam:
Closing for new #2924

PR #2499, opened 2021-02-09
Update repoagent documentation

deadeyegoodwin:


deadeyegoodwin:
@GuanLuo read carefully the triton steps listed in this doc, especially the forward and reverse orders. I don't think this matches the implementation so we may need to discuss.

docs/repository_agents.md
```diff
@@ -0,0 +1,168 @@
...
+
+Your repository agent must implement the repository agent API as
+documented in
+[tritonrepoagent.h](https://github.com/triton-inference-server/core/tree/main/include/triton/core/tritonrepoagent.h).
+
+Triton follows these steps when loading a model:
+
+* Load the model's configuration file (config.pbtxt) and extract the
+  *ModelRepositoryAgents* settings. Even if a repository agent
+  modifies the config.pbtxt file, the repository agent settings from
```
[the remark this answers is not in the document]
deadeyegoodwin: Yes, initially the repository manager should only need the repoagent part of the initial config to determine which (if any) repo agents are specified for the model. After invoking all the repoagents with LOAD successfully then the repository manager should treat the final repository location as if it was directly presented by the user. So config.pbtxt should be used, etc.

I will file a ticket for the reverse order change

docs/repository_agents.md
```diff
@@ -0,0 +1,168 @@
...
+    successfully or TRITONREPOAGENT_ACTION_LOAD_FAIL if the model
+    failed to load.
+
+Triton follows these steps when unloading a model:
+
+* Triton uses the repository agent settings from the initial
+  config.pbtxt file, even if during loading one or more agents
+  modified its contents.
+
+* For each repository agent specified, in reverse order:
```
deadeyegoodwin, line 156: @GuanLuo  I'm going to change this line to "For each repository agent that was invoked with TRITONREPOAGENT_ACTION_LOAD:"
So UNLOAD action will be in forward order, the same as LOAD action

docs/repository_agents.md
```diff
@@ -0,0 +1,168 @@
...
+
+  * Triton invokes the repository agent's
+    *TRITONREPOAGENT_ModelAction* function with action
+    TRITONREPOAGENT_ACTION_UNLOAD.
+
+* Triton unloads the model.
+
+* For each repository agent specified, in reverse order:
```
deadeyegoodwin, line 164: @GuanLuo  I'm changing to: 
"For each repository agent that was invoked with TRITONREPOAGENT_ACTION_UNLOAD, in reverse order:"

PR #2504, opened 2021-02-10
Fix expected test calculation

CoderHam:


qa/L0_infer/test.sh
```diff
@@ -39,7 +39,6 @@ export CUDA_VISIBLE_DEVICES=0
 
 CLIENT_LOG_BASE="./client"
 INFER_TEST=infer_test.py
-EXPECTED_NUM_TESTS=${EXPECTED_NUM_TESTS:="42"}
```
deadeyegoodwin, line 42: It seems like we should keep this so that we have a default value for EXPECTED

deadeyegoodwin (changes requested):

qa/L0_infer/test.sh
CoderHam: It is unnecessary since we have that case covered here:
https://github.com/triton-inference-server/server/blob/605d4aa72b80f3289a4078813fcb91d02fa2ce1d/qa/L0_infer/test.sh#L63-L67

deadeyegoodwin (approved):

PR #2505, opened 2021-02-10
Install repoagents into /opt/tritonserver for container build

deadeyegoodwin:


GuanLuo (changes requested):
Since repo agents are now only in global search path, you can remove the copy of checksum agent to L0_repoagent_checksum directory in `dockerfile.QA`

deadeyegoodwin:
> 
> 
> Since repo agents are now only in global search path, you can remove the copy of checksum agent to L0_repoagent_checksum directory in `dockerfile.QA`

@GuanLuo That is in the commit, unless I missed a case.

GuanLuo (approved):
My bad, I missed that part

PR #2507, opened 2021-02-10
Modify build.py to support zero container build on jetson

CoderHam:


build.py
```diff
@@ -254,7 +253,8 @@ def core_cmake_args(components, backends, install_dir):
             else:
                 fail('unknown core backend {}'.format(be))
 
-    cargs.append('-DTRITON_EXTRA_LIB_PATHS=/opt/tritonserver/lib')
```
deadeyegoodwin, line 257: I don't think we need TRITON_EXTRA_LIB_PATHS anymore. Can you try removing it. We should also not need --opt-dir flag.

build.py
```diff
@@ -264,6 +264,8 @@ def core_cmake_args(components, backends, install_dir):
 
     if platform.system() == 'Windows':
```
deadeyegoodwin, line 265: We should generalize the target platform. Can you replace --jetson-build flag with --target-platform arg that takes either "ubuntu", "windows" or "jetpack". After parsing if it is not specified then call a function to automatically determine it. For windows the test is already elsewhere in this file. For jetson is there a similar test we can do. It would get good to auto-detect as much as possible so that users don't actually need to specify --target-platform. Replace current windows test in all places with a test against FLAGS.target_platform.

build.py
```diff
@@ -264,6 +264,8 @@ def core_cmake_args(components, backends, install_dir):
...
         cargs.append('c:/workspace/build')
+    if FLAGS.jetson_build:
+        cargs.append(FLAGS.cmake_dir)
```
deadeyegoodwin, line 268: --cmake-dir is a good addition, but use it to get rid of all these hardwired paths. The in CI find our other uses of build.py and pass in the appropriate value.

build.py
```diff
@@ -380,22 +384,25 @@ def tensorrt_cmake_args():
...
 
 def tensorflow_cmake_args(ver, images):
-    # If a specific TF image is specified use it, otherwise pull from
-    # NGC.
-    image_name = "tensorflow{}".format(ver)
-    if image_name in images:
-        image = images[image_name]
+    # If jetson build is specified do not use docker images
+    if FLAGS.jetson_build:
+        opt_dir = "/opt" if FLAGS.opt_dir is None else FLAGS.opt_dir
```
deadeyegoodwin, line 389: Instead of --opt-dir add a new options that are like --image in that they provide a way to give library and include paths for certain parts of the build. Add --library-paths and --include-paths that are used in the same way as --image. For example --library-paths=tensorflow1:/opt/tritonserver/backends/tensorflow1

build.py
```diff
@@ -1062,10 +1081,11 @@ def container_build(backends, images):
         log('image "{}": "{}"'.format(parts[0], parts[1]))
         images[parts[0]] = parts[1]
 
-    # If --container-build is specified then we perform the actual
+    # If --no-container-build or jetson_build is not specified then we perform the actual
```
deadeyegoodwin, line 1084: Your jetson build should just explicitly add --no-container-build

deadeyegoodwin (changes requested):

build.py
CoderHam: Yes, will do!

build.py
CoderHam: For now we do not have any include paths that we need to specify in the build, we can add them in as needed. Added the library-paths arg and remove opt-dir

build.py

build.py

build.py
```diff
@@ -375,31 +373,38 @@ def onnxruntime_cmake_args():
...
-def tensorflow_cmake_args(ver, images):
-    # If a specific TF image is specified use it, otherwise pull from
-    # NGC.
-    image_name = "tensorflow{}".format(ver)
-    if image_name in images:
-        image = images[image_name]
+def tensorflow_cmake_args(ver, images, library_paths):
+    backend_name = "tensorflow{}".format(ver)
+
+    # If platform is jetpack build is specified do not use docker images
```
deadeyegoodwin, line 387: extra_args = [ ]

build.py
```diff
@@ -857,6 +862,13 @@ def container_build(images, backends, repoagents):
         action="store_true",
         required=False,
         help='Do not use Docker --pull argument when building container.')
+    parser.add_argument(
+        '--target-platform',
+        required=False,
+        default=None,
+        help=
+        'Target Platform, can be "ubuntu", "windows" or "jetpack". If not specified, auto-detect the same.'
```
deadeyegoodwin, line 870: "Target for build, can be ... If not specified, build targets the current platform"

build.py
```diff
@@ -875,9 +887,22 @@ def container_build(images, backends, repoagents):
...
     parser.add_argument(
         '--install-dir',
+        type=str,
         required=False,
         default=None,
         help='Install directory, default is <builddir>/opt/tritonserver.')
+    parser.add_argument('--cmake-dir',
+                        type=str,
+                        required=True,
+                        help='CMake directory. Folder for CMakeLists.txt.')
```
deadeyegoodwin, line 897: "Directory containing the CMakeLists.txt file for Triton server."

build.py
```diff
@@ -875,9 +887,22 @@ def container_build(images, backends, repoagents):
...
+    parser.add_argument(
+        '--library-paths',
+        action='append',
+        required=False,
+        default=None,
+        help=
+        'Include specified library paths for respective backends in build as <backend-name>[:<library_path>].'
```
deadeyegoodwin, line 904: "Specify library paths for ..."

build.py
```diff
@@ -1071,11 +1098,20 @@ def container_build(images, backends, repoagents):
         log('image "{}": "{}"'.format(parts[0], parts[1]))
         images[parts[0]] = parts[1]
 
+    # Initialize map of repo agents to build and repo-tag for each.
```
deadeyegoodwin, line 1101: Fix comment

build.py
```diff
@@ -1071,11 +1098,20 @@ def container_build(images, backends, repoagents):
...
+    for lpath in FLAGS.library_paths:
+        parts = lpath.split(':')
+        if len(parts) == 2:
+            log('backend "{}" library path "{}"'.format(parts[0], parts[1]))
+            library_paths[parts[0]] = parts[1]
+
     # If --container-build is specified then we perform the actual
-    # build within a build container and then from that create a
-    # tritonserver container holding the results of the build.
+    # then we perform the actual build within a build container and
```
deadeyegoodwin, line 1110: repeated phrase. Revert to original text.

docs/build.md
```diff
@@ -55,7 +55,7 @@ By default no Triton features are enabled. The following build.py
...
+$ ./build.py --build-dir=/tmp/citritonbuild --enable-logging --enable-stats --enable-tracing --enable-metrics --enable-gpu-metrics --enable-gpu --filesystem=gcs --filesystem=azure_storage --filesystem=s3 --endpoint=http --endpoint=grpc --repo-tag=common:<container tag> --repo-tag=core:<container tag> --repo-tag=backend:<container tag> --backend=custom --backend=ensemble --backend=tensorrt --backend=identity:<container tag> --backend=repeat:<container tag> --backend=square:<container tag> --backend=onnxruntime:<container tag> --backend=pytorch:<container tag> --backend=tensorflow1:<container tag> --backend=tensorflow2:<container tag> --backend=python:<container tag> --backend=dali:<container tag> --cmake-dir=/workspace/build
```
deadeyegoodwin, line 58: --cmake-dir=<path/to/repo>/build

deadeyegoodwin (changes requested):

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #2510, opened 2021-02-11
Enhance BackendInputCollector to use CUDA kernel for gathering device-accessible data to device buffer

GuanLuo:


src/core/CMakeLists.txt
```diff
@@ -151,6 +151,15 @@ if(${TRITON_ENABLE_S3})
...
 endif()
 
+if(${TRITON_ENABLE_GPU})
+  set(CUDA_LIBRARIES PUBLIC ${CUDA_LIBRARIES})
+  cuda_add_library(
+    gather_kernel 
+    kernel.cu kernel.h
+    STATIC
+  )
+endif() # TRITON_ENABLE_GPU
```
deadeyegoodwin, line 161: set_property(TARGET gather-kernel PROPERTY CUDA_PTX_COMPILATION ON)

src/core/CMakeLists.txt
```diff
@@ -151,6 +151,15 @@ if(${TRITON_ENABLE_S3})
   message(STATUS "Using aws-sdk-cpp ${AWSSDK_VERSION}")
 endif()
 
+if(${TRITON_ENABLE_GPU})
+  set(CUDA_LIBRARIES PUBLIC ${CUDA_LIBRARIES})
+  cuda_add_library(
+    gather_kernel
```
deadeyegoodwin, line 157: naming convention, use dash: "gather-kernel"

src/core/CMakeLists.txt
```diff
@@ -240,6 +250,9 @@ add_library(
   ${SERVER_SRCS} ${SERVER_HDRS}
 )
 add_dependencies(server-library proto-library)
+if(${TRITON_ENABLE_GPU})
+  add_dependencies(server-library gather_kernel)
```
deadeyegoodwin, line 254: can you use "target-link-libraries server-library gather-kernel"
and then I don't think you need the change in src/servers/CMakeLists.txt

src/core/backend_context.h
```diff
@@ -295,9 +306,18 @@ class BackendInputCollector {
   size_t pending_pinned_offset_;
   RequestsList pending_pinned_inputs_;
 
-  // Pinned memories that need to live over the lifetime of this
-  // BackendResponder object.
-  std::list<std::unique_ptr<AllocatedMemory>> pinned_memories_;
+  // Allocated memories that need to live over the lifetime of this
+  // BackendInputCollector object.
+  std::list<std::unique_ptr<AllocatedMemory>> in_used_memories_;
```
deadeyegoodwin, line 311: in_use_memories_

deadeyegoodwin (changes requested):

src/core/CMakeLists.txt
GuanLuo: It doesn't work so I still link it in `tritonserver` target. But I remove the dependency here because the symbol will be resolved in `tritonserver`

src/core/CMakeLists.txt
GuanLuo: I change the name to a more general name `kernel-library`, I can imagine we may try to expose the similar feature in output side (i.e. scatter kernel)

deadeyegoodwin (approved):

Issue #2488, opened 2021-02-04
Using TensorRT acceleration with model running on both CPU and GPU


jishminor:
**Description**
When specifying a tensorflow graphdef model configuration with instance group for both CPU and GPU, and applying tensorrt gpu acceleration accelerator, Triton complains about the configuration being invalid, with error message:
`E0204 21:14:55.078078 1 model_repository_manager.cc:963] failed to load 'inceptionv3' version 1: Invalid argument: GPU Execution Accelerator can only be set on non-CPU backend context`

**Triton Information**
What version of Triton are you using?
2.6.0

Are you using the Triton container or did you build it yourself?
Built myself for Jetson AGX Xavier Jetpack 4.4.1

**To Reproduce**
Steps to reproduce the behavior.

Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).
```bash
name: "inceptionv3"
platform: "tensorflow_graphdef"
max_batch_size: 8
input [
    {
    name: "input"
    data_type: TYPE_FP32
    dims: [ 299, 299, 3 ]
    }    
]
output [
    {
    name: "InceptionV3/Predictions/Softmax"
    data_type: TYPE_FP32
    dims: [ 1001 ]
    }
]
instance_group [
    {
    count: 2
    kind: KIND_GPU
    }
    ,
    {
    count: 3
    kind: KIND_CPU
    }
]
optimization { execution_accelerators {
    gpu_execution_accelerator : [ {
    name : "tensorrt"
    parameters { key: "precision_mode" value: "FP32" }}]
}}
dynamic_batching {
    preferred_batch_size: [ 3 ]
}
```

**Expected behavior**
TensorRT optimizations should only be applied to GPU instances.

deadeyegoodwin:
The error needs to be changed into a warning to support this use case. We will look into that change.

CoderHam:
We should not be using the 'gpu_execution_accelerator" on CPU devices. It is ignored and an appropriate error is logged in the aforementioned case. Fix: https://github.com/triton-inference-server/tensorflow_backend/pull/14

PR #2516, opened 2021-02-12
Fix to check for warning increase of failure of gpu exec acc on CPU device

CoderHam:


CoderHam:
Fix test to be compatible with PR: https://github.com/triton-inference-server/tensorflow_backend/pull/14 that fixes a bug where GPU exec accelerators are being specified for models that run on CPU: https://github.com/triton-inference-server/server/issues/2488

deadeyegoodwin (approved):

PR #2524, opened 2021-02-17
Fix np.object uninteded cast in Python client

Tabrizian:


qa/L0_string_io/string_client_test.py
```diff
@@ -64,22 +64,22 @@ def test_tf_unicode_bytes(self):
...
+                b'\n3\n\x13\n\x01a\x12\x0e\x1a\x0c\n\n<\xe2\x05\x8a\x01\xb3\x07?\xfd\x01\n\r\n\x01b\x12\x08\x1a\x06\n\x04\xf6\xa2\xc5\x01\n\r\n\x01c\x12\x08\x12\x06\n\x04\x1b\x931\xbf\x00\x00\x00'
             ],
             [
-                b'\n&\n\x07\n\x01a\x12\x02\x1a\x00\n\x0c\n\x01b\x12\x07\x1a\x05\n\x03\x89\xcc=\n\r\n\x01c\x12\x08\x12\x06\n\x04{\xbc\x0e>'
+                b'\n&\n\x07\n\x01a\x12\x02\x1a\x00\n\x0c\n\x01b\x12\x07\x1a\x05\n\x03\x89\xcc=\n\r\n\x01c\x12\x08\x12\x06\n\x04{\xbc\x0e>\x00'
             ],
             [
-                b'\nF\n\'\n\x01a\x12"\x1a \n\x1e\x97\x01\x93\x02\x9e\x01\xac\x06\xff\x01\xd8\x05\xe1\x07\xd8\x04g]\x9a\x05\xff\x06\xde\x07\x8f\x04\x97\x04\xda\x03\n\x0c\n\x01b\x12\x07\x1a\x05\n\x03\x9a\xb7I\n\r\n\x01c\x12\x08\x12\x06\n\x04\xfb\x87\x83\xbf'
+                b'\nF\n\'\n\x01a\x12"\x1a \n\x1e\x97\x01\x93\x02\x9e\x01\xac\x06\xff\x01\xd8\x05\xe1\x07\xd8\x04g]\x9a\x05\xff\x06\xde\x07\x8f\x04\x97\x04\xda\x03\n\x0c\n\x01b\x12\x07\x1a\x05\n\x03\x9a\xb7I\n\r\n\x01c\x12\x08\x12\x06\n\x04\xfb\x87\x83\xbf\x00'
```
deadeyegoodwin, line 79: We should test some strings that don't end in /x00

qa/L0_backend_python/python_test.py
```diff
@@ -265,7 +265,7 @@ def test_unicode(self):
         with httpclient.InferenceServerClient("localhost:8000") as client:
             utf8 = '😀'
             input_data = np.array([bytes(utf8, encoding='utf-8')],
-                                  dtype=np.bytes_)
+                                  dtype=np.object_)
```
deadeyegoodwin, line 268: Does bytes_ not work anymore? If so that would break backwards compat. If both work then should have a test for both.

deadeyegoodwin, line 268: What is the usage required? Using np.object in all cases doesn't work, at least not based on findings here: https://jirasw.nvidia.com/browse/DLIS-1955?focusedCommentId=2855169&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2855169

From that it seems like using np.object will a binary str will cause it to be serialized as the string representation of the binary, which is not what is wanted. The fix is to use np.bytes but that will string ending zeros so shouldn't it really be np.void. We need some testing for all these cases and to make sure that the wire representation of these is the actual binary data (or correct string data) and not some other representation.

deadeyegoodwin (changes requested):

qa/L0_backend_python/python_test.py
Tabrizian: It works, if it doesn't end with `\x00`. I will add tests in `L0_string_io` to test for both cases.

src/clients/python/library/tritonclient/utils/__init__.py
```diff
@@ -186,14 +185,14 @@ def serialize_byte_tensor(input_tensor):
...
-                                             == np.bytes_):
+    if (input_tensor.dtype == np.object) or \
+       (input_tensor.dtype.type == np.bytes_):
         flattened = bytes()
         for obj in np.nditer(input_tensor, flags=["refs_ok"], order='C'):
             # If directly passing bytes to BYTES type,
             # don't convert it to str as Python will encode the
             # bytes which may distort the meaning
-            if obj.dtype.type == np.bytes_:
+            if obj.dtype.type == np.bytes_ or obj.dtype.type == np.object_:
```
Tabrizian, line 195: @deadeyegoodwin The problem that you mentioned with `np.object_` is fixed here.

qa/L0_backend_python/python_test.py
Tabrizian: I understand the issue in DLIS-1955. The problem there is also addressed as a part of this PR. I originally tried to use the `np.void` as we discussed. But the problem with `np.void` was that it raised the error below after serialization:

```
Invalid type promotion with void datatypes of different lengths. Use the `np.bytes_` datatype instead to pad the shorter value with trailing zero bytes.
```

The problem in DLIS-1955 is fixed in the comment I added below. I will add more tests to L0_string_io to make sure all different variations work.

qa/L0_string_io/string_client_test.py
```diff
@@ -92,12 +90,36 @@ def test_tf_unicode_bytes(self):
...
-                                      model_version=model_version)
+                                      outputs=outputs)
 
+        out0 = results.as_numpy('OUTPUT0')
         # We expect there to be 1 results (with batch-size 1). Verify
         # that all 8 result elements are the same as the input.
-        self.assertTrue(np.array_equal(in0, results.as_numpy('OUTPUT0')))
+        self.assertTrue(np.array_equal(in0, out0))
+
+        in0_object = in0.astype(np.object)
```
deadeyegoodwin, line 100: So this is testing that np.object works for both strings and binary blobs?
Does np.bytes_ work for both cases as well? Can you test that?

deadeyegoodwin (changes requested):
Doesn't the title of this PR need to be changed?

qa/L0_string_io/string_client_test.py
Tabrizian: `|S78` is like `np.bytes_`. I'll add a test case to be more explicit.

deadeyegoodwin (dismissed):

src/clients/python/library/tritonclient/utils/__init__.py
```diff
@@ -159,25 +159,27 @@ def triton_to_np_dtype(dtype):
 
 def serialize_byte_tensor(input_tensor):
     """
-        Serializes a bytes tensor into a flat numpy array of length prepend bytes.
-        Can pass bytes tensor as numpy array of bytes with dtype of np.bytes_,
-        numpy strings with dtype of np.str_ or python strings with dtype of np.object.
+    Serializes a bytes tensor into a flat numpy array of length prepended bytes.
+    Can pass bytes tensor as numpy array of bytes with dtype of np.bytes_,
```
deadeyegoodwin, line 163: We want to recommend only np.object for both strings and byte blobs. We can also say "A numpy array of dtype np.str_ or np.bytes_ can also be used but are not recommended because numpy will remove trailing zeros at the end of the byte sequence for those types"

deadeyegoodwin (changes requested):
Don't need to rerun CI

deadeyegoodwin (approved):

PR #2529, opened 2021-02-18
Add client memory leak and memory growth tests

CoderHam:
- L0_client_valgrind
- L0_client_memory_growth

Only with C API for now, Will add python client tests in a follow up PR


qa/L0_client_valgrind/test.sh
```diff
@@ -0,0 +1,104 @@
...
+RET=0
+
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+# Run test for both HTTP and GRPC, synchronous and asynchronous. 1000 inferences each.
```
deadeyegoodwin, line 67: sync and async -> reuse and no reuse ?

src/clients/c++/library/grpc_client.cc
```diff
@@ -46,8 +46,12 @@ namespace {
 
 // Use map to keep track of GRPC channels. <key, value> : <url, Channel*>
 // If context is created on url that has established Channel, then reuse it.
-std::map<std::string, std::shared_ptr<grpc::Channel>> grpc_channel_map_;
```
deadeyegoodwin, line 49: Can you put the fix in a different PR than the testing.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #2547, opened 2021-02-23
Add option to control the use of CUDA gather kernel

GuanLuo:


src/core/backend_context.cc
```diff
@@ -1219,11 +1219,19 @@ BackendInputCollector::FlushPendingCopyKernel(
...
   }
-  auto status = LaunchCopyKernel(
-      tensor_buffer, tensor_buffer_byte_size, tensor_memory_type,
-      tensor_memory_type_id);
-  bool cuda_copy = status.IsOk();
 
+  bool cuda_copy = false;
+  Status status = Status(Status::Code::INTERNAL, "");
+  // empirical value of the minimum buffer count
+  static size_t MIN_BUFFER_COUNT = 64;
```
CoderHam, line 1226: Would this not be GPU/case dependent? What is the delta in performance we are seeing with this?
If so we should allow the user to control this parameter via a command line argument.

GuanLuo: I agree that this should be exposed as a model config option. This is a quick fix that has minimum code change on both the Triton and the harness around it, I will see if there is time to make the change more robust

GuanLuo:
The gather kernel will be turned off by default as more results showed that it only helped on limited use case.

deadeyegoodwin (approved):

PR #2550, opened 2021-02-23
Invoke agent model load action at polling so the final location is treated as if it was directly presented by the user

GuanLuo:


deadeyegoodwin (commented):
Need to add some test where the repo agent changes the config.pbtxt and make sure the final one is used. Also a test where agent doesn't make any update to the repo and make sure that initial repo config.pbtxt is used.

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #2551, opened 2021-02-23
Fix trailing zeros for type BYTES

Tabrizian:
Backward compatibility notes:

* All the shared memory operations on string need to be updated to fix the number of bytes calculation (we can include the function used in the `infer_util` in the Python package to facilitate this `get_number_of_bytes_for_npobject`).
* If a user code relied on the encoding to `utf-8` be performed in the client code for `np.object` that will no longer be the case (It was merged in the previous PR).
* The serialized tensor will be `np.object` instead of `np.bytes`.
* `np.bytes_` still works for regular inference (without shared memory).

### DIFF
--- a/qa/L0_backend_identity/identity_test.py
+++ b/qa/L0_backend_identity/identity_test.py
@@ -28,8 +28,6 @@
 
 import argparse
 import numpy as np
-import os
-import re
 import sys
 import requests as httpreq
 from builtins import range
@@ -160,7 +158,7 @@ if __name__ == '__main__':
     with client_util.InferenceServerClient(FLAGS.url,
                                            verbose=FLAGS.verbose) as client:
         for model_name, np_dtype, shape in (
-            # yapf: disable
+                # yapf: disable
             ("identity_fp32", np.float32, [1, 0]),
             ("identity_fp32", np.float32, [1, 5]),
             ("identity_uint32", np.uint32, [4, 0]),
@@ -173,8 +171,9 @@ if __name__ == '__main__':
                 input_data = (16384 * np.random.randn(*shape)).astype(np_dtype)
             else:
                 in0 = (16384 * np.ones(shape, dtype='int'))
-                in0n = np.array([str(x) for x in in0.reshape(in0.size)],
-                                dtype=object)
+                in0n = np.array(
+                    [str(x).encode('utf-8') for x in in0.reshape(in0.size)],
+                    dtype=object)
                 input_data = in0n.reshape(in0.shape)
             inputs = [
                 client_util.InferInput("INPUT0", input_data.shape,
@@ -190,9 +189,6 @@ if __name__ == '__main__':
             if output_data is None:
                 print("error: expected 'OUTPUT0'")
                 sys.exit(1)
-
-            if np_dtype == object:
-                output_data = np.char.decode(output_data)
 
             if not np.array_equal(output_data, input_data):
                 print("error: expected output {} to match input {}".format(
--- a/qa/L0_infer_reshape/infer_reshape_test.py
+++ b/qa/L0_infer_reshape/infer_reshape_test.py
@@ -308,7 +308,7 @@ class InferReshapeTest(tu.TestResultCollector):
         self._full_reshape(np.int32, input_shapes=([1, 4, 1], [8], [2, 2, 3]))
 
     def test_oo1(self):
-        self._full_reshape(np.object, input_shapes=([1],), no_batch=False)
+        self._full_reshape(np.object_, input_shapes=([1],), no_batch=False)
 
 
 if __name__ == '__main__':
--- a/qa/common/infer_util.py
+++ b/qa/common/infer_util.py
@@ -25,7 +25,6 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import sys
-import os
 import numpy as np
 import test_util as tu
 import shm_util as su
@@ -63,7 +62,7 @@ def _range_repr_dtype(dtype):
         return np.int16
     elif dtype == np.float16:
         return np.int8
-    elif dtype == np.object:  # TYPE_STRING
+    elif dtype == np.object_:  # TYPE_STRING
         return np.int32
     return dtype
 
@@ -157,7 +156,7 @@ def infer_exact(tester,
                                      high=val_max,
                                      size=tensor_shape,
                                      dtype=rinput_dtype)
-    if input_dtype != np.object:
+    if input_dtype != np.object_:
         input0_array = input0_array.astype(input_dtype)
         input1_array = input1_array.astype(input_dtype)
 
@@ -168,7 +167,7 @@ def infer_exact(tester,
         output0_array = input0_array - input1_array
         output1_array = input0_array + input1_array
 
-    if output0_dtype == np.object:
+    if output0_dtype == np.object_:
         output0_array = np.array([
             unicode(str(x), encoding='utf-8') for x in (output0_array.flatten())
         ],
@@ -183,18 +182,22 @@ def infer_exact(tester,
     else:
         output1_array = output1_array.astype(output1_dtype)
 
-    if input_dtype == np.object:
-        in0n = np.array(
-            [str(x) for x in input0_array.reshape(input0_array.size)],
-            dtype=object)
+    if input_dtype == np.object_:
+        in0n = np.array([
+            str(x).encode('utf-8')
+            for x in input0_array.reshape(input0_array.size)
+        ],
+                        dtype=object)
         input0_array = in0n.reshape(input0_array.shape)
-        in1n = np.array(
-            [str(x) for x in input1_array.reshape(input1_array.size)],
-            dtype=object)
+        in1n = np.array([
+            str(x).encode('utf-8')
+            for x in input1_array.reshape(input1_array.size)
+        ],
+                        dtype=object)
         input1_array = in1n.reshape(input1_array.shape)
 
     # prepend size of string to output string data
-    if output0_dtype == np.object:
+    if output0_dtype == np.object_:
         if batch_size == 1:
             output0_array_tmp = serialize_byte_tensor_list([output0_array])
         else:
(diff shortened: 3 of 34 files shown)


src/clients/python/examples/simple_grpc_shm_string_client.py
```diff
@@ -36,6 +36,30 @@
 
 FLAGS = None
 
+
+def get_number_of_bytes_for_npobject(tensor_value):
```
Tabrizian, line 40: This is the function that we can potentially include in the Triton Python client.

src/clients/python/library/tritonclient/utils/cuda_shared_memory/__init__.py
```diff
@@ -176,10 +176,17 @@ def set_shared_memory_region(cuda_shm_handle, input_values):
     offset_current = 0
     for input_value in input_values:
         input_value = np.ascontiguousarray(input_value).flatten()
-        byte_size = input_value.size * input_value.itemsize
-        _raise_if_error(
-            c_int(_ccudashm_shared_memory_region_set(cuda_shm_handle, c_uint64(offset_current), \
-                c_uint64(byte_size), input_value.ctypes.data_as(c_void_p))))
+        if input_value.dtype == np.object_:
+            input_value = input_value.item()
+            byte_size = np.dtype(np.byte).itemsize * len(input_value)
```
Tabrizian, line 181: @deadeyegoodwin / @tanmayv25 Is it ok if I pass the pointer to a temporary variable? I was not sure about the valid duration of a pointer in python.

qa/L0_infer_reshape/infer_reshape_test.py
```diff
@@ -308,7 +308,7 @@ def test_ii3(self):
         self._full_reshape(np.int32, input_shapes=([1, 4, 1], [8], [2, 2, 3]))
 
     def test_oo1(self):
-        self._full_reshape(np.object, input_shapes=([1],), no_batch=False)
+        self._full_reshape(np.object_, input_shapes=([1],), no_batch=False)
```
deadeyegoodwin, line 311: Why do we use "object_" here but elsewhere use "object"?

qa/common/infer_util.py
```diff
@@ -75,6 +75,16 @@ def serialize_byte_tensor_list(tensor_values):
     return tensor_list
 
 
+def get_number_of_bytes_for_npobject(tensor_value):
+    if tensor_value.size > 0:
```
deadeyegoodwin, line 79: You should verify that the tensor dtype is np.object

src/clients/python/examples/simple_grpc_shm_string_client.py
```diff
@@ -36,6 +36,30 @@
...
+    ----------
+    tensor_value : numpy.ndarray
+        Numpy array to calculate the number of bytes for.
+
+    Returns
+    -------
+    int
+        Number of bytes present in this tensor
+    """
+    if tensor_value.size > 0:
```
deadeyegoodwin, line 54: raise exception is not np.object

src/clients/python/library/tritonclient/utils/__init__.py
```diff
@@ -122,7 +122,7 @@ def np_to_triton_dtype(np_dtype):
         return "FP32"
     elif np_dtype == np.float64:
         return "FP64"
-    elif np_dtype == np.object or np_dtype.type == np.bytes_:
+    elif np_dtype == np.object_ or np_dtype.type == np.bytes_:
```
deadeyegoodwin, line 125: Does this break backwards compat?

qa/L0_backend_identity/identity_test.py
```diff
@@ -173,8 +171,9 @@
                 input_data = (16384 * np.random.randn(*shape)).astype(np_dtype)
             else:
                 in0 = (16384 * np.ones(shape, dtype='int'))
-                in0n = np.array([str(x) for x in in0.reshape(in0.size)],
-                                dtype=object)
+                in0n = np.array(
+                    [str(x).encode('utf-8') for x in in0.reshape(in0.size)],
```
deadeyegoodwin, line 175: Why is explicit utf-8 encoding needed? Is it because python add-sub is expecting that?

deadeyegoodwin (changes requested):

src/clients/python/examples/simple_grpc_shm_string_client.py
CoderHam: Do you mean `tritonclient.utils`?

qa/common/shm_util.py
```diff
@@ -52,15 +53,26 @@ def create_set_shm_regions(input0_list, input1_list, output0_byte_size,
                            output1_byte_size, outputs, shm_region_names,
                            precreated_shm_regions, use_system_shared_memory,
                            use_cuda_shared_memory):
+    from infer_util import get_number_of_bytes_for_npobject
```
CoderHam, line 55: Why not include this at the top? after `from tritonclient.utils import *`

qa/common/shm_util.py
```diff
@@ -52,15 +53,26 @@ def create_set_shm_regions(input0_list, input1_list, output0_byte_size,
...
     if use_system_shared_memory and use_cuda_shared_memory:
         raise ValueError(
             "Cannot set both System and CUDA shared memory flags to 1")
 
     if not (use_system_shared_memory or use_cuda_shared_memory):
         return [], []
 
-    input0_byte_size = sum([i0.nbytes for i0 in input0_list])
-    input1_byte_size = sum([i1.nbytes for i1 in input1_list])
+    if input0_list[0].dtype == np.object:
```
CoderHam: should this be `np.object` or `np.object_`?

src/clients/python/library/tritonclient/http/__init__.py
```diff
@@ -1402,7 +1402,11 @@ def set_data_from_numpy(self, input_tensor, binary_data=True):
         else:
             self._data = None
             if self._datatype == "BYTES":
-                self._raw_data = serialize_byte_tensor(input_tensor).tobytes()
+                serialized_output = serialize_byte_tensor(input_tensor)
+                if serialized_output.size > 0:
```
CoderHam, line 1406: We already have a check in `serialize_byte_tensor()`:
```
    if input_tensor.size == 0:
        return np.empty([0])
```
Why not just modify that as needed?

CoderHam (changes requested):

src/clients/python/library/tritonclient/utils/__init__.py
Tabrizian: I don't think it breaks the backward compat. I'll add a test for this in `L0_string_io`. The reason that I changed this in every location is that `np.object` is deprecated and should be replaced with `np.object_`. It looks like it is only a convention that `numpy` is trying to follow. All the types that their equivalent type exists in Python are superseded with an underscore.

qa/L0_backend_identity/identity_test.py
Tabrizian: This is not Python add_sub. This is the L0_backend_identity test. It was required before too but it was performed in the client code. Previously, whenever the type of the ndarray was `np.object` it was encoded using UTF-8. The client does not require string to be encoded in UTF-8 but this is what the model expects. Also, because the HTTP client needs the final string to be JSON serializable, it requires it to be UTF-8 encoded (or encodable). But in the client code itself, we don't require the byte array to be encoded in UTF-8.

qa/common/shm_util.py
Tabrizian: It creates a circular dependency. infer_util -> shm_util -> infer_util

src/clients/python/examples/simple_grpc_shm_string_client.py
Tabrizian: Yes

qa/common/shm_util.py
Tabrizian: I think you have commented on the previous commit. It is fixed in the newer commit.

qa/L0_infer_reshape/infer_reshape_test.py
Tabrizian: I tried to use `np.object_` everywhere. `np.object_` and `object` are interchangeable.

src/clients/python/library/tritonclient/utils/cuda_shared_memory/__init__.py
deadeyegoodwin: I assume you are talking about 'input_value'. You do have to be careful that the python object lifetime is longer than any use you have in C code. But in this case it seems that you haven't changed that compared to what was before, correct? Did you change _ccudashm_shared_memory_region_set?

Tabrizian: No, I didn't change the `_ccudashm_shared_memory_region_set`. I just used `input_value.item()` instead.

deadeyegoodwin: Then should be ok. I believe that function just copies data out of input_value.item and into shared memory. So it doesn't keep reference to item() pointer.

Tabrizian: I see.

src/clients/python/examples/simple_grpc_shm_string_client.py
CoderHam: I agree. But then we should change the name to serialized_byte_size and make it generic for all both np.object and non-np.object type. If there a reason we should have this for a specific tensor value instead of a tenor of values?

qa/common/shm_util.py
CoderHam: All the more reason for it to be moved to tritonclient.utils

src/clients/python/library/tritonclient/http/__init__.py
Tabrizian: `serialize_byte_tensor` is used in other places too. That's why I had to do it here.

qa/common/shm_util.py
Tabrizian: I moved it to the tritonclient.utils.

src/clients/python/examples/simple_grpc_shm_string_client.py
Tabrizian: I changed the name to `serialized_byte_size` but I guess because of the name, it shouldn't work for other types. I have raised an exception if a different type is provided.

deadeyegoodwin (approved):

PR #2561, opened 2021-02-25
Fix non-GPU build

GuanLuo:


src/core/backend_context.cc
```diff
@@ -1241,6 +1254,7 @@ BackendInputCollector::FlushPendingCopyKernel(
     const TRITONSERVER_MemoryType tensor_memory_type,
     const int64_t tensor_memory_type_id)
 {
+  LOG_ERROR << "before launching gather kernel";
```
CoderHam, line 1257: Did you mean to add these temporarily for logging. I think they should be removed.

src/core/backend_context.cc
```diff
@@ -1254,6 +1268,7 @@ BackendInputCollector::FlushPendingCopyKernel(
         tensor_buffer, tensor_buffer_byte_size, tensor_memory_type,
         tensor_memory_type_id);
     cuda_copy = status.IsOk();
+    LOG_ERROR << "gather kernel launched with status: " << status.AsString();
```
CoderHam, line 1271: Same here.

CoderHam (changes requested):

CoderHam (approved):

deadeyegoodwin:
@GuanLuo  It seems like that extra logging for every request could have had an impact of perf. What do you think?

GuanLuo:
That is debug logging for checking PTX.. Removed from this PR

deadeyegoodwin (approved):

PR #2569, opened 2021-02-27
Add SSL/TLS Mutual Authentication support for GRPC

CoderHam:


src/servers/main.cc
```diff
@@ -308,6 +309,9 @@ std::vector<Option> options_
        "request/response objects."},
       {OPTION_GRPC_USE_SSL, "grpc-use-ssl", Option::ArgBool,
        "Use SSL authentication for GRPC requests. Default is false."},
+      {OPTION_GRPC_USE_SSL_MUTUAL, "grpc-use-ssl-mutual", Option::ArgBool,
+       "Use mututal SSL authentication for GRPC requests. Default is false. If "
+       "true, set --grpc-use-ssl to true."},
```
deadeyegoodwin, line 314: I would remove the "If true..." sentence. Internally we will set options as necessary.

src/servers/main.cc
```diff
@@ -385,7 +389,8 @@ std::vector<Option> options_
        "The global directory searched for backend shared libraries. Default is "
        "'/opt/tritonserver/backends'."},
       {OPTION_REPOAGENT_DIR, "repoagent-directory", Option::ArgStr,
-       "The global directory searched for repository agent shared libraries. Default is "
+       "The global directory searched for repository agent shared libraries. "
+       "Default is "
```
deadeyegoodwin, line 393: merge this line with next

deadeyegoodwin (changes requested):
Don't need to rerun CI

deadeyegoodwin (approved):

PR #2577, opened 2021-03-02
Add python backend to nomodel perf benchmark

CoderHam:
- Replace legacy custom model

### DIFF
--- a/qa/L0_perf_nomodel/run_test.sh
+++ b/qa/L0_perf_nomodel/run_test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019-2021, NVIDIA CORPORATION. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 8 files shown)


qa/L0_perf_nomodel/run_test.sh
```diff
@@ -81,9 +84,15 @@ for BACKEND in $BACKENDS; do
...
     fi
 
-    # set naming (special case for libtorch model)
-    INPUT_NAME="INPUT0" && [ $BACKEND == "libtorch" ] && INPUT_NAME="INPUT__0"
-
+    # set input name (special case for libtorch and python model)
+    if [ $BACKEND == "libtorch" ]; then
+        INPUT_NAME="INPUT__0"
+    elif [ $BACKEND == "python" ]; then
+        INPUT_NAME="IN"
```
Tabrizian, line 91: You can rename "IN" to "INPUT0" so that this if condition is not required.

Tabrizian (changes requested):

qa/L0_perf_nomodel/run_test.sh
CoderHam: Solved in follow up PR. Is there a reason the python backend uses a different naming convention? I think we should keep the naming consistent i.e. INPUT0 and OUTPUT0 instead of IN and OUT

Tabrizian: Great, no there is not any reason. My bad, I should be more consistent in the naming of inputs and outputs for QA models.

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #2580, opened 2021-03-03
Added failure testing for L0_secure_grpc

CoderHam:


deadeyegoodwin (approved):

qa/L0_secure_grpc/test.sh
```diff
@@ -90,15 +94,72 @@ for CASE in server mutual both; do
...
+        echo -e "\n***\n*** Expected test failure\n***"
+    else
         RET=1
     fi
 
-    $TEST_CLIENT -v --ssl --root-certificates ca.crt --private-key client.key --certificate-chain client.crt >> ${CLIENT_LOG}.c++.ssl_infer 2>&1
+    $TEST_CLIENT -v >> ${CLIENT_LOG}.${CASE}.no_ssl_fail_infer 2>&1
     if [ $? -ne 0 ]; then
-        cat ${CLIENT_LOG}.c++.ssl_infer
+        cat ${CLIENT_LOG}.${CASE}.c++.no_ssl_fail_infer
```
deadeyegoodwin, line 143: The CI is showing a couple of cases where the "cat" is failing because the filename is not correct. After fixing you can just test locally, no need to rerun the entire CI

deadeyegoodwin (changes requested):

qa/L0_secure_grpc/test.sh
CoderHam: Fixed with last commit

PR #2585, opened 2021-03-03
Build gather kernel with cross device portability. Enable kernel build for Windows and Jetson

GuanLuo:


src/core/CMakeLists.txt
```diff
@@ -152,13 +152,19 @@ if(${TRITON_ENABLE_S3})
...
+  if (WIN32)
+    set(HOST_COMPILER_FLAGS "/MD")
+  endif()
+
   set(CUDA_LIBRARIES PUBLIC ${CUDA_LIBRARIES})
   cuda_add_library(
     kernel-library 
     kernel.cu kernel.h
+    OPTIONS -arch compute_60
+    OPTIONS -code compute_60,sm_60,sm_61,sm_62,sm_70,sm_72,sm_75
```
deadeyegoodwin, line 165: Ampere is sm_80 and sm_86.

deadeyegoodwin (changes requested):

src/core/CMakeLists.txt
CoderHam: Some of the jetson machines have cuda compute capability under 6.0
Nano and TX1 have 5.3
We can either add them in optionally or customize this based on the min cuda compute flag from build.py

deadeyegoodwin: Oh yes, add sm_53 as well.

GuanLuo: So lower the compute to 53?

I added sm_80 before but the Jetson build complains about it, it seems like it can't recognize it yet.

deadeyegoodwin: Yes, it should be sufficient to use compute_53

deadeyegoodwin (approved):

PR #2589, opened 2021-03-03
Add openvino to nomodel perf benchmark

CoderHam:
- Only batch size 1 tests for now

### DIFF
--- a/qa/L0_perf_nomodel/run_test.sh
+++ b/qa/L0_perf_nomodel/run_test.sh
@@ -90,7 +90,8 @@ for BACKEND in $BACKENDS; do
     MAX_LATENCY=300
     MAX_BATCH=${STATIC_BATCH} && [ $DYNAMIC_BATCH > $STATIC_BATCH ] && MAX_BATCH=${DYNAMIC_BATCH}
 
-    # openvino model does also not support batching/dynamic shapes
+    # TODO Add openvino identity model that supports batching/dynamic batching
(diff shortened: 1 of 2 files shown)


deadeyegoodwin (dismissed):

qa/L0_perf_nomodel/run_test.sh
```diff
@@ -90,14 +90,24 @@ for BACKEND in $BACKENDS; do
     MAX_LATENCY=300
     MAX_BATCH=${STATIC_BATCH} && [ $DYNAMIC_BATCH > $STATIC_BATCH ] && MAX_BATCH=${DYNAMIC_BATCH}
 
+    # openvino model does also not support batching/dynamic shapes
```
tanmayv25, line 93: Not all the layers support dynamic batch sizes. But I think we can still support nomodel.
See this for supported layers: https://docs.openvinotoolkit.org/latest/openvino_docs_IE_DG_DynamicBatching.html#limitations

tanmayv25: Just fix the comment for factual accuracy. We can address the case in later PRs.

deadeyegoodwin (approved):

qa/L0_perf_nomodel/run_test.sh
CoderHam: Fixed

PR #2590, opened 2021-03-04
Add OpenVINO documentation

deadeyegoodwin:


tanmayv25:
Don't we need to mention the OV backend is still in release beta?

deadeyegoodwin:
> 
> 
> Don't we need to mention the OV backend is still in release beta somewhere?

I'll make sure that is clearly stated in the OV backend README, and we will mention it in the release notes

PR #2602, opened 2021-03-09
Fix TRT backend to capture timestamp as soon as possible

GuanLuo:
The timestamp captured for each phase is more realistic:
resnet before:
`Avg request latency: 4254 usec (overhead 2 usec + queue 72 usec + compute input 920 usec + compute infer 3244 usec + compute output 16 usec)`
resnet after:
`Avg request latency: 4348 usec (overhead 2 usec + queue 73 usec + compute input 312 usec + compute infer 3930 usec + compute output 31 usec)`

But there is additional overhead introduced by the fix, which is obvious for small model (end-to-end < 1ms):
plan_zero_1_float32 before:
`Avg request latency: 101 usec (overhead 1 usec + queue 13 usec + compute input 76 usec + compute infer 7 usec + compute output 4 usec)`
after:
`Avg request latency: 130 usec (overhead 3 usec + queue 13 usec + compute input 58 usec + compute infer 23 usec + compute output 33 usec)`

### DIFF
--- a/src/backends/tensorrt/plan_backend.cc
+++ b/src/backends/tensorrt/plan_backend.cc
@@ -594,8 +594,10 @@ PlanBackend::CreateExecutionContext(
   const int cuda_stream_priority =
       GetCudaStreamPriority(Config().optimization().priority());
   RETURN_IF_ERROR(context->CreateCudaStream(cuda_stream_priority));
+#ifdef TRITON_ENABLE_STATS
   RETURN_IF_ERROR(context->CreateCudaStream(
       cuda_stream_priority, &context->signal_stream_));
+#endif  // TRITON_ENABLE_STATS
   RETURN_IF_ERROR(context->CreateCudaStream(
       cuda_stream_priority, &context->input_copy_stream_));
   if (separate_output_stream) {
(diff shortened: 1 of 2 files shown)


GuanLuo:
Reverted part of the change to still let `ProcessResponse()` check "input consumed event". This is to prevent the `ProcessResponse()` backlog in the edge case where the responses take time to be finalized and to be sent: consider a case where model execution time < response thread's turn around time, decoupling the check from response thread will cause the backend keep taking batches from scheduler while the response thread is backlogged.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -575,6 +594,8 @@ PlanBackend::CreateExecutionContext(
   const int cuda_stream_priority =
       GetCudaStreamPriority(Config().optimization().priority());
   RETURN_IF_ERROR(context->CreateCudaStream(cuda_stream_priority));
+  RETURN_IF_ERROR(context->CreateCudaStream(
```
deadeyegoodwin, line 597: TRITON_ENABLE_STATS

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3202,6 +3233,9 @@ PlanBackend::Context::InitEventSet(bool busy_wait_events)
     RETURN_IF_ERROR(CreateCudaEvent(
         "Set " + std::to_string(idx) + " output ready", event_flags,
         &events_[idx].output_ready_));
+    RETURN_IF_ERROR(CreateCudaEvent(
```
deadeyegoodwin, line 3236: TRITON_ENABLE_STATS

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #2615, opened 2021-03-12
Add simple  java client test

CoderHam:
Test both java and scala client and add validation for outputs in both client examples


qa/L0_client_java/test.sh
```diff
@@ -0,0 +1,91 @@
...
+# Build simple java client example 
+cd src/clients/java/examples && mvn clean install
+
+# Test simple java client example
+mvn exec:java -Dexec.mainClass=clients.SimpleJavaClient -Dexec.args="localhost 8001" >> $CLIENT_LOG 2>&1
+if [ $? -ne 0 ]; then
+    cat $CLIENT_LOG
+    RET=1
+fi
+
```
deadeyegoodwin, line 79: Also run the scala test.
Can you update both  the java and scala code to check the response to make sure the correct result is received.

deadeyegoodwin (changes requested):

qa/L0_client_java/test.sh
CoderHam: Yes I can add checks to both and add the scala test as well

CoderHam: Done, triggered a new CI

deadeyegoodwin (approved):

PR #2619, opened 2021-03-12
Remove deprecated numpy usages from client

jbkyang-nvi:
This PR removes the [deprecated numpy usages](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) from the codebase. 
Specifically `numpy.bool` usages and fixes one mis-use of `numpy.object`.

# Verified that deprecation warnings are gone

### DIFF
--- a/qa/L0_backend_python/python_test.py
+++ b/qa/L0_backend_python/python_test.py
@@ -123,7 +123,7 @@ class PythonTest(tu.TestResultCollector):
     def test_bool(self):
         model_name = 'identity_bool'
         with httpclient.InferenceServerClient("localhost:8000") as client:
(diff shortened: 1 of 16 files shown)


Tabrizian (commented):
Can you make similar changes to Python backend too?

https://github.com/triton-inference-server/python_backend/blob/main/src/resources/triton_python_backend_utils.py#L59

https://github.com/triton-inference-server/python_backend/blob/main/src/resources/triton_python_backend_utils.py#L31

CoderHam (approved):
@jbkyang-nvi yes, you could create a PR with similar fixes for the python_backend as well. I will let you know how to run the CI for that as well. 
These changes look good to me.

dzier (approved):

PR #2628, opened 2021-03-16
Support compression in gRPC transactions

tanmayv25:
Uses per call compression only for inference requests. Using compression for non-infer related requests is not a good idea.

client-side compression:
`I0316 12:05:30.316294326    7617 message_compress_filter.cc:265] Compressed[deflate] 204 bytes vs. 102 bytes (50.00% savings)`

server-side compression:
`I0316 12:50:35.188782589   10734 message_compress_filter.cc:265] Compressed[deflate] 187 bytes vs. 99 bytes (47.06% savings)`

performance for sure seems to suffer. The numbers are collected on RN50 saved model:

```
No Compression: Concurrency: 1, throughput: 69 infer/sec, latency 14488 usec
Deflate: Concurrency: 1, throughput: 27.8 infer/sec, latency 35990 usec
Gzip: Concurrency: 1, throughput: 27.8 infer/sec, latency 35957 usec
```

### DIFF
--- a/src/clients/python/examples/simple_grpc_infer_client.py
+++ b/src/clients/python/examples/simple_grpc_infer_client.py
@@ -80,7 +80,7 @@ if __name__ == '__main__':
         help='File holding PEM-encoded certicate chain. Default is None.')
     parser.add_argument(
         '-C',
-        '--compression_algorithm',
+        '--grpc-compression-algorithm',
         type=str,
         required=False,
         default=None,
@@ -129,7 +129,7 @@ if __name__ == '__main__':
         outputs=outputs,
         client_timeout=FLAGS.client_timeout,
         headers={'test': '1'},
-        compression_algorithm=FLAGS.compression_algorithm)
+        compression_algorithm=FLAGS.grpc_compression_algorithm)
 
     statistics = triton_client.get_inference_statistics(model_name=model_name)
     print(statistics)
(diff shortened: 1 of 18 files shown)


CoderHam:
Can you try this with a large payload size to see if it makes a difference for large payloads?

src/clients/python/examples/simple_grpc_infer_client.py
```diff
@@ -78,6 +78,15 @@
         required=False,
         default=None,
         help='File holding PEM-encoded certicate chain. Default is None.')
+    parser.add_argument(
+        '-C',
+        '--compression_algorithm',
```
CoderHam, line 83: Since this is just for GRPC maybe call it `--grpc-compression-algorithm` instead?

CoderHam (changes requested):

src/clients/python/examples/simple_grpc_infer_client.py
tanmayv25: The client is itself grpc hence I omitted grpc here... But I can make this change for uniformity with perf_analyzer ( and perf_analyzer in the future)

CoderHam: Yes, I agree that having uniform flags will help 😄

tanmayv25:
@CoderHam More than the tensor size, we will see improvements with larger concurrency so that the time spent in compressing and decompressing the messages are overlapped. The above perf numbers are for RN50 model with input size ~4MB.

tanmayv25:
For the add sub model at a higher concurrency we see similar perf as the compression/decompression time is pipelined:
```
Deflate: Concurrency: 50, throughput: 2457.2 infer/sec, latency 20346 usec
Gzip: Concurrency: 50, throughput: 2494.4 infer/sec, latency 20044 usec
None: Concurrency: 50, throughput: 2461.6 infer/sec, latency 20308 usec
```

CoderHam (approved):

PR #2630, opened 2021-03-17
Add generation scripts for OpenVino models (No variable shapes)

CoderHam:
~https://github.com/openvinotoolkit/openvino/issues/4798 prevents this from working till the next release i.e. 2021.3~
Moving forward to Openvino 2021.4 
Openvino does not support variable shape tensors 
- This impacts batch as well as variable shape [-1] models such as identity models.
- Hence for now while we generate all non-variable models but enable just L0_infer

PS: Bug filed against Openvino for multi dims models with different I/O dtypes such as addsub with inputs of FP32 dtype and outputs of INT32 dtype
https://github.com/openvinotoolkit/openvino/issues/7173


CoderHam:
Update: Openvino https://github.com/openvinotoolkit/openvino/releases/tag/2021.3 is out but must wait on Onnxruntime to move forward to pick up the same

deadeyegoodwin (approved):

tanmayv25 (approved):

PR #2639, opened 2021-03-18
Improve Cloud Storage Filesystem error reporting

CoderHam:
- Restructure S3 to report error regarding s3 path during initialization

### DIFF
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -1161,6 +1161,7 @@ class S3FileSystem : public FileSystem {
  public:
   S3FileSystem(const Aws::SDKOptions& options, const std::string& s3_path);
   ~S3FileSystem();
+  Status IntializeAndCheckClient();
   Status FileExists(const std::string& path, bool* exists) override;
   Status IsDirectory(const std::string& path, bool* is_dir) override;
   Status FileModificationTime(
@@ -1275,6 +1276,11 @@ S3FileSystem::S3FileSystem(
                              "s3://([0-9a-zA-Z\\-.]+):([0-9]+)/"
                              "([0-9a-z.\\-]+)(((/[0-9a-zA-Z.\\-_]+)*)?)")
 {
+}
+
+Status
+S3FileSystem::IntializeAndCheckClient()
+{
   Aws::Client::ClientConfiguration config;
   Aws::Auth::AWSCredentials credentials;
 
@@ -1299,9 +1305,9 @@ S3FileSystem::S3FileSystem(
     config = Aws::Client::ClientConfiguration("default");
   }
 
-  // Cleanup extra slashes
+  // Cleanup extra slashes, return error if path is invalid
   std::string clean_path;
-  LOG_STATUS_ERROR(CleanPath(s3_path, &clean_path), "failed to parse S3 path");
+  RETURN_IF_ERROR(CleanPath(s3_path, &clean_path));
 
   std::string host_name, host_port, bucket, object;
   if (RE2::FullMatch(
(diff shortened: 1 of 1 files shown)


src/core/filesystem.cc
```diff
@@ -895,7 +899,9 @@ Status
 ASFileSystem::CheckClient()
 {
   if (client_ == nullptr) {
-    return Status(Status::Code::INTERNAL, "blob client initialize failed.");
+    return Status(
+        Status::Code::INTERNAL,
+        "Unable to create Azure filesystem client. Check account credentials.");
```
tanmayv25, line 904: Just curious, if the client_ is nullptr then does that definitely means problem with account credentials? If not, can we get finer details of the error of why the client creation failed?

tanmayv25 (changes requested):

src/core/filesystem.cc
```diff
@@ -909,6 +915,12 @@ ASFileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   RETURN_IF_ERROR(ParsePath(path, &container, &object_path));
 
   auto blobProperty = bc.get_blob_property(container, object_path);
+  if (!blobProperty.valid()) {
+    return Status(
+        Status::Code::INTERNAL,
+        "Unable to get file modification time for " + path);
```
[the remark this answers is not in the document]
CoderHam: Done

src/core/filesystem.cc
CoderHam: The client is null if not initialized. This could be for many reasons. It is not possible to get the exact reason for the failure. Meanwhile even when credentials are valid (need not be correct for that specific blob storage) the client will be initialized correctly.

tanmayv25 (dismissed):

src/core/filesystem.cc
```diff
@@ -1342,14 +1342,9 @@ S3FileSystem::IntializeAndCheckClient(const std::string& s3_path)
...
-  bool is_dir;
-  RETURN_IF_ERROR(IsDirectory(clean_path, &is_dir));
-  if (!is_dir) {
-    return Status(
-        Status::Code::INVALID_ARG, "No bucket/folder found at " + clean_path +
-                                       ". Check account credentials.");
-  }
+  // Verify that the path exists and 'client_' is initialized correctly.
+  bool exists;
+  RETURN_IF_ERROR(FileExists(clean_path, &exists));
```
tanmayv25, line 1347: > if (!head_object_outcome.IsSuccess()) {
>     return Status(
>         Status::Code::INTERNAL,
>         "Could not get MetaData for object at " + path + " due to exception: " +
>             head_object_outcome.GetError().GetExceptionName() +
>             ", error message: " + head_object_outcome.GetError().GetMessage());
>   } else {
>     *exists = true;
>   }
I believe we don't need exists output parameter here. Either the `exists` is set true or the FileExists returns an erroneous status.

CoderHam: We need to return that the file exists. The file can be a directory,. The flow is such:
1.  Check if Path is a valid Path (Can be parsed correctly)
2. Check if path is a directory. If directory then exists is set to True and return success, else it is set to False. 
3. Reach here if not a directory. Check if path is a file. If so, exists is set to True.

src/core/filesystem.cc
```diff
@@ -670,8 +672,8 @@ GCSFileSystem::ReadTextFile(const std::string& path, std::string* contents)
 
   if (!stream) {
     return Status(
-        Status::Code::INTERNAL,
-        "Failed to open object read stream for " + path);
+        Status::Code::INTERNAL, "Failed to open object read stream for " +
+                                    path + " : " + stream.status().message());
```
deadeyegoodwin, line 676: "stream" is defined even if !stream ?

CoderHam: Yes. I followed the examples [here](https://github.com/googleapis/google-cloud-cpp/blob/master/google/cloud/storage/examples/storage_bucket_samples.cc) and [here](https://googleapis.dev/cpp/google-cloud-storage/latest/).

deadeyegoodwin (dismissed):

tanmayv25 (dismissed):

deadeyegoodwin (approved):

PR #2645, opened 2021-03-18
Added dependency for Model Analyzer Reports

aramesh7:
Model Analyzer's report generation depends on `pdfkit` which depends on `wkhtmltopdf`

### DIFF
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -59,7 +59,8 @@ RUN apt-get update && \
             python3-dev \
             rapidjson-dev \
             vim \
-            wget && \
+            wget \
+            python3-pdfkit && \
     pip3 install --upgrade wheel setuptools && \
     pip3 install --upgrade grpcio-tools && \
     pip3 install --upgrade pip


Dockerfile.sdk
```diff
@@ -59,7 +59,8 @@ RUN apt-get update && \
             python3-dev \
             rapidjson-dev \
             vim \
-            wget && \
+            wget 
+            wkhtmltopdf && \
```
CoderHam, line 63: Shouldn't this be `python3-pdfkit`?
Additionally the same can be removed from **L0_kibana_summary/test.sh**

CoderHam (changes requested):

Dockerfile.sdk
aramesh7: I see I will change this

Tabrizian: I think we install pdfkit as the dependency in `requiremenets.txt`. If `wkhtmltopdf ` is not required, we can probably remove it from the MA docker file too.

CoderHam: For the kibana summary I did not need to manually install wkhtmltopdf. It was installed with python-pdfkit as a dep. as I recall.

CoderHam: @aramesh7 can you create a gitlab PR to remove this install from the kibana summary's test.sh ?

aramesh7: Done

CoderHam (approved):

Dockerfile.sdk
CoderHam: I do not see an MR for this on Gitlab

PR #2646, opened 2021-03-19
Added python script to check the valgrind logs, removed bash function

aramesh7:


tanmayv25:
Why is this needed? To clarify, I do prefer python script to a bash function. Trying to understand the motivation.

aramesh7:
> Why is this needed? To clarify, I do prefer python script to a bash function. Trying to understand the motivation.

The way the `check_valgrind_log` bash function was splitting up the leak records required that they be of a minimum length. Sometimes this would result in missed leaks, or false positives. With the python script, you can cleanly separate the leak records. In the future, if we encounter leaks with hidden stack traces, we can increase the number of callers shown on the stack trace, and the script can already accomodate this. With the bash function you would need to go back and change the number of lines emitted by grep.

CoderHam:
@aramesh7 was the CI failure fixed for this?

aramesh7:
> @aramesh7 was the CI failure fixed for this?

Yes, all the related tests pass.

qa/common/check_valgrind_log.py
```diff
@@ -0,0 +1,115 @@
+# Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
```
CoderHam, line 1: Fix Copyright

CoderHam (changes requested):

CoderHam (approved):

deadeyegoodwin (approved):

PR #2647, opened 2021-03-19
Fix minor typos in docs/architecture.md

qpakzk:


tanmayv25 (approved):

docs/architecture.md
```diff
@@ -246,7 +246,7 @@ sequence_batching {
   the corresponding batch slot has an inference request ready for
   inference. In the example configuration, fp32_false_true indicates
   that a sequence ready is indicated by tensor element equal to 1, and
-  non-start is indicated by tensor element equal to 0.
+  non-ready is indicated by tensor element equal to 0.
```
CoderHam, line 249: I believe this should be `non-start` and not `non-ready`. Is there a reason for you to think otherwise?

CoderHam (changes requested):

docs/architecture.md
qpakzk: Hello. I thought it is a typo because `non-start` is for the start input tensor.

deadeyegoodwin (approved):

PR #2654, opened 2021-03-23
Update client_libraries.md

cathy-kim:
pip install tritonclient[all] or tritonclient[http] not working now


tanmayv25:
@cathy-kim 
`pip install tritonclient[http]`,  `pip install tritonclient[grpc]` and `pip install tritonclient[all]`  works just fine for me. Which operating system are you using?
If you find any of these not working then file an issue rather than this PR to remove documentation.

PR #2655, opened 2021-03-23
Include triton headers in container

deadeyegoodwin:


CoderHam:
@deadeyegoodwin why is this required?

deadeyegoodwin:
When using libtritonserver.so directly the header file is needed. The server container is the easiest way for users to get libtritonserver.so. So even though using Triton as a containerized server doesn't require the inlcludes, including them is a nice convenience (and they are very small).

CoderHam:
I understand. Thanks for explaining the same :)

CoderHam (approved):

PR #2664, opened 2021-03-25
Update README and versions for 21.03 release

deadeyegoodwin:


README.md
```diff
@@ -30,5 +30,296 @@
...
+  OpenVINO models on CPUs.
+  https://github.com/triton-inference-server/openvino_backend
+
+* The PyTorch backend is now maintained in its own repository:
+  https://github.com/triton-inference-server/pytorch_backend
+
+* The ONNX Runtime backend is now maintained in its own repository:
+  https://github.com/triton-inference-server/onnxruntime_backend
+
+* The Jetson release of Triton now supports the shared-memory protocol
```
CoderHam, line 60: Can we clarify that this is "system" shared memory? If the users try to use cuda shared memory, it will not work

CoderHam (changes requested):

README.md
deadeyegoodwin: Fixed

PR #2666, opened 2021-03-25
WIP: RateLimiter integration with Dynamic Batcher

tanmayv25:
Opening this PR to start receiving some preliminary review comments. 

Points to note:
1. Rate Limiter will be supported only by the backends that use Triton's Backend API.
2. Much of the weird artifacts related to lifetime management can now be removed.
3. There is a chance that TritonModelInsatnce_Execute is executed by different threads.. Whichever happen to invoke the callback by using RateLimiter::ModelInstance::Release function.
4. Most of the implementations are already laid out in the code. Most challenging part is to implement actual dynamic batching..
        a. The scheduler should be very careful while requesting the model instance from the RateLimiter so we don't waste resources which would have been better allocated to other models.
        b. In callback, we must search for any requests that can help us in building better batches. 
The additional staging queue helps in addressing the two but requires entire rewrite of the batching logic and might have an affect on the performance. The requests will go through another queue before being sent for execution.

- [x] Single Instance, no dynamic batcher
- [x] Mutiple instance, no dynamic batcher - Trivial
- [ ] Dynamic Batching - Requires care...
- [ ] Add support for WarmUp. - Should be Trivial
- [ ] Fix artifacts related to tests 
- [ ] Enable RateLimiter unit tests - Trivial

Diff between dynamic_batch_scheduler_v2.cc and dynamic_batch_scheduler.cc.. 
[diff.txt](https://github.com/triton-inference-server/server/files/6207463/diff.txt)

### DIFF
--- a/src/core/server.h
+++ b/src/core/server.h
@@ -47,7 +47,7 @@ class InferenceRequest;
 enum class ModelControlMode { MODE_NONE, MODE_POLL, MODE_EXPLICIT };
 
 // The configured mode for rate limiting
-enum class RateLimitMode { RL_OFF, RL_EXEC_COUNT };
+enum class RateLimitMode { RL_EXEC_COUNT, RL_OFF };
 
 // Readiness status for the inference server.
 enum class ServerReadyState {
--- a/src/core/server.h
+++ b/src/core/server.h
@@ -36,6 +36,7 @@
 #include "src/core/model_config.pb.h"
 #include "src/core/model_repository_manager.h"
 #include "src/core/persistent_backend_manager.h"
+#include "src/core/rate_limiter.h"
 #include "src/core/status.h"
 
 namespace nvidia { namespace inferenceserver {
@@ -44,6 +45,9 @@ class InferenceBackend;
 class InferenceRequest;
 
 enum class ModelControlMode { MODE_NONE, MODE_POLL, MODE_EXPLICIT };
+
+// The configured mode for rate limiting
+enum class RateLimitMode { RL_OFF, RL_EXEC_COUNT };
 
 // Readiness status for the inference server.
 enum class ServerReadyState {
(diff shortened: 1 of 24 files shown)


src/core/server.h
```diff
@@ -154,6 +158,20 @@ class InferenceServer {
   bool StrictModelConfigEnabled() const { return strict_model_config_; }
   void SetStrictModelConfigEnabled(bool e) { strict_model_config_ = e; }
 
+  // Get / set rate limit mode.
+  RateLimitMode GetRateLimitMode() const { return rate_limit_mode_; }
```
deadeyegoodwin, line 162: We don't typically use "Get". In this case ok since RateLimitMode is an enum... or change to RateLimiterMode()

src/core/server.h
```diff
@@ -154,6 +158,20 @@ class InferenceServer {
...
+  void SetRateLimitMode(RateLimitMode m) { rate_limit_mode_ = m; }
+
+  // Get / set rate limit resource counts
+  const RateLimiter::ResourceMap& GetRateLimitResources() const
```
deadeyegoodwin, line 166: RateLimiterResources ?

src/servers/main.cc
```diff
@@ -366,6 +368,27 @@ std::vector<Option> options_
        "Name of the model to be loaded on server startup. It may be specified "
        "multiple times to add multiple models. Note that this option will only "
        "take affect if --model-control-mode=explicit is true."},
+      {OPTION_RATE_LIMIT, "rate-limit", Option::ArgStr,
+       "Specify the mode for rate limiting. Options are \"execution_count\" "
```
deadeyegoodwin, line 372: I think the default should be "off"

deadeyegoodwin (changes requested):

src/servers/main.cc
tanmayv25: I might have confused something. 
Can you explain what did you meant here: 
https://github.com/triton-inference-server/core/pull/6#discussion_r515357510

deadeyegoodwin: You are right, I got confused. It should be on by default... so leave this like you have it.

PR #2669, opened 2021-03-26
Improve documentation of Python client string tensors

deadeyegoodwin:


docs/client_examples.md
```diff
@@ -83,13 +83,24 @@ Docker](client_libraries.md#build-using-docker) or
...
+Some frameworks support tensors where each element in the tensor is
+variable-length binary data. Each element can hold a string or an
+arbitrary sequence of bytes. On the client this datatype is BYTES (see
+[Datatypes](model_configuration.md#datatypes) for information on
+supported datatypes).
+
+The Python client library uses numpy to represent input and output
+tensors. For BYTES tensors the dtype of the numpy array should be
+'np.object_' as shown in the examples. For backwards compatibility
+with previous versions of the client library, np.bytes_ can also be
```
Tabrizian, line 97: np.bytes_ -> `np.bytes_`

src/clients/python/examples/simple_grpc_string_infer_client.py
```diff
@@ -108,4 +97,76 @@
...
+        sys.exit()
+
+    # Example using BYTES input tensors with utf-8 encoded string
+    # elements
+    simple_string_inference(triton_client)
+
+    # Example using BYTES input tensor with utf-8 encoded string that
+    # has an embedded null character.
+    null_chars_array = np.array(
+        ["he\x00llo".encode('utf-8') for i in range(16)], dtype=np.object_)
```
Tabrizian, line 160: Actually, this worked even before the new changes were introduced. It would be great if we can have two arrays here where one of them uses `np.object` and the other one uses `np.bytes_`. Then, we can show in the examples that trailing zeros are lost when using `np.bytes_`.

```python
null_chars_object = np.array(
        ["hello\x00".encode('utf-8') for i in range(16)], dtype=np.object_)
identity_inference(null_chars_object)
null_chars_bytes = np.array(
        ["hello\x00".encode('utf-8') for i in range(16)], dtype=np.bytes_)
incorrect_identity_inference(null_chars_object)
```

Where the `incorrect_identity_inference` passes when results don't match.

Tabrizian (changes requested):
Do we want to include the `serialized_byte_size` documentation changes in this PR or it will be addressed later?

src/clients/python/examples/simple_grpc_string_infer_client.py
deadeyegoodwin: Can you file a ticket for this addition.

docs/client_examples.md
deadeyegoodwin: Fixed

deadeyegoodwin:
> 
> 
> Do we want to include the `serialized_byte_size` documentation changes in this PR or it will be addressed later?

There is something in the release notes. You can check it when we do the github release and we can adjust if necessary.

src/clients/python/examples/simple_grpc_string_infer_client.py
Tabrizian: https://jirasw.nvidia.com/browse/DLIS-2213

PR #2679, opened 2021-03-30
Make byte tensor serialization faster for python client

VibhuJawa:
This pr improves byte tensor serialization by order of magnitude. (`6.5s` -> `10.4 ms`), thus making `set_data_from_numpy` for byte tensors much faster. 

### Todo:
- [x] Ensure tests and build pass

### Perf Comparision
```python
import numpy as np


log_ls = ['test log '*100]*10000
log_ls = [l.encode('utf-8') for l in log_ls]
log_ar = np.array(log_ls).reshape(1,len(log_ls))

input_grpc = grpcclient.InferInput("raw_log",log_ar.shape,"BYTES")
```

##### Without this PR:
```python
input_grpc.set_data_from_numpy(log_ar)
```
```
Wall time:  6.55 s
```

##### With this PR:
```python
input_grpc.set_data_from_numpy(log_ar)
```
```
CPU times: user 9.24 ms, sys: 1.26 ms, total: 10.5 ms
Wall time: 10.4 ms
```

CC: @BartleyR, @bsuryadevara

### DIFF
--- a/src/clients/python/library/tritonclient/utils/__init__.py
+++ b/src/clients/python/library/tritonclient/utils/__init__.py
@@ -216,7 +216,7 @@ def serialize_byte_tensor(input_tensor):
     # order.
     if (input_tensor.dtype == np.object_) or (input_tensor.dtype.type
                                               == np.bytes_):
-        flattened = bytes()
+        flattened_ls = []
         for obj in np.nditer(input_tensor, flags=["refs_ok"], order='C'):
             # If directly passing bytes to BYTES type,
             # don't convert it to str as Python will encode the
@@ -228,8 +228,9 @@ def serialize_byte_tensor(input_tensor):
                     s = str(obj.item()).encode('utf-8')
             else:
                 s = obj.item()
-            flattened += struct.pack("<I", len(s))
-            flattened += s
+            flattened.append(struct.pack("<I", len(s)))
+            flattened.append(s)
+        flattened = b''.join(flattened_ls)
         flattened_array = np.asarray(flattened, dtype=np.object_)
         if not flattened_array.flags['C_CONTIGUOUS']:
             flattened_array = np.ascontiguousarray(flattened_array,
--- a/src/clients/python/library/tritonclient/utils/__init__.py
+++ b/src/clients/python/library/tritonclient/utils/__init__.py
@@ -228,8 +228,8 @@ def serialize_byte_tensor(input_tensor):
                     s = str(obj.item()).encode('utf-8')
             else:
                 s = obj.item()
-            flattened.append(struct.pack("<I", len(s)))
-            flattened.append(s)
+            flattened_ls.append(struct.pack("<I", len(s)))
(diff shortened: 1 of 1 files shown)


VibhuJawa:
@tanmayv25 , 

If this change is acceptable we should also target it to the `python_backend`. 

https://github.com/triton-inference-server/python_backend/blob/15ca4cf1d3336844d88ab3f31b930639d47dfd4d/src/resources/startup.py#L54-L108

Tabrizian:
Good catch and thank you for the PR. What do you think is the underlying reason for this speedup? Is it because `bytes` is immutable and lists are mutable? Do you think using `bytearray` will lead to the same speedup?

Tabrizian:
Regarding the python_backend, we are in the process of redesigning the python backend and that file will be removed when those changes are complete. But thanks for mentioning this.

CoderHam:
@VibhuJawa can you remove the WIP from the PR title if it is no longer a WIP?

tanmayv25:
@VibhuJawa Thanks for the PR. I have launched a CI tests with your changes. I will let you know if there any issues pop-up. Otherwise, changes look good to me.

tanmayv25:
CI run was successful: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/2221297

tanmayv25 (approved):

VibhuJawa:
> Good catch and thank you for the PR. What do you think is the underlying reason for this speedup? Is it because `bytes` is immutable and lists are mutable? Do you think using `bytearray` will lead to the same speedup?

@Tabrizian,

Kind of, I think this is because of the number of memory allocations that are happening as well as the increase in size of allocations. 

If we have something like below , we have 2*n big allocation and 2*n small allocation and the cost of each subsequence allocation keeps on increasing. 

```python
flattened = bytes()
s = 'test string' * 1024
n = 1000

for _ in range(n):
   flattened =  flattened + struct.pack("<I", len(s)) ## 2 allocations , 1 big , 1 small 
   flattened = flattened + s ## 2 allocations , 1 big , 1 small 
``` 
```
1.19 s ± 26.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
``` 

### Graph of cost of  allocations (increases incrementally due to increase increase in size of allocation) 
```
flattened = flattened + s 
```

image


          
### Alternate Solution: 
This should just have just 2n small allocations and 1 big allocation  :
```python

s = 'test string' * 1024
n = 1000
flattened_ls = []

for _ in range(n):
  flattened_ls.append(struct.pack("<I", len(s))) # 1 small allocation 
  flattened_ls.append(s) ## 1 small allocation 
  
flattened = b''.join(flattened_ls) ## 1 big allocation 
```
```
1.27 ms ± 43.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Tabrizian:
@VibhuJawa Thanks for the detailed answer!

PR #2693, opened 2021-04-02
Remove legacy custom backend 

CoderHam:


qa/L0_dyna_sequence_batcher/dyna_sequence_batcher_test.py
```diff
@@ -46,14 +46,12 @@
...
 _no_batching = (int(os.environ.get('NO_BATCHING', 0)) == 1)
 
-_trials = ("custom", "savedmodel", "graphdef", "plan", "onnx", "libtorch")
+_trials = ("savedmodel", "graphdef", "plan", "onnx", "libtorch")
 if _no_batching:
     _trials += ("savedmodel_nobatch", "graphdef_nobatch", "plan_nobatch",
                 "onnx_nobatch", "libtorch_nobatch")
 
 _ragged_batch_supported_trials = list()
-if "custom" in _trials:
```
deadeyegoodwin, line 55: @GuanLuo  should we be testing TensorRT and TF (and soon ORT) here since they support ragged batching?

qa/L0_infer/infer_test.py
```diff
@@ -107,25 +107,25 @@ def _infer_exact_helper(tester,
...
+                    use_streaming=use_streaming,
+                    correlation_id=correlation_id,
                     use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
                     use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY)
 
         input_size = 16
 
         all_ensemble_prefix = ["simple_", "sequence_", "fan_"]
         ensemble_prefix = [""]
-        if ENSEMBLES and ("custom" in BACKENDS):
```
deadeyegoodwin, line 128: Interesting... looks like we were not testing ensembles for non-custom models?

qa/L0_large_payload/large_payload_test.py
```diff
@@ -157,6 +157,7 @@ def test_onnx(self):
         for client in self._clients:
             model_name = tu.get_zero_model_name("onnx_nobatch", 1,
                                                 self._data_type)
+            self._test_helper(client, model_name)
```
deadeyegoodwin, line 160: So onnx was not running the large payload test!?!

qa/L0_perf_nomodel/test.sh
```diff
@@ -139,7 +139,7 @@ TEST_CONCURRENCY+=(
     1
     16
     16)
-TEST_BACKENDS="plan custom graphdef savedmodel onnx libtorch python"
+TEST_BACKENDS="custom"
```
deadeyegoodwin, line 142: Don't you want to run nomodel test on all backends?

qa/L0_sequence_batcher/test.sh
```diff
@@ -152,19 +156,20 @@ function get_datatype () {
...
 MODELS=""
 for BACKEND in $BACKENDS; do
-  if [[ $BACKEND == "custom" ]]; then
+  DTYPE=$(get_datatype $BACKEND)
+  if [ "$BACKEND" == "custom" ]; then
     MODELS="$MODELS ../custom_models/custom_sequence_int32"
-  else
-    DTYPE=$(get_datatype $BACKEND)
-    MODELS="$MODELS $DATADIR/qa_sequence_model_repository/${BACKEND}_sequence_${DTYPE}"
+    continue
```
deadeyegoodwin, line 162: ? We don't want to skip this, do we? Similar for continues below.

qa/L0_sequence_batcher/test.sh
```diff
@@ -200,27 +205,36 @@ for MODEL in $MODELS; do
       (cd queue_delay_models/$(basename $MODEL)_full && \
         sed -i "s/$(basename $MODEL)/$(basename $MODEL)_full/" config.pbtxt && \
         sed -i "s/minimum_slot_utilization: 0/minimum_slot_utilization: 1/" config.pbtxt)
+  else
+    cp -r $MODEL queue_delay_models/$(basename $MODEL)_full && \
```
deadeyegoodwin, line 209: Why wasn't this needed before?

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -80,7 +80,7 @@
     res = []
```
deadeyegoodwin, line 80: Earlier in this file there is another case where custom is used as the only ragged batch testcase. @GuanLuo this seems like another case of missed ragged batching testing? We should file a ticket to track these and get them enabled.

qa/L0_simple_lib/test.sh
```diff
@@ -85,37 +85,13 @@ for SIMPLE_CLIENT in simple ; do
...
+    # Use savedmodel for addsub ensemble
     mkdir -p $MODELSDIR/simple/1
-    cp -r ../custom_models/custom_float32_float32_float32/1/* $MODELSDIR/simple/1/.
-    cp ../custom_models/custom_float32_float32_float32/config.pbtxt $MODELSDIR/simple/.
+    cp -r $DATADIR/savedmodel_float32_float32_float32/1/* $MODELSDIR/simple/1/.
+    cp $DATADIR/savedmodel_float32_float32_float32/config.pbtxt $MODELSDIR/simple/.
     (cd $MODELSDIR/simple && \
             sed -i "s/^name:.*/name: \"simple\"/" config.pbtxt && \
             sed -i "s/label_filename:.*//" config.pbtxt)
-    full=custom_float32_float32_float32
```
deadeyegoodwin, line 96: Why are these removed?

src/core/model_config_utils.cc
```diff
@@ -873,6 +868,17 @@ ValidateModelIOConfig(const inference::ModelConfig& config)
...
 ValidateBatchIO(const inference::ModelConfig& config)
 {
+  if (
+#ifdef TRITON_ENABLE_TENSORRT
+      (config.platform() != kTensorRTPlanPlatform) &&
+#endif  // TRITON_ENABLE_TENSORRT
+      ((config.batch_input_size() != 0) || (config.batch_output_size() != 0))) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "batch inputs and batch outputs are only supported for custom "
```
deadeyegoodwin, line 878: Now supported for Tensorflow as well, and soon ORT.

"batch inputs and batch outputs are not support for this platform/backend"

src/core/model_config_utils.cc
```diff
@@ -1364,6 +1364,17 @@ ValidateModelInput(
...
   }
 
+  if (
+#ifdef TRITON_ENABLE_TENSORRT
+      (platform != kTensorRTPlanPlatform) &&
+#endif  // TRITON_ENABLE_TENSORRT
+      io.allow_ragged_batch()) {
+    return Status(
+        Status::Code::INVALID_ARG,
+        "ragged-batch input tensors are only supported for custom platform"
```
deadeyegoodwin, line 1374: Now supported for Tensorflow as well, and soon ORT.

"ragged-batch input tensors are not support for this platform/backend"

deadeyegoodwin (changes requested):

qa/L0_infer/infer_test.py
GuanLuo: For ensemble, you must name the models in the model config, so a lot of ensemble models will need to be generated to cover the backend combinations. And we don't really need to test all combinations as ensemble is backend insensitive.

qa/L0_infer/test.sh
```diff
@@ -173,26 +169,23 @@ for TARGET in cpu gpu; do
...
+
+      create_nop_version_dir `pwd`/models
 
       if [[ $BACKENDS == *"graphdef"* ]]; then
         ENSEMBLE_MODELS="wrong_label_int32_float32_float32 label_override_int32_float32_float32 mix_type_int32_float32_float32"
 
-        if [[ $BACKENDS == *"custom"* ]]; then
-          ENSEMBLE_MODELS="${ENSEMBLE_MODELS} batch_to_nobatch_float32_float32_float32 batch_to_nobatch_nobatch_float32_float32_float32 nobatch_to_batch_float32_float32_float32 nobatch_to_batch_nobatch_float32_float32_float32 mix_nobatch_batch_float32_float32_float32"
-        fi
+        ENSEMBLE_MODELS="${ENSEMBLE_MODELS} batch_to_nobatch_float32_float32_float32 batch_to_nobatch_nobatch_float32_float32_float32 nobatch_to_batch_float32_float32_float32 nobatch_to_batch_nobatch_float32_float32_float32 mix_nobatch_batch_float32_float32_float32"
```
GuanLuo, line 193: @CoderHam I think now you use savedmodel for the ensemble? Just change the conditional to check if "savedmodel" is in BACKENDS here and elsewhere. We don't want to run "savedmodel" if the script is invoked without specifying it.

qa/L0_sequence_batcher/sequence_batcher_test.py
GuanLuo: Agree. We have L0_batch_input which tests ragged batching as well. We may want to either extract the related tests to one test folder so we can manage it in one place or place the names of backend with ragged support in a common area (test util?) and the tests will read the backends to be tested from that common area.

qa/L0_batcher/test.sh
```diff
@@ -403,10 +399,12 @@ done
 # by comparing the "response send" timestamps.
 TEST_CASE=test_multi_batch_preserve_ordering
 
-if [[ $BACKENDS == *"custom"* ]]; then
+# FIXME sleep_for does not delay for expected time when passing 400ms
```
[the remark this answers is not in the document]
CoderHam: Here is the ticket https://jirasw.nvidia.com/browse/DLIS-2315
The execution delay works fine for system and no-shared memory but for cuda shared memory case I see an execution delay of 100 and 280ms instead of 100 and 400ms.

src/core/model_config_utils.cc
GuanLuo: Why adding these checks? This type of check ought to be done by the backend really, as now Triton doesn't aware whether the backend of the model config support these features or not.

qa/L0_infer/infer_test.py
CoderHam: We do have some ensembles that are purely graphdef and savedmodel. We test these under test_ensemble_mix_platform and test_ensemble_mix_type

qa/L0_large_payload/large_payload_test.py
CoderHam: Yes it was extremely odd. I did not see a comment as to why it was not there so I added it. 
I was able to trace back the removal to https://github.com/triton-inference-server/server/pull/2207

qa/L0_perf_nomodel/test.sh
CoderHam: Fixing, missed this in the cleanup

qa/L0_perf_analyzer/test.sh
```diff
@@ -94,9 +93,6 @@ cp -r /data/inferenceserver/${REPO_VERSION}/tf_model_store/inception_v1_graphdef
 # Copy resnet50v1.5_fp16
 cp -r /data/inferenceserver/${REPO_VERSION}/perf_model_store/resnet50v1.5_fp16_savedmodel $DATADIR
 
-# Set up the ensemble model repository
-cp -r ../ensemble_models/image_preprocess_ensemble_example ensemble_model_repository
```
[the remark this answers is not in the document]
CoderHam: The image process example is to be removed. I had discussed this with @deadeyegoodwin. It is necessary since we are deprecating the custom backend.

qa/L0_simple_lib/test.sh
CoderHam: The custom_float32_float32_float32 is an addsub model that used the legacy addsub.so. Since we do not have a replacement for it yet we have to remove it. As we discussed we do not want to add a new addsub backend right?

qa/L0_batcher/batcher_test.py
```diff
@@ -56,14 +56,10 @@
...
-BACKENDS = os.environ.get('BACKENDS',
-                          "graphdef savedmodel onnx libtorch plan custom")
+BACKENDS = os.environ.get('BACKENDS', "graphdef savedmodel onnx libtorch plan")
 
 _trials = BACKENDS.split(" ")
-if "custom" in BACKENDS:
-    _ragged_batch_supported_trials = ("custom",)
-else:
-    _ragged_batch_supported_trials = ()
+_ragged_batch_supported_trials = ()
```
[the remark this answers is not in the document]
CoderHam: @GuanLuo I think we should enable ragged batch testing for the other backends in a different ticket since there is already way to much going on here

qa/L0_grpc/test.sh
```diff
@@ -142,14 +138,6 @@ for i in \
             cat $CLIENT_LOG.${SUFFIX}
             RET=1
         fi
-    elif [ $SUFFIX == "ensemble_image_client" ]; then
```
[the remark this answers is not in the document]
CoderHam: Filed a follow up ticket to check this out [DLIS-2316](https://jirasw.nvidia.com/browse/DLIS-2316)

src/core/model_config_utils.cc
CoderHam: Removing. These changes are outdated and unnecessary

src/core/model_config_utils.cc
CoderHam: Removing. These changes are outdated and unnecessary

qa/L0_decoupled/decoupled_test.py
```diff
@@ -59,7 +59,6 @@ class DecoupledTest(tu.TestResultCollector):
     def setUp(self):
         self.trials_ = [("repeat_int32", None), ("simple_repeat", None),
                         ("sequence_repeat", None),
-                        ("fan_repeat", self._fan_validate),
```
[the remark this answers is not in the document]
CoderHam: The fan_repeat ensemble uses custom_nobatch_int32_int32_int32 which is a legacy custom addsub model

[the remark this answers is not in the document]
GuanLuo: can't you replace it with other addsub like what you have done for other places?

qa/L0_dyna_sequence_batcher/dyna_sequence_batcher_test.py
CoderHam: [DLIS-2318](https://jirasw.nvidia.com/browse/DLIS-2318)

qa/L0_infer/test.sh
CoderHam: These ensembles now use the identity backend model instead of the legacy custom backend addsub

qa/L0_grpc/test.sh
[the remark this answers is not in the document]
GuanLuo: Since we plan to keep an example of using ensemble for image classification, I am inclined to keep the test for that as well. So just comment this part out so it is easier to re-enable the test after the ensemble example is updated.

qa/L0_perf_analyzer/test.sh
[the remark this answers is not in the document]
GuanLuo: Make sure it is also mentioned in DLIS-2316

qa/L0_batcher/test.sh
```diff
@@ -403,10 +401,12 @@ done
 # by comparing the "response send" timestamps.
 TEST_CASE=test_multi_batch_preserve_ordering
 
-if [[ $BACKENDS == *"custom"* ]]; then
+# FIXME sleep_for does not delay for expected time when passing 400ms 
+# when using I/O is in GPU memory. Skip for Windows test.
+if [ "$OS_WINDOWS" -eq "0" ] && [ "$TEST_CUDA_SHARED_MEMORY" -eq 0 ]; then
```
GuanLuo, line 406: Why Windows is skipped? Is that because we don't build identity for Windows?

And I noticed that a lot of ensemble related tests are now disabled for Windows, but we should have been able to test that on Windows, right? Perhaps instead of using TF models, you should use ONNX addsub as ORT backend is supported on Windows.

CoderHam: Yes, we do not build identity backend for windows. We earlier used the custom backend to signify if we were to test for models. The custom backend was not listed at the backend to test for on windows and hence these tests were earlier disabled as well. In short, no tests are being additionally disabled that were not already being run on windows.

qa/L0_decoupled/decoupled_test.py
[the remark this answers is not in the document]
CoderHam: Follow up: DLIS-2320

qa/L0_grpc/test.sh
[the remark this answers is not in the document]
CoderHam: Done

qa/L0_sequence_batcher/test.sh
CoderHam: When testing for valgrind, the QUEUE_DELAY_TESTS is set to test_queue_delay_full_min_util which does indeed need the _full models.. I am not sure why this test was passing before but it should not have. 
@aramesh7 you added the valgrind test coverage that required the _full models [here](https://github.com/triton-inference-server/server/commit/5feabbde052e61cab7c822a53fbbb58effa7a622#diff-13d0a7efc37da7e6d411c65de0b4b4d91ed150a97ef5e885d92395f6506177fb).
Can you confirm why the _full models were not being created for the valgrind variant even though they appear to be needed by the test_queue_delay_full_min_util test?

qa/L0_batcher/test.sh
GuanLuo: Can you file a ticket to address this and remove the Windows conditionals here and in other tests? I don't think we intended to skip Windows for ensemble tests and there is really no reason to do skip that (i.e. all parts needed for ensemble should be able to build for Windows)

CoderHam: DLIS-2321 - Windows ensemble testing
DLIS-2315 - Cuda shared memory delay time failure

qa/L0_model_queue/model_queue_test.py
```diff
@@ -115,12 +114,12 @@ def test_max_queue_size(self):
         # model execution. Then sends 10 requests to overload the model queue,
         # expecting 2 of the requests are returned with error code immediately.
         dtype = np.float32
-        shapes = ([16],)
+        shapes = ([16], )
 
         for trial in self.trials_:
             preceding_thread = threading.Thread(
                 target=self.check_response,
-                args=(8, dtype, shapes, 0, 0, (1999, 1000)),
```
GuanLuo, line 123: Why do you change the threshold values?

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -2267,7 +2266,7 @@ def test_queue_delay_full_min_util(self):
                             model_name,
                             dtype,
                             1001,
-                            (4000, 3000),
+                            (6000, 3000),
```
GuanLuo, line 2269: Why do you change the threshold?

GuanLuo (changes requested):

deadeyegoodwin (approved):

qa/L0_sequence_batcher/sequence_batcher_test.py
CoderHam: There is a performance drop from the custom legacy sequence model to the new sequence backend is the need to increase this threshold.

GuanLuo: That is a concerning drop, from 4000ms to 6000ms (4 s to 6 s). I don't think the performance drop in sequence model change can be this dramatic, but in either way, we will likely want to fix this. Can you file a ticket for that and leave the original threshold values as inline comments?

qa/L0_model_queue/model_queue_test.py
GuanLuo: This test isn't failing on master (without the change in threshold)

qa/L0_sequence_batcher/sequence_batcher_test.py
CoderHam: Done

qa/L0_model_queue/model_queue_test.py
CoderHam: Reverted these changes. Yes they were not needed.

GuanLuo (approved):

PR #2694, opened 2021-04-02
Fix ONNX version in gen script

CoderHam:
- Fix typo in comment


qa/common/gen_qa_model_repository
```diff
@@ -51,7 +51,7 @@
 TRITON_VERSION=21.03
 
 # ONNX. Use ONNX_OPSET 0 to use the default for ONNX version
-ONNX_VERSION=1.6.0
+ONNX_VERSION=1.7.0
```
deadeyegoodwin, line 54: Shouldn't it be 1.7.1

CoderHam: Yes but there is no [pip candidate](https://github.com/triton-inference-server/server/blob/master/qa/common/gen_qa_model_repository#L156) for 1.7.1 so I am using 1.7.0 instead

PR #2704, opened 2021-04-06
delete meaningless if True condition

qpakzk:


CoderHam:
We plan to move the core server out to a separate repository. As mentioned in the comments above this is a temporary placeholder for the same.

PR #2705, opened 2021-04-06
Minor test fixes for Python backend

Tabrizian:


src/core/shared_library.cc
```diff
@@ -85,7 +85,14 @@ OpenLibraryHandle(const std::string& path, void** handle)
         Status::Code::NOT_FOUND, "unable to load backend library: " + errstr);
   }
 #else
-  *handle = dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL);
+  // Need to open the Python backend shared library using RTLD_GLOBAL flag so
+  // that subsequent dlopens in the user's Python model work properly
+  if (path.find("libtriton_python") != std::string::npos) {
```
deadeyegoodwin, line 90: Since the stub is going to be a completely separate process now, I think this is not needed?

Tabrizian: Yes, this is no longer required when the stub becomes completely separate. I'll mark this PR as WIP for now.

CoderHam (approved):

PR #2712, opened 2021-04-08
Fix dependencies in QA container

tanmayv25:


tanmayv25:
Passes the tests like `L0_grpc` and `L0_http` which uses `image_client.py`. 
https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/2247564

deadeyegoodwin (approved):

PR #2718, opened 2021-04-10
Support A100 MIG Partition in GKE application 

mengdong:
@deadeyegoodwin Hi David, could you please review this pull request? It adds support for A100 MIG in GKE. We offer a bert large TRT model for 5gb partition with Triton App.


mengdong:
Regarding 21.04 update, could we update the deployer code when we push the 21.04 NGC image? I need to refresh the deployer at that time as well.

deadeyegoodwin:
If you have changes that should be included in the 21.04 release then you can put them on the r21.04 branch so that they are released with the 21.04 release.

deadeyegoodwin (approved):

PR #2734, opened 2021-04-15
Moving AsyncWorkQueue to common repo

tanmayv25:


src/core/status.h
```diff
@@ -27,6 +27,7 @@
 
 #include <string>
 #include "src/core/tritonserver_apis.h"
+#include "triton/common/status.h"
```
deadeyegoodwin, line 30: It is confusing to have this Status and also common::Status. How about name the common class "Error". so it is triton::common::Error.

src/core/status.h
```diff
@@ -51,6 +52,9 @@ class Status {
   // Construct a status from a code and message.
```
deadeyegoodwin, line 52: Do we need to replicate the Code class above. Can't you just inherit that from common::Error? Can't everything be inherited except the TRITONSERVER_* conversions?

deadeyegoodwin (changes requested):

GuanLuo (approved):

src/core/status.h
GuanLuo: Do we want to move `TRITONSERVER_*` conversions to `common::Error` as well? For users of `common::Error`, i.e. implementing a backend, it is likely that they will need to convert it to `TRITONSERVER_Error` at the end

tanmayv25: @GuanLuo I was thinking in that direction. Like moving status.cc/.h to common-repo from core. But it will expose the entire TRTITONSERVER_*  api to all the places where status.h is being used. I decided against this for better encapsulation. It would be good to have self-sufficient utilities in common-repo. 

However, I do see benefit in breaking the encapsulation and including TRITONSERVER_* conversions in `triton::common::`. If you find it a better solution, I am in for migrating entire `status.cc/status.h` to common::repo.

deadeyegoodwin (approved):

PR #2736, opened 2021-04-16
Support data compression in HTTP

GuanLuo:


src/servers/http_server.cc
```diff
@@ -2738,6 +2770,53 @@ HTTPAPIServer::InferRequestClass::FinalizeResponse(
                                std::to_string(buffer.Size()).c_str(), 1, 1));
   }
 
+  evbuffer* response_body = response_placeholder;
+  // Find Accept-Encoding in header. Try to compress if found
+  const char* accept_encoding_c_str =
+      evhtp_kv_find(req_->headers_in, kAcceptEncodingHTTPHeader);
+  if (accept_encoding_c_str != NULL) {
+    // FIXME compress only work for the simplest case for now
```
deadeyegoodwin, line 2778: Should fix this

src/servers/http_server.cc
```diff
@@ -2471,6 +2497,11 @@ HTTPAPIServer::InferRequestClass::InferResponseComplete(
   HTTPAPIServer::InferRequestClass* infer_request =
       reinterpret_cast<HTTPAPIServer::InferRequestClass*>(userp);
 
+  // Always specify supported compression as Accept-Encoding
+  evhtp_headers_add_header(
```
deadeyegoodwin, line 2501: The response doesn't need the Accept-Encoding header

deadeyegoodwin (changes requested):

src/servers/http_server.cc
GuanLuo: How would the client knows what compression type is supported by the server?

deadeyegoodwin: https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation#server-driven_content_negotiation

It looks like we should return 415 if client has Content-Encoding that Triton can support. That will tell the client that they need to try the request again with a different Content-Encoding. We don't need to automatically do the retry in our client libraries.

GuanLuo: So only on client sending content with unsupported type, the server returns 415 and attach the supported type as `Accept-Encoding` in the response header?

deadeyegoodwin: Client-side compression negitiation doesn't seem to be very well standardized. It should be ok to include Accept-Encoding in that error response.

deadeyegoodwin (approved):

PR #2749, opened 2021-04-21
Improve python backend testing

Tabrizian:
- [x] Add tests for tritonserver flags
- [x] Add test for shared memory growth
- [x] Add test for error in the batch.

### DIFF
--- a/qa/L0_backend_python/python_test.py
+++ b/qa/L0_backend_python/python_test.py
@@ -42,6 +42,45 @@ import tritonclient.http as httpclient
 
 class PythonTest(tu.TestResultCollector):
 
+    def _infer_help(self, model_name, shape, data_type):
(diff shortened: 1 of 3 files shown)


deadeyegoodwin (approved):

CoderHam (changes requested):
```
| python_nobatch_int8_int8_int8                       | 1       | UNAVAILABLE: Internal: Failed to initialize the model instance python_nobatch_int8_int8_int8_0 because it run out of shared memory. Increase the shared memory size to fix this error.          |
```
L0_infer failures. You might need to increase the shared memory size just for that test. Additionally this might be true for other tests so just have a deeper look.

deadeyegoodwin (approved):

Tabrizian:
I fixed the error on the gitlab side by increasing the shared memory size.

CoderHam (approved):

PR #2753, opened 2021-04-21
Set ownership correctly for /opt/tritonserver/include

deadeyegoodwin:


build.py
```diff
@@ -648,6 +648,10 @@ def create_dockerfile(ddir, dockerfile_name, argmap, backends, repoagents):
 COPY --chown=1000:1000 --from=tritonserver_build /tmp/tritonbuild/install/bin/tritonserver bin/
 COPY --chown=1000:1000 --from=tritonserver_build /tmp/tritonbuild/install/lib/libtritonserver.so lib/
 COPY --chown=1000:1000 --from=tritonserver_build /tmp/tritonbuild/install/include/triton/core include/triton/core
+
+# Top-level include/core not copied so --chown does not set it correctly,
+# so explicit set on all of include
+RUN chown -R triton-server:triton-server include
```
CoderHam, line 654: Why use triton-server here instead of 1000?

CoderHam (changes requested):

build.py
deadeyegoodwin: We have to use 1000 for the docker --chown because it doesn't know the actual user names as those are internal to the contents of the container. But here we are running a command inside the container so we can use the actual name. There are some comments earlier in the dockerfile where the triton-server user is created.

PR #2762, opened 2021-04-23
Use min container as base for the SDK container

deadeyegoodwin:


CoderHam:
Why do we want to use the min container (it likely contains other deps that aren't needed.. like TRT) as the base for the sdk container?

Tabrizian:
Not sure why this change was initially required but it is beneficial to model analyzer too. The quick start guide will become easier as MA can use triton which is inside the container instead of launching another separate container which is a bit difficult and requires some tricks.

Dockerfile.sdk
```diff
@@ -25,7 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 # Default setting is building on nvidia/cuda:11.1-devel-ubuntu20.04
```
tanmayv25, line 27: Would need to update the comment.

deadeyegoodwin:
The specific reason for this change is because the min container (which is same as base container) is managed to ensure it has all the latest security fixes, etc. The container we were using does not have this guarantee.

tanmayv25 (approved):

PR #2764, opened 2021-04-23
Add Python backend to L0_io

Tabrizian:
~~Needs to be rebased after https://github.com/triton-inference-server/server/pull/2749 is merged.~~ Done


qa/L0_backend_python/python_test.py
```diff
@@ -184,24 +223,41 @@ def test_infer_pytorch(self):
...
-            inputs[0].set_data_from_numpy(input_data)
-            try:
-                client.infer(model_name, inputs)
-            except InferenceServerException as e:
-                print(e)
-                self.assertTrue(
-                    e.message().startswith("An error occured during execution"),
-                    "Exception message is not correct")
-            else:
+        request_parallelism = 4
```
tanmayv25, line 226: Why are we using `request_parallelism = 4`? Why not `2`?
I think it is testing something beyond just infer_output_error.. Generating error response for the first requests and then successful response for the rest. This looks little confusing. May be add the justification as a comment here.

tanmayv25 (changes requested):

qa/L0_backend_python/python_test.py
Tabrizian: `2` does work too. This tests for the case when there is an error in one of the requests in the request batch. The backend should not fail the whole batch and should only fail the request that has an error.

tanmayv25 (dismissed):

deadeyegoodwin (approved):

PR #2772, opened 2021-04-26
Expand ONNXRuntime backend for Jetson/zero-container build

CoderHam:


build.py
```diff
@@ -367,28 +367,37 @@ def pytorch_cmake_args(images):
     ]
 
 
-def onnxruntime_cmake_args(images):
+def onnxruntime_cmake_args(images, library_paths):
+    # If platform is jetpack do not use docker based build
```
tanmayv25, line 371: Move the comment right above L377.

tanmayv25 (changes requested):

PR #2782, opened 2021-04-28
Fix sequence backend cuda shared memory support

CoderHam:


Tabrizian (commented):
Why don't you use `CopyBuffer`?

https://github.com/triton-inference-server/backend/blob/e29286a15560a1bb08cdff1fc35e30e1678bc968/include/triton/backend/backend_common.h#L423

CoderHam:
CopyBuffer manages copies for inputs and outputs but for that the inputs and outputs need to be in contiguous memory. A copy is only needed for the input and output since:
1. Inputs are conditionally set and cannot directly be modified in GPU memory
2. Outputs are copied one element at a time from accumulator

Tabrizian (approved):

Tabrizian (approved):

PR #2789, opened 2021-04-29
Fix priority and timeout passing in ensemble model. Add ensemble in L0_model_queue test

GuanLuo:


deadeyegoodwin (approved):

src/core/ensemble_scheduler.cc
```diff
@@ -818,11 +820,6 @@ EnsembleContext::InitStep(
   auto irequest = std::unique_ptr<InferenceRequest>(
       new InferenceRequest(backend, istep.model_version_));
 
-  // Request for ensemble model cannot override the timeout values for the
-  // composing models. Thus currently the timeout field in request has no
-  // effect until we support an overall ensemble timeout.
-  irequest->SetTimeoutMicroseconds(0);
```
GuanLuo, line 824: @deadeyegoodwin I think we had a discussion about what will ensemble do if a request to it has timeout set and came up with that. I updated the policy queue test with ensemble model and I need to change the behavior to passing along the composing models like priority as some of the tests rely on the timeout being overwritten.

Do you think it is okay to make this behavior change or I should exclude ensemble from the tests related to timeout override?

deadeyegoodwin:
I think having each model in the ensemble using the timeout is better than just ignoring the timeout completely. So I think that part of the change is ok.

Issue #2794, opened 2021-04-30
memory leak in s3 filesystem


yushcs:
**Description**
There is a memory leak in s3 filesystem.
Here is the information given by valgrind
```
==627== 344 bytes in 1 blocks are definitely lost in loss record 1,637 of 2,513
==627==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==627==    by 0x535D205: Aws::InitializeEnumOverflowContainer() (in /opt/tritonserver/lib/libtritonserver.so)
==627==    by 0x53880B3: Aws::InitAPI(Aws::SDKOptions const&) (in /opt/tritonserver/lib/libtritonserver.so)
==627==    by 0x4BFB5E9: nvidia::inferenceserver::(anonymous namespace)::GetFileSystem(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, nvidia::inferenceserver::(anonymous namespace)::FileSystem**) (filesystem.cc:1589)
==627==    by 0x4BFBFB3: nvidia::inferenceserver::IsDirectory(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool*) (filesystem.cc:1715)
[...]

PR #2796, opened 2021-04-30
Fix memory leak caused by repeated calling Aws::InitAPI

CoderHam:
Fixes issues outlined in #2794

### DIFF
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -1753,7 +1753,8 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
         "-DTRITON_ENABLE_S3=ON.");
 #else
     Aws::SDKOptions options;
-    Aws::InitAPI(options);
+    std::once_flag onceFlag;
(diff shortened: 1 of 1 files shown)


src/core/filesystem.cc
```diff
@@ -1753,7 +1759,10 @@ GetFileSystem(const std::string& path, FileSystem** file_system)
         "-DTRITON_ENABLE_S3=ON.");
 #else
     Aws::SDKOptions options;
-    Aws::InitAPI(options);
```
deadeyegoodwin, line 1756: c++11 call_once is probably a better, thread-safe way to make sure InitAPI is only called once.

deadeyegoodwin (changes requested):

src/core/filesystem.cc
CoderHam: Good point. Fixed 👍

deadeyegoodwin (approved):

PR #2797, opened 2021-04-30
Remove TF check for KIND_MODEL from Triton

GuanLuo:
As mentioned in other PR, I think we should remove backend specific check from Triton core. This particular check prevents other backends from reflecting the model placement in the appropriate way, i.e. the model config of the backend will specify `KIND_CPU` as WAR for now while that is not true.


CoderHam (approved):
These changes look good to me.
@tanmayv25 Since you are moving out the TensorRT backend to a separate repo could you also look into removing similar TRT checks from ValidateModelConfig()?

PR #2799, opened 2021-05-01
Fix nvtx annotations

AntoineFroger:
For performance reasons, some tools will return a fixed value when
implementing the callback for `nvtxRangePushA`. Keeping track of the
nesting level for each thread requires accessing thread local data which
does not come for free.

For instance, Nsight Systems always returns `-1`. Following the previous
code logic, all ranges would be pushed but never popped.

Applications should not to detect failures in the tool via return codes
of NVTX API functions. It is intentionally something NVTX was designed to
prevent. We want the application to always take the same branches,
whether a tool is profiling the application or not.

This PR:
* Modifies the `NvtxRange` class to ignore the return value of
  `nvtxRangePushA`.
* Overload the constructor of `NvtxRange` to prevent potential
  unnecessary string copies
* Format the `nvtx.h` file by running `clang-format`.

### DIFF
--- a/src/core/nvtx.h
+++ b/src/core/nvtx.h
@@ -28,7 +28,6 @@
 #ifdef TRITON_ENABLE_NVTX
 
 #include <nvtx3/nvToolsExt.h>
-#include "src/core/logging.h"
 
 namespace nvidia { namespace inferenceserver {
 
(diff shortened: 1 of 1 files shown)


AntoineFroger:
I tested my change by running Nsight Systems and checking that the NVTX ranges are now aligned correctly on the timeline.

GuanLuo (approved):
Looks good, will merge once we verify on our side that the nsys profile is generated properly.

PR #2806, opened 2021-05-04
Enable ragged test in L0_batcher, L0_dyna_sequence_batcher

GuanLuo:
L0_sequence_batcher already has ragged test enabled, so no change is needed.

Only the ragged test in L0_batcher added additional TRT and ONNX because we can use existing models to construct the "ragged models" expected by the test, but extra amount of effort will be needed for the test in (dyna) sequence. However, this should be fine since the batcher tests are mainly focusing on testing whether the requests will be batched appropriately, testing with one backend (sequence / dyna_sequence test model) will be sufficient.


deadeyegoodwin (approved):
ONNX ragged batch support was just merged so make sure you are picking up that change

CoderHam (approved):

PR #2814, opened 2021-05-05
Fix L0_backend_identity failure

CoderHam:
~~@tanmayv25 I fixed a smaller perf issue like earlier i.e. .find() in set is much faster than vector using the stl std::find~~
I also ran through most of the code base but no other places existed where indices were being compared as string instead of integers


GuanLuo (dismissed):

CoderHam:
Needed to re-base, please re-approve

GuanLuo (dismissed):

tanmayv25 (approved):

CoderHam:
@removed the change to dynamic sequence batcher since we cannot use set. Position for the corrid is required, hence reverted to old vector form.

deadeyegoodwin (approved):

Issue #2803, opened 2021-05-03
Windows support for C API sample


gvashishtha:
**Is your feature request related to a problem? Please describe.**
Today [simple.cc](https://github.com/triton-inference-server/server/blob/master/src/servers/simple.cc) is not included in Windows build because it uses some libraries and functions not supported on Windows, including unistd.h, getopt,  optarg, usleep.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

CoderHam:
Can you help us understand the motivation to have a simple.cc for Windows? It was meant to serve as an example.
This would help us prioritize expanding the simple.cc to be supported on Windows.

gvashishtha:
Our devs want to see if they can successfully use Triton on Windows. Having a known-good Windows example would help them move faster (rather than having to come up with something themselves)

Issue #2800, opened 2021-05-03
Exclude disabled cloud dependencies from CMAKE


gvashishtha:
**Is your feature request related to a problem? Please describe.**
Exclude cloud dependencies like Azure, AWS, Google Cloud from build target in CMAKE if they’re disabled. They will still be compiled even they’re disabled in CMAKE options and they failed to build on Windows today.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Specific to Windows build

CoderHam:
Can you elaborate how you are building on windows?

Yes, I do agree that the ExternalProject_Add and other flags corresponding to the Cloud deps should be disabled when the TRITON_ENABLE_GCS, TRITON_ENABLE_S3 and TRITON_ENABLE_AZURE_STORAGE flags are OFF.
Thanks for bringing this to our attention.

gvashishtha:
Working on getting the exact build script we are using, will follow up soon

deadeyegoodwin:
You should build using build.py as that is the supported way to build Triton, and then you won't see non-needed components being built.

YongHuangSJTU:
Hi @deadeyegoodwin, I did use build.py. I got this [error](https://github.com/triton-inference-server/server/issues/2834) from grpc build even I turned off grpc and http. These cloud dependencies will be built but not used in Triton even it's turned off since they're included by ExternalProject_Add.

deadeyegoodwin:
Are you following the build documentation for windows? https://github.com/triton-inference-server/server/blob/master/docs/build.md#building-for-windows-10
Following those docs I don't see errors and I also don't see grpc being build (assuming I don't add --endpoint:grpc flag to build.py) nor do I see cloud sdk builds because I don't add those flags to build.py (as they are not supported on windows at this point).

You may see the grpc repo being pulled as part of the build... that is because we build protobuf from that repo and protobuf is required even if grpc is not enabled. But we do not build anything related to grpc itself.

Can you share your build.py command.

YongHuangSJTU:
I set TRITON_ENABLE_HTTP and TRITON_ENABLE_GRPC to OFF, and run build.py --no-container-build --build-dir=./build_1 --cmake-dir=D:/github/triton2/server/build

PR #2816, opened 2021-05-05
WIP: Simplify and document windows server build

deadeyegoodwin:


YongHuangSJTU:
Thank you for working on windows build, this is very valuable for our project!
We want to integrate Triton using C API with our inference service, so I have a few suggestions based on our needs:

1.  Can we move the steps in Dockerfile to build.py or any executable script? I need to build Triton automatically, I know I can write a simple script to do it myself, but it may break when the code changes, so I’d like to have a single script in Triton repo to build it. 

2. Can we build this demo [simple.cc](https://github.com/triton-inference-server/server/blob/master/src/servers/simple.cc) (it’s excluded [here](https://github.com/triton-inference-server/server/blob/master/src/servers/CMakeLists.txt#L258)) on Windows? I managed to build triton on my Windows once, but I can’t build with this simple.cc, thus I can’t run Triton with the demo code. It uses some libraries and functions not supported on Windows, like unistd.h, getopt,  optarg, usleep. I want to play with it to see how Triton works, other Triton customers may have same need.

3. Can we exclude cloud (AWS, Azure, Google Cloud), http, grpc related dependencies in build if they are disabled? My project won't use these dependencies, it’s not necessary to build them if they're disabled. Can we add some conditions in CMakeLists.txt to exclude them?

deadeyegoodwin:
1. Why don't you just build with docker as is described in the build documentation? That is the most dependable way to build Triton. We don't want the build.py script to modify the users system by installing new packages, etc. That would be unexpected and not a good thing to do in general. That is why we capture all these dependencies in the Dockerfile.win10.min and use that for building.
2. simple.cc needs to be ported to windows so we need to schedule that work, we see you issue #2803 
3. You should build with build.py as that is the supported way of building. The cloud sdks will not be built if you do not enable them when using build.py.

Dockerfile.win10.min
```diff
@@ -0,0 +1,107 @@
...
+                   cusolver_${CUDA_MAJOR}.${CUDA_MINOR} cusolver_dev_${CUDA_MAJOR}.${CUDA_MINOR} \
+                   cusparse_${CUDA_MAJOR}.${CUDA_MINOR} cusparse_dev_${CUDA_MAJOR}.${CUDA_MINOR} \
+                   visual_studio_integration_${CUDA_MAJOR}.${CUDA_MINOR}"
+ARG CUDA_INSTALL_ROOT_WP="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${CUDA_MAJOR}.${CUDA_MINOR}"
+
+ARG TENSORRT_VERSION=7.2.2.3
+ARG TENSORRT_ZIP=TensorRT-${TENSORRT_VERSION}.Windows10.x86_64.cuda-11.1.cudnn8.0.zip
+
+ARG CUDNN_VERSION=8.0.5.39
+ARG CUDNN_ZIP=cudnn-11.1-windows-x64-v${CUDNN_VERSION}.zip
```
GuanLuo, line 52: Does 11.1 also refer to CUDA version? Should it be changed to variables?

deadeyegoodwin: Yes, but these names are from the public download area for cudnn and they will need to be explicitly found and downloaded each time we need to change them. So I don't want to use a variable to make it seem like it is automatically determined just by changing the cuda version.

GuanLuo: Okay

GuanLuo (approved):

YongHuangSJTU:
Hi @deadeyegoodwin , we are considering integrating Triton C API into an existing system which doesn't use docker, so I need to build without docker

I replied the issue of cloud dependency https://github.com/triton-inference-server/server/issues/2800, and posted two other issues about windows build https://github.com/triton-inference-server/server/issues/2834, https://github.com/triton-inference-server/server/issues/2833

Thank you for your support

PR #2820, opened 2021-05-06
Add SageMaker endpoints. Add test against sagemaker endpoints

GuanLuo:
Also remove the old checking on `port == -1` because now we use explicit cmdline option for enabling endpoints


build/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ option(TRITON_ENABLE_ENSEMBLE "Include ensemble support in server" OFF)
 # Endpoints
 option(TRITON_ENABLE_HTTP "Include HTTP API in server" ON)
 option(TRITON_ENABLE_GRPC "Include GRPC API in server" ON)
+option(TRITON_ENABLE_SAGEMAKER "Include AWS SageMaker API in server" ON)
```
CoderHam, line 49: Not sure if SageMaker has a GRPC API but might be useful to call this SageMaker HTTP?

src/servers/http_server.cc
```diff
@@ -2909,4 +2975,20 @@ HTTPServer::CreateMetricsServer(
 #endif  // TRITON_ENABLE_METRICS
 }
 
+TRITONSERVER_Error*
```
CoderHam, line 2978: wrap this in TRITON_ENABLE_SAGEMAKER?

src/servers/http_server.cc
```diff
@@ -2868,11 +2772,173 @@ HTTPAPIServer::InferRequestClass::FinalizeResponse(
...
+  }
+
+  LOG_VERBOSE(1) << "HTTP error: " << req->method << " " << req->uri->path->full
+                 << " - " << static_cast<int>(EVHTP_RES_BADREQ);
+
+  evhtp_send_reply(req, EVHTP_RES_BADREQ);
+}
+
+
+// Handle Sagemaker HTTP requests to inference server APIs
```
CoderHam, line 2892: wrap this in TRITON_ENABLE_SAGEMAKER?

src/servers/http_server.h
```diff
@@ -47,6 +47,13 @@ class HTTPServer {
       const std::shared_ptr<TRITONSERVER_Server>& server, int32_t port,
       int thread_cnt, std::unique_ptr<HTTPServer>* metrics_server);
 
+  static TRITONSERVER_Error* CreateSagemakerAPIServer(
```
CoderHam, line 50: Here too?

src/servers/main.cc
```diff
@@ -69,14 +69,19 @@ namespace {
 int32_t repository_poll_secs_ = 15;
 
 // The HTTP, GRPC and metrics service/s and ports. Initialized to
-// default values and modifyied based on command-line args. Set to -1
```
CoderHam, line 72: What is the motivation to remove the -1 logic for disabling the protocol?

CoderHam (changes requested):

src/servers/main.cc
GuanLuo: It's confusing to have two ways to control the endpoints and we don't really publicly document that setting port to -1 means disabling the endpoint, so this is for internal use and only using `allow_xxx` parameter will be more clear. Thus I remove it

build/CMakeLists.txt
GuanLuo: No, sagemaker endpoint will always use HTTP network protocol. SageMaker is a protocol on its own, from the user perspective it is actually different from the HTTP / GRPC protocol that we [defined](https://github.com/triton-inference-server/server/tree/master/docs/protocol), so I think we should keep it as it is so it is not confused with HTTP protocol.

src/servers/main.cc
CoderHam: But if the user builds the server with both endpoints, it will always launch both/all endpoints. I don't think removing this because it isn't documented is the best idea. I think it's a nice feature that should instead of documented

build/CMakeLists.txt
CoderHam: Makes sense

src/servers/http_server.cc
```diff
@@ -2868,11 +2772,173 @@ HTTPAPIServer::InferRequestClass::FinalizeResponse(
...
+class SagemakerAPIServer : public BaseAPIServer {
+ public:
```
tanmayv25, line 2894: I would advice placing SagemakeAPIServer class in a file of its own.. 
call it `sagemaker_server.cc`.

src/servers/main.cc
GuanLuo: An example to disable endpoint: https://github.com/triton-inference-server/server/pull/2820/files#diff-e15379ef713d35b70b5e1def42a8b3078083b9c68e2e25cf5ef4957fb52e29aaR86

Won't this be a nicer way to do so? `allow-xxx` is listed as a option so we need to ensure backward compatibility, but `port == -1` is not. Again, the functionality is duplicated and I think it is better to keep either one but not both, since backward compatibility is not a concern.

CoderHam: Oh yes, I see what you mean now. Go ahead

src/servers/http_server.cc
GuanLuo: I agree, the `http_server.cc` seems to be able to be split into files for different endpoints (HTTP / metrics / sagemaker). But I think I will keep `http_server.h` as it is so conceptually it lists the endpoints that uses HTTP as the underlying network protocol. The alternative is also separating `http_server.h` into `xxx_server.h`, then the endpoints are represented separately, which I think is better from the encapsulation perspective.

GuanLuo:
@deadeyegoodwin The SageMaker endpoint PR is ready for review.

deadeyegoodwin (approved):

CoderHam (approved):

philschmid:
Hello, 

Thank you for implementing this!🚀  Do you already know when this will be included in an official release?

deadeyegoodwin:
> 
> 
> Hello,
> 
> Thank you for implementing this!🚀 Do you already know when this will be included in an official release?

We still have a little work left to do before it is complete, but we are planning on having it available in the 21.06 release (late June).

gilinachum:
Will there be an example on how to deploy Triton as a realtime endpoint in SageMaker as part of the 21.06 release?
Additionally, will model concurrency / GPU, be supported?

deadeyegoodwin:
There will be documentation available with the release. This initial implementation supports only the sagemaker single-model mode. Multi-model support will come in the future.
Note that even with a single model you can have multiple instances of that model, and GPU is supported.

PR #2823, opened 2021-05-06
Increase stub timeout for python backend

Tabrizian:


deadeyegoodwin (approved):

qa/L0_infer/test.sh
```diff
@@ -88,7 +88,7 @@ else
 fi
 
 # Allow more time to exit. Ensemble brings in too many models
-SERVER_ARGS_EXTRA="--exit-timeout-secs=120 --backend-directory=${BACKEND_DIR} --backend-config=tensorflow,version=${TF_VERSION}"
+SERVER_ARGS_EXTRA="--exit-timeout-secs=120 --backend-directory=${BACKEND_DIR} --backend-config=tensorflow,version=${TF_VERSION} --backend-config=python,stub-timeout-seconds=120"
```
CoderHam, line 91: I think both the TF and python backend config args should be only added if the corresponding BACKENDS are enabled. Maybe you can fix this as a part of this change?

CoderHam (changes requested):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #2825, opened 2021-05-06
Fix trace lifecycle in GRPC

GuanLuo:
@tanmayv25 The idea is that a trace object will be created whenever GRPC frontend is ready for next request, so there will always be trace objects that the GRPC frontend is responsible to delete, this change is addressing that.

### DIFF
--- a/src/servers/grpc_server.cc
+++ b/src/servers/grpc_server.cc
@@ -1965,6 +1965,13 @@ class InferHandlerState {
       const std::shared_ptr<Context>& context, Steps start_step = Steps::START)
       : tritonserver_(tritonserver)
   {
+#ifdef TRITON_ENABLE_TRACING
+    trace_manager_ = nullptr;
+    trace_ = nullptr;
+    trace_id_ = 0;
+    trace_userp_ = nullptr;
+#endif  // TRITON_ENABLE_TRACING
+
     // For debugging and testing,
     const char* dstr = getenv("TRITONSERVER_DELAY_GRPC_RESPONSE");
     delay_response_ms_ = 0;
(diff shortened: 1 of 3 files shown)


src/servers/tracer.h
```diff
@@ -80,6 +81,7 @@ class TraceManager {
   std::mutex mu_;
   uint32_t trace_cnt_;
 
+
```
tanmayv25, line 84: Remove this additional line

src/servers/grpc_server.cc
```diff
@@ -3448,8 +3477,9 @@ ModelStreamInferHandler::StartNewRequest()
   state->trace_manager_ = nullptr;
   state->trace_ = nullptr;
   state->trace_id_ = 0;
+  state->trace_userp_ = nullptr;
   if (trace_manager_ != nullptr) {
-    state->trace_ = trace_manager_->SampleTrace();
+    state->trace_ = trace_manager_->SampleTrace(&state->trace_userp_);
```
tanmayv25, line 3482: Correct me if I am wrong, isn't `trace_userp`  passed to `SampleTrace` always nullptr?

src/servers/tracer.cc
```diff
@@ -101,18 +101,21 @@ TraceManager::SampleTrace()
...
       &trace, level_, 0 /* parent_id */, TraceActivity, TraceRelease,
-      userp.get());
+      luserp.get());
   if (err != nullptr) {
     LOG_TRITONSERVER_ERROR(err, "creating inference trace object");
     return nullptr;
   }
 
-  userp.release();
+  if (userp != nullptr) {
```
tanmayv25, line 115: In which case is userp not nullptr?

tanmayv25 (changes requested):

src/servers/grpc_server.cc
GuanLuo: Note that the pointer to the class member is passed, so it will not be nullptr. This is similar to `TRITONBACKEND_InputProperties` where you will pass an non-nullptr if you want to know the property associated with the object, here is the userp which the trace manager newed.

For HTTP frontend, it will be nullptr because the HTTP frontend doesn't need to release the trace and related data directly. The trace object will only be created when an inference request is received and ownership of the trace object will be passed into server core.

tanmayv25: Ahh.. Makes sense!

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #2829, opened 2021-05-07
Fix minor typo

CoderHam:
For the timestamp for batcher_cudashm to be reported corrected we need to use GPU instance and not CPU. The cudaSetDevice is only triggered for GPU instances

### DIFF
--- a/qa/L0_batcher/test.sh
+++ b/qa/L0_batcher/test.sh
@@ -159,7 +159,7 @@ done
 # Create allow-ragged model to variable-size model repository
 cp -r ../custom_models/custom_zero_1_float32 var_models/. && \
     (cd var_models/custom_zero_1_float32 && mkdir 1 && \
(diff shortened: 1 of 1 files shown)


GuanLuo:
I would rather send a dummy request to the model to warm it up (and document why) instead of changing the test. So it seems like any CPU model will experience this delay just be cause the I/O data will be in device memory, why we didn't observe that previously.

CoderHam:
The cudaSetDevice was called previously even for non-GPU models. It was called from the custom backend and not the identity model created using it. 

In the new identity model it is its own backend. In this case we would indeed see the delay for a CPU model. We didn't modify the test. We modified the backend to behave correctly. It should not call cudaSetDevice for a CPU model and hence it would init cuda only when needed i.e. when it needs to copy the input data from GPU

GuanLuo (approved):

PR #2839, opened 2021-05-10
Fix L0_batcher for windows

deadeyegoodwin:


CoderHam (changes requested):
You should also wrap https://github.com/triton-inference-server/server/blob/687397d89386d4c96088149151a776a301e09259/qa/L0_batcher/test.sh#L160
with 
```
if [ "$OS_WINDOWS" -eq "0" ]; then
...
fi
```

PR #2859, opened 2021-05-14
Add the Optimization Profile APIs for TRT backend

tanmayv25:
I have separated this change in a separate PR from tanmay-dev as it can go for now. I have tested it for TRT backend repo and it passes L0_infer_variable.
This will prevent frequent rebases.
Linked PR on core: https://github.com/triton-inference-server/core/pull/20

### DIFF
--- a/src/backends/backend/triton_model_instance.cc
+++ b/src/backends/backend/triton_model_instance.cc
@@ -36,7 +36,7 @@ namespace nvidia { namespace inferenceserver {
 TritonModelInstance::TritonModelInstance(
     TritonModel* model, const std::string& name, const size_t index,
     const TRITONSERVER_InstanceGroupKind kind, const int32_t device_id,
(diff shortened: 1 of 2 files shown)


src/backends/backend/triton_model_instance.cc
```diff
@@ -69,6 +70,10 @@ TritonModelInstance::CreateInstances(
     TritonModel* model, const inference::ModelConfig& model_config)
 {
   for (const auto& group : model_config.instance_group()) {
+    std::set<std::string> profile_names;
```
deadeyegoodwin, line 73: The order of the profile names is important since you access them by index and because we should present them in the same order as they appear in the model configuration, so you should use a vector here. You can't rely on the ordering in a set.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #2861, opened 2021-05-14
CAPI for perf analyzer unit tests

jbkyang-nvi:
adding unit test for https://github.com/triton-inference-server/client/pull/8


qa/L0_perf_analyzer_capi/json_input_data_files/float_data_with_shape.json
```diff
@@ -0,0 +1,69 @@
+{
```
CoderHam, line 1: This duplicates files present in the L0_perf_analyzer dir

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+fi
+if [ $(cat $CLIENT_LOG |  grep "${ERROR_STRING}" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+
+# Shape tensor I/O model (server needs the shape tensor on the CPU)
+
```
CoderHam, line 332: Extra empty line

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+fi
+if [ $(cat $CLIENT_LOG |  grep "${ERROR_STRING}" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+
+# Testing with variable inputs
+
```
CoderHam, line 289: Extra empty line

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+    RET=1
+fi
+if [ $(cat $CLIENT_LOG |  grep "${ERROR_STRING}" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Testing with different file inputs
+set +e
```
CoderHam, line 261: `set +e` must be accompanied by `set -e`

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Test perf client behavior on different model with different batch size
+for MODEL in graphdef_nobatch_int32_int32_int32 graphdef_int32_int32_int32; do
+    # Valid batch size
+    $PERF_ANALYZER -v  -m $MODEL -t 1 -p2000 -b 1 \
+    --service-kind=triton_local --model-repo=$DATADIR --library-name=$SERVER_LIBRARY_PATH >$CLIENT_LOG 2>&1
```
CoderHam, line 213: Reminder to modify triton_local to triton_c_api as based on suggestions on client side PR

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+--service-kind=triton_local --model-repo=$DATADIR --library-name=$SERVER_LIBRARY_PATH >$CLIENT_LOG 2>&1
+if [ $? -eq 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+
+# Testing with inception model
+
```
CoderHam, line 180: Extra line

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+cp -r /data/inferenceserver/${REPO_VERSION}/perf_model_store/resnet50v1.5_fp16_savedmodel $DATADIR
+
+# Generating test data
+mkdir -p $TESTDATADIR
+for INPUT in INPUT0 INPUT1; do
+    for i in {1..16}; do
+        echo '1' >> $TESTDATADIR/${INPUT}
+    done
+done
+RET=0
```
CoderHam, line 103: Add a blank line before and after this

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+# C API tests
+# Make sure tritonserver is not running first
+SERVER_PID=$(pidof tritonserver)
+if [ $? -ne 1 ]; then
+echo -e "\n There was a previous instance of tritonserver, killing \n"
+  kill $SERVER_PID
+  wait $SERVER_PID
+fi
+
+# #Testing simple configurations
```
CoderHam, line 114: Fix comments here and later

CoderHam (changes requested):

deadeyegoodwin (commented):
Is this just a copy of L0_perf_analyer? Can we instead just make a loop within that test that runs triton and triton-c-api modes?

qa/L0_perf_analyzer_capi/test.sh
tanmayv25: Still needs this to be fixed..

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+if [ $(cat $CLIENT_LOG | grep "${ERROR_STRING}" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+
+#Testing that async does NOT work
+
+$PERF_ANALYZER -v -m graphdef_int32_int32_int32 -t 1 -p2000 -b 1 -a \
```
tanmayv25, line 146: A better test would be to check for the specific error message.

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,396 @@
...
+
+# Generating test data
+mkdir -p $TESTDATADIR
+for INPUT in INPUT0 INPUT1; do
+    for i in {1..16}; do
+        echo '1' >> $TESTDATADIR/${INPUT}
+    done
+done
+RET=0
+########## Test C API #############
```
tanmayv25, line 104: Most of these tests are just testing the operation of different options with perf_analyzer. I don't think it is a good idea to completely replicate the L0_perf_analyzer to work with L0_perf_analyzer_capi. A better way would be to extract just the cases of interest for c_api client backend. We can be more intelligent in designing these tests for more targeted testing..

CoderHam:
> Is this just a copy of L0_perf_analyer? Can we instead just make a loop within that test that runs triton and triton-c-api modes?

I agree. If it indeed is identical to L0_perf_analyzer we can merge both. You can use conditionals to enable the testing for the C API that is supported with your current changes.

qa/L0_perf_analyzer_capi/json_input_data_files/float_data_with_shape.json
tanmayv25: As @CoderHam  suggests, you can reuse these files from L0_perf_analyzer. If you are worried about aesthetics then place the json_input_data_files to `qa/common` with name `perf_analyzer_test_files/json_input_data`. 
Both the solutions will be better than replicating image_data.json.

tanmayv25 (changes requested):

tanmayv25:
> Is this just a copy of L0_perf_analyer? Can we instead just make a loop within that test that runs triton and triton-c-api modes?

I think it will test lots of unnecessary cases in the name of coverage. But yes having a single test file will be helpful in developing test for future enhancements. I think `test.sh` has already grown out of proportions. It might be a good idea to break it into python subtests.

jbkyang-nvi:
> > Is this just a copy of L0_perf_analyer? Can we instead just make a loop within that test that runs triton and triton-c-api modes?
> 
> I agree. If it indeed is identical to L0_perf_analyzer we can merge both. You can use conditionals to enable the testing for the C API that is supported with your current changes.

Since we're starting the server again, I wanted to make sure the server that's started is indeed from the C API and not because we run `run_server`. I agree this is non-ideal for repeated cases. However, since the C API does not yet have the same functionality as the other backend, I think it might be cleaner this way, and will be easier to see what this different feature supports. What do you think @CoderHam  @deadeyegoodwin

qa/L0_perf_analyzer_capi/test.sh
jbkyang-nvi: Since we do a test loop for `grpc` and `http`, I think it is fair if we do the same tests for c_api as well. That being said, I'm also unhappy with how there's a lot of duplicated code to just do testing.

jbkyang-nvi: Please let me know how to do this better?

tanmayv25: The reason for looping `grpc` and `http` was because we don't really test our C++ client library in CI and use python client library. Our C++ client testing is limited to examples and perf_analyzer tests. C_API on the other hand is being tested by all our end to end tests.
 It's not such a bad idea to add another loop around C_API tests for its simplicity. But if we are going to have a separate L0_per_analyzer_capi then I wanted to highlight that we are not testing anything new with these cases and can be safely removed:
 1. Different input_data options for perf_analyzer. (json, synthetic and directory)
 2. I don't think large tensor size case will add any benefit either.
 3. Request rate manager. (why is this not supported? I believe it interacts with the same interface with client backend)
 4. Runs with a range of concurrency or request rate values.
 The cases which we must be most interested in:
 1. Basic inference with multiple I/O.
 2. String input
 3. Variable shaped tensors
 4. Shape tensors
 5. Async inference (whenever it is supported)
 6. Sequence Model (whenever it is supported)
 7. Shared Memory (whenever it is supported)

tanmayv25: > However, since the C API does not yet have the same functionality as the other backend, I think it might be cleaner this way, and will be easier to see what this different feature supports. 

I see merit in this point. If you think it will be easier with follow-on work then I am fine resolving this conversation,

CoderHam:
I am ok having a different L0_perf_analyzer_capi for now if there are enough differences to make the current L0_perf_anlyzer look cluttered. 
Additionally I would then try to test that perf analyzer starts server itself in the beginning and then kill the server. Then go on to run these cases after that in a since server session. (Can you confirm that it can indeed do this?)
It would definitely shorten the test a little to not have to spawn and kill the server again and again.@tanmayv25 @deadeyegoodwin what do you think?

tanmayv25:
I am fine with the separate L0_perf_analyzer_capi for now too. It saves clutter from L0_perf_analyzer test and helps @jbkyang-nvi  adding follow-up features with the best coverage. We must revisit these tests sometime in future when tritoncapi functionality in perf_analyzer is complete.

qa/L0_perf_analyzer/test.sh
```diff
@@ -43,14 +43,14 @@ PERF_ANALYZER=../clients/perf_analyzer
...
 
-INT_JSONDATAFILE=`pwd`/json_input_data_files/int_data.json
-INT_DIFFSHAPE_JSONDATAFILE=`pwd`/json_input_data_files/int_data_diff_shape.json
-FLOAT_DIFFSHAPE_JSONDATAFILE=`pwd`/json_input_data_files/float_data_with_shape.json
-STRING_JSONDATAFILE=`pwd`/json_input_data_files/string_data.json
-STRING_WITHSHAPE_JSONDATAFILE=`pwd`/json_input_data_files/string_data_with_shape.json
-SEQ_JSONDATAFILE=`pwd`/json_input_data_files/seq_data.json
-SHAPETENSORADTAFILE=`pwd`/json_input_data_files/shape_tensor_data.json
-IMAGE_JSONDATAFILE=`pwd`/json_input_data_files/image_data.json
+INT_JSONDATAFILE=`pwd`/../common/json_input_data_files/int_data.json
```
tanmayv25, line 46: json_input_data_files --> perf_analyzer_input_data_json

tanmayv25:
You would also need to raise a gitlab side MR that enable L0_perf_analyzer_capi on our CI.

tanmayv25 (approved):

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,296 @@
...
+fi
+set -e
+
+if [ $RET -eq 0 ]; then
+  echo -e "\n***\n*** Test Passed\n***"
+else
+  echo -e "\n***\n*** Test FAILED\n***"
+fi
+exit $RET
+set +x
```
CoderHam, line 296: Add a newline at the end

CoderHam (changes requested):

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,295 @@
...
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+if [ $(cat $CLIENT_LOG | grep ": 0 infer/sec\|: 0 usec" | wc -l) -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# # TODO: simple_savedmodel_sequence_object segfaults with CAPI
```
CoderHam, line 206: Remove double comments and modify to "TODO: Re-enable after sequence model support if fixed for CAPI"

qa/L0_perf_analyzer_capi/test.sh
```diff
@@ -0,0 +1,295 @@
...
+#     echo -e "\n***\n*** Test Failed\n***"
+#     RET=1
+# fi
+# if [ $(cat $CLIENT_LOG |  grep "${ERROR_STRING}" | wc -l) -ne 0 ]; then
+#     cat $CLIENT_LOG
+#     echo -e "\n***\n*** Test Failed\n***"
+#     RET=1
+# fi
+# 
+# # TODO: Testing with variable ensemble model doesn't work
```
CoderHam, line 222: "TODO: Re-enable after variable model support if fixed for CAPI"

CoderHam (changes requested):

CoderHam (approved):

PR #2862, opened 2021-05-14
Add custom+ensemble testing to windows

CoderHam:


qa/L0_batcher/test.sh
```diff
@@ -432,7 +427,6 @@ done
 # by comparing the "response send" timestamps.
 TEST_CASE=test_multi_batch_preserve_ordering
 
-# Skip test for Windows. Does not build identity backend
 if [[ "$(< /proc/sys/kernel/osrelease)" != *Microsoft ]]; then
```
deadeyegoodwin, line 430: You removed the comment but we still seem to be skipping the test?

CoderHam: Oh yes, fixing it now. We cannot enable this yet because of the trace file issue -> concats at 8192 chars.

deadeyegoodwin (approved):

PR #2863, opened 2021-05-14
[New Threading Model] For Single Instance, Dynamic Batcher, Execution…

tanmayv25:
… Blocking

I will use `tanmayv-dev` for all the developments related to  TRT backend, new threading model and rate limiter.`*` 
This PR includes changes to be made in dynamic_batcher_scheduler and a brief idea of new organization.

 For multiple instances, all the remaining logic will be implemented in TritonModelInstance class which includes separate threads for executing inference. The warmup logic will also be moved to TritonModelInstance. Even after these future enhancements, the threading behavior for single instance will remain the same. i.e. for a single instance, scheduler thread itself will jump into backend for execution.
 
 Please ignore `#if 0 ... #endif ` and warmup related sections in the code. I will clean that up once I get to it. 
 
 These changes passes most of the tests working on TRT backend. 
 
 `*` Until and unless all these 3 features are implemented fully master will be in broken state.

### DIFF
--- a/src/backends/backend/triton_model_instance.cc
+++ b/src/backends/backend/triton_model_instance.cc
@@ -171,222 +171,220 @@ TritonModelInstance::WarmUp()
 {
   LOG_ERROR << "WarmUp is not yet supported";
   return Status::Success;
-#if 0
-  std::vector<TRITONBACKEND_Request*> triton_requests(1024);
-  triton_requests.clear();
-  for (auto& request : sample.requests_) {
-    // Capture timestamp before run to avoid incorrect accumulation from
-    // sequential warmup runs
-#ifdef TRITON_ENABLE_STATS
-    request->CaptureRequestStartNs();
-#endif  // TRITON_ENABLE_STATS
-    request->CaptureQueueStartNs();
-    triton_requests.push_back(
-        reinterpret_cast<TRITONBACKEND_Request*>(request.release()));
-  }
-  TRITONBACKEND_ModelInstance* triton_model_instance =
-      reinterpret_cast<TRITONBACKEND_ModelInstance*>(this);
-  TritonBackend::TritonModelInstanceExecFn_t inst_exec_fn =
-      backend_->ModelInstanceExecFn();
-
-  // If there is an error then we retain ownership of 'requests'
-  // and must send error responses.
-  TRITONSERVER_Error* err = inst_exec_fn(
-      triton_model_instance, &triton_requests[0], triton_requests.size());
-  if (err != nullptr) {
-    Status status = Status(
-        TritonCodeToStatusCode(TRITONSERVER_ErrorCode(err)),
-        TRITONSERVER_ErrorMessage(err));
-    for (TRITONBACKEND_Request* tr : triton_requests) {
-      std::unique_ptr<InferenceRequest> ur(
-          reinterpret_cast<InferenceRequest*>(tr));
-      InferenceRequest::RespondIfError(ur, status, true /* release_requests */);
-    }
-
-    TRITONSERVER_ErrorDelete(err);
-  }
-#endif
+//  std::vector<TRITONBACKEND_Request*> triton_requests(1024);
+//  triton_requests.clear();
+//  for (auto& request : sample.requests_) {
+//    // Capture timestamp before run to avoid incorrect accumulation from
+//    // sequential warmup runs
+//#ifdef TRITON_ENABLE_STATS
+//    request->CaptureRequestStartNs();
+//#endif  // TRITON_ENABLE_STATS
+//    request->CaptureQueueStartNs();
+//    triton_requests.push_back(
+//        reinterpret_cast<TRITONBACKEND_Request*>(request.release()));
+//  }
+//  TRITONBACKEND_ModelInstance* triton_model_instance =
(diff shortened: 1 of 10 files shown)


CoderHam (changes requested):
Could you comment the sections of code instead of using `#if 0 ... #endif`?

src/core/dynamic_batch_scheduler.cc
```diff
@@ -198,60 +155,44 @@ DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
     // Assuming no error is returned, this call takes ownership of
     // 'request' and so we can't use it after this point.
     RETURN_IF_ERROR(queue_.Enqueue(request->Priority(), request));
-
-    // If there are any idle runners and the queued batch size is greater or
-    // equal to next preferred batch size, then wake one up to service this
-    // request. We do the actual wake outside of the lock to avoid having the
-    // woken thread immediately block on the lock
-    wake_runner = (idle_scheduler_thread_cnt_ > 0);
```
deadeyegoodwin, line 206: I think you want to keep this logic. We don't want to wakeup the scheduler thread unless these conditions are true.

CoderHam (approved):

src/core/dynamic_batch_scheduler.cc
```diff
@@ -198,60 +155,44 @@ DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
...
-                 << " at default nice...";
+  LOG_VERBOSE(1) << "Starting dynamic-batch scheduler thread for "
+                 << model->Name() << " at default nice...";
 #endif
 
   // Initialize using the thread. If error then just exit this thread
   // now... that means the corresponding model instance will not have
   // any runner and so will not get used for execution.
-  Status startup_status = OnInit_(runner_id);
+  Status startup_status = model->Initialize();
```
deadeyegoodwin, line 186: This initialize and warmup are ok for now, but given the new reorg we are moving towards they don't belong here. When a TritonModel object is created it can initialize and warmup itself, and then create the scheduler for itself.

deadeyegoodwin (changes requested):

src/core/dynamic_batch_scheduler.cc
tanmayv25: We will not hit this logic as there is only a single scheduler thread in the system. 
For multiple instances, if a request is being allocated to the last available BackendThread, then I will block the scheduler thread just after the hand off till a BackendThread returns from execution.

src/core/dynamic_batch_scheduler.cc
tanmayv25: I agree... I will keep note of this in my next PR for handling multiple instances.

src/core/dynamic_batch_scheduler.cc
deadeyegoodwin: I think it may be better to not block the scheduler thread. There is work for the scheduler thread to do as requests come in and it would be good to allow it do continue doing that work even if no instances are available. The rate limiter will make sure that the scheduler thread does not call back into the schedule/execute function until an instance is actually available.

src/backends/backend/triton_model_instance.cc
```diff
@@ -128,6 +128,264 @@ TritonModelInstance::CreateInstance(
   return Status::Success;
 }
 
+Status
+TritonModelInstance::Schedule(
+    std::vector<std::unique_ptr<InferenceRequest>>&& requests)
+{
+  // Use a thread local vector to avoid needing to malloc each
+  // time an inference is run.
+  thread_local std::vector<TRITONBACKEND_Request*> triton_requests(1024);
```
GuanLuo, line 137: Note that this may need to change later on when multiple instances share the same thread, or at least need to be documented in `TRITONBACKEND_ModelInstanceExecute()` that the `requests` may be overwritten once the function returns. For instance, in TRT the requests are released in the `ProcessResponse()` and that may happen after the next batch is scheduled, so instead of passing `requests` to the `ProcessResponse()`, the request handles in `requests` should be copied.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -411,20 +355,20 @@ DynamicBatchScheduler::SchedulerThread(
...
       }
     }
 
-    if (wake_thread) {
-      cv_.notify_one();
-    }
-
     if (!requests.empty()) {
-      OnSchedule_(runner_id, std::move(requests));
+      model->Instances()[instance_index]->Schedule(std::move(requests));
```
GuanLuo, line 364: Just want to make sure that later on the thread pool will be invoked in `Schedule` either to delegate the execution to other thread or to let the scheduler thread to call into the instance execution?

tanmayv25: Yes.. For multiple instances, the requests will be delegated to BackendThread within TritonModelInstance

src/backends/backend/triton_model_instance.cc
tanmayv25: Thanks for the insight! I will keep that in mind for DEVICE_BLOCKING implementation.

deadeyegoodwin (approved):

PR #2869, opened 2021-05-17
Fix documentation typos

avant1:


CoderHam (approved):
Thanks for these fixes @avant1

PR #2878, opened 2021-05-18
[Rate Limiter] Single Instance, Exec blocking: Integrate rate limiter with dynamic b…

tanmayv25:
…atch scheduler

This has some logic into scheduler for handling mulitple instances. The threading hand off for BackendThreads with mulitple instances will be shared soon.

RateLimiter unit tests needs to be updated to handle triton_model and triton_model_instance abstractions.

Linked core PR: https://github.com/triton-inference-server/core/pull/22

### DIFF
--- a/src/core/dynamic_batch_scheduler.cc
+++ b/src/core/dynamic_batch_scheduler.cc
@@ -387,10 +387,11 @@ DynamicBatchScheduler::SchedulerThread(
     }
 
     if (!requests.empty()) {
-      auto sched_cb = [&requests, &pending_request_rate_limiter](
+      auto sched_cb = [this, &requests, &pending_request_rate_limiter](
                           RateLimiter::ModelInstance* mi) {
         mi->ScheduleNow(std::move(requests));
         pending_request_rate_limiter.store(false);
+        this->cv_.notify_all();
       };
       pending_request_rate_limiter.store(true);
       rate_limiter_->RequestModelInstance(sched_cb, model_);
--- a/src/core/dynamic_batch_scheduler.cc
+++ b/src/core/dynamic_batch_scheduler.cc
@@ -35,18 +35,20 @@
 #include "src/core/logging.h"
 #include "src/core/model_config.h"
 #include "src/core/nvtx.h"
+#include "src/core/server.h"
 
 namespace nvidia { namespace inferenceserver {
 
 DynamicBatchScheduler::DynamicBatchScheduler(
-    const bool dynamic_batching_enabled, const int32_t max_batch_size,
+    TritonModel* model, const bool dynamic_batching_enabled,
+    const int32_t max_batch_size,
     const std::unordered_map<std::string, bool>& enforce_equal_shape_tensors,
     const bool preserve_ordering,
     const std::set<int32_t>& preferred_batch_sizes,
     const uint64_t max_queue_delay_microseconds,
     const inference::ModelQueuePolicy& default_queue_policy,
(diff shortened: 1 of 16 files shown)


src/core/dynamic_batch_scheduler.cc
```diff
@@ -35,18 +35,20 @@
 #include "src/core/logging.h"
 #include "src/core/model_config.h"
 #include "src/core/nvtx.h"
+#include "src/core/server.h"
 
 namespace nvidia { namespace inferenceserver {
 
 DynamicBatchScheduler::DynamicBatchScheduler(
-    const bool dynamic_batching_enabled, const int32_t max_batch_size,
+    TritonModel* model, const bool dynamic_batching_enabled,
```
deadeyegoodwin, line 43: Should use shared_ptr for 'model', not a raw pointer.

src/core/dynamic_batch_scheduler.h
```diff
@@ -103,6 +106,8 @@ class DynamicBatchScheduler : public Scheduler {
   std::mutex mu_;
   std::condition_variable cv_;
 
+  RateLimiter* rate_limiter_;
```
deadeyegoodwin, line 109: Use shared_ptr, avoid raw pointers...

src/core/server.cc
```diff
@@ -149,6 +149,12 @@ InferenceServer::Init()
     status = CommonErrorToStatus(triton::common::AsyncWorkQueue::Initialize(
         buffer_manager_thread_count_));
   }
+
+  bool ignore_resources_and_priority =
+      (rate_limit_mode_ == RateLimitMode::RL_OFF);
```
deadeyegoodwin, line 154: Default to RL_EXEC_COUNT

src/core/tritonserver.cc
```diff
@@ -1635,6 +1724,32 @@ TRITONSERVER_ServerNew(
   options_table.InsertRow(std::vector<std::string>{
       "strict_model_config",
       std::to_string(lserver->StrictModelConfigEnabled())});
+  std::string rate_limit;
```
deadeyegoodwin, line 1727: Make RateLimitModeToString function.

deadeyegoodwin (changes requested):

src/core/dynamic_batch_scheduler.cc
tanmayv25: The scheduler object is held within TritonModel(actually InferenceBackend from which TritonModel extends from). Currently we can not expose the TritonModel class to backend.h as it interferes with ensemble_scheduler. The raw pointer to model should be safe inside Scheduler as the scheduler is owned by TritonModel object itself. I understand the need to avoid raw pointers. But I think we can address it as a separate clean up effort to remove InferenceBackend. I will document the requirement.
Forcing a use of shared_ptr will destroy backend alongwith scheduler.

src/core/dynamic_batch_scheduler.h
tanmayv25: Same as above. rate_limiter is actually held within server object already held as a raw pointer inside TritonModel.

src/core/server.cc
tanmayv25: Default mode is indeed RL_EXEC_COUNT.  This is done just to ignore resources and priority settings in case RateLimitMode is OFF. I will move this logic within rate limiter implementation to avoid confusion.

tanmayv25: Actually, that mismatches with how we handle model_control_mode_. I will leave the operation as is.

deadeyegoodwin (approved):

PR #2881, opened 2021-05-19
Remove nvidia from client namespace

jbkyang-nvi:
Change unit test to reflect changes in the client

Changed code:
https://github.com/triton-inference-server/client/pull/14

### DIFF
--- a/qa/L0_sdk/grpc_test.cc
+++ b/qa/L0_sdk/grpc_test.cc
@@ -26,13 +26,15 @@
 
 #include "grpc_client.h"
 
+namespace tc = triton::client;
+
 int
 main(int argc, char* argv[])
 {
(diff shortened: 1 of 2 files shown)


deadeyegoodwin (dismissed):

qa/L0_sdk/grpc_test.cc
```diff
@@ -26,15 +26,13 @@
 
 #include "grpc_client.h"
 
-namespace nvic = nvidia::inferenceserver::client;
```
CoderHam, line 29: Can you keep this namespace. Personally I think its nice that we use it. 
```
tc= triton::client;
```

CoderHam (changes requested):

CoderHam (approved):

PR #2891, opened 2021-05-20
Support multiple instances with TritonBackendThread. Fix dynamic sche…

tanmayv25:
…duler

Adds the TritonBackendThread implementation within TritonModelInstances. This does not implement DEVICE_BLOCKING yet. For instance_count > 1, the BackendThread corresponding to ModelInstance will be used to call model Initialize, WarmUp and Schedule.

Also fixed the dynamic batch scheduler to ensure the scheduler thread corresponding to the same TritonModel gets to run its inference everytime. I have successfully done some functional testing and am looking on the affect on performance of this new threading model on non-TRT backends.

### DIFF
--- a/src/backends/backend/triton_model_instance.cc
+++ b/src/backends/backend/triton_model_instance.cc
@@ -162,7 +162,6 @@ TritonModelInstance::SetBackendThread()
   // TODO: Currently each instance has a dedicated TritonBackendThread.
   // For device blocking execution policy we must share the
   // TritonBackendThread object with the instances on same device.
-  // FIXME: Get the nice value for the thread from config
   std::unique_ptr<TritonBackendThread> local_backend_thread;
   RETURN_IF_ERROR(TritonBackendThread::CreateBackendThread(
       Name(), 0 /* nice */, &local_backend_thread));
--- a/src/backends/backend/triton_model_instance.cc
+++ b/src/backends/backend/triton_model_instance.cc
@@ -26,10 +26,16 @@
 
 #include "src/backends/backend/triton_model_instance.h"
 
+#ifndef _WIN32
+#include <sys/resource.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+#endif
 #include "model_config.pb.h"
 #include "src/backends/backend/triton_model.h"
 #include "src/core/logging.h"
 #include "src/core/metrics.h"
+#include "src/core/nvtx.h"
 
 namespace nvidia { namespace inferenceserver {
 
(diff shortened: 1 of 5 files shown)


src/backends/backend/triton_model_instance.cc
```diff
@@ -131,9 +157,69 @@ TritonModelInstance::CreateInstance(
 }
 
 Status
+TritonModelInstance::SetBackendThread()
+{
+  // TODO: Currently each instance has a dedicated TritonBackendThread.
+  // For device blocking execution policy we must share the
+  // TritonBackendThread object with the instances on same device.
+  // FIXME: Get the nice value for the thread from config
```
deadeyegoodwin, line 165: We can remove the "nice" setting that we are currently using for prioritization as it is not really effective and we now have the rate limiter for managing prioritization.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -251,169 +246,217 @@ DynamicBatchScheduler::SchedulerThread(
...
-            if (status.IsOk()) {
-              requests.emplace_back(std::move(request));
-            } else {
-              // The queue is empty which conflicts with pending batch count.
-              // Send the current batch if any and reset related variables.
-              LOG_ERROR << "Failed to retrieve request from scheduler queue: "
+    RateLimiter::ModelInstance* allocated_instance = nullptr;
+    state.store(SchedState::RETRIEVING_REQUEST);
+
+    while (state != READY_FOR_NEW_REQUESTS && !scheduler_thread_exit_.load()) {
```
deadeyegoodwin, line 266: parens

deadeyegoodwin (changes requested):
You are making a lot of changes to the dynamic batcher... need to be very careful to make sure existing logic is preserved...

src/core/dynamic_batch_scheduler.cc
```diff
@@ -251,169 +246,217 @@ DynamicBatchScheduler::SchedulerThread(
...
+        case SchedState::RETRIEVING_REQUEST: {
+          uint64_t wait_microseconds = 0;
+
+          // Hold the lock for as short a time as possible.
+          {
+            std::unique_lock<std::mutex> lock(mu_);
+            if (delay_cnt > 0) {
+              // Debugging/testing... wait until queue contains 'delay_cnt'
+              // items...
+              wait_microseconds = 10 * 1000;
```
CoderHam, line 277: Do we want to hardcode this wait?

CoderHam (changes requested):

src/core/dynamic_batch_scheduler.cc
tanmayv25: I have just ported this logic from our current dynamic batcher implementation. I have tried to keep the control flow for single instance as similar as possible.

tanmayv25:
@deadeyegoodwin  @CoderHam  I understand the risk and have tried my best to keep this implementation of dynamic batch scheduler functionally identical to our current one for a single instance case. There are a couple of locks and atomic::load/store added in the pipeline to correctly interact with the rate limiter.
We do need to do some tuning here to identify bottlenecks and obtain similar or hopefully better perf.

deadeyegoodwin (approved):

PR #2896, opened 2021-05-20
Move WarmUp inside TritonModelInstance class

tanmayv25:
@GuanLuo With this change and new TensorRT Backend, I don't think we need the special warmup handling as we do now within backend. Each instance will WarmUp itself with their respective TritonBackendThread. The only difference is in eager_batching mode. The current implementation would warmup each instance twice because they are duplicated in the queue. I think this is not required, Let me know if you think otherwise.


src/core/dynamic_batch_scheduler.cc
```diff
@@ -197,9 +197,6 @@ DynamicBatchScheduler::SchedulerThread(
                  << model_->Name() << " at default nice...";
 #endif
 
-  // TODO: Move Initialize and WarmUp to triton_model and
```
tanmayv25, line 200: @deadeyegoodwin  I think we would need to leave the `Initialize` and `WarmUp` call as is instead of earlier proposed idea of moving into construction logic. With @GuanLuo changes we would want to call these functions on the thread which is going to enter the backend exec. What do you think?

deadeyegoodwin: ok

PR #2897, opened 2021-05-20
Add unittests for custom Python environments

Tabrizian:


qa/L0_backend_python/test.sh
```diff
@@ -405,14 +447,82 @@ $page_size."
...
+wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | \
+	gpg --dearmor - |  \
+	tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null && \
+	apt-add-repository 'deb https://apt.kitware.com/ubuntu/ focal main' && \
+	apt-get update && \
+	apt-get install -y --no-install-recommends \
+	cmake-data=3.18.4-0kitware1ubuntu20.04.1 cmake=3.18.4-0kitware1ubuntu20.04.1
+install_conda
+
+# Create a model with python 3.9 version
```
deadeyegoodwin, line 468: Can you install a different version of some package (numpy) in each of these environments and then in the print statement output both the python and numpy version so that the grep makes sure that the python/numpy combination is what is expected.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

GuanLuo (approved):

PR #2913, opened 2021-05-25
Rename master -> main

Tabrizian:


CoderHam (changes requested):
Nice catch Iman, could you also make these changes in the rest of the docs?
Also I believe master/main would be clearer but I think that defeats the purpose of the change of the name. 
cc @deadeyegoodwin

CoderHam (approved):

PR #2914, opened 2021-05-25
Add OpenVINO backend to windows build

deadeyegoodwin:


build.py
```diff
@@ -64,7 +65,13 @@
...
 #
 TRITON_VERSION_MAP = {
-    '2.11.0dev': ('21.06dev', '21.04', '1.7.1', '2021.2.200', '2021.2.200')
+    '2.11.0dev':
+      ('21.06dev',   # triton container
+       '21.04',      # upstream container
+       '1.7.1',      # ORT
+       '2021.2.200', # ORT OpenVINO
+       '2021.2.200', # Standalone OpenVINO (non-windows)
+       '2021.3.0')   # Standalone OpenVINO (windows)
```
tanmayv25, line 74: So, it's ok to have separate OV version than ORT OpenVINO on windows? We can update the comment above that describes the problem with dlopen and how it doesn't exists in windows(or with dlls).

Still, I wonder whether shouldn't we use the identical OV version on windows as well to maintain consistency.

build.py
```diff
@@ -64,7 +65,13 @@
...
 #
 TRITON_VERSION_MAP = {
-    '2.11.0dev': ('21.06dev', '21.04', '1.7.1', '2021.2.200', '2021.2.200')
+    '2.11.0dev':
+      ('21.06dev',   # triton container
+       '21.04',      # upstream container
+       '1.7.1',      # ORT
+       '2021.2.200', # ORT OpenVINO
+       '2021.2.200', # Standalone OpenVINO (non-windows)
+       '2021.2')     # Standalone OpenVINO (windows)
```
CoderHam, line 74: How is 2021.2 different from 2021.2.200, does something stop us from using 2021.2.200 on windows?

deadeyegoodwin: 2021.2 is the actual branch that we build from and I will be moving linux to build from that branch as well in a separate change. Right now we are using OV from a prebuilt package that has version 2021.2.200... not sure where the 200 comes from.

deadeyegoodwin: Eventually we need to get the 3 openvino entries down to 1 so that we are using the same OV version for everything

build.py
deadeyegoodwin: Once I get windows and linux building OV then I'll work on getting all OV users to use the same version/build of OV and these 3 versions will reduce to 1 and that will make sure that we use the same version of OV for all uses.

build.py
tanmayv25: 200 is the build number of the release.

PR #2916, opened 2021-05-25
ArmNN TFLite backend build support

jishminor:
This PR adds support to the triton build infrastructure to build-in the new ArmNN Tensorflow Lite backend, against an Ubuntu 20.04 arm64 base image. 

In these changes we:
* Add some extra argument options to build.py to support tflite with ArmNN delegate support
* Build CMake 3.19.1 from source for all docker builds
* Remove the `defensive wiring for CUDA Compat lib` if we are building for non-cuda devices
* Add the build Flag TRITON_ENABLE_MALI_GPU. This flag is to enable non-cuda gpu features when TRITON_ENABLE_GPU is not set in a build. This allows the TFLite backend model configurations to specify KIND_GPU for instance types without throwing errors for builds which disabled TRITON_ENABLE_GPU
* Modifies nvidia_entrypoint.sh to support non-cuda Arm devices

This PR depends on https://github.com/triton-inference-server/backend/pull/25 being merged into main.


build.py
```diff
@@ -696,13 +714,15 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
 COPY --chown=1000:1000 --from=tritonserver_build /tmp/tritonbuild/install/repoagents repoagents
 '''
 
+    if target_platform() != 'ubuntu/arm64':
```
jishminor, line 717: I will note here the docker build breaks on a non-cuda x86 machine (where you would pass the target platform "ubuntu") with error: /bin/bash: ${{_CUDA_COMPAT_PATH}}/lib.real: bad substitution. I guess the assumption is that on x86 platforms you must have the nvidia container runtime.

build.py
```diff
@@ -71,7 +71,7 @@
 CORE_BACKENDS = ['tensorrt', 'ensemble']
 NONCORE_BACKENDS = [
     'tensorflow1', 'tensorflow2', 'onnxruntime', 'python', 'dali', 'pytorch',
-    'openvino', 'fil'
+    'openvino', 'fil', 'tflite'
```
deadeyegoodwin, line 74: Instead if using 'tflite' can we instead refer to the backend as 'armnn' or some similar arm-specific term. Similarly the backend should be called 'armnn_backend" or similar. We don't want to represent as tflite as we are not claiming general tflite support.

deadeyegoodwin (changes requested):

build.py
jishminor: Well this backend is for TFLite, and enables the use of the ArmNN and XNNPACK delegate to accelerate any operations which they each implement respectively. If the delegates are not used the default TFLite runtime is used. See: https://gitlab.com/arm-research/smarter/tflite_backend

build.py
```diff
@@ -65,21 +65,22 @@
 # incorrectly load the other version of the openvino libraries.
 #
 TRITON_VERSION_MAP = {
-    '2.12.0dev':
-      ('21.07dev',   # triton container
-       '21.06',      # upstream container
-       '1.8.0',      # ORT
-       '2021.2.200', # ORT OpenVINO
-       '2021.2')     # Standalone OpenVINO
+    "2.12.0dev": (
```
deadeyegoodwin, line 68: Rebase to main...

build.py
```diff
@@ -65,21 +65,22 @@
...
+        "21.06",  # upstream container
+        "1.8.0",  # ORT
+        "2021.2.200",  # ORT OpenVINO
+        "2021.2",
+    )  # Standalone OpenVINO
 }
 
-EXAMPLE_BACKENDS = ['identity', 'square', 'repeat']
-CORE_BACKENDS = ['tensorrt', 'ensemble']
+EXAMPLE_BACKENDS = ["identity", "square", "repeat"]
```
deadeyegoodwin, line 77: Keep these as single quotes

build.py
```diff
@@ -65,21 +65,22 @@
...
+CORE_BACKENDS = ["tensorrt", "ensemble"]
 NONCORE_BACKENDS = [
     'tensorflow1', 'tensorflow2', 'onnxruntime', 'python', 'dali', 'pytorch',
-    'openvino', 'fil', 'fastertransformer'
+    'openvino', 'fil', 'fastertransformer', 'tflite'
```
deadeyegoodwin, line 81: I thought we had discussed that this should be "armnn" as we don't want to advertise and support a general tflite backend as part of this effort.

build.py
```diff
@@ -65,21 +65,22 @@
...
 ]
-EXAMPLE_REPOAGENTS = ['checksum']
+EXAMPLE_REPOAGENTS = ["checksum"]
```
deadeyegoodwin, line 83: don't change quote style, here or below...

build.py
```diff
@@ -108,338 +109,400 @@ def target_platform():
...
+        fail_if(p.returncode != 0, 'git checkout of branch "tritonbuildref" failed')
 
     else:
         log_verbose('git clone of repo "{}" at tag "{}"'.format(repo, tag))
-        p = subprocess.Popen([
-            'git', 'clone', '--recursive', '--single-branch', '--depth=1', '-b',
-            tag, '{}/{}.git'.format(FLAGS.github_organization, repo), subdir
-        ],
-                             cwd=cwd)
+        p = subprocess.Popen(
```
deadeyegoodwin, line 178: Are you formatting as described here? https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#coding-convention

deadeyegoodwin (changes requested):
There are too many formatting changes to be sure what the real changes are. See my review comment for a link. Or just don't format at all so the actual changes are clear.

jishminor:
In a previous commit to this branch `black` formatting was applied to `build.py` on accident. Changes now reflect the correctly formatted `build.py`. Naming of backend still under review.

build.py
```diff
@@ -529,7 +544,6 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
 
     df += '''
 FROM ${BASE_IMAGE}
-
```
deadeyegoodwin, line 532: Don't remove these blank lines.

build.py
```diff
@@ -1396,7 +1420,13 @@ def container_build(images, backends, repoagents, endpoints):
         repo_install_dir = os.path.join(FLAGS.build_dir, be, 'install')
 
         mkdir(FLAGS.build_dir)
-        gitclone(FLAGS.build_dir, backend_repo(be), backends[be], be)
+        # If tflite backend, source from external repo for git clone
+        if be == 'tflite':
+            gitclone(FLAGS.build_dir, backend_repo(be), backends[be], be,
+                     'https://gitlab.com/arm-research/smarter/')
```
deadeyegoodwin, line 1426: It would be better to set github-organization default to None in the argparse, and then manually set a default value for it based on platform if the user doesn't set it explicitly. Then you don't need any of these org changes.

deadeyegoodwin (changes requested):

build.py
jishminor: @deadeyegoodwin  Changes are made as per your request, however tying the github org to the platform and not the backend makes it such that the tflite backend can ONLY be built into the `ubuntu/arm64` build, although it should build fine into the jetpack build as well.

deadeyegoodwin: That's a good point. Ideally we'd have an option to set github org for each backend independently... but not worth doing that at this point. I would say you should just revert fdc20a0 to go back to what you had. That means that someone can't build the armnn backend from their fork without editing build.py but I think that is ok for now.

jishminor: Agreed. `fdc20a0` reverted.

deadeyegoodwin (dismissed):

build.py
```diff
@@ -584,14 +601,17 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
...
 
-# Server build requires recent version of CMake (FetchContent required)
-RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | \
-      gpg --dearmor - |  \
-      tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null && \
-    apt-add-repository 'deb https://apt.kitware.com/ubuntu/ focal main' && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends \
-      cmake-data=3.18.4-0kitware1ubuntu20.04.1 cmake=3.18.4-0kitware1ubuntu20.04.1
+# Server build requires recent version of CMake (FetchContent required), build from source
```
deadeyegoodwin, line 604: I think you are building from source for availability on ARM? But more recent packages are available for ARM, so I will try using 3.21.1 when I run CI

jishminor: Ah ok. I wasn't aware there was an arm build available for later cmake versions, if that's the case, then avoiding the source build is preferred.

build.py
```diff
@@ -796,6 +816,14 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu):
     pip3 install --upgrade wheel setuptools && \
     pip3 install --upgrade numpy && \
     rm -rf /var/lib/apt/lists/*
+'''
```
deadeyegoodwin, line 819: Remove this change... it is already 20 lines above.

deadeyegoodwin (changes requested):

build.py
deadeyegoodwin: Revert this back to the original code and change the cmake version to 3.21.1

nvidia_entrypoint.sh
```diff
@@ -69,7 +69,14 @@ else
   fi
 fi
 
+avx=1
```
deadeyegoodwin, line 72: Can you revert all changes in this file. The avx and isarm dereferences need '$' but we will fix this file ourselves once we officially support SBSA.

deadeyegoodwin (changes requested):
Linux CI looks good once you make the 2 requested changes.

deadeyegoodwin (approved):

PR #2923, opened 2021-05-26
Fix some synchronization points and prevent extraneous notifications to scheduler thread

tanmayv25:
This fixes the perf_issues for low request concurrencies and multiple instances(instance_count=2).
For request concurrency 10 and instance_count=2, these changes improve perf from 3400 infer/sec to  3600 infer/sec.
The old DB scheduler produces ~4000 infer/sec for the same model and settings.


tanmayv25:
By reducing the number of notifications, I see the perf jumps to 3800 infer/sec.  This is with `dynamic_batching_enabled_ = false`
Scheduler thread keeps running as long as queue is not empty without waiting... The notification is also issue only when the queue is empty and scheduler thread might be waiting.

PR #2924, opened 2021-05-26
Change usage of NVML to DCGM

jbkyang-nvi:
Changes on top of https://github.com/triton-inference-server/server/compare/hemantj-dcgm
 to change monitoring with NVML to DCGM.


Dockerfile.QA
```diff
@@ -283,6 +283,18 @@ ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
 RUN GO111MODULE=off go get github.com/golang/protobuf/protoc-gen-go && \
     GO111MODULE=off go get google.golang.org/grpc
 
+# DCGM needed for GPU Metrics
```
CoderHam, line 286: Can you double check if we need to install this in the QA container if we do it in our buildbase container?

build.py
```diff
@@ -552,6 +563,15 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
...
 '''
+    if target_platform() != 'windows':
+        df += '''
+# Install DCGM - patch
+# Note: The current version is missing header files so we applied a patch to fix the build. This patch should be removed
+# after DCGM team fixes this issue in their new release.
+RUN cp /workspace/tools/dcgm_api_export.h /usr/include/
+'''
+
+
```
CoderHam, line 574: Extra line

build.py
```diff
@@ -522,6 +522,17 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
...
+# Note: The current version is missing header files so we applied a patch to fix the build. This patch should be removed
+# after DCGM team fixes this issue in their new release.
+RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin \
+    && mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600 \
+    && apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub \
+    && add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
+RUN apt-get update \
+    && apt-get install -y datacenter-gpu-manager
+  
+
```
CoderHam, line 535: Extra line

CoderHam (changes requested):

build.py
```diff
@@ -552,6 +563,15 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
 ENV TRITON_SERVER_VERSION ${TRITON_VERSION}
 ENV NVIDIA_TRITON_SERVER_VERSION ${TRITON_CONTAINER_VERSION}
 '''
+    if target_platform() != 'windows':
+        df += '''
+# Install DCGM - patch
+# Note: The current version is missing header files so we applied a patch to fix the build. This patch should be removed
+# after DCGM team fixes this issue in their new release.
+RUN cp /workspace/tools/dcgm_api_export.h /usr/include/
```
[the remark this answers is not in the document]
jbkyang-nvi: I think we only get /workspace set up a few lines above this one in line 549

Dockerfile.QA
jbkyang-nvi: yep I checked. If I don't have it installed, there's tritonserver: error while loading shared libraries: libdcgm.so.2: cannot open shared object file: No such file or directory errors.

deadeyegoodwin: This final stage of the QA container build is based on BASE_IMAGE which should be the production tritonserver image and that should have everything needed for tritonserver to run. So you shouldn't need to install libdcgm. The other installs you see here are installing things needed for the tests, not for tritonserver itself.

tools/dcgm_api_export.h
```diff
@@ -0,0 +1,14 @@
+#ifndef DCGM_DCGM_API_EXPORT_H
```
deadeyegoodwin, line 1: This is an entire file, not a patch. It is less safe to overwrite the entire file than to just apply a patch. The patch will fail if the file changes and lead to a failed build which will allow us to notice that something changed. Also this should live in "build/libgcgm/dcgm_api_export.patch". See onnxruntime_backend for example of applying the patch.

build.py
```diff
@@ -522,6 +522,17 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
     pip3 install --upgrade wheel setuptools docker && \
     pip3 install grpcio-tools grpcio-channelz
 
+# Install DCGM
```
deadeyegoodwin, line 525: This install is needed in buildbase container so that we can build tritonserver. You should move it down with the other related code so that it is all together. The install is also needed in create_dockerfile_linux() as that is the dockerfile that builds the actual production container. Move it into a separate function since it needs to be used in 2 places.

deadeyegoodwin (changes requested):

Dockerfile.QA
jbkyang-nvi: @deadeyegoodwin would I need to do anything to make this happen? e.g. apply patches, or would all the installs from the `build` container automatically install everything

tools/dcgm_api_export.h
jbkyang-nvi: @deadeyegoodwin we're just adding a new file (it's not actually a patch). Libdcgm does not currently have a release with this change. Should we just wait for them?

Dockerfile.QA
deadeyegoodwin: The base container is not the build container, see my other comments in build.py. We must make sure that the QA container is just the production container (i.e. the base container as in master-py3-base) plus some installs needed for testing. Otherwise we risk testing tritonserver in a way that differs from how it is actually delivered.

build.py
jbkyang-nvi: Since we added in the buildbase would that information not be included in the linux version?

deadeyegoodwin: No, look at create_dockerfile_linux and you can see that we start from the min container and just copy some things over from the build container. We do that because we don't want the production container to contain all the build dependencies and artifacts. That also means that we need to explicitly install any dependencies that are needed to run tritonserver, as you can see us doing in create_dockerfile_linux(). In general the production tritonserver container is not based on the build containers... it just copies out the built libraries and executables from the build container.

build.py
```diff
@@ -545,12 +563,19 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap, backends):
 RUN rm -fr *
 COPY . .
 ENTRYPOINT []
+'''
+        df += install_dcgm_libraries()
+        df += fail_if_dcgm_api_export_exists()
+        df += '''
+WORKDIR /workspace
```
CoderHam, line 570: I believe this line is not needed. The WORKDIR is already set to /workspace earlier

build.py
```diff
@@ -667,6 +692,12 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
 # Top-level include/core not copied so --chown does not set it correctly,
 # so explicit set on all of include
 RUN chown -R triton-server:triton-server include
+'''
+
```
CoderHam, line 696: Extra empty line

tools/dcgm_api_export.h
CoderHam: The addition of this file can be the patch. I agree with @deadeyegoodwin that it should be in `build/libgcgm/dcgm_api_export.patch` and that you should apply this patch. Refer to how we do this in our ORT backend.

Dockerfile.QA
```diff
@@ -284,7 +284,7 @@ ENV GOPATH /root/go
 ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
 RUN GO111MODULE=off go get github.com/golang/protobuf/protoc-gen-go && \
     GO111MODULE=off go get google.golang.org/grpc
-
+
```
CoderHam, line 287: Remove extra spaces added

CoderHam (changes requested):

build.py
```diff
@@ -462,6 +462,24 @@ def dali_cmake_args():
...
+&& add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
+RUN apt-get update \
+&& apt-get install -y datacenter-gpu-manager
+'''
+
+def fail_if_dcgm_api_export_exists():
+    return '''
+# Note: The current version is missing header files so we applied a dcgm_api_export.h to fix the build. This fix should be removed
+# after DCGM team fixes this issue in their new release.
+RUN if test -s /usr/include/dcgm_api_export.h;then echo "/usr/include/dcgm_api_export.h exists! Remove copy from build.py" | exit 1;fi
```
deadeyegoodwin, line 481: Replace '|' with ';'? Why are you piping the echo to "exit 1". Did you test this to make sure the build fails as expected.

build.py
```diff
@@ -667,6 +692,12 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
 # Top-level include/core not copied so --chown does not set it correctly,
 # so explicit set on all of include
 RUN chown -R triton-server:triton-server include
+'''
+
+    df += install_dcgm_libraries()
+    df += fail_if_dcgm_api_export_exists()
```
deadeyegoodwin, line 698: Don't need to call 'fail...' here. We just want the libraries installed... not even going to use the includes.

build.py
```diff
@@ -667,6 +692,12 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+    df += '''
+COPY --from=tritonserver_build /workspace/tools/dcgm_api_export.h /usr/include/
```
deadeyegoodwin, line 700: Remove this, don't care about include in the production container.

tools/dcgm_api_export.h
deadeyegoodwin: We talked about this offline. With the above build.py changes copying a complete file is ok. But it still needs to live in build/libdcgm

deadeyegoodwin (changes requested):

build.py
jbkyang-nvi: I did. Figured out a way to make a patch so removing all of this anyway

Dockerfile.QA
```diff
@@ -284,7 +284,6 @@ ENV GOPATH /root/go
 ENV PATH $PATH:/usr/local/go/bin:$GOPATH/bin
 RUN GO111MODULE=off go get github.com/golang/protobuf/protoc-gen-go && \
     GO111MODULE=off go get google.golang.org/grpc
```
deadeyegoodwin, line 286: You have no changes in this file, remove it from the PR

build.py
```diff
@@ -668,6 +684,7 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
 # so explicit set on all of include
 RUN chown -R triton-server:triton-server include
 '''
+    df += install_dcgm_libraries()
```
deadeyegoodwin, line 687: This should be up at line 659 where the other common dependencies are installed.

deadeyegoodwin (changes requested):

src/core/metrics.cc
```diff
@@ -163,205 +182,307 @@ Metrics::EnableGPUMetrics()
...
+      unsigned short fields[field_count] = {
+          DCGM_FI_DEV_POWER_MGMT_LIMIT,          // power limit, watts
+          DCGM_FI_DEV_POWER_USAGE,               // power usage, watts
+          DCGM_FI_DEV_TOTAL_ENERGY_CONSUMPTION,  // Total energy consumption, mJ
+          util_flag,                             // util ratio, 1 = 1%
+          DCGM_FI_DEV_FB_USED,                   // Frame buffer used, MB
+          DCGM_FI_DEV_FB_TOTAL,                  // Frame buffer used, MB
+      };
+
+      char fieldName[16] = "myfields";
```
CoderHam, line 367: Why is this 16 again?

CoderHam: Also use "field_group"
In general keep possession words out of variable names unless they signify meaning,

src/core/metrics.cc
```diff
@@ -163,205 +182,307 @@ Metrics::EnableGPUMetrics()
...
+      dcgmFieldGrp_t fieldGroupId;
+      dcgmReturn_t dcgmerr = dcgmFieldGroupCreate(
+          handle, field_count, &fields[0], fieldName, &fieldGroupId);
+      if (dcgmerr != DCGM_ST_OK) {
+        LOG_WARNING << "error, cannot cannot make field group. Err:"
+                    << errorString(dcgmerr);
+      }
+      dcgmerr = dcgmWatchFields(
+          handle, groupId, fieldGroupId, 2000000 /*update period, usec*/,
```
CoderHam, line 376: Is watching every 2 seconds the suggested amount?

src/core/metrics.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright 2018-2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 1: Here and in other places keep (c) after Copyright. Also the convention is:
"Copyright (c) 2020-21 ..."

CoderHam (changes requested):

src/servers/CMakeLists.txt
```diff
@@ -528,7 +528,6 @@ if(${TRITON_ENABLE_GPU})
   find_library(CNMEM_LIBRARY NAMES cnmem PATHS ${CNMEM_PATH}/lib)
   target_link_libraries(
     tritonserver
-    PUBLIC ${CUDA_nvml_LIBRARY}
```
CoderHam, line 531: Also need to make these changes for simple and memory_alloc

CoderHam (changes requested):

src/core/metrics.cc
```diff
@@ -163,205 +182,307 @@ Metrics::EnableGPUMetrics()
...
 
-    std::string uuid;
-    char uuid_str[NVML_DEVICE_UUID_BUFFER_SIZE + 1];
-    nvmlerr = nvmlDeviceGetUUID(gpu, uuid_str, NVML_DEVICE_UUID_BUFFER_SIZE);
-    if (nvmlerr == NVML_SUCCESS) {
-      uuid = uuid_str;
-    } else {
-      uuid = "unknown";
+  // create a gpu group
+  char groupName[16] = "mygroup";
```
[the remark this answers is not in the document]
jbkyang-nvi: Yeah, we do sadly

src/core/metrics.cc
```diff
@@ -163,205 +182,307 @@ Metrics::EnableGPUMetrics()
...
+    char pciFieldName[16] = "pciFields";
+    dcgmerr = dcgmFieldGroupCreate(
+        dcgm_handle_, field_count, &pciField[0], pciFieldName, &pciFieldId);
+    if (dcgmerr != DCGM_ST_OK) {
+      LOG_WARNING
+          << "error, cannot cannot make field group for PCI fields. Err:"
+          << errorString(dcgmerr);
+    }
+    // we only watch once here
+    dcgmerr = dcgmWatchFields(
```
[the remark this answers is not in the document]
jbkyang-nvi: It only updates (since we're in manual mode) when `dcgmUpdateAllFields` is called. We need do specify this at least

src/core/metrics.cc
jbkyang-nvi: Since we watch every 2 seconds, `std::this_thread::sleep_for(std::chrono::milliseconds(2000));` I think this will be the right amount

build.py
```diff
@@ -964,15 +982,16 @@ def container_build(images, backends, repoagents, endpoints):
...
-            create_dockerfile_linux(FLAGS.build_dir, 'Dockerfile', dockerfileargmap,
-                              backends, repoagents, endpoints)
+            create_dockerfile_linux(FLAGS.build_dir, 'Dockerfile',
+                                    dockerfileargmap, backends, repoagents,
+                                    endpoints)
         p = subprocess.Popen([
-                'docker', 'build', '-f',
-                os.path.join(FLAGS.build_dir, 'Dockerfile')
-            ] + ['-t', 'tritonserver', '.'])
+            'docker', 'build', '-f',
```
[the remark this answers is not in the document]
jbkyang-nvi: I did run the formatter and this is what it gave me... @CoderHam should I just use the previous format?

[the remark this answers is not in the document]
CoderHam: If this is what the formatter gave, it should be fine

CoderHam (approved):

build.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright (c) 2020-2021, NVIDIA CORPORATION. All rights reserved.
+# Copyright 2018-2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 2: Still needs to be fixed to be "2018-21". Check other places as well

CoderHam (changes requested):

build.py
deadeyegoodwin: Why? Please don't add older copyright years. This file did not exist in 2018. Also, why are you removing the (c). We should just keep the existing copyright heading that we have always had but add "& AFFILICATES" and add "2021" if you are making a change this year.

build.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2018-2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 2: This needs to be 2020-21. Why are you added a comma?

src/servers/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2021, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 1: Similarly This needs to be 2019-21. No comma.

CoderHam (changes requested):

build.py
deadeyegoodwin: Actually the comma is required. Here are the regex that we check for (not including AFFILICATES change):
COPYRIGHT_YEAR_RE0 = 'Copyright \\(c\\) (20[0-9][0-9]), NVIDIA CORPORATION. All rights reserved.'
COPYRIGHT_YEAR_RE1 = 'Copyright \\(c\\) (20[0-9][0-9])-(20[0-9][0-9]), NVIDIA CORPORATION. All rights reserved.'

In main branch the comma is there so curious why it is not showing up here in this diff?

CoderHam (dismissed):

src/core/metrics.cc
```diff
@@ -129,10 +132,26 @@ Metrics::HashLabels(const std::map<std::string, std::string>& labels)
...
+    dcgmGroupDestroy(dcgm_handle_, groupId_);
+    // Stop and shutdown DCGM
+    dcgmReturn_t derr;
+    if (standalone_) {
+      derr = dcgmDisconnect(dcgm_handle_);
+    } else {
+      derr = dcgmStopEmbedded(dcgm_handle_);
+    }
+    if (derr != DCGM_ST_OK) {
+      LOG_WARNING << "error, unable to stop DCGM: " << errorString(derr);
```
deadeyegoodwin, line 148: Here and elsewhere. Don't say  "error" when it is a warning. Experience shows that we will bet issues and bugs filed saying an error happened. just put "unable to stop... ".

src/core/metrics.cc
```diff
@@ -163,205 +182,270 @@ Metrics::EnableGPUMetrics()
...
+    char pcibusid_str[64];
+    cudaError_t cudaerr =
+        cudaDeviceGetPCIBusId(pcibusid_str, sizeof(pcibusid_str) - 1, i);
+    if (cudaerr == cudaSuccess) {
+      pciBusId.append(pcibusid_str);
+      pciBusIds.emplace(pciBusId);
     } else {
-      LOG_INFO << "Collecting metrics for GPU " << didx;
+      LOG_WARNING << "error, CUDA unable to get PCI bus information for GPU:"
+                  << i;
```
deadeyegoodwin, line 250: "GPU metrics will not be available for this device" follow same pattern as similar messages

src/core/metrics.cc
```diff
@@ -163,205 +182,270 @@ Metrics::EnableGPUMetrics()
...
 
-    nvmlDevice_t gpu;
-    nvmlReturn_t nvmlerr = nvmlDeviceGetHandleByPciBusId_v2(pcibusid_str, &gpu);
-    if (nvmlerr != NVML_SUCCESS) {
-      LOG_WARNING << "failed to get device from Bus ID "
-                  << ", GPU metrics will not be available for this device: "
-                  << nvmlErrorString(nvmlerr);
-      continue;
-    }
+  unsigned int all_gpu_ids[DCGM_MAX_NUM_DEVICES];
```
deadeyegoodwin, line 228: how about dcgm_gpu_ids and dcgm_gpu_count

src/core/metrics.cc
```diff
@@ -163,205 +182,270 @@ Metrics::EnableGPUMetrics()
...
-    std::string uuid;
-    char uuid_str[NVML_DEVICE_UUID_BUFFER_SIZE + 1];
-    nvmlerr = nvmlDeviceGetUUID(gpu, uuid_str, NVML_DEVICE_UUID_BUFFER_SIZE);
-    if (nvmlerr == NVML_SUCCESS) {
-      uuid = uuid_str;
+  // Get DCGM metrics for each GPU. Some devices may have problems using DCGM
+  // API and thus these devices needs to be ignored.
+  std::vector<uint32_t> available_gpu_ids;
+  dcgmDeviceAttributes_t gpu_attributes[DCGM_MAX_NUM_DEVICES];
+  for (int i = 0; i < gpu_count; i++) {
```
deadeyegoodwin, line 258: You are iterating through the DCGM devices and reportin DCGM device IDs. But that is not what the user understands. They understand CUDA device IDs. So this code seems a little backwards. Think about creating a map from pcibusid to DCGM device ID. Then iterate through the cuda devices and for each look up the DCGM id and from that the DCGM info. If you can't find DCGM id for a pcibusid then you can report the CUDA device id as not having metrics available.

deadeyegoodwin (changes requested):

src/core/metrics.cc
jbkyang-nvi: Should we use the CUDA id for all devices then? Right now, we are reporting all device ids from the DCGM standpoint.

deadeyegoodwin: Yes. The user does not known anything about dcgm numbering of the GPUs. If they understand anything it will be the CUDA number of the GPUs. For example, if they run on a system with multiple GPUs but only one is visible to Triton, then they expect that GPU to be referred to as GPU 0, no matter what the DCGM id of that GPU is.

deadeyegoodwin (changes requested):

src/core/metrics.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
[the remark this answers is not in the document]
jbkyang-nvi: @deadeyegoodwin  Yep the formatter did this...

src/core/metrics.cc
```diff
@@ -163,205 +183,286 @@ Metrics::EnableGPUMetrics()
...
+          5.0 /*maxKeepAge, sec*/, 5 /*maxKeepSamples*/);
+      if (dcgmerr != DCGM_ST_OK) {
+        LOG_WARNING << "Cannot start watching fields: " << errorString(dcgmerr);
+      } else {
+        while (!dcgm_thread_exit_.load()) {
+          std::this_thread::sleep_for(std::chrono::milliseconds(2000));
+          dcgmUpdateAllFields(handle, 1 /* wait for update*/);
+          for (int didx = 0; didx < available_gpu_count; ++didx) {
+            uint32_t dcgm_id = available_gpu_ids[didx];
+            if (dcgm_ids_to_cuda_ids.count(dcgm_id) <= 0) {
```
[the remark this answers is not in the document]
jbkyang-nvi: the `dcgm_ids_to_cuda_ids.at()` will segfault if `dcgm_id` doesn't exist. Are we ok with this?

src/core/metrics.cc
```diff
@@ -163,205 +183,290 @@ Metrics::EnableGPUMetrics()
...
-    if (nvmlerr == NVML_SUCCESS) {
-      uuid = uuid_str;
+
+  // Get CUDA-visible PCI Bus Ids and get DCGM metrics for each CUDA-visible GPU
+  std::map<uint32_t, uint32_t> cuda_ids_to_dcgm_ids;
+  std::vector<uint32_t> available_cuda_gpu_ids;
+  int cuda_gpu_count;
+  cudaError_t cudaerr = cudaGetDeviceCount(&cuda_gpu_count);
+  if (cudaerr != cudaSuccess) {
+    LOG_WARNING << "Cannot get cuda device count";
```
deadeyegoodwin, line 275: "... GPU metrics will not be available".
Also, at this point cuda_gpu_count is undefined so you can't continue. You should "return false" here.

src/core/metrics.cc
```diff
@@ -163,205 +183,290 @@ Metrics::EnableGPUMetrics()
...
+                    (!DCGM_INT64_IS_BLANK(memory_total))) {
+                  mem_fail_cnt[didx] = 0;
+                } else {
+                  memory_total = 0;
+                  memory_used = 0;
+                  mem_fail_cnt[didx]++;
+                  LOG_WARNING << "Unable to get memory usage for GPU "
+                              << cuda_id << ": " << errorString(dcgmerr);
+                }
+                gpu_memory_total_[didx]->Set(memory_total * 1e6);  // bytes
```
deadeyegoodwin, line 466: Don't use "1e6" because that is a float value which then does math as floats and converts back to int. Use "1000 * 1000". Or likely you need to use "1024 * 1024" because the field value is in MiB... you need to check that.

deadeyegoodwin (changes requested):

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

PR #2933, opened 2021-05-28
Fix HTTP to pass proper content length in the case of compression

GuanLuo:


deadeyegoodwin (approved):

src/servers/http_server.cc
```diff
@@ -2203,6 +2181,41 @@ HTTPAPIServer::HandleInfer(
         // Do nothing
         break;
     }
+  }
+
+  // Get the header length
+  size_t header_length;
+  if (err == nullptr) {
+    // Set to large value in case there is no Content-Length to compare with
```
CoderHam, line 2189: "Set to large value in case there is no Content-Length in the request header"?

CoderHam (changes requested):

CoderHam (approved):

Issue #2853, opened 2021-05-12
Triton does not load the latest available model in explicit mode.


nsiddharth:
**Description**
using model control mode as explicit, it is expected that Triton will load the latest available model when a new model version is added and load_model is called. However, this does not happen and load_model returns only the already loaded model. This can be tested by adding a new version directory for the simple model provided in the examples. Note that a new version directory is added after Triton has started up. However, this works when unload_model is called followed by load model
**Triton Information**
What version of Triton are you using?
2.8.0
Are you using the Triton container or did you build it yourself?
Using the container
**To Reproduce**
Steps to reproduce the behavior.
Use a simple script like so to test this behavior:
```
import tritonclient.grpc as grpcclient
from tritonclient.utils import InferenceServerException

MODEL_NAME = "simple"

URL = "localhost:8001"
triton_client = grpcclient.InferenceServerClient(url=URL, verbose=True)

triton_client.is_server_live()
print(triton_client.is_model_ready(MODEL_NAME))
triton_client.get_model_repository_index().models

"""
model loading and unloading 
This should pickup the latest available model
without having to call:
triton_client.unload_model(MODEL_NAME)
"""
triton_client.load_model(MODEL_NAME)
"""This should show that the latest model is available"""
triton_client.get_model_repository_index().models

```
Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).
The model used to test this was the simple model provided in the examples
**Expected behavior**
when a new version is added to the model repo like so:
`aws s3 cp --recursive 2 s3://<Model_repo>/simple/2`
and load_model is called, the latest model should be loaded and made available. This does not happen. 
However, this takes effect if unload_model is called before calling load_model again. We do not want this behavior in production where you need to unload the model causing downtime before you are able to load a new one.

dzier:
Your expectation for explicit mode is incorrect. What you are describing is the POLL mode. In EXPLICIT mode, you need to explicitly unload the model before you can load the model again with the new changes. In your example code, it appears you are commenting out the `triton_client.unload_model(MODEL_NAME)`. 

Additional information can be found here: https://github.com/triton-inference-server/server/blob/master/docs/model_management.md#modifying-the-model-repository

nsiddharth:
[...]

PR #2936, opened 2021-05-29
convert Millis to Nanoseconds for correct comparison.

nsiddharth:
Fixes [issue #2853](https://github.com/triton-inference-server/server/issues/2853) which was prematurely closed

### DIFF
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -1470,7 +1470,7 @@ S3FileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   // If request succeeds, copy over the modification time
   auto head_object_outcome = client_.HeadObject(head_request);
   if (head_object_outcome.IsSuccess()) {
-    *mtime_ns = head_object_outcome.GetResult().GetLastModified().Millis();
+    *mtime_ns = head_object_outcome.GetResult().GetLastModified().Millis()*1000000;
   } else {
     return Status(
         Status::Code::INTERNAL,


CoderHam:
Thank you for your contribution @nsiddharth. To clarify this is unrelated to the explicit model control mode but still a fix for the polling model control mode. 

I stand by my earlier belief that explicit model control works as expected (loads latest available model for models on reloading) with S3. That being said I can look into it again to verify and re-open the issue you linked as needed.

nsiddharth:
I see, In that case I am not quite sure why it enters that code path. It seemed like that it determines when a model is modifed using `getModifiedTime()`. I have posted some server logs with added print statements in the other ticket if it helps.

CoderHam:
@nsiddharth while working on expanding our testing and looking into this issue I zeroed in on the true case. Since the Origin was inside [GetModifiedTime()](https://github.com/triton-inference-server/server/blob/main/src/core/model_repository_manager.cc#L267). [S3FileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)](https://github.com/triton-inference-server/server/blob/main/src/core/filesystem.cc#L1453-L1460) does not update the value of mtime_ns for directories. This causes the garbage value in [mtime](https://github.com/triton-inference-server/server/blob/main/src/core/model_repository_manager.cc#L283) not get updated and the following comparison to use the garbage value in the mtime_ns and when garbage value > the last modified time we see the behavior in your issue. 

The behavior seen in your issue is often seen in S3 because prior to your changes in this PR we were comparing millisecond times instead of nanoseconds. Using nanoseconds greatly reduces the odds of the garbage value to be larger than the actual last modified time. Hence it appeared that this resolved the issue. 

Initializing mtime to 0 before the call to FileModificationTime() resolves the issue.

CoderHam:
@nsiddharth I picked up your change and added the other fix here: https://github.com/triton-inference-server/server/pull/2957
I'd appreciate your review. Thanks!

CoderHam:
Closed by https://github.com/triton-inference-server/server/pull/2957

nsiddharth:
I was able to build and test the latest image from main and it works as expected, thanks for fixing this.

PR #2945, opened 2021-05-31
Refactor Python backend tests

Tabrizian:
This PR breaks the Python backend tests into multiple subsections. It also fixes a minor bug in server where the Triton server was not calling the backend destructors after it has become ready.


src/servers/main.cc
```diff
@@ -1486,6 +1481,11 @@ main(int argc, char** argv)
     exit(1);
   }
 
+  // Start the HTTP, GRPC, and metrics endpoints.
+  if (!StartEndpoints(server, trace_manager, shm_manager)) {
+    exit(1);
+  }
+
```
Tabrizian, line 1488: I moved the registration of endpoints after the registration of signal handling so that when the server is ready the destructors are called properly.

CoderHam (approved):

PR #2946, opened 2021-06-01
Add tests for automatic restart of Python backend stubs

Tabrizian:
This should be merged after #2945

### DIFF
--- a/qa/L0_backend_python/restart/restart_test.py
+++ b/qa/L0_backend_python/restart/restart_test.py
@@ -1,0 +1,66 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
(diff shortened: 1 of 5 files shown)


qa/L0_backend_python/common.sh
```diff
@@ -0,0 +1,58 @@
+# Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
```
CoderHam, line 1: So now inside multiple test scripts? For ensemble, environment and lifecycle. And we keep the generic tests in python_test.py? 
Why not have a single test.sh with all the above?

qa/L0_backend_python/test.sh
```diff
@@ -447,85 +278,35 @@ $page_size."
...
-          echo "parameters: {key: \"EXECUTION_ENV_PATH\", value: {string_value: \"$path_to_conda_pack\"}}" >> config.pbtxt)
-cp ../python_models/python_version/model.py ./models/python_3_6/1/
-cp python_backend/builddir/triton_python_backend_stub ./models/python_3_6
-
-run_server
-if [ "$SERVER_PID" == "0" ]; then
-    echo -e "\n***\n*** Failed to start $SERVER\n***"
-    cat $SERVER_LOG
-    exit 1
+(cd env/ && bash -ex ./test.sh)
```
CoderHam, line 285: extra '/' after env. Here and in other places as well. Also './' is not needed before test.sh

qa/L0_backend_python/common.sh
CoderHam, line 1: Why not keep a single test.sh?

qa/L0_backend_python/restart/restart_test.py
```diff
@@ -0,0 +1,66 @@
...
+        shape = [1, 16]
+        model_name = 'identity_fp32'
+        dtype = np.float32
+
+        # Since the stub process has been killed, the first request
+        # will return an exception.
+        with self.assertRaises(InferenceServerException):
+            self._infer_helper(model_name, shape, dtype)
+        
+        # The second request should work properly since the stub process should
```
CoderHam, line 60: This looks like it could be a cause for intermittence. You should wait to check if the stub process is alive before this second inference call

CoderHam (changes requested):

src/servers/main.cc
```diff
@@ -1486,6 +1481,11 @@ main(int argc, char** argv)
     exit(1);
   }
 
+  // Start the HTTP, GRPC, and metrics endpoints.
```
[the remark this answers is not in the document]
Tabrizian: I explained this in the meeting. If we don't move this, the destructors will not be called properly even if the server is ready.

qa/L0_backend_python/common.sh
Tabrizian: Yes, that is correct. Because  `test.sh` becomes very large and hard to maintain.

qa/L0_backend_python/restart/restart_test.py
Tabrizian: Python backend takes of making sure that the stub is alive. The previous request will not be released until the stub has come alive. The source of intermittence was because of signal handlers not being registered before the server is ready.

qa/L0_backend_python/test.sh
Tabrizian: Fixed.

CoderHam (dismissed):

CoderHam (dismissed):

deadeyegoodwin (dismissed):

CoderHam (approved):

CoderHam (approved):

PR #2955, opened 2021-06-03
Update copyright checker to handle AFFILIATES

deadeyegoodwin:


qa/common/check_copyright.py
```diff
@@ -52,8 +52,8 @@
     'src/clients/c++/library/cencode.c', 'src/clients/c++/library/cencode.h',
     'src/clients/java', 'TRITON_VERSION')
 
-COPYRIGHT_YEAR_RE0 = 'Copyright \\(c\\) (20[0-9][0-9]), NVIDIA CORPORATION. All rights reserved.'
-COPYRIGHT_YEAR_RE1 = 'Copyright \\(c\\) (20[0-9][0-9])-(20[0-9][0-9]), NVIDIA CORPORATION. All rights reserved.'
+COPYRIGHT_YEAR_RE0 = 'Copyright \\(c\\) (20[0-9][0-9]), NVIDIA CORPORATION(?: & AFFILIATES)?. All rights reserved.'
+COPYRIGHT_YEAR_RE1 = 'Copyright \\(c\\) (20[0-9][0-9])-(20[0-9][0-9]), NVIDIA CORPORATION(?: & AFFILIATES)?. All rights reserved.'
```
CoderHam, line 56: I believe this does not support the case where years are as below:
2018,19-21
Additionally only the first year has the 20 prefix.

CoderHam (changes requested):

qa/common/check_copyright.py
deadeyegoodwin: Yes, but the script has never accepted that and we haven't yet had a case where we skip a year in the copyright. I we need that enhancement we can do it in a separate change.

CoderHam: Sure, @jbkyang-nvi has a similar PR up for model analyzer: https://github.com/triton-inference-server/model_analyzer/pull/158

deadeyegoodwin: I think you mean @kthui . If that other PR subsumes this one then we should bring it over to this repo and close this PR.

CoderHam: Makes sense. @kthui would you please do the following?

PR #2956, opened 2021-06-03
Update TRT backend based on TRT 8 APIs

GuanLuo:
This change should re-target to 21.06 branch once created as the main branch is still building against 21.05 containers.

### DIFF
--- a/src/backends/tensorrt/plan_backend.cc
+++ b/src/backends/tensorrt/plan_backend.cc
@@ -145,8 +145,7 @@ PlanBackend::Context::Context(
           name, gpu_device, max_batch_size, enable_pinned_input,
           enable_pinned_output, gather_kernel_buffer_threshold,
           std::move(metric_reporter)),
-      engine_(nullptr), is_shared_engine_(true), total_bindings_(0),
-      num_expected_bindings_(0),
+      engine_(nullptr), total_bindings_(0), num_expected_bindings_(0),
       use_output_copy_stream_(separate_output_stream),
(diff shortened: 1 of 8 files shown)


src/backends/tensorrt/plan_backend.cc
```diff
@@ -561,16 +549,15 @@ PlanBackend::Context::InitOptimizationProfiles(
...
       } else {
-        res.first->second.context_ = engine_->createExecutionContext();
+        res.first->second.context_.reset(engine_->createExecutionContext());
         if (res.first->second.context_ == nullptr) {
           return Status(
               Status::Code::INTERNAL, "unable to create TensorRT context");
         }
-        if (!res.first->second.context_->setOptimizationProfile(
-                profile_index)) {
+        if (!res.first->second.context_->setOptimizationProfileAsync(
```
deadeyegoodwin, line 559: So this is the only API change needed?

deadeyegoodwin:
> This change should re-target to 21.06 branch once created as the main branch is still building against 21.05 containers.

Do you expect this change not to build correctly on main? Isn't the new API we are using also valid for TRT 7.2?

src/backends/tensorrt/plan_backend.cc
GuanLuo: The destroy functions are also marked deprecated in TRT 8. In TRT 8 style, the runtime / engine / context are destroyed via their destructor, which is hidden in TRT 7.

PR #2957, opened 2021-06-03
Fix bug in GetModifiedTime that causes garbage modification time to be reported

CoderHam:
convert Millis to Nanoseconds for correct comparison

- [x] Add test coverage for model reload in L0_storage_S3

### DIFF
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -552,6 +552,7 @@ GCSFileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   bool is_dir;
   RETURN_IF_ERROR(IsDirectory(path, &is_dir));
   if (is_dir) {
+    *mtime_ns = 0;
     return Status::Success;
   }
 
@@ -1456,6 +1457,7 @@ S3FileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   bool is_dir;
   RETURN_IF_ERROR(IsDirectory(path, &is_dir));
   if (is_dir) {
+    *mtime_ns = 0;
     return Status::Success;
   }
 
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -1470,7 +1470,8 @@ S3FileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   // If request succeeds, copy over the modification time
   auto head_object_outcome = client_.HeadObject(head_request);
   if (head_object_outcome.IsSuccess()) {
-    *mtime_ns = head_object_outcome.GetResult().GetLastModified().Millis();
+    *mtime_ns =
+        head_object_outcome.GetResult().GetLastModified().Millis() * 1000000;
   } else {
     return Status(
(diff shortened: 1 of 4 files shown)


src/core/filesystem.cc
```diff
@@ -1470,7 +1470,8 @@ S3FileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   // If request succeeds, copy over the modification time
   auto head_object_outcome = client_.HeadObject(head_request);
   if (head_object_outcome.IsSuccess()) {
-    *mtime_ns = head_object_outcome.GetResult().GetLastModified().Millis();
+    *mtime_ns =
+        head_object_outcome.GetResult().GetLastModified().Millis() * 1000000;
```
deadeyegoodwin, line 1474: Use NANOS_PER_MILLIS from constants.h. Or better, is there a Nanos() call that can replace Millis()?

src/core/model_repository_manager.cc
```diff
@@ -280,7 +280,7 @@ GetModifiedTime(const std::string& path)
 
   // If 'path' is a file return its mtime. Otherwise, using the modification
   // time of the directory as baseline in case of file deletion
-  int64_t mtime;
+  int64_t mtime = 0;
   status = FileModificationTime(path, &mtime);
```
deadeyegoodwin, line 284: We need to fix the definition of FileModificationTime. If it returns success it should return a valid file modification time. If we ask for modification time for a file/folder and none is available we should have some defined behavior, like setting the modification time to 0 or returning a specific error code.

deadeyegoodwin (changes requested):

src/core/filesystem.cc
CoderHam: Will use NANOS_PER_MILLIS. No there is no Nanos() call for the AWS DateTime object
https://sdk.amazonaws.com/cpp/api/0.12.9/df/dc4/class_aws_1_1_utils_1_1_date_time.html

src/core/model_repository_manager.cc
CoderHam: I think we need not return an error for S3 and other filesystems that do not have a concept of timestamps for folders. The logic inside GetModifiedTime would log an unnecessary error in such a case. I'd rather we return 0 in the case where the path corresponds to a directory which does not have a modification time. 

That being said 0 is still not a valid file modification time.

src/core/filesystem.cc
```diff
@@ -552,6 +552,7 @@ GCSFileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   bool is_dir;
```
deadeyegoodwin, line 552: Update filesystem.h. "for some filesystems a file or folder may not have a modification time, in that case m_time returns 0"

deadeyegoodwin (changes requested):

src/core/filesystem.cc
CoderHam: Done

deadeyegoodwin (approved):

PR #2964, opened 2021-06-04
Serialize shared-library operations like dlopen and SetDllDirectory

deadeyegoodwin:


GuanLuo (dismissed):

GuanLuo (approved):

src/core/shared_library.cc
```diff
@@ -60,16 +72,55 @@ OpenLibraryHandle(const std::string& path, void** handle)
...
   }
+#endif
+
+  return Status::Success;
+}
+
+Status
+SharedLibrary::ResetLibraryDirectory()
+{
+#ifdef _WIN32
```
CoderHam, line 85: Why do we need to reset library path for windows?

deadeyegoodwin: We don't want the path we set for one backend to be used for anything else. More generally we could probably try to read what the original path was and restore it. But that is not necessary for the use cases we care about.

PR #2973, opened 2021-06-07
Fix null requests for TYPE_STRING

### LINKED ISSUE #2855: python backend streaming pybind11::error_already_set for perf_analyzer with concurrency > 1
**Description**
Try to perf_analyze a simple sequence model with concurrency > 1  (=1 no issues), get error message:
```
terminate called after throwing an instance of 'pybind11::error_already_set'
  what():  error: unpack_from requires a buffer of at least 3489662884 bytes for unpacking 3489662880 bytes at offset 4 (actual buffer size is 132)

At:
  /opt/tritonserver/backends/python/triton_python_backend_utils.py(117): deserialize_bytes_tensor
```

**Triton Information**
21.04

Are you using the Triton container or did you build it yourself?   Triton container

**To Reproduce**
Paste the model.py
```
import numpy as np
import json

# triton_python_backend_utils is available in every Triton Python model. You
# need to use this module to create inference requests and responses. It also
# contains some utility functions for extracting information from model_config
# and converting Triton input/output types to numpy types.
import triton_python_backend_utils as pb_utils
from multiprocessing.pool import ThreadPool

class Decoder(object):
    def __init__(self, blank):
        self.prev = ''
        self.result = ''
        self.blank_symbol = blank
    
    def decode(self, input, start, ready):
        """
        input: a list of characters/a string
        """
        if start:
            self.prev = ''
            self.result = ''
        
        if ready:
            for li in input.decode("utf-8"):
                if li != self.prev:
                    if li != self.blank_symbol:
                        self.result += li
                self.prev = li
        r = np.array([[self.result]], dtype=np.dtype(object))
        return r

class TritonPythonModel:
    """Your Python model must use the same class name. Every Python model
    that is created must have "TritonPythonModel" as the class name.
    """

    def initialize(self, args):
        """`initialize` is called only once when the model is being loaded.
        Implementing `initialize` function is optional. This function allows
        the model to intialize any state associated with this model.
        Parameters
        ----------
        args : dict
          Both keys and values are strings. The dictionary keys and values are:
          * model_config: A JSON string containing the model configuration
          * model_instance_kind: A string containing model instance kind
          * model_instance_device_id: A string containing model instance device ID
          * model_repository: Model repository path
          * model_version: Model version
          * model_name: Model name
        """

        # You must parse model_config. JSON string is not parsed here
        self.model_config = model_config = json.loads(args['model_config'])

        # get max batch size
        max_batch_size = max(model_config["max_batch_size"], 1)

        # get blank symbol from config
        blank = self.model_config.get("blank_id", '-')

        # initialize decoders
        self.decoders = [Decoder(blank) for i in range(max_batch_size)]

        # Get OUTPUT0 configuration
        output0_config = pb_utils.get_output_config_by_name(
            model_config, "OUTPUT0")
            
        # Convert Triton types to numpy types
        self.output0_dtype = pb_utils.triton_string_to_numpy(
            output0_config['data_type'])

    def batch_decode(self, batch_input, batch_start, batch_ready):
        responses = []
        args = []
        idx = 0
        for i,r,s in zip(batch_input, batch_ready, batch_start):
            args.append([idx, i, r, s])
            idx += 1
        
        with ThreadPool() as p:
            responses = p.map(self.process_single_request, args)

        return responses

    def process_single_request(self, inp):
        decoder_idx, input, ready, start = inp
        response = self.decoders[decoder_idx].decode(input[0], start[0], ready[0])
        output0_dtype = self.output0_dtype
        #print("output dtype:", output0_dtype, type(response))
        #print(response.astype(output0_dtype))
        out_tensor_0 = pb_utils.Tensor("OUTPUT0", response.astype(output0_dtype))
        inference_response = pb_utils.InferenceResponse(
                output_tensors=[out_tensor_0])
        return inference_response

    def execute(self, requests):
        """`execute` MUST be implemented in every Python model. `execute`
        function receives a list of pb_utils.InferenceRequest as the only
        argument. This function is called when an inference request is made
        for this model. Depending on the batching configuration (e.g. Dynamic
        Batching) used, `requests` may contain multiple requests. Every
        Python model, must create one pb_utils.InferenceResponse for every
        pb_utils.InferenceRequest in `requests`. If there is an error, you can
        set the error argument when creating a pb_utils.InferenceResponse
        Parameters
        ----------
        requests : list
          A list of pb_utils.InferenceRequest
        Returns
        -------
        list
          A list of pb_utils.InferenceResponse. The length of this list must
          be the same as `requests`
        """
        #print("START NEW")
        responses = []

        batch_input = []
        batch_ready = []
        batch_start = []
        batch_corrid = []
        # Every Python backend must iterate over everyone of the requests
        # and create a pb_utils.InferenceResponse for each of them.
        for request in requests:
            # Get INPUT0
            in_0 = pb_utils.get_input_tensor_by_name(request, "INPUT0")
            #in_0 ->  <triton_python_backend_utils.Tensor object
            # in_0 -> ndarray['xxx']

            batch_input += in_0.as_numpy().tolist()
            
            in_start = pb_utils.get_input_tensor_by_name(request, "START")
            batch_start += in_start.as_numpy().tolist()
            
            in_ready = pb_utils.get_input_tensor_by_name(request, "READY")
            batch_ready += in_ready.as_numpy().tolist()

            in_corrid = pb_utils.get_input_tensor_by_name(request, "CORRID")
            batch_corrid += in_corrid.as_numpy().tolist()

        #print("corrid", batch_corrid)
        #print("batch input", batch_input)
        responses = self.batch_decode(batch_input, batch_start, batch_ready)
        #print("batch response", responses)
        # You should return a list of pb_utils.InferenceResponse. Length
        # of this list must match the length of `requests` list.
        assert len(requests) == len(responses)
        #print("send responses", responses)
        return responses

    def finalize(self):
        """`finalize` is called only once when the model is being unloaded.
        Implementing `finalize` function is OPTIONAL. This function allows
        the model to perform any necessary clean ups before exit.
        """
        print('Cleaning up...')
```

The config.pbtxt
```
name: "ctc_decode_model"
backend: "python"
max_batch_size: 4
sequence_batching{
    max_sequence_idle_microseconds: 5000000
    direct {
      minimum_slot_utilization: 0.1
      max_queue_delay_microseconds: 100000
    }
    control_input [
        {
            name: "START",
            control [
                {
                    kind: CONTROL_SEQUENCE_START
                    fp32_false_true: [0,1]
                }
            ]
        },
        {
            name: "READY"
            control [
                {
                    kind: CONTROL_SEQUENCE_READY
                    fp32_false_true: [0, 1]
                }
            ]
        },
        {
            name: "CORRID",
            control [
                {
                    kind: CONTROL_SEQUENCE_CORRID
                    data_type: TYPE_UINT64
                }
            ]
        }
    ]
}
input [
  {
    name: "INPUT0"
    data_type: TYPE_STRING
    dims: [1] 
  }
]

output [
  {
    name: "OUTPUT0"
    data_type: TYPE_STRING
    dims: [-1]
  }
]

instance_group [{ kind: KIND_CPU }]
```

The commandline is 
```
perf_client --streaming -i gRPC --concurrency-range 2:2:1 -m ctc_decode_model
```

Tabrizian:
Fix the first 4 bytes to correctly represent the total number of bytes available in a null request. Closes #2855.

### DIFF
--- a/src/core/infer_request.cc
+++ b/src/core/infer_request.cc
@@ -251,9 +251,8 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
 
   // Second pass
   size_t max_byte_size = 0;
+  size_t max_str_byte_size = 0;
   const std::string* max_input_name;
-
-  bool includes_type_bytes = false;
   for (const auto& input : from.OriginalInputs()) {
     // Skip shape tensors in this pass
     if (input.second.IsShapeTensor()) {
@@ -261,16 +260,12 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
     }
 
     if (input.second.DType() == inference::DataType::TYPE_STRING) {
-      includes_type_bytes = true;
       int64_t element_count = GetElementCount(input.second.Shape());
-      if (element_count == -1) {
-        LOG_WARNING
-            << "Tensor should not have a variable dimension at this stage.";
-      }
 
       size_t str_byte_size = static_cast<size_t>(4 * element_count);
       if (str_byte_size > max_byte_size) {
         max_byte_size = str_byte_size;
+        max_str_byte_size = max_byte_size;
         max_input_name = &(input.first);
       }
     } else {
@@ -289,12 +284,12 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
       std::make_shared<AllocatedMemory>(max_byte_size, mem_type, mem_id);
   auto data_base = data->BufferAt(0, &max_byte_size, &mem_type, &mem_id);
 
-  // Zero initialization is only required when there is
-  // a TYPE_BYTES tensor in the request.
-  if (includes_type_bytes) {
+  // Zero initialization is only required when there is a TYPE_BYTES tensor in
+  // the request. Only set the required number of bytes to zero.
+  if (max_str_byte_size > 0) {
     std::fill(
         data->MutableBuffer(),
-        data->MutableBuffer() + max_byte_size, 0);
+        data->MutableBuffer() + max_str_byte_size, 0);
   }
 
   for (const auto& input : from.OriginalInputs()) {
--- a/src/core/infer_request.cc
+++ b/src/core/infer_request.cc
@@ -26,6 +26,7 @@
 
 #include "src/core/infer_request.h"
 
+#include <algorithm>
 #include <deque>
 #include "src/core/backend.h"
 #include "src/core/logging.h"
@@ -263,9 +264,9 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
       int64_t element_count = GetElementCount(input.second.Shape());
 
       size_t str_byte_size = static_cast<size_t>(4 * element_count);
+      max_str_byte_size = std::max(str_byte_size, max_str_byte_size);
       if (str_byte_size > max_byte_size) {
         max_byte_size = str_byte_size;
-        max_str_byte_size = max_byte_size;
         max_input_name = &(input.first);
       }
     } else {
@@ -288,8 +289,7 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
   // the request. Only set the required number of bytes to zero.
   if (max_str_byte_size > 0) {
     std::fill(
-        data->MutableBuffer(),
-        data->MutableBuffer() + max_str_byte_size, 0);
+        data->MutableBuffer(), data->MutableBuffer() + max_str_byte_size, 0);
   }
 
   for (const auto& input : from.OriginalInputs()) {
--- a/src/core/infer_request.cc
+++ b/src/core/infer_request.cc
@@ -249,18 +249,33 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
     new_input->SetData(data);
   }
 
-
   // Second pass
   size_t max_byte_size = 0;
+  int64_t max_str_element_count = 0;
+  const std::string* max_str_input_name;
   const std::string* max_input_name;
   for (const auto& input : from.OriginalInputs()) {
     // Skip shape tensors in this pass
     if (input.second.IsShapeTensor()) {
       continue;
     }
-    if (input.second.Data()->TotalByteSize() >= max_byte_size) {
-      max_byte_size = input.second.Data()->TotalByteSize();
-      max_input_name = &(input.first);
+
+    if (input.second.DType() == inference::DataType::TYPE_STRING) {
+      int64_t element_count = GetElementCount(input.second.Shape());
+      if (element_count == -1) {
+        LOG_WARNING
+            << "Tensor should not have a variable dimension at this stage.";
+      }
+
+      if (element_count > max_str_element_count) {
+        max_str_element_count = element_count;
+        max_str_input_name = &(input.first);
+      }
+    } else {
+      if (input.second.Data()->TotalByteSize() >= max_byte_size) {
+        max_byte_size = input.second.Data()->TotalByteSize();
+        max_input_name = &(input.first);
+      }
     }
   }
 
@@ -271,6 +286,25 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
   std::shared_ptr<Memory> data =
       std::make_shared<AllocatedMemory>(max_byte_size, mem_type, mem_id);
   auto data_base = data->BufferAt(0, &max_byte_size, &mem_type, &mem_id);
+
+  // Number of bytes required for the string tensor is 4 times the number of
+  // elements in the tensor. Each byte represent the length of the string in
+  // each element which in the case of null request is zero.
+  size_t max_str_byte_size = max_str_element_count * 4;
+
+  std::shared_ptr<MutableMemory> data_str;
+  const char* data_base_str;
+
+  if (max_str_element_count > 0) {
+    data_str =
+        std::make_shared<AllocatedMemory>(max_str_byte_size, mem_type, mem_id);
+    std::fill(
+        data_str->MutableBuffer(),
+        data_str->MutableBuffer() + max_str_byte_size, 0);
+    data_base_str =
+        data_str->BufferAt(0, &max_str_byte_size, &mem_type, &mem_id);
+  }
+
   for (const auto& input : from.OriginalInputs()) {
     // skip shape tensors in this pass
     if (input.second.IsShapeTensor()) {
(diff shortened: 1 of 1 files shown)


src/core/infer_request.cc
```diff
@@ -284,14 +284,41 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
     *new_input->MutableShape() = input.second.Shape();
     *new_input->MutableShapeWithBatchDim() = input.second.ShapeWithBatchDim();
 
+    std::shared_ptr<MutableMemory> new_data_str;
+
+    // Custom handling for null requests in TYPE_STRING. The null requests in
+    // type string have to follow a certain structure. We set the first 4 bytes
+    // in the Tensor to represent the number of total available bytes.
+    if (inference::DataType::TYPE_STRING == input.second.DType()) {
```
deadeyegoodwin, line 292: It seems like this only works if a string input has shape [ -1 ]. What if the shape is [ 2, 2 ]. Then you need to initialize the data to be 4 empty strings. So 16 bytes of zeros.

My thoughts:
- In second pass above when you have a string tensor you calculate its data size as "num_elements * 4" because you are going to make each element an empty string so only need 4 bytes per element. Remember the max across all string inputs as max_string_byte_size. 
- When allocating 'data' use the max of max_byte_size and max_string_byte_size.
- If max_string_byte_size > 0 then zero that many bytes at the beginning of data so that it can be used for the string tensor data
- In the below when appending data for string tensor you must use "num_elements * 4" instead of TotalByteSize.

deadeyegoodwin (changes requested):

src/core/infer_request.cc
```diff
@@ -284,14 +284,41 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
...
+      new_data_str = std::make_shared<AllocatedMemory>(
+          input.second.Data()->TotalByteSize(), mem_type, mem_id);
+      uint32_t* new_data_base =
+          reinterpret_cast<uint32_t*>(new_data_str->MutableBuffer());
```
GuanLuo, line 296: The `new_data_base` doesn't seem to be used.

src/core/infer_request.cc
```diff
@@ -284,14 +284,41 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
...
+        new_input->SetData(new_data_str);
+      } else {
+        new_input->SetData(data);
+      }
     } else {
-      new_input->AppendData(
-          data_base, input.second.Data()->TotalByteSize(), mem_type, mem_id);
+      if (inference::DataType::TYPE_STRING == input.second.DType()) {
+        new_input->AppendData(
+            new_data_str->MutableBuffer(), input.second.Data()->TotalByteSize(),
```
GuanLuo, line 316: I don't think using the total byte size from the actual request is correct, should be calculated by `num_elements * 4`

GuanLuo (changes requested):

src/core/infer_request.cc
Tabrizian: @deadeyegoodwin I see. I missed the multi-dimensional case. I'll fix this.

src/core/infer_request.cc
Tabrizian: It is used in the next line.

GuanLuo: My bad, I misread that. I think given that you are just allocating buffer for "num_elements" of empty string, you just need to initialize the whole buffer to zeros.

src/core/infer_request.cc
```diff
@@ -284,14 +284,41 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
...
+    if (inference::DataType::TYPE_STRING == input.second.DType()) {
+      new_data_str = std::make_shared<AllocatedMemory>(
+          input.second.Data()->TotalByteSize(), mem_type, mem_id);
+      uint32_t* new_data_base =
+          reinterpret_cast<uint32_t*>(new_data_str->MutableBuffer());
+
+      // Total number of bytes available inside the Tensor is equal to the
+      // tensor's total byte size minus 4 bytes. The first four bytes are used
+      // to determine the string length.
+      *new_data_base = input.second.Data()->TotalByteSize() - 4;
```
CoderHam, line 301: You first point to the base address of the mutable buffer with new_data_base and then update it to be the total size of the buffer - length of the string. Why?

CoderHam (changes requested):

src/core/infer_request.cc
Tabrizian: @CoderHam The problem is that null requests for TYPE_STRING do not represent a correct Tensor. This was an attempt to fix it.

Tabrizian:
I tested this with multi-dimensional tensors as well as tensors with variable shapes in them and seems to be working fine.

src/core/infer_request.cc
```diff
@@ -249,18 +249,33 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
     new_input->SetData(data);
   }
 
-
   // Second pass
   size_t max_byte_size = 0;
+  int64_t max_str_element_count = 0;
```
deadeyegoodwin, line 254: Why not just use max_byte_size and max_input_name for both strings and non-strings. Then you just need a single alloc and don't need a lot of the other changes below? It will also be more performant and this code already has performance problems so should try not to make it slower.

deadeyegoodwin (changes requested):

src/core/infer_request.cc
Tabrizian: Addressed this comment in the next commit.

src/core/infer_request.cc
```diff
@@ -249,28 +249,54 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
     new_input->SetData(data);
   }
 
-
   // Second pass
   size_t max_byte_size = 0;
   const std::string* max_input_name;
+
+  bool includes_type_bytes = false;
```
deadeyegoodwin, line 256: Use a "zero_bytes_count" variable initialized to 0 and update it for TYPE_STRING to be the max str_byte_size. Then you only need to zero out the first 'zero_bytes_count' bytes of the buffer instead of zeroing the whole thing.

deadeyegoodwin (changes requested):

src/core/infer_request.cc
```diff
@@ -249,28 +249,54 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
...
-    if (input.second.Data()->TotalByteSize() >= max_byte_size) {
-      max_byte_size = input.second.Data()->TotalByteSize();
-      max_input_name = &(input.first);
+
+    if (input.second.DType() == inference::DataType::TYPE_STRING) {
+      includes_type_bytes = true;
+      int64_t element_count = GetElementCount(input.second.Shape());
+      if (element_count == -1) {
+        LOG_WARNING
+            << "Tensor should not have a variable dimension at this stage.";
```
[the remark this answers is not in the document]
GuanLuo: Okay, then I don't know if you actually need to add this check..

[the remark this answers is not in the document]
Tabrizian: Ok, I'll remove this warning.

src/core/infer_request.cc
```diff
@@ -249,28 +249,49 @@ InferenceRequest::CopyAsNull(const InferenceRequest& from)
...
     }
-    if (input.second.Data()->TotalByteSize() >= max_byte_size) {
-      max_byte_size = input.second.Data()->TotalByteSize();
-      max_input_name = &(input.first);
+
+    if (input.second.DType() == inference::DataType::TYPE_STRING) {
+      int64_t element_count = GetElementCount(input.second.Shape());
+
+      size_t str_byte_size = static_cast<size_t>(4 * element_count);
+      if (str_byte_size > max_byte_size) {
```
deadeyegoodwin, line 266: You need to unconditionally use "max_str_byte_size = max(str_byte_size, max_str_byte_size)" because this could be a bigger string tensor but not be bigger than max_byte_size.

deadeyegoodwin (changes requested):

src/core/infer_request.cc
Tabrizian: Good catch, sorry about this. I fixed this in the next commit and tested this with the config below:

```
input [
  {
    name: "INPUT0"
    data_type: TYPE_STRING
    dims: [1]
  },
  {
    name: "INPUT1"
    data_type: TYPE_FP32
    dims: [7]
  },
  {
    name: "INPUT2"
    data_type: TYPE_STRING
    dims: [2]
  },
  {
    name: "INPUT3"
    data_type: TYPE_STRING
    dims: [2, 2]
  },
  {
    name: "INPUT4"
    data_type: TYPE_STRING
    dims: [2]
  }
]

output [
  {
    name: "OUTPUT0"
    data_type: TYPE_STRING
    dims: [-1]
  }
]

instance_group [{ kind: KIND_CPU }]
```

I hope it covers all the corner cases and there are not any further issues.

deadeyegoodwin (approved):

deadeyegoodwin (approved):

CoderHam (approved):

GuanLuo (approved):

PR #2981, opened 2021-06-08
Fix cuda memory manager destructor when not initialized

CoderHam:
- cnmem is not initialized when memory pool sizes are 0

### DIFF
--- a/src/core/cuda_memory_manager.cc
+++ b/src/core/cuda_memory_manager.cc
@@ -61,10 +61,12 @@ std::mutex CudaMemoryManager::instance_mu_;
 
 CudaMemoryManager::~CudaMemoryManager()
 {
-  auto status = cnmemFinalize();
-  if (status != CNMEM_STATUS_SUCCESS) {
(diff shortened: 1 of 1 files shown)


src/core/cuda_memory_manager.h
```diff
@@ -76,8 +76,13 @@ class CudaMemoryManager {
...
 
  private:
-  CudaMemoryManager(bool has_allocation) : has_allocation_(has_allocation) {}
+  CudaMemoryManager(bool has_allocation)
+      : has_allocation_(has_allocation), initialized_(false)
+  {
+  }
+
   bool has_allocation_;
+  bool initialized_;
```
GuanLuo, line 85: Why not just use `has_allocation_`

CoderHam: We need a case for when CudaMemoryManager has been created but cnmemInit has not been called i.e. when memory pool sizes are 0.

deadeyegoodwin (dismissed):

GuanLuo (dismissed):

deadeyegoodwin (approved):

PR #2984, opened 2021-06-09
Fix calculation of the NUMA node count

GuanLuo:


src/core/numa_utils.cc
```diff
@@ -124,10 +124,10 @@ SetNumaMemoryPolicy(const HostPolicyCmdlineConfig& host_policy)
     RETURN_IF_ERROR(
         ParseIntOption("Parsing 'numa-node' value", it->second, &node_id));
     LOG_VERBOSE(1) << "Thread is binding to NUMA node " << it->second
-                   << ". Max NUMA node count: " << numa_max_node();
+                   << ". Max NUMA node count: " << (numa_max_node() + 1);
     numa_set = true;
     unsigned long node_mask = 1UL << node_id;
-    if (set_mempolicy(MPOL_BIND, &node_mask, numa_max_node() + 1) != 0) {
+    if (set_mempolicy(MPOL_BIND, &node_mask, (numa_max_node() + 1) + 1) != 0) {
```
CoderHam, line 130: change to numa_max_node() + 2

CoderHam (changes requested):

deadeyegoodwin (approved):

src/core/numa_utils.cc
GuanLuo: The two "+" has different meanings, the first is getting the number of nodes, second is one-plus needed to describe the mask

CoderHam: I see. So you have it written as so for clarity

CoderHam (approved):

PR #2990, opened 2021-06-09
Fix classification label error

hoshimura:
I believe the class name lookup should be indexed by class instead of the global index in the probabilities file.

Change-Id: I508892dda626a024dda8ffe589f321de0523a55b

### DIFF
--- a/src/servers/classification.cc
+++ b/src/servers/classification.cc
@@ -56,7 +56,7 @@ AddClassResults(
 
     const char* label;
     RETURN_IF_ERR(TRITONSERVER_InferenceResponseOutputClassificationLabel(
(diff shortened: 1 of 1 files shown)


deadeyegoodwin:
Do you have an example of why this change is needed. I believe we have many tests where the labels are being used correctly with the current implementation.

hoshimura:
I released as you posed your question that the classification extension is for classification only :smile: . I have a detection model with a probs tensor with probabilities per class and detection. So what I would like to have is a classification of every row or detection obviously. And you are right that this pull request will not be able to fix that for me and on the way destroy the classification functionality :woozy_face: !  Maybe there are other extensions for this? Otherwise we can close this and I will have to

PR #2991, opened 2021-06-09
New design for better performance of no batching case with multiple i…

tanmayv25:
…nstances

So, in this re-design we have modified the interface of rate limiter quite a bit. Instead of earlier callback based signaling, we will use the Enqueue and Dequeue methods to control when each instance gets to run inference.

The current implementation is just a proposal for a simple no batching case. I have verified that it keeps the same performance as our current threading model when the RL is disabled. However, there are some weirdness that I can not yet explain. More on this at the end.

Following are some points about this design:
1. The scheduler threads will just be batcher threads and will come into the picture only when a batching is enabled.
2. The request queue is maintained by RateLimiter. This will be replaced with PriorityQueue from scheduler_utils in the future to include functionalities like timeouts and priorities.
3. BackendThreads will wait on RL to receive a payload to use for inference. In case of the high load and no rate limiting, this reduces to just a pop out of the queue.
4. Payload infrastructure is kept because later for device blocking case the same thread can be used to intialize and warmup the multiple model instances whenever required.
5. In order to interact with the BackendThread, user must request for a payload oobject from RL, configure the object and then enqueue to the RL. The RL will then pass it to the waiting BackendThread whenever scheduling for inference.

### DIFF
--- a/src/backends/backend/triton_model_instance.cc
+++ b/src/backends/backend/triton_model_instance.cc
@@ -36,6 +36,7 @@
 #include "src/core/logging.h"
 #include "src/core/metrics.h"
 #include "src/core/nvtx.h"
+#include "src/core/server.h"
 
 namespace nvidia { namespace inferenceserver {
 
@@ -141,17 +142,6 @@ TritonModelInstance::CreateInstances(
     TritonModel* model, const inference::ModelConfig& model_config,
     const bool device_blocking)
 {
-  bool use_backend_threads = false;
-  size_t count = 0;
-  for (const auto& group : model_config.instance_group()) {
-    if (!group.passive()) {
-      count += group.count();
-      if (count > 1) {
-        use_backend_threads = true;
-        break;
-      }
-    }
-  }
   // This structure is used to allocate TritonBackendThread to instances on same
   // device for device blocking execution policy.
   std::map<uint32_t, std::shared_ptr<TritonBackendThread>> device_to_thread_map;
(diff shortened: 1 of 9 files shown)


src/backends/backend/triton_model_instance.cc
```diff
@@ -231,11 +218,12 @@ TritonModelInstance::CreateInstance(
...
 Status
 TritonModelInstance::SetBackendThread(
-    const int32_t device_id, const bool device_blocking,
+    const TRITONSERVER_InstanceGroupKind kind, const int32_t device_id,
+    const bool device_blocking,
     std::map<uint32_t, std::shared_ptr<TritonBackendThread>>*
         device_to_thread_map)
 {
-  if (device_blocking) {
+  if (device_blocking && kind == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
```
deadeyegoodwin, line 226: parens

src/backends/backend/triton_model_instance.cc
```diff
@@ -246,10 +234,13 @@ TritonModelInstance::SetBackendThread(
   if (triton_backend_thread_.get() == nullptr) {
     std::unique_ptr<TritonBackendThread> local_backend_thread;
     RETURN_IF_ERROR(TritonBackendThread::CreateBackendThread(
-        Name(), 5 /* nice */, device_id, &local_backend_thread));
+        Name(), model_, 5 /* nice */, device_id, &local_backend_thread));
```
deadeyegoodwin, line 237: don't use nice (nice = 0)

deadeyegoodwin (changes requested):

tanmayv25:
I do have to implement the preserve ordering case as well.

deadeyegoodwin:
> 
> 
> I do have to implement the preserve ordering case as well.

It seems like that is completely handled by the batcher and the inference request... I wouldn't expect your changes to have any interaction

PR #3002, opened 2021-06-10
Add explicit model control test for Azure storage

CoderHam:
Fix case errno is 0 but the blob_property.valid() returns false. It appears errno is the correct way to check for errors based on the azure sdk [sample](https://github.com/Azure/azure-storage-cpplite/blob/master/sample/sample2.cpp).

### DIFF
--- a/src/core/filesystem.cc
+++ b/src/core/filesystem.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
+// Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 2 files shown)


deadeyegoodwin (approved):

tanmayv25 (dismissed):

src/core/filesystem.cc
```diff
@@ -916,7 +916,7 @@ ASFileSystem::FileModificationTime(const std::string& path, int64_t* mtime_ns)
   RETURN_IF_ERROR(ParsePath(path, &container, &object_path));
 
   auto blobProperty = bc.get_blob_property(container, object_path);
-  if (!blobProperty.valid()) {
+  if (errno != 0) {
```
deadeyegoodwin, line 919: Very strange for a C++ API to use errno...

deadeyegoodwin (approved):

src/core/filesystem.cc
CoderHam: That is what I thought as well. Additionally there are some cases where the valid is set to false even when errno is set to 0. While I was not able to reproduce such a case this is what the caused the error that the user was seeing in issue 2966

tanmayv25 (approved):

PR #3005, opened 2021-06-11
Add logging for creation of new runtimes and engines

CoderHam:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -414,6 +413,14 @@ PlanBackend::CreateExecutionContexts(
               mn_itr->second, dla_core_id, &eit->second.first,
               &eit->second.second));
 
+          if (dla_core_id != -1) {
```
deadeyegoodwin, line 416: Shouldn't all these logs be in LoadPlan where the engine and runtime are actually created?
Also, you should always print the gpu_device, even when dla is used.

deadeyegoodwin (changes requested):

src/backends/tensorrt/plan_backend.cc
CoderHam: The reason we cannot log this inside LoadPlan is because in the dynamic shape we destroy the engine after creation. 
We also do not pass the gpu_id, and model name to LoadPlan().

deadeyegoodwin: I see, then can just reduce it to a single line. Since it is verbose log it is ok to just always print gpu id and dla id even when dla id is -1

tanmayv25 (dismissed):

tanmayv25 (approved):

CoderHam:
@deadeyegoodwin I cannot merge until I get a re-review from you

PR #3016, opened 2021-06-15
Fix wrapped copyright in src/core/metrics.cc

Tabrizian:


src/core/metrics.cc
```diff
@@ -1,5 +1,4 @@
-// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights
-// reserved.
+// Copyright (c) 2018-21, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 1: Just remove the '(c)' after Copyright since it is causing the copyright to be split into 2 lines because of the formatter.

CoderHam (changes requested):

src/core/metrics.cc
Tabrizian: Fixed this in other places too.

CoderHam (approved):

PR #3017, opened 2021-06-15
Add tests for non-contiguous tensors for Python backend

Tabrizian:
Tests for triton-inference-server/python_backend#61.


qa/L0_backend_python/python_test.py
```diff
@@ -257,6 +257,30 @@ def test_string(self):
                 self.assertTrue(output0 is not None)
                 self.assertTrue(output0[0] == input_data.astype(np.bytes_))
 
+    def test_contiguous(self):
```
CoderHam, line 260: This tests appears to be testing support for non contiguous output tensors. Thus test_non_contiguous would be more appropriate.

CoderHam (changes requested):

CoderHam (approved):

PR #3024, opened 2021-06-17
Build dynamic batches using BatcherThread

tanmayv25:
This enables the dynamic batches to the models. I have fixed the perf difference for most of the cases. There are few cases performance of this DB still lags. The reason is now the batcher thread acquires the lock to the priority queue more frequently. I have to do some more tuning and believe after which the performance might actually improve.

UPDATE: I have done some of the tune ups to batcher. Take a look at the performance analysis and further scope with DLIS-2534.

### DIFF
--- a/src/core/dynamic_batch_scheduler.cc
+++ b/src/core/dynamic_batch_scheduler.cc
@@ -168,7 +168,7 @@ DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
       // this request. We do the actual wake outside of the lock to avoid having
       // the woken thread immediately block on the lock
       wake_batcher =
-          model_->Server()->GetRateLimiter()->IdleInstanceCount(model_) > 0;
+          model_->Server()->GetRateLimiter()->PayloadSlotAvailable(model_);
 
       // We may wake up runner less often if we don't enforce equal shape within
       // a batch, otherwise must always wake up runner to check it
@@ -223,7 +223,6 @@ DynamicBatchScheduler::BatcherThread(const int nice)
   }
 
   NewPayload();
-  const uint64_t default_wait_microseconds = 5 * 1000;
 
   while (!scheduler_thread_exit_.load()) {
     NVTX_RANGE(nvtx_, "DynamicBatcher " + model_->Name());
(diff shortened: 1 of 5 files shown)


src/core/dynamic_batch_scheduler.cc
```diff
@@ -153,210 +153,202 @@ DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
...
+                       << " queued requests, current total = " << queue_.Size();
+      } else if (queue_.Empty()) {
+        wait_microseconds = 0;
+      } else {
+        if (payload_saturated_) {
+          continue;
+        }
+        auto wait_for_slots = [this]() {
+          return model_->Server()->GetRateLimiter()->PayloadSlotAvailable(
+              model_);
```
GuanLuo, line 269: The lambda can be moved outside the while loop?

src/core/dynamic_batch_scheduler.cc
```diff
@@ -153,210 +153,202 @@ DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
...
+          if ((wait_microseconds == 0) && (pending_batch_queue_cnt != 0)) {
+            curr_payload_->ReserveRequests(pending_batch_queue_cnt);
+            for (size_t idx = 0; idx < pending_batch_queue_cnt; ++idx) {
+              std::unique_ptr<InferenceRequest> request;
+              auto status = queue_.Dequeue(&request);
+              if (status.IsOk()) {
+                if (preserve_ordering_) {
+                  DelegateResponse(request);
+                }
+                curr_payload_->AddRequest(std::move(request));
```
GuanLuo, line 299: It may be more clear to have a SetState() function and calls it at the end of adding pending requests

src/core/rate_limiter.h
```diff
@@ -96,6 +96,8 @@ class RateLimiter {
   Status UnregisterModel(const TritonModel* model);
 
   void InitializePayloadQueues(const TritonModelInstance* instance);
+  size_t IdleInstanceCount(const TritonModel* model);
```
GuanLuo, line 99: Where does this function be used?

tanmayv25: It was needed for one of the variations I tried. But it was not giving best performance. I will clean this up.

GuanLuo (approved):

deadeyegoodwin (approved):

PR #3028, opened 2021-06-21
Test for PA binary within tritonclient pip wheel

tanmayv25:
Check for cases like: https://github.com/triton-inference-server/client/pull/18.

### DIFF
--- a/qa/L0_sdk/test.sh
+++ b/qa/L0_sdk/test.sh
@@ -166,7 +166,7 @@ python -c """import tritonclient; import tritonclient.grpc; import tritonclient.
           import tritonclient.utils; import tritonclient.grpc.model_config_pb2; \
           import tritonclient.grpc.service_pb2; import tritonclient.grpc.service_pb2_grpc; \
           import tritonclient.utils.cuda_shared_memory; import tritonclient.utils.shared_memory"""
-RET_VAL=$(($RET_VAL+$?))
+RET=$(($RET+$?))
 EXECUTABLES="perf_analyzer perf_client"
 for l in $EXECUTABLES; do
   if [ $(which -a $l | grep "/usr/local/bin/$l" | wc -l) -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


qa/L0_sdk/test.sh
```diff
@@ -161,6 +161,21 @@ for l in $WHLS; do
...
+# Check wheel installation
+python -c """import tritonclient; import tritonclient.grpc; import tritonclient.http; \
+          import tritonclient.utils; import tritonclient.grpc.model_config_pb2; \
+          import tritonclient.grpc.service_pb2; import tritonclient.grpc.service_pb2_grpc; \
+          import tritonclient.utils.cuda_shared_memory; import tritonclient.utils.shared_memory"""
+RET_VAL=$(($RET_VAL+$?))
+EXECUTABLES="perf_analyzer perf_client"
+for l in $EXECUTABLES; do
+  if [ $(which -a $l | grep "/usr/local/bin/$l" | wc -l) -ne 1 ]; then
+    which -a $l
```
CoderHam, line 173: Is this line need?

CoderHam (changes requested):

qa/L0_sdk/test.sh
```diff
@@ -161,6 +161,21 @@ for l in $WHLS; do
     fi
 done
 
+# Check wheel installation
+python -c """import tritonclient; import tritonclient.grpc; import tritonclient.http; \
+          import tritonclient.utils; import tritonclient.grpc.model_config_pb2; \
+          import tritonclient.grpc.service_pb2; import tritonclient.grpc.service_pb2_grpc; \
+          import tritonclient.utils.cuda_shared_memory; import tritonclient.utils.shared_memory"""
+RET_VAL=$(($RET_VAL+$?))
```
CoderHam, line 169: I don't see this being used anywhere

CoderHam (changes requested):

qa/L0_sdk/test.sh
tanmayv25: In case of failure we will print all the locations of executables in the system path. This might help us in understanding what is wrong right away.

CoderHam (approved):

PR #3039, opened 2021-06-23
Custom container composition script

jbkyang-nvi:
Make compose.py so users can compose custom containers with different backends themselves.
Gitlab test PR: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/merge_requests/327


compose.py
```diff
@@ -0,0 +1,305 @@
...
+    [ `id -g $TRITON_SERVER_USER` -eq 1000 ]
+
+WORKDIR /opt/tritonserver
+RUN rm -fr /opt/tritonserver/*
+COPY --chown=1000:1000 --from=full /opt/tritonserver/LICENSE .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/TRITON_VERSION .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/NVIDIA_Deep_Learning_Container_License.pdf .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/bin bin/
+COPY --chown=1000:1000 --from=full /opt/tritonserver/lib lib/
+COPY --chown=1000:1000 --from=full /opt/tritonserver/include/triton/core include/triton/core
```
GuanLuo, line 118: Wouldn't your `min` based image then still have everything from `full` image?

compose.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+### add additional backends needed
+def add_requested_backends(ddir, dockerfile_name, backends):
+    df = "# Copying over backends \n"
+    for backend in backends:
+        if (backend.lower()
+                in (EXAMPLE_BACKENDS + CORE_BACKENDS + NONCORE_BACKENDS)):
+            df += '''COPY --chown=1000:1000 --from=full /opt/tritonserver/backends/{} /opt/tritonserver/backends/{}    
+'''.format(backend, backend)
```
GuanLuo, line 134: Since everything is copied from the release image, I suggest currently this compose script doesn't support composition of custom builds? i.e. no GPU, no cloud storage, etc.?

qa/L0_infer/infer_test.py
```diff
@@ -39,8 +39,8 @@
     int(os.environ.get('TEST_SYSTEM_SHARED_MEMORY', 0)))
 TEST_CUDA_SHARED_MEMORY = bool(int(os.environ.get('TEST_CUDA_SHARED_MEMORY',
                                                   0)))
-CPU_ONLY = (os.environ.get('TRITON_SERVER_CPU_ONLY') is not None)
-
+#CPU_ONLY = (os.environ.get('TRITON_SERVER_CPU_ONLY') is not None)
+CPU_ONLY = False
```
GuanLuo, line 43: Why you need to change the test? Won't that break existing tests?

GuanLuo (changes requested):

compose.py
```diff
@@ -0,0 +1,305 @@
...
+    df += '''
+# Add entrypoint script
+WORKDIR /opt/tritonserver
+COPY --chown=1000:1000 --from=full /opt/tritonserver/nvidia_entrypoint.sh .
+ENTRYPOINT ["/opt/tritonserver/nvidia_entrypoint.sh"]
+'''
+    with open(os.path.join(ddir, dockerfile_name), "a") as dfile:
+        dfile.write(df)
+
+### create build container
```
deadeyegoodwin, line 196: remove build terminology

compose.py
```diff
@@ -0,0 +1,305 @@
...
+def build_docker_container(ddir, dockerfile_name, container_name):
+    # Before attempting to run the new image, make sure any
```
deadeyegoodwin, line 198: You shouldn't have to remove existing image before building.

compose.py
```diff
@@ -0,0 +1,305 @@
...
+        help=
+        'Include specified backend in build as <backend-name>. To have multiple backends built, specify --backend <foo> --backend <bar>'
+    )
+    parser.add_argument(
+        '--repoagent',
+        action='append',
+        required=False,
+        help='Include specified repo agent in build as <repoagent-name>')
+    FLAGS = parser.parse_args()
+
```
deadeyegoodwin, line 273: Add --enable-gpu as in build.py. For now not supporting cpu-only build so should raise error if --enable-gpu is not specified.

compose.py
```diff
@@ -0,0 +1,305 @@
...
+    dockerfile_name = 'Dockerfile.buildbase'
+    workdir_path="/opt/tritonserver"
+    
+    if FLAGS.backend is None:
+        FLAGS.backend = []
+    if FLAGS.repoagent is None:
+        FLAGS.repoagent = []
+
+    make_dockerfile(FLAGS.build_dir, dockerfile_name)
```
deadeyegoodwin, line 300: This breakout of functions is confusing. I see why you want add_requested_backends and add_requested_repoagents separately... so you can use them in the future cpu-only dockerfile... but other than show I would just have "start_gpu_dockerfile" and "end_gpu_dockerfile" or similar.

docs/compose.md
```diff
@@ -39,8 +39,32 @@ releases will increase the amount of customization that is available.
 It is also possible to [build Triton](build.md#building-triton)
 yourself to get more exact customization.
```
deadeyegoodwin, line 40: change "yourself" to "from source"

docs/compose.md
```diff
@@ -39,8 +39,32 @@ releases will increase the amount of customization that is available.
...
 
-The two Docker images used for customization are retrieved using the
-following commands.
+## Use the compose.py script
+
+The `compose.py` script uses the existing NGC containers
```
deadeyegoodwin, line 44: "containers" -> "Docker images"

compose.py
jbkyang-nvi: I was wondering if I need to copy over the `bin/`, `lib/` or `include/triton/core`. I think  the benefit for this script is that it doesn't include some of the backends in the `full` image

docs/compose.md
```diff
@@ -39,8 +39,32 @@ releases will increase the amount of customization that is available.
...
-The two Docker images used for customization are retrieved using the
-following commands.
+## Use the compose.py script
+
+The `compose.py` script uses the existing NGC containers 
+```
+nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
+nvcr.io/nvidia/tritonserver:<xx.yy>-py3
+```
+to build the new custom container. The resulting container version will be `xx.yy` as it inherits from the pre-mentioned builds.
```
deadeyegoodwin, line 49: "new custom Triton Docker image".

I would drop "The resulting ..." sentence

docs/compose.md
```diff
@@ -39,8 +39,32 @@ releases will increase the amount of customization that is available.
...
+
+### Customizing backends and repo-agents
```
deadeyegoodwin, line 51: I would not create a new section here but just continue the previous paragraph...
"`compose.py` provides `--backend` and `--repoagent` options that allow you to specify which backends and repository agents you want to include in the custom image. The --enable-gpu flag indicates that you want to create an image that supports NVIDIA GPUs. For example, the following creates a new docker image that contains only the TensorFlow 1 and TensorFlow 2 backends and the checksum repository agent.

docs/compose.md
```diff
@@ -39,8 +39,32 @@ releases will increase the amount of customization that is available.
...
+
+`compose.py` provides a `--backend` option to specify the backend and a `--repoagent` option to specify the repository agent for the new container. The requested backend needs to be an officialy supported tritonbackend.
+ 
+Example:
+```
+python3 compose.py --build-dir . --backend tensorflow1 --backend tensorflow2 --repoagent checksum
```
deadeyegoodwin, line 57: Add --enable-gpu and remove --build-dir

compose.py
jbkyang-nvi: @GuanLuo can you enlighten me about what backends shouldn't be supported? We do need support for both CPU only and GPU only. Which backend is allowed for cloud storage?

qa/L0_infer/infer_test.py
jbkyang-nvi: Sorry this is to get the unit test run properly. I'll change this back

docs/compose.md
```diff
@@ -39,8 +39,32 @@ releases will increase the amount of customization that is available.
...
+```
+will provide a container `tritonserver-build` locally. You can access the container with
```
deadeyegoodwin, line 59: 'tritonserver'

qa/L0_infer/infer_test.py
```diff
@@ -39,8 +39,8 @@
     int(os.environ.get('TEST_SYSTEM_SHARED_MEMORY', 0)))
 TEST_CUDA_SHARED_MEMORY = bool(int(os.environ.get('TEST_CUDA_SHARED_MEMORY',
                                                   0)))
-CPU_ONLY = (os.environ.get('TRITON_SERVER_CPU_ONLY') is not None)
-
+#CPU_ONLY = (os.environ.get('TRITON_SERVER_CPU_ONLY') is not None)
```
deadeyegoodwin, line 42: I don't think we should be changing L0_infer test. We are going to need additional testing for "separate triton images for invidual backends" ticket and that should be sufficient for compose.py (at least for now). For "individual backend images" testing as part of TRITON_NIGHTLY CI runs we should invoke a new build step that uses compose.py to create the individual images. We will then need to build some kind of devel container  for each and run L0_infer appropriately.... but all those changes can be done for the other ticket in a different PR.

deadeyegoodwin (changes requested):

compose.py
GuanLuo: I think cloud storage is on the Triton core, and other backends mostly just have configuration on whether GPU should be supported. But the idea is that eventually we will need to have the Triton and other components be built in configuration as requested and compose with them instead of using the ones in the release container.

CoderHam (changes requested):

compose.py
```diff
@@ -0,0 +1,305 @@
+#!/usr/bin/env python3
+# Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
jbkyang-nvi: It doesn't wrap line for `#` started comments. I've removed the `(c)` for consistency

compose.py
jbkyang-nvi: Will change this for next round

compose.py
jbkyang-nvi: This logic was copied over from `container_build()` though. A bit confused why we don't need to do it here?

deadeyegoodwin: I believe the code you copied was for the case where we wanted to create an image by saving a container... and in that case you need to delete any existing image. In this case you are just running "docker build" and that will overwrite an existing image so no need to delete it first.

compose.py
deadeyegoodwin: Not sure what is being asked here but the compose.py will do no builds. It will just take the entire triton core from the release image and include it.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import argparse
+import os
+import subprocess
+import sys
+import docker
+
+### Global variables
+TRITON_VERSION_MAP = {
+    '2.11.0dev': (
```
deadeyegoodwin, line 35: Update to match what is in main branch now.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+        '21.06dev',  # triton container
+        '21.05')  # upstream container     
+}
+
+EXAMPLE_BACKENDS = ['identity', 'square', 'repeat']
```
deadeyegoodwin, line 40: It would be good to not have these backend and repoagents defines. Instead it seems like the docker build will just fail if user requests some backend that is not actually available in the full image.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+CORE_BACKENDS = ['tensorrt', 'ensemble']
+NONCORE_BACKENDS = [
+    'tensorflow1', 'tensorflow2', 'onnxruntime', 'python', 'dali', 'pytorch',
+    'openvino', 'fil'
+]
+EXAMPLE_REPOAGENTS = ['checksum']
+FLAGS = None
+
+DEPENDENCY_MAP = {'python': ('python3-pip'), 'pytorch': ('python3-dev')}
```
deadeyegoodwin, line 49: Is this used?

compose.py
```diff
@@ -0,0 +1,297 @@
...
+                          '--quiet',
+                          action="store_true",
+                          required=False,
+                          help='Disable console output.')
+    group_qv.add_argument('-v',
+                          '--verbose',
+                          action="store_true",
+                          required=False,
+                          help='Enable verbose output.')
+    parser.add_argument('--build-name',
```
deadeyegoodwin, line 214: There was a previous comment on this... Use --output-name and describe as "Name for the generated Docker image."

compose.py
```diff
@@ -0,0 +1,297 @@
...
+        required=False,
+        help=
+        'The Triton version. If not specified defaults to the value in the TRITON_VERSION file.'
+    )
+    parser.add_argument(
+        '--container-version',
+        type=str,
+        required=False,
+        help=
+        'The Triton container version to build. If not specified the container version will be chosen automatically based on --version value.'
```
deadeyegoodwin, line 237: "The version to use for the generated Docker image."
Don't say "build"

compose.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    parser.add_argument(
+        '--upstream-container-version',
+        type=str,
+        required=False,
+        help=
+        'The upstream container version to use for the build. If not specified the upstream container version will be chosen automatically based on --version value.'
```
deadeyegoodwin, line 244: "The version of the NGC Docker images to use as a base for this image. If not ..."

compose.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    parser.add_argument(
+        '--enable-gpu',
+        action="store_true",
+        required=True,
+        help=
+        'Flag to enable gpus for the build. Must be specified since only-cpu is not supported'
```
deadeyegoodwin, line 251: "Generate a Triton image that suppors GPUs." Don't put the rest just raise an appropriate error below.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    parser.add_argument(
+        '--backend',
+        action='append',
+        required=False,
+        help=
+        'Include specified backend in build as <backend-name>. To have multiple backends built, specify --backend <foo> --backend <bar>'
```
deadeyegoodwin, line 258: ".. in the generated Docker image. The flag may be specified multiple times."

compose.py
```diff
@@ -0,0 +1,297 @@
...
+    )
+    parser.add_argument(
+        '--repoagent',
+        action='append',
+        required=False,
+        help='Include specified repo agent in build as <repoagent-name>')
```
deadeyegoodwin, line 264: ".. in the generated Docker image. The flag may be specified multiple times."

compose.py
```diff
@@ -0,0 +1,297 @@
...
+    if FLAGS.backend is None:
+        FLAGS.backend = []
+    if FLAGS.repoagent is None:
+        FLAGS.repoagent = []
+
+    start_gpu_dockerfile(FLAGS.work_dir, dockerfile_name)
+    add_requested_backends(FLAGS.work_dir, dockerfile_name, FLAGS.backend)
+    add_requested_repoagents(FLAGS.work_dir, dockerfile_name, FLAGS.repoagent)
+    end_gpu_dockerfile(FLAGS.work_dir, dockerfile_name)
+    build_docker_container(FLAGS.work_dir, dockerfile_name, FLAGS.build_name)
```
deadeyegoodwin, line 297: build_docker_image

compose.py
```diff
@@ -0,0 +1,297 @@
...
+        help=
+        'Include specified backend in build as <backend-name>. To have multiple backends built, specify --backend <foo> --backend <bar>'
+    )
+    parser.add_argument(
+        '--repoagent',
+        action='append',
+        required=False,
+        help='Include specified repo agent in build as <repoagent-name>')
+    FLAGS = parser.parse_args()
+
```
deadeyegoodwin, line 266: Make sure FLAGS.enable_gpu is true.

compose.py
deadeyegoodwin: Min starts with nothing triton related. So the above copy over just the core parts of triton which are always required. The backends and repo-agents are the things that are included conditionally.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+    [ `id -g $TRITON_SERVER_USER` -eq 1000 ]
+
+WORKDIR /opt/tritonserver
+RUN rm -fr /opt/tritonserver/*
+COPY --chown=1000:1000 --from=full /opt/tritonserver/LICENSE .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/TRITON_VERSION .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/NVIDIA_Deep_Learning_Container_License.pdf .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/bin bin/
+COPY --chown=1000:1000 --from=full /opt/tritonserver/lib lib/
+COPY --chown=1000:1000 --from=full /opt/tritonserver/include/triton/core include/triton/core
```
deadeyegoodwin, line 116: Just copy the entire include directory.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+# Top-level include/core not copied so --chown does not set it correctly,
+# so explicit set on all of include
+RUN chown -R triton-server:triton-server include
```
deadeyegoodwin, line 119: You shouldn't need this if you copy the entire include directory, but you should check after removing it.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+'''
+    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
+        dfile.write(df)
+
+
+### add additional backends needed
+def add_requested_backends(ddir, dockerfile_name, backends):
+    df = "# Copying over backends \n"
+    for backend in backends:
+        if (backend.lower()
```
deadeyegoodwin, line 129: Just unconditionallly copy. Docker build will fail if the backend doesn't exist in full.

compose.py
```diff
@@ -0,0 +1,297 @@
...
+            log("Cannot create container from unsupported backend: " + backend)
+
+    with open(os.path.join(ddir, dockerfile_name), "a") as dfile:
+        dfile.write(df)
+
+
+def add_requested_repoagents(ddir, dockerfile_name, repoagents):
+    df = "#  Copying over repoagents \n"
+    for ra in repoagents:
+        if (ra in EXAMPLE_REPOAGENTS):
```
deadeyegoodwin, line 143: unconditional

compose.py
```diff
@@ -0,0 +1,297 @@
...
+        else:
+            log("Cannot create container from unsupported repoagent: " + ra)
+
+    with open(os.path.join(ddir, dockerfile_name), "a") as dfile:
+        dfile.write(df)
+
+
+
+# Install dependencies and run entrypoint script
+def end_gpu_dockerfile(ddir, dockerfile_name):
```
deadeyegoodwin, line 155: This needs to have all the steps from build.py:create_dockerfile_linux for the production stage (excluding the copying already done above). We don't want to replicate that code so you will need to refactor build.py to group together the non-copy parts of that function to the end and then extract them to a different function. Then compose.py can import build.py and call that function.

deadeyegoodwin (changes requested):

compose.py
jbkyang-nvi: As in, should we just not have this check? Instead, we'll just let the dockerbuild fail?

compose.py
jbkyang-nvi: No not right now

compose.py
jbkyang-nvi: What am I checking?

compose.py
deadeyegoodwin: Yes, that will require less maintenance on the script. Also, at some point we might need to support including a locally built backend into the container.

compose.py
jbkyang-nvi: Is there something I need to configure for each backend so it won't run in GPU? Is @GuanLuo suggesting changing the backend builds so that they are optimized for CPU only?

compose.py
deadeyegoodwin: If you remove this statement and copy the entire include/ dir above, then the owner:group for the entire include subdirectory contents should be user 1000

compose.py
deadeyegoodwin: We can't work on the CPU-only support yet. In general we may need CPU-only builds of the backends (so a separate full image that is built for cpu only)

compose.py
jbkyang-nvi: I specified `required=True` so It will always raise an error here.

compose.py
jbkyang-nvi: Thanks David. Do we need build_sha and build_id?

build.py
```diff
@@ -698,59 +751,23 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
-    # Add feature labels for SageMaker endpoint
-    if 'sagemaker' in endpoints:
-        df += '''
-LABEL com.amazonaws.sagemaker.capabilities.accept-bind-to-port=true
-COPY --chown=1000:1000 --from=tritonserver_build /workspace/build/sagemaker/serve /usr/bin/.
-'''
-
-    mkdir(ddir)
-    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
-        dfile.write(df)
```
jbkyang-nvi, line 753: moved this to the only linux build function

build.py
```diff
@@ -649,6 +649,59 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+'''
+            break
+
+    if len(repoagents) > 0:
+        df += '''
+COPY --chown=1000:1000 --from=tritonserver_build /tmp/tritonbuild/install/repoagents repoagents
+'''
+    df += dockerfile_add_installation_linux(argmap, backends)
+    
+    # Add feature labels for SageMaker endpoint
```
deadeyegoodwin, line 681: This sagemaker part needs to be in dockerfile_add_installation_linux

compose.py
```diff
@@ -0,0 +1,276 @@
...
+COPY --chown=1000:1000 --from=full /opt/tritonserver/TRITON_VERSION .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/NVIDIA_Deep_Learning_Container_License.pdf .
+COPY --chown=1000:1000 --from=full /opt/tritonserver/bin bin/
+COPY --chown=1000:1000 --from=full /opt/tritonserver/lib lib/
+COPY --chown=1000:1000 --from=full /opt/tritonserver/include include/
+'''
+    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
+        dfile.write(df)
+
+def start_cpu_dockerfile(ddir, dockerfile_name):
```
deadeyegoodwin, line 111: Remove, nothing CPU yet

compose.py
```diff
@@ -0,0 +1,276 @@
...
+    df = "#  Copying over repoagents \n"
+    for ra in repoagents:
+        df += '''COPY --chown=1000:1000 --from=full /opt/tritonserver/repoagents/{} /opt/tritonserver/repoagents/{}    
+'''.format(ra, ra)
+
+    with open(os.path.join(ddir, dockerfile_name), "a") as dfile:
+        dfile.write(df)
+
+
+def end_cpu_dockerfile(ddir, dockerfile_name):
```
deadeyegoodwin, line 138: remove

compose.py
```diff
@@ -0,0 +1,276 @@
...
+    df = ""
+    with open(os.path.join(ddir, dockerfile_name), "a") as dfile:
+        dfile.write(df)
+
+# Install dependencies and run entrypoint script
+def end_gpu_dockerfile(ddir, dockerfile_name, argmap, backends):
+    # TODO: remove libnvidia-ml-dev after 21.06 is launched
```
deadeyegoodwin, line 145: You should rebase and remove nvml stuff

compose.py
```diff
@@ -0,0 +1,276 @@
...
+        action='append',
+        required=False,
+        help='Include <repoagent-name> in the generated Docker image. The flag may be specified multiple times.')
+    FLAGS = parser.parse_args()
+    
+    if FLAGS.work_dir is None:
+        FLAGS.work_dir = "."
+    if FLAGS.output_name is None:
+        FLAGS.output_name = "tritonserver"
+    # Determine the versions. Start with Triton version, if --version
```
deadeyegoodwin, line 233: You only need a single version flag, --container-version, from which you determine the min and server container to pull from NGC. All the other version and build information you need to extract like you do for the build information below.

compose.py
```diff
@@ -0,0 +1,276 @@
...
+    if FLAGS.backend is None:
+        FLAGS.backend = []
+    if FLAGS.repoagent is None:
+        FLAGS.repoagent = []
+
+    start_gpu_dockerfile(FLAGS.work_dir, dockerfile_name)
+    add_requested_backends(FLAGS.work_dir, dockerfile_name, FLAGS.backend)
+    add_requested_repoagents(FLAGS.work_dir, dockerfile_name, FLAGS.repoagent)
+
+    upstreamDockerImage = 'nvcr.io/nvidia/tritonserver:{}-py3'.format(FLAGS.upstream_container_version)
```
deadeyegoodwin, line 259: Put this in a function (like create_argmap()). You will need to call it before start_gpu_dockerfile() because you need the version information in that function.

deadeyegoodwin (changes requested):

build.py
```diff
@@ -649,6 +649,53 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+    df += dockerfile_add_installation_linux(argmap, backends, endpoints)
+    
+    mkdir(ddir)
+    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
+        dfile.write(df)
+
+
+def dockerfile_add_installation_linux(argmap, backends, endpoints):
+    df='''
+ARG TRITON_VERSION={}
```
deadeyegoodwin, line 688: It's good that you put this version information in the common area shared by compose and build, but you also need to remove it from above (line 640-647)

build.py
```diff
@@ -692,59 +739,31 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+'''
+    df += '''
 ENV NVIDIA_BUILD_ID {}
 LABEL com.nvidia.build.id={}
 LABEL com.nvidia.build.ref={}
 '''.format(argmap['NVIDIA_BUILD_ID'], argmap['NVIDIA_BUILD_ID'],
            argmap['NVIDIA_BUILD_REF'])
 
-    # Add feature labels for SageMaker endpoint
+# Add feature labels for SageMaker endpoint
```
deadeyegoodwin, line 759: indent

build.py
```diff
@@ -649,6 +649,53 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+'''
+    df += dockerfile_add_installation_linux(argmap, backends, endpoints)
+    
+    mkdir(ddir)
+    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
+        dfile.write(df)
+
+
+def dockerfile_add_installation_linux(argmap, backends, endpoints):
+    df='''
```
deadeyegoodwin, line 687: Add description:
"""Common steps for production docker image, shared by build.py and compose.py"""

compose.py
```diff
@@ -0,0 +1,248 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import argparse
+import os
+import subprocess
+import sys
+import docker
+
+### Global variables
+TRITON_VERSION_MAP = {
```
deadeyegoodwin, line 34: Remove, you should use this from build.py when needed

compose.py
```diff
@@ -0,0 +1,248 @@
...
+##### create base image for gpu
+def start_gpu_dockerfile(ddir, argmap, dockerfile_name):
+    # Set enviroment variables
+    df = '''
+#
+# Multistage build.
+#
+FROM nvcr.io/nvidia/tritonserver:{}-py3 as full
+FROM nvcr.io/nvidia/tritonserver:{}-py3-min
+ARG TRITON_VERSION={}
```
deadeyegoodwin, line 75: Isn't this version information added by the common function shared with build.py? Why are you doing it again here?

compose.py
```diff
@@ -0,0 +1,248 @@
...
+
+    dockerfile_name = 'Dockerfile.compose'
+
+    if FLAGS.backend is None:
+        FLAGS.backend = []
+    if FLAGS.repoagent is None:
+        FLAGS.repoagent = []
+    if FLAGS.endpoint is None:
+        FLAGS.endpoint = []
+
```
deadeyegoodwin, line 242: I was wrong before in saying we needed a --container-version flag. We don't want that flag. Instead we must get the container version from the branch that we are on. You will need to use the logic from build.py  that get the triton version from the TRITON_VERSION file, and then looks up in build.py:TRITON_VERSION_MAP to get the container version.

docs/compose.md
```diff
@@ -37,10 +37,32 @@ functionality that you don't require.
...
+## Use the compose.py script
+
+The `compose.py` script uses the existing NGC Docker images 
+```
+nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
+nvcr.io/nvidia/tritonserver:<xx.yy>-py3
+```
+to create the new custom Triton Docker image.
+
+`compose.py` provides `--container-version`, `--backend`, `--repoagent` and `--endpoint` options that allow you to specify which upstream container to use and which backends, repository agents and endpoints to include in the custom image. The `--enable-gpu` flag indicates that you want to create an image that supports NVIDIA GPUs. For example, the following creates a new docker image that contains only the TensorFlow 1 and TensorFlow 2 backends and the checksum repository agent.
```
deadeyegoodwin, line 51: remove --container-version

docs/compose.md
```diff
@@ -37,10 +37,32 @@ functionality that you don't require.
...
-The two Docker images used for customization are retrieved using the
-following commands.
+## Use the compose.py script
+
+The `compose.py` script uses the existing NGC Docker images 
+```
+nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
+nvcr.io/nvidia/tritonserver:<xx.yy>-py3
+```
+to create the new custom Triton Docker image.
```
deadeyegoodwin, line 49: describe here how you need to "git clone" the release branch of the server repo that corresponds to the docker container release that you want to use to compose a new image

deadeyegoodwin (changes requested):

build.py
```diff
@@ -692,59 +739,31 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
-
     df += '''
 # Extra defensive wiring for CUDA Compat lib
-RUN ln -sf ${{_CUDA_COMPAT_PATH}}/lib.real ${{_CUDA_COMPAT_PATH}}/lib \
- && echo ${{_CUDA_COMPAT_PATH}}/lib > /etc/ld.so.conf.d/00-cuda-compat.conf \
+RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
+ && echo ${_CUDA_COMPAT_PATH}/lib > /etc/ld.so.conf.d/00-cuda-compat.conf \
  && ldconfig \
- && rm -f ${{_CUDA_COMPAT_PATH}}/lib
+ && rm -f ${_CUDA_COMPAT_PATH}/lib
```
GuanLuo, line 747: Is the character escaping not an issue anymore? If so, you can remove the double brackets from the environment variables above (line 691)

compose.py
```diff
@@ -0,0 +1,248 @@
...
+    argmap = create_argmap(FLAGS.container_version)
+    start_gpu_dockerfile(FLAGS.work_dir, argmap, dockerfile_name)
```
GuanLuo, line 244: Seems like we always create GPU enable image now, perhaps add a check on the flags above that returns error if `--enable-gpu` is not set. And remove that later when we add CPU-only image.

compose.py
jbkyang-nvi: yep need to delete. Thanks for catching

build.py
jbkyang-nvi: There's a problem when I keep the character escaping. Will remove the double brackets

compose.py
jbkyang-nvi: The `--enable-gpu` is enforced to always exist right now. I'll just add a check to make that more clear

docs/compose.md
jbkyang-nvi: Would it not be easier to ask the user specify the upstream container version and if they don't specify, then fallback to branch specific version? This way, people don't have to clone the entire repo just for this script

deadeyegoodwin: We can make --container-version optional. The potential problem is that the branch they run from (e.g. "main") may have changes in compose.py that are not valid for the release container they target. So the default should be to work using TRITON_VERSION but it is ok to have --container-version flag to override if the user knows what they are doing.

build.py
```diff
@@ -637,8 +647,49 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+    if len(repoagents) > 0:
+        df += '''
+COPY --chown=1000:1000 --from=tritonserver_build /tmp/tritonbuild/install/repoagents repoagents
+'''
+    mkdir(ddir)
+    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
+        dfile.write(df)
+
+
+"""Common steps for production docker image, shared by build.py and compose.py"""
```
deadeyegoodwin, line 686: In python function comments go after the def, so this should be after line 689.

build.py
```diff
@@ -742,9 +766,7 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
 COPY --chown=1000:1000 --from=tritonserver_build /workspace/build/sagemaker/serve /usr/bin/.
```
deadeyegoodwin, line 766: There is no "tritonserver_build" when running from compose.py. So you likely need to move the sagemaker out of the common function and modify it as needed in compose.py.

compose.py
```diff
@@ -0,0 +1,268 @@
...
+        'Include <backend-name> in the generated Docker image. The flag may be specified multiple times.'
+    )
+    parser.add_argument(
+        '--repoagent',
+        action='append',
+        required=False,
+        help=
+        'Include <repoagent-name> in the generated Docker image. The flag may be specified multiple times.'
+    )
+    parser.add_argument(
```
deadeyegoodwin, line 224: Don't add --endpoint flag because what endpoints are included in Triton cannot be controlled by compose. For sagemaker you need to look into the server image and see if the /usr/bin/serve and the do the sagemaker setup correctly for composed image (this code will be in end_gpu_dockerfile)

compose.py
```diff
@@ -0,0 +1,268 @@
...
+        required=False,
+        help='Name for the generated Docker image. Default is "tritonserver".')
+    parser.add_argument(
+        '--work-dir',
+        type=str,
+        required=False,
+        help=
+        'Generated dockerfiles are placed here. Default to current directory.')
+    parser.add_argument(
+        '--upstream-container-version',
```
deadeyegoodwin, line 200: just --container-version not upstream-container-version

docs/compose.md
```diff
@@ -37,10 +37,27 @@ functionality that you don't require.
...
 releases will increase the amount of customization that is available.
 It is also possible to [build Triton](build.md#building-triton)
-yourself to get more exact customization.
+from source to get more exact customization.
 
-The two Docker images used for customization are retrieved using the
-following commands.
+## Use the compose.py script
+
+The `compose.py` script can be found in the [server repository](https://github.com/triton-inference-server/server). Simply clone the repository and run `compose.py` to create a custom container.
```
deadeyegoodwin, line 44: Need to help them know which branch to clone. They need to clone the release branch corresponding to the NGC release that they want to base their custom image on. For example they need to use r21.06 branch if they want to create an image based on the NGC 21.06 Triton release.

deadeyegoodwin (changes requested):

build.py
```diff
@@ -65,12 +65,12 @@
...
 #
 TRITON_VERSION_MAP = {
-    '2.12.0dev':
-      ('21.07dev',   # triton container
-       '21.06',      # upstream container
-       '1.8.0',      # ORT
-       '2021.2.200', # ORT OpenVINO
-       '2021.2')     # Standalone OpenVINO
+    '2.12.0dev': (
+        '21.06',  # triton container
```
deadeyegoodwin, line 69: 21.07dev

compose.py
```diff
@@ -0,0 +1,280 @@
...
+        capture_output=True,
+        text=True)
+    f = re.search("serve", p_find.stdout)
+    fail_if(p_find.returncode != 0, "Cannot search for 'serve' in /usr/bin")
+    argmap = {
+        'NVIDIA_BUILD_REF': p_sha.stdout.rstrip(),
+        'NVIDIA_BUILD_ID': p_build.stdout.rstrip(),
+        'TRITON_VERSION': version,
+        'TRITON_CONTAINER_VERSION': container_version,
+        'ENDPOINTS': f is not None,
```
deadeyegoodwin, line 173: Can you name this "SAGEMAKER_ENDPOINT" to be more descriptive

compose.py
```diff
@@ -0,0 +1,280 @@
...
+        type=str,
+        required=False,
+        help=
+        'Generated dockerfiles are placed here. Default to current directory.')
+    parser.add_argument(
+        '--container-version',
+        type=str,
+        required=False,
+        help=
+        'The version to use for the generated Docker image. If not specified the container version will be chosen automatically based on --version value.'
```
deadeyegoodwin, line 228: There is no --version flag, so change desc to be "... will be chosen automatically based on the repository branch."

compose.py
```diff
@@ -0,0 +1,280 @@
...
+        action='append',
+        required=False,
+        help=
+        'Include <repoagent-name> in the generated Docker image. The flag may be specified multiple times.'
+    )
+    parser.add_argument(
+        '--dry-run',
+        action="store_true",
+        required=False,
+        help='Only create Dockerfile, does not create a container.')
```
deadeyegoodwin, line 252: "Only create Dockerfile.compose, do not build the Docker image"

docs/compose.md
```diff
@@ -37,78 +37,28 @@ functionality that you don't require.
...
-$ docker pull nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
-$ docker pull nvcr.io/nvidia/tritonserver:<xx.yy>-py3
-```
-
-Where \<xx.yy\> is the version of Triton that you want to customize. The
-\<xx.yy\>-py3-min image is a minimal, base image that contains the CUDA,
-cuDNN, etc. dependencies that are required to run Triton. The
-\<xx.yy\>-py3 image contains the complete Triton with all options and
-backends.
+The `compose.py` script can be found in the [server repository](https://github.com/triton-inference-server/server). Simply clone the repository and run `compose.py` to create a custom container. Note created container version will depend on the branch that was cloned. For example branch [r21.06](https://github.com/triton-inference-server/server/tree/r21.06) is needed to create a image based on the NGC 21.06 Triton release.
```
deadeyegoodwin, line 44: "is needed to create" -> "should be used to create"

docs/compose.md
```diff
@@ -37,78 +37,28 @@ functionality that you don't require.
...
 
-### Minimum Triton
-
-To create an image containing the minimal possible Triton use the
-following multi-stage Dockerfile. As mentioned above the amount of
-customization currently available is limited. As a result the minimum
-Triton still contains both HTTP/REST and GRPC endpoints; S3, GCS and
-Azure Storage filesystem support; and the TensorRT backend.
+`compose.py` provides `--backend`, `--repoagent` and `--endpoint` options that allow you to specify which backends, repository agents and endpoints to include in the custom image. The `--enable-gpu` flag indicates that you want to create an image that supports NVIDIA GPUs. For example, the following creates a new docker image that contains only the TensorFlow 1 and TensorFlow 2 backends and the checksum repository agent.
```
deadeyegoodwin, line 46: Remove --endpoint

deadeyegoodwin (changes requested):

deadeyegoodwin (dismissed):

build.py
```diff
@@ -65,12 +65,12 @@
 # incorrectly load the other version of the openvino libraries.
 #
 TRITON_VERSION_MAP = {
-    '2.12.0dev':
-      ('21.07dev',   # triton container
-       '21.06',      # upstream container
-       '1.8.0',      # ORT
-       '2021.2.200', # ORT OpenVINO
-       '2021.2')     # Standalone OpenVINO
+    '2.12.0dev': (
```
CoderHam, line 68: format fixes

GuanLuo (approved):

docs/compose.md
```diff
@@ -117,12 +67,14 @@ backend](https://github.com/triton-inference-server/backend).  The
...
-directory is "./mkbackend", the following Dockerfile will create a
-Triton image that contains all the supported Triton backends plus your
+directory is "./mkbackend", adding the following to the Dockerfile `compose.py`
+created will create a Triton image that contains all the supported Triton backends plus your
 custom backend.
 
 ```
-FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3 as full
+FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3 as full # populated by compose.py
+...
```
CoderHam, line 76: Extra 
```
"```"
```

docs/compose.md
```diff
@@ -131,5 +83,5 @@ backend as part of the Dockerfile. Then use Docker to create the
 image.
 
 ```
-$ docker build -t tritonserver_custom .
+$ docker build -t tritonserver_custom -f Dockerfile.compose .
```
CoderHam, line 86: Remove $ as it is removed from other places

qa/L0_infer/install_and_test.sh
```diff
@@ -0,0 +1,40 @@
...
+    apt-get install -y --no-install-recommends \
+         python3 \
+         python3-pip && \
+pip3 install --upgrade pip
+# install client libraries
+pip3 install nvidia-pyindex
+pip3 install tritonclient[all]
+
+# Run the actual test
+./test.sh
```
CoderHam, line 40: You might want to use bash -x test.sh instead.

qa/L0_infer/install_and_test.sh
```diff
@@ -0,0 +1,40 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Note: This script is to be used with triton containers that need
```
CoderHam, line 28: with _customized_ Triton containers

CoderHam (changes requested):

build.py
jbkyang-nvi: This is with the `format.py` formatting...

docs/compose.md
jbkyang-nvi: these are '...' 
like dot dot dot

docs/compose.md
jbkyang-nvi: Other stuff is from the Dockerfiles. Would it be more clear if we have the `$` to indicate its a bash command?

docs/compose.md
```diff
@@ -73,7 +73,7 @@ custom backend.
 
 ```
 FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3 as full # populated by compose.py
-...
+... # Other compose commands
```
CoderHam, line 76: Still looks like a typo here

CoderHam (changes requested):

docs/compose.md
CoderHam: I don't think it is necessary to have $. It should be clear. In fact you removed it [elsewhere](https://github.com/triton-inference-server/server/pull/3039/files/e6b3371694294fb60ff21385bafd3bbd391348c1#diff-13411a91b1a348eab8a65b80e6d6709ca1530ab8217d53df8b90fbb07552eb8cR54) in this PR

GuanLuo: I think then we should add back the `$` in elsewhere, the convention is using `$` to indicate a bash command (see other places in docs)

CoderHam (approved):

tanmayv25 (dismissed):

compose.py
```diff
@@ -0,0 +1,275 @@
+#!/usr/bin/env python3
+# Copyright 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 2: Does this need to be just 2021?

docs/compose.md
```diff
@@ -37,78 +37,28 @@ functionality that you don't require.
 Currently the customization is limited as described below but future
```
CoderHam, line 37: Update copyright, add affiliates

CoderHam (changes requested):
Minor copyright change else LGTM. No need to re-run CI for these fixes

docs/compose.md
CoderHam: Missed this

CoderHam (dismissed):

deadeyegoodwin (dismissed):

build.py
```diff
@@ -77,7 +77,7 @@
 CORE_BACKENDS = ['tensorrt', 'ensemble']
 NONCORE_BACKENDS = [
     'tensorflow1', 'tensorflow2', 'onnxruntime', 'python', 'dali', 'pytorch',
-    'openvino', 'fil', 'fastertransformer'
```
deadeyegoodwin, line 80: It looks like  you are incorrectly reverting this earlier change: https://github.com/triton-inference-server/server/pull/3077

deadeyegoodwin (changes requested):

build.py
jbkyang-nvi: Must have been a rebase issue. Thanks for pointing this out

deadeyegoodwin (dismissed):

CoderHam (changes requested):

build.py
```diff
@@ -638,19 +651,60 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
...
+COPY --chown=1000:1000 --from=tritonserver_build /workspace/build/sagemaker/serve /usr/bin/.
+'''
+    mkdir(ddir)
+    with open(os.path.join(ddir, dockerfile_name), "w") as dfile:
+        dfile.write(df)
+
+
+def dockerfile_add_installation_linux(argmap, backends):
+    # Common steps for production docker image, shared by build.py and compose.py
+    # set enviroment variables and TRITON_SERVER_USER so should be added earlier
```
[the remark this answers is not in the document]
CoderHam: Was the second line not needed then?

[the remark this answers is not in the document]
jbkyang-nvi: I don't think it is very clear. Do you think it was useful?

CoderHam (dismissed):

CoderHam (dismissed):

deadeyegoodwin (dismissed):

CoderHam (dismissed):

docs/compose.md
```diff
@@ -41,7 +41,7 @@ from source to get more exact customization.
 
 ## Use the compose.py script
 
-The `compose.py` script can be found in the [server repository](https://github.com/triton-inference-server/server). Simply clone the repository and run `compose.py` to create a custom container. Note created container version will depend on the branch that was cloned. For example branch [r21.06](https://github.com/triton-inference-server/server/tree/r21.06) should be used to create a image based on the NGC 21.06 Triton release. 
+The `compose.py` script can be found in the [server repository](https://github.com/triton-inference-server/server). Simply clone the repository and run `compose.py` to create a custom container. Note created container version will depend on the branch that was cloned. For example branch [r21.08](https://github.com/triton-inference-server/server/tree/r21.06) should be used to create a image based on the NGC 21.08 Triton release.
```
CoderHam, line 44: Arent the doc updates to reflect 21.08 does during the 21.08 codefreeze? These should be 21.07 for now

docs/compose.md
```diff
@@ -54,6 +54,7 @@ will provide a container `tritonserver` locally. You can access the container wi
 $ docker run -it tritonserver:latest
 ```
 
+Note: If `compose.py` is run on release versions older than r21.08, the resulting container will have DCGM version 2.2.8 installed. This may result in different GPU statistic reporting behavior.
```
CoderHam, line 57: Same here. This should be added during codefreeze. Additionally won't 21.07 have 2.2.3 with your latest changes?

CoderHam (changes requested):

docs/compose.md
jbkyang-nvi: This is not going to be merged in before the 21.07 release.. lets change it back to 21.06.

compose.py
```diff
@@ -0,0 +1,350 @@
...
+    fail_if(
+        len(container_version) == 0,
+        'docker inspect to find triton container version failed, {}'.format(
+            vars))
+
+    vars = p_version.stdout
+    dcgm_ver = re.search("DCGM_VERSION=([\S]{4,}) ", vars)
+    dcgm_version = ""
+    if dcgm_ver == None:
+        dcgm_version = "2.2.3"
```
CoderHam, line 189: Why was this changed back to 2.2.3?

CoderHam (changes requested):

compose.py
jbkyang-nvi: Since the two versions that matter -- 21.06 and 21.07 will have 2.2.3, we should just use that for the older versions. 21.08 will have 2.2.8,

jbkyang-nvi: Note the buildbase is the only time we will need the patch anyway, so this should work (and additionally make the composed container have the same version of dcgm as 21.06 and 21.07

CoderHam (approved):

PR #3047, opened 2021-06-24
Update TRT model gen to support TRT 8

CoderHam:


CoderHam:
Tested with the latest TRT container (with 8.0.1.6) and the models were generated correctly. 
cc @msalehiNV

deadeyegoodwin (commented):
Looks good but not approving for now because this must go into the r21.07 branch, not main.

CoderHam:
Right. I'll share the destination branch once r21.07 is ready
cc @dzier

CoderHam:
> Right. I'll share the destination branch once r21.07 is ready
> cc @dzier

Done

CoderHam:
Closing since already in 21.07

PR #3049, opened 2021-06-25
Add test for host policy. Fix edge case

GuanLuo:


GuanLuo:
Will create follow up PR to update L0_io when we apply the same to other backends.

src/backends/backend/triton_backend_manager.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Remove the "(c)" to avoid this line wrap... it will cause L0_copyright to fail

src/backends/backend/triton_model_instance.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Remove the "(c)" to avoid this line wrap... it will cause L0_copyright to fail

deadeyegoodwin (changes requested):

deadeyegoodwin (dismissed):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

src/core/infer_request.cc
```diff
@@ -808,7 +808,7 @@ InferenceRequest::Input::SetIsShapeTensor(const bool is_shape_tensor)
 }
```
CoderHam, line 808: Please update copyright, and remaining source files that require updates

CoderHam (changes requested):

CoderHam (approved):

deadeyegoodwin (approved):

PR #3051, opened 2021-06-25
WIP: Expand device_to_runner_map for DLA

CoderHam:
- Fixes TRT models to run concurrently on different DLA cores

### DIFF
--- a/src/backends/tensorrt/plan_backend.cc
+++ b/src/backends/tensorrt/plan_backend.cc
@@ -280,7 +280,7 @@ PlanBackend::CreateExecutionContexts(
 
   // Only need to map device to runner when creating contexts, after that,
   // only runner idx is needed.
-  std::map<int, size_t> device_to_runner_map;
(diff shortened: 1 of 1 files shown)


src/backends/tensorrt/plan_backend.cc
```diff
@@ -280,7 +280,7 @@ PlanBackend::CreateExecutionContexts(
 
   // Only need to map device to runner when creating contexts, after that,
   // only runner idx is needed.
-  std::map<int, size_t> device_to_runner_map;
+  std::map<std::pair<int, int64_t>, size_t> device_to_runner_map;
```
tanmayv25, line 283: Can you add a comment here about using the pair? It will help me when making changes to the TRT backend repo. Basically a pair of primary device and secondary device will compose the **"device"** which will have their own context.

CoderHam: Done, marked as WIP for now till other iGPU issues are resolved.

CoderHam:
No longer needed after changes in #3067

PR #3057, opened 2021-06-28
Integrate SequenceBatcher with new threading abstraction

tanmayv25:
There is a big performance slump.

**Config:**
model: simple_savedmodel_sequence_object
instance count: 1
max_batch_size: 8
sequence_batching {
  max_sequence_idle_microseconds: 5000000
}
concurrent sequences: 4
request concurrencies: 5

**New SequenceBatcher:**
Concurrency: 5, throughput: 1830 infer/sec, latency 2701 usec
Avg request latency: 2543 usec (overhead 91 usec + **queue 2008 usec** + compute input 32 usec + compute infer 403 usec + compute output 9 usec)

**Current SequenceBatcher**
Concurrency: 5, throughput: 3619 infer/sec, latency 1348 usec
Avg request latency: 1171 usec (overhead 178 usec + **queue 512 usec** + compute input 37 usec + compute infer 431 usec + compute output 13 usec)

The nsight traces  show some serious missed notification. I will address the perf slump as a part of another ticket. Let me know if something looks seriously flawed.

### DIFF
--- a/src/core/backend.cc
+++ b/src/core/backend.cc
@@ -143,25 +143,21 @@ InferenceBackend::SetConfiguredScheduler(void* model)
   // If 'sequence_batching' is configured use the SequenceBatchScheduler,
   // otherwise use the default DynamicBatchScheduler.
   if (config_.has_sequence_batching()) {
-    return Status(
-        Status::Code::INTERNAL, "SequenceBatcher is not yet supported");
-    /*
     // Sequence batcher
     RETURN_IF_ERROR(SequenceBatchScheduler::Create(
-        config_, runner_cnt, OnInit, OnWarmup, OnRun,
-        enforce_equal_shape_tensors, &scheduler));
-    */
+        static_cast<TritonModel*>(model), enforce_equal_shape_tensors,
+        &scheduler));
   } else if (config_.has_dynamic_batching()) {
     // Dynamic batcher
     RETURN_IF_ERROR(DynamicBatchScheduler::Create(
-        static_cast<TritonModel*>(model), GetCpuNiceLevel(config_),
+        static_cast<TritonModel*>(model), nullptr, GetCpuNiceLevel(config_),
         true /* dynamic_batching_enabled */, config_.max_batch_size(),
         enforce_equal_shape_tensors, config_.dynamic_batching(), &scheduler));
   } else {
     // Default scheduler. Use dynamic batch scheduler (with batching
     // disabled) as the default scheduler.
     RETURN_IF_ERROR(DynamicBatchScheduler::Create(
-        static_cast<TritonModel*>(model), GetCpuNiceLevel(config_),
+        static_cast<TritonModel*>(model), nullptr, GetCpuNiceLevel(config_),
         false /* dynamic_batching_enabled */, 1 /* max_batch_size */,
         std::unordered_map<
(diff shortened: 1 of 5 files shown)


src/core/backend.cc
```diff
@@ -143,25 +143,21 @@ InferenceBackend::SetConfiguredScheduler(void* model)
...
-        config_, runner_cnt, OnInit, OnWarmup, OnRun,
-        enforce_equal_shape_tensors, &scheduler));
-    */
+        static_cast<TritonModel*>(model), enforce_equal_shape_tensors,
+        &scheduler));
   } else if (config_.has_dynamic_batching()) {
     // Dynamic batcher
     RETURN_IF_ERROR(DynamicBatchScheduler::Create(
-        static_cast<TritonModel*>(model), GetCpuNiceLevel(config_),
+        static_cast<TritonModel*>(model), nullptr, GetCpuNiceLevel(config_),
```
deadeyegoodwin, line 153: We shouldn't use nice

src/core/backend.cc
```diff
@@ -143,25 +143,21 @@ InferenceBackend::SetConfiguredScheduler(void* model)
...
         true /* dynamic_batching_enabled */, config_.max_batch_size(),
         enforce_equal_shape_tensors, config_.dynamic_batching(), &scheduler));
   } else {
     // Default scheduler. Use dynamic batch scheduler (with batching
     // disabled) as the default scheduler.
     RETURN_IF_ERROR(DynamicBatchScheduler::Create(
-        static_cast<TritonModel*>(model), GetCpuNiceLevel(config_),
+        static_cast<TritonModel*>(model), nullptr, GetCpuNiceLevel(config_),
```
deadeyegoodwin, line 160: no nice

src/core/sequence_batch_scheduler.cc
```diff
@@ -1150,7 +1128,7 @@ DirectSequenceBatch::SchedulerThread(
             // just use zero for that.
             SetControlTensors(
                 ni, seq_slot, 0 /* corrid */, true /* not_ready */);
-            requests.emplace_back(std::move(ni));
+            curr_payload_->AddRequest(std::move(ni));
```
deadeyegoodwin, line 1131: Does AddRequest take a lock or anything similar? The Direct sequence batcher should be able to just form the entire batch ("requests") and then give the whole batch to the RL and not have to update it.

deadeyegoodwin (changes requested):

src/core/sequence_batch_scheduler.cc
tanmayv25: No. The `AddRequest` is a function which is specific to Payload class. Only one thread is building the requests for a payload at a time. Hence, no need to lock. The curr_payload_ object exists for each DirectSequenceBatch object. 

The Payload interface was designed in a way to accommodate the requirement of DB where the payload can be grown even after enqueueing to RL. I have kept the interface same here for simplicity.  For DirectSequenceBatch, we update the curr_payload_ as soon as the previous one is enqueued to the RL. Instead of local `std::vector<std::unique_ptr<InferenceRequest>> requests;` we build the request vector directly in the payload.

deadeyegoodwin (approved):

PR #3059, opened 2021-06-28
Add documentation for making custom min container

jbkyang-nvi:
Addressing /github.com/triton-inference-server/server/issues/2910

### DIFF
--- a/docs/compose.md
+++ b/docs/compose.md
@@ -67,8 +67,8 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
 COPY --from=full /opt/tritonserver/bin /opt/tritonserver/bin
 COPY --from=full /opt/tritonserver/lib /opt/tritonserver/lib
 ```
-Then install dependencies outlined in [build.py](https://github.com/triton-inference-server/server/blob/main/build.py#L670-L694)
-
+Then install dependencies outlined in [build.py](https://github.com/triton-inference-server/server/blob/main/build.py), `dockerfile_add_installation_linux()` function.
+ 
 Example Dockerfile that only uses `Tensorflow 1` with the 21.06 build:
 ```
 FROM nvcr.io/nvidia/tritonserver:21.06-py3 as full
--- a/docs/compose.md
+++ b/docs/compose.md
@@ -67,11 +67,33 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
 COPY --from=full /opt/tritonserver/bin /opt/tritonserver/bin
 COPY --from=full /opt/tritonserver/lib /opt/tritonserver/lib
 ```
+Then install dependencies outlined in [build.py](https://github.com/triton-inference-server/server/blob/main/build.py#L670-L694)
 
+Example Dockerfile that only uses `Tensorflow 1` with the 21.06 build:
+```
+FROM nvcr.io/nvidia/tritonserver:21.06-py3 as full
+FROM nvcr.io/nvidia/tritonserver:21.06-py3-min
+COPY --from=full /opt/tritonserver/bin /opt/tritonserver/bin
+COPY --from=full /opt/tritonserver/lib /opt/tritonserver/lib
+COPY --from=full /opt/tritonserver/backends/tensorflow1 /opt/tritonserver/backends/tensorflow1
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+         software-properties-common \
+         libb64-0d \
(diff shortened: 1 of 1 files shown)


docs/compose.md
```diff
@@ -67,11 +67,33 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
...
+         software-properties-common \
+         libb64-0d \
+         libre2-5 && \
+    rm -rf /var/lib/apt/lists/*
+RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin \
+&& mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600 \
+&& apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub \
+&& add-apt-repository "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /"
+RUN apt-get update \
+&& apt-get install -y datacenter-gpu-manager
```
CoderHam, line 91: Fix indents here and above

docs/compose.md
```diff
@@ -67,11 +67,33 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
...
+```
 Then use Docker to create the image.
 
 ```
-$ docker build -t tritonserver_min .
+$ docker build -t tritonserver_min -f <Dockerfile name> .
```
CoderHam, line 96: Why is ":-f <Dockerfile name>" needed?

CoderHam (changes requested):

docs/compose.md
```diff
@@ -67,11 +67,33 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
 COPY --from=full /opt/tritonserver/bin /opt/tritonserver/bin
 COPY --from=full /opt/tritonserver/lib /opt/tritonserver/lib
 ```
+Then install dependencies outlined in [build.py](https://github.com/triton-inference-server/server/blob/main/build.py#L670-L694)
```
deadeyegoodwin, line 70: Refer to function name instead of line numbers since the line numbers will quickly become out of date.

deadeyegoodwin (changes requested):

docs/compose.md
jbkyang-nvi: `Dockerbuild` assumes default dockerfile name to be `PATH/Dockerfile` so we want to specify here

docs/compose.md
jbkyang-nvi: did you want me to indent everything? Not sure if this matters?

docs/compose.md
```diff
@@ -67,11 +67,33 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
 COPY --from=full /opt/tritonserver/bin /opt/tritonserver/bin
 COPY --from=full /opt/tritonserver/lib /opt/tritonserver/lib
 ```
-
+Then install dependencies outlined in [build.py](https://github.com/triton-inference-server/server/blob/main/build.py), `dockerfile_add_installation_linux()` function.
+ 
+Example Dockerfile that only uses `Tensorflow 1` with the 21.06 build:
```
deadeyegoodwin, line 72: Example Dockerfile that creates a Triton image that contains only the TensorFlow 1 backend.

docs/compose.md
```diff
@@ -67,11 +67,33 @@ FROM nvcr.io/nvidia/tritonserver:<xx.yy>-py3-min
...
+```
+FROM nvcr.io/nvidia/tritonserver:21.06-py3 as full
+FROM nvcr.io/nvidia/tritonserver:21.06-py3-min
+COPY --from=full /opt/tritonserver/bin /opt/tritonserver/bin
```
deadeyegoodwin, line 76: I don't like this because it is not complete. I think we should stop documenting the manual way of doing compose and instead just change this documentation to only refer to the compose.py script and how to use it.

deadeyegoodwin (changes requested):

deadeyegoodwin:
I wasn't clear in my earlier comment. I would not make any of these changes at all. As part of your compose.py change you should update the documentation so that it only talks about using compose.py and not about doing it yourself. One thing we can do is add a --dryrun flag to compose.py that stops after generating the dockerfile, so that anyone that wants to use that as a starting point to add additional changes can use that flag and then modify the generated dockerfile.

jbkyang-nvi:
all changes will be addressed in https://github.com/triton-inference-server/server/pull/3039

PR #3063, opened 2021-06-30
Fix test-util build to use correct BACKEND_REPO_TAG

deadeyegoodwin:


build/test-util/CMakeLists.txt
```diff
@@ -57,9 +57,19 @@ FetchContent_Declare(
   GIT_TAG ${TRITON_CORE_REPO_TAG}
   GIT_SHALLOW ON
 )
+FetchContent_Declare(
```
CoderHam, line 60: We do this inside the test/example backends. I think we can remove this change and keep the one in CMakeLists.txt

CoderHam (changes requested):

CoderHam (approved):

PR #3067, opened 2021-06-30
Fix copy of tensors for TRT on iGPU 

CoderHam:


src/backends/tensorrt/plan_backend.cc
```diff
@@ -3117,14 +3165,23 @@ PlanBackend::Context::Run(
...
-          static_cast<const char*>(io_binding_info.device_buffer_),
-          TRITONSERVER_MEMORY_GPU, gpu_device_);
+      // Use pinned memory address if zero copy is supported, else use device
+      // memory address.
+      if (zero_copy_support) {
+        LOG_VERBOSE(1) << "Normal ProcessTensor";
+        payload_->responder_->ProcessTensor(
+            name, io_binding_info.io_shape_mapping_.first, dt,
+            io_binding_info.io_shape_mapping_.second,
+            static_cast<const char*>(io_binding_info.buffer_),
```
deadeyegoodwin, line 3175: Only this line differs? The just conditionalize this argument on zero_copy_support instead of the entire call

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3024,6 +3054,12 @@ PlanBackend::Context::Run(
         output_copy_stream_, events_[next_set_].ready_for_output_, 0);
   }
 
+  bool zero_copy_support = false;
+  FAIL_ALL_AND_RETURN_IF_ERROR(
```
deadeyegoodwin, line 3058: Isn't this saved in the class, we shouldn't calculated it against for every request

deadeyegoodwin (changes requested):

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3161,13 +3206,26 @@ PlanBackend::Context::Run(
...
+        LOG_VERBOSE(1) << "cudaLaunchHostFunc ProcessTensor";
+        cudaStreamWaitEvent(stream_, events_[next_set_].ready_for_output_, 0);
+        processTensor_tuples_.push_back(
+            {name, dt, batchn_shape,
+             static_cast<const char*>(io_binding_info.buffer_),
+             io_binding_info.memory_type_, io_binding_info.memory_type_id_,
+             payload_->responder_.get()});
+        cudaLaunchHostFunc(
+            stream_, ProcessTensorCudaHost,
+            reinterpret_cast<void*>(&processTensor_tuples_.back()));
```
tanmayv25, line 3222: I would not pass the address of an element of vector to this function. The cudaLaunchHostFunc will be executed at a later point when ready_for_output_ event is recorded. The  addresses of processTensor_tuples_ are likely to get invalidated with growing tuples.

tanmayv25 (changes requested):

src/backends/tensorrt/plan_backend.h
```diff
@@ -417,10 +417,15 @@ class PlanBackend : public InferenceBackend {
...
-    // The aer Context::num_expected_bindings_ number of IOBindingInfo elements
-    // for copy stream
+    // There are Context::num_expected_bindings_ number of IOBindingInfo
+    // elements for copy stream
     std::vector<std::vector<IOBindingInfo>> io_binding_infos_;
 
+    std::vector<std::tuple<
+        std::string, inference::DataType, std::vector<int64_t>, const char*,
+        TRITONSERVER_MemoryType, int64_t, BackendResponder*>>
+        processTensor_tuples_;
```
tanmayv25, line 427: You also must not use the processTensor_tuples_ as a class member variable. The plan backend context can issue multiple requests. The tuples will be over ridden by the latest request issue. This can be a part of the payload.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3161,15 +3214,39 @@ PlanBackend::Context::Run(
...
+            static_cast<const char*>(io_binding_info.device_buffer_),
+            io_binding_info.memory_type_, io_binding_info.memory_type_id_);
+      }
     }
   }
+
+  if (zero_copy_support_) {
+    cudaStreamWaitEvent(stream_, events_[next_set_].output_ready_, 0);
+    LOG_ERROR << "cudaStreamWaitEvent for output_ready";
+    cudaStreamSynchronize(stream_);
```
tanmayv25, line 3246: I don't think you should be syncing here.

CoderHam: Fixed

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3044,6 +3041,21 @@ PlanBackend::Context::Run(
     nvinfer1::Dims dims;
     dims = citr->second.context_->getBindingDimensions(binding_index);
 
+    // Wait for model execution to process the output tensors with pinned memory
```
tanmayv25, line 3044: The comment here needs to be updated..  No need to discuss waiting here. It is already documented at the ProcessTensor below.

src/core/backend_context.h
```diff
@@ -138,11 +138,12 @@ class BackendResponder {
       const std::vector<std::unique_ptr<InferenceRequest>>& requests,
       std::vector<std::unique_ptr<InferenceResponse>>* responses,
       const int max_batch_size, const bool pinned_enabled, cudaStream_t stream,
-      cudaEvent_t event)
+      cudaEvent_t event, bool zero_copy_support = false)
```
tanmayv25, line 141: I understand how this completely simplifies the logic but it exposes the zero_copy_support to the backend utilities API. Hence, this will be further exposed to other backends. We must be careful with naming. zero_copy_support in generic context shows there are no copies involved.. However, there is still a "memcpy" being enqueued to the stream.

I am okay for now but we must reconsider the nomenclature here. Probably as part of my changes to update backend utilities again with the core backend_context.h.

src/core/cuda_utils.cc
```diff
@@ -87,7 +96,14 @@ CopyBuffer(
   // the src buffer is valid.
   if ((src_memory_type != TRITONSERVER_MEMORY_GPU) &&
       (dst_memory_type != TRITONSERVER_MEMORY_GPU)) {
-    memcpy(dst, src, byte_size);
+    if (zero_copy_support) {
+      auto params = new CopyParams(dst, src, byte_size);
+      cudaLaunchHostFunc(
+          cuda_stream, MemcpyHost, reinterpret_cast<void*>(params));
+      *cuda_used = true;
```
tanmayv25, line 103: This should take care of the synchronization at the end with Finalize() call.. Hence, I don't think we need any other synchronization steps.

tanmayv25 (changes requested):

src/core/cuda_utils.cc
CoderHam: Yes that is correct. I have remove the extra synchronization I had added.

src/core/backend_context.h
CoderHam: Yes I agree. While the generic feature is zero copy support, we are not leveraging this completely since we still need the copies to prepare the contiguous buffers. 

I will consider renaming it to integrated_map_host_memory_support_ to be more accurate?

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3044,6 +3041,21 @@ PlanBackend::Context::Run(
...
+    // address if zero-copy is supported. Otherwise use device memory address,
+    // so that the memory copies perform asynchronously and wait for model
+    // execution.
+    void* buffer = nullptr;
+    if (zero_copy_support_) {
+      buffer = io_binding_info.buffer_;
+      io_binding_info.memory_type_ = TRITONSERVER_MEMORY_CPU_PINNED;
```
GuanLuo, line 3051: You shouldn't overwrite anything in `io_binding_info` during run, this information should be able to set once for all during initialization. You may need to change the information that `io_binding_info` carries if needed, but I think what it holds for this change is the device pointer used for TRT enqueue and the buffer / mem type / id (pinned in this case) for process tensor.

src/core/backend_context.h
GuanLuo: From the responder's perspective, this variable is to determine whether a system `memcpy` is queued in the given copy stream, so maybe something like `copy_on_stream`

src/core/cuda_utils.cc
```diff
@@ -87,7 +96,14 @@ CopyBuffer(
   // the src buffer is valid.
   if ((src_memory_type != TRITONSERVER_MEMORY_GPU) &&
       (dst_memory_type != TRITONSERVER_MEMORY_GPU)) {
-    memcpy(dst, src, byte_size);
+    if (zero_copy_support) {
+      auto params = new CopyParams(dst, src, byte_size);
```
GuanLuo, line 100: Considering that the synchronization will be done in `Finalize()`, in other words, the responder object will be valid during the execution of the host function, you don't need to `new / delete` the params object: it can just be a member of responder and you pass the pointer to the member as usrp.

GuanLuo (changes requested):

src/core/backend_context.h
CoderHam: @GuanLuo I am open to that name as well. It is more accurate to the behavior

src/backends/tensorrt/plan_backend.cc
```diff
@@ -170,6 +170,12 @@ PlanBackend::Context::Context(
     events_[idx].timestamp_signal_ = nullptr;
   }
   support_batching_ = (max_batch_size != NO_BATCHING);
+
+  Status status = SupportsIntegratedZeroCopy(gpu_device_, &zero_copy_support_);
+  if (!status.IsOk()) {
+    LOG_ERROR << "Failed to check if zero copy is supported";
```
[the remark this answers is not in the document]
CoderHam: I think it might be useful to know why zero copy is not supported. That is why I retained the reason in the log.

src/core/cuda_utils.cc
CoderHam: Sure, I will try a build without the new/delete for params

src/backends/tensorrt/plan_backend.cc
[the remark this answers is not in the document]
GuanLuo: I don't think you have printed the reason in log. By the way, it may be better to mention the optimization for zero copy will be disabled (not a fatal error, just that the performance may be less optimized)

src/core/cuda_utils.cc
CoderHam: I verified that the new/delete is indeed needed.

src/backends/tensorrt/plan_backend.cc
CoderHam: Yes I had already set this earlier. These updates are not needed.

src/backends/tensorrt/plan_backend.cc
GuanLuo: ^^^ And the waiting is no longer the case: memcpy is async and sync in collector in zero-copy case, the same as the handling in the regular case

src/backends/tensorrt/plan_backend.cc
```diff
@@ -3044,6 +3047,17 @@ PlanBackend::Context::Run(
...
 
+    // Wait for model execution to process the output tensors with pinned memory
+    // address if zero-copy is supported. Otherwise use device memory address,
+    // so that the memory copies perform asynchronously and wait for model
+    // execution.
+    void* buffer = nullptr;
+    if (zero_copy_support_) {
+      buffer = io_binding_info.buffer_;
+    } else {
+      buffer = io_binding_info.device_buffer_;
```
GuanLuo, line 3058: You don't need this, just pass `io_binding_info.buffer_` to responder's ProcessTensor

src/core/backend_context.h
GuanLuo: ^^^ change the name here and elsewhere

src/core/cuda_utils.cc
GuanLuo: Why? Responder is deleted before host function returns?

src/core/cuda_utils.h
```diff
@@ -79,7 +79,8 @@ Status CopyBuffer(
     const int64_t src_memory_type_id,
     const TRITONSERVER_MemoryType dst_memory_type,
     const int64_t dst_memory_type_id, const size_t byte_size, const void* src,
-    void* dst, cudaStream_t cuda_stream, bool* cuda_used);
+    void* dst, cudaStream_t cuda_stream, bool* cuda_used,
+    bool zero_copy_support = false);
```
GuanLuo, line 83: Change name

GuanLuo (changes requested):

src/backends/tensorrt/plan_backend.cc
CoderHam: Doesn't it require to be device_buffer_ for non iGPU case?

GuanLuo: See what `buffer_` is set to in regular case

src/core/cuda_utils.cc
CoderHam: The lifecycle of the CopyParams is such that the object is destroyed before the memcpy call inside the cuda host function

GuanLuo: My mistake, it doesn't reside in responder

tanmayv25:
Looks good to me. Just address Guan's changes..

src/core/cuda_utils.h
CoderHam: Done

src/backends/tensorrt/plan_backend.cc
CoderHam: I see what you mean. Fixed it now

src/backends/tensorrt/plan_backend.cc
```diff
@@ -170,6 +170,14 @@ PlanBackend::Context::Context(
     events_[idx].timestamp_signal_ = nullptr;
   }
   support_batching_ = (max_batch_size != NO_BATCHING);
+
+  Status status = SupportsIntegratedZeroCopy(gpu_device_, &zero_copy_support_);
+  if ((!status.IsOk()) || (!zero_copy_support_)) {
```
GuanLuo, line 175: I thought you said you were going to print the error message if there is an error?

CoderHam: Ah right. Fixed now.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -172,11 +172,14 @@ PlanBackend::Context::Context(
   support_batching_ = (max_batch_size != NO_BATCHING);
 
   Status status = SupportsIntegratedZeroCopy(gpu_device_, &zero_copy_support_);
-  if ((!status.IsOk()) || (!zero_copy_support_)) {
-    LOG_INFO << "Zero copy optimization is disabled";
+  if (!status.IsOk()) {
+    LOG_ERROR << "Failed to check if zero copy is supported: "
+              << status.Message();
```
GuanLuo, line 177: Mention "Zero copy optimization is disabled" here as well, so user know this is not a fatal error.

src/backends/tensorrt/plan_backend.cc
```diff
@@ -172,11 +172,14 @@ PlanBackend::Context::Context(
...
     zero_copy_support_ = false;
-  } else {
+  } else if (zero_copy_support_) {
     LOG_INFO << "Zero copy optimization is enabled";
+  } else {
+    LOG_INFO << "Zero copy optimization is disabled";
```
GuanLuo, line 182: maybe changing logging level to verbose?

src/backends/tensorrt/plan_backend.cc
CoderHam: I prefer the failed to check. The wording itself makes it clear that it is just a check.

GuanLuo: Okay

GuanLuo (approved):

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #3071, opened 2021-07-01
Redirect test results to another file

Tabrizian:


deadeyegoodwin (approved):

CoderHam (changes requested):
We need to run the rest of the nightly tests to catch other cases. PS: L0_batcher_input will need this fix. There may be others as well.

Tabrizian:
I fixed the L0_batcher_input and started nightly.

deadeyegoodwin (approved):

PR #3072, opened 2021-07-01
increase precision to 9

outtoin:
When using classification API, it only gives 6-digit precision logit because `std::to_string` return 6-digit precision.
Some of our model is really sensitive, so we want to see more precise result on classification API result.
So add simple code in `src/servers/classification.cc` to make precision to 9-digit.

### DIFF
--- a/src/servers/classification.cc
+++ b/src/servers/classification.cc
@@ -28,11 +28,21 @@
 
 #include <algorithm>
 #include <numeric>
+#include <sstream>
 #include "src/servers/common.h"
 
 namespace nvidia { namespace inferenceserver {
 
 namespace {
+
+template <typename T>
+std::string to_string_with_precision(const T a_value, const int n = 9)
+{
+    std::ostringstream out;
+    out.precision(n);
+    out << std::fixed << a_value;
+    return out.str();
+}
 
 template <typename T>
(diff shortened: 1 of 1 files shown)


tanmayv25:
@deadeyegoodwin  I guess this change is ok, but one can argue how much precision is enough. Does it make sense where in classification API instead of string we return logit as FP32?

outtoin:
> Can you just add some scaling to your model to the values as a way to increase the significant digits returned by the classification.

So...should I need to fix this code...?

outtoin:
> Can you just add some scaling to your model to the values as a way to increase the significant digits returned by the classification.

@deadeyegoodwin So...should I need to fix this code...?

deadeyegoodwin:
> 
> 
> > Can you just add some scaling to your model to the values as a way to increase the significant digits returned by the classification.
> 
> @deadeyegoodwin So...how can I fix this code?

You will need to propose an enhancement to the classification extension  that adds an optional precision specifier and then make the corresponding changes in the code. https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_classification.md

Why not just multiply your logit values by 1000 in the model itself if you need to expose more decimal places of precision.

PR #3088, opened 2021-07-06
Repopulating request_id in the Stream Inference Response with errors

rakib-hasan:


src/servers/grpc_server.cc
```diff
@@ -3291,6 +3291,27 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
+    // since grpc drops msg if status is not ok,
+    // attach request id to the error message
+    // grpc_client needs to parse it appropriately
+    const char *id;
+    LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceResponseId(res_, &id),
+      "couldn't retrieve id for failed request");
+    TRITONSERVER_Error_Code err_code = TRITONSERVER_ErrorCode(err_);
+    std::string encoded_msg(TRITONSERVER_ErrorMessage(err_));
+    encoded_msg.append(" request_id = # ");
```
tanmayv25, line 3305: The request_id field is optional in request message. If id is empty, then the `request_id = #` should not be attached to the error message. See my comment below. You should not need to add such encoded messages.

src/servers/grpc_server.cc
```diff
@@ -3291,6 +3291,27 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
+    encoded_msg.append(id);
+    // encoded_msg looks like "<ErrorMessage> request_id = # <Request_ID>"
+    // TODO: Maybe a better way would be to send in json format.
```
tanmayv25, line 3308: Instead of encoding the id into the error message, why not populate the just the [`id`](https://github.com/triton-inference-server/common/blob/main/protobuf/grpc_service.proto#L743) field of [infer_response ](https://github.com/triton-inference-server/common/blob/main/protobuf/grpc_service.proto#L804) when returning erroneous response? The protobuf message is already there.
This will save client from unnecessary decoding logic.

In other words, in the case of error, server will return ModelStreamInferResponse with only error_message and infer_response.id populated.

tanmayv25 (changes requested):

tanmayv25:
@rakib-hasan  Can you also add the following after L3667 in `grpc_server.cc`?

```
// repopulate the id so that client knows which request failed.
response->mutable_infer_response()->set_id(request.id());
```

tanmayv25 (approved):

PR #3092, opened 2021-07-07
Exposing port 8000 for Gitlab CI

Leem0sh:
Hey, working around with Triton, Gitlab CI end to end testing. Sadly I'm not able to access the container from Gitlab CI bcs no port is exposed in the dockerfile. Exposing the port should be a solution bcs Trition is running healthy in the container and there is just no way of accessing it.

### DIFF
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -41,7 +41,7 @@ ARG TRITON_BACKEND_REPO_TAG=main
 ## into QA area.
 ############################################################################
 FROM ${BUILD_IMAGE} AS build
-
+EXPOSE 9000
 # Ensure apt-get won't prompt for selecting options
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


dzier:
We don't implicitly expose ports in the Dockerfile. You can explicitly expose the port when running the docker container with the `-p8000:8000` flag for example:

```
docker run ... -p8000:8000 ... -f Dockerfile.QA ...
```

See documentation on running the docker container with GPUS: https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md#run-on-system-with-gpus

Leem0sh:
> We don't implicitly expose ports in the Dockerfile. You can explicitly expose the port when running the docker container with the `-p8000:8000` flag for example:
> 
> ```
> docker run ... -p8000:8000 ... -f Dockerfile.QA ...
> ```
> 
> See documentation on running the docker container with GPUS: https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md#run-on-system-with-gpus

Hello! Yes, that is a solution but sadly not for a Gitlab CI as I'm trying to run the image as a service. Services needs to have ports preconfigured in dockerfile and Gitlab CI doesn't have the option for exposing explicit port. :/ 

Thanks anyway! :)

PR #3118, opened 2021-07-14
Remove platform checks when the backend is present.

dyastremsky:
Removing the platform checks when backend is present. This will allow people to put custom backends onto the server after building it, rather than building with the backend flags enabled.

### DIFF
--- a/src/core/model_config_utils.cc
+++ b/src/core/model_config_utils.cc
@@ -998,7 +998,8 @@ ValidateModelConfig(
         "'max_batch_size' must be non-negative value for " + config.name());
   }
 
-  if (!config.platform().empty()) {
+  //If backend is  empty, validate the platform
+  if (!config.platform().empty() && config.backend().empty()) {
     if (!expected_platform.empty() &&
         (config.platform() != expected_platform)) {
       return Status(
--- a/src/core/model_config_utils.cc
+++ b/src/core/model_config_utils.cc
@@ -998,7 +998,8 @@ ValidateModelConfig(
         "'max_batch_size' must be non-negative value for " + config.name());
   }
 
-  //If backend is  empty, validate the platform
+  // If backend is  empty, validate the platform.
+  // Otherwise, the platform validation should be performed by the backend, if necessary, as Triton may not know about the backend's requirements.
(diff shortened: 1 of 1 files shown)


src/core/model_config_utils.cc
```diff
@@ -998,7 +998,8 @@ ValidateModelConfig(
         "'max_batch_size' must be non-negative value for " + config.name());
   }
 
-  if (!config.platform().empty()) {
+  //If backend is  empty, validate the platform
```
GuanLuo, line 1001: explain why it is not needed if backend is set, something like "Otherwise, the platform validation should be performed by the backend if necessary as Triton may not know about the backend's requirements"

dyastremsky: Good catch. Added the explanation.

src/core/model_config_utils.cc
```diff
@@ -998,7 +998,9 @@ ValidateModelConfig(
         "'max_batch_size' must be non-negative value for " + config.name());
   }
 
-  if (!config.platform().empty()) {
+  // If backend is  empty, validate the platform.
+  // Otherwise, the platform validation should be performed by the backend, if necessary, as Triton may not know about the backend's requirements.
```
GuanLuo, line 1002: Format, too long for one line

src/core/model_config_utils.cc
```diff
@@ -998,7 +998,9 @@ ValidateModelConfig(
         "'max_batch_size' must be non-negative value for " + config.name());
   }
 
-  if (!config.platform().empty()) {
+  // If backend is  empty, validate the platform.
```
CoderHam, line 1001: Extra space between "is  empty"

CoderHam (changes requested):

src/core/model_config_utils.cc
dyastremsky: Thanks! Updated.

src/core/model_config_utils.cc
dyastremsky: Thank you! Removed the extra space.

CoderHam (approved):

tanmayv25 (approved):

PR #3125, opened 2021-07-15
Create long-running stress test

krishung5:
Different types of tests that needs to be added using stress_thread :
- [x] timeouts
- [x] batched & non-batched request (using nomodels & resnet models)
- [x] misbehaving clients


qa/L0_long_running_stress/nobatch_stress.py
```diff
@@ -0,0 +1,120 @@
+# Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 1: If this was added in 2021 it should just be 2021

qa/L0_long_running_stress/nobatch_stress.py
```diff
@@ -0,0 +1,120 @@
...
+                          httpclient.InferenceServerClient('localhost:8000')),
+                         (grpcclient,
+                          grpcclient.InferenceServerClient('localhost:8001')))
+
+    def _test_helper(self,
+                     client,
+                     model_name,
+                     input_name='INPUT0',
+                     output_name='OUTPUT0'):
+
```
CoderHam, line 58: Unneeded blank line

qa/L0_long_running_stress/sequence_stress.py
```diff
@@ -0,0 +1,610 @@
+# Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 1: Same here. Also rename to just stress.py

qa/L0_long_running_stress/sequence_stress.py
```diff
@@ -0,0 +1,610 @@
...
+                ) == expected, "{}: expected result {}, got {}".format(
+                    sequence_name, expected, int(result))
+            else:
+                assert result == expected, "{}: expected result {}, got {}".format(
+                    sequence_name, expected, result)
+    triton_client.stop_stream()
+
+
+def get_datatype(trial):
+    # Get the datatype to use based on what models are available (see test.sh)
```
CoderHam, line 172: Remove (see test.sh)

qa/L0_long_running_stress/test.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+# Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
CoderHam, line 2: Here as well

qa/L0_long_running_stress/test.sh
```diff
@@ -0,0 +1,124 @@
...
+
+
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+for TARGET in graphdef savedmodel onnx libtorch plan python; do
```
CoderHam, line 91: I'd store "graphdef savedmodel onnx libtorch plan python" in a variable called BACKENDS so that we can easily run the test for all or a subset of backends.

qa/L0_long_running_stress/test.sh
```diff
@@ -0,0 +1,124 @@
...
+        (cd all_models/custom_sequence_int32 && \
+            sed -i "s/max_sequence_idle_microseconds:.*/max_sequence_idle_microseconds: 1000000/" config.pbtxt && \
+            sed -i "s/^max_batch_size:.*/max_batch_size: 1/" config.pbtxt && \
+            sed -i "s/kind: KIND_GPU/kind: KIND_GPU\\ncount: 4/" config.pbtxt && \
+            sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 4/" config.pbtxt)
+done
+
+#Setup batched models
+
+
```
CoderHam, line 82: Extra empty lines

qa/L0_long_running_stress/test.sh
```diff
@@ -0,0 +1,124 @@
...
+for m in ../custom_models/custom_sequence_int32 ; do
+    cp -r $m all_models/ && \
+        (cd all_models/custom_sequence_int32 && \
+            sed -i "s/max_sequence_idle_microseconds:.*/max_sequence_idle_microseconds: 1000000/" config.pbtxt && \
+            sed -i "s/^max_batch_size:.*/max_batch_size: 1/" config.pbtxt && \
+            sed -i "s/kind: KIND_GPU/kind: KIND_GPU\\ncount: 4/" config.pbtxt && \
+            sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 4/" config.pbtxt)
+done
+
+#Setup batched models
```
CoderHam, line 80: Add  space after #

qa/L0_long_running_stress/test.sh
```diff
@@ -0,0 +1,124 @@
...
+    set -e
+done
+
+kill $SERVER_PID
+wait $SERVER_PID
+
+if [ $RET -eq 0 ]; then
+    echo -e "\n***\n*** Test Passed\n***"
+else
+    cat $CLIENT_LOG
```
CoderHam, line 120: Where is CLIENT_LOG defined?

CoderHam (changes requested):

qa/L0_long_running_stress/test.sh
CoderHam: Still needs to be fixed

qa/L0_long_running_stress/sequence_stress.sh
```diff
@@ -0,0 +1,36 @@
...
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
```
CoderHam, line 27: Keep naming symmetric: stress.sh -> stress.py

qa/L0_long_running_stress/stress.py
```diff
@@ -0,0 +1,610 @@
...
+FLAGS = None
+CORRELATION_ID_BLOCK_SIZE = 100
+DEFAULT_TIMEOUT_MS = 5000
+SEQUENCE_LENGTH_MEAN = 16
+SEQUENCE_LENGTH_STDEV = 8
+
+_thread_exceptions = []
+_thread_exceptions_mutex = threading.Lock()
+
+
```
CoderHam, line 58: We want the scripts to run for specific backends. We need to capture and condition on the BACKENDS env var here as well. Look at how this in done in L0_infer/infer_test.py

qa/L0_long_running_stress/nobatch_stress.py
```diff
@@ -0,0 +1,118 @@
...
+import math
+import unittest
+import numpy as np
+import test_util as tu
+import tritongrpcclient as grpcclient
+import tritonhttpclient as httpclient
+from tritonclientutils import np_to_triton_dtype, InferenceServerException
+
+
```
CoderHam, line 38: Similar here. Use BACKENDS env var

CoderHam (changes requested):

CoderHam:
@krishung5 can you also remove the draft status and mark this ready for review?

qa/L0_long_running_stress/nobatch_stress.py
krishung5: I use BACKENDS var in nobatch_stress.sh and iterate through all the backend in BACKENDS. Do I still need to add BACKENDS env var to this script?

CoderHam:
Yes. How else will the python script know to send requests only for the specific backends?

qa/L0_long_running_stress/stress.py
krishung5: Modified, please check the updated PR. Originally the test was only using custom model. Will add other trials if needed.

tanmayv25 (changes requested):
@krishung5  @CoderHam  I may not have complete understanding of the kind of stress test we are designing here but it looks like we want to test how our server is robust while serving various kinds of inference requests. If yes, then I don't understand why is it being called long-running test? It is little confusing. How about `L0_stress` or `L0_stress_infer`? 

Also, if what we want is to run multiple workload simultaneously, then I don't think this is the best infrastructure,
Does stress.sh, nobatch_stress.sh ad timeout_stress.sh will take same amount of time in their execution? If not, then I think the test is only running one or two of these scripts most of the time which may be kind of defeating the original purpose. Ideally, we would want shorter running jobs to repeat if one of these jobs have not completed their single execution.

I might have wrong understanding so please feel free to elaborate the use case and why my above suggestions might not make sense.

tanmayv25:
> Yes. How else will the python script know to send requests only for the specific backends?

I think her bash script is invoking the python subtests.

qa/L0_long_running_stress/stress.py
```diff
@@ -0,0 +1,612 @@
...
+                        type=int,
+                        required=False,
+                        help='Random seed.')
+    parser.add_argument('-t',
+                        '--concurrency',
+                        type=int,
+                        required=False,
+                        default=8,
+                        help='Request concurrency. Default is 8.')
+    parser.add_argument(
```
deadeyegoodwin, line 549: Change from using iterations to using time as a measure of how long to run the test.

qa/L0_long_running_stress/stress.py
```diff
@@ -0,0 +1,612 @@
...
+            # reproducible order off of the initial seed we will get
+            # reproducible results when given the same seed.
+            seed = np.random.randint(2**32)
+
+            # Each thread is reserved a block of correlation IDs or size
+            # CORRELATION_ID_BLOCK_SIZE
+            correlation_id_base = 1 + (idx * CORRELATION_ID_BLOCK_SIZE)
+
+            threads.append(
+                threading.Thread(target=stress_thread,
```
deadeyegoodwin, line 590: You should expand stress_thread to have more cases to cover different types of tests. For example, add some cases for timeouts, some cases for batched and non-batched requests (using nomodels but also using some real models like resnet). For misbehaving clients that crash you will need to run separate python scripts. But I would launch them as a separate case in stress_thread. You can then launch some clients that have bad behavior (like close connection, exit early, etc.). You can also launch some clients that start a transaction and then this script can kill them, etc. In general everything should be driven from this script.

deadeyegoodwin (changes requested):

qa/L0_long_running_stress/timeout_stress.sh
```diff
@@ -0,0 +1,96 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+TIMEOUT_CLIENT_LOG_BASE="./timeout_client"
+CLIENT_TIMEOUT_TEST_CPP=../clients/client_timeout_test
+
+# Provide too small a timeout and expect a failure.
+# Note, the custom_identity_int32 is configured with a delay of 3 sec.
+# Test request timeout in grpc synchronous inference
+$CLIENT_TIMEOUT_TEST_CPP -t 1000 -v -i grpc >> ${TIMEOUT_CLIENT_LOG_BASE}.c++.grpc_infer.log 2>&1
```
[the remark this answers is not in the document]
krishung5: Thank you for these great suggestions! I think the reason why this test is called long-running stress is because it will be extended to run for a longer time later. For the infrastructure, I think you are right that the test is only running one or two scripts most of the time because of their different execution time. David G just gave me the suggestions about how I should utilize stress_thread to cover more cases, so I think I will modify the infrastructure accordingly. Really appreciate your helpful comments!

qa/L0_long_running_stress/stress.py
krishung5: Will modify.

deadeyegoodwin (approved):

CoderHam (approved):

tanmayv25 (approved):

qa/L0_long_running_stress/test.sh
```diff
@@ -0,0 +1,164 @@
...
+SERVER=/opt/tritonserver/bin/tritonserver
+source ../common/util.sh
+
+RET=0
+
+# If BACKENDS not specified, set to all
+BACKENDS=${BACKENDS:="graphdef savedmodel custom"}
+export BACKENDS
+
+# If MODEL_TRIALS not specified set to 1 2 4
```
deadeyegoodwin, line 53: The goal of this stress test is to run as much different models as possible at the same time in different ways. For the sequence stuff you started with it is sufficient to keep just the "model2" set of models. Add to that resnet models for all the backends to create a single model repository with all those models. Then run stress.py one time and have it use all those models at the same time. We will run this on a multi-gpu system so we will also need to adjust the model configurations to have multiple instances... but we can do that with a followup change. 

There is no need to use "custom" model for this test. With the new backend API that testing is now covered by any of the other backends... so remove all "custom" stuff.

deadeyegoodwin (changes requested):

qa/L0_long_running_stress/crashing_client.py
```diff
@@ -0,0 +1,71 @@
...
+sys.path.append("../common")
+
+import numpy as np
+import multiprocessing
+import time
+import tritongrpcclient as grpcclient
+
+
+def crashing_client(triton_client):
+    model_name_ = "custom_identity_int32"
```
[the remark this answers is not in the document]
krishung5: I tested that it would ran for around 10 seconds so I was thinking it should have been terminated before completed. But running this in a loop should be better, thanks for the comment, will modify.

qa/L0_long_running_stress/test.sh
krishung5: I see. Will modify.

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #3140, opened 2021-07-21
Remove DCGM patch and enforce version from version map

jbkyang-nvi:
Remove DCGM patch and enforce version from version map


build.py
```diff
@@ -70,7 +70,8 @@
        '21.06',      # upstream container
        '1.8.0',      # ORT
        '2021.2.200', # ORT OpenVINO
-       '2021.2')     # Standalone OpenVINO
+       '2021.2',     # Standalone OpenVINO
+       '2.2.8')      # DCGM version
```
CoderHam, line 74: Need to update on L59

CoderHam (changes requested):

build.py
CoderHam: @dzier / @deadeyegoodwin since we are using a specific version of DCGM maybe we should add this to the release notes?

deadeyegoodwin: Good idea. Can you open a ticket as a reminder for this.

CoderHam: Done: DLIS-2710

deadeyegoodwin: @jbkyang-nvi  while you are in this code can you also combine 58 and 59 to just "Standalone OpenVINO version"

build.py
```diff
@@ -474,16 +475,21 @@ def dali_cmake_args():
 
 
 def install_dcgm_libraries():
+    dcgm_version = ''
+    if FLAGS.version not in TRITON_VERSION_MAP:
+        fail('unable to determine default repo-tag, container version not known for {}'.format(FLAGS.version))
```
deadeyegoodwin, line 480: "... DCGM version not know for ..."

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

CoderHam (approved):

PR #3143, opened 2021-07-22
Change DCGM version to release version

jbkyang-nvi:


CoderHam:
@jbkyang-nvi can you include DCGM in the PR title?

CoderHam:
Also the target branch should be 21.07 as I understand

build.py
```diff
@@ -65,13 +65,13 @@
...
 TRITON_VERSION_MAP = {
-    '2.13.0dev':
-      ('21.08dev',   # triton container
-       '21.06',      # upstream container
-       '1.8.0',      # ORT
-       '2021.2.200', # ORT OpenVINO
-       '2021.2',     # Standalone OpenVINO
-       '2.2.8')      # DCGM version
+    '2.13.0dev': (
+        '21.08dev',  # triton container
```
CoderHam, line 69: The target branch should be r21.07 as I understand
It is ok for us to move to 2.2.8 for 21.08, we need this change in 21.07. 
@deadeyegoodwin feel free to correct me

CoderHam (changes requested):

jbkyang-nvi:
> Also the target branch should be 21.07 as I understand

@dzier should I do both?

build.py
deadeyegoodwin: The revert to 2.2.3 and re-application of the patch is only needed on 21.07. We can keep the main branch as it is (i.e. with 2.2.8 and with no patch)

Issue #3119, opened 2021-07-15
Triton server crashes silently with invalid input data


NonStatic2014:
**Description**
Sending an invalid input json to the REST API interface can cause triton server crashes silently

**Triton Information**
21.05-py3 container pulled from nvidia, on a CPU-only machine

**To Reproduce**
1. Follow instruction [here](https://www.tensorflow.org/tfx/tutorials/serving/rest_simple) to train the mnist_fashion model
2. create a folder for the trained model and make sure layout is correct
3. Run triton: `docker run --rm -p8000:8000 -p8001:8001 -p8002:8002 -v$(pwd)/model:/model nvcr.io/nvidia/tritonserver:21.05-py3 tritonserver --model-repository=/model --log-verbose=1 --strict-model-config=false`
4. Command to send request: `curl -d @"<filename>.json" http://localhost:8000/v2/models/fashion/versions/1/infer -v`

Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).
* Trained Model: https://ufile.io/dde0s62m
* Good input json: https://ufile.io/inpve8py
* Bad input json (causing crash): https://ufile.io/0kfmjc2p

**Expected behavior**
Error message when receive invalid input.

deadeyegoodwin:
Thanks for the bug report and repo instructions. We will look into the issue.

PR #3158, opened 2021-07-27
Add check for input shape

krishung5:
Should fix the issue #3119

### DIFF
--- a/qa/L0_http/test.sh
+++ b/qa/L0_http/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -364,6 +364,34 @@ if [ `grep -c "\[2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32\]" ./curl.out` != "
     RET=1
 fi
 if [ `grep -c "\[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0\]" ./curl.out` != "1" ]; then
+    RET=1
+fi
+
+# send bad request with invalid json
+# the 'data' field misaligns with the 'shape' field of the input
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -d'{"inputs":[{"name":"INPUT0","datatype":"INT32","shape":[1,16],"data":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]}]}' localhost:8000/v2/models/simple/infer`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+if [ `grep -c "\{\"error\":\"Unable to parse 'data': Shape does not match true shape of 'data' field\"\}" ./curl.out` != "1" ]; then
+    RET=1
+fi
+
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -d'{"inputs":[{"name":"INPUT0","datatype":"INT32","shape":[1,16],"data":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18]}]}' localhost:8000/v2/models/simple/infer`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+if [ `grep -c "\{\"error\":\"Unable to parse 'data': Shape does not match true shape of 'data' field\"\}" ./curl.out` != "1" ]; then
     RET=1
 fi
 
--- a/qa/L0_http/test.sh
+++ b/qa/L0_http/test.sh
@@ -393,6 +393,7 @@ fi
 if [ `grep -c "\{\"error\":\"Unable to parse 'data': Shape does not match true shape of 'data' field\"\}" ./curl.out` != "1" ]; then
     RET=1
 fi
+
 # Check if the server is still working after the above bad requests
 rm -f ./curl.out
 set +e
--- a/qa/L0_http/test.sh
+++ b/qa/L0_http/test.sh
@@ -367,8 +367,7 @@ if [ `grep -c "\[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0\]" ./curl.out` != "1" ]; then
     RET=1
 fi
 
-# send bad request with invalid json
-# the 'data' field misaligns with the 'shape' field of the input
+# Send bad request where the 'data' field misaligns with the 'shape' field of the input
 rm -f ./curl.out
 set +e
 code=`curl -s -w %{http_code} -o ./curl.out -d'{"inputs":[{"name":"INPUT0","datatype":"INT32","shape":[1,16],"data":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15]}]}' localhost:8000/v2/models/simple/infer`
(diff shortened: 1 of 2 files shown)


src/servers/http_server.cc
```diff
@@ -801,6 +801,23 @@ ValidateOutputParameter(triton::common::TritonJson::Value& io)
   return nullptr;  // success
 }
 
+TRITONSERVER_Error*
+CheckInputShape(
```
CoderHam, line 805: You are checking only if the total number of elements is the same and not if each dimension is the same. Maybe rename to CheckInputElementCount or something that more clearly explains what this function is doing.

src/servers/http_server.cc
```diff
@@ -801,6 +801,23 @@ ValidateOutputParameter(triton::common::TritonJson::Value& io)
...
+    const std::vector<int64_t>& shape,
+    triton::common::TritonJson::Value& tensor_data)
+{
+  int shape_product = 1;
+  for (size_t i = 0; i < shape.size(); i++) {
+    shape_product *= shape[i];
+  }
+  if (shape_product != static_cast<int>(tensor_data.ArraySize())) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL, "Shape size is mismatched with data size");
```
CoderHam, line 815: Shape does not match true shape of "data" field.

CoderHam (changes requested):

src/servers/http_server.cc
```diff
@@ -801,6 +801,24 @@ ValidateOutputParameter(triton::common::TritonJson::Value& io)
...
+TRITONSERVER_Error*
+CheckInputElementCount(
+    const std::vector<int64_t>& shape,
+    triton::common::TritonJson::Value& tensor_data)
+{
+  int shape_product = 1;
+  for (size_t i = 0; i < shape.size(); i++) {
+    shape_product *= shape[i];
+  }
+  if (shape_product != static_cast<int>(tensor_data.ArraySize())) {
```
NonStatic2014, line 813: Convert it to `int` may loss precision?

src/servers/http_server.cc
```diff
@@ -801,6 +801,24 @@ ValidateOutputParameter(triton::common::TritonJson::Value& io)
...
 }
 
+TRITONSERVER_Error*
+CheckInputElementCount(
+    const std::vector<int64_t>& shape,
+    triton::common::TritonJson::Value& tensor_data)
+{
+  int shape_product = 1;
+  for (size_t i = 0; i < shape.size(); i++) {
+    shape_product *= shape[i];
```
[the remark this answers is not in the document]
NonStatic2014: Can you add UTs?

NonStatic2014 (changes requested):

src/servers/http_server.cc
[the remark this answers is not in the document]
GuanLuo: Using int64_t should be safe in terms of overflowing

src/servers/http_server.cc
GuanLuo: Note that [TensorData](https://github.com/kubeflow/kfserving/blob/master/docs/predict-api/v2/required_api.md#tensor-data) can be represent in two ways, so the function should check which representation it uses and compare the dimensions accordingly:

- if 1-D, then get element count of the shape and compare the value
- if > 1-D, then expect the number of dimensions is the same and iterate each dimension to ensure they have the same shape.

GuanLuo (changes requested):
Is the issue due to the content in "data" field misaligns with the "shape" field of the input? If so, it seems to be the same issue that this [FIXME](https://github.com/triton-inference-server/server/blob/afa2cd415bfdf35f21d3eab219a31e7a9ad4bf70/src/servers/http_server.cc#L2026-L2031) refers to. I think it is better to have a explicit check instead of what is suggested in the FIXME, but you should clean up the FIXME and related section as your change will address them.

src/servers/http_server.cc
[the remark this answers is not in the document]
NonStatic2014: but i will still suggest adding UTs to prevent regressions later.

src/servers/http_server.cc
CoderHam: Yes. Good catch @GuanLuo!

CoderHam (changes requested):

src/servers/http_server.cc
```diff
@@ -379,6 +381,13 @@ ReadDataFromJsonHelper(
         default:
           break;
       }
+      // Check if writing to 'serialized' is overrunning the expected byte_size
+      uint64_t bytes_read = *counter * TRITONSERVER_DataTypeByteSize(dtype);
```
GuanLuo, line 385: You don't need to compute the byte size on every element, in fact, you can just keep track of the expected total counter (expected number of elements) and compare with `*counter` directly, as `byte_size` is simply computed by `element_cnt * TRITONSERVER_DataTypeByteSize(dtype)`

From this perspective, note that the expected counter is slightly different from `TRITONSERVER_TYPE_BYTES`, which is just the byte size, so you need to condition to set it properly before calling into `ReadDataFromJson()`

GuanLuo (changes requested):

src/servers/http_server.cc
krishung5: I see. Script is modified, please review.

src/servers/http_server.cc
```diff
@@ -269,8 +270,15 @@ ReadDataFromJsonHelper(
     TRITONSERVER_Error* assert_err =
         el.AssertType(triton::common::TritonJson::ValueType::ARRAY);
     if (assert_err == nullptr) {
-      RETURN_IF_ERR(ReadDataFromJsonHelper(base, dtype, el, counter));
+      RETURN_IF_ERR(
+          ReadDataFromJsonHelper(base, dtype, el, counter, expected_cnt));
     } else {
+      // Check if writing to 'serialized' is overrunning the expected byte_size
+      if (*counter == expected_cnt) {
```
GuanLuo, line 277: Use >=, counter for `TRITONSERVER_TYPE_BYTES` can increment more than 1 value. You probably need to add a similar check after line 380 after you know how much more bytes are needed to store the value (i.e. need `*counter + len + sizeof(uint32_t) <= expected_cnt`)

GuanLuo (changes requested):
Otherwise looks good except for one more change. And @NonStatic2014 raised a point that we should add test coverage, testing with invalid request JSON, around it. You can add it to this PR or open a follow-up PR for that

src/servers/http_server.cc
krishung5: Modified. Test cases for invalid request JSON are added. Please reveiw.

GuanLuo (approved):

qa/L0_http/test.sh
```diff
@@ -367,6 +367,34 @@ if [ `grep -c "\[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0\]" ./curl.out` != "1" ]; then
     RET=1
 fi
 
+# send bad request with invalid json
```
CoderHam, line 370: The json itself is not invalid. Fix:
Send bad request where the 'data' field misaligns with the 'shape' field of the input

qa/L0_http/test.sh
```diff
@@ -367,6 +367,34 @@ if [ `grep -c "\[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0\]" ./curl.out` != "1" ]; then
...
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+if [ `grep -c "\{\"error\":\"Unable to parse 'data': Shape does not match true shape of 'data' field\"\}" ./curl.out` != "1" ]; then
+    RET=1
+fi
+
```
CoderHam, line 397: Add a check here to ensure the sever is running and can successful infer even after the above bad request.

src/servers/http_server.cc
```diff
@@ -411,10 +425,19 @@ ReadDataFromJson(
 
     default:
       RETURN_MSG_IF_ERR(
-          ReadDataFromJsonHelper(base, dtype, tensor_data, &counter),
+          ReadDataFromJsonHelper(
+              base, dtype, tensor_data, &counter, expected_cnt),
           "Unable to parse 'data'");
       break;
   }
+  // Check if 'ReadDataFromJsonHelper' reads
```
CoderHam, line 433: Run formatter

CoderHam (changes requested):

src/servers/http_server.cc
CoderHam: Also add empty line before this comment

krishung5: Updated

qa/L0_http/test.sh
```diff
@@ -394,6 +393,22 @@ fi
 if [ `grep -c "\{\"error\":\"Unable to parse 'data': Shape does not match true shape of 'data' field\"\}" ./curl.out` != "1" ]; then
     RET=1
 fi
+# Check if the server is still working after the above bad requests
```
CoderHam, line 396: Add empty line before comment here

CoderHam (changes requested):
Minor formatting else LGTM

qa/L0_http/test.sh
CoderHam: The check should be an additional successful inference -> not just checking the output from the current inference.

CoderHam (changes requested):

CoderHam (approved):

PR #3166, opened 2021-07-28
Extend L0_perf_resnet and L0_perf_nomodel to support jetson

CoderHam:
Replace use of OPTDIR with TRITON_DIR


qa/L0_perf_nomodel/test.sh
```diff
@@ -139,7 +141,13 @@ TEST_CONCURRENCY+=(
     1
     16
     16)
-TEST_BACKENDS="plan custom graphdef savedmodel onnx libtorch python"
+
+# Only TF, Onnx and TRT are supported on Jetson
+if [ "$TEST_JETSON" -eq 1 ]; then
+    TEST_BACKENDS="plan custom graphdef savedmodel onnx"
```
GuanLuo, line 147: Do we still have "custom"?

qa/L0_perf_nomodel/run_test.sh
```diff
@@ -42,26 +41,38 @@ TENSOR_SIZE=${TENSOR_SIZE:=1}
 SHARED_MEMORY=${SHARED_MEMORY:="none"}
 REPORTER=../common/reporter.py
 
-DATADIR=/data/inferenceserver/${REPO_VERSION}
+DATADIR=${DATADIR:="/data/inferenceserver/${REPO_VERSION}"}
 RESULTDIR=${RESULTDIR:=.}
 
-SERVER=/opt/tritonserver/bin/tritonserver
-SERVER_ARGS="--model-repository=`pwd`/models"
+OPTDIR=${OPTDIR:="/opt"}
```
GuanLuo, line 47: I think you should change this to be "TRITON_DIR" and default to `/opt/tritonserver`, which simply means the directory that is structured in the way you expect as below. Currently it feels to be in the middle, as the user can configure the path to the directory but not the name of the directory as well.

GuanLuo (changes requested):

qa/L0_perf_nomodel/test.sh
CoderHam: Yes. We use the new backend api for the identity / addsub models and refer to these as custom models

qa/L0_perf_nomodel/run_test.sh
CoderHam: Sure, making this change

qa/L0_perf_nomodel/run_test.sh
```diff
@@ -42,26 +41,38 @@ TENSOR_SIZE=${TENSOR_SIZE:=1}
...
+fi
+
 # Select the single GPU that will be available to the inference server
 export CUDA_VISIBLE_DEVICES=0
 
 mkdir -p ${RESULTDIR}
 RET=0
 
-cp /opt/tritonserver/backends/python/triton_python_backend_utils.py .
+if [ "$TEST_JETSON" -eq 0 ]; then
```
GuanLuo, line 67: Check if the backends are under test and prepare the model accordingly instead of checking TEST_JETSON.

qa/L0_perf_nomodel/run_test.sh
```diff
@@ -42,26 +41,38 @@ TENSOR_SIZE=${TENSOR_SIZE:=1}
...
+BACKEND_DIR=${OPTDIR}/tritonserver/backends
+SERVER_ARGS="--model-repository=`pwd`/models --backend-directory=${BACKEND_DIR}" 
 source ../common/util.sh
 
+if [ "$TEST_JETSON" -eq 1 ]; then
+    ARCH="aarch64"
+    PERF_CLIENT=${OPTDIR}/tritonserver/clients/bin/perf_client
+else
+    ARCH="x86_64"
+    PERF_CLIENT=../clients/perf_client
```
GuanLuo, line 58: Personally I would make these variables (`ARCH` / `PERF_CLIENT`) to be configurable like `DATADIR`, and transfer the setting to top level where the test is configured in CI. So the script doesn't tie with a particular setup, think of adding yet another environment for testing, currently you will need to modify the `test.sh` / `run_test.h` / CI script to add the conditional, while it can just be changes in CI script (knowing what environment it is running against and set the variables at the top level)

GuanLuo (changes requested):

qa/L0_perf_nomodel/run_test.sh
CoderHam: Done

qa/L0_perf_nomodel/run_test.sh
CoderHam: I think PERF_CLIENT (just like server) should be configured based on TRITON_DIR. 
I did make the changes for ARCH and got rid of TEST_JETSON

tanmayv25: Make sure you maintain the consistency in all L0_* tests running on jetson.

qa/L0_perf_resnet/test.sh
```diff
@@ -177,23 +188,27 @@ for MODEL_NAME in $OPTIMIZED_MODEL_NAMES; do
...
+        MODEL_FRAMEWORK=${FRAMEWORK} \
+        MODEL_PATH="$REPO/${MODEL_NAME}" \
+        STATIC_BATCH=${STATIC_BATCH} \
+        PERF_CLIENT_PROTOCOL="grpc" \
+        INSTANCE_CNT=${INSTANCE_CNT} \
+        CONCURRENCY=${CONCURRENCY} \
+        ARCH=${ARCH} \
+        BACKEND_CONFIG=" --backend-config=tensorflow,version=2" \
+        bash -x run_test.sh
+fi
```
tanmayv25, line 214: Add newline

qa/L0_perf_nomodel/test.sh
```diff
@@ -35,6 +35,8 @@ if [ -z "$REPO_VERSION" ]; then
     exit 1
 fi
 
+TEST_JETSON=${TEST_JETSON:="0"}
```
GuanLuo, line 38: I think it is no longer needed

GuanLuo (changes requested):

GuanLuo (approved):

tanmayv25 (approved):

tanmayv25 (approved):

PR #3168, opened 2021-07-29
CPU only build with only onnx runtime

jbkyang-nvi:
Changes to enable CPU only build with onnx runtime. 
Onnx Runtime PR: https://github.com/triton-inference-server/onnxruntime_backend/pull/54

### DIFF
--- a/build.py
+++ b/build.py
@@ -327,6 +327,7 @@ def backend_repo(be):
 def backend_cmake_args(images, components, be, install_dir, library_paths):
     if be == 'onnxruntime':
         args = onnxruntime_cmake_args(images, library_paths)
+        print(args)
     elif be == 'openvino':
         args = openvino_cmake_args()
     elif be == 'tensorflow1':
@@ -382,10 +383,13 @@ def pytorch_cmake_args(images):
 
 def onnxruntime_cmake_args(images, library_paths):
     cargs = [
-        '-DTRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON',
         '-DTRITON_BUILD_ONNXRUNTIME_VERSION={}'.format(
             TRITON_VERSION_MAP[FLAGS.version][2])
     ]
+    if FLAGS.enable_gpu:
+        cargs.append('-DTRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON')
+    else:
+        cargs.append('-DTRITON_ENABLE_GPU=OFF')
 
     # If platform is jetpack do not use docker based build
     if target_platform() == 'jetpack':
@@ -409,7 +413,8 @@ def onnxruntime_cmake_args(images, library_paths):
                 cargs.append('-DTRITON_BUILD_CONTAINER_VERSION={}'.format(
                     TRITON_VERSION_MAP[FLAGS.version][1]))
 
-            if TRITON_VERSION_MAP[FLAGS.version][3] is not None:
+            if TRITON_VERSION_MAP[
+                    FLAGS.version][3] is not None and FLAGS.enable_gpu:
                 cargs.append('-DTRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON')
                 cargs.append(
                     '-DTRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION={}'.format(
@@ -602,7 +607,16 @@ RUN rm -fr *
 COPY . .
 ENTRYPOINT []
 '''
-        df += install_dcgm_libraries(argmap['DCGM_VERSION'])
+        if FLAGS.enable_gpu:
+            df += install_dcgm_libraries(argmap['DCGM_VERSION'])
+        else:
+            df += '''
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+            git \
+            libnuma-dev && \
+    rm -rf /var/lib/apt/lists/*
+'''
 
     df += '''
 ENV TRITON_SERVER_VERSION ${TRITON_VERSION}
@@ -660,7 +674,7 @@ FROM ${{BASE_IMAGE}}
 '''.format(argmap['TRITON_VERSION'], argmap['TRITON_CONTAINER_VERSION'],
            argmap['BASE_IMAGE'])
 
-    df += dockerfile_prepare_container_linux(argmap, backends)
+    df += dockerfile_prepare_container_linux(argmap, backends, FLAGS.enable_gpu)
 
     df += '''
 WORKDIR /opt/tritonserver
@@ -698,7 +712,7 @@ COPY --chown=1000:1000 --from=tritonserver_build /workspace/build/sagemaker/serv
         dfile.write(df)
 
 
-def dockerfile_prepare_container_linux(argmap, backends):
+def dockerfile_prepare_container_linux(argmap, backends, enable_gpu):
     # Common steps to produce docker images shared by build.py and compose.py.
     # Sets enviroment variables, installs dependencies and adds entrypoint
     df = '''
@@ -740,7 +754,24 @@ RUN apt-get update && \
          libre2-5 && \
     rm -rf /var/lib/apt/lists/*
 '''
-    df += install_dcgm_libraries(argmap['DCGM_VERSION'])
+    if enable_gpu:
+        df += install_dcgm_libraries(argmap['DCGM_VERSION'])
+        df += '''
+# Extra defensive wiring for CUDA Compat lib
+RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
+ && echo ${_CUDA_COMPAT_PATH}/lib > /etc/ld.so.conf.d/00-cuda-compat.conf \
+ && ldconfig \
+ && rm -f ${_CUDA_COMPAT_PATH}/lib 
+'''
+    else:
+        df += '''
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        git \
+        dirmngr \ 
+        libnuma-dev && \
+    rm -rf /var/lib/apt/lists/*
+'''
     # Add dependencies needed for python backend
     if 'python' in backends:
         df += '''
@@ -755,12 +786,6 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 '''
     df += '''
-# Extra defensive wiring for CUDA Compat lib
-RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
- && echo ${_CUDA_COMPAT_PATH}/lib > /etc/ld.so.conf.d/00-cuda-compat.conf \
- && ldconfig \
- && rm -f ${_CUDA_COMPAT_PATH}/lib
-
 WORKDIR /opt/tritonserver
 RUN rm -fr /opt/tritonserver/*
 COPY --chown=1000:1000 nvidia_entrypoint.sh .
--- a/build.py
+++ b/build.py
@@ -769,9 +769,18 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         git \
         dirmngr \ 
-        libnuma-dev && \
+        libnuma-dev \
+        curl && \
     rm -rf /var/lib/apt/lists/*
 '''
+        if 'onnxruntime' in backends:
+            df += '''
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        libgomp1 && \
+    rm -rf /var/lib/apt/lists
+'''
+
     # Add dependencies needed for python backend
     if 'python' in backends:
         df += '''
--- a/compose.py
+++ b/compose.py
@@ -71,7 +71,8 @@ FROM {}
            images["full"], images["min"])
 
     import build
-    df += build.dockerfile_prepare_container_linux(argmap, backends)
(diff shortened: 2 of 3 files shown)


build.py
```diff
@@ -327,6 +327,7 @@ def backend_repo(be):
 def backend_cmake_args(images, components, be, install_dir, library_paths):
     if be == 'onnxruntime':
         args = onnxruntime_cmake_args(images, library_paths)
+        print(args)
```
GuanLuo, line 330: Remove debug log

build.py
```diff
@@ -409,7 +413,8 @@ def onnxruntime_cmake_args(images, library_paths):
                 cargs.append('-DTRITON_BUILD_CONTAINER_VERSION={}'.format(
                     TRITON_VERSION_MAP[FLAGS.version][1]))
 
-            if TRITON_VERSION_MAP[FLAGS.version][3] is not None:
+            if TRITON_VERSION_MAP[
+                    FLAGS.version][3] is not None and FLAGS.enable_gpu:
```
GuanLuo, line 417: I don't think openvino is GPU based. @tanmayv25 any thoughts?

build.py
```diff
@@ -698,7 +712,7 @@ def create_dockerfile_linux(ddir, dockerfile_name, argmap, backends, repoagents,
         dfile.write(df)
 
 
-def dockerfile_prepare_container_linux(argmap, backends):
+def dockerfile_prepare_container_linux(argmap, backends, enable_gpu):
```
GuanLuo, line 715: You don't need to add an extra argument, just get value from `FLAGS.enable_gpu` when needed like how it is done elsewhere

build.py
```diff
@@ -740,7 +754,33 @@ def dockerfile_prepare_container_linux(argmap, backends):
...
+        libnuma-dev \
+        curl && \
+    rm -rf /var/lib/apt/lists/*
+'''
+        if 'onnxruntime' in backends:
+            df += '''
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+        libgomp1 && \
+    rm -rf /var/lib/apt/lists
```
GuanLuo, line 781: These are dependencies required for CPU only build? I wonder why we don't need them before (I couldn't find their reference in other place of `build.py`)

qa/L0_infer/install_and_test.sh
```diff
@@ -29,11 +29,13 @@
 # dependencies to run L0_infer tests
 apt-get update && \
     apt-get install -y --no-install-recommends \
+         curl \
+         jq \
          python3 \
          python3-pip && \
-pip3 install --upgrade pip
+pip3 install --upgrade pip && \
```
GuanLuo, line 36: I don't think `&&` is needed

GuanLuo (changes requested):

build.py
tanmayv25: There are openvino plugins for GPU. See [here](https://github.com/openvinotoolkit/openvino/blob/master/docs/IE_PLUGIN_DG/Plugin.md). They even have a plugin wrapping around cuDNN for supporting NVIDIA GPUs. Though the support is marked experimental.

I would confirm whether onnx-openvino support GPU plugins.

tanmayv25: Anyways, onnx-openvino is supported on CPU. Hence, I would advice adding the `-DTRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON` for cpu only build too.

build.py
jbkyang-nvi: This specific function is used in `compose.py` that's why I'm passing it in as a param

build.py
jbkyang-nvi: It's because the base image is ubuntu:20.04 instead of cuda. I think the Nvidia cuda container installs OpenMP

build.py
```diff
@@ -602,7 +607,16 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
...
 ENTRYPOINT []
 '''
-        df += install_dcgm_libraries(argmap['DCGM_VERSION'])
+        if FLAGS.enable_gpu:
+            df += install_dcgm_libraries(argmap['DCGM_VERSION'])
+        else:
+            df += '''
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+            git \
```
CoderHam, line 616: Isnt git installed earlier?
Also why is numa installed only for non-GPU build?

qa/L0_infer/install_and_test.sh
```diff
@@ -29,11 +29,13 @@
...
 # install client libraries
-pip3 install nvidia-pyindex
+pip3 install nvidia-pyindex && \
```
CoderHam, line 38: Add indent if using && \ in previous line. Not really sure this change is necessary

build.py
CoderHam: I think these deps we install in ubuntu:20.04 instead of our current base container should be handled elsewhere. We would want a single function that installs these additional deps and ensure that we do NOT try to install them again unless the base container is not our current tritonserver base container.

CoderHam (changes requested):

build.py
jbkyang-nvi: numa is a required dependency. Since numa_utils is included everywhere. I can remove the dependency in a separate PR.

https://jirasw.nvidia.com/browse/DLIS-2749

GuanLuo: NUMA is a generic optimization

jbkyang-nvi:
@CoderHam  with respect to https://github.com/triton-inference-server/server/pull/3168#discussion_r679422483 I agree we don't think we want to install additional dependencies if we don't have to. However adding things into a single function might make dependencies diverge.
> ensure that we do NOT try to install them again unless the base container is not our current tritonserver base container.
I am a bit confused by what you mean. Do you think I have too many `RUN` commands in the dockerfile? Or do you mean we should do a hard check on if the base image has `*-min` in it?

CoderHam:
My problem is not with the number of RUN commands. It is with the readability. A user should easily be able to see from build.py that ABC deps will be installed based on his configuration. 
A dedicated function that performs the necessary checks and installs the additional deps will be useful.

build.py
CoderHam: @jbkyang-nvi why are is the numa install only for when --enable-gpu if off?

jbkyang-nvi: It's because CUDA image installs NUMA. In can make the check on if it's UBUNTU 20.04

deadeyegoodwin: It is confusing why git and numa are installed only for non-gpu. Can we just install them unconditionally with other deps. If they are already installed then apt-get will just skip them.

deadeyegoodwin (changes requested):

build.py
```diff
@@ -733,34 +740,44 @@ def dockerfile_prepare_container_linux(argmap, backends):
...
     # Add dependencies needed for python backend
     if 'python' in backends:
         df += '''
 # python3, python3-pip and some pip installs required for the python backend
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-         python3 libarchive-dev \
-         python3-pip && \
+            python3 libarchive-dev \
+            python3-pip && \
```
jbkyang-nvi, line 774: aligning everything to install, like the previous `apt-get` commands

build.py
```diff
@@ -733,34 +740,44 @@ def dockerfile_prepare_container_linux(argmap, backends):
 
 # Common dependencies. FIXME (can any of these be conditional? For
 # example libcurl only needed for GCS?)
+# libgomp1 is used for Onnxruntime
```
GuanLuo, line 743: Should this be conditioned on whether "onnxruntime" is in `backends`?

jbkyang-nvi: I wasn't sure if it was worth adding an extra RUN command in the layer. I think the GPU min container installs it by default

GuanLuo: Can you elaborate on the worry of adding `RUN` instructions? Regardless of that, there is actually ways to resolve the backend required package at the Python script level:
```
RUN apt-get update && \
            libb64-0d \
            libcurl4-openssl-dev \
            libre2-5 \
            git \
            dirmngr \
            libnuma-dev \
            curl \
            {ort_dependencies} && \
    rm -rf /var/lib/apt/lists/*
'''.format(ort_dependencies=ort_dependencies)
```
And you prepare the `ort_dependencies` variable earlier to be something like:
```
ort_dependencies = "libgomp1" if "onnxruntime" in backneds else ""
```

jbkyang-nvi: Run adds additional layers, which is discouraged in dockerfiles I think. 
ref: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

deadeyegoodwin: Each layer in docker adds some overhead... but I agree with Guan in the best way to address conditionally adding libgomp1 that does not require an extra RUN.

GuanLuo (dismissed):
Looks good to me, you can remove comment "libgomp1 is used for Onnxruntime" in line 744 because it is handled explicitly now.

tanmayv25 (dismissed):

deadeyegoodwin (commented):
Looks ok, do not merge yet.

build.py
```diff
@@ -334,7 +334,7 @@ def backend_cmake_args(images, components, be, install_dir, library_paths):
     elif be == 'tensorflow2':
         args = tensorflow_cmake_args(2, images, library_paths)
     elif be == 'python':
-        args = []
+        args = ['-DTRITON_ENABLE_GPU={}'.format(cmake_enable(FLAGS.enable_gpu))]
```
CoderHam, line 337: Make this symmetric to other backends and use python_cmake_args function

CoderHam: Actually we already add this to cargs below:
```
    cargs.append('-DTRITON_ENABLE_GPU:BOOL={}'.format(
        cmake_enable(FLAGS.enable_gpu)))
```

build.py
```diff
@@ -550,6 +553,7 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
 # python3-dev is needed by Torchvision
 # python3-pip and libarchive-dev is needed by python backend
 # uuid-dev and pkg-config is needed for Azure Storage
+
```
CoderHam, line 556: Remove unnecessary space

build.py
```diff
@@ -756,32 +764,42 @@ def dockerfile_prepare_container_linux(argmap, backends):
 # example libcurl only needed for GCS?)
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-         libb64-0d \
-         libcurl4-openssl-dev \
-         libre2-5 && \
+            libb64-0d \
```
CoderHam, line 767: Fix indent

build.py
```diff
@@ -756,32 +764,42 @@ def dockerfile_prepare_container_linux(argmap, backends):
...
+
     # Add dependencies needed for python backend
     if 'python' in backends:
         df += '''
 # python3, python3-pip and some pip installs required for the python backend
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-         python3 libarchive-dev \
-         python3-pip && \
+            python3 libarchive-dev \
```
CoderHam, line 794: Here too

compose.py
```diff
@@ -91,7 +92,8 @@ def add_requested_backends(ddir, dockerfile_name, backends):
     for backend in backends:
         df += '''COPY --chown=1000:1000 --from=full /opt/tritonserver/backends/{} /opt/tritonserver/backends/{}    
 '''.format(backend, backend)
-    df += '''
+    if len(backends) > 0:
```
CoderHam, line 95: len is always +ve (>= 0), so len != 0 should suffice

qa/L0_infer/install_and_test.sh
```diff
@@ -29,8 +29,10 @@
 # dependencies to run L0_infer tests
 apt-get update && \
     apt-get install -y --no-install-recommends \
+         curl \
+         jq \
          python3 \
-         python3-pip && \
+         python3-pip
```
CoderHam, line 35: Why break these compound statements?

CoderHam (changes requested):

build.py
jbkyang-nvi: I have commented above, This is to keep all the apt-get indents the same

build.py
jbkyang-nvi: If we are looking for consistency in the dockerfile, this apt-get does not align with the rest of the apt-get indents

qa/L0_infer/install_and_test.sh
jbkyang-nvi: It doesn't have to be connected. @GuanLuo suggested I remove it

compose.py
jbkyang-nvi: Is this just a style preference?

deadeyegoodwin (approved):

build.py
```diff
@@ -574,7 +577,9 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
             zlib1g-dev \
             libarchive-dev \
             pkg-config \
-            uuid-dev && \
+            uuid-dev \
+            libnuma-dev \
+            libnuma1 && \
```
CoderHam, line 582: Do we need libnuma1 if we have libnuma-dev?

CoderHam (changes requested):

GuanLuo (approved):

build.py
jbkyang-nvi: Yeah.

CoderHam (approved):

PR #3180, opened 2021-07-30
Include test duration in emails

krishung5:


qa/L0_client_memory_growth/client_memory_mail.py
```diff
@@ -37,7 +37,7 @@
     today = date.today().strftime("%Y-%m-%d")
     subject = "Triton Client Memory Growth " + sys.argv[1] + " Summary: " + today
     memory_graphs = glob.glob("client_memory_growth*.log")
-    write_up = "<p>This test is run for both HTTP and GRPC protocols using C++ and Python test scripts. The max-allowed difference between mean and maximum memory usage is set to 10MB and 1MB for C++ and Python tests individually.</p>"
+    write_up = "<p>This test is run for both HTTP and GRPC protocols using C++ and Python test scripts for around 8 hours. The max-allowed difference between mean and maximum memory usage is set to 10MB and 1MB for C++ and Python tests individually.</p>"
```
CoderHam, line 40: Why do we want to include the test duration? If we do it's better to call the approximate time for east test case instead of the overall runtime of the test

CoderHam (changes requested):

qa/L0_client_memory_growth/client_memory_mail.py
krishung5: David G suggests that it would be great if the test duration is included in the emails. Will change to showing the time for each test case.

qa/L0_client_memory_growth/client_memory_mail.py
```diff
@@ -38,7 +38,7 @@
     subject = "Triton Client Memory Growth " + sys.argv[1] + " Summary: " + today
     memory_graphs = glob.glob("client_memory_growth*.log")
     write_up = "<p>This test is run for both HTTP and GRPC protocols using C++ and Python test scripts. The max-allowed difference between mean and maximum memory usage is set to 10MB and 1MB for C++ and Python tests individually.</p>"
-    write_up += "<p><b>&#8226 What to look for</b><br>A linear memory growth in the beginning of the graph is acceptable only when it is followed by a flat memory usage. If a linear memory growth is observed during the entire test then there is possibly a memory leak.</p>"
+    write_up += "<p><b>&#8226 What to look for</b><br>The approximate time for each test case is 1.7 hours. A linear memory growth in the beginning of the graph is acceptable only when it is followed by a flat memory usage. If a linear memory growth is observed during the entire test then there is possibly a memory leak.</p>"
```
deadeyegoodwin, line 41: Sorry, I wasn't specific enough before. Is it possible to update the script to actually measure the time for each test run, and then include that measured time in the information you output for each test just above the graph?

deadeyegoodwin (changes requested):

qa/L0_client_memory_growth/client_memory_mail.py
krishung5: Sure. Will update.

qa/L0_client_memory_growth/test.sh
```diff
@@ -96,7 +96,9 @@ for PROTOCOL in http grpc; do
             EXTRA_ARGS="-r ${REPETITION_PY} -i ${PROTOCOL}"
         fi
 
+        SECONDS=0
```
CoderHam, line 99: Where is SECONDS used?

CoderHam (changes requested):

qa/L0_client_memory_growth/test.sh
krishung5: SECONDS is a builtin variable in Bash that tracks the number of seconds that have passed since the shell was started, so I set SECONDS to 0 before starting the following events and calculate the time afterwards.

tanmayv25: It is an internal bash variable. @krishung5  is setting it to 0 so that $SECONDS captures the test runtime.

qa/L0_client_memory_growth/test.sh
```diff
@@ -117,6 +119,7 @@ for PROTOCOL in http grpc; do
             fi
 
             # Log the graph for memory growth and the change between Average and Max memory usage
```
tanmayv25, line 121: Can you update the above comment to include test duration?

qa/L0_client_memory_growth/test.sh
```diff
@@ -117,6 +119,7 @@ for PROTOCOL in http grpc; do
...
+            echo -e "Test Duration: $(($TEST_DURATION / 3600)):$((($TEST_DURATION / 60) % 60)):$(($TEST_DURATION % 60)) (H:M:S)" >> ${GRAPH_LOG}
```
tanmayv25, line 122: HH::MM::SS?

tanmayv25 (changes requested):

qa/L0_client_memory_growth/test.sh
CoderHam: Thanks for clarifying that. It is a pretty neat feature :)

qa/L0_client_memory_growth/test.sh
krishung5: Updated

qa/L0_client_memory_growth/test.sh
krishung5: Updated, please review.

qa/L0_client_memory_growth/test.sh
```diff
@@ -116,7 +118,8 @@ for PROTOCOL in http grpc; do
                 RET=1
             fi
 
-            # Log the graph for memory growth and the change between Average and Max memory usage
+            # Log test duration, the graph for memory growth and the change between Average and Max memory usage
+            echo -e "Test Duration: $(($TEST_DURATION / 3600)):$((($TEST_DURATION / 60) % 60)):$(($TEST_DURATION % 60)) (HH::MM::SS)" >> ${GRAPH_LOG}
```
tanmayv25, line 122: Sorry for not being clear enough.

You must use formatted outputs to actually get time reported in HH:MM:SS.
See [here ](https://askubuntu.com/questions/1028924/how-do-i-use-seconds-inside-a-bash-script) for example.

So instead of 0:1:2 the output should be like 00:01:02.

krishung5: I see. Thanks for clarifying this, modifying it now.

tanmayv25 (approved):

CoderHam (approved):

deadeyegoodwin (approved):

PR #3181, opened 2021-07-30
Clarify some metric descriptions

deadeyegoodwin:


docs/metrics.md
```diff
@@ -54,9 +54,9 @@ The following table describes the available metrics.
 |              |GPU Utilization |GPU utilization rate (0.0 - 1.0)       |Per GPU    |Per second   |
 |GPU Memory    |GPU Total Memory|Total GPU memory, in bytes             |Per GPU    |Per second   |
 |              |GPU Used Memory |Used GPU memory, in bytes              |Per GPU    |Per second   |
-|Count         |Request Count   |Number of inference requests           |Per model  |Per request  |
-|              |Execution Count |Number of inference executions (request count / execution count = average dynamic batch size)|Per model|Per request|
-|              |Inference Count |Number of inferences performed (one request counts as "batch size" inferences)|Per model|Per request|
+|Count         |Request Count   |Number of inference requests received by Triton (each request is counted as 1, even if the request contains a batch) |Per model  |Per request  |
+|              |Inference Count |Number of inferences performed (a batch of "n" is counted as "n" inferences)|Per model|Per request|
+|              |Execution Count |Number of inference executions (inference count / execution count = average executed batch size)|Per model|Per request|
```
tanmayv25, line 59: Shouldn't this be better? 
`execution count =  inference count / average executed batch size`

GuanLuo: I think `average executed batch size` should be the result of calculation because we don't report it explicitly

tanmayv25: Still it is a bit confusing to see the above expression in parenthesis for Execution Count. The content in parenthesis should be detailing more about the target metric that is "Execution Count" and should not be introducing any new variable. 

How about having a section after the table which explains the relationship between the metrics reported?

tanmayv25 (approved):

CoderHam (approved):

GuanLuo (approved):

PR #3185, opened 2021-08-02
Modify default value of allow-sagemaker to be false

GuanLuo:


deadeyegoodwin (dismissed):

src/servers/main.cc
```diff
@@ -393,7 +393,7 @@ std::vector<Option> options_
 #endif  // TRITON_ENABLE_GRPC
 #if defined(TRITON_ENABLE_SAGEMAKER)
       {OPTION_ALLOW_SAGEMAKER, "allow-sagemaker", Option::ArgBool,
-       "Allow the server to listen for Sagemaker requests."},
+       "Allow the server to listen for Sagemaker requests. Default is false."},
```
tanmayv25, line 396: We must mention the defaults for other sagemaker options as well.

deadeyegoodwin (approved):

src/servers/main.cc
```diff
@@ -393,15 +389,16 @@ std::vector<Option> options_
...
       {OPTION_SAGEMAKER_PORT, "sagemaker-port", Option::ArgInt,
-       "The port for the server to listen on for Sagemaker requests."},
+       "The port for the server to listen on for Sagemaker requests. Default "
+       "is 8080."},
```
CoderHam, line 395: Can we default it to a different (not 8080) port since it is a commonly used port for many web applications? Or do we default to 8080 intentionally?

CoderHam (changes requested):

src/servers/main.cc
GuanLuo: Yes

CoderHam (approved):

tanmayv25 (approved):

PR #3190, opened 2021-08-03
Fix runtime for L0_client_memory_growth

krishung5:


qa/L0_client_memory_growth/test.sh
```diff
@@ -56,8 +56,8 @@ source ../common/util.sh
 # Set the number of repetitions in nightly and weekly tests
 # Set the email subject for nightly and weekly tests
 if [ "$TRITON_PERF_WEEKLY" == 1 ]; then
-    REPETITION_CPP=2000000
```
CoderHam, line 59: Add a comment to clarify how much time these reps correspond to approximately

CoderHam (changes requested):

qa/L0_client_memory_growth/test.sh
krishung5: Comments updated.

qa/L0_client_memory_growth/test.sh
```diff
@@ -56,8 +56,9 @@ source ../common/util.sh
...
-    REPETITION_PY=2400000
+    # Run the test for approximately 7 hours
```
CoderHam, line 59: All tests cumulatively run for 7 hours or does each test case run for 7 hours? I think it would be better to mention the time for each test case that corr. to the reps for that test case.

CoderHam (changes requested):

krishung5:
Just found that test duration of c++_grpc test case is less than what I got before. Repetition is modified, running the test locally. Will update the repetition and comments if needed.

CoderHam (dismissed):

CoderHam (approved):

PR #3201, opened 2021-08-04
Implement backend API for accessing secondary devices

tanmayv25:
needed for dla support in TRT backend.

### DIFF
--- a/src/backends/backend/triton_model_instance.h
+++ b/src/backends/backend/triton_model_instance.h
@@ -65,6 +65,19 @@ class TritonModelInstance {
   bool IsPassive() const { return passive_; }
   const std::vector<std::string>& Profiles() const { return profile_names_; }
 
+  struct SecondaryDevice {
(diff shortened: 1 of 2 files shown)


src/backends/backend/triton_model_instance.h
```diff
@@ -65,6 +65,19 @@ class TritonModelInstance {
...
 
+  struct SecondaryDevice {
+    SecondaryDevice(const std::string kind, const int64_t id)
+        : kind_(kind), id_(id)
+    {
+    }
+    const std::string kind_;
+    const int64_t id_;
+  };
+  const std::vector<SecondaryDevice> SecondaryDevices() const
```
deadeyegoodwin, line 76: Return as const&

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #3208, opened 2021-08-06
Extend START, END, READY controls to allow BOOL type

krishung5:


src/core/sequence_batch_scheduler.cc
```diff
@@ -296,15 +305,16 @@ SequenceBatchScheduler::CreateBooleanControlTensors(
...
+        &bool_false_value, &bool_true_value));
     if (!tensor_name.empty()) {
       std::shared_ptr<InferenceRequest::Input> true_override;
       std::shared_ptr<InferenceRequest::Input> false_override;
 
       RETURN_IF_ERROR(GetBooleanOverrideInputs(
           tensor_name, config.max_batch_size() != 0, tensor_datatype,
-          fp32_false_value, fp32_true_value, int32_false_value,
-          int32_true_value, &true_override, &false_override));
+          fp32_false_value, fp32_true_value, int32_false_value, int32_true_value,
```
CoderHam, line 316: I think you need to run formatter.

CoderHam (changes requested):

src/core/sequence_batch_scheduler.cc
krishung5: Thanks, updated.

src/core/model_config_utils.cc
```diff
@@ -444,11 +444,12 @@ GetBooleanSequenceControlProperties(
         seen_control = true;
 
         if (c.int32_false_true_size() > 0) {
-          if (c.fp32_false_true_size() != 0) {
+          if (c.fp32_false_true_size() != 0 || c.bool_false_true_size() != 0) {
```
deadeyegoodwin, line 447: parens around sub-expressions

src/core/model_config_utils.cc
```diff
@@ -444,11 +444,12 @@ GetBooleanSequenceControlProperties(
         seen_control = true;
 
         if (c.int32_false_true_size() > 0) {
```
deadeyegoodwin, line 446: Have a single test to make sure that exactly one of int, float, or bool is specified.

deadeyegoodwin (changes requested):
I general these changes need to be made on top of Tanmay's upcoming change which changes many of these files.

src/core/model_config_utils.cc
krishung5: Could you elaborate more on that? So far the if -else statements should be able to test if only one type is specified and return error message if not.

src/core/model_config_utils.cc
krishung5: Updated.

src/core/model_config_utils.cc
deadeyegoodwin: Before the code was structured around their being only two choices. If A then must not be B. You've changed it to: if A then must not be B or C, else if B then must not be C, else must be C. Seems better to say. If not (A || B || C) then error. If (A&&B)||(A&&C)||(B&&C) then error.

krishung5: I see. Thank you for the explanation, will modify.

src/core/model_config_utils.cc
```diff
@@ -404,7 +404,7 @@ GetBooleanSequenceControlProperties(
     const bool required, std::string* tensor_name,
     inference::DataType* tensor_datatype, float* fp32_false_value,
     float* fp32_true_value, int32_t* int32_false_value,
-    int32_t* int32_true_value)
+    int32_t* int32_true_value, bool* bool_false_value, bool* bool_true_value)
```
GuanLuo, line 407: Can the function declaration be changed to taking a pointer of union of `int / float / bool` for `false_value` / `true_value`? The actual type stored can be inferred from `tensor_datatype`

krishung5: I think I can make this change. @deadeyegoodwin I would like to confirm that this can be part of the changes in implicit state management implementation.

deadeyegoodwin: I think you can defer this kind of change to a later PR, be it part of implicit state or something else.  It is somewhat annoying to have so many args but I'm also not too happy about having a generic void* value.

krishung5: Got it, thanks!

src/core/model_config_utils.cc
```diff
@@ -572,11 +600,12 @@ GetTypedSequenceControlProperties(
...
         seen_control = true;
 
-        if ((c.int32_false_true_size() > 0) || (c.fp32_false_true_size() > 0)) {
+        if ((c.int32_false_true_size() > 0) || (c.fp32_false_true_size() > 0) ||
+            (c.bool_false_true_size() > 0)) {
           return Status(
               Status::Code::INVALID_ARG,
-              "sequence batching must not specify either 'int32_false_true' "
-              "nor 'fp32_false_true' for " +
+              "sequence batching must not specify neither 'int32_false_true' "
```
CoderHam, line 607: sequence batching must not specify either .... or ...

CoderHam (changes requested):

src/core/model_config_utils.cc
```diff
@@ -609,7 +605,7 @@ GetTypedSequenceControlProperties(
           return Status(
               Status::Code::INVALID_ARG,
               "sequence batching must not specify neither 'int32_false_true' "
```
CoderHam, line 607: not specify either ... or ...

CoderHam (changes requested):

CoderHam (dismissed):

deadeyegoodwin (approved):

deadeyegoodwin (approved):

PR #3216, opened 2021-08-09
Cleanup L0_trt_dla test

CoderHam:


deadeyegoodwin (approved):

tanmayv25:
Why are we using a separate dla_client instead of image_client?  Whether a model runs on GPU or DLA, the client should stay agnostic to it.

qa/L0_trt_dla/dla_client.py
```diff
@@ -0,0 +1,135 @@
...
+import tritonclient.http as httpclient
+from tritonclient.utils import InferenceServerException
+
+
+def preprocess(img, dtype, c, h, w, scaling):
+    """
+    Pre-process an image to meet the size and type
+    requirements specified by the parameters.
+    """
+    # np.set_printoptions(threshold='nan')
```
tanmayv25, line 42: Remove this line?

qa/L0_trt_dla/dla_client.py
```diff
@@ -0,0 +1,135 @@
...
+    image_data = preprocess(img, np.int8, 3, 224, 224, FLAGS.scaling)
+    image_data = np.expand_dims(image_data, axis=0)
+
+    batched_image_data = image_data
+    for i in range(1, batch_size):
+        batched_image_data = np.concatenate((batched_image_data, image_data), axis=0)
+
+    inputs = [httpclient.InferInput('input_tensor_0', [batch_size, 3, 224, 224], 'INT8')]
+    inputs[0].set_data_from_numpy(batched_image_data, binary_data=True)
+    # inputs[0].set_data_from_numpy([image_data], binary_data=True)
```
tanmayv25, line 116: This line as well?

GuanLuo (commented):
Same as Tanmay's concern, why don't we use image client directly?

CoderHam:
> Why don't we use image client directly?

Since the image client only deals with models with FP32 outputs / inputs and models that output a 1D label of shape [1000] we would need to modify the client for the same. 
The input for the dla model only accepts a batch size of 32 (INT8) and outputs a 3D of shape (1,1,1) for the output where the output is the label and not the softmax for the 1000 labels.

In an effort to keep the image client for being more complicated I stuck to using a different client.

qa/L0_trt_dla/dla_client.py
```diff
@@ -0,0 +1,133 @@
...
+    # Validate the results by comparing with precomputed values.
+    # VULTURE class corresponds with index 23
+    EXPECTED_CLASS_INDEX = 23
+
+    output_data = results.as_numpy('topk_layer_output_index')
+    print(output_data)
+    for i in range(batch_size):
+        if output_data[i][0][0] != EXPECTED_CLASS_INDEX:
+            print("Infer error: Incorrect output label")
+            sys.exit(1)
```
GuanLuo, line 133: If you are already verifying the returned value in Python script, then you don't need to print and check that again in the bash script.

tanmayv25:
If the requirement was to handle a very specific model, then instead of dla_client.py that brings a lot of artifacts from image_client,py, I would prefer a dla_test.py following python unittest framework like most of our tests.

It can prepare specific inputs and process specific output.

qa/L0_trt_dla/dla_client.py
```diff
@@ -0,0 +1,133 @@
...
+
+    FLAGS = parser.parse_args()
+    try:
+        triton_client = httpclient.InferenceServerClient(
+            url=FLAGS.url, verbose=FLAGS.verbose)
+    except Exception as e:
+        print("channel creation failed: " + str(e))
+        sys.exit(1)
+
+    model_name = "resnet50_plan"
```
GuanLuo, line 102: Since the Python script is targeting for this particular model and input in the test, it doesn't need to be constructed like the client examples, a unit test style makes more sense as the arguments above are actually determined and thus the `preprocess()` helper function can also be simplified.

CoderHam: Done

qa/L0_trt_dla/dla_client.py
CoderHam: Done

qa/L0_trt_dla/test.sh
```diff
@@ -35,19 +35,15 @@ if [ -z "$REPO_VERSION" ]; then
...
 # PLAN. Without this test will fail on a heterogeneous system.
 export CUDA_VISIBLE_DEVICES=0
 
-IMAGE_CLIENT=../clients/image_client
-IMAGE=../images/vulture.jpeg
-
-CAFFE2PLAN=../common/caffe2plan
+# Only need to set paths for jetson since this test runs only on jetson
+TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
+IMAGE_CLIENT=./dla_test.py
```
GuanLuo, line 44: change variable name to something like DLA_TEST

tanmayv25 (approved):

PR #3220, opened 2021-08-10
Re-enable GPU tensor test for PB

Tabrizian:


qa/L0_backend_python/python_unittest.py
```diff
@@ -30,17 +30,17 @@
 
 import test_util as tu
 import unittest
-import tritonclient.http as httpclient
```
CoderHam, line 33: Why move from http to grpc?

CoderHam (approved):
1 small comment else LGTM

qa/L0_backend_python/python_unittest.py
Tabrizian: It didn't support the `client_timeout` parameter in `infer`.

PR #3230, opened 2021-08-13
Remove the unneeded plan backend file

tanmayv25:
Looks like the file was added back after the rebase of this PR: https://github.com/triton-inference-server/server/pull/3208

PR removes this file as it is no longer required with the move to separate tensorrt_backend repo.


krishung5 (approved):
Thanks for the catch!

CoderHam (approved):

PR #3236, opened 2021-08-16
Fix L0_sequence_batcher_jetson

krishung5:


qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -1186,6 +1194,7 @@ def test_backlog(self):
 
                 for t in threads:
                     t.start()
+                    time.sleep(1)
```
CoderHam, line 1197: This sleep will prevent concurrent inference requests. The goal of this test case is to:
```
# Send 5 equal-length sequences in
# parallel and make sure they get completely batched into
# batch-size 4 inferences plus the 5th should go in the
# backlog and then get handled once there is a free slot.
```

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -2052,6 +2061,12 @@ def test_queue_delay_no_min_util(self):
         # execution to be a batch of 'null, seq 2'. The executions should not be
         # waited.
 
+        # Adjust the values of 'sequence_thresholds' due to Jetson intermittence
+        if TEST_JETSON:
+            sequence_thresholds = (6000, None)
```
CoderHam, line 2066: 2000 -> 6000 is a big jump
I don't see why we need to set the lt_ms part of the sequence thresholds to a different value here when we already multiply it by the _jetson_slowdown_factor if the platform is jetson.

qa/common/sequence_util.py
```diff
@@ -63,7 +63,7 @@
 
 _deferred_exceptions_lock = threading.Lock()
 _deferred_exceptions = None
-_jetson_slowdown_factor = 1.6
+_jetson_slowdown_factor = 2
```
CoderHam, line 66: You should ideally be able to modify just this variable and the sequence thresholds should be adjusted by this on jetson

CoderHam (changes requested):

qa/L0_sequence_batcher/sequence_batcher_test.py
krishung5: The reason why I changed this is because the response time becomes really long when running this on jetson. (i.e. 13000 ms while originally it should be less than 4000 ms) I didn't make lots of changes on '_jetson_slowdown_factor' is because the variable will affect other tests, but I agree that we should modify only this variable instead of the sequence thresholds since the latter one is already multiplied by the factor. Will modify.

qa/L0_sequence_batcher/sequence_batcher_test.py
krishung5: Ah... didn't think it clearly, updated, thanks!

qa/L0_sequence_batcher/sequence_batcher_test.py
CoderHam: _jetson_slowdown_factor is only used inside sequence_util.py

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -1838,7 +1838,7 @@ def test_backlog_same_correlation_id_no_end(self):
                             trial,
                             model_name,
                             dtype,
-                            1001,
+                            1002,
```
CoderHam, line 1841: What is the reason for this change?

qa/common/sequence_util.py
```diff
@@ -63,7 +63,7 @@
 
 _deferred_exceptions_lock = threading.Lock()
 _deferred_exceptions = None
-_jetson_slowdown_factor = 1.6
+_jetson_slowdown_factor = 6
```
CoderHam, line 66: 6 feels a little extreme. Add a comment above to revisit why jetson is 6x slower for trivial inferences.

CoderHam (changes requested):

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -1790,9 +1791,8 @@ def test_backlog_same_correlation_id_no_end(self):
                             1002,
                             (None, None),
                             # (flag_str, value, pre_delay_ms)
-                            (("start", 11, None), (None, 12, None),
-                             (None, 12, None), ("end", 13, None)),
```
GuanLuo, line 1794: Why switching the order. You are changing the test behavior so something seems wrong

qa/L0_sequence_batcher/sequence_batcher_test.py
krishung5: I found that CORRID 1001 would exceed max_sequence_idle_microseconds while waiting for sending all the other sequences, and then got force-ended, which leads the scheduler to hang because it couldn't receive the enough number of inferences. I was thinking that maybe switching the order of the first two threads can prevent the sequence without an end marker from getting terminated.

qa/L0_sequence_batcher/sequence_batcher_test.py
krishung5: Please take a look at the above comment.

qa/common/sequence_util.py
krishung5: Will do.

krishung5: I think for most of the cases, changing the factor to 3 should be enough, but sometimes there is only one test case exceeding timeout, making the whole test to fail. Setting the factor to 6 can cover the largest test duration I got, but it is actually too much for general cases. Should I still set it to 6 and add a comment for explanation?

qa/L0_sequence_batcher/sequence_batcher_test.py
GuanLuo: What is the `max_sequence_idle_microseconds` for the model? From this test's [description](https://github.com/triton-inference-server/server/blob/main/qa/L0_sequence_batcher/sequence_batcher_test.py#L1723-L1730), I would think that the timeout is quite long to test that "the new sequence must start without waiting for the timeout"

GuanLuo: It is also strange that changing the order helps, in this test the first two threads are started [back-to-back](https://github.com/triton-inference-server/server/blob/main/qa/L0_sequence_batcher/sequence_batcher_test.py#L1852-L1857), unless the Python's thread launch has significant overhead in Jetson.

krishung5: `max_sequence_idle_microseconds` is set to 5000000 based on model configuration. In normal cases, the new sequence does start without waiting for the timeout, but I found that this issue occurred in lots of pipelines since last week, and my guess would be that it took longer than we expected for the threads being launched. 
For example, I was checking "[test_backlog_fill_no_end](https://github.com/triton-inference-server/server/blob/main/qa/L0_sequence_batcher/sequence_batcher_test.py#L1536-L1546)" test case. Originally the delay between each thread is set to 2 seconds, and the order of sequences the server received should be CORRID 1001, 1002 ... 1006. But from the loggings I got, the order could be wrong sometimes. The server could actually receive 1005 before receiving 1004, and thus the scheduler got hang. So I changed the delay to 4 seconds, and seems like this happens lesser than before.

GuanLuo: That is 5 seconds, substracting the 2 seconds that the script waits for before launching 5th sequence, 3 seconds is a long time for sending requests. That looks like a regression that we want to resolve rather than just making the test less time sensitive. The test runs for all backends, does this timeout happen for all backends or only for a subset of backends?

GuanLuo (dismissed):
I agree with the discussion that this change is okay as long as the test behavior is unchanged. I think `test_backlog_same_correlation_id_no_end` is fine, and `test_backlog_fill_no_end` should be fine as well since the batch size checking enforces the batching order.

You need to resolve the slowdown factor discussion with Hemant as that can alter the test without notice. Otherwise the change looks good on my end, you can decide whether to merge it or not.

CoderHam:
@krishung5 lets keep the slowdown factor at 3 for now. Just run through the tests to confirm a slowdown of 3 does not impact the behavior. I think accepting a small level of intermittence is acceptable as long we don't change what the test is attempting to test.

qa/common/sequence_util.py
CoderHam: Set it to 3 for now. I still feel 6x slowdown should not be observed on jetson unless due to machine specific issues. How often does the test fail when using 3 instead of 6?

krishung5:
Thank you for the comments.
@GuanLuo I think I can run the test for more times with slowdown factor set to 3 and see if the result is better. If the changes have not much effect on the test then I guess we don't have to merge in.
@CoderHam Will run through the tests and confirm.

qa/common/sequence_util.py
krishung5: Checking.

CoderHam (approved):

krishung5:
@CoderHam @GuanLuo After a few tries, there was no significant improvements with my changes now, so I think we can stick to the original settings. And changing slowdown factor to 3 seems fine as I only got timeout one out of five.

GuanLuo (approved):

PR #3238, opened 2021-08-17
Extend testing to cover START, END, READY controls for BOOL type

krishung5:


qa/L0_model_config/autofill_noplatform_success/pytorch/control_bool_type/config.pbtxt
```diff
@@ -0,0 +1,44 @@
+
```
CoderHam, line 1: Extra empty line

qa/L0_model_config/autofill_noplatform_success/tensorflow_graphdef/control_bool_type/config.pbtxt
```diff
@@ -0,0 +1,44 @@
+
```
CoderHam, line 1: same here

CoderHam (changes requested):

qa/L0_model_config/test.sh
```diff
@@ -361,6 +361,46 @@ for TARGET_DIR in `ls -d autofill_noplatform_success/*/*`; do
     fi
 done
 
+# Run all noautofill tests that are expected to be successful. These
+# tests add the platform to the model config before running.
```
CoderHam, line 365: These changes appear unrelated to your PR.

CoderHam (commented):
PS: You don't need to use Draft if you are using WIP

krishung5:
Having some problems when trying to generate bool type tensorrt model using `nvcr.io/nvidia/tensorrt:21.07-py3` image:
`[TensorRT] ERROR: 4: [network.cpp::validate::2445] Error Code 4: Internal Error (Boolean tensor: INPUT is not supported on this configuration. This may be because the requested platform is not supported, or because you are statically linking TensorRT.)`. Still working on it.

CoderHam:
Following up offline/on slack
> Having some problems when trying to generate bool type tensorrt model using `nvcr.io/nvidia/tensorrt:21.07-py3` image:

GuanLuo (commented):
This is the right place to test the misconfigured model config and check if proper error is returned, but not the place for validating the success case. L0_sequence_batcher is a better place for success case because we not only want to verify that the model can be loaded but also want to verify that the sequence batcher is forming boolean control inputs properly.

krishung5:
@GuanLuo I think that makes more sense. I can move all bool type models to L0_sequence_batcher test.

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -2310,6 +2316,281 @@ def test_queue_delay_full_min_util(self):
...
 
+    def test_simple_sequence_bool(self):
+        # Send one sequence and check for correct result.
+        # The result should be returned immediately.
+        for trial in _bool_trials:
+            # Run on different protocols.
+            for idx, protocol in enumerate(_protocols):
+                self.clear_deferred_exceptions()
+                try:
+                    dtype = np.bool
```
GuanLuo, line 2327: You shouldn't need to duplicate the test for bool, what you can do is to have a list of data type to be tested, and then add a loop to iterate the types for all tests.

qa/L0_sequence_batcher/bool_type_sequence_models/graphdef_sequence_bool/config.pbtxt
```diff
@@ -0,0 +1,45 @@
+name: "graphdef_sequence_bool"
```
GuanLuo, line 1: I forgot to mention this in the previous round of review: we generate the models used for `L0_sequence_batcher` with script (qa/common/gen_qa_sequence_models.py), and you should be able to generate boolean sequence model as well with minor modifications on the gen script. Then you would want to upload the new models to our test model directory.

And then with the comment regarding test script, the boolean case should be covered without the change in `test.sh`

GuanLuo (changes requested):

qa/L0_sequence_batcher/bool_type_sequence_models/graphdef_sequence_bool/config.pbtxt
krishung5: I see, thanks for the comments, will modify.

qa/L0_sequence_batcher/sequence_batcher_test.py
```diff
@@ -1573,151 +1897,188 @@ def test_backlog_same_correlation_id(self):
...
+                                model_name,
+                                dtype,
+                                1003,
+                                (None, None),
+                                # (flag_str, value, pre_delay_ms)
+                                (("start", True, None), (None, False, None),
+                                 ("end", False, None)) if dtype == np.bool else
+                                (("start", 111, None), (None, 112, None),
+                                 ("end", 113, None)),
+                                False if dtype == np.bool else
```
GuanLuo, line 1999: You may not need to add this, see my comment in the generation script

qa/common/gen_qa_sequence_models.py
```diff
@@ -185,8 +185,15 @@ def create_tf_modelfile(create_savedmodel, models_dir, model_version, max_batch,
         acc = tf.get_variable("ACC", [
             1,
         ], dtype=tf_dtype)
-        tmp = tf.where(tf.equal(start0, 1), input0, tf.add(acc, input0))
-        newacc = tf.where(tf.equal(ready0, 1), tmp, acc)
+
+        if tf_dtype == tf.bool:
+            tmp = tf.where(tf.equal(start0, True), input0,
+                           tf.math.logical_and(acc, input0))
+            newacc = tf.where(tf.equal(ready0, True), tmp, acc)
```
GuanLuo, line 192: The bool control models you generate here are different from the models with other control types, that is why you found that you still need to modify the test script at the end.

What you can do is to make the bool control models behave the same as the other models (same I/O types and operators), except for the places where the "control tensors" are used (comparing to `True` instead of `1`). Take this as example, if you have the following, then from the Triton client's perspective, the model should be identical to the model with other control type.
```
if tf_dtype == tf.bool:
            tmp = tf.where(tf.equal(start0, True), input0, tf.add(acc, input0))
            newacc = tf.where(tf.equal(ready0, True), tmp, acc)
```

qa/L0_sequence_batcher/bool_type_sequence_models/graphdef_sequence_bool/config.pbtxt
GuanLuo: You can remove the `bool_type_sequence_models` folder as now it is not longer needed.

GuanLuo (approved):

CoderHam (approved):

PR #3242, opened 2021-08-18
Wait for execution to finish before exitting

tanmayv25:
Most likely fixes L0_decoupled test.

### DIFF
--- a/src/backends/backend/triton_model_instance.cc
+++ b/src/backends/backend/triton_model_instance.cc
@@ -130,10 +130,15 @@ TritonModelInstance::TritonModelInstance(
         &reporter_);
   }
 #endif  // TRITON_ENABLE_METRICS
+  ongoing_requests_count_.store(0);
 }
 
 TritonModelInstance::~TritonModelInstance()
 {
+  while (ongoing_requests_count_.load() != 0) {
+    LOG_VERBOSE(1) << "Waiting for the execution on instance to finish...";
+    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+  }
   // Model finalization is optional...
   if (model_->Backend()->ModelInstanceFiniFn() != nullptr) {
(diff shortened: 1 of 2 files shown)


src/backends/backend/triton_model_instance.cc
```diff
@@ -130,10 +130,15 @@ TritonModelInstance::TritonModelInstance(
         &reporter_);
   }
 #endif  // TRITON_ENABLE_METRICS
+  ongoing_requests_count_.store(0);
 }
 
 TritonModelInstance::~TritonModelInstance()
 {
+  while (ongoing_requests_count_.load() != 0) {
+    LOG_VERBOSE(1) << "Waiting for the execution on instance to finish...";
```
CoderHam, line 139: You should add the Instance / Model name here

CoderHam (changes requested):

src/backends/backend/triton_model_instance.cc
```diff
@@ -580,6 +587,8 @@ TritonModelInstance::Execute(
 
     TRITONSERVER_ErrorDelete(err);
   }
+
+  ongoing_requests_count_--;
```
GuanLuo, line 591: The decrement is done in `Execute()`, if the request is delegated to a separate thread, Does the `count` track the ongoing request properly? In this case, the `inst_exec_fn` returns "immediately", right?

tanmayv25:
Closing this PR as there might be a better way to handle the clean up. I am pretty sure where the race condition is.

src/backends/backend/triton_model_instance.cc
tanmayv25: It doesn't return immediately. We have subtests that have WAIT time specified as large as 10 sec. That being said I was too much focused on triaging the issue that I came up with this very specific solution. I will share a proper fix soon.

PR #3257, opened 2021-08-20
Print stack trace before the server crashes

krishung5:


src/servers/signal.cc
```diff
@@ -107,6 +110,8 @@ void
 SignalHandler(int signum)
 {
   std::cout << "Signal (" << signum << ") received." << std::endl;
+  std::cout << boost::stacktrace::stacktrace() << std::endl;
```
deadeyegoodwin, line 113: Don't dump stack trace for the existing signals that are handled...

src/servers/signal.cc
```diff
@@ -115,9 +120,12 @@ SignalHandler(int signum)
 TRITONSERVER_Error*
 RegisterSignalHandler()
 {
-  // Trap SIGINT and SIGTERM to allow server to exit gracefully
+  // Trap SIGINT, SIGTERM, SIGSEGV and SIGABRT to allow server to exit
+  // gracefully
   signal(SIGINT, SignalHandler);
   signal(SIGTERM, SignalHandler);
+  signal(SIGSEGV, SignalHandler);
```
deadeyegoodwin, line 127: For SEGV and ABRT we don't want to use the existing signalhandler that tries to exit gracefully. Instead we should use a new handler for those that just dumps the stack trace and then does exit(1)

deadeyegoodwin (changes requested):

src/servers/signal.cc
krishung5: Got it, will modify.

src/servers/signal.cc
```diff
@@ -110,6 +113,15 @@ SignalHandler(int signum)
   CommonSignalHandler();
 }
 
+void
+ErrorSignalHandler(int signum)
+{
+  std::cout << "Signal (" << signum << ") received." << std::endl;
```
deadeyegoodwin, line 119: Missed this before, output to cerr

deadeyegoodwin (changes requested):

src/servers/signal.cc
krishung5: Updated.

deadeyegoodwin (approved):
Make sure to run CI for this

krishung5:
@deadeyegoodwin I have a CI running now, will not merge it in until passes.

PR #3266, opened 2021-08-23
Email reporting for long-running stress

krishung5:


CoderHam:
Can you add a small sample of the prettytable format to the issue?

krishung5:
A sample of the prettytable format:

```
Test Duration: 04:00:05 (HH:MM:SS)
+-----------------------+--------------------+-------------+---------------------------------------------------+
| Test Case             | Number of Failures | Total Times | Test Case Description                             |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| sequence_valid        | 0                  | 18241       | Send a sequence with "start" and "end" flags.     |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| sequence_valid_valid  | 0                  | 2310        | Send two sequences back to back using the same    |
|                       |                    |             | correlation ID with "start" and "end" flags.      |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| sequence_valid_no_end | 0                  | 2374        | Send two sequences back to back using the same    |
|                       |                    |             | correlation ID. The first with "start" and "end"  |
|                       |                    |             | flags, and the second with no "end" flag.         |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| sequence_no_start     | 0                  | 205         | Send a sequence without a "start" flag. Sequence  |
|                       |                    |             | should get an error from the server.              |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| sequence_no_end       | 0                  | 2160        | Send a sequence with "start" flag but that never  |
|                       |                    |             | ends. The sequence should be aborted by the       |
|                       |                    |             | server and its slot reused for another sequence.  |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| timeout_client        | 0                  | 4380        | Expect an exception for small timeout values.     |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| resnet_model_request  | 0                  | 2822        | Send a request using resnet model.                |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| crashing_client       | 0                  | 7220        | Client crashes in the middle of inferences.       |
+-----------------------+--------------------+-------------+---------------------------------------------------+
| TOTAL                 | 0                  | 39712       | X                                                 |
+-----------------------+--------------------+-------------+---------------------------------------------------+
```

krishung5:
@CoderHam Updated, please see the above comment.

CoderHam (dismissed):

qa/L0_long_running_stress/stress.py
```diff
@@ -219,6 +226,30 @@ def get_trial(is_sequence=True):
         return np.random.choice(_trials)
 
 
+def count_test_case(test_case_name):
+    # Count the times each test case runs
+    _test_case_count_mutex.acquire()
```
deadeyegoodwin, line 231: We want the threads to send requests as fast as possible, and taking these locks is potentially going to slow things down. Can you instead give each thread their own "test case count" hashes and then you can update without a lock. At the end of the run you can accumulate all the individual thread counts.

qa/L0_long_running_stress/stress.py
```diff
@@ -514,6 +546,8 @@ def stress_thread(name, seed, test_duration, correlation_id_base):
                 # exercise the idle sequence path of the sequence
                 # scheduler
                 if choice < 0.33:
+                    count_test_case("sequence_no_end")
```
deadeyegoodwin, line 549: For the sequences also count the number of individual requests. In the table you will then have a column for the number of times that test case was chosen ("Count"), and another for how many total requests were sent for that case ("Request Count")

deadeyegoodwin (changes requested):

qa/L0_long_running_stress/stress.py
krishung5: I see, will modify.

qa/L0_long_running_stress/stress.py
krishung5: Got it, will add new column for that.

krishung5:
Updated table:

```
Test Duration: 00:03:05 (HH:MM:SS)
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| Test Case             | Number of Failures | Count | Request Count | Test Case Description                             |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| sequence_valid        | 0                  | 265   | 4129          | Send a sequence with "start" and "end" flags.     |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| sequence_valid_valid  | 0                  | 52    | 806           | Send two sequences back to back using the same    |
|                       |                    |       |               | correlation ID with "start" and "end" flags.      |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| sequence_valid_no_end | 0                  | 27    | 421           | Send two sequences back to back using the same    |
|                       |                    |       |               | correlation ID. The first with "start" and "end"  |
|                       |                    |       |               | flags, and the second with no "end" flag.         |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| sequence_no_start     | 0                  | 2     | 2             | Send a sequence without a "start" flag. Sequence  |
|                       |                    |       |               | should get an error from the server.              |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| sequence_no_end       | 0                  | 32    | 542           | Send a sequence with "start" flag but that never  |
|                       |                    |       |               | ends. The sequence should be aborted by the       |
|                       |                    |       |               | server and its slot reused for another sequence.  |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| timeout_client        | 0                  | 53    | X             | Expect an exception for small timeout values.     |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| resnet_model_request  | 0                  | 43    | X             | Send a request using resnet model.                |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| crashing_client       | 0                  | 102   | X             | Client crashes in the middle of inferences.       |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+
| TOTAL                 | 0                  | 576   | 5900          | X                                                 |
+-----------------------+--------------------+-------+---------------+---------------------------------------------------+

```

qa/L0_long_running_stress/stress.py
CoderHam: I think it would be better to use "Test Count" instead of "Count" since Count is not as descriptive.
>  the number of times that test case was chosen ("Count")

deadeyegoodwin:
@krishung5  Instead of putting 'X' in the request count colunm... just put the request count. It is ok if it is the same as the test count.

CoderHam (approved):

PR #3293, opened 2021-08-30
Update NGC versions post-21.08 release

dzier:


build.py
```diff
@@ -65,8 +65,8 @@
 # incorrectly load the other version of the openvino libraries.
 #
 TRITON_VERSION_MAP = {
-    '2.14.0dev': (
-        '21.09dev',  # triton container
```
deadeyegoodwin, line 69: You want to keep this as 2.14.0dev and 21.09dev

deadeyegoodwin (changes requested):

PR #3296, opened 2021-08-30
Fix build_variants

krishung5:
Because the dl library is needed by boost::stacktrace, the dl library should be linked after the Boost stacktrace libraries are linked.

### DIFF
--- a/src/servers/CMakeLists.txt
+++ b/src/servers/CMakeLists.txt
@@ -241,6 +241,8 @@ target_link_libraries(
   PRIVATE triton-common-error
   PRIVATE triton-core-serverapi
   PRIVATE tritonserver
+  PRIVATE pthread
+  PRIVATE dl
 )
 if (NOT WIN32)
 target_link_libraries(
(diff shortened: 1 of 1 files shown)


src/servers/CMakeLists.txt
```diff
@@ -241,6 +241,7 @@ target_link_libraries(
   PRIVATE triton-common-error
   PRIVATE triton-core-serverapi
   PRIVATE tritonserver
+  PUBLIC pthread dl
```
deadeyegoodwin, line 244: One library per line, and you should be able to use PRIVATE, so

PRIVATE pthread
PRIVATE dl

deadeyegoodwin (changes requested):

src/servers/CMakeLists.txt
krishung5: Got it, will update

krishung5: Updated

deadeyegoodwin (approved):

CoderHam (approved):

PR #3298, opened 2021-08-31
Fix DB to prefer batches with preferred batch size for default max_queue_d…

tanmayv25:
…elay

### DIFF
--- a/qa/L0_batcher/batcher_test.py
+++ b/qa/L0_batcher/batcher_test.py
@@ -187,6 +187,8 @@ class BatcherTest(tu.TestResultCollector):
         # Make sure test.sh set up the correct batcher settings
         config = self.triton_client_.get_model_config(model_name).config
         bconfig = config.dynamic_batching
+        self.assertEqual(len(bconfig.preferred_batch_size),
+                         len(preferred_batch_sizes))
         for i in preferred_batch_sizes:
             self.assertTrue(i in bconfig.preferred_batch_size)
         self.assertEqual(bconfig.max_queue_delay_microseconds,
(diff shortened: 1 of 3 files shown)


qa/L0_batcher/batcher_test.py
```diff
@@ -182,14 +182,15 @@ def check_response(self,
...
 
-    def check_setup(self, model_name):
+    def check_setup(self, model_name, preferred_batch_sizes,
+                    max_queue_delay_us):
         # Make sure test.sh set up the correct batcher settings
         config = self.triton_client_.get_model_config(model_name).config
         bconfig = config.dynamic_batching
-        self.assertTrue(2 in bconfig.preferred_batch_size)
-        self.assertTrue(6 in bconfig.preferred_batch_size)
+        for i in preferred_batch_sizes:
```
deadeyegoodwin, line 190: You should also check if len of preferred_batch_sizes matches the len of same in bconfig

deadeyegoodwin (dismissed):

deadeyegoodwin:
@tanmayv25 oops, didn't mean to approve yet. I had one requested change...

src/core/dynamic_batch_scheduler.cc
```diff
@@ -467,7 +464,7 @@ DynamicBatchScheduler::GetDynamicBatch()
     return 0;
   }
 
-  if (delay_is_exceeded) {
+  if (delay_is_exceeded || (pending_batch_delay_ns_ == 0)) {
```
GuanLuo, line 467: if `pending_batch_delay_ns_ == 0`, shouldn't the batcher NOT return 0 to wait for preferred batch or max batch?

tanmayv25: But the question then becomes how long should it wait for forming the preferred batch size. These changes only let batcher prefer a preferred batch size if one can be formed. Otherwise it will issue the request with pending batches.

deadeyegoodwin: When delay is 0, the batcher should form the largest possible preferred batch, or if can't form any preferred batch then just the largest possible batch, and execute it immediately.

Tanmay, we should have a test with no delay where there are multiple preferred sizes and make sure that the largest preferred size is used, or when no preferred can be formed that the largest possible batch is formed. We already have these kind of tests but I guess they all use delay.

GuanLuo: I see.

tanmayv25: I am adding the tests...

deadeyegoodwin (approved):

PR #3309, opened 2021-09-01
Add test for BLS GPU tensor

Tabrizian:


Tabrizian:
The branch was incorrect.

PR #3310, opened 2021-09-01
Add test for BLS GPU tensor

Tabrizian:


deadeyegoodwin (approved):
Does this allow L0_io test with python backend with all combinations of tensor memory types?

Tabrizian:
L0_io tests the input/output tensor combinations but it doesn't test BLS tensor combinations. There is a [subtest ](https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/io/io_test.py) in L0_backend_python test that was used to test passing the identity tensors in Python backend. It was created to cover some corner cases were the model returned the same GPU tensor as the final tensor. I have extended that test here to do the same for BLS.

PR #3314, opened 2021-09-02
Add test for relative path in execution env

Tabrizian:


qa/L0_backend_python/env/test.sh
```diff
@@ -65,10 +65,13 @@ create_conda_env "3.6" "python-3-6"
...
 conda install tensorflow=2.1.0 -y
 conda-pack -o python3.6.tar.gz
-path_to_conda_pack=`pwd`/python3.6.tar.gz
+
+# Test relative execution env path
+path_to_conda_pack='$$TRITON_MODEL_DIRECTORY/python3.6.tar.gz'
 create_python_backend_stub
 mkdir -p models/python_3_6/1/
 cp ../../python_models/python_version/config.pbtxt ./models/python_3_6
+cp python3.6.tar.gz models/python_3_6
```
deadeyegoodwin, line 74: It would be a somewhat better test if you copy it into models/python_3_6 as a different named file to avoid any change of accidentally picking up the original file.

deadeyegoodwin (changes requested):

qa/L0_backend_python/env/test.sh
Tabrizian: Fixed.

deadeyegoodwin (approved):

PR #3317, opened 2021-09-02
Fix default values for backend configs when not set

CoderHam:
- MinComputeCapability
- BackendDir
- StrictModelConfig

### DIFF
--- a/src/core/tritonserver.cc
+++ b/src/core/tritonserver.cc
@@ -205,13 +205,7 @@ class TritonServerOptions {
   void SetExitOnError(bool b) { exit_on_error_ = b; }
 
   bool StrictModelConfig() const { return strict_model_config_; }
-  void SetStrictModelConfig(bool b)
-  {
-    strict_model_config_ = b;
-    // Note the condition is reverted due to setting name is different
-    AddBackendConfig(
-        std::string(), "auto-complete-config", b ? "false" : "true");
-  }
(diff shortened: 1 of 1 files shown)


src/core/tritonserver.cc
```diff
@@ -1717,13 +1705,27 @@ TRITONSERVER_ServerNew(
...
+  lserver->SetMinSupportedComputeCapability(min_compute_capability);
+  loptions->AddBackendConfig(
+      std::string(), "min-compute-capability",
+      std::to_string(min_compute_capability));
+  loptions->AddBackendConfig(
+      std::string(), "backend-directory", loptions->BackendDir());
+
   lserver->SetStrictReadinessEnabled(loptions->StrictReadiness());
   lserver->SetExitTimeoutSeconds(loptions->ExitTimeout());
   lserver->SetBackendCmdlineConfig(loptions->BackendCmdlineConfigMap());
```
deadeyegoodwin, line 1731: This line is where all the AddBackendConfig commands must be completed before. Can you move the AddBackendConfig calls and this line does to after the other option setting (line1740) and put a comment indicating as much.

deadeyegoodwin (changes requested):

src/core/tritonserver.cc
CoderHam: Done

deadeyegoodwin (approved):

PR #3319, opened 2021-09-02
Update DCGM version and add better error logging

jbkyang-nvi:
Update DCGM to 2.2.9 and add better error logging functionalities. 
Also adjusted installation method to be same as [build.py](https://github.com/triton-inference-server/server/blob/main/build.py)
Related PR: https://github.com/triton-inference-server/model_analyzer/pull/206

### DIFF
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -42,7 +42,7 @@ FROM ${BASE_IMAGE}
 ENV DEBIAN_FRONTEND=noninteractive
 
 # DCGM version to install for Model Analyzer
-ARG DCGM_VERSION=2.0.13
+ARG DCGM_VERSION=2.2.9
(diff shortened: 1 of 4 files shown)


deadeyegoodwin (changes requested):
We also install dcgm in Dockerfile.sdk so version needs update there as well. Also look to see if there is any other place we use it.

jbkyang-nvi:
> We also install dcgm in Dockerfile.sdk so version needs update there as well. Also look to see if there is any other place we use it.

@deadeyegoodwin  Dockerfile.sdk's DCGM is for the model analyzer. Updated both

Tabrizian (approved):

deadeyegoodwin (approved):

PR #3324, opened 2021-09-03
Fix perf test warmup to ensure all model instances ready

deadeyegoodwin:


tanmayv25:
I don't think it fixes the actual problem. I believe with my changes in core the subsequent requests will run on the model instances in a round-robin fashion. The available instances are kept in a deque and the instance at the front is allocated to the payload. https://github.com/triton-inference-server/server/blob/main/src/core/rate_limiter.cc#L242.
And after execution the instance is pushed to the back of the available queue.https://github.com/triton-inference-server/server/blob/main/src/backends/backend/triton_model_instance.cc#L686

So, even for a single concurrency all the instances should ideally be warmed up.

tanmayv25:
However, using the higher concurrency, would create that many Payload objects, which would not have been created for single concurrency. So, I see how this might be useful.

tanmayv25 (approved):

qa/L0_perf_resnet/run_test.sh
```diff
@@ -69,16 +70,21 @@ if [ "$ARCH" == "aarch64" ]; then
     fi
 else
     PERF_CLIENT=../clients/perf_client
+    if [ "$MODEL_FRAMEWORK" == "savedmodel_trt" ] || [ "$MODEL_FRAMEWORK" == "onnx_trt" ]; then
+        MEASUREMENT_WINDOW=20000
```
tanmayv25, line 74: Instead of increasing measurement windows for such model, how about adding a mode in perf analyzer that sends the request to the model and wait for response before starting the actual profiling. 
This will reduce the overall runtime.  There will be no need to run 3-4 cycles of 20 sec (runtime: 60s) as the same can take 20sec of init time and 15sec(3 cycles of default 5sec) of profiling (runtime: 45sec)

tanmayv25 (approved):

qa/L0_perf_resnet/run_test.sh
CoderHam: @tanmayv25 can you also try to use 20000 as the measurement window for onnx and onnx-trt?
Additionally can you check if this slowdown impacted aarch64 as well?

CoderHam (changes requested):

qa/L0_perf_resnet/run_test.sh
tanmayv25: I think the PR is already trying to use 20000 sec measurement window for onnx-trt,
Do you see failures on aarch64 on CI?

CoderHam: You are right. I missed that. I do however see aarch64 failures for perf resnet on master because of onnx-trt even with 20000 ms as measurement window. This occurs even with concurrency 1.

PR #3329, opened 2021-09-07
Add test for zero length bytes tensor in Python backend

Tabrizian:


deadeyegoodwin (dismissed):

Tabrizian:
Just rebased to pick up some test changes.

deadeyegoodwin (approved):

PR #3338, opened 2021-09-08
Allow changing metrics update frequency by specifying metrics-interval-secs

aramesh7:


qa/L0_metrics/test.sh
```diff
@@ -94,6 +94,33 @@ kill $SERVER_PID
...
+  cat $SERVER_LOG
+  exit 1
+fi
+
+num_iterations=10
+prev_energy=`curl -s localhost:8002/metrics | awk '/nv_energy_consumption{/ {print $2}'`
+for (( i = 0; i < $num_iterations; ++i )); do
+  sleep $METRICS_INTERVAL_SECS
+  current_energy=`curl -s localhost:8002/metrics | awk '/nv_energy_consumption{/ {print $2}'`
+  if [ $current_energy == $prev_energy ]; then
```
deadeyegoodwin, line 112: This seems like a very unstable way to test this. If energy just happens to be the same across 2 samples then it will fail. Or what stops you from sening all 10 curls in < 0.5 seconds and then you will definitely get the same value.

Tabrizian, line 112: What if the updated value is the same as previous value?

src/core/model_config_utils.h
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
Tabrizian, line 1: Remove `(c)` to avoid copyright wrap.

src/servers/signal.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2020-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
Tabrizian, line 1: Same

qa/L0_metrics/test.sh
aramesh7: The assumption is that the energy consumed will increase over time, monotonically

src/core/metrics.cc
```diff
@@ -188,6 +188,13 @@ Metrics::EnableGPUMetrics()
   singleton->gpu_metrics_enabled_ = true;
 }
 
+void
+Metrics::SetMetricsInterval(float metrics_interval_secs)
```
deadeyegoodwin, line 192: Use uint64_t and milliseconds

src/core/model_config_utils.h
aramesh7: This change came from the formatter I think

src/servers/signal.cc
aramesh7: Sam as above

src/core/metrics.cc
```diff
@@ -355,13 +362,15 @@ Metrics::InitializeDcgmMetrics()
...
-          handle, groupId, fieldGroupId, 2000000 /*update period, usec*/,
+          handle, groupId, fieldGroupId,
+          (long long)(metrics_interval_secs_ * 1e6) /*update period, usec*/,
           5.0 /*maxKeepAge, sec*/, 5 /*maxKeepSamples*/);
       if (dcgmerr != DCGM_ST_OK) {
         LOG_WARNING << "Cannot start watching fields: " << errorString(dcgmerr);
       } else {
         while (!dcgm_thread_exit_.load()) {
-          std::this_thread::sleep_for(std::chrono::milliseconds(2000));
+          std::this_thread::sleep_for(std::chrono::milliseconds(
```
deadeyegoodwin, line 372: You only need to update at 2x the sample rate you use above to avoid missing readings. So this should be 0.5 * the interval.

src/core/model_config_utils.h
deadeyegoodwin, line 1: Remove the '(c)' to avoid line wrapping.

src/core/tritonserver.cc
```diff
@@ -1706,6 +1725,7 @@ TRITONSERVER_ServerNew(
   if (loptions->Metrics()) {
     ni::Metrics::EnableMetrics();
   }
+  ni::Metrics::SetMetricsInterval(loptions->MetricsInterval());
```
deadeyegoodwin, line 1728: This can be inside the above conditional

src/servers/signal.cc
deadeyegoodwin, line 1: Remove (c) to avoid line wrap

deadeyegoodwin (changes requested):

qa/L0_metrics/test.sh
aramesh7: From the tests I've run, the energy seems to be be increasing pretty quickly, it would have to be a very low power GPU for it to fail. 

I didn't get the question regarding all 10 curls in < 0.5 seconds. I sleep the test thread for 0.5ms between curls.

deadeyegoodwin: Sorry, I missed the sleep...
I guess this is ok.

aramesh7: If we had timestamps, i think we could make this better, but I didn't see any.

src/core/metrics.cc
```diff
@@ -355,13 +362,15 @@ Metrics::InitializeDcgmMetrics()
         LOG_WARNING << "Cannot make field group: " << errorString(dcgmerr);
       }
       dcgmerr = dcgmWatchFields(
-          handle, groupId, fieldGroupId, 2000000 /*update period, usec*/,
+          handle, groupId, fieldGroupId,
+          (long long)(metrics_interval_ms_ * 1e3) /*update period, usec*/,
```
deadeyegoodwin, line 366: Should no longer need cast, change 1e3 to 1000

src/core/metrics.h
```diff
@@ -173,6 +176,7 @@ class Metrics {
   bool metrics_enabled_;
   bool gpu_metrics_enabled_;
   std::mutex gpu_metrics_enabling_;
+  uint64_t metrics_interval_ms_;
```
deadeyegoodwin, line 179: Set to default value in constructor.

src/servers/main.cc
```diff
@@ -410,6 +412,9 @@ std::vector<Option> options_
        "--allow-metrics is true."},
       {OPTION_METRICS_PORT, "metrics-port", Option::ArgInt,
        "The port reporting prometheus metrics."},
+      {OPTION_METRICS_INTERVAL_MS, "metrics-interval-ms", Option::ArgFloat,
+       "Metrics will be collected once every <metrics-interval-ms> "
+       "milliseconds."},
```
deadeyegoodwin, line 417: "Default is 2000 milliseconds."

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

src/servers/http_server.cc
```diff
@@ -150,7 +150,6 @@ HTTPMetricsServer::Handle(evhtp_request_t* req)
   evhtp_send_reply(req, res);
 }
 
-#ifdef TRITON_ENABLE_METRICS
```
CoderHam, line 153: Why are we removing this ifdef?

CoderHam (changes requested):

src/servers/http_server.cc
aramesh7: It seems to be duplicated

GuanLuo (approved):

CoderHam (approved):

PR #3346, opened 2021-09-09
Fix long-running stress test

krishung5:


qa/L0_long_running_stress/stress.py
```diff
@@ -634,13 +643,7 @@ def stress_thread(name, seed, test_duration, correlation_id_base,
 
                 choice = rng.rand()
 
-                # no-start cannot follow no-end since the server will
-                # just assume that the no-start is a continuation of
-                # the no-end sequence instead of being a sequence
-                # missing start flag.
-                if ((last_choice != "sequence_no_end") and
```
deadeyegoodwin, line 641: I think that this code is not longer working as before because you added non-sequence cases below that set last_choices. You should rename last_choices to last_seq_choices and set it only for the sequence cases and then this original logic should be ok. You will also need a 'last_choice that you set everywhere that you can use when calling count_failed_test_case.

deadeyegoodwin (changes requested):

qa/L0_long_running_stress/stress.py
```diff
@@ -109,6 +109,15 @@ def check_sequence_async(client_metadata,
...
     sequence_id = client_metadata[1]
 
+    # Need to use a unique ID for 'sequence_no_start' case since we
+    # need to prevent sequences with no 'start' flags from following
+    # some sequences with no 'end' flags since the server will just
+    # assume that the no-start is a continuation of the no-end sequence
+    # instead of being a sequence missing start flag, which gives results
+    # not expected.
+    if test_case_name == "sequence_no_start":
+        sequence_id *= 10
```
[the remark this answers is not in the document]
krishung5: Will remove this part and use the original approach mentioned below.

qa/L0_long_running_stress/stress.py
krishung5: I see, will make the changes.

deadeyegoodwin (dismissed):

CoderHam (approved):

CoderHam (approved):

krishung5:
@deadeyegoodwin I think the original logic doesn't work well is because in the original sequence stress test, we only serve one model, so we can just use the last_seq_choice to see if the last sequence case is some no-end cases. But we are serving multiple models in this stress test, so we need to remember which model is used for the no-end cases.

deadeyegoodwin (approved):

PR #3351, opened 2021-09-10
provide the link to the devblog and scripts for autoscaling Triton deployment with Kubernetes (on MIG) and load balacing using NGNIX Plus.

maggiezha:
This repo provides the scripts for autoscaling Triton deployment with Kubernetes (on MIG) and load balacing using NGNIX Plus.

For step by step instructions, please refer to this developer blog:

* [Deploying NVIDIA Triton at Scale with MIG and Kubernetes](https://developer.nvidia.com/blog/deploying-nvidia-triton-at-scale-with-mig-and-kubernetes/).


deploy/K8s-onprem-MIG/Readme
```diff
@@ -0,0 +1,5 @@
+This repo provides the scripts for autoscaling Triton deployment with Kubernetes (on MIG) and load balacing using NGNIX Plus.
```
deadeyegoodwin, line 1: Every file needs nvidia copyright. You can find examples in the other deploy/ subdirectories

deadeyegoodwin, line 1: This directory should follow existing naming conventions, so no caps. Also, why is this onprem specific. Also, why is it MIG specific. It seems that mig is only mentioned in one resource limit... but couldn't that instead just be a variable that users can set to deploy to many different types of resources. Certainly you can call out how this example can be used to deploy to mig but it doesn't need to be only mig.

The fundamental contribution seems to be that it autoscales based on some metrics and that it uses nginx. So can we name directory nginx_autoscale?

deploy/K8s-onprem-MIG/custom-metrics-server-config.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: adapter-config
+  namespace: monitoring
+data:
+  triton-adapter-config.yaml: |
+    rules:
+    - seriesQuery: 'nv_inference_queue_duration_us{namespace="default",pod!=""}'
```
deadeyegoodwin, line 9: So you are scaling on queue duration. What if there are multiple models?

deploy/K8s-onprem-MIG/flower-hpa.yml
```diff
@@ -0,0 +1,18 @@
...
+      name: flower
+    minReplicas: 1
+    maxReplicas: 7
+    metrics:
+    # use a "Pods" metric, which takes the average of the
+    # given metric across all pods controlled by the autoscaling target
+    - type: Pods
+      pods:
+        metricName: avg_time_queue_ms
+        targetAverageValue: 50
```
deadeyegoodwin, line 18: Seems very model specific. Can all the fixed values become variables and then the README can describe what they mean.

deploy/K8s-onprem-MIG/flower-replicas3.yml
```diff
@@ -0,0 +1,41 @@
...
+      containers:
+        - name: flower
+          ports:
+          - containerPort: 8000
+            name: http-triton
+          - containerPort: 8001
+            name: grpc-triton
+          - containerPort: 8002
+            name: metrics-triton
+          image: "nvcr.io/nvidia/tritonserver:20.12-py3"
```
deadeyegoodwin, line 32: Should use latest version of Triton.

deploy/K8s-onprem-MIG/flower-service-nginx.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: flower-nginx
+  labels:
+    app: flower
```
deadeyegoodwin, line 6: So this is the flower demo? Isn't there a client application that is needed to drive the entire demo?

deadeyegoodwin (changes requested):

deploy/K8s-onprem-MIG/Readme
maggiezha: added to each file

deploy/K8s-onprem-MIG/flower-hpa.yml
maggiezha: Explained in Readme that users can define their own custom metric for HPA.

deploy/K8s-onprem-MIG/flower-replicas3.yml
maggiezha: changed.

deploy/K8s-onprem-MIG/Readme
maggiezha: I tried to rename it but seems I don't have the permission, could you please help rename it to hpa_autoscale, or hpa_autoscale_nginx? Thanks

maggiezha:
Made a few changes as requested: copyright, Triton version, GPU resource, max/min Replica. Also explained to the users about GPU resource, Replica number, HPA custom metric etc.

maggiezha (commented):
thanks

deploy/K8s-onprem-MIG/custom-metrics-server-config.yaml
maggiezha: multiple models are not used in this demo, but explained by River in the email

deadeyegoodwin: This cannot just be a demo. As discussed in the email this must be a general example that shows auto-scaling. If you are just looking for a place to host the exact example used in the blog then this is not the proper location.

PR #3363, opened 2021-09-13
Update SDK dockerfile to support ARM64

deadeyegoodwin:


Dockerfile.sdk
```diff
@@ -38,6 +38,8 @@ ARG TRITON_ENABLE_GPU=ON
 
 FROM ${BASE_IMAGE}
 
+ARG TARGETPLATFORM
```
CoderHam, line 41: TARGET_PLATFORM to make it consistent with the way others ARGs are named?

Dockerfile.sdk
```diff
@@ -80,7 +82,7 @@ RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/nul
       apt-add-repository 'deb https://apt.kitware.com/ubuntu/ focal main' && \
       apt-get update && \
       apt-get install -y --no-install-recommends \
-        cmake-data=3.18.4-0kitware1ubuntu20.04.1 cmake=3.18.4-0kitware1ubuntu20.04.1; \
+        cmake-data=3.21.1-0kitware1ubuntu20.04.1 cmake=3.21.1-0kitware1ubuntu20.04.1; \
```
CoderHam, line 85: Move to self to make similar change in jetson build script (if possible) instead of building from source.

CoderHam (changes requested):
One small fix else LGTM

Dockerfile.sdk
deadeyegoodwin: This variable comes from "docker buildx" so we don't control it name.

CoderHam (dismissed):

build.py
```diff
@@ -912,8 +928,6 @@ def container_build(images, backends, repoagents, endpoints):
         base_image = images['base']
     elif target_platform() == 'windows':
         base_image = 'mcr.microsoft.com/dotnet/framework/sdk:4.8'
-    elif target_platform() == 'ubuntu/arm64':
```
CoderHam, line 915: Don't we want the aarch64 base image here when platform is ARM?

CoderHam (changes requested):

build.py
deadeyegoodwin: The base image is now a manifest that points to both the amd64 and arm64 images. Docker will pull the appropriate one based on the host platform

PR #3375, opened 2021-09-17
[DO NOT MERGE] Updated sequence extensions docs to support sequence_id as a string

aramesh7:


docs/protocol/extension_sequence.md
```diff
@@ -35,24 +35,26 @@ supported, Triton reports “sequence” in the extensions field of its
 Server Metadata.
```
deadeyegoodwin, line 35: You should add sentences about "sequence(string_id)" extension and note what it adds. Similar to what is done in the model_repository extension. https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_model_repository.md#model-repository-extension

deadeyegoodwin (changes requested):

docs/protocol/extension_sequence.md
aramesh7: Added

docs/protocol/extension_sequence.md
```diff
@@ -28,31 +28,35 @@
...
 
 An inference request can specify that it is part of a sequence using
 the “sequence_id” parameter in the request and by using the
 “sequence_start” and “sequence_end” parameters to indicate the start
 and end of sequences.
 
-- "sequence_id" : uint64 value that indicates which sequence a request
-  belongs to. All inference requests that belong to the same sequence
-  must use the same sequence ID. A sequence ID of 0 indicates the
+Because this extension is supported, Triton reports either "sequence(uint64)"
```
deadeyegoodwin, line 40: You can change the name of this extension to not be "sequence" as that would break backwards compatibility. So you need to keep "sequence" and add "sequence(<name>)" where <name> describes the new functionality you are adding. Just "string" is likely too generic. So perhaps "sequence(string_id)".

deadeyegoodwin (changes requested):

docs/protocol/extension_sequence.md
```diff
@@ -28,31 +28,36 @@
...
 
 An inference request can specify that it is part of a sequence using
 the “sequence_id” parameter in the request and by using the
 “sequence_start” and “sequence_end” parameters to indicate the start
 and end of sequences.
 
-- "sequence_id" : uint64 value that indicates which sequence a request
-  belongs to. All inference requests that belong to the same sequence
-  must use the same sequence ID. A sequence ID of 0 indicates the
+Because this extension is supported, Triton reports either "sequence"
```
deadeyegoodwin, line 40: Sorry to be pedantic... but "sequence" must always be reported if any of sequence extension is supported. *Additionally* "sequence(string_id)" can be reported if server also supports sequence IDs as strings.

deadeyegoodwin (changes requested):

docs/protocol/extension_sequence.md
aramesh7: Oh i see. I should have understood that, apologies.

deadeyegoodwin (approved):

PR #3387, opened 2021-09-21
Update entrypoint to support arm64

deadeyegoodwin:


nvidia_entrypoint.sh
```diff
@@ -69,7 +69,7 @@ else
   fi
 fi
 
-if ! cat /proc/cpuinfo | grep flags | sort -u | grep avx >& /dev/null; then
```
CoderHam, line 72: Add comment about AVX not being supported on ARM/aarch64

CoderHam (changes requested):

PR #3411, opened 2021-09-27
Add support for implicit state

Tabrizian:


deadeyegoodwin (commented):
Can  you update based on the TRITONBACKEND API changes I requested and then I will review.

src/backends/backend/triton_model.cc
```diff
@@ -564,6 +564,67 @@ TRITONBACKEND_RequestRelease(
...
+    TRITONBACKEND_State** state, TRITONBACKEND_Request* request,
+    const char* name, const TRITONSERVER_DataType datatype,
+    const int64_t* shape, const uint32_t dims_count)
+{
+  InferenceRequest* tr = reinterpret_cast<InferenceRequest*>(request);
+  std::vector<int64_t> lshape(shape, shape + dims_count);
+  InferenceRequest::State* lstate;
+  Status status =
+      tr->AddState(name, TritonToDataType(datatype), lshape, &lstate);
+  *lstate->MutableShape() = lshape;
```
deadeyegoodwin, line 597: Why do you need to call this, doesn't AddState set the shape? In any case it should be after the status check.

src/backends/backend/triton_model.cc
```diff
@@ -564,6 +564,67 @@ TRITONBACKEND_RequestRelease(
...
+TRITONSERVER_Error*
+TRITONBACKEND_StateBuffer(
+    TRITONBACKEND_State* state, void** buffer, const uint64_t buffer_byte_size,
+    TRITONSERVER_MemoryType* memory_type, int64_t* memory_type_id)
+{
+  InferenceRequest::State* to =
+      reinterpret_cast<InferenceRequest::State*>(state);
+  std::shared_ptr<AllocatedMemory> memory = std::make_shared<AllocatedMemory>(
+      buffer_byte_size, *memory_type, *memory_type_id);
+  *buffer = memory->MutableBuffer(memory_type, memory_type_id);
```
deadeyegoodwin, line 617: Shouldn't the state just have a persistent buffer that is used throughout the sequence lifetime... and only change when it needs to grow?

src/core/infer_request.h
```diff
@@ -194,6 +194,27 @@ class InferenceRequest {
     std::map<std::string, std::shared_ptr<Memory>> host_policy_data_map_;
   };
 
+  // State tensor
```
deadeyegoodwin, line 197: State itself shouldn't live inside a request, should it? A State is associated with a sequence and so likely should live inside the sequence batcher. It seems like a request should just have a reference to a "sequence object" and whenever need to get a state object it should go though that sequence object. For now the sequence object may just hold maps of input_name->State and output_name ->State

deadeyegoodwin (changes requested):

src/core/sequence_batch_scheduler.cc
```diff
@@ -827,6 +843,82 @@ SequenceBatch::SetControlTensors(
...
+        input->SetData(data);
+
+        // Insert a dummy state for the first request in the sequence.
+        const auto& pair = states->emplace(
+            std::piecewise_construct,
+            std::forward_as_tuple(state_config.output_name()),
+            std::forward_as_tuple(new InferenceRequest::State(
+                state_config.output_name(), state.second.data_type(), dims)));
+        pair.first->second->SetData(data);
+        *pair.first->second->MutableShape() = dims;
```
tanmayv25, line 891: This can go within the the InferenceRequest::State constructor.

tanmayv25 (changes requested):

src/backends/backend/triton_model.cc
Tabrizian: I'll modify this. Because State is based on input. The shape will be added to `original_shape_` instead of `shapes_` attribute.

src/core/infer_request.h
Tabrizian: The state temporarily lives inside the request before `TRITONBACKEND_UpdateState` is called. There are state slots in the `SequenceBatch` object that the state will be transferred to after `TRITONBACKEND_UpdateState` is called. I will modify this according to your comment.

src/core/sequence_batch_scheduler.cc
```diff
@@ -827,6 +843,82 @@ SequenceBatch::SetControlTensors(
...
+        std::vector<int64_t> dims;
+        for (auto& dim : state_config.dims()) {
+          if (dim == -1) {
+            dims.push_back(1);
+          } else {
+            dims.push_back(dim);
+          }
+        }
+
+        // Replace the dynamic shapes with 1 for the first inference request.
```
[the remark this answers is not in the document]
Tabrizian: This is so that the model can know beforehand what shape to expect for the first inference request for the implicit state. This was part of the proposal.

> Update comment explaining why this won't be an issue.

You mean add a comment explaining why the garbage data won't be an issue?

src/backends/backend/triton_model.cc
Tabrizian: In the current implementation it will be completely replaced with the new buffer. I'll modify it so that memory is allocated only when needed.

src/core/sequence_batch_scheduler.cc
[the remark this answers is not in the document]
tanmayv25: You are right. I missed it. Make sure we document this somewhere.

src/core/sequence_batch_scheduler.cc
```diff
@@ -59,12 +60,23 @@ SequenceBatchScheduler::Create(
...
   // Max sequence idle...
   sched->max_sequence_idle_microseconds_ =
       config.sequence_batching().max_sequence_idle_microseconds();
 
+  // Implicit States
+  auto& states = config.sequence_batching().state();
+  sched->has_implicit_state_ = states.size() > 0;
+
+  if (sched->has_implicit_state_) {
+    for (const inference::ModelSequenceBatching_State& state : states) {
```
deadeyegoodwin, line 74: Redundant? If 'states' is size 0 this won't iterator.

src/core/sequence_state.h
```diff
@@ -0,0 +1,107 @@
+// Copyright 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
```
deadeyegoodwin, line 4: The sequence state logic seems unnecessrily distributed across infer_request, triton_model, sequence_batcher. Can you instead add a SequenceStates abstraction here that along with State (perhaps renamed to SequenceState so it is not so general) can hold all that logic for creating, managing, etc. sequence state. Then, in the other places you will just hold references to SequenceStates and call into methods in that class to perform all the state related functions.

deadeyegoodwin (changes requested):

PR #3412, opened 2021-09-28
update ort version and add tests

askhade:
Update ORT version to 1.9 and add tests for newly added optimization configs for CUDA and CPU provider

### DIFF
--- a/qa/L0_onnx_optimization/test.sh
+++ b/qa/L0_onnx_optimization/test.sh
@@ -60,6 +60,20 @@ for MODEL in \
     rm -fr models/${MODEL}_test/3 && \
     (cd models/${MODEL}_test && \
             sed -i 's/_float32_float32_float32/&_test/' config.pbtxt) && \
+    # CUDA EP optimization params
+    cp -r models/${MODEL}_test models/${MODEL}_cuda_config && \
+    (cd models/${MODEL}_cuda_config && \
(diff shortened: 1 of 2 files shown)


deadeyegoodwin (commented):
Looks good. Will approve after CI

qa/L0_onnx_optimization/test.sh
```diff
@@ -60,6 +60,20 @@ for MODEL in \
     rm -fr models/${MODEL}_test/3 && \
     (cd models/${MODEL}_test && \
             sed -i 's/_float32_float32_float32/&_test/' config.pbtxt) && \
+    # CUDA EP optimization params
+    cp -r models/${MODEL}_test models/${MODEL}_cuda_config && \
```
deadeyegoodwin, line 64: You've changed the directory name from L0_onnxtrt_optimization to L0_onnx_optimization... is that intentional? If so I need to adjust on CI side as well.

deadeyegoodwin (changes requested):

qa/L0_onnx_optimization/test.sh
askhade: Yes it is intentional. I made this change because now I added tests for CUDA provider as well... so this is not more a onnx-trt dedicated test.

PR #3427, opened 2021-10-01
Set rate limiter priority to default to 1

tanmayv25:


qa/L0_rate_limiter/test.sh
```diff
@@ -66,7 +66,7 @@ cp -r ./custom_models/custom_zero_1_float32 ./custom_models/custom_zero_1_float3
         sed -i "s/max_batch_size:.*/max_batch_size: 4/g" config.pbtxt && \
         echo "instance_group [{"  >> config.pbtxt && \
         echo "kind: KIND_GPU count: 1"  >> config.pbtxt && \
-        echo "rate_limiter { resources [{name: \"resource1\" count: 4 }] priority: 1}"  >> config.pbtxt && \
```
CoderHam, line 69: In the model config protobuf, we should call out priority as an optional field.

src/core/rate_limiter.cc
```diff
@@ -674,7 +674,7 @@ RateLimiter::ModelInstanceContext::ScaledPriority()
 {
   // TODO: Different schemes for the prioritization of
   // model instance can be added here.
-  return (exec_count_ * rate_limiter_config_.priority());
+  return (exec_count_ * std::max(rate_limiter_config_.priority(), 1u));
```
CoderHam, line 677: Add a comment about priority taking default value of 0 when not specified in model config.

CoderHam (changes requested):

src/core/rate_limiter.cc
deadeyegoodwin: Default value of 1 ?

deadeyegoodwin (approved):

PR #3432, opened 2021-10-05
Fix test REPO_VERSION to allow optional TEST_REPO_ARCH suffix

deadeyegoodwin:


deadeyegoodwin:
> 
> 
> Do you want to make the changes to the jetson test scripts on gitlab to adopt these changes or should I?

@CoderHam It would be good if you could make the changes.

CoderHam:
> > Do you want to make the changes to the jetson test scripts on gitlab to adopt these changes or should I?
> 
> @CoderHam It would be good if you could make the changes.

On it

PR #3476, opened 2021-10-15
Updated L0_grpc and L0_http to use string sequence ID examples

aramesh7:


aramesh7:
> When adding models to the model repo there is a check for the number of models in the grpc and http model control scripts. You will need to increment the counts there accordingly. The examples are in the client repo so you should create a corresponding PR for that repo

This has been done and merged into the clients side

PR #3516, opened 2021-10-29
Add testing for implicit state in TensorRT backend

Tabrizian:


deadeyegoodwin (approved):

src/backends/backend/triton_model.cc
```diff
@@ -170,9 +170,11 @@ TritonModel::Create(
   }
   local_model->initialized_ = true;
 
-  const bool device_blocking =
-      (local_model->backend_->ExecutionPolicy() ==
-       TRITONBACKEND_EXECUTION_DEVICE_BLOCKING);
+  // NOTE, we don't use device blocking transaction policy for
+  // sequence batching models.
```
GuanLuo, line 174: What is the reasoning? I thought we had sequence test and that was working fine.

tanmayv25: This is specifically for implicit state management. It greatly simplifies the logic to update state per runner(backend thread) and get to execute the next instance quickly.
Interestingly, I also observed better performance without device blocking on other sequence models. Hence, sticking to the older method of having per instance runner for sequence model.

GuanLuo: Okay, then I think we should log / document this as the policy will not be device blocking in this case and the backend implementer may assume it is still device blocking

GuanLuo (approved):

PR #3526, opened 2021-11-02
Fix entrypoint banner for SDK container

rmccorm4:


Dockerfile.sdk
```diff
@@ -168,5 +168,13 @@ ARG TRITON_MODEL_ANALYZER_REPO_TAG
 ARG TRITON_MODEL_ANALYZER_REPO="https://github.com/triton-inference-server/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}"
 RUN pip3 install "git+${TRITON_MODEL_ANALYZER_REPO}"
 
+# Entrypoint Banner
+ARG NVIDIA_TRITON_INFERENCE_SERVER_SDK_VERSION=21.10
```
deadeyegoodwin, line 172: We don't want to hardwire this verision. Can you instead get it from the TRITON_VERSION file?

deadeyegoodwin (changes requested):

Dockerfile.sdk
```diff
@@ -168,5 +168,13 @@ ARG TRITON_MODEL_ANALYZER_REPO_TAG
...
+ARG NVIDIA_BUILD_ID=unknown
```
CoderHam, line 173: Is the convention to use 'unknown' as the default build id?

Dockerfile.sdk
rmccorm4: Will be passed in from CI, changed 21.10 default to `unknown`

Dockerfile.sdk
rmccorm4: I took `unknown` as the default from our `build.py` convention: https://github.com/triton-inference-server/server/blob/400cc0cf7287de8f19c227442f1b98f5abe86e55/build.py#L959

deadeyegoodwin (approved):

PR #3549, opened 2021-11-09
Remove documentation on composing CPU-only images

deadeyegoodwin:
We are not yet publishing the CPU-only full image, so composition
cannot work without that.


jbkyang-nvi:
Thanks for the change David! Can you maybe merge this with `kyang-fix-readme` since I made #3550

PR #3556, opened 2021-11-10
Disable int8_int32_int32 for pytorch due to https://github.com/pytorch/pytorch/issues/66930

deadeyegoodwin:


CoderHam (approved):
We should remember to relnote this as well

deadeyegoodwin:
> We should remember to relnote this as well

I filed a ticket to re-enable. I'll file another to make sure we relnote in 21.12 if it doesn't get fixed.

PR #3568, opened 2021-11-12
Update test docs to describe sanity testing

deadeyegoodwin:


docs/test.md
```diff
@@ -79,19 +79,47 @@ Lastly, build the actual QA image.
...
+are parameterized so that you can run them even if you have built a
+Triton that contains only a subset of Triton backends. These tests are
+L0_infer, L0_batcher and L0_sequence_batcher. For these tests the
+following envvars are available to control how the tests behave:
+
+* BACKENDS: Control which backends are tested. Look in the test.sh
+  file of the test to see the default and allowed values.
+
+* ENSEMBLES: Enable testing of ensembles. Set to "0" to disable, set
+  to "1" to enable. If enabled you must have the *identity* backend
```
CoderHam, line 112: We need sequence backend for sequence batcher test if "custom" backend is specified

docs/test.md
```diff
@@ -79,19 +79,47 @@ Lastly, build the actual QA image.
...
+* EXPECTED_NUM_TESTS: The tests perform a check of the total number of
+  test sub-cases. The exact number of sub-cases that run will depend
+  on the values you use for BACKENDS and ENSEMBLES. So you will need
+  to adjust this as appropriate for your testing.
+
+For example, if you build a Triton that has only the TensorRT backend
+you can run L0_infer as follows:
+
+```
+$ BACKENDS="plan" ENSEMBLES=0 bash -x ./test.sh
```
CoderHam, line 124: Should give `EXPECTED_NUM_TESTS` in the example as well since it will not be the usual number if only TRT is enabled.

CoderHam (changes requested):

docs/test.md
deadeyegoodwin: The sequence backend is always built as part of the test-utils for the QA container, isn't it?

docs/test.md
deadeyegoodwin: I don't know what the correct value is and it may change over time. But I will add some additional text.

docs/test.md
CoderHam: You are right. Resolving

CoderHam (approved):

PR #3582, opened 2021-11-18
Refactor Testing: Unused Variables, Imprecise Asserts

dyastremsky:
A small refactor of some of the testing and models to get rid of unused variables, imprecise asserts, unnecessary imports, and non-recommended syntax.

### DIFF
--- a/qa/L0_rate_limiter/rate_limiter_test.py
+++ b/qa/L0_rate_limiter/rate_limiter_test.py
@@ -26,7 +26,6 @@
 
 import sys
 
-from tritonclient.grpc.model_config_pb2 import ModelEnsembling
 sys.path.append("../common")
 
 import functools
@@ -35,8 +34,6 @@ import os
 import unittest
 import threading
 import time
-import infer_util as iu
-import test_util as tu
 import sequence_util as su
 import tritongrpcclient as grpcclient
 from tritonclientutils import *
@@ -232,7 +229,7 @@ class RateLimiterTest(su.SequenceBatcherTestUtil):
     def test_single_model_dynamic_batching(self):
         # Send all the inference requests with a delay to a model
 
-        self.assertTrue("TRITONSERVER_DELAY_SCHEDULER" in os.environ)
+        self.assertIn("TRITONSERVER_DELAY_SCHEDULER", os.environ)
         model_names = ["custom_zero_1_float32"]
         infer_counts = self.stress_models(model_names, delay_ms=100)
 
@@ -252,7 +249,6 @@ class RateLimiterTest(su.SequenceBatcherTestUtil):
             self.assertEqual(
                 batch_stat.batch_size, 4,
                 "unexpected batch-size {}".format(batch_stat.batch_size))
-            bc = batch_stat.compute_infer.count
             # Get count from one of the stats
             self.assertEqual(
                 batch_stat.compute_infer.count, _inference_count / 4,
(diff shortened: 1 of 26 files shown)


tanmayv25:
Looks good to me. But @GuanLuo can review mlflow side changes. 

@dyastremsky  Did you use any tool to detect the issues? Can you do something similar for Python backend?

dyastremsky:
@tanmayv25 Yup! I saw someone put our repo into LGTM, so I refactored some of the higher-significance alerts. Just added the python_backend repo, it might take some time to analyze.

LGTM link here: https://lgtm.com/projects/g/triton-inference-server/server/

qa/common/nightly_email_helper.py
```diff
@@ -39,7 +39,7 @@
 
 def send(subject: str,
          content: str,
-         attachments=[],
```
CoderHam, line 42: Why is this change needed?

CoderHam (changes requested):
Good work with the cleanup. Is the general convention to not have [] as a default value for params?

qa/common/nightly_email_helper.py
```diff
@@ -39,7 +39,7 @@
```
CoderHam, line 39: 'base64' is not used. You should be able to remove that import

CoderHam (changes requested):

qa/common/nightly_email_helper.py
dyastremsky: It's because of the attachments.append. Default arguments are instantiated once, not once per call. By using .append, the default argument is changed. So the behavior would vary for every call (i.e. once a new attachment is appended when using the default value, that attachment would already be appended for the next call with the default attachments value).

qa/common/reporter.py
```diff
@@ -139,7 +139,8 @@ def post_to_url(url, data):
```
CoderHam, line 139: Import of 'sys' is not used.

CoderHam (changes requested):

CoderHam:
> @dyastremsky Did you use any tool to detect the issues? Can you do something similar for Python backend?

@tanmayv25 Similar tool for python backend
https://lgtm.com/projects/g/triton-inference-server/python_backend?mode=list

We should do a similar run through the client repo

qa/common/nightly_email_helper.py
CoderHam: TIL 😄

qa/common/reporter.py
dyastremsky: Removed.

qa/common/nightly_email_helper.py
dyastremsky: Removed.

CoderHam (dismissed):

dyastremsky:
Thank you both! Addressed the feedback regarding other imports to get rid of. There will likely be more later on, as I only addressed a hundred or so of the higher-priority alerts. Didn't want to do them all at once, as I'm already trying to figure out why [rate_limiter--base is failing](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/29411262), so figured it might make sense to split into a couple PRs.

GuanLuo (approved):

tanmayv25 (dismissed):

qa/L0_rate_limiter/rate_limiter_test.py
```diff
@@ -267,7 +263,7 @@ def test_single_model_sequence_batching(self):
 
         try:
             model_name = "custom_sequence_int32"
-            self.assertFalse("TRITONSERVER_DELAY_SCHEDULER" in os.environ)
```
tanmayv25, line 270: Note: this was assertFalse.. We are checking whether TRITONSERVER_DELAY_SCHEDULER was properly unset before running this test.

dyastremsky: Thanks for finding this!

CoderHam (approved):

tanmayv25 (approved):

PR #3598, opened 2021-11-22
Support tracing tensors in triton

ZhuYuJin:
A draft implementation of tracing tensors according to the design doc(https://docs.google.com/document/d/1yL40ctSccNMnbhiAkR-Wg6T0zSaXrfoC/edit) and the PR of triton core(https://github.com/triton-inference-server/core/pull/36).

The output of trace_summary.py is shown as follow.
image
image

Please check whether the implementation is reasonable. @deadeyegoodwin


qa/common/trace_summary.py
```diff
@@ -259,6 +261,117 @@ def summarize(frontend, traces):
...
+                      get_numpy_array(trace["tensor"])))
+            elif trace["id"] == id and trace["activity"] == "8":
+                print("{0}{1}:".format(tabs, "OUTPUT"))
+                print("{0}\t{1}: {2}".format(tabs, trace["tensor"]["name"],
+                      get_numpy_array(trace["tensor"])))
+    print("{0}{1}".format(tabs, "*"*50))
+
+
+TRITON_TYPE_TO_NUMPY = {
+    1: bool,
```
deadeyegoodwin, line 350: Wha are these magic numbers. At the least put comments indicating the corresponding enum name

src/core/dynamic_batch_scheduler.h
```diff
@@ -34,6 +34,7 @@
 #include <queue>
 #include <set>
 #include <thread>
+
```
deadeyegoodwin, line 37: Remove extra lines between includes added here and elsewhere. Are you running the formatter? It should not be adding these lines.

src/core/infer_request.h
```diff
@@ -39,6 +40,10 @@
 #include "src/core/status.h"
 #include "src/core/tritonserver_apis.h"
 
+#ifdef TRITON_ENABLE_GPU
+#include <cuda_runtime_api.h>
```
deadeyegoodwin, line 44: Why do you need this include?

src/core/infer_request.h
```diff
@@ -304,7 +309,12 @@ class InferenceRequest {
   void SetTrace(std::unique_ptr<InferenceTrace>&& trace)
   {
     trace_ = std::move(trace);
+#ifdef TRITON_ENABLE_TRACING
+    response_factory_.SetTrace(std::move(trace_->CopyTrace()));
```
deadeyegoodwin, line 313: If we need to have the trace object be used by both request and response(s), then we should change the trace to be a shared pointer everywhere. We should not be creating copies of the trace object. The potential complication is that the activity collection and other functions may need additional synchronization as the response(s) may be reporting trace activities/tensors interleaved with each other and with the request.

src/core/infer_trace.h
```diff
@@ -52,7 +53,31 @@ class InferenceTrace {
   {
   }
 
+  InferenceTrace(
```
deadeyegoodwin, line 56: Make this the only constructor and change callers to explicitly pass tensor_activity_fn = nullptr when necessary

src/servers/tracer.h
```diff
@@ -31,8 +31,13 @@
 #include <mutex>
 #include <sstream>
 #include <string>
+
 #include "triton/core/tritonserver.h"
 
+#ifdef TRITON_ENABLE_GPU
+#include <cuda_runtime_api.h>
```
deadeyegoodwin, line 38: Why did you add this? I don't see any cuda api usage in this file.

deadeyegoodwin (changes requested):

ZhuYuJin:
@deadeyegoodwin Please check the latest commit.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -155,6 +155,7 @@ DynamicBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& request)
   INFER_TRACE_ACTIVITY(
       request->Trace(), TRITONSERVER_TRACE_QUEUE_START,
       request->QueueStartNs());
+  request->TraceTensor();
```
deadeyegoodwin, line 158: Surround with "#ifdef TRITON_ENABLE_TRACING"

deadeyegoodwin, line 158: Doesn't TraceTensor need an argument indicating at what point you are collecting the tensor. Here should be at "Queue Start".

src/core/infer_response.cc
```diff
@@ -225,6 +230,39 @@ InferenceResponse::SendWithStatus(
   return InferenceResponse::Send(std::move(response), flags);
 }
 
+void
+InferenceResponse::TraceTensor()
```
deadeyegoodwin, line 234: Same comments as for InferenceRequest::TraceTensor.
Plus in both cases you need to include error handing. For now can just output something in the log will be sufficient I think,

src/core/infer_response.h
```diff
@@ -74,6 +76,11 @@ class InferenceResponseFactory {
   // Send a "null" response with 'flags'.
   Status SendFlags(const uint32_t flags) const;
 
+#ifdef TRITON_ENABLE_TRACING
+  const std::shared_ptr<InferenceTrace>& Trace() const { return trace_; }
+  void SetTrace(std::shared_ptr<InferenceTrace> trace) { trace_ = trace; }
```
deadeyegoodwin, line 81: SetTrace(const std::shared_ptr<InferenceTrace>& trace)

src/core/infer_response.h
```diff
@@ -102,6 +109,11 @@ class InferenceResponseFactory {
   // Delegator to be invoked on sending responses.
   std::function<void(std::unique_ptr<InferenceResponse>&&, const uint32_t)>
       response_delegator_;
+
+#ifdef TRITON_ENABLE_TRACING
+  // Inference trace associated with this request.
```
deadeyegoodwin, line 114: ... "associated with this response."

src/core/infer_response.h
```diff
@@ -303,6 +320,11 @@ class InferenceResponse {
       response_delegator_;
 
   bool null_response_;
+
+#ifdef TRITON_ENABLE_TRACING
+  // Inference trace associated with this request.
```
deadeyegoodwin, line 325: ... "associated with this response."

src/core/infer_trace.h
```diff
@@ -46,13 +46,24 @@ class InferenceTrace {
   InferenceTrace(
       const TRITONSERVER_InferenceTraceLevel level, const uint64_t parent_id,
       TRITONSERVER_InferenceTraceActivityFn_t activity_fn,
-      TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* userp)
+      TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* userp,
+      TRITONSERVER_InferenceTraceTensorActivityFn_t tensor_activity_fn =
```
deadeyegoodwin, line 50: Move tensor_activity_fn argument to after activity_fn. You will not be able to use a default augument value (nullptr) so you will need to change the callers to explicitly pass nullptr.

src/servers/tracer.cc
```diff
@@ -26,9 +26,13 @@
 
 #include "src/servers/tracer.h"
 
+#include <stdlib.h>
 #include <unordered_map>
 #include "src/core/constants.h"
 #include "src/core/logging.h"
+#ifdef TRITON_ENABLE_GPU
+#include <cuda_runtime_api.h>
```
deadeyegoodwin, line 34: Why this include?

src/servers/tracer.cc
```diff
@@ -242,4 +246,215 @@ TraceManager::TraceActivity(
...
+    const int64_t* shape, uint64_t dim_count,
+    TRITONSERVER_MemoryType memory_type, int64_t memory_type_id, void* userp)
+{
+  if ((activity != TRITONSERVER_TRACE_TENSOR_INPUT) &&
+      (activity != TRITONSERVER_TRACE_TENSOR_OUTPUT)) {
+    return;
+  }
+
+  void* buffer_base = nullptr;
+  if (memory_type == TRITONSERVER_MEMORY_GPU) {
```
deadeyegoodwin, line 263: What is the design plan for how tensor data is handled? Above we discussed using  CopyBuffer but why are you copying within core Triton at all. If you are synchronously calling this function with 'base', 'memory_type' and 'memory_type_id' then you should just do all the copying here to get the tensor data into CPU memory so that you can output it.

deadeyegoodwin (changes requested):

ZhuYuJin:
@deadeyegoodwin The above comments should get solved in the latest commits. Please check if there are any other problems.

src/core/infer_request.cc
```diff
@@ -180,7 +248,6 @@ InferenceRequest::Release(
   // and the callback may interact with upper level trace.
   if (request->trace_ != nullptr) {
     request->trace_->ReportNow(TRITONSERVER_TRACE_REQUEST_END);
-    InferenceTrace::Release(std::move(request->trace_));
```
deadeyegoodwin, line 183: You should "request->trace_ = nullptr" to drop the shared_ptr reference. That would be the equivalent of the current code.

src/core/infer_request.cc
```diff
@@ -110,6 +113,71 @@ InferenceRequest::SetPriority(uint32_t p)
   }
 }
 
+#ifdef TRITON_ENABLE_TRACING
+Status
+InferenceRequest::TraceTensor(const std::string& msg)
```
deadeyegoodwin, line 118: It is ok to have "msg" for logging, but you also need an argument of type TRITONSERVER_InferenceTraceActivity. Those values uniquely identify what the trace record represents. Right now we have TRITONSERVER_TRACE_TENSOR_INPUT and TRITONSERVER_TRACE_TENSOR_OUTPUT. But it seems like instead we should have:
TRITONSERVER_TRACE_TENSOR_QUEUE_INPUT
TRITONSERVER_TRACE_TENSOR_BACKEND_INPUT
TRITONSERVER_TRACE_TENSOR_BACKEND_OUTPUT

src/core/tritonserver.cc
```diff
@@ -2400,17 +2421,6 @@ TRITONSERVER_ServerInferAsync(
   // Run inference...
   ni::Status status = lserver->InferAsync(ureq);
 
-  // If there is an error then must explicitly release any trace
-  // object associated with the inference request above.
-#ifdef TRITON_ENABLE_TRACING
-  if (!status.IsOk()) {
-    std::unique_ptr<ni::InferenceTrace>* trace = ureq->MutableTrace();
-    if (*trace != nullptr) {
-      ni::InferenceTrace::Release(std::move(*trace));
```
deadeyegoodwin, line 2409: You still need to "*trace = nullptr" here to release the reference to the trace object which will cause the trace release_fn_ to be called. Otherwise we are changing the existing behavior

src/servers/tracer.cc
```diff
@@ -181,9 +182,6 @@ TraceManager::TraceRelease(TRITONSERVER_InferenceTrace* trace, void* userp)
   if (parent_id == 0) {
     delete ts;
   }
-
-  LOG_TRITONSERVER_ERROR(
```
deadeyegoodwin, line 185: Why are you removing this?

src/servers/tracer.cc
```diff
@@ -242,4 +240,203 @@ TraceManager::TraceActivity(
...
+  uint64_t id;
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceTraceId(trace, &id), "getting trace id");
+
+  // The function may be called with different traces but the same 'userp',
+  // group the activity of the same trace together for more readable output.
+  auto ts = reinterpret_cast<TraceStreams*>(userp);
+  std::stringstream* ss = nullptr;
+  {
+    if (activity == TRITONSERVER_TRACE_REQUEST_START) {
```
deadeyegoodwin, line 276: You are never going to get this activity in this function. Instead you should just see if ts->streams_[id] exists and if it doesn't you need to create it. Also, because we can now get this function called by concurrently by request/response you need to have the ts->mtx_ lock stay active until update to 'ss' in complete.

You need to make similar change to TraceActivity function above.

src/servers/tracer.cc
```diff
@@ -242,4 +240,203 @@ TraceManager::TraceActivity(
...
+    } else {
+      std::lock_guard<std::mutex> lk(ts->mtx_);
+      ss = ts->streams_[id].get();
+    }
+  }
+
+  std::stringstream ss_tmp;
+
+  // collect and serialize trace details.
+  ss_tmp << ",{\"id\":" << id << ",\"activity\":\""
```
deadeyegoodwin, line 290: Why not write directly to 'ss'... doing it this way introduces unnecessary copies.

deadeyegoodwin (changes requested):

src/servers/tracer.cc
ZhuYuJin: TraceRelease is called in destructor of InferenceTrace by now. So it's no need to release it again. Otherwise, it will cause deadlock.

ZhuYuJin:
@deadeyegoodwin The above comments should get resolved. Please check this PR again. Thanks for your patience.

deadeyegoodwin (changes requested):

src/core/infer_request.cc
```diff
@@ -110,6 +113,73 @@ InferenceRequest::SetPriority(uint32_t p)
...
+          b, &buffer, &buffer_size, &src_memory_type, &src_memory_type_id);
+      if (!status.IsOk()) {
+        LOG_STATUS_ERROR(
+            status,
+            std::string(TRITONSERVER_InferenceTraceActivityString(activity)) +
+                ": " + msg + ": fail to get data buffer: " + status.Message());
+        return status;
+      }
+
+      status = CopyBuffer(
```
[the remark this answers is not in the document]
ZhuYuJin: Each input tensor consists of several chunk buffers. So, I need to copy all chunks into a contiguous buffer before calling trace_tensor_activity_fn_.

ZhuYuJin:
@deadeyegoodwin I have explained both the comments. Do you agree my opinions?

src/core/infer_request.cc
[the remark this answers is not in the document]
deadeyegoodwin: Yes, good point, that is correct. But the most common case is that there will be a single chunk, and in that case we should not have a CopyBuffer call.

deadeyegoodwin:
> @deadeyegoodwin I have explained both the comments. Do you agree my opinions?

I only saw one response from you, and I agree. See my comment above.

ZhuYuJin:
image
@deadeyegoodwin Please check this response.

deadeyegoodwin:
"the release timing of InferenceTrace is determined, which should be the release timing of InferenceRequest". That is not true.
Previous to this change, we used a unique_ptr for InferenceTrace because its lifetime was closely tied to InferenceRequest.
With this change the lifetime of InferenceTrace is now complicated. InferrenceTrace is required by both InferenceRequest and (one or more) InferenceResponse. The lifetimes of InferenceRequest and InferenceResponse are completely decoupled... an InferenceResponse can live much longer than the corresponding InferenceRequest, and vice-versa. So we now use a shared_ptr for InferenceTrace and we can no longer assume that its lifetime is tied to InferenceRequest. An InferenceTrace reference can be held by an InferenceRequest object, and one or more InferenceResponse objects, and by the client using the C API. Once once all these references are released should the InferenceTrace object be released and finalized. Careful reading of the entire InferenceTrace lifetime/flow is required to make sure that this new lifetime is correctly handled.... and whatever we do must remain backwards compatible at the API level.

deadeyegoodwin:
BTW, the reason I couldn't see you response is because I don't think you actually posted it... note the "Pending" next to it.

ZhuYuJin:
> "the release timing of InferenceTrace is determined, which should be the release timing of InferenceRequest". That is not true. Previous to this change, we used a unique_ptr for InferenceTrace because its lifetime was closely tied to InferenceRequest. With this change the lifetime of InferenceTrace is now complicated. InferrenceTrace is required by both InferenceRequest and (one or more) InferenceResponse. The lifetimes of InferenceRequest and InferenceResponse are completely decoupled... an InferenceResponse can live much longer than the corresponding InferenceRequest, and vice-versa. So we now use a shared_ptr for InferenceTrace and we can no longer assume that its lifetime is tied to InferenceRequest. An InferenceTrace reference can be held by an InferenceRequest object, and one or more InferenceResponse objects, and by the client using the C API. Once once all these references are released should the InferenceTrace object be released and finalized. Careful reading of the entire InferenceTrace lifetime/flow is required to make sure that this new lifetime is correctly handled.... and whatever we do must remain backwards compatible at the API level.

src/servers/tracer.cc
ZhuYuJin: I split original TraceRelease into two functions: TraceRelease and TraceStreamRelease.
TraceRelease should delete InferenceTrace object.
TraceStreamRelease should be called in InferenceTrace destructor.

ZhuYuJin:
@deadeyegoodwin I have made two modification:
1. I simply split original TraceRelease into two functions: TraceRelease and TraceStreamRelease.
   TraceRelease should delete InferenceTrace object.
   TraceStreamRelease should be called in InferenceTrace destructor.
2. If there is less than one chunk, TraceTensor in InferenceRequest will not copy buffer.

deadeyegoodwin:
> @deadeyegoodwin I have made two modification:
> 
>     1. I simply split original TraceRelease into two functions: TraceRelease and TraceStreamRelease.
>        TraceRelease should delete InferenceTrace object.
>        TraceStreamRelease should be called in InferenceTrace destructor.
> 
>     2. If there is less than one chunk, TraceTensor in InferenceRequest will not copy buffer.

The #2 change looks good. I'm not sure about #1 change. The first thing you need to do is to address the comments I made above:
"But now we need TRITONSERVER_InferenceTrace needs to represent shared_ptr so that TRITONSERVER_InferenceTrace* is a cast of shared_ptr*"
I don't see that you have done that change. Have you carefully audited the full life-cycle of trace objects along with making that change?

ZhuYuJin:
> > @deadeyegoodwin I have made two modification:
> > ```
> > 1. I simply split original TraceRelease into two functions: TraceRelease and TraceStreamRelease.
> >    TraceRelease should delete InferenceTrace object.
> >    TraceStreamRelease should be called in InferenceTrace destructor.
> > 
> > 2. If there is less than one chunk, TraceTensor in InferenceRequest will not copy buffer.
> > ```
> 
> The #2 change looks good. I'm not sure about #1 change. The first thing you need to do is to address the comments I made above: "But now we need TRITONSERVER_InferenceTrace needs to represent shared_ptr so that TRITONSERVER_InferenceTrace* is a cast of shared_ptr*" I don't see that you have done that change. Have you carefully audited the full life-cycle of trace objects along with making that change?

Hi. I think I don't get your point that we should represent TRITONSERVER_InferenceTrace* as shared_ptr*. What's the difference that the user keep a shared_ptr* or InferenceTrace*? 

In the current change, I simply split original TraceRelease into two functions: TraceRelease and TraceStreamRelease. TraceRelease will delete InferenceTrace object. InferenceTrace destructor will call TraceStreamRelease. And TraceStreamRelease will release TraceStream.

Since InferenceTrace destructor will call TraceStreamRelease to release TraceStream, we only need to release InferenceTrace.

The grpc/http server will call trace_manager->SampleTrace or TRITONSERVER_InferenceTraceNew to get a InferenceTrace*, and call trace_manager->TraceRelease or TRITONSERVER_InferenceTraceDelete to delete InferenceTrace*. 

In Triton, we will call trace->SpawnChildTrace to get a child shared_ptr<InferenceTrace>, which is maintained by InferenceRequest and InferenceResponse. After InferenceRequest and InferenceResponse are released, the InferenceTrace will get auto released.

ZhuYuJin:
> > > @deadeyegoodwin I have made two modification:
> > > ```
> > > 1. I simply split original TraceRelease into two functions: TraceRelease and TraceStreamRelease.
> > >    TraceRelease should delete InferenceTrace object.
> > >    TraceStreamRelease should be called in InferenceTrace destructor.
> > > 
> > > 2. If there is less than one chunk, TraceTensor in InferenceRequest will not copy buffer.
> > > ```
> > 
> > 
> > The #2 change looks good. I'm not sure about #1 change. The first thing you need to do is to address the comments I made above: "But now we need TRITONSERVER_InferenceTrace needs to represent shared_ptr so that TRITONSERVER_InferenceTrace* is a cast of shared_ptr*" I don't see that you have done that change. Have you carefully audited the full life-cycle of trace objects along with making that change?
> 
> @deadeyegoodwin Hi. I think I don't get your point that we should represent TRITONSERVER_InferenceTrace* as shared_ptr*. What's the difference between shared_ptr* and InferenceTrace*?
> 
> In the current change, I simply split original TraceRelease into two functions: **_TraceRelease_** and **_TraceStreamRelease_**. **_TraceRelease_** will delete InferenceTrace object. InferenceTrace destructor will call **_TraceStreamRelease_**. And **_TraceStreamRelease_** will release TraceStream.
> 
> Since InferenceTrace destructor will call **_TraceStreamRelease_** to release TraceStream, we only need to release InferenceTrace.
> 
> The grpc/http server will call trace_manager->SampleTrace or TRITONSERVER_InferenceTraceNew to get a InferenceTrace*, and call **_TraceRelease_** or TRITONSERVER_InferenceTraceDelete to delete InferenceTrace*.
> 
> Inside Triton, we will call trace->SpawnChildTrace to get a child shared_ptr, which is maintained by InferenceRequest and InferenceResponse. After InferenceRequest and InferenceResponse are released, the InferenceTrace will get auto released.

Hi. Please check this comment if you are free. I'm confused of our last divergence point.

deadeyegoodwin:
> > > > @deadeyegoodwin I have made two modification:
> > > > ```
> > > > 1. I simply split original TraceRelease into two functions: TraceRelease and TraceStreamRelease.
> > > >    TraceRelease should delete InferenceTrace object.
> > > >    TraceStreamRelease should be called in InferenceTrace destructor.
> > > > 
> > > > 2. If there is less than one chunk, TraceTensor in InferenceRequest will not copy buffer.
> > > > ```
> > > 
> > > 
> > > The #2 change looks good. I'm not sure about #1 change. The first thing you need to do is to address the comments I made above: "But now we need TRITONSERVER_InferenceTrace needs to represent shared_ptr so that TRITONSERVER_InferenceTrace* is a cast of shared_ptr*" I don't see that you have done that change. Have you carefully audited the full life-cycle of trace objects along with making that change?
> > 
> > 
> > @deadeyegoodwin Hi. I think I don't get your point that we should represent TRITONSERVER_InferenceTrace* as shared_ptr*. What's the difference between shared_ptr* and InferenceTrace*?
> > In the current change, I simply split original TraceRelease into two functions: **_TraceRelease_** and **_TraceStreamRelease_**. **_TraceRelease_** will delete InferenceTrace object. InferenceTrace destructor will call **_TraceStreamRelease_**. And **_TraceStreamRelease_** will release TraceStream.
> > Since InferenceTrace destructor will call **_TraceStreamRelease_** to release TraceStream, we only need to release InferenceTrace.
> > The grpc/http server will call trace_manager->SampleTrace or TRITONSERVER_InferenceTraceNew to get a InferenceTrace*, and call **_TraceRelease_** or TRITONSERVER_InferenceTraceDelete to delete InferenceTrace*.
> > Inside Triton, we will call trace->SpawnChildTrace to get a child shared_ptr, which is maintained by InferenceRequest and InferenceResponse. After InferenceRequest and InferenceResponse are released, the InferenceTrace will get auto released.
> 
> @deadeyegoodwin Hi. Please check this comment if you are free. I'm confused of our last divergence point.

I'll need to look through the entire trace code to understand what you are proposing, as it has been a long time since I worked on that part of the code. I likely won't be able to provide feedback until later this week.

deadeyegoodwin:
After reading through the entire trace lifecycle I think the changes you need to make are fairly limited. The primary issue is that the Trace object's lifetime is no longer linked to the Request object's lifetime and so we can't release the Trace object when we release the Request object. I would implement InferenceTrace as below. InferenceTraceProxy is the object that needs to be attached as a shared_ptr to the Request and all created Response objects. When all of those objects are released then the shared_ptr will destruct the InferenceTraceProxy object and that in turn will release the InferenceTrace object. In infer_request.cc we should not explicitly release the trace object.

```
diff --git a/src/core/infer_trace.h b/src/core/infer_trace.h
index 28787c5..e3c4188 100644
--- a/src/core/infer_trace.h
+++ b/src/core/infer_trace.h
@@ -81,10 +81,8 @@ class InferenceTrace {
                       .count());
   }

-  // Release the trace. Call the trace release callback and transfer
-  // ownership of the trace to the callback. On return 'trace' is
-  // nullptr.
-  static void Release(std::unique_ptr<InferenceTrace>&& trace);
+  // Release the trace and pass ownership back to release_fn_.
+  void Release();

  private:
   const TRITONSERVER_InferenceTraceLevel level_;
@@ -103,6 +101,45 @@ class InferenceTrace {
   static std::atomic<uint64_t> next_id_;
 };

+//
+// InferenceTraceProxy
+//
+// Object attached as shared_ptr to InferenceRequest and
+// InferenceResponse(s) being traced as part of a single inference
+// request.
+//
+class InferenceTraceProxy {
+ public:
+  InferenceTraceProxy(InferenceTrace* trace) : trace_(trace) { }
+  ~InferenceTraceProxy()
+  {
+    trace_->Release();
+  }
+
+  int64_t Id() const { return trace_->Id(); }
+  int64_t ParentId() const { return trace_->ParentId(); }
+
+  const std::string& ModelName() const { return trace_->ModelName(); }
+  int64_t ModelVersion() const { return trace_->ModelVersion(); }
+
+  void SetModelName(const std::string& n) { trace_->SetModelName(n); }
+  void SetModelVersion(int64_t v) { trace_->SetModelVersion(v); }
+
+  void Report(
+      const TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns)
+  {
+    trace_->Report(activity, timestamp_ns);
+  }
+
+  void ReportNow(const TRITONSERVER_InferenceTraceActivity activity)
+  {
+    trace_->ReportNow(activity);
+  }
+
+ private:
+  InferenceTrace* trace_;
+};
+

 //

```

ZhuYuJin:
> After reading through the entire trace lifecycle I think the changes you need to make are fairly limited. The primary issue is that the Trace object's lifetime is no longer linked to the Request object's lifetime and so we can't release the Trace object when we release the Request object. I would implement InferenceTrace as below. InferenceTraceProxy is the object that needs to be attached as a shared_ptr to the Request and all created Response objects. When all of those objects are released then the shared_ptr will destruct the InferenceTraceProxy object and that in turn will release the InferenceTrace object. In infer_request.cc we should not explicitly release the trace object.
> 
> ```
> diff --git a/src/core/infer_trace.h b/src/core/infer_trace.h
> index 28787c5..e3c4188 100644
> --- a/src/core/infer_trace.h
> +++ b/src/core/infer_trace.h
> @@ -81,10 +81,8 @@ class InferenceTrace {
>                        .count());
>    }
> 
> -  // Release the trace. Call the trace release callback and transfer
> -  // ownership of the trace to the callback. On return 'trace' is
> -  // nullptr.
> -  static void Release(std::unique_ptr<InferenceTrace>&& trace);
> +  // Release the trace and pass ownership back to release_fn_.
> +  void Release();
> 
>   private:
>    const TRITONSERVER_InferenceTraceLevel level_;
> @@ -103,6 +101,45 @@ class InferenceTrace {
>    static std::atomic<uint64_t> next_id_;
>  };
> 
> +//
> +// InferenceTraceProxy
> +//
> +// Object attached as shared_ptr to InferenceRequest and
> +// InferenceResponse(s) being traced as part of a single inference
> +// request.
> +//
> +class InferenceTraceProxy {
> + public:
> +  InferenceTraceProxy(InferenceTrace* trace) : trace_(trace) { }
> +  ~InferenceTraceProxy()
> +  {
> +    trace_->Release();
> +  }
> +
> +  int64_t Id() const { return trace_->Id(); }
> +  int64_t ParentId() const { return trace_->ParentId(); }
> +
> +  const std::string& ModelName() const { return trace_->ModelName(); }
> +  int64_t ModelVersion() const { return trace_->ModelVersion(); }
> +
> +  void SetModelName(const std::string& n) { trace_->SetModelName(n); }
> +  void SetModelVersion(int64_t v) { trace_->SetModelVersion(v); }
> +
> +  void Report(
> +      const TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns)
> +  {
> +    trace_->Report(activity, timestamp_ns);
> +  }
> +
> +  void ReportNow(const TRITONSERVER_InferenceTraceActivity activity)
> +  {
> +    trace_->ReportNow(activity);
> +  }
> +
> + private:
> +  InferenceTrace* trace_;
> +};
> +
>  #endif  // TRITON_ENABLE_TRACING
> 
>  //
> ```

@deadeyegoodwin Please check the latest commit. I have made the changes you mentioned.

ZhuYuJin:
@deadeyegoodwin Hi, please check the latest commit if you are free.

src/core/infer_trace.h
```diff
@@ -81,17 +83,31 @@ class InferenceTrace {
...
+      TRITONSERVER_MemoryType memory_type, int64_t memory_type_id)
+  {
+    tensor_activity_fn_(
+        reinterpret_cast<TRITONSERVER_InferenceTrace*>(this), activity, name,
+        datatype, base, byte_size, shape, dim_count, memory_type,
+        memory_type_id, userp_);
+  }
+
   // Release the trace. Call the trace release callback and transfer
   // ownership of the trace to the callback. On return 'trace' is
```
deadeyegoodwin, line 100: Remove last sentence as it is no longer relevant.

src/core/infer_trace.cc
```diff
@@ -34,21 +34,27 @@ namespace nvidia { namespace inferenceserver {
...
-  release_fn(
-      reinterpret_cast<TRITONSERVER_InferenceTrace*>(trace.release()), userp);
+  release_fn_(reinterpret_cast<TRITONSERVER_InferenceTrace*>(this), userp_);
+}
+
+std::shared_ptr<InferenceTraceProxy>
+InferenceTraceProxy::SpawnChildTrace()
+{
+  InferenceTrace* child_trace = trace_->SpawnChildTrace();
+  InferenceTraceProxy* trace_proxy = new InferenceTraceProxy(child_trace);
```
deadeyegoodwin, line 55: These 2 lines can be replaced with std::make_shared

src/core/tritonserver.cc
```diff
@@ -2395,8 +2418,9 @@ TRITONSERVER_ServerInferAsync(
     ltrace->SetModelName(lrequest->ModelName());
     ltrace->SetModelVersion(lrequest->ActualModelVersion());
 
-    std::unique_ptr<ni::InferenceTrace> utrace(ltrace);
-    lrequest->SetTrace(std::move(utrace));
+    ni::InferenceTraceProxy* trace_proxy = new ni::InferenceTraceProxy(ltrace);
+    std::shared_ptr<ni::InferenceTraceProxy> strace_proxy(trace_proxy);
+    lrequest->SetTrace(strace_proxy);
```
deadeyegoodwin, line 2423: I believe you can replace these 3 lines with:
lrequest->SetTrace(std::make_shared<ni::InferenceTraceProxy>(ltrace))

src/core/infer_request.cc
```diff
@@ -110,6 +113,111 @@ InferenceRequest::SetPriority(uint32_t p)
   }
 }
 
+#ifdef TRITON_ENABLE_TRACING
+Status
+InferenceRequest::TraceTensor(
```
deadeyegoodwin, line 118: TraceInputTensors

src/core/infer_request.cc
```diff
@@ -199,7 +307,7 @@ InferenceRequest::Release(
   // and the callback may interact with upper level trace.
```
deadeyegoodwin, line 307: Note that this layering will be broken now because setting trace_ = nullptr doesn't necessarily cause the trace release callback to be called. Not sure what the impact of this will be.

src/core/infer_request.cc
```diff
@@ -199,7 +307,7 @@ InferenceRequest::Release(
...
   if (request->trace_ != nullptr) {
     request->trace_->ReportNow(TRITONSERVER_TRACE_REQUEST_END);
-    InferenceTrace::Release(std::move(request->trace_));
+    request->trace_ = nullptr;
   }
```
deadeyegoodwin, line 311: Also need to call
response_factory_.SetTrace(nullptr) or the equivalent to drop the trace reference that is held by the response factory.

src/core/tritonserver.cc
```diff
@@ -2414,9 +2438,9 @@ TRITONSERVER_ServerInferAsync(
   // object associated with the inference request above.
 #ifdef TRITON_ENABLE_TRACING
   if (!status.IsOk()) {
-    std::unique_ptr<ni::InferenceTrace>* trace = ureq->MutableTrace();
+    std::shared_ptr<ni::InferenceTraceProxy>* trace = ureq->MutableTrace();
     if (*trace != nullptr) {
-      ni::InferenceTrace::Release(std::move(*trace));
+      *trace = nullptr;
```
deadeyegoodwin, line 2443: Need to add a InferenceRequest::ReleaseTrace() function and have that release both the InferenceRequest and the InferenceResponseFactory trace object.

deadeyegoodwin (changes requested):

src/core/infer_request.cc
ZhuYuJin: I think this note is not necessary.
The layering will be the same as before. We only change the lifetime of trace, which controls the timing to write TraceStream into trace file. And we do not change any record timing. So, each record will stay the same sequence as before in the TraceStream.

src/core/infer_request.cc
ZhuYuJin: change to call request->ReleaseTrace

src/core/tritonserver.cc
ZhuYuJin: change to call ureq->ReleaseTrace()

ZhuYuJin:
@deadeyegoodwin Please check the latest commit again.

src/core/infer_request.cc
deadeyegoodwin: We will see once we start running some existing and new tests.

deadeyegoodwin (commented):
These changes look good. Have you built and run the existing trace tests to ensure there are no regressions? We will need to next update the tritonserver command-line arguments to allow this new tracing to be enabled in a way that is backwards compatible.

ZhuYuJin:
>

ZhuYuJin:
> These changes look good. Have you built and run the existing trace tests to ensure there are no regressions? We will need to next update the tritonserver command-line arguments to allow this new tracing to be enabled in a way that is backwards compatible.

image
I'm trying to built and run the existing trace tests.
Besides, we have three trace levels by now, which are "disabled", "min" and "max". Should we add a trace level called "tensor" between "min" and "max"?

deadeyegoodwin:
I put some related comments here: https://github.com/triton-inference-server/core/pull/36

ParseTraceLevelOption and cmdline documentation need to be updated to match. We want min and max deprecated and replaced with "timestamps".

ZhuYuJin:
@deadeyegoodwin Hi, I have updated cmdline and InferenceTrace. Please check the latest commit again.

BTW, I have run the existing trace tests to ensure there are no regressions.
image

src/core/infer_trace.h
```diff
@@ -69,6 +69,10 @@ class InferenceTrace {
   void Report(
       const TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns)
   {
+    if (level_ < TRITONSERVER_TRACE_LEVEL_TIMESTAMPS) {
```
deadeyegoodwin, line 72: Do not use '<'. level_ is a bitmask of enabled trace levels. So you need to mask TRACE_LEVEL_TIMESTAMPS and only report if that bit is set.

src/core/infer_trace.h
```diff
@@ -90,6 +98,10 @@ class InferenceTrace {
       const int64_t* shape, uint64_t dim_count,
       TRITONSERVER_MemoryType memory_type, int64_t memory_type_id)
   {
+    if (level_ < TRITONSERVER_TRACE_LEVEL_TENSORS) {
```
deadeyegoodwin, line 101: Do not use '<'. level_ is a bitmask of enabled trace levels. So you need to mask TRACE_LEVEL_TENSORS and only report if that bit is set.

src/servers/main.cc
```diff
@@ -462,8 +462,9 @@ std::vector<Option> options_
       {OPTION_TRACE_FILEPATH, "trace-file", Option::ArgStr,
        "Set the file where trace output will be saved."},
       {OPTION_TRACE_LEVEL, "trace-level", Option::ArgStr,
-       "Set the trace level. OFF to disable tracing, MIN for minimal tracing, "
-       "MAX for maximal tracing. Default is OFF."},
+       "Set the trace level. OFF to disable tracing, TIMESTAMPES to trace "
```
deadeyegoodwin, line 465: We need to allow --trace-level to be specified more than once because there are multiple trace levels that can be enabled.  For example, see how --backend-config is handled. TIMESTAMPS enabled only timestamps and TENSORS enables only tensors.

src/core/infer_trace.h
```diff
@@ -69,6 +69,10 @@ class InferenceTrace {
   void Report(
```
deadeyegoodwin, line 69: In the contructor you need to set level_ in a way that handles the deprecated MIN and MAX settings.

deadeyegoodwin (changes requested):
In TraceManager as well you need to fix the initialization and handling of TraceLevel to account for deprecated MIN and MAX and the new TIMESTAMPS and TENSORS values.

ZhuYuJin:
@deadeyegoodwin 
By now, I simply transfer both the deprecated MIN and MAX to the new TIMESTAMPS. So, I do not need to handle them.

Besides, If we take trace_level_ as a bitmask of enabled trace levels, we have to change the definition of the old interface `TRITONSERVER_InferenceTraceNew` in `tritonserver.h` because we take level as an enum in the current interface.
image

Maybe it should look like:
TRITONSERVER_DECLSPEC TRITONSERVER_Error* TRITONSERVER_InferenceTraceNew(
    TRITONSERVER_InferenceTrace** trace, TRITONSERVER_InferenceTraceLevel* level_settings,
    uint64_t parent_id, TRITONSERVER_InferenceTraceActivityFn_t activity_fn,
    TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* trace_userp);

But, we can keep the old interface `TRITONSERVER_InferenceTraceNew` as before, and only fit the new level bitmap into the new interface `TRITONSERVER_InferenceTraceTensorNew`. Then, `TRITONSERVER_InferenceTraceNew` will only support one trace level. And `TRITONSERVER_InferenceTraceTensorNew` support multiple trace level.

However, if we take trace_level_ as an enum and use comparison operators to check the trace level, we should not have the above problem.

What's your suggestion?

deadeyegoodwin:
The trace level remains an enum. It is just that we only allow power-of-2 enum values (except for the old deprecated values). A "trace level" is still a "trace level" and it is still an enum... the documentation is updated already to indicate what the trace level enum means. On the cmdline and C API we must still accept min and max for backwards compatibility, but internally we should convert those to TIMESTAMPS, so that internally in the code we can assume that trace level will never be min or max and so we can always treat it as a bitmask. There is no need to change the InfrenceTraceNew APIs.

ZhuYuJin:
@deadeyegoodwin Please check the latest commit.

I convert MIN and MAX to TIMESTAMPS on the cmdline, C APIs and grpc/http server.
Besides, I convert trace level bitmap to the enum type, and continue the rest logits.

src/core/tritonserver.cc
```diff
@@ -807,6 +807,10 @@ TRITONSERVER_InferenceTraceNew(
     TRITONSERVER_InferenceTraceReleaseFn_t release_fn, void* trace_userp)
 {
 #ifdef TRITON_ENABLE_TRACING
+  if ((level == TRITONSERVER_TRACE_LEVEL_MIN) ||
```
deadeyegoodwin, line 810: This is the right idea, but level could be (TRITONSERVER_TRACE_LEVEL_TIMESTAMPS | TRITONSERVER_TRACE_LEVEL_TENSORS). So if either min or max is set you need to turn on TIMESTAMPS (level |= TRITONSERVER_TRACE_LEVEL_TIMESTAMPS), and unconditionally you need to mask out MIN and MAX.

src/servers/main.cc
```diff
@@ -462,9 +462,8 @@ std::vector<Option> options_
       {OPTION_TRACE_FILEPATH, "trace-file", Option::ArgStr,
        "Set the file where trace output will be saved."},
       {OPTION_TRACE_LEVEL, "trace-level", Option::ArgStr,
-       "Set the trace level. OFF to disable tracing, TIMESTAMPES to trace "
-       "timestamps, TENSORS to trace both timestamps and tensors. Default is "
-       "OFF."},
+       "Specify a trace level. OFF to disable tracing, TIMESTAMPES to "
```
deadeyegoodwin, line 465: Fix spelling of TIMESTAMPS. Also you need to say something about how the flag can be specified multiple times. That is needed so that you can turn on both TIMESTAMPS and TENSORS.

deadeyegoodwin (changes requested):

ZhuYuJin:
@deadeyegoodwin Please check the latest commit. I have masked out MIN and MAX, and added some description of trace-level command.

src/servers/main.cc
deadeyegoodwin: Still need to fix TIMESTAMPES -> TIMESTAMPS

deadeyegoodwin (commented):
Just need to fix the one spelling issue, rebuild and make sure existing tests still work and then I will try your changes in a full CI run. The last step will be to add new testing for the TENSORS trace.

qa/L0_cmdline_trace/test.sh
```diff
@@ -104,8 +104,8 @@ fi
 
 set -e
 
-# trace-rate == 1, trace-level=MIN make sure every request is traced
```
deadeyegoodwin, line 107: Keep one of the MIN and one of the MAX test cases to cover backwards compatibility

qa/L0_cmdline_trace/test.sh
```diff
@@ -333,4 +333,79 @@ else
...
+
+if [ `grep -o TENSOR_QUEUE_INPUT trace_ensemble_tensor.log | wc -l` != "18" ]; then
+    echo -e "Ensemble trace tensors log expects 18 TENSOR_QUEUE_INPUTs"
+    RET=1
+fi
+
+if [ `grep -o TENSOR_BACKEND_OUTPUT trace_ensemble_tensor.log | wc -l` != "14" ]; then
+    echo -e "Ensemble trace tensors log expects 14 TENSOR_BACKEND_OUTPUTs"
+    RET=1
+fi
```
deadeyegoodwin, line 399: Need to test that the traced tensor values are what is expected.

deadeyegoodwin (changes requested):

deadeyegoodwin:
Just need a couple of test updated. Then you need to rebase everything to main branch and resolve any conflicts, and then I will run CI.

ZhuYuJin:
@deadeyegoodwin I have merged main branch in the server/core repos. Please check the latest commit.

deadeyegoodwin:
I ran the CI and linux looks good. There is a strange windows build failure that I will need to triage before we can merge.

deadeyegoodwin:
I fixed the windows build issue. It wasn't directly related to your change but was somehow triggered by it.
You need to rebase again and then I can run another CI.

deadeyegoodwin:
The CI passed so this is ready to merge. I can merge my local rebased copy of these changes but I would prefer to just merge your PR. We have other, conflicting changes pending so we need to get this merged ASAP. Are you able to rebase it soon?

ZhuYuJin:
@deadeyegoodwin I have merged the latest main branch and also fixed a print problem in `trace_summary.py`.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -37,36 +37,40 @@
...
-namespace nvidia { namespace inferenceserver {
-
-bool
-IsStaleState(Payload::State payload_state)
-{
-  return (
-      (payload_state == Payload::State::EXECUTING) ||
-      (payload_state == Payload::State::RELEASED));
+namespace nvidia {
+namespace inferenceserver {
```
deadeyegoodwin, line 41: Why did you change this formatting and other formatting within the file? This need to be reverted.

src/core/dynamic_batch_scheduler.cc
```diff
@@ -152,18 +158,28 @@ DynamicBatchScheduler::~DynamicBatchScheduler()
...
+  // oldest sequence batcher, this will overwrite the value that was previously
+  // set by sequence batcher, which is okay as by this point, the previous
+  // batcher won't be needing this value and it can be safely reused by
+  // the dynamic batcher.
+  request->CaptureBatcherStartNs();
+
 #ifdef TRITON_ENABLE_TRACING
-  request->TraceInputTensors(
-      TRITONSERVER_TRACE_TENSOR_QUEUE_INPUT, "DynamicBatchScheduler Enqueue");
+  request->TraceInputTensors(TRITONSERVER_TRACE_TENSOR_QUEUE_INPUT,
```
deadeyegoodwin, line 181: Move this to just after line 170, it needs to be inside of the conditional above.

deadeyegoodwin (changes requested):

PR #3600, opened 2021-11-22
Add test coverage for pytorch too few inputs/outputs

CoderHam:
Autofill should return error if no inputs / outputs are provided in model config


GuanLuo (approved):

GuanLuo:
Does this require a model file? If so you need to update the test.sh

CoderHam:
> Does this require a model file? If so you need to update the test.sh

Done

GuanLuo (approved):

PR #3611, opened 2021-11-24
Fix L0_long_running_stress

GuanLuo:


deadeyegoodwin (dismissed):

CoderHam (dismissed):

dyastremsky (dismissed):

dyastremsky (dismissed):

CoderHam (approved):

deadeyegoodwin (approved):

qa/L0_long_running_stress/scenarios.py
```diff
@@ -263,7 +263,7 @@ def generate_sequence_data(self, trial, dtype, data_filename):
                 if dtype == np.float32:
                     res = float(res)
                 elif dtype == np.int32:
-                    # Do nothing
+                    res = int(res)
```
tanmayv25, line 266: Why do we need `int` cast here?

dyastremsky: Good find. Heads up that https://github.com/triton-inference-server/server/pull/3674 was just merged and addressed this with pass (an allowed no-op). Might want to rebase this PR off main.

dyastremsky: Ignore my comment above, looks like you did rebase. Pass might be clearer/more efficient, since it's a no-op, but really up to preference.

GuanLuo:
Seeing START flag issue in perf analyzer is probably due to not running it with streaming flag and thus the request sequence can be out of order. Verifying the fix by running the script with default duration (7h)..

dyastremsky (approved):

tanmayv25 (approved):

PR #3635, opened 2021-11-30
Add tests for explicit model control in Python backend

Tabrizian:


CoderHam (changes requested):
Can you change the test name from explicit to model_control / explicit_model_control?

tanmayv25:
@Tabrizian Should we add these tests in `lifecycle`? We have explicit model load/reload for other platforms in L0_lifecycle.

qa/L0_backend_python/explicit/explicit_model_test.py
```diff
@@ -0,0 +1,65 @@
...
+            model_name=model_name,
+            inputs=inputs,
+            outputs=[httpclient.InferRequestedOutput('OUTPUT0')])
+        output_numpy = result.as_numpy('OUTPUT0')
+        self.assertTrue(np.all(input0_data == output_numpy))
+
+    def test_model_reload(self):
+        model_name = "identity_fp32"
+        with httpclient.InferenceServerClient("localhost:8000") as client:
+            for _ in range(5):
```
tanmayv25, line 56: You can check the `model_name` is not already loaded in the server before proceeding.

Tabrizian:
@tanmayv25 I looked into L0_lifecycle. It looks like a good place to add Python backend but the tests need some changes to make them compatible with other backends so it can take some time. I created a ticket for it.

tanmayv25 (approved):

CoderHam (approved):

PR #3636, opened 2021-11-30
Use compose.py --gpu-enable=false for cpu-only build and test

jbkyang-nvi:
Remove support of cpu-only containers for compose.py

### DIFF
--- a/compose.py
+++ b/compose.py
@@ -73,7 +73,7 @@ FROM {}
 
     import build
     df += build.dockerfile_prepare_container_linux(argmap, backends,
-                                                   FLAGS.enable_gpu,
+                                                   not FLAGS.disable_gpu,
                                                    platform.machine().lower())
     # Copy over files
     df += '''
@@ -154,7 +154,7 @@ def create_argmap(images):
     # use
     full_docker_image = images["full"]
     min_docker_image = images["min"]
-    enable_gpu = FLAGS.enable_gpu
+    enable_gpu = not FLAGS.disable_gpu
     # Docker inspect enviroment variables
     base_run_args = ['docker', 'inspect', '-f']
     import re  # parse all PATH enviroment variables
@@ -187,7 +187,7 @@ def create_argmap(images):
         gpu_enabled = False if e is None else True
         fail_if(
             not gpu_enabled,
-            '\'enable-gpu\' flag specified but min container provided does not have CUDA installed'
+            '\'disable-gpu\' flag not specified but min container provided does not have CUDA installed'
         )
 
     # Check full container enviroment variables
@@ -213,7 +213,7 @@ def create_argmap(images):
         gpu_enabled = True
     fail_if(
         gpu_enabled != enable_gpu,
-        'Error: full container provided was build with \'enable_gpu\' as {} and you are composing container with \'enable_gpu\' as {}'
+        'Error: full container provided was build with \'TRITON_SERVER_GPU_ENABLED\' as {} and you are composing container with \'TRITON_SERVER_GPU_ENABLED\' as {}'
         .format(gpu_enabled, enable_gpu))
     e = re.search("TRITON_SERVER_VERSION=([\S]{6,}) ", vars)
     version = "" if e is None else e.group(1)
@@ -314,10 +314,10 @@ if __name__ == '__main__':
         help=
         'Use specified Docker image to generate Docker image. Specified as <image-name>,<full-image-name>. <image-name> can be "min" or "full". Both "min" and "full" need to be specified at the same time. This will override "--container-version".'
     )
-    parser.add_argument('--enable-gpu',
+    parser.add_argument('--disable-gpu',
                         action="store_true",
                         required=False,
-                        help='Generate a Triton image that supports GPU.')
+                        help='Generate a cpu-only Triton image.')
     parser.add_argument(
         '--backend',
         action='append',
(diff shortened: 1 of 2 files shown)


compose.py
```diff
@@ -73,7 +73,7 @@ def start_dockerfile(ddir, images, argmap, dockerfile_name, backends):
 
     import build
     df += build.dockerfile_prepare_container_linux(argmap, backends,
-                                                   FLAGS.enable_gpu,
+                                                   not FLAGS.disable_gpu,
```
CoderHam, line 76: We might want to rename the variables and re-work the checks for disable_gpu instead of using enable_gpu

CoderHam (changes requested):

compose.py
```diff
@@ -314,10 +314,10 @@ def create_argmap(images):
         help=
         'Use specified Docker image to generate Docker image. Specified as <image-name>,<full-image-name>. <image-name> can be "min" or "full". Both "min" and "full" need to be specified at the same time. This will override "--container-version".'
     )
-    parser.add_argument('--enable-gpu',
```
deadeyegoodwin, line 317: We need to keep the --enable-gpu flag for backward compat. We want it to be a flag that can be set to true or false but default to true. As cpu-only build is not supported the flag isn't needed and we don't want to document it. So set the appropriate setting so that it does not appear in the --help output.

docs/compose.md
```diff
@@ -44,26 +44,25 @@ from source to get more exact customization.
 The `compose.py` script can be found in the [server repository](https://github.com/triton-inference-server/server).
 Simply clone the repository and run `compose.py` to create a custom container.
 Note: Created container version will depend on the branch that was cloned.
-For example branch [r21.11](https://github.com/triton-inference-server/server/tree/r21.11)
-should be used to create a image based on the NGC 21.11 Triton release.
+For example branch [r21.12](https://github.com/triton-inference-server/server/tree/r21.12)
```
deadeyegoodwin, line 47: Don't change these versions. 21.12 is not yet available.

deadeyegoodwin (changes requested):
This looks good. But we have just discussed some related changes where

compose.py
```diff
@@ -315,9 +315,10 @@ def create_argmap(images):
         'Use specified Docker image to generate Docker image. Specified as <image-name>,<full-image-name>. <image-name> can be "min" or "full". Both "min" and "full" need to be specified at the same time. This will override "--container-version".'
     )
     parser.add_argument('--enable-gpu',
-                        action="store_true",
-                        required=False,
-                        help='Generate a Triton image that supports GPU.')
+                        nargs='?',
+                        type=lambda x: (str(x).lower() == 'true'),
```
jbkyang-nvi, line 319: this is needed as type=bool always returns true

compose.py
```diff
@@ -315,9 +315,10 @@ def create_argmap(images):
...
+                        const=True,
+                        required=True)
```
deadeyegoodwin, line 321: This flag doesn't need to be required, does it?

docs/compose.md
```diff
@@ -49,8 +49,7 @@ should be used to create a image based on the NGC 21.11 Triton release.
 
 `compose.py` provides `--backend`, `--repoagent` options that allow you to
 specify which backends and repository agents to include in the custom image.
-The `--enable-gpu` flag indicates that you want to create an image that supports
-NVIDIA GPUs. For example, the following creates a new docker image that
+For example, the following creates a new docker image that
```
deadeyegoodwin, line 52: Remove --enable-gpu from the below commands. We don't want any mention of it in docs or --help

compose.py
```diff
@@ -315,9 +315,10 @@ def create_argmap(images):
         'Use specified Docker image to generate Docker image. Specified as <image-name>,<full-image-name>. <image-name> can be "min" or "full". Both "min" and "full" need to be specified at the same time. This will override "--container-version".'
     )
     parser.add_argument('--enable-gpu',
```
deadeyegoodwin, line 317: Doesn't it still show up in --help?

deadeyegoodwin (changes requested):

compose.py
jbkyang-nvi: I thought the whole point of keeping --enable-gpu is to keep backward compatibility

compose.py
CoderHam: Addition of help=argparse.SUPPRESS ensures it does not show up in --help

CoderHam (approved):

compose.py
```diff
@@ -187,7 +187,7 @@ def create_argmap(images):
         gpu_enabled = False if e is None else True
         fail_if(
             not gpu_enabled,
-            '\'enable-gpu\' flag specified but min container provided does not have CUDA installed'
+            'Composing container with gpu enabled but min container provided does not have CUDA installed'
```
CoderHam, line 190: gpu _support_ enabled

CoderHam (changes requested):

CoderHam (dismissed):

compose.py
```diff
@@ -315,9 +315,12 @@ def create_argmap(images):
         'Use specified Docker image to generate Docker image. Specified as <image-name>,<full-image-name>. <image-name> can be "min" or "full". Both "min" and "full" need to be specified at the same time. This will override "--container-version".'
     )
     parser.add_argument('--enable-gpu',
```
deadeyegoodwin, line 317: Does this allow both --enable-gpu and --enable-gpu=true/false? We need to accept --enable-gpu for backwards compat.

jbkyang-nvi: @deadeyegoodwin both works. Please see CI run :)

deadeyegoodwin (approved):

PR #3654, opened 2021-12-03
Improve Python backend execution env tests.

Tabrizian:


qa/L0_backend_python/env/test.sh
```diff
@@ -75,10 +75,24 @@ cp python3.6.tar.gz models/python_3_6/python_3_6_environment.tar.gz
 (cd models/python_3_6 && \
           sed -i "s/^name:.*/name: \"python_3_6\"/" config.pbtxt && \
           echo "parameters: {key: \"EXECUTION_ENV_PATH\", value: {string_value: \"$path_to_conda_pack\"}}" >> config.pbtxt)
-rm -rf ./miniconda
 cp ../../python_models/python_version/model.py ./models/python_3_6/1/
 cp python_backend/builddir/triton_python_backend_stub ./models/python_3_6
 
+# Test conda env without custom Python backend stub
+path_to_conda_pack='$$TRITON_MODEL_DIRECTORY/python_3_8_environment.tar.gz'
+create_conda_env "3.8" "python-3-8"
```
tanmayv25, line 83: We would want to use the python version in the container for this right? Instead of fixing it to 3.8 can't we point it to whatever python version is being used in our container. This would future-proof us and prevent us from updating the test whenever we update the python version.

On the same lines, instead of naming everything python_3_8, can we make it just python and move the generation above python3_6 env creation?

qa/L0_backend_python/env/test.sh
```diff
@@ -104,6 +118,13 @@ if [ $? -ne 0 ]; then
     RET=1
 fi
 
+grep "Python version is 3.8, NumPy version is 1.19.1, and Tensorflow version is 2.3.0" $SERVER_LOG
```
tanmayv25, line 121: Dynamically get the system python version here...

qa/L0_backend_python/env/test.sh
Tabrizian: @tanmayv25 I'd rather not make it dynamic since if the Python is upgraded in container, this test may still break because NumPy or Tensorflow that is used in this test may not work with the future Python versions. For example, the Tensorflow that is installed with Python 3.6 doesn't work with Python 3.8. Also, when the Python version is upgraded, we need to manually update the documentation for versions mentioned in Python backend docs anyway.

tanmayv25: Can you document the requirement here? It will be easier to fix the test when versions are updated?

Tabrizian: Added a comment about this.

PR #3676, opened 2021-12-09
Add support for state initialization

Tabrizian:


src/core/sequence_batch_scheduler.cc
```diff
@@ -149,6 +165,101 @@ SequenceBatchScheduler::Create(
...
+  if (initial_state_itr != initial_state_.end()) {
+    return Status(
+        Status::Code::INVALID_ARG, std::string("State input name '") +
+                                       state.input_name() +
+                                       "' specified more than once.");
+  }
+
+  // Check the dimensions to make sure it doesn't have variable-sized dims.
+  for (auto& dim : initial_state.dims()) {
+    if (dim == -1) {
```
GuanLuo, line 198: I think you should also check if the dim is consistent when the correspondence in both `state` and `initial_state` are not -1?

CoderHam (dismissed):

src/core/sequence_batch_scheduler.cc
```diff
@@ -149,6 +165,101 @@ SequenceBatchScheduler::Create(
...
+    case inference::ModelSequenceBatching_InitialState::StateDataCase::
+        kZeroData: {
+      initial_state_data.data_ = std::make_shared<AllocatedMemory>(
+          total_byte_size, TRITONSERVER_MEMORY_CPU /* memory_type */,
+          0 /* memory_type_id */);
+      break;
+    }
+    case inference::ModelSequenceBatching_InitialState::StateDataCase::
+        kDataFile: {
+      initial_state_data.file_data_.reset(new std::string());
```
GuanLuo, line 231: I don't think you need to store the file data explicitly. Instead of making `data_` a reference to `file_data_`, you can allocate memory of file size in `data_` and copy the file data to `data_`. There is extra copy but it is at load time so it is probably okay, and this drop the extra `file_data_` variable.

GuanLuo (changes requested):

src/core/sequence_batch_scheduler.cc
Tabrizian: Yes, exactly. I wanted to avoid the data_file copy. Ok, I'll change this accordingly.

GuanLuo (approved):

PR #3680, opened 2021-12-09
Add optional input support

GuanLuo:


src/core/infer_request.h
```diff
@@ -201,6 +201,7 @@ class InferenceRequest {
     std::vector<int64_t> shape_with_batch_dim_;
     bool is_shape_tensor_;
     std::shared_ptr<Memory> data_;
+    bool optional_;
```
deadeyegoodwin, line 204: For consistency can you move after is_shape_tensor_ and call it is_optional_tensor_.

src/core/scheduler_utils.cc
```diff
@@ -47,6 +47,10 @@ InitRequiredEqualInputs(
       required_equal_inputs->emplace(
           std::piecewise_construct, std::forward_as_tuple(input->Name()),
           std::forward_as_tuple(input, itr->second));
+    } else {
+      required_equal_inputs->emplace(
```
deadeyegoodwin, line 51: For perf reasons it would be good not to do this unless the model had optional inputs.

src/core/infer_request.cc
```diff
@@ -590,14 +590,17 @@ InferenceRequest::Normalize()
...
   // as is expected by the model.
-  if (original_inputs_.size() != (size_t)model_config.input_size()) {
+  if ((original_inputs_.size() > (size_t)model_config.input_size()) ||
+      (original_inputs_.size() < model_raw_->RequiredInputCount())) {
     return Status(
         Status::Code::INVALID_ARG,
-        "expected " + std::to_string(model_config.input_size()) +
-            " inputs but got " + std::to_string(original_inputs_.size()) +
-            " inputs for model '" + ModelName() + "'");
+        "expected number of inputs between [" +
```
deadeyegoodwin, line 599: Use this message only if input_size != RequiredInputCount, otherwise keep the original message. So models not using optional inputs will not have any change in their error reporting.

src/core/infer_request.cc
```diff
@@ -590,14 +590,17 @@ InferenceRequest::Normalize()
...
+            std::to_string(model_raw_->RequiredInputCount()) + ", " +
```
deadeyegoodwin, line 600: I would not use [ ], just say "... inputs between # and # but got # inputs for model"

deadeyegoodwin (changes requested):

src/core/infer_request.h
GuanLuo: Actually this variable can be removed, this is a spot I missed after I removed 
 the "optional" getter/setter in `InferenceRequest`. As Triton core doesn't really need this information and the backend can check if a input is optional via the `BackendModel` function

deadeyegoodwin (approved):

PR #3684, opened 2021-12-10
Add support for building openVINO 2021.4

tanmayv25:


build.py
```diff
@@ -64,21 +64,28 @@
...
+# To use openvino_21_4, the model configuration should specify
+# the "openvino_21_4" in the "backend" field.
+#
 TRITON_VERSION_MAP = {
     '2.18.0dev': (
         '22.01dev',  # triton container
         '21.11',  # upstream container
         '1.10.0', # ORT
         '2021.2.200',  # ORT OpenVINO
         '2021.2',  # Standalone OpenVINO
```
deadeyegoodwin, line 78: Can we generalize this entry to be a list of OpenVINO versions. Or since we need to support both building from source and installing prebuilt perhaps we can express like (('2021.2', None /* build from source */), ('2021.4', '2021.4.582'))
Also update the documentation above to describe. Also have the first entry in the list be the default so that 'openvino' backend refers to the default.

build.py
```diff
@@ -64,21 +64,28 @@
...
+        '2021.4.582') # Standalone OpenVINO for openvino_21_4
 }
 
 EXAMPLE_BACKENDS = ['identity', 'square', 'repeat']
 CORE_BACKENDS = ['ensemble']
 NONCORE_BACKENDS = [
     'tensorflow1', 'tensorflow2', 'onnxruntime', 'python', 'dali', 'pytorch',
-    'openvino', 'fil', 'fastertransformer', 'tensorrt', 'armnn_tflite'
+    'openvino', 'fil', 'fastertransformer', 'tensorrt', 'armnn_tflite',
+    'openvino_21_4'
```
deadeyegoodwin, line 88: From the list above generate 'openvino_2021_2' and 'openvino_2021_4' backends here instead of hard-coding them

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #3698, opened 2021-12-14
Cleanup s3 local test and add polling coverage for model version updates from S3

CoderHam:
The polling test passes i.e. adding a new version directory to a model triggers the polling and the new version is loaded as expected.

### DIFF
--- a/qa/L0_s3_local/test.sh
+++ b/qa/L0_s3_local/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


qa/L0_s3_local/test.sh
```diff
@@ -93,56 +97,60 @@ awslocal $ENDPOINT_FLAG s3 rm s3://demo-bucket1.0 --recursive --include "*" && \
...
-fi
-set -e
-
-kill $SERVER_PID
-wait $SERVER_PID
-
-# Test with IP
-SERVER_ARGS="--model-repository=s3://127.0.0.1:4572/demo-bucket1.0 --model-control-mode=explicit"
-SERVER_LOG="./inference_server_ip.log"
+# Test with Polling
```
GuanLuo, line 151: Should test the same in explicit mode

GuanLuo (changes requested):
I don't see changes on the source code, so it is polling from the correct cloud repository?

qa/L0_s3_local/test.sh
CoderHam: Already tested in explicit mode in L0_storage_S3

CoderHam:
> I don't see changes on the source code, so it is polling from the correct cloud repository?

Yes that is correct. It appears my earlier observation was incorrect due to sleep that was too low.

GuanLuo (approved):
> Already tested in explicit mode in L0_storage_S3

Can you create a follow up ticket to group the S3 tests into one? I realize we have different test folders that just test different flavors of S3, which can be hard to track. For instance, with this change, the generic model loading functionality tests is across L0_s3_local and L0_storage_S3, which can be confusing. What we can do is having the single test where the first half tests the generic cases against the common S3 usage, and the second half would be sanity check on other flavors are handled properly

qa/L0_s3_local/test.sh
```diff
@@ -44,15 +44,19 @@ CLIENT_LOG="./client.log"
...
 DATADIR="/data/inferenceserver/${REPO_VERSION}/qa_model_repository"
+BACKENDS="graphdef libtorch onnx plan savedmodel"
 
-MODELDIR="models"
-rm -rf ./$MODELDIR && mkdir $MODELDIR
-cp -r $DATADIR/*_float32_float32_float32 $MODELDIR/.
+rm -rf models && mkdir models
+for BACKEND in $BACKENDS; do
+    cp -r $DATADIR/${BACKEND}_float32_float32_float32 models/.
+    # Remove version policy from config.pbtxt
```
tanmayv25, line 52: Why is this needed?

CoderHam:
> Can you create a follow up ticket to group the S3 tests into one?

Sure

qa/L0_s3_local/test.sh
CoderHam: The version policy allows 1 and 3 but we want to autoload 4 when and unload 3 when 4 is added to model repo and polling is enabled

tanmayv25 (approved):

PR #3699, opened 2021-12-14
Update Lifecycle Tests to Check Specific Errors

dyastremsky:
This updates the lifecycle tests to check specific errors are returned when a model load fails. The tests check for errors in the server logs for model control modes NONE and EXPLICIT.

I am not seeing errors returned for the tests in POLL mode (which load the models successfully, then modify the models to fail reloading). If we expect those to return errors too, the tests can be updated.

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -288,12 +288,6 @@ class LifeCycleTest(tu.TestResultCollector):
                 self.assertIn(
                     "Request for unknown model: 'graphdef_float32_float32_float32' is not found",
                     ex.message())
-                with open(os.environ["SERVER_LOG"]) as log:
-                    log_text = log.read()
-                    self.assertIn("failed to open text file for read", log_text)
-                    self.assertIn(
-                        "{}/config.pbtxt: No such file or directory".format(
-                            model_name), log_text)
 
         # And other models should be loaded successfully
         try:
@@ -611,8 +605,8 @@ class LifeCycleTest(tu.TestResultCollector):
                 "expected error for unavailable model " + savedmodel_name)
         except Exception as ex:
             self.assertIn(
-                "Request for unknown model: 'savedmodel_float32_float32_float32' has no available versions",
-                ex.message())
+                "Request for unknown model: '{}' has no available versions".
+                format(savedmodel_name), ex.message())
 
         # Add back the same model. The status/stats should be reset.
         try:
@@ -1862,10 +1856,6 @@ class LifeCycleTest(tu.TestResultCollector):
         except Exception as ex:
             self.assertIn("version 2: Internal: GPU instances not supported",
                           ex.message())
-            with open(os.environ["SERVER_LOG"]) as log:
-                self.assertIn(
-                    "failed to load '{}' version 2: Internal: GPU instances not supported"
-                    .format(model_name), log.read())
 
         # Make sure version 1 of the model is available, and version 2 is not
         try:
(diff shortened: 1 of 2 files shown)


GuanLuo:
> I am not seeing errors returned for the tests in POLL mode (which load the models successfully, then modify the models to fail reloading). If we expect those to return errors too, the tests can be updated.

Which tests are you referring to? In POLL mode, the server should log the error if model loading failed, and we should check the error is logged if we haven't had test to cover that.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -284,9 +285,15 @@ def test_parse_error_no_model_config(self):
                     "' to be ignored due to polling failure")
 
             except Exception as ex:
-                self.assertTrue(ex.message().startswith(
-                    "Request for unknown model: 'graphdef_float32_float32_float32' is not found"
-                ))
+                self.assertIn(
+                    "Request for unknown model: 'graphdef_float32_float32_float32' is not found",
+                    ex.message())
+                with open(os.environ["SERVER_LOG"]) as log:
```
GuanLuo, line 291: If we want to check if certain message is in the server log, the convention is to do so in the bash script, look up "grep" for usage.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -1854,7 +1860,12 @@ def test_model_reload_fail(self):
             triton_client.load_model(model_name)
             self.assertTrue(False, "expecting load failure")
         except Exception as ex:
-            pass
+            self.assertIn("version 2: Internal: GPU instances not supported",
+                          ex.message())
+            with open(os.environ["SERVER_LOG"]) as log:
```
GuanLuo, line 1865: same

GuanLuo (changes requested):

dyastremsky:
> > I am not seeing errors returned for the tests in POLL mode (which load the models successfully, then modify the models to fail reloading). If we expect those to return errors too, the tests can be updated.



> > I am not seeing errors returned for the tests in POLL mode (which load the models successfully, then modify the models to fail reloading). If we expect those to return errors too, the tests can be updated.
> 
> Which tests are you referring to? In POLL mode, the server should log the error if model loading failed, and we should check the error is logged if we haven't had test to cover that.

test_dynamic_model_load_unload
test_dynamic_version_load_unload
test_dynamic_model_modify
test_dynamic_file_delete
test_multiple_model_repository_polling

qa/L0_lifecycle/lifecycle_test.py
dyastremsky: Thanks for this tip. I'll update the ticket to reflect the convention. If we ever want to think about changing the convention in the future, it'd likely be helpful to provide the exact error via the Python script (logs) rather than "Test Failed," since there are multiple points of failure.

GuanLuo: That actually reminds me something I found when looking at the other lifecycle ticket. The client log is actually not being printed (cat) at the end in the case of error (while other tests do), you should probably fix that as you are working on it. Note that you may need to remove the `rm -f $CLIENT_LOG` so the client log at the end is complete, or you can clear the log before every test and print it as soon as the test failed. Either way is fine as long as we do report the logs on failure

GuanLuo:
I don't think these tests is targeting the case of failing to load a model, they are testing if the model manager loads/unloads model accordingly and the result is checked by the model liveness / infer in the Python script. I guess you can double check the server log if you want to, as the model manager also log its actions.

qa/L0_lifecycle/lifecycle_test.py
dyastremsky: Added.

dyastremsky:
@GuanLuo The changes you mentioned should be reflected now.

GuanLuo (approved):

dyastremsky:
Hey @GuanLuo, could you submit another review? I hadn't realized that rebasing off the main branch (for CI purposes; there were no merge conflicts) would dismiss your review.

GuanLuo (approved):

PR #3703, opened 2021-12-14
Update entry point to launch tritonserver directly if detect Vertex AI

GuanLuo:


GuanLuo:
@mengdong for viz

nvidia_entrypoint.sh
```diff
@@ -82,7 +82,12 @@ fi
 echo
 
 if [[ $# -eq 0 ]]; then
-  exec "/bin/bash"
+  # If detect Vertex AI environment, launch tritonserver with supplied arguments
```
deadeyegoodwin, line 85: I think this needs to be the outer conditional. Right now the new code is only running when there are no args.

deadeyegoodwin (changes requested):
What is your plan to test this?

GuanLuo:
> What is your plan to test this?

I added a test that is different to the typical test structure as Triton is tied with the container when launch via entry point, so we actually perform minimum liveness checks from outside the container.

deadeyegoodwin (approved):

nvidia_entrypoint.sh
```diff
@@ -81,8 +81,13 @@ fi
 
 echo
 
-if [[ $# -eq 0 ]]; then
-  exec "/bin/bash"
+# If detect Vertex AI environment, launch tritonserver with supplied arguments
+if [[ ! -z "$AIP_MODE" ]]; then
+  exec /opt/tritonserver/bin/tritonserver "$@"
```
deadeyegoodwin, line 86: Strange that you need to full path, because "docker run <image> tritonserver --model-repository=<x>" works.
I'm concerned that if full path is required then does that mean that the env is not set property as well?

GuanLuo: It may not need the full path, I changed it to resolve the "not found" issue but it later seems to be related to exec usage and the the way quote ("") should be used (quote on the whole thing will make linux use the full string for executable name). I didn't revert the change just in case.

deadeyegoodwin (approved):

Issue #3135, opened 2021-07-20
Triton requiring config.pbtxt when loading models from s3 (MinIO)?


LMarino1:
Using MinIO for S3, it seems starting with release 21.04, Triton is no longer automatically generating config files for models that lack them and instead errors out.

Logs from attempting to run version 21.06.1:
<pre>
=============================
== Triton Inference Server ==
=============================

NVIDIA Release 21.06 (build 24449615)

Copyright (c) 2018-2021, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.

Various files include modifications (c) NVIDIA CORPORATION.  All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

WARNING: The NVIDIA Driver was not detected.  GPU functionality will not be available.
   Use Docker with NVIDIA Container Toolkit to start this container; see
   https://github.com/NVIDIA/nvidia-docker.

I0720 15:30:54.928850 1 libtorch.cc:987] TRITONBACKEND_Initialize: pytorch
I0720 15:30:54.928886 1 libtorch.cc:997] Triton TRITONBACKEND API version: 1.4
I0720 15:30:54.928892 1 libtorch.cc:1003] 'pytorch' TRITONBACKEND API version: 1.4
2021-07-20 15:30:58.469303: I tensorflow/stream_executor/platform/default/dso_loader.cc:49] Successfully opened dynamic library libcudart.so.11.0
I0720 15:30:58.518754 1 tensorflow.cc:2165] TRITONBACKEND_Initialize: tensorflow
I0720 15:30:58.518777 1 tensorflow.cc:2175] Triton TRITONBACKEND API version: 1.4
I0720 15:30:58.518785 1 tensorflow.cc:2181] 'tensorflow' TRITONBACKEND API version: 1.4
I0720 15:30:58.518809 1 tensorflow.cc:2205] backend configuration:
{}
I0720 15:30:58.645847 1 onnxruntime.cc:1969] TRITONBACKEND_Initialize: onnxruntime
I0720 15:30:58.645922 1 onnxruntime.cc:1979] Triton TRITONBACKEND API version: 1.4
I0720 15:30:58.645949 1 onnxruntime.cc:1985] 'onnxruntime' TRITONBACKEND API version: 1.4
[...]

PR #3704, opened 2021-12-14
Fix model load failure when using cloud storage without model config file and disabled strict model config

CoderHam:
Fixes https://github.com/triton-inference-server/server/issues/3135

### DIFF
--- a/src/core/model_config_utils.cc
+++ b/src/core/model_config_utils.cc
@@ -637,10 +637,8 @@ GetNormalizedModelConfig(
 {
   // If 'autofill' then the configuration file can be empty.
   const auto config_path = JoinPath({path, kModelConfigPbTxt});
-
   bool model_config_exists;
   RETURN_IF_ERROR(FileExists(config_path, &model_config_exists));
-
   if (autofill && !model_config_exists) {
     config->Clear();
   } else {
--- a/src/core/model_config_utils.cc
+++ b/src/core/model_config_utils.cc
@@ -638,13 +638,8 @@ GetNormalizedModelConfig(
   // If 'autofill' then the configuration file can be empty.
   const auto config_path = JoinPath({path, kModelConfigPbTxt});
 
-  // Cloud file systems may report error when config file is missing. As a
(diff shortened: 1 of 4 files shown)


src/core/model_config_utils.cc
```diff
@@ -637,8 +637,15 @@ GetNormalizedModelConfig(
 {
   // If 'autofill' then the configuration file can be empty.
   const auto config_path = JoinPath({path, kModelConfigPbTxt});
+
+  // Cloud file systems may report error when config file is missing. As a
+  // workaround, we assume that an error indicates that the file is missing.
```
GuanLuo, line 642: Shouldn't we improve `FileExists()` function for cloud file systems that it sets `model_config_exists` and return Status::Success if the file is missing? So when `FileExists()` does return error (i.e. can't connect to storage) and we will report it instead of silently ignoring it.

CoderHam: In `S3FileSystem::FileExists()`
For the case when `head_object_outcome.IsSuccess()` is false in the filesystem we cannot know if it is due to a credential 
 or access error or if the file itself is missing. Due to this reason the handling for the model config when autofill is enabled with be different since we can proceed with the assumption that the config file is missing. 
 
 The same is also true for GCS / Azure. 
 
 We can log the error that we were unable to access the <config file path> if you think that is necessary.

GuanLuo: `head_object_outcome` doesn't have a corresponding error code for "file not found"? In either case, I think at least the logic you have here should be moved to `FileExists()` function so the [interface](https://github.com/triton-inference-server/server/blob/main/src/core/filesystem.h#L89-L92) is consistent: `exists` is set to whether the file exists and error is returned if unable to perform the check. Otherwise your change here implies returning error also means the file doe not exist..

CoderHam: Found something for S3, looking for similar error codes for GCS and Azure

tanmayv25 (approved):

PR #3707, opened 2021-12-15
Check input shape is fixed on infer request normalization

GuanLuo:
This must be enforce otherwise CopyAsNull() is operating on invalid shape information

### DIFF
--- a/src/core/infer_request.cc
+++ b/src/core/infer_request.cc
@@ -692,30 +692,43 @@ InferenceRequest::Normalize()
               "' for '" + ModelName() + "'");
     }
 
-    if (!CompareDimsWithWildcard(input_config->dims(), *shape)) {
-      DimsList full_dims;
-      if (model_config.max_batch_size() > 0) {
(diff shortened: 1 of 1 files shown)


src/core/infer_request.cc
```diff
@@ -707,6 +707,18 @@ InferenceRequest::Normalize()
               ", got " + DimsListToString(input.OriginalShape()));
     }
 
+    // Must make sure input shape is specified which is not checked in
+    // compare function above
+    for (const auto& dim : input.OriginalShape()) {
```
deadeyegoodwin, line 712: Instead of iterating over the shape again can we just replace CompareDimsWithWildcard above with a local compare function that does exactly what we want and gives appropriate error messages.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #3712, opened 2021-12-17
Fix compose script by tagging openvino backend dir name before trying to copy it out

tanmayv25:
openvino backend directory name is now tagged to reflect the version. In order to compose image with mulitple openvino backends user must explicitly specify `--backend openvino_<vers_str>`. To support --backend openvino, we tag the name with default ov version from the build.py.

### DIFF
--- a/compose.py
+++ b/compose.py
@@ -92,6 +92,10 @@ COPY --chown=1000:1000 --from=full /opt/tritonserver/include include/
 def add_requested_backends(ddir, dockerfile_name, backends):
     df = "# Copying over backends \n"
     for backend in backends:
+        if backend == 'openvino':
+            import build
+            ver = next(iter(build.TRITON_VERSION_MAP.values()))
+            backend = build.get_tagged_backend(backend, ver[4][0])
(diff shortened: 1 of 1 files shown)


deadeyegoodwin (approved):

compose.py
```diff
@@ -92,6 +92,10 @@ def start_dockerfile(ddir, images, argmap, dockerfile_name, backends):
 def add_requested_backends(ddir, dockerfile_name, backends):
     df = "# Copying over backends \n"
     for backend in backends:
+        if backend == 'openvino':
+            import build
+            ver = next(iter(build.TRITON_VERSION_MAP.values()))
```
jbkyang-nvi, line 97: For my understanding.. why is this better than `ver = build.TRITON_VERSION_MAP.values()[4][0]`? Does `ver` need to be an iterator?

tanmayv25: `dict.values()` in python 3.x returns `dict_values` object which is not subscriptable. You can read more about dict-views from [here](https://docs.python.org/3.7/library/stdtypes.html?highlight=dict%20values#dict-views).

PR #3760, opened 2022-01-05
Add test for empty io in BLS

Tabrizian:


qa/python_models/bls/model.py
```diff
@@ -140,6 +140,20 @@ def _get_gpu_bls_outputs(self, input0_pb, input1_pb):
 
         return output0.to_dlpack(), output1.to_dlpack()
 
+    def test_zero_length_io(self):
+        model_name = 'identity_fp32'
+        input0 = np.zeros([1, 0], dtype=np.float32)
```
CoderHam, line 145: Is not `input0 = np.zeros([0], dtype=np.float32)` sufficient?

CoderHam (changes requested):

qa/python_models/bls/model.py
Tabrizian: @CoderHam Since identity model has `max_batch_size > 0`  it needs to have two dims.

CoderHam (approved):

PR #3766, opened 2022-01-05
Update windows min with newer versions

deadeyegoodwin:


CoderHam (approved):
Note to make corresponding update in the release notes Windows Support section

PR #3768, opened 2022-01-06
Move triton SDK building to its own section and add reference to it i…

matthewkotila:
…n build doc


matthewkotila:
It seems good enough to just include reference to SDK building in the build doc, rather than also adding the SDK building reference to the top-level readme

@deadeyegoodwin thoughts?

deadeyegoodwin (approved):

PR #3771, opened 2022-01-06
Simple Java Tests

dyastremsky:
The simple_example and simple_ensemble tests for the Java simple client. They test the different command line arguments and datatypes available for the simple model. They also test running an ensemble model.


dyastremsky:
@jbkyang-nvi This should be all set to push to the Java prerelease branch! Please review if it's good to merge.

qa/L0_java_simple_example/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+    <repositories>
+        <repository>
+            <id>sonatype-nexus-snapshots</id>
+            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+        </repository>
+    </repositories>
+    <pluginRepositories>
+        <pluginRepository>
+            <id>sonatype-nexus-snapshots</id>
+            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
```
jbkyang-nvi, line 32: Would it be better to make the changes in javacpp-presets instead? So that the users can use the pom.xml there too

dyastremsky: I can do that! Would you be okay waiting until tomorrow's meeting to discuss that with Jack? I'm assuming the pom.xml works for Jack and/or he knows why it's not working for us. Based on that conversation, we can update the test to use the javacpp-presets pom.xml (and our instructions for how users can run this example).

qa/L0_java_simple_example/test.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+# Copyright 2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jbkyang-nvi, line 2: 2022

dyastremsky: Nice catch. Updated.

qa/L0_java_simple_example/pom.xml
dyastremsky: Since we found a way to build using the javacpp-presets pom, I removed it here.

qa/L0_java_simple_example/test.sh
```diff
@@ -0,0 +1,110 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Set up test files
```
jbkyang-nvi, line 28: Can you add a line that points to the installation instructions here?

dyastremsky: Pushed. Let me know if that's what you had in mind.

qa/L0_java_simple_example/test.sh
```diff
@@ -0,0 +1,111 @@
...
+rm -f *.log
+RET=0
+
+# Run with default settings
+$BASE_COMMAND -Dexec.args="-r $MODEL_REPO" >>client.log 2>&1
+if [ $? -ne 0 ]; then
+    RET=1
+fi
+
+if [ `grep -c "1 - 1 = 0" client.log` != "18" ]; then
```
CoderHam, line 50: Are all input values for Input0 and Input1 = 1? Why 18 and not 16?

qa/L0_java_simple_example/test.sh
```diff
@@ -0,0 +1,111 @@
...
+if [ $? -ne 0 ]; then
+    RET=1
+fi
+
+if [ `grep -c "Server side auto-completed config" client.log` != "2" ]; then
+    echo -e "\n***\n*** Failed. Expected 'Server side auto-completed config'\n***"
+    RET=1
+fi
+
+# Run with memory set to system
```
CoderHam, line 66: Run with I/O using system shared memory

CoderHam: Do we not want a cuda shared memory version of the same?

qa/L0_java_simple_example/test.sh
```diff
@@ -0,0 +1,111 @@
...
+
+if [ `grep -c "data_type: TYPE_FP32" client.log` != "8" ]; then
+    echo -e "\n***\n*** Failed. Expected 4 'data_type: TYPE_FP32'\n***"
+    RET=1
+fi
+
+# Run ensemble
+sed -i 's/"simple"/"ensemble_add_sub_int32_int32_int32"/g' $SAMPLES_REPO/Simple.java
+$BASE_COMMAND -Dexec.args="-r $MODEL_REPO -v" >>client.log 2>&1
+sed -i 's/"ensemble_add_sub_int32_int32_int32"/"simple"/g' $SAMPLES_REPO/Simple.java
```
CoderHam, line 95: Why not pass the model name as an argument instead?

CoderHam (changes requested):

qa/L0_java_simple_example/test.sh
dyastremsky: Jack wrote simple.java, which was already merged into the presets repo.  It's based on simple.cc, so the command line arguments are similar (the model name is defined in simple.cc rather than as a command line argument). This is just using that program for the test.

qa/L0_java_simple_example/test.sh
dyastremsky: In the current version of simple.java, pinned and GPU memory are not options. Since Java does not have preprocessor flags like C++, when simple.java was written, the flags were missing. As a result, simple.java acts as if TRITON_ENABLE_GPU is not set. We have a [ticket ](https://jirasw.nvidia.com/browse/DLIS-3357?filter=-3)to fix the logic around preprocessor flags; I added a reminder to add these tests then.

I made a mock solution but we want to figure out the design, which is why we put it into a ticket. I did manually check the GPU and pinned cases, which work.

qa/L0_java_simple_example/test.sh
dyastremsky: In one of the cases that are run, they are. So that would be 16. In the other two inferences that run, input1 is always 1 and input0 has different values including one 1. So each of those have one "1 - 1 = 0". That gets you two 18.

Simple.java checks the outputs and fails if any are incorrect, so this is a check to make sure it ran until the end. "1 - 1 = 0" would be the last checked and printed value, so this requires simple.java to run and check the inference successfully.

CoderHam (approved):

jbkyang-nvi (approved):

PR #3792, opened 2022-01-11
Java cpp readme

jbkyang-nvi:
merge after https://github.com/bytedeco/javacpp-presets/pull/1124

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <!--
-# Copyright 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ release is
   linked directly into your application for edge and other in-process
   use cases.
 
+* A [Java API](docs/inference_protocols.md#java-api) is similar to C API and
+  allows Triton to be linked directly into your Java application.
+
 * [Metrics](docs/metrics.md) indicating GPU utilization, server
   throughput, and server latency. The metrics are provided in
   Prometheus data format.
--- a/README.md
+++ b/README.md
@@ -103,11 +103,11 @@ release is
   protocol](https://github.com/kserve/kserve/tree/master/docs/predict-api/v2).
 
 * A [C API](docs/inference_protocols.md#c-api) allows Triton to be
-  linked directly into your application for edge and other in-process
+  linked directly to your application for edge and other in-process
   use cases.
 
 * A [Java API](docs/inference_protocols.md#java-api) is similar to C API and
-  allows Triton to be linked directly into your Java application.
+  allows Triton to be linked directly to your Java application.
 
 * [Metrics](docs/metrics.md) indicating GPU utilization, server
   throughput, and server latency. The metrics are provided in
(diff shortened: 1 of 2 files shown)


docs/inference_protocols.md
```diff
@@ -111,3 +111,42 @@ endpoints for Triton. These endpoints use the C API to communicate
...
+generated from `tritonserver.java`.
+A simple example using the Java API can be found in
+[Samples folder](https://github.com/bytedeco/javacpp-presets/tree/master/tritonserver/samples)
+which includes `Simple.java` which is similar to `simple.cc`. Please refer to
+[sample usage documentation](https://github.com/bytedeco/javacpp-presets/tree/master/tritonserver#sample-usage)
+to learn about how to build and run `Simple.java`.
+
+### Java API setup instructions
+
+The current snapshot is based on Triton container version `21.12`. To setup your
```
dyastremsky, line 133: Nit: "To set up your environment"

README.md
```diff
@@ -106,6 +106,9 @@ release is
   linked directly into your application for edge and other in-process
   use cases.
 
+* A [Java API](docs/inference_protocols.md#java-api) is similar to C API and
+  allows Triton to be linked directly into your Java application.
```
dyastremsky, line 110: Nit: "linked directly to"

docs/inference_protocols.md
```diff
@@ -111,3 +111,42 @@ endpoints for Triton. These endpoints use the C API to communicate
...
+to create bindings around Tritonserver to create Java API. It requires the user
+to link a pre-built library to their application using Maven.   
+
+The API is documented in 
+[tritonserver.java](https://github.com/bytedeco/javacpp-presets/blob/master/tritonserver/src/gen/java/org/bytedeco/tritonserver/global/tritonserver.java).
+Alternatively, the user can refer to the web version [API docs](http://bytedeco.org/javacpp-presets/tritonserver/apidocs/)
+generated from `tritonserver.java`.
+A simple example using the Java API can be found in
+[Samples folder](https://github.com/bytedeco/javacpp-presets/tree/master/tritonserver/samples)
+which includes `Simple.java` which is similar to `simple.cc`. Please refer to
```
dyastremsky, line 127: Could be good to add the link to simple.cc here (https://github.com/triton-inference-server/server/blob/main/src/servers/simple.cc)

docs/inference_protocols.md
```diff
@@ -111,3 +111,42 @@ endpoints for Triton. These endpoints use the C API to communicate
...
+ $ docker run -it --gpus=all -v ${pwd}:/workspace nvcr.io/nvidia/tritonserver:21.12-py3 bash
+```
+2. Then install `java sdk` and `maven`:
+```
+ $ apt update && apt install -y openjdk-11-jdk
+ $ wget https://archive.apache.org/dist/maven/maven-3/3.8.4/binaries/apache-maven-3.8.4-bin.tar.gz
+ $ tar zxvf apache-maven-3.8.4-bin.tar.gz
+ $ export PATH=/opt/tritonserver/apache-maven-3.8.4/bin:$PATH
+```
+3. Maven requires a `pom.xml` file to compile. The pre-built snapshots are hosted
```
dyastremsky, line 146: For this, related to your [comment in the simple test PR](https://github.com/triton-inference-server/server/pull/3771#pullrequestreview-849843641), we should decide whether we:

1. Update the pom.xml in javacpp-presets to work with the provided instructions.
2. Provide the pom.xml for the simple example (e.g. linking to the test pom.xml), so use can base their poms off it.

I think we may want to wait until we talk to Jack, since the current instructions did not work for us. It's helpful that you write which snapshot to use, but users may run into the same issue as we did. It'd be great to include the sample working pom to avoid that.

dyastremsky (commented):
Great work! Added a few small comments.

dyastremsky (approved):

docs/inference_protocols.md
dyastremsky: Great updates on building with the current pom. Approving. These build instructions are clear and should work for the user, including if they use the linked samples pom.xml for simple (as an example).

PR #3793, opened 2022-01-12
Fix type bytes for implicit state management

Tabrizian:
The ONNX model tests using the string datatype only put the INPUT in the OUTPUT and OUTPUT_STATE. I couldn't find an operator that performs concatenation of two strings so that the cast to INT32 doesn't fail when the string is empty.


GuanLuo (dismissed):

qa/L0_implicit_state/test.sh
```diff
@@ -26,7 +26,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 export ENSEMBLES=0
-BACKENDS=${BACKENDS:="onnx plan"}
+BACKENDS=${BACKENDS:="onnx"}
```
tanmayv25, line 29: Why are we not running implicit state tests for tensorrt?

Tabrizian: Good catch. Fixed. I'm not going to re-run the CI since the previous commit didn't have this issue and it passed the CI.

tanmayv25 (dismissed):

GuanLuo (approved):

PR #3814, opened 2022-01-14
Fix parsing of --override-backend-cmake-arg

CoderHam:
Prior to change:
Using `--override-backend-cmake-arg=python:TRITON_ENABLE_GPU=OFF` resulted in incorrect parsing
```
backend "python" CMake override "-Dpython=TRITON_ENABLE_GPU=OFF"
```

After change:

```
backend "python" CMake override "-DTRITON_ENABLE_GPU=OFF"
```

### DIFF
--- a/build.py
+++ b/build.py
@@ -1830,7 +1830,6 @@ if __name__ == '__main__':
             len(parts) != 2,
             '--override-backend-cmake-arg must specify <backend>:<name>=<value>'
         )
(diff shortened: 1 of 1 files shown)


CoderHam:
@deadeyegoodwin @dzier this should be cherry picked into r22.01

deadeyegoodwin (approved):

PR #3826, opened 2022-01-19
Add testing for sequence models with BLS

Tabrizian:


CoderHam (dismissed):

qa/python_models/bls/model.py
```diff
@@ -77,6 +77,60 @@ def test_bls_wrong_inputs(self):
         infer_response = infer_request.exec()
         self.assertTrue(infer_response.has_error())
 
+    def test_bls_sequence(self):
```
tanmayv25, line 80: Shouldn't we test with multiple sequences(correlation ids) as well? Two should be enough I guess. The requests from these two should be interleaved.

tanmayv25 (changes requested):

tanmayv25 (approved):

PR #3827, opened 2022-01-19
Memory Growth Tests for Java API

dyastremsky:
Add a test to validate that memory growth does not occur over time for the Java bindings.

### DIFF
--- a/qa/L0_java_memory_growth/Simple.java
+++ b/qa/L0_java_memory_growth/Simple.java
@@ -36,7 +36,6 @@ import static org.bytedeco.tritonserver.global.tritonserver.*;
 
 public class Simple {
     static final double TRITON_MIN_COMPUTE_CAPABILITY = 6.0;
-    private static boolean done = false;
 
     static void FAIL(String MSG) {
         System.err.println("Cuda failure: " + MSG);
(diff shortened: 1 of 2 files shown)


dyastremsky:
Currently, there is high memory growth (from 25MB to 1.7GB). I am investigating to see if I can find a memory leak or any issues within the test. The example is based on the simple.java example used as a sample.

CoderHam:
Maybe I missed this but why do we elect to not put on java simple example in the client repo?

dyastremsky:
> Maybe I missed this but why do we elect to not put on java simple example in the client repo?

This is for the Java API, not the Java client. Will defer to @jbkyang-nvi in case the client makes more sense.

jbkyang-nvi:
We plan to move all of [javacpp-presets/tritonserver] related things to Triton eventually. I think its reasonable to have only 1 copy for now, before we move everything over.

qa/L0_java_memory_growth/Simple.java
```diff
@@ -0,0 +1,1018 @@
...
+          if(ValidateMemoryGrowth(.1f)){
+            System.out.println("Memory growth test passed");
+          } else {
+            System.out.println("Memory growth test FAILED");
+          }
+        };
+      Thread memory_thread = new Thread(runnable);
+      memory_thread.start();
+
+      for(int i = 0; i < 1000000; i++){
```
GuanLuo, line 1008: Can you expose command line option to set the number of iteration? Then we can indirectly control the length of the test in `test.sh`

dyastremsky: Good recommendation. Done.

jbkyang-nvi (commented):
Can we have a single instance of Simple, and call different functions for each instance? It is a bit hard to review all of them.

jackyh:
as you know, I've already tried the "vibration version", since the mem allocated will still has some range of 0 to 100MB, so that sort of test will still fail. But, that should not be an issue, since it will not grow gradually.

dyastremsky:
Closing due to https://github.com/triton-inference-server/server/pull/4259 (i.e. Java API memory growth test) being merged.

PR #3835, opened 2022-01-20
Cleanup L0_backend_python test and enable testing on Jetson

CoderHam:


qa/L0_backend_python/examples/test.sh
```diff
@@ -64,7 +67,7 @@ fi
 grep "PASS" add_sub_client.log
 if [ $? -ne 0 ]; then
     echo -e "\n***\n*** Failed to verify pytorch example. \n***"
-    cat pytorch_client.log
+    cat add_sub_client.log
```
tanmayv25, line 70: Define a variable CLIENT_LOG before running each example and then use that in those example section.

qa/L0_backend_python/test.sh
```diff
@@ -298,44 +308,53 @@ done
...
 if [ $? -ne 0 ]; then
   RET=1
 fi
 
-(cd bls && bash -ex test.sh)
-if [ $? -ne 0 ]; then
-  RET=1
+# Skip variants for Jetson since already built without GPU Tensors support
+if [ "$TEST_JETSON" == "0" ]; then
+  (cd variants && bash -ex test.sh)
```
tanmayv25, line 354: Why are there separate blocks for skipping on jetson? Why not have a single $TEST_JSON == "0" block and collect all the tests to be skipped there?

tanmayv25 (changes requested):

Tabrizian (approved):

qa/L0_backend_python/test.sh
CoderHam: Done

tanmayv25 (approved):

qa/L0_backend_python/test.sh
```diff
@@ -298,42 +308,47 @@ done
 kill $SERVER_PID
 wait $SERVER_PID
 
-(cd env && bash -ex test.sh)
-if [ $? -ne 0 ]; then
-  RET=1
-fi
+# Disable env test for Jetson since build is non-dockerized and cloud storage repos are not supported
```
Tabrizian, line 311: One comment that I forgot, please file a ticket for enabling a subset of the env tests in Jetson since it touches on various platform specific settings and it might not work on jetson. It is a widely used feature in Python backend and I think we should have some testing for it.

CoderHam: Sure. We will ideally need to wait for a dockerized jetson build because we do not want to play with the environments on the CI machine. Additionally the AWS cloud storage support is not added on jetson yet so they will remain disabled.

CoderHam: PS: Filed the ticket - DLIS 3377

Tabrizian: Thanks Hemant.

tanmayv25 (approved):

dyastremsky (approved):

qa/L0_backend_python/python_test.py
```diff
@@ -53,36 +56,40 @@ def _infer_help(self, model_name, shape, data_type):
             output0 = result.as_numpy('OUTPUT0')
             self.assertTrue(np.all(input_data_0 == output0))
```
dyastremsky, line 57: May not be worth it for this PR, but heads up for the future that Numpy has its own testing class that returns detailed messages for assertions like this one. Example for this case: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

CoderHam:
@Tabrizian @tanmayv25 I had to disable async BLS on Jetson since it is not supported with python3.6. Having multiple versions of python are causing build issues where cmake picks the newer version despite using symlinks to have python point to python3.6.

I tried using anaconda but it is not officially supported on jetson.

tanmayv25 (approved):

qa/L0_backend_python/python_test.py
Tabrizian: Filed a ticket.

tanmayv25 (approved):

PR #3836, opened 2022-01-21
Add more testing for implicit state

Tabrizian:
I found one minor bug as a result of this additional testing. The bug was that it was not possible to change the device of the internal state in the middle of the sequence. I added test coverage for it. For example, if you requested CPU in the middle of the sequence and all the internal state was stored in GPU you would still get GPU memory back.


qa/L0_implicit_state/implicit_state.py
```diff
@@ -0,0 +1,135 @@
...
+        inputs[0].set_data_from_numpy(np.random.randint(5, size=[1], dtype=np.int32))
+        inputs[1].set_data_from_numpy(np.asarray([0], dtype=np.int32))
+
+        with self.assertRaises(InferenceServerException) as e:
+            triton_client.infer(model_name="wrong_internal_state", inputs=inputs, sequence_id=2, sequence_start=True)
+
+        self.assertEqual(str(e.exception), "state 'undefined_state' is not a valid state name.")
+
+    def test_no_update(self):
+	    # Test implicit state without updating any state
```
GuanLuo, line 72: Can you elaborate what this test is doing? Why the implicit state won't be updated when it receives a sequence of requests? Is this just how the model work? i.e. it always output the same data for the "state".

qa/L0_implicit_state/implicit_state.py
```diff
@@ -0,0 +1,135 @@
...
+            self.assertTrue(str(e.exception).startswith("unexpected inference output 'OUTPUT_STATE' for model"))
+
+    def test_request_output(self):
+        triton_client = tritonhttpclient.InferenceServerClient("localhost:8000")
+        inputs = []
+        inputs.append(tritonhttpclient.InferInput('INPUT', [1], 'INT32'))
+        inputs[0].set_data_from_numpy(np.asarray([1], dtype=np.int32))
+
+        outputs = []
+        outputs.append(tritonhttpclient.InferRequestedOutput('OUTPUT_STATE'))
```
GuanLuo, line 118: So different from the previous case,  in this test, the request for the state works because the model config specifies the state in output field as well?

GuanLuo (approved):
The fix LGTM, just want some clarification on the test cases

qa/L0_implicit_state/implicit_state.py
Tabrizian: I added this test to make sure that `TRITONBACKEND_StateUpdate` must be called to update the state. In the phase 2 of implicit state, there is a feature request to add the capability to not update some state based on some inputs (cloning of implicit state ticket). I wanted to make sure that this is behaving as expected.

qa/L0_implicit_state/implicit_state.py
Tabrizian: Right. The model must specify the state in the output field if it wants to allow the client to request it.

qa/L0_implicit_state/implicit_state.py
GuanLuo: Are you saying that `TRITONBACKEND_StateUpdate` must NOT be called?

Tabrizian: No, I wanted to make sure that not calling it will lead to the state not being updated. It should be called if you want to update the internal state.

PR #3843, opened 2022-01-21
Add batcher timestamp to report queue time properly and to maintain batcher functionality

GuanLuo:
Now in the case of oldest sequence batcher, the queue time is tracked (not counting towards overhead)
Before:
```
...
Avg request latency: 68538 usec (overhead 60622 usec + queue 2080 usec + compute input 172 usec + compute infer 5525 usec + compute output 139 usec)
...
```
After:
```
...
Avg request latency: 69777 usec (overhead 168 usec + queue 63749 usec + compute input 168 usec + compute infer 5554 usec + compute output 138 usec)
...
```

### DIFF
--- a/src/core/sequence_batch_scheduler.cc
+++ b/src/core/sequence_batch_scheduler.cc
@@ -1,4 +1,4 @@
-// Copyright 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2018-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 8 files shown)


tanmayv25 (dismissed):

src/core/sequence_batch_scheduler.cc
```diff
@@ -513,6 +513,9 @@ SequenceBatchScheduler::Enqueue(std::unique_ptr<InferenceRequest>& irequest)
       irequest->Trace(), TRITONSERVER_TRACE_QUEUE_START,
       irequest->QueueStartNs());
 
+  // Record time at the beginning of the batcher queueing
+  irequest->CaptureBatcherStartNs();
```
deadeyegoodwin, line 517: I guess it is ok, and expected, that the dynaseqbatcher will overwrite this for oldest? You may want to add something in the comment about that.

GuanLuo: Yes, batcher start time should be overwritten for the use of current batcher, as the previous batcher has finished its scheduling of the request by this point. Added comment

deadeyegoodwin (approved):

CoderHam (approved):

PR #3848, opened 2022-01-25
Block the DirectSequenceBatcher thread to form larger batches

tanmayv25:
This change improves performance of DirectSequenceBatcher and brings it at par with r21.08.

### Benchmarking parameters:
perf_analyzer -u localhost:8001 --streaming -i gRPC -m citrinet-1024-en-US-asr-streaming-throughput -b 1 --input-data ensemble_chunk800/ --shape=AUDIO_SIGNAL:12800 --shape=CUSTOM_CONFIGURATION:1,2 --sequence-length=4  --concurrency-range=128:128:1 -p 1000

### Model Configuration:
`citrinet-1024-en-US-asr-streaming-throughput` is an ensemble model with 5 composing sequence models. Only 1 of these five  uses DirectSequenceBatcher, others use Oldest scheme.

### With Fix

```
Concurrency: 128, throughput: 983 infer/sec, latency 136770 usec
```

### Without fix

```
Concurrency: 128, throughput: 644 infer/sec, latency 199690 usec
```

### 21.08

```
*** Measurement Settings ***
Concurrency: 128, throughput: 846 infer/sec, latency 147019 usec
```

### DIFF
--- a/src/core/sequence_batch_scheduler.cc
+++ b/src/core/sequence_batch_scheduler.cc
@@ -1202,7 +1202,7 @@ DirectSequenceBatch::BatcherThread(const int nice)
   while (!scheduler_thread_exit_) {
     uint64_t wait_microseconds = default_wait_microseconds;
 
-    // Wait till execution of the last enqueued payload is not
+    // Wait till execution of the last enqueued payload is
     // complete.
     {
       std::unique_lock<std::mutex> lk(payload_mu_);
--- a/src/core/sequence_batch_scheduler.cc
+++ b/src/core/sequence_batch_scheduler.cc
@@ -1193,6 +1193,7 @@ DirectSequenceBatch::BatcherThread(const int nice)
   const uint64_t default_wait_microseconds = 500 * 1000;
 
   NewPayload();
+  exec_complete_ = true;
 
   // When there is optional input or input shape must be enforced,
   // the inputs in the requests must be examined for forming a batch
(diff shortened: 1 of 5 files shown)


src/core/sequence_batch_scheduler.cc
```diff
@@ -1201,6 +1202,13 @@ DirectSequenceBatch::BatcherThread(const int nice)
   while (!scheduler_thread_exit_) {
     uint64_t wait_microseconds = default_wait_microseconds;
 
+    // Wait till execution of the last enqueued payload is not
```
deadeyegoodwin, line 1205: This comment is confusing. "wait till is complete"?

deadeyegoodwin (changes requested):

GuanLuo (commented):
Looks good except needing to clarify the comment

GuanLuo (approved):

deadeyegoodwin (dismissed):

src/core/sequence_batch_scheduler.cc
```diff
@@ -1201,6 +1202,13 @@ DirectSequenceBatch::BatcherThread(const int nice)
   while (!scheduler_thread_exit_) {
     uint64_t wait_microseconds = default_wait_microseconds;
 
+    // Wait till execution of the last enqueued payload is
+    // complete.
+    {
+      std::unique_lock<std::mutex> lk(payload_mu_);
+      payload_cv_.wait(lk, [this] { return exec_complete_; });
+    }
```
Tabrizian, line 1210: I was just curious why waiting for completion of the previous enqueued payload leads to higher performance?

CoderHam: As I understand this allows us to delay the enqueue by the maximum amount of time. This additional time allows formation of larger batches that in turn yield a greater throughput. 
@tanmayv25 feel free to correct my understanding

CoderHam (approved):

src/core/sequence_batch_scheduler.cc
tanmayv25: Hemant is correct. If we give more time for the requests to accumulate, we can form larger batches. We wait till the Instance thread is actually done processing the payload and then only try to form the batch.
So, instead of too many small sized requests, we will have few larger batch sized requests.

PR #3849, opened 2022-01-26
Add trace endpoints. Add trace log frequency

GuanLuo:
@deadeyegoodwin look-and-feel of the trace endpoints and the new trace configuration requested

### DIFF
--- a/docs/protocol/extension_trace.md
+++ b/docs/protocol/extension_trace.md
@@ -83,7 +83,9 @@ trace multiple informations.
 will one trace be sampled from. For example, if the trace rate is "1000",
 1 trace will be sampled for every 1000 requests.
 - "trace_count" : the number of remaining traces to be sampled. Once the value
-becomes "0", no more traces will be sampled for the trace setting.
+becomes "0", no more traces will be sampled for the trace setting, and the
+collected traces will be written to indexed trace file in the format described
+in "log_frequency", regardless of the "log_frequencey" status.
 If the value is "-1", the number of traces to be sampled will not be limited.
 - "log_frequency" : the frequency that Triton will log the
 trace output to the files. If the value is "0", Triton will only log
--- a/docs/protocol/extension_trace.md
+++ b/docs/protocol/extension_trace.md
@@ -82,9 +82,9 @@ trace multiple informations.
 - "trace_rate" : the trace sampling rate. The value represents how many requests
 will one trace be sampled from. For example, if the trace rate is "1000",
 1 trace will be sampled for every 1000 requests.
-- "trace_count" : the number of traces to be sampled. Once the specified number
-of traces are sampled, 'trace_level' will be set to "OFF" to disable tracing.
-If the value is "0", the number of traces to be sampled will not be limited.
+- "trace_count" : the number of remaining traces to be sampled. Once the value
+becomes "0", no more traces will be sampled for the trace setting.
+If the value is "-1", the number of traces to be sampled will not be limited.
 - "log_frequency" : the frequency that Triton will log the
 trace output to the files. If the value is "0", Triton will only log
 the trace output to ${trace_file} when shutting down. Otherwise, Triton will log
--- a/docs/protocol/extension_trace.md
+++ b/docs/protocol/extension_trace.md
@@ -167,6 +167,11 @@ message TraceRequest
   // The name of the model to apply the new trace settings.
   // If not given, the new settings will be applied globally.
   string model_name = 2;
+
+  // Clear the current model specific setting if specified, and
+  // any new setting values will not be applied.
+  // This option will be ignored if 'model_name' is not given.
+  bool clear_setting = 3;
 }
 
 message TraceResponse
--- a/docs/protocol/extension_trace.md
+++ b/docs/protocol/extension_trace.md
@@ -44,12 +44,14 @@ HTTP GET request to retrieve the current trace setting. A HTTP POST request
 will modify the trace setting, and the endpoint will return the updated trace
 setting on success or an error in the case of failure. Optional model name and
 model version can be provided to get or to set the trace settings for specific
-model or model version.
+model or model version. Note that when a model or model version is specified
+without providing any trace settings in the request body, it will be a request
+to remove the specification. 
 
 ```
-GET v2/models[/${MODEL_NAME}[/versions/${MODEL_VERSION}]]/trace
+GET v2[/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]]/trace
 
-POST v2/models[/${MODEL_NAME}[/versions/${MODEL_VERSION}]]/trace
+POST v2[/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]]/trace
 ```
 
 ### Trace Response JSON Object
@@ -114,8 +116,7 @@ A trace configure request is made with a HTTP POST to
 the trace endpoint. In the corresponding response the HTTP body contains the
 response JSON. A successful request is indicated by a 200 HTTP status code.
 
-The request object, identified as
-$trace_request must be provided in the HTTP
+The request object, identified as $trace_request must be provided in the HTTP
 body.
 
 ```
@@ -127,7 +128,10 @@ $trace_request =
 
 The $trace_setting JSON is defined in
 [Trace Response JSON Object](#Trace-Response-JSON-bject), only the specified
-settings will be updated.
+settings will be updated. Note that if this is the first request to initalize
+a model (version) trace settings, for the trace settings that are not specified
+in the request, the value will be copied from the current setting in the
+following order: version settings, model settings, global settings.
 
 ## GRPC
 
@@ -158,11 +162,12 @@ message TraceRequest
 
   // The name of the model to apply the new trace settings.
   // If not given, the new settings will be applied globally.
-  string name = 1;
+  string model_name = 1;
 
   // The version of the model to apply the new trace settings.
   // If not given, the new settings will be applied to all version of the model.
-  string version = 2;
+  // 'model_version' is ignored if 'model_name' is not specified.
+  string model_version = 2;
 }
 
 message TraceResponse
@@ -174,3 +179,7 @@ message TraceResponse
 
 The trace settings are mentioned in
 [Trace Response JSON Object](#Trace-Response-JSON-bject).
+Note that if this is the first request to initalize
+a model (version) trace settings, for the trace settings that are not specified
+in the request, the value will be copied from the current setting in the
+following order: version settings, model settings, global settings.
--- a/docs/protocol/extension_trace.md
+++ b/docs/protocol/extension_trace.md
@@ -42,16 +42,15 @@ indicates an optional JSON field.
 Triton exposes the trace endpoint at the following URL. The client may use
 HTTP GET request to retrieve the current trace setting. A HTTP POST request
 will modify the trace setting, and the endpoint will return the updated trace
-setting on success or an error in the case of failure. Optional model name and
-model version can be provided to get or to set the trace settings for specific
-model or model version. Note that when a model or model version is specified
-without providing any trace settings in the request body, it will be a request
-to remove the specification. 
+setting on success or an error in the case of failure. Optional model name
+can be provided to get or to set the trace settings for specific model.
+Note that when a model is specified without providing any trace settings
+in the request body, it will be a request to remove the specification. 
 
 ```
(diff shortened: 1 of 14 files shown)


deadeyegoodwin (changes requested):

docs/protocol/extension_trace.md
```diff
@@ -0,0 +1,164 @@
...
+$object and $array refer to the fundamental JSON types. #optional
+indicates an optional JSON field.
+
+Triton exposes the trace endpoint at the following URL. The client may use
+HTTP GET request to retrieve the current trace setting. A HTTP POST request
+will modify the trace setting, and the endpoint will return the updated trace
+setting on success or an error in the case of failure. 
+
+```
+GET v2/trace
```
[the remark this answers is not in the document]
GuanLuo: So all trace settings can be configured at model (/version) level? i.e. it can have say 100 sample rate globally and for model A, we can set 1 sample rate so all requests to model A will be traced?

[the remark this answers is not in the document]
deadeyegoodwin: Yes, you should use the commandline values as the initial global defaults that apply to all models/versions. The /v2/trace API can be used to get/set those global defaults. Per-model/version overrides can be set with v2/trace/... API. No need to have per-model controls from command line. So when you are getting the settings for a particular model the order of precedence is model/version settings, model settings, global settings.

docs/protocol/extension_trace.md
```diff
@@ -0,0 +1,176 @@
...
+
+Triton exposes the trace endpoint at the following URL. The client may use
+HTTP GET request to retrieve the current trace setting. A HTTP POST request
+will modify the trace setting, and the endpoint will return the updated trace
+setting on success or an error in the case of failure. Optional model name and
+model version can be provided to get or to set the trace settings for specific
+model or model version.
+
+```
+GET v2/models[/${MODEL_NAME}[/versions/${MODEL_VERSION}]]/trace
```
deadeyegoodwin, line 50: v2[/models/...

docs/protocol/extension_trace.md
```diff
@@ -0,0 +1,176 @@
...
+```
+message TraceRequest
+{
+  // The new setting values to be updated,
+  // settings that are not specified will remain unchanged.
+  map<string, string> settings = 1;
+
+  // The name of the model to apply the new trace settings.
+  // If not given, the new settings will be applied globally.
+  string name = 1;
```
deadeyegoodwin, line 161: model_name

docs/protocol/extension_trace.md
```diff
@@ -0,0 +1,176 @@
...
+
+  // The version of the model to apply the new trace settings.
+  // If not given, the new settings will be applied to all version of the model.
+  string version = 2;
```
deadeyegoodwin, line 165: model_version
"model_version is ignored if 'model_name' is not specified"

deadeyegoodwin (changes requested):

docs/protocol/extension_trace.md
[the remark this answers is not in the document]
GuanLuo: Do we also want a "clear" API on model and model version to remove the specific settings so user can reset it to use global settings?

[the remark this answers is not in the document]
deadeyegoodwin: Yes, that would be good. Not sure if we should just have .../trace/clear endpoint for that or something else.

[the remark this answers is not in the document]
GuanLuo: We may add something like when using the POST to update the specific model or model version without providing any parameters, then it will clear the previous specification, although it sounds less clear than having a clear endpoint while I am typing it...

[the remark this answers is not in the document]
deadeyegoodwin: I like it better than adding a .../clear endpoint.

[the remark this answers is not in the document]
GuanLuo: Pushed a commit to document the clear behavior, and more clarification on specifying model (version) setting without providing all trace setting values

GuanLuo:
The trace logic has been reworked to the following:

- When a trace should be sampled, a `Trace` object is returned and it contains the Triton trace pointer (which will be transferred to Triton on calling `InferAsync()` as Triton API documented). The `Trace` object is owned by the server frontends and the frontends must ensure its lifecycle is longer than Triton trace's because the Triton trace will log Triton timestamps through the `Trace` object, which is doable as the frontends have been maintaining meta data associated with the requests (request's lifecycle includes Triton trace's lifecycle).
  - The frontends will log its own timestamps associated with the request via returned `Trace` object
- The frontends will only interact with `TraceManger` for trace APIs, `TraceSetting` and `TraceFile` are abstraction to help `TraceManger` to keep track of the different model settings and versioned logging functionality.

docs/protocol/extension_trace.md
```diff
@@ -0,0 +1,193 @@
...
+```
+$trace_request =
+{
+  $trace_setting, ...
+}
+```
+
+The $trace_setting JSON is defined in
+[Trace Response JSON Object](#Trace-Response-JSON-bject), only the specified
+settings will be updated. Note that if this is the first request to initalize
```
deadeyegoodwin, line 131: Need to say something about is a setting is listed with a 'null' value then it is cleared. By 'null' I mean the json defined null

src/servers/http_server.h
```diff
@@ -204,6 +203,9 @@ class HTTPAPIServer : public HTTPServer {
     // lifetime of the request.
     std::list<std::vector<char>> serialized_data_;
 
+    std::atomic<uint32_t> complete_count_;
```
deadeyegoodwin, line 206: Why does the request lifetime management get more complicated?

docs/protocol/extension_trace.md
```diff
@@ -0,0 +1,193 @@
...
+
+- "trace_file" : the file where the trace output will be saved. If
+"log frequency" is set, this will be the prefix of the files to save the
+trace output, resulting files in name "${trace_file}.0", "${trace_file}.1"...,
+see trace setting "log frequency" below for detail.
+- "trace_level" : the trace level. "OFF" to disable tracing,
+TIMESTAMPS" to trace timestamps, "TENSORS" to trace tensors.
+This value is an array of string whhere user may specify multiple levels to
+trace multiple informations.
+- "trace_rate" : the trace sampling rate. The value represents how many requests
```
deadeyegoodwin, line 84: Need to add "trace_count" here and also in the cmdline. trace_count defaults to 0 which means to continue collecting trace indefinitely. But it can be set to non-zero to cause trace_level to go to OFF after collecting trace_count traces.

deadeyegoodwin (changes requested):

docs/protocol/extension_trace.md
GuanLuo: So for the "clear" mentioned before, it actually only clears a set of fields specified? Or we want to be able to clear the whole model setting as well as a subset of fields inside it? And when we "clear" a field, it actually means use the field value in global setting

deadeyegoodwin: Ideally, a setting that has never been explicitly set for a model will just pass through to the global settings. So
- clearing a global setting mean that the global default will now be used for that setting
- clearing a model setting means that the global value will now be used for that setting
Depending on how you implement, this means that when a model setting is cleared, you just set its value to whatever the current global setting is. And when a global setting is changed you go through all the models and the ones that don't have a model-specific value for that setting you set it to the new global setting.
A common use case could be that I set a global trace rate that I want for all models. Then for one specific model I temporarily set the trace rate to be higher because I want more trace data. When I no longer want the high trace rate for that model I want to be able to just clear the trace rate for that model and have it automatically pick up the current global trace rate.

src/servers/http_server.h
GuanLuo: The HTTP request lifecycle wasn't being managed correctly, before it assumes the requests can be released when response complete callback is called, which is not the case because the backend can release the request after response is sent, so I think the lifecycle should be done when both the response and request are done.

In this particular case, Triton core will call "response complete -> request complete (log timestamp to `Trace`)", so with the previous HTTP request lifecycle, the trace will operate on released `Trace` object.

It was working before because the frontend and the Triton trace report timestamps to `TraceManager` directly and there is no dependency between the Triton trace and HTTP request, whereas now they both report to `Trace` which is owned by the HTTP request.

deadeyegoodwin: In the old lifecycle was the HTTP request needed to live until the response completed? Or is that a new requirement due to the new way of handling Trace? I'm worried about introducing a memory leak... the existing HTTP endpoint is quite convoluted and makes these kind of changes risky.

docs/protocol/extension_trace.md
GuanLuo: I am going to use "field" to represent each of the configurables ("trace_rate", "trace_file", etc.) and "setting" for all the configurables associated with a model or global. So conceptually there are 3 levels of setting:
- global default (cmdline, all fields considered "specified")
- global setting (global update via API)
- model setting (model update via API)

For any level, the unspecified field will use value from the previous level, and the change of the field in the previous level will propagate down.

When a field is listed in the request, it will either specify the field (with the value) or un-specify the field ("null" in JSON / empty value list in GPRC protobuf).

We don't provide additional way to un-specify all fields in a setting, this can drop the `clear_setting` in GRPC protobuf. If the user want to un-specify all fields, the user will send the following JSON request:
```
{
"trace_rate": null,
"log_frequency": null,
"trace_file": null,
"trace_level": null
}
```

deadeyegoodwin: That sounds good.

src/servers/http_server.h
GuanLuo: > In the old lifecycle was the HTTP request needed to live until the response completed?

Yes, the old way requires HTTP request lives until the response completed. The new way extends the requirement to that HTTP request lives until both response and request are completed.

I can make the `TRITONSERVER_InferenceTrace` to carry the pointer to a `shared_ptr<Trace>` and delete it when Triton trace finish its part. Then HTTP request lifecycle won't need to be changed, even if it is released before Triton trace, the `Trace` object will still be valid.

deadeyegoodwin: So the old way is ok as long as we don't actually need the HTTP request object on request completion. Is that the case?
If so, I would prefer to use shared_ptr<Trace> idea and not change the HTTP request object lifecycle handling.

docs/protocol/extension_trace.md
GuanLuo: If `trace_level` is updated to OFF when `trace_count` exceeds, I assume the following will be true?
1. Turning OFF global `trace_level` by `trace_count` does not cause any model specific setting to turn off its `trace_level` even if it wasn't specified for the model. In other words, the `trace_level` in model setting will be marked as "specified" once the global `trace_level` is affected by `trace_count`
2. The user can turn `trace_level` back on with the update API. For global setting, due to the behavior in 1., this change will not be reflected to model setting unless the user "clear" the model `trace_level` first.

deadeyegoodwin: Would it be easier to explain and implement if we don't actually change trace_level when trace_count is done? In the implementation we would trace only if trace_level was not OFF and trace_count > 0. We can use trace_count = -1 to mean unlimited trace_count.

Also be easier for user to just reset trace_count if they want to restart or change it (no need to also reset trace_level). Also, we can be consistent withe trace_count, when it is set in global and has not been specifically set in model then we can just set it in the model as well.

deadeyegoodwin (changes requested):

docs/protocol/extension_trace.md
GuanLuo: So we use `trace_count` to directly indicate the remaining trace to be sampled for the particular setting, and a -1 value is used for no trace limit? Yes, that will be more clear to know why exactly the trace stops.

deadeyegoodwin: Yes, and we don't adjust trace_level ourselves.

src/servers/main.cc
```diff
@@ -476,8 +476,8 @@ std::vector<Option> options_
       {OPTION_TRACE_COUNT, "trace-count", Option::ArgInt,
        "Set the number of traces to be sampled. Once the specified number "
```
deadeyegoodwin, line 477: Delete "Once the specified... to disable tracing"

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #3850, opened 2022-01-26
Fix windows build to avoid tritonserver.lib collision

deadeyegoodwin:


src/servers/CMakeLists.txt
```diff
@@ -277,10 +286,23 @@ target_link_libraries(
...
+if (NOT WIN32)
 install(
   TARGETS main
   RUNTIME DESTINATION bin
 )
+else()
+  # See explanation above as to why we need to rename main.exe to
+  # tritonserver.exe as part of the install process on windows.
+  install(
+    PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_BUILD_TYPE}/main.exe
```
GuanLuo, line 299: So when we generate the `.exe` on Windows, it also generates a `.lib` with the same name? And here because we already have `tritonserver.lib` generated in previous step, we use a different name (main) to ensure the side-effect goes through (`main.lib` is generated but we don't need it)?

GuanLuo (approved):
LGTM. Just want to know more about the detail.

CoderHam (approved):

src/servers/CMakeLists.txt
deadeyegoodwin: Yes. I think the *.lib is only being generated for tritonserver.exe because we are accidentally exporting a symbol. It still doesn't make sense why you want a *.lib for an executable, but that is what windows does. I think it has something to do with src/servers using a source file from src/core that includes tritonserver_api.h, which in turn causes C APIs to be exported instead of imported.

PR #3853, opened 2022-01-26
Enable variable dimension tests for TRT backend in sequence batcher

Tabrizian:
Enable the variable dimension tests both for the L0_sequence_batcher and L0_implicit_state.

### DIFF
--- a/qa/common/gen_qa_implicit_models.py
+++ b/qa/common/gen_qa_implicit_models.py
@@ -573,7 +573,6 @@ def create_plan_fixed_modelfile(models_dir, model_version, max_batch, dtype,
     not_start = network.add_elementwise(constant_1.get_output(0), start0,
                                         trt.ElementWiseOperation.SUB)
     not_start.set_output_type(0, trt_dtype)
-
     internal_state = network.add_elementwise(in_state0, not_start.get_output(0),
                                              trt.ElementWiseOperation.PROD)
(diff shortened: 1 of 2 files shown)


qa/common/gen_qa_implicit_models.py
```diff
@@ -690,17 +689,236 @@ def create_plan_fixed_rf_modelfile(models_dir, model_version, max_batch, dtype,
...
-                                    shape)
+        if (not tu.shape_is_fixed(shape)):
+            create_plan_dynamic_modelfile(models_dir, model_version, max_batch,
+                                          dtype, shape)
+        else:
+            create_plan_fixed_modelfile(models_dir, model_version, max_batch,
+                                        dtype, shape)
 
 
 def create_plan_modelconfig(models_dir, model_version, max_batch, dtype, shape):
```
GuanLuo, line 924: Do you need to generate the config for the new dynamic shape models?

Tabrizian: I generated them in the DLcluster.

tanmayv25: I think Guan meant do you need to write a different function for creating plan_modelconfig file for variable dimension case. 
I think the same function should work as the information regarding dynamic shape is communicated via shape parameter.

tanmayv25 (approved):

PR #3862, opened 2022-01-27
Add SSL testing on C++ client library. Enhance L0_https.

tanmayv25:


qa/L0_https/test.sh
```diff
@@ -89,12 +98,67 @@ service nginx restart
...
+$TEST_CLIENT -v -u https://localhost:443 --verifyhost 0 --verifypeer 0 >> ${CLIENT_LOG}.c++.ssl_infer_insecure 2>&1
+if [ $? -ne 0 ]; then
+    cat ${CLIENT_LOG}.c++.ssl_infer_insecure
+    RET=1
+fi
+
+# Test failure cases for SSL
+# Try without SSL 
+$SIMPLE_INFER_CLIENT_PY -v -u localhost >> ${CLIENT_LOG}.no_ssl_fail_infer 2>&1
+if [ $? -ne 0 ]; then
```
matthewkotila, line 129: Kind of a small detail, but if I'm understanding this correctly, this test expects the exit code of line 128 to be 0, correct? And if it's not 0, then the test fails. But the comment and echo suggest that line 128 is supposed to fail--i.e. a non-zero exit code?

CoderHam: Check condition checks if there is a failure, comments that the same was expected and if there is no failure sets RET to 1. Looks correct.

CoderHam (dismissed):

qa/L0_https/test.sh
matthewkotila: oh, the `if` is the test success case, not the test failure case. the `else` is the test failure case. 😵‍💫

CoderHam (approved):

PR #3872, opened 2022-01-31
Update NGC versions post-22.01 release

dzier:


dzier:
@tanmayv25  and @GuanLuo  can you review your cherry picked commits above to make sure they were merged correctly?

src/core/payload.cc
```diff
@@ -142,6 +142,12 @@ Payload::MarkSaturated()
   saturated_ = true;
 }
 
+void
```
tanmayv25, line 145: MarkSaturated got duplicated. We can remove this one.

dzier: fixed

tanmayv25 (approved):

PR #3873, opened 2022-01-31
Fix mismatch between the override correlation ID and request correlation ID

Tabrizian:
The root cause was that we were reusing the same override input for all the sequence slots. If the sequence ended but the request was not executed yet, the slot will belong to a new sequence and override the correlation id of the previous sequence which has not been executed which lead to mismatch between the request correlation ID and input override correlation ID.


GuanLuo (approved):
The change looks good to me, basically we need dedicated override tensor for the request to avoid overwrite in the case you describe, should the same be applied to other override tensors?

GuanLuo:
Right.. if I remember it correctly, we have the true/false tensor and simplify assign the proper one to the request, the value of the tensor is not changed

PR #3886, opened 2022-02-03
Fix generation script for variable size TRT models

Tabrizian:


rmccorm4 (approved):
Thanks for catching the `__pycache__`

PR #3888, opened 2022-02-04
Implement testing for running PA with SSL options

matthewkotila:


qa/L0_perf_analyzer/test.sh
```diff
@@ -718,6 +718,113 @@ set -e
...
+    --ssl-grpc-private-key-file=client.key \
+    --ssl-grpc-certificate-chain-file=client2.crt \
+    > ${CLIENT_LOG}.grpc_failure 2>&1
+if [ $? -eq 0 ]; then
+    cat ${CLIENT_LOG}.grpc_failure
+    echo -e "\n***\n*** Expected test failure\n***"
+    RET=1
+fi
+
+# Test that HTTP protocol with SSL works correctly
```
tanmayv25, line 809: Add one more test case.. Don't provide the certs to PA. Just provide: 
```
$PERF_ANALYZER -v -u https://localhost:443 -i http -m graphdef_int16_int16_int16 --percentile=95 \
    --ssl-https-verify-peer 0 --ssl-https-verify-host 0
```
Most of the users will be interested in the same.

qa/L0_perf_analyzer/test.sh
```diff
@@ -718,6 +718,130 @@ set -e
...
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+# Test gRPC SSL
+set +e
+
+# Test that gRPC protocol with SSL works correctly
+$PERF_ANALYZER -v -i grpc -m graphdef_int16_int16_int16 --percentile=95 \
```
tanmayv25, line 758: Remove `--percentile` from here and below.

tanmayv25 (changes requested):

qa/L0_perf_analyzer/test.sh
matthewkotila: added that test case 👍

qa/L0_perf_analyzer/test.sh
matthewkotila: removed 👍

tanmayv25 (approved):

PR #3889, opened 2022-02-04
Add new option to run stress and memory growth tests longer

GuanLuo:
must revert the change in L0_long_running_stress added for CI experiment


CoderHam (approved):

qa/L0_long_running_stress/stress.py
```diff
@@ -392,6 +441,12 @@ def generate_report(elapsed_time, _test_case_count, _failed_test_case_count,
                         required=False,
                         default=8,
                         help='Request concurrency. Default is 8.')
+    parser.add_argument('--load-thread',
```
tanmayv25, line 444: What is the purpose of having load_thread? What would a load thread count of 0 entail? Is it to keep perf_analyzer's running in the background when stress_thread is running? If yes, then we should clarify in the help message.

tanmayv25 (changes requested):

tanmayv25 (dismissed):

CoderHam (dismissed):

tanmayv25 (approved):

PR #3892, opened 2022-02-04
Add Java Simple Tests - CPU-Only

dyastremsky:
This PR adds CPU-only tests to L0_java_simple.


dyastremsky:
@jbkyang-nvi Could you review this PR?

It only modifies the test script, not the interface or sample files, so it should be separate from getting SimpleCPUOnly.java to be CPU-only. It only tests that the CPU and GPU capabilities work as expected.

jbkyang-nvi (approved):
looks good!

PR #3893, opened 2022-02-04
Improve the L0_io to test for peer access

Tabrizian:
@deadeyegoodwin Only disabling the peer access was not useful for the case that we were trying to test. I changed the L0_io such that it will run L0_io on two GPUs that cannot enable peer access. If we only disabled peer access, it will be automatically enabled by Triton and Python backend.


dyastremsky:
Hey Iman. Should this PR be closed?

rmccorm4:
Does this have any action like merge/close after Indrajit's recent "enable peer access" flag?

PR #3900, opened 2022-02-07
Fix L0_model_config pytorch failures

CoderHam:


qa/L0_model_config/autofill_noplatform/pytorch/too_few_outputs/config.pbtxt
```diff
@@ -4,6 +4,11 @@ input [
     name: "INPUT__0"
     data_type: TYPE_INT32
     dims: [ 16 ]
+  },
+  {
+    name: "INPUT__1"
+    data_type: TYPE_INT32
+    dims: [ 16 ]
```
GuanLuo, line 11: This is not right.. Look at what the model is.. This test is checking the backend validation will return error when model has outputs but the model config doesn't specify any

GuanLuo (approved):

qa/L0_model_config/autofill_noplatform/pytorch/too_few_outputs/config.pbtxt
CoderHam: The test will report an error when there are no outputs. For pytorch, in ValidateOutputs, we can only report an error when there are no outputs. We cannot know the number of outputs till the model has been run.

PR #3933, opened 2022-02-14
Add Java Sequence Batcher Tests

dyastremsky:
The tests cover using the sequence batcher with the Java Bindings. The test covers the ONNX, TensorFlow, and PyTorch backends.

### DIFF
--- a/qa/L0_java_sequence_batcher/SequenceTest.java
+++ b/qa/L0_java_sequence_batcher/SequenceTest.java
@@ -561,10 +561,13 @@ public class SequenceTest {
           TRITONSERVER_InferenceRequestAddRequestedOutput(irequest, output0),
           "requesting output 0 for the request");
 
+      // Non-zero ID for the sequence requests
       long correlation_id = 5;
-      int iterations = 9;
-      int expected_result = 45;
-      boolean sequence_start = true, sequence_end = false;
+      int num_requests = 9;
+      // The expected_result is  1+2+3+...+num_requests
+      int expected_result = num_requests * (1 + num_requests) / 2;
+      boolean sequence_start = true;
+      boolean sequence_end = false;
 
       // Create the initial data for the input tensor.
       IntPointer[] p0 = {new IntPointer(1)};
@@ -577,7 +580,7 @@ public class SequenceTest {
                 0 /* memory_type_id */),
             "assigning INPUT0 data");
 
-      for(int i = 0; i < iterations; i++) {
+      for(int i = 0; i < num_requests; i++) {
         // Update input value
         int input = i + 1;
         p0[0].put(0, input);
@@ -586,7 +589,7 @@ public class SequenceTest {
         if(i == 1) {
           sequence_start = false;
         }
-        if(i == iterations - 1) {
+        if(i == num_requests - 1) {
           sequence_end = true;
         }
         SetSequenceMetadata(irequest, correlation_id, sequence_start, sequence_end);
--- a/qa/L0_java_sequence_batcher/SequenceTest.java
+++ b/qa/L0_java_sequence_batcher/SequenceTest.java
@@ -563,8 +563,9 @@ public class SequenceTest {
 
       // Non-zero ID for the sequence requests
       long correlation_id = 5;
+      // Number of requests in the sequence
       int num_requests = 9;
-      // The expected_result is  1+2+3+...+num_requests
+      // Expected_result is  1+2+3+...+num_requests
       int expected_result = num_requests * (1 + num_requests) / 2;
       boolean sequence_start = true;
       boolean sequence_end = false;
--- a/qa/L0_java_sequence_batcher/SequenceTest.java
+++ b/qa/L0_java_sequence_batcher/SequenceTest.java
@@ -32,7 +32,7 @@ import org.bytedeco.javacpp.*;
 import org.bytedeco.tritonserver.tritonserver.*;
 import static org.bytedeco.tritonserver.global.tritonserver.*;
 
-public class Simple {
+public class SequenceTest {
 
     // Boilerplate code for setting up Triton
     static void FAIL(String MSG) {
@@ -67,7 +67,7 @@ public class Simple {
         System.err.println(msg);
       }
 
-      System.err.println("Usage: java " + Simple.class.getSimpleName() + " [options]");
+      System.err.println("Usage: java " + SequenceTest.class.getSimpleName() + " [options]");
       System.err.println("\t-m [model name]");
       System.err.println("\t-v Enable verbose logging");
       System.err.println("\t-r [model repository absolute path]");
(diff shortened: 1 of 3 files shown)


qa/L0_java_sequence_batcher/Simple.java
```diff
@@ -0,0 +1,676 @@
...
+
+    static void
+    SetSequenceMetadata(TRITONSERVER_InferenceRequest irequest, long correlation_id, boolean sequence_start, boolean sequence_end)
+    throws Exception
+    {
+
+      FAIL_IF_ERR(
+          TRITONSERVER_InferenceRequestSetCorrelationId(
+              irequest, correlation_id), "Unable to set correlation ID");
+      if(sequence_start) {
```
Tabrizian, line 212: I think we shouldn't use `else if` for `sequence_end` because a sequence can be both starting and ending at the same time and we may need to set both of the flags.

dyastremsky: Thanks for reviewing, great catch. I updated the code to allow for a request to be both start and end.

qa/L0_java_sequence_batcher/Simple.java
```diff
@@ -0,0 +1,671 @@
+// Copyright (c) 2021-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
jbkyang-nvi, line 1: If we're not making any changes to Simple.java, can we move this into `common`?

qa/L0_java_sequence_batcher/Simple.java
```diff
@@ -0,0 +1,671 @@
...
+//
```
jbkyang-nvi, line 2: This is a change of this from the original Simple.java code right? Can we change the name to something like simple_sequence.java or similar? Also change copyright

jbkyang-nvi: But also, maybe it's better to just have one Simple.java that calls a separate function unique for that test.

dyastremsky: That's correct.

We could probably wrap a lot of the functionality in Simple, and have each test call a child class of Simple specific to it with those functions overridden. Would that solution be ideal? Are you doing L0_java_infer differently; if not, let me know what customizations you would need for that.

dyastremsky: @jbkyang-nvi As discussed, I renamed Simple and added comments to describe what is boilerplate vs custom code.

Once we have L0_java_sequence_batcher, L0_java_infer, and L0_java_memory_growth merged, I could look into turning them into one test framework class file to avoid repeating boilerplate code, if we think it makes sense.

qa/L0_java_sequence_batcher/SequenceTest.java
```diff
@@ -0,0 +1,636 @@
...
+      String output0 = is_torch_model[0] ? "OUTPUT__0" : "OUTPUT";
+
+      FAIL_IF_ERR(
+          TRITONSERVER_InferenceRequestAddRequestedOutput(irequest, output0),
+          "requesting output 0 for the request");
+
+      long correlation_id = 5;
+      int iterations = 9;
+      int expected_result = 45;
+      boolean sequence_start = true, sequence_end = false;
```
jbkyang-nvi, line 567: nit: can we not use the comma here? Declare two booleans instead

qa/L0_java_sequence_batcher/SequenceTest.java
```diff
@@ -0,0 +1,636 @@
...
+
+      String output0 = is_torch_model[0] ? "OUTPUT__0" : "OUTPUT";
+
+      FAIL_IF_ERR(
+          TRITONSERVER_InferenceRequestAddRequestedOutput(irequest, output0),
+          "requesting output 0 for the request");
+
+      long correlation_id = 5;
+      int iterations = 9;
+      int expected_result = 45;
```
jbkyang-nvi, line 566: where do these values come from? Can they be input params?

jbkyang-nvi (commented):
looks good so far! Two comments

qa/L0_java_sequence_batcher/SequenceTest.java
dyastremsky: Done!

qa/L0_java_sequence_batcher/SequenceTest.java
dyastremsky: Correlation_id doesn't really matter, it just needs to be a non-zero number so that the sequence of requests has an ID tying them together.

As for iterations and expected result, I made them match sequence_batcher_test.py in terms of setting a value of 9 and 45. I can switch them to input parameters if that's better... it doesn't make a big difference, as these models essentially sum 1 to n. In this case, 1+2+...+8+9 = 45. 

I renamed these variables and added comments for clarity. Let me know if that works. If you still want me to switch any of these to input params, let me know. I don't think it'd affect what we're testing in the sequence batcher, but it's an easy change.

qa/L0_java_sequence_batcher/SequenceTest.java
```diff
@@ -0,0 +1,640 @@
...
+      is_torch_model[0] =
+          model_metadata.get("platform").getAsString().equals("pytorch_libtorch");
+      return null;
+    }
+
+    // Custom function to set metadata required for
+    // sequence batcher
+    static void
+    SetSequenceMetadata(TRITONSERVER_InferenceRequest irequest, long correlation_id, boolean sequence_start, boolean sequence_end)
+    throws Exception
```
jbkyang-nvi, line 209: do we need throws Exception here?

dyastremsky: Good catch (pun intended). Must have gotten mixed in during a previous commit... thanks, fixed!

qa/L0_java_sequence_batcher/SequenceTest.java
```diff
@@ -0,0 +1,638 @@
...
+          System.out.println(model_name + " test PASSED");
+        }
+      }
+    }
+
+    // Boilerplate main function to run inference
+    // for provided model, custom setting of
+    // sequence metadata
+    public static void
+    main(String[] args) throws Exception
```
jbkyang-nvi, line 340: here too

jbkyang-nvi (approved):
nit on the exception throwing otherwise looks good

qa/L0_java_sequence_batcher/SequenceTest.java
dyastremsky: This one does need to be there. It's from the original Simple. We could get rid of it by try-catching the two Thread.sleeps (during server start-up and model readiness) and the CompletableFuture.get() instead. If we do that, it'd be good to do it to our original sample and across all the tests for uniformity (perhaps in the big java-api-prerelease -> main PR).

dyastremsky: If it's okay, I'll merge for now. If you'd like me to address this across our Simple examples and tests, let me know and I can do so in the big PR ticket or can do so sooner in a separate ticket.

PR #3934, opened 2022-02-15
Grow payload even if preferred batch size is not found. Fix bug in Payload class

GuanLuo:


tanmayv25 (dismissed):

tanmayv25 (dismissed):

deadeyegoodwin (commented):
So does this fix an issue that only happens when the model specifies a preferred_batch_size < max_batch_size? Or can it happen to any model?

GuanLuo:
It can happen to any model. This is useful in the edge case where the request consumption / arrival rate makes preferred batch size will never be formed. In the particular case that I looked into, the queue delay is smaller than model execution: when the model is running, the queue delay will expire and enqueue next payload (non-preferred batch). Before this change the dynamic batcher will only add new requests to the payload if it forms preferred batch, which can results in pending requests in scheduler queue when the model is going to execute the next payload.

tanmayv25 (approved):

PR #3945, opened 2022-02-16
Build additional openvino backend with specific commit id

tanmayv25:


build.py
```diff
@@ -87,7 +87,9 @@
         '22.01',  # upstream container
         '1.10.0',  # ORT
         '2021.2.200',  # ORT OpenVINO
-        (('2021.2', None), ('2021.4', '2021.4.582')),  # Standalone OpenVINO
+        (('2021.2', None),
+         ('2021.4', '2021.4.582'),
+         ('SPECIFIC','f2f281e6')),  # Standalone OpenVINO
```
deadeyegoodwin, line 92: Add comment above describing use of "SPECIFIC"

deadeyegoodwin (changes requested):

build.py
CoderHam: Can we use COMMIT_SHA instead of specific to make it more accurate?

build.py
```diff
@@ -615,17 +617,30 @@ def onnxruntime_cmake_args(images, library_paths):
 
 
 def openvino_cmake_args(be, variant_index):
+    using_specific_commit = False
```
CoderHam, line 620: using_commit_sha?

CoderHam (changes requested):

build.py
GuanLuo: I think the first value is meant to be some version number, `COMMIT_SHA` doesn't fit that well

CoderHam: Then maybe just adding a commet like David G said is enough

deadeyegoodwin (approved):

CoderHam (approved):

CoderHam (dismissed):

CoderHam (approved):

PR #3946, opened 2022-02-16
Extend model repository API for config override

GuanLuo:
@deadeyegoodwin @tanmayv25 @CoderHam Look and feel of the extended API. In terms of GRPC API, I think we could `ModelRepositoryParameter` extend to have `ModelConfig` as one of the value option, but requiring JSON string for now to match the HTTP side.

### DIFF
--- a/docs/protocol/extension_model_repository.md
+++ b/docs/protocol/extension_model_repository.md
@@ -137,7 +137,8 @@ The load API accepts the following parameters:
 
 - "config" : string parameter that contains a JSON representation of the model
 configuration, which must be able to be parsed into [ModelConfig message from
-model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto). This config will be used for loading the model instead of the one in
+model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto).
+This config will be used for loading the model instead of the one in
 the model directory.
 
 A failed load request must be indicated by an HTTP error status
@@ -305,7 +306,8 @@ The RepositoryModelLoad API accepts the following parameters:
 
 - "config" : string parameter that contains a JSON representation of the model
 configuration, which must be able to be parsed into [ModelConfig message from
-model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto). This config will be used for loading the model instead of the one in
+model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto).
+This config will be used for loading the model instead of the one in
 the model directory.
 
 ### Unload
--- a/docs/protocol/extension_model_repository.md
+++ b/docs/protocol/extension_model_repository.md
@@ -139,7 +139,9 @@ The load API accepts the following parameters:
 configuration, which must be able to be parsed into [ModelConfig message from
 model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto).
 This config will be used for loading the model instead of the one in
-the model directory.
+the model directory. If config is provided, the (re-)load will be triggered as
+the model metadata has been updated, and the same (re-)load behavior will be
+applied.
 
 A failed load request must be indicated by an HTTP error status
 (typically 400). The HTTP body must contain the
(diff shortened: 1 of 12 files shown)


docs/protocol/extension_model_repository.md
```diff
@@ -116,8 +116,29 @@ $repository_index_error_response =
...
+- "parameters" : An object containing zero or more parameters for this
+  request expressed as key/value pairs. See
+  [Parameters](https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#parameters)
+  for more information.
+
+The load API accepts the following parameters:
+
+- "config" : string parameter that contains a JSON representation of the model
+configuration, which must be able to be parsed into [ModelConfig message from
+model_config.proto](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto). This config will be used for loading the model instead of the one in
```
deadeyegoodwin, line 140: format long lines to match existing

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

docs/protocol/extension_model_repository.md
```diff
@@ -116,8 +116,30 @@ $repository_index_error_response =
...
+```
+
+- "parameters" : An object containing zero or more parameters for this
+  request expressed as key/value pairs. See
+  [Parameters](https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#parameters)
+  for more information.
+
+The load API accepts the following parameters:
+
+- "config" : string parameter that contains a JSON representation of the model
```
CoderHam, line 138: IF a load operation is called on a model with a specified config, what will be the expected behavior?

docs/protocol/extension_model_repository.md
```diff
@@ -303,3 +336,10 @@ message RepositoryModelUnloadResponse
 {
 }
 ```
+
+The RepositoryModelUnload API accepts the following parameters:
+
+- "unload_dependents" : boolean parameter indicating that in addition
+  to unloading the requested model, also unload any dependent model
+  that was loaded along with the requested model (for example, the
```
CoderHam, line 344: For example, the ensemble models that were composed of the requested model.

CoderHam (changes requested):

docs/protocol/extension_model_repository.md
GuanLuo: "This config will be used for loading the model instead of the one in
the model directory."

tanmayv25: @GuanLuo If the load request with a config parameter is issued over a model that is already loaded, my guess is the model with previous config would be unloaded and the re-loaded model will use the config from the request. Am I right? 

May be just nitpicking. If this happens on an ensemble model where the config parameter modifies the composing model. The load API can trigger multiple model unloads as well. May be it is a good idea to call out these cases explicitly in the documentation.

docs/protocol/extension_model_repository.md
```diff
@@ -116,8 +116,32 @@ $repository_index_error_response =
 
 The load API requests that a model be loaded into Triton, or reloaded
 if the model is already loaded. A load request is made with an HTTP
-POST to a load endpoint. The HTTP body must be empty. A successful
-load request is indicated by a 200 HTTP status.
+POST to a load endpoint.  The HTTP body may be empty or may contain
```
CoderHam, line 119: Extra space before `The HTTP`

src/core/model_repository_manager.cc
```diff
@@ -1221,8 +1221,13 @@ ModelRepositoryManager::Create(
     // model loading / unloading error will be printed but ignored
     RETURN_IF_ERROR(local_manager->PollAndUpdateInternal(&all_models_polled));
   } else {
+    std::unordered_map<std::string, std::vector<const InferenceParameter*>>
+        models;
+    for (const auto& model_name : startup_models) {
+      models[model_name];
```
CoderHam, line 1227: Is this typo?

src/servers/http_server.cc
```diff
@@ -1265,7 +1265,48 @@ HTTPAPIServer::HandleRepositoryControl(
...
+            HTTP_RESPOND_IF_ERR(req, param.AsString(&config));
+            auto param = TRITONSERVER_ParameterNew(
+                "config", TRITONSERVER_PARAMETER_STRING, config.c_str());
+            if (param != nullptr) {
+              params.emplace_back(param);
+            } else {
+              HTTP_RESPOND_IF_ERR(
+                  req, TRITONSERVER_ErrorNew(
+                           TRITONSERVER_ERROR_INTERNAL,
+                           "unexpected error on creating Triton parameter"));
```
CoderHam, line 1303: Instead of "unexpected error on creating Triton parameter", maybe we can report a more verbose error?

CoderHam (changes requested):

src/core/model_repository_manager.cc
GuanLuo: This is to create a default map element, which is model to load without any additional load parameters

CoderHam (approved):

PR #3949, opened 2022-02-16
Protect completion queue from consecutive access

tanmayv25:
The conflict exists between batcher thread and executor thread.


CoderHam (approved):
Why were the mutex locks removed earlier? I remember we had completion_queue_mtx_ a few releases back (removed between 21.08 and 21.09.

GuanLuo (approved):

PR #3957, opened 2022-02-17
Build Java jar in sdk and QA containers

jbkyang-nvi:
Multistage build to add Java bindings jar to SDK and QA container
Actual build in client repo: https://github.com/triton-inference-server/client/pull/68

### DIFF
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -1,4 +1,4 @@
-# Copyright 2018-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -321,6 +321,9 @@ RUN GO111MODULE=off go get github.com/golang/protobuf/protoc-gen-go && \
 # user.
 WORKDIR /opt/tritonserver
 COPY --chown=1000:1000 --from=sdk /workspace/qa/ qa/
+# Copy Java API jar
+COPY --from=sdk /workspace/install/java-api-bindings/tritonserver-java-bindings.jar qa/pkgs/.
+
 
 # Remove CI tests that are meant to run only on build image and
 # install the tritonserver/triton python client APIs.
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -246,7 +246,7 @@ RUN cp client/src/grpc_generated/go/*.go qa/L0_simple_go_client/. && \
     cp client/src/grpc_generated/javascript/*.json qa/L0_simple_nodejs_client/. && \
     cp -r client/src/grpc_generated/java qa/L0_client_java/.
 # Copy Java API jar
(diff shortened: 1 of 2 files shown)


Dockerfile.sdk
```diff
@@ -179,3 +192,4 @@ ENV NVIDIA_BUILD_ID=${NVIDIA_BUILD_ID}
 
 ENV PATH /workspace/install/bin:${PATH}
 ENV LD_LIBRARY_PATH /workspace/install/lib:${LD_LIBRARY_PATH}
+
```
dyastremsky, line 195: Accidental extra newline at the end?

Dockerfile.QA
```diff
@@ -242,6 +243,10 @@ RUN cp client/src/grpc_generated/go/*.go qa/L0_simple_go_client/. && \
     cp client/src/grpc_generated/javascript/*.js qa/L0_simple_nodejs_client/. && \
     cp client/src/grpc_generated/javascript/*.json qa/L0_simple_nodejs_client/. && \
     cp -r client/src/grpc_generated/java qa/L0_client_java/.
+# Copy Java API jar
+RUN if [ "$TARGETPLATFORM" != "linux/arm64" ]; then \
```
dyastremsky, line 247: I didn't realize this won't work on arm64. Will this work on every other platform we have? (Windows, Jetpack, etc.)

dyastremsky (changes requested):

Dockerfile.QA
jbkyang-nvi: Currently no...

dyastremsky: Would it make sense to make the logic inclusive rather than exclusive here then? E.g. if $TARGETPLATFORM is in <list of supported platforms>; then...

CoderHam: Agreed. See what I did in https://github.com/triton-inference-server/server/blob/48cb049883783b7860f6e6e7cc9fc39ce24457b5/Dockerfile.sdk

Maybe we want to add a TRITON_ENABLE_JAVA ARG?

CoderHam (changes requested):

Dockerfile.QA
jbkyang-nvi: I would rather to be dependent on platforms since I think adding so many variables introduce more clutter. Is this ok?

dyastremsky (approved):

Dockerfile.QA
```diff
@@ -242,6 +243,10 @@ RUN cp client/src/grpc_generated/go/*.go qa/L0_simple_go_client/. && \
     cp client/src/grpc_generated/javascript/*.js qa/L0_simple_nodejs_client/. && \
     cp client/src/grpc_generated/javascript/*.json qa/L0_simple_nodejs_client/. && \
     cp -r client/src/grpc_generated/java qa/L0_client_java/.
+# Copy Java API jar
```
CoderHam, line 246: Java client jar?

CoderHam (changes requested):

CoderHam (approved):

CoderHam:
LGTM. @jbkyang-nvi please update the copyrights for these files

CoderHam (dismissed):

CoderHam (approved):

dyastremsky (approved):

PR #3966, opened 2022-02-22
Fix an old broken link to a model config protobuf.

lazykyama:
This is a trivial fix.
It looks like one link to a model config protobuf is still old and 404 error happens. This PR replaces it with a newer correct link.


Tabrizian (approved):

Tabrizian:
Thanks for your contribution.

PR #3977, opened 2022-02-23
Upgrade OpenVino version to 2021.4

CoderHam:


build.py
```diff
@@ -89,9 +89,8 @@
         '22.03dev',  # triton container
         '22.01',  # upstream container
         '1.10.0',  # ORT
-        '2021.2.200',  # ORT OpenVINO
-        (('2021.2', None), ('2021.4', '2021.4.582'),
-         ('SPECIFIC', 'f2f281e6')),  # Standalone OpenVINO
+        '2021.4.582',  # ORT OpenVINO
```
tanmayv25, line 92: I need to confirm with George whether they still need 2021.2 for mlperf. In this case we would have to just reorder the versions instead of removing 2021.2. It would be something like this:
```
(('2021.4', '2021.4.582'), ('2021.2', None), ('SPECIFIC', 'f2f281e6'))
```

Please hold off this change till I confirm.

tanmayv25: I confirmed with George. Please change the version string as:

```
(('2021.4', None), ('2021.4', '2021.4.582'), ('SPECIFIC', 'f2f281e6'))
```

CoderHam: Will do. Thanks for confirming the same

CoderHam: @tanmayv25 Do you mean `(('2021.4', None), ('2021.2', '2021.2.200'), ('SPECIFIC', 'f2f281e6'))`?

tanmayv25: 2021.2 is no longer needed. But they still need 2021.4 flavor with TBB. I meant the above only.

tanmayv25:
With 2021.2 gone, youcan remove all the logic that was added to support both the APIs from OV backend.

tanmayv25 (dismissed):

tanmayv25 (approved):

PR #3981, opened 2022-02-24
Add client unit test in L0_http / L0_grpc

GuanLuo:


CoderHam (dismissed):

tanmayv25:
What are these tests for? What are the clients being used in the test?

GuanLuo:
@tanmayv25 This is for https://github.com/triton-inference-server/client/pull/71, currently it is just unit testing the multi infer API, but I gave it a generic name as ideally we should adopt more on unit testing the C++ client API instead of creating additional sample clients, in this way it will be more modularized and easier to extend test coverage.

CoderHam (dismissed):

tanmayv25 (dismissed):

CoderHam (approved):

tanmayv25 (approved):

qa/L0_grpc/test.sh
```diff
@@ -82,8 +83,11 @@ if [[ "$(< /proc/sys/kernel/osrelease)" == *microsoft* ]]; then
     SIMPLE_IMAGE_CLIENT=${SDKDIR}/python/image_client
     # SIMPLE_ENSEMBLE_IMAGE_CLIENT=${SDKDIR}/python/ensemble_image_client
     SIMPLE_REUSE_INFER_OBJECTS_CLIENT=${SDKDIR}/python/reuse_infer_objects_client
+    # [FIXME] point to proper client
```
rmccorm4, line 86: Do these FIXME's need changes?

qa/L0_http/test.sh
```diff
@@ -73,8 +74,11 @@ if [[ "$(< /proc/sys/kernel/osrelease)" == *microsoft* ]]; then
     SIMPLE_SHM_CLIENT=${SDKDIR}/python/simple_http_shm_client
     SIMPLE_CUDASHM_CLIENT=${SDKDIR}/python/simple_http_cudashm_client
     SIMPLE_REUSE_INFER_OBJECTS_CLIENT=${SDKDIR}/python/reuse_infer_objects_client
+    # [FIXME] point to proper client
```
rmccorm4, line 77: Do these FIXME's need changes?

qa/L0_grpc/test.sh
GuanLuo: Not in this change, this should be taken care of when we port this test to Windows platform as well, note that the paths for the other clients above also need to be fixed ("python" path for cpp clients)

PR #3995, opened 2022-03-01
Add sanity check on new batch input types. Add model generation and test.

GuanLuo:


Tabrizian (commented):
Do we need to update the ragged batching docs too https://github.com/triton-inference-server/server/blob/main/docs/ragged_batching.md?

qa/L0_batch_input/batch_input_test.py
```diff
@@ -244,6 +244,93 @@ def test_max_element_count_as_shape(self):
...
 
+    def test_batch_item_shape_flatten(self):
+        # Use 4 set of inputs with shape
+        # [1, 4, 1], [1, 1, 2], [1, 1, 2], [1, 2, 2]
+        self.inputs = []
+        for value in [[1, 4, 1], [1, 1, 2], [1, 1, 2], [1, 2, 2]]:
+            self.inputs.append(
+                [tritonhttpclient.InferInput('RAGGED_INPUT', value, "FP32")])
+            self.inputs[-1][0].set_data_from_numpy(
+                np.full(value, value[0], np.float32))
```
CoderHam, line 255: Why are we using np.full instead of np.ones?

qa/L0_batch_input/batch_input_test.py
```diff
@@ -244,6 +244,93 @@ def test_max_element_count_as_shape(self):
...
+            self.assertTrue(False, "unexpected error {}".format(ex))
+
+    def test_batch_item_shape(self):
+        # Use 3 set of inputs with shape [2, 1, 2], [1, 1, 2], [1, 2, 2]
+        inputs = []
+        for value in [[2, 1, 2], [1, 1, 2], [1, 2, 2]]:
+            inputs.append(
+                [tritonhttpclient.InferInput('RAGGED_INPUT', value, "FP32")])
+            inputs[-1][0].set_data_from_numpy(
+                np.full(value, value[0], np.float32))
```
CoderHam, line 295: Same here. Does the value of the inputs matter?

GuanLuo:
The ragged batching docs doesn't describe all batch input types in detail, currently it points to the model config if the user wants to learn more about it, so I think the current doc is still accurate.

CoderHam (changes requested):

qa/L0_batch_input/batch_input_test.py
CoderHam: I'd change the np.full to be np.ones or hardcode the value being filled instead of making it value[0]

CoderHam (approved):

qa/common/gen_qa_ragged_models.py
```diff
@@ -471,20 +488,236 @@ def create_modelconfig(models_dir, max_batch, model_version, dtype, backend,
...
+
+    backend_spec = '''
+backend: "{}"
+'''.format(backend)
+    if backend == "tensorflow":
+        backend_spec += '''
+platform: "{}_{}"
+'''.format(backend, platform)
+
+    # Use a different model name for the non-batching variant
```
dyastremsky, line 648: Should this comment be somewhere else? This function seems to be for the batching variant.

Or maybe "a different model name from the non-batching variant"?

dyastremsky (changes requested):

dyastremsky (approved):

qa/L0_batch_input/batch_input_test.py
GuanLuo: What is the reason for this change? If it is to highlight the value is irrelevant, I think the better way is to clarify it with a comment, using hardcode value may give the opposite impression (i.e. a special value is required)

dyastremsky (dismissed):

qa/L0_batch_input/batch_input_test.py
CoderHam: Sure, you can use a comment to clarify the same.

tanmayv25 (dismissed):

dyastremsky (approved):

PR #4025, opened 2022-03-08
Add buffer attributes API

Tabrizian:


qa/L0_buffer_attributes/buffer_attributes_test.py
```diff
@@ -0,0 +1,105 @@
...
+   def test_buffer_attributes(self):
+    model_name = 'bls'
+
+    # Infer
+    clients = [httpclient.InferenceServerClient(url='localhost:8000'), grpcclient.InferenceServerClient(url='localhost:8001')]
+    triton_clients = [httpclient, grpcclient]
+    for i, client in enumerate(clients):
+
+      # To make sure no shared memory regions are registered with the
+      # server.
```
CoderHam, line 51: I don't think this needs to be on the next line

qa/L0_buffer_attributes/buffer_attributes_test.py
```diff
@@ -0,0 +1,105 @@
...
+      
+      input_byte_size = input0_data.size * input0_data.itemsize
+      output_byte_size = input_byte_size
+
+      shm_ip0_handle = cudashm.create_shared_memory_region(
+          "input0_data", input_byte_size, 0)
+      shm_op0_handle = cudashm.create_shared_memory_region(
+          "output0_data", output_byte_size, 0)
+      
+      client.register_cuda_shared_memory("input0_data", cudashm.get_raw_handle(shm_ip0_handle), 0, input_byte_size)
```
CoderHam, line 71: Run formatter

qa/L0_buffer_attributes/buffer_attributes_test.py
```diff
@@ -0,0 +1,105 @@
...
+        output0_data = cudashm.get_contents_as_numpy(shm_op0_handle, triton_to_np_dtype(output0['datatype']), output0['shape'])
+      self.assertTrue(np.all(output0_data == input0_data))
+
+      # To make sure no shared memory regions are registered with the
+      # server.
+      client.unregister_system_shared_memory()
+      client.unregister_cuda_shared_memory()
+
+if __name__ == '__main__':
+    unittest.main()
```
CoderHam, line 105: New line at end

qa/L0_buffer_attributes/models/bls/1/model.py
```diff
@@ -0,0 +1,51 @@
...
+            infer_response = infer_request.exec()
+
+            if infer_response.has_error():
+                raise pb_utils.TritonModelException(
+                    infer_response.error().message())
+
+            inference_response = pb_utils.InferenceResponse(output_tensors=[pb_utils.get_output_tensor_by_name(infer_response, 'OUTPUT0')])
+            responses.append(inference_response)
+
+        return responses
```
CoderHam, line 51: New line at end + run formatter

src/core/buffer_attributes.h
```diff
@@ -0,0 +1,78 @@
...
+  // Get the memory type
+  TRITONSERVER_MemoryType MemoryType() const;
+
+  // Get the memory type id
+  int64_t MemoryTypeId() const;
+
+ private:
+  size_t byte_size_;
+  TRITONSERVER_MemoryType memory_type_;
+  int64_t memory_type_id_;
```
CoderHam, line 75: Why do we use an int id instead of the TRITONSERVER_MemoryType enum for the memory type? A magic int mapping to the memory type should be avoided or documented clearly if necessary

CoderHam (changes requested):

qa/L0_buffer_attributes/models/bls/1/model.py
Tabrizian: It does have a newline at the end of file. When it doesn't have a new line it shows a red sign in the end.

src/core/buffer_attributes.h
Tabrizian: This is the memory_type_id not memory_type. For memory_type we are indeed using the enum.

CoderHam (approved):

PR #4029, opened 2022-03-09
Remove hardcoded tmp dir for Docker builds

CoderHam:


Dockerfile.QA
```diff
@@ -46,7 +47,7 @@ FROM ${BUILD_IMAGE} AS build
 # Ensure apt-get won't prompt for selecting options
 ENV DEBIAN_FRONTEND=noninteractive
 
-WORKDIR /tmp/tritonbuild/tritonserver/build
+WORKDIR ${TRITONBUILD_DIR}/tritonserver/build
```
GuanLuo, line 50: You don't need to call `ARG TRITONBUILD_DIR` first to re-declare it?

GuanLuo (dismissed):

Dockerfile.QA
CoderHam: Good catch

Dockerfile.sdk
```diff
@@ -134,10 +134,8 @@ RUN cd install && \
     export VERSION=`cat /workspace/TRITON_VERSION` && \
     tar zcf /workspace/v$VERSION.clients.tar.gz *
 
-# For CI testing need to install a test script.
-COPY qa/L0_sdk/test.sh /tmp/test.sh
-COPY qa/L0_sdk/grpc_test.cc /tmp/grpc_test.cc
-COPY qa/L0_sdk/http_test.cc /tmp/http_test.cc
+# For CI testing need to copy over L0_sdk test.
+COPY qa/L0_sdk L0_sdk
```
deadeyegoodwin, line 138: So just going to have a /workspace/L0_sdk directory in the SDK image? That doesn't seem great. Also, the L0_sdk test will need to be updated for this new location.

build.py
```diff
@@ -1525,6 +1525,14 @@ def get_tagged_backend(be, version):
         type=str,
         required=False,
         help='Directory containing the CMakeLists.txt file for Triton server.')
+    parser.add_argument(
+        '--tmp-dir',
+        type=str,
+        required=False,
+        default="/tmp/tritonbuild",
```
deadeyegoodwin, line 1532: tmpdir should just be /tmp

deadeyegoodwin (changes requested):

Dockerfile.sdk
CoderHam: Updated the SDK test on GitLab and created a PR for this. We want to avoid using /tmp or at least making it configurable. For this case I believe it does not make sense to add an additional param to dockerfile for just 1 test. 
How about /workspace/qa/L0_sdk?

build.py

Dockerfile.sdk
deadeyegoodwin: I agree. /workspace/qa/L0_sdk is a good idea.

deadeyegoodwin (dismissed):

Dockerfile.QA
```diff
@@ -61,10 +63,10 @@ RUN cmake -DTRITON_ENABLE_GPU=ON \
 # Add inception_graphdef model to example repo
 WORKDIR /workspace/docs/examples/model_repository
 RUN mkdir -p inception_graphdef/1 && \
-    wget -O /tmp/inception_v3_2016_08_28_frozen.pb.tar.gz \
+    wget -O ${TRITONBUILD_DIR}/../inception_v3_2016_08_28_frozen.pb.tar.gz \
         https://storage.googleapis.com/download.tensorflow.org/models/inception_v3_2016_08_28_frozen.pb.tar.gz && \
-    (cd /tmp && tar xzf inception_v3_2016_08_28_frozen.pb.tar.gz) && \
-    mv /tmp/inception_v3_2016_08_28_frozen.pb inception_graphdef/1/model.graphdef
+    (cd ${TRITONBUILD_DIR}/.. && tar xzf inception_v3_2016_08_28_frozen.pb.tar.gz) && \
+    mv ${TRITONBUILD_DIR}/../inception_v3_2016_08_28_frozen.pb inception_graphdef/1/model.graphdef
```
GuanLuo, line 69: This assumes `TRITONBUILD_DIR` has a parent folder, should just create temporary location within and clean up afterwards

CoderHam: Removed this assumption by using /tmp as default to make it consistent with tmp dir usage in build.py

GuanLuo (approved):

build.py
```diff
@@ -1171,9 +1171,10 @@ def create_dockerfile_windows(ddir, dockerfile_name, argmap, backends,
 
 def container_build(images, backends, repoagents, endpoints):
     # The cmake, build and install directories within the container.
-    build_dir = os.path.join(FLAGS.tmp_dir, 'tritonbuild')
+    build_dir = os.path.join(FLAGS.tmp_dir, 'tritonbuild').replace("\\", "/")
```
GuanLuo, line 1174: Add comment on why added the "replace", something about "\\" is added for Windows path but Docker doesn't accept that.

GuanLuo (approved):

PR #4034, opened 2022-03-10
Use decoupled gRPC streaming in preserve ordering testing

tanmayv25:
The libevhtp event loop does not ensure the defered actions are performed exactly in the same order. 
~~Using the decoupled streaming API ensures that we test preserve ordering functionality in dynamic batcher accurately.~~
(Decoupled streaming + config.pbtxt preserve_ordering) frontend can also not ensure the responses are delivered in the exact order as invoked in the callbacks in Triton core. 
Using the async GRPC non-streaming is robust as the responses are written in the callback itself ensuring the correct order.

### DIFF
--- a/qa/L0_batcher/batcher_test.py
+++ b/qa/L0_batcher/batcher_test.py
@@ -35,8 +35,15 @@ import unittest
 import numpy as np
 import infer_util as iu
 import test_util as tu
+from functools import partial
 
 import tritonclient.grpc as grpcclient
+from tritonclient.utils import *
+
+if sys.version_info >= (3, 0):
+    import queue
+else:
+    import Queue as queue
 
 # By default, find tritonserver on "localhost", but can be overridden
 # with TRITONSERVER_IPADDR envvar
@@ -75,6 +82,16 @@ _max_queue_delay_ms = 10000
 _deferred_exceptions_lock = threading.Lock()
 _deferred_exceptions = []
 
+class UserData:
+
+    def __init__(self):
+        self._completed_requests = queue.Queue()
+
+
+# Callback function used for async_stream_infer()
+def completion_callback(user_data, result, error):
+    # passing error raise and handling out
+    user_data._completed_requests.put((result, error))
 
 class BatcherTest(tu.TestResultCollector):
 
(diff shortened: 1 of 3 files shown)


qa/L0_batcher/batcher_test.py
```diff
@@ -35,8 +35,15 @@
 import numpy as np
 import infer_util as iu
 import test_util as tu
+from functools import partial
 
 import tritonclient.grpc as grpcclient
+from tritonclient.utils import *
```
GuanLuo, line 41: I believe `import *` is not a good practice as it may pollute the namespace and override functions (although not likely to happen in this case)

qa/L0_batcher/test.sh
```diff
@@ -559,10 +559,13 @@ if [[ "$(< /proc/sys/kernel/osrelease)" != *microsoft* ]]; then
 
     # Two instances will be created for the custom model, one delays 100 ms while
     # the other delays 400 ms
+    # Using decoupled mode as the non-decoupled streaming adds preserve ordering
```
GuanLuo, line 562: What does that mean? Isn't "preserve_ordering" a model config setting?

tanmayv25: The grpc frontend in streaming non-decoupled case ensures that the responses are written to the stream in order the corresponding requests were obtained. The preserver_ordering in a model config setting that ensure the Response callbacks are called in order. But even if the setting is not true a non-decoupled grpc streaming inference ensures the order in which with response is sent to client is preserved.

GuanLuo: So streaming in GRPC frontend has already preserved ordering for the whole workflow, is this intended? Or we should let it align with GRPC streaming concept where only the order over network is preserved (and user need to set `preserve_ordering` in model config to ensure the ordering within Triton)

tanmayv25: This was completely intentional. It is from the times when we didn't have the preserve_ordering in the dynamic batcher. We can do what you are suggesting but there are few problems:
1. Many users depend upon the ordering in GRPC frontend, the change will not be backwards compatible.
2. The change is likely not going to be trivial.

In the past in order to break away the ordering, we have proposed users to use decoupled transaction policy. However, as it is now (decoupled + DB preserve_ordering) is also races over the network and can not ensure the order in frontend.

GuanLuo: Doesn't GRPC streaming already ensure the ordering at network level? My mental model was that if user want to ensure the ordering, the user will use "streaming API + perserve_ordering" to ensure the order of "request received by the frontend" (streaming), "response returned by Triton core" (preserve ordering) and "response received by the client" (streaming), which together provides the guarantee on the overall ordering.

Based on your comment, the second ordering guarantee is already handled by GRPC frontend (in non-decoupled case) which surprises me, I am okay with doing so, but just want to refresh my understanding to align with the intended use case.

GuanLuo (dismissed):

qa/L0_batcher/test.sh
tanmayv25: > Doesn't GRPC streaming already ensure the ordering at network level? 

It does. The order of responses are maintained within the stream. That is the order in which messages are written in one end, the same order they would come out of the other end. Note that the ResponseCallback from the frontend does not write the message on the stream. It prepares the response message and marks it ready. Then the state objects cycles through and are selected to be written to the stream.

The non-decoupled GRPC Streaming uses a queue to hold the states in the order corresponding to the requests arrival from the stream. See here: https://github.com/triton-inference-server/server/blob/main/src/servers/grpc_server.cc#L3898-L3903
The GRPC streaming worked only like this before decoupled transaction policy was added. It does not require the user to place preserve_ordering.

CoderHam (approved):

PR #4047, opened 2022-03-11
Fix conditional call to update cache miss/hit stats only when TRITON_ENABLE_STATS is ON

rmccorm4:


CoderHam (approved):

rmccorm4:
Builds locally with
```
cmake -D TRITON_ENABLE_STATS=OFF -D TRITON_ENABLE_METRICS=OFF -D TRITON_ENABLE_METRICS_GPU=OFF ..
```

PR #4054, opened 2022-03-14
add java binding memory growth test

jackyh:
add memory growth test with the fix of memory leak based on David Y.'s original test.

### DIFF
--- a/qa/L0_java_memory_growth/Simple.java
+++ b/qa/L0_java_memory_growth/Simple.java
@@ -197,7 +197,7 @@ public class Simple {
             // releasing the buffer.
             if (!allocated_ptr.isNull()) {
               buffer.put(0, allocated_ptr);
-              buffer_userp.put(0, new BytePointer(tensor_name));
+              buffer_userp.put(0, new Loader.newGlobalRef(tensor_name));
             }
           }
 
@@ -210,11 +210,11 @@ public class Simple {
             TRITONSERVER_ResponseAllocator allocator, Pointer buffer, Pointer buffer_userp,
             long byte_size, int memory_type, long memory_type_id)
         {
-          BytePointer name = null;
+          String name = null;
           if (buffer_userp != null) {
-            name = new BytePointer(buffer_userp);
+            name = (String)Loader.accessGlobalRef(buffer_userp);
           } else {
-            name = new BytePointer("<unknown>");
+            name = new "<unknown>";
           }
 
           switch (memory_type) {
@@ -248,7 +248,7 @@ public class Simple {
               break;
           }
 
-          name.deallocate();
+          Loader.deleteGlobalRef(buffer_userp);
 
           return null;  // Success
         }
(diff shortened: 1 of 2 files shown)


jackyh:
Notes here:
1) This test is based on the one designed by David Yastremsky.
2) The main difference is:
    a), In file of "Simple.java", use pointerscope to scope the following:
        try (PointerScope scope = new PointerScope()) {
          RunInference(server, model_name, is_int, is_torch_model);
        }
       because some of the pointers in this function are failed to be GCed by JVM which will cause memory leak. If we add PointerScope here, this will explicitly tell the system that the scope of the pointers is just the function of "RunInference".
        Also, after each RunInferece, we sleep the thread for 100ms, this is tunable, but less than 10ms seems doesn't work.
      b) In file of "test.sh":
        two parameter of Java were added: -Dorg.bytedeco.javacpp.maxPhysicalBytes=2000m -Dorg.bytedeco.javacpp.maxRetries=100
        "maxPhysicalBytes" is the max allowable off-heap memory; "maxRetries" is the max retries of System.gc( ) and trim( ) before Java is able to free enough off-heap memory, if, after 100 retries, Java fails to free enough off-heap memory to make it lower than 2000M, an out-of-memory exception will be triggered and test will fail.
3) to verify the test:
    a) make the model repositories ready, something like this: https://github.com/jackyh/server/blob/java-api-prerelease/docs/test.md#generate-qa-model-repositories
    b) run test.sh
    c) you can modify "-i 1000000" to some "-i 10000000" in test.sh to do some over night tests.

jackyh:
> Fantastic work. Interesting that you're able to get rid of the memory leak even with the CUDA/TensorRT bindings included.
> 
> Do you know why we need to run Thread.sleep()? Is there an alternative approach? The bulk of the test time will be sleeping... we should make sure we know why.
> 
The reason here is because "RunInference" allocate off-heap memory too fast, and, it is the only thread that will keep allocating. But, there's no memory leak anymore if "RunInference" is pointerscoped, because if there's still memory leak, Thread.sleep() is useless, it will throw "out of mem" exception eventually as you know. Another thing is, I used JMC, and there's no potential memory leak.

> Also, are we sure this is fixing the memory leak? It looks to me like we're allowing off-heap memory of 2GB, on-heap memory of over 750MB. That's huge for this model. It seems like we **need** to get this number down. And if we're only using "out of memory" as the gauge of the success, for 3GB, we'd need to run this hundreds of millions of times to be sure we're not masking the issue by setting very high memory maximums.

I checked with memory leak with the tool of JMC and there's no potential memory leak. for "2GB" and "750MB", these are tunable, looks like our app will use "quite a lot" of off-heap memory. Since to test more configuration will take lots of time because it's need over-night tests.

dyastremsky:
@jackyh This PR should be able to be closed since the memory growth test was merged here: https://github.com/triton-inference-server/server/pull/4259

PR #4070, opened 2022-03-18
Add HTTP raw binary request support. Add documentation in binary extension

GuanLuo:


dyastremsky:
Looks good! Would it make sense to add an error test case? Like if the user requests the wrong number of bytes (too many, too little). What's expected behavior and can we verify it with testing?

src/servers/http_server.cc
```diff
@@ -2023,7 +2023,7 @@ HTTPAPIServer::GetInferenceHeaderLength(
     evhtp_request_t* req, int32_t content_length, size_t* header_length)
 {
   // Find Inference-Header-Content-Length in header. If missing set to 0
```
deadeyegoodwin, line 2025: I think this comment is outdated

qa/L0_http/http_test.py
```diff
@@ -0,0 +1,78 @@
...
+import sys
+sys.path.append("../common")
+
+import unittest
+import numpy as np
+import test_util as tu
+import requests
+
+
+class HttpTest(tu.TestResultCollector):
```
deadeyegoodwin, line 37: You should add a successful test case with BYTE type model. Also add failing test cases for BYTE type but not shape [1], fail if multiple variable-sized dimensions, fail if model has more than one input.

deadeyegoodwin (changes requested):

tanmayv25 (changes requested):

deadeyegoodwin (dismissed):

tanmayv25 (approved):

CoderHam (changes requested):
@GuanLuo can you rebase to resolve branch conflicts?

CoderHam (approved):

PR #4071, opened 2022-03-18
Port Collision Detection

dyastremsky:
Detect whether the port assigned to GRPC, HTTP, or metrics is already in use when server is starting up. If so, return a error.

### DIFF
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -222,8 +222,6 @@ if [ `grep -c "failed to start GRPC service: Unavailable - Port '0.0.0.0:8001' a
 fi
 
 SERVER_PID=$SAVED_SERVER_PID
-set -e
-
 kill $SERVER_PID
 wait $SERVER_PID
 
@@ -254,7 +252,6 @@ if [ `grep -c "failed to start HTTP service: Unavailable - Port '0.0.0.0:8000' a
 fi
 
 SERVER_PID=$SAVED_SERVER_PID
-set -e
 
 kill $SERVER_PID
 wait $SERVER_PID
@@ -286,7 +283,6 @@ if [ `grep -c "failed to start Metrics service: Unavailable - Port '0.0.0.0:8002
 fi
 
 SERVER_PID=$SAVED_SERVER_PID
-set -e
 
 kill $SERVER_PID
 wait $SERVER_PID
@@ -306,13 +302,13 @@ if [ "$SERVER_PID" == "0" ]; then
 fi
 SAVED_SERVER_PID=$SERVER_PID
 SERVER_ARGS="--model-repository=`pwd`/models --grpc-port 8003 --http-port 8004 --metrics-port 8005"
-run_server
+# Use nowait to avoid curl to port 8000
+run_server_nowait
 if [ "$SERVER_PID" == "0" ]; then
     echo -e "\n***\n*** Failed to start $SERVER\n***"
     cat $SERVER_LOG
     exit 1
 fi
-set -e
 
 kill $SERVER_PID
 wait $SERVER_PID
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -40,13 +40,12 @@ fi
 
 export CUDA_VISIBLE_DEVICES=0
 
-TEST_RESULT_FILE='test_results.txt'
 CLIENT_LOG="./client.log"
-LC_TEST=lifecycle_test.py
-
 DATADIR=/data/inferenceserver/${REPO_VERSION}
-
+LC_TEST=lifecycle_test.py
+SLEEP_TIME=10
 SERVER=/opt/tritonserver/bin/tritonserver
+TEST_RESULT_FILE='test_results.txt'
 source ../common/util.sh
 
 RET=0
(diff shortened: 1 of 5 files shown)


src/servers/grpc_server.cc
```diff
@@ -4491,6 +4493,12 @@ GRPCServer::Start()
   model_infer_cq_ = grpc_builder_.AddCompletionQueue();
   model_stream_infer_cq_ = grpc_builder_.AddCompletionQueue();
   grpc_server_ = grpc_builder_.BuildAndStart();
+  // Check if binding port failed
+  if (grpc_server_ == nullptr) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_UNAVAILABLE,
+        (std::string("Port '") + server_addr_ + "' already in use ").c_str());
+  }
```
rmccorm4, line 4501: I'm not sure this is the correct spot to check. I believe `grpc_server_ == nullptr` indicates a failure to start the server, but it didn't indicate the reason for failure.

I believe `AddListeningPort()` will specifically tell you whether we failed due to trying to bind a port that's already in use: https://github.com/triton-inference-server/server/blob/afeb2a630b83ca204f8962b832db22d06e316204/src/servers/grpc_server.cc#L4452

Reference: https://github.com/grpc/grpc/blob/1dff2dc6604f98bc09a8a3677646c196025cd475/include/grpcpp/impl/codegen/server_interface.h#L136-L148

I think that may be more correct to verify a port issue specifically.

src/servers/http_server.cc
```diff
@@ -35,10 +35,10 @@
 #include <algorithm>
 #include <list>
 #include <thread>
-#include "triton/common/constants.h"
-#include "triton/common/logging.h"
 #include "classification.h"
 #include "data_compressor.h"
+#include "triton/common/constants.h"
+#include "triton/common/logging.h"
```
rmccorm4, line 41: Looks like these are already included in the header "http_server.h"

src/servers/http_server.cc
```diff
@@ -78,7 +78,13 @@ HTTPServer::Start()
     evhtp_enable_flag(htp_, EVHTP_FLAG_ENABLE_NODELAY);
     evhtp_set_gencb(htp_, HTTPServer::Dispatch, this);
     evhtp_use_threads_wexit(htp_, NULL, NULL, thread_cnt_, NULL);
-    evhtp_bind_socket(htp_, "0.0.0.0", port_, 1024);
+    if (evhtp_bind_socket(htp_, "0.0.0.0", port_, 1024) != 0) {
```
Tabrizian, line 81: Only some of the `evhtp` functions return an error? Do we need to do error checking for other functions too?

Tabrizian: Thanks for checking David!

> Do we want to make a ticket for all the changes? 

Yes, I think we need to have a separate ticket to check the error of evhtp calls that do return an error.

dyastremsky: Any time! Created the ticket here: https://jirasw.nvidia.com/browse/DLIS-3580

src/servers/http_server.cc
dyastremsky: Good catch! Removed.

src/servers/grpc_server.cc
dyastremsky: The GitHub documentation can be a bit misleading. The error is only returned [after BuildAndStart is called](https://grpc.github.io/grpc/cpp/classgrpc_1_1_server_builder.html#ad8c7eff4f5747333b7ffbb09cef71b5e). So it would need to be checked here.

When I moved it to this part, I dropped the bound port to avoid an unnecessary variable. You raise a great point about specificity of the error, so I pushed a change to check the whether the port was bound specifically for this error. Thanks for finding this and providing documentation!

qa/L0_lifecycle/test.sh
```diff
@@ -197,6 +197,126 @@ wait $SERVER_PID
...
+fi
+
+SERVER_PID=$SAVED_SERVER_PID
+
+kill $SERVER_PID
+wait $SERVER_PID
+
+LOG_IDX=$((LOG_IDX+1))
+
+# No Port Collision Test
```
CoderHam, line 292: Maybe add/update a test case for multiple port collisions

CoderHam (approved):

Tabrizian (dismissed):

rmccorm4 (dismissed):

qa/L0_lifecycle/test.sh
```diff
@@ -197,6 +203,160 @@ wait $SERVER_PID
...
+    exit 1
+fi
+
+LOG_IDX=$((LOG_IDX+1))
+SERVER_LOG="./inference_server_$LOG_IDX.log"
+
+SAVED_SERVER_PID=$SERVER_PID
+SERVER_ARGS="--model-repository=`pwd`/models --grpc-port 8003 --http-port 8004 --metrics-port 8005"
+run_server
+sleep 10
```
jbkyang-nvi, line 346: Nit: can you add timeout as a configurable parameter at the top of the test?

dyastremsky: Thank you for the feedback. Done. :)

qa/L0_lifecycle/test.sh
dyastremsky: Thanks for providing feedback! Done.

CoderHam (dismissed):

GuanLuo (approved):

CoderHam (approved):

PR #4073, opened 2022-03-18
Fix namespace and constants

deadeyegoodwin:


rmccorm4 (approved):
Small nit that instances of `namespace ni = triton::server;` might want to be `namespace ts` or something instead, but hardly necessary at the moment.

CoderHam (approved):

PR #4086, opened 2022-03-21
fix l0 infer for cpu only containers

jbkyang-nvi:


docker/entrypoint.d/50-gpu-driver-check2.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
-if [[ "${NVIDIA_CPU_ONLY:-0}" == "1" ]]; then
```
deadeyegoodwin, line 4: Don't change this file, it is correct for the non cpu-only triton image.
You need to copy this file into docker/cpu_only/entrypoint.d and then remove the condition as noted in the other comment.

deadeyegoodwin (changes requested):

CoderHam (dismissed):

deadeyegoodwin (approved):

PR #4090, opened 2022-03-22
Multi-stage docker build for client container

krishung5:
Use multi-stage docker build to reduce size of client container


Dockerfile.sdk
```diff
@@ -39,16 +43,15 @@ ARG TRITON_ENABLE_GPU=ON
 ARG NVIDIA_TRITON_SERVER_SDK_VERSION=unknown
 ARG NVIDIA_BUILD_ID=unknown
 
-FROM ${BASE_IMAGE}
+############################################################################
+##  Build image
+############################################################################
 
-ARG TARGETPLATFORM
+FROM ${BASE_IMAGE} AS sdk_build
```
jbkyang-nvi, line 50: I don't think TARGETPLATFORM should be removed. It's used later

Dockerfile.sdk
```diff
@@ -129,7 +132,48 @@ RUN cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
...
+############################################################################
+## Create sdk container
+############################################################################
+FROM ${BASE_IMAGE}
+
+# Ensure apt-get won't prompt for selecting options
+ENV DEBIAN_FRONTEND=noninteractive
+
+# DCGM version to install for Model Analyzer
+ARG DCGM_VERSION=2.2.9
```
jbkyang-nvi, line 144: Nit: Should we add DCGM_VERSION at the top as well? Can --build-arg propagate through stages for docker build?

jbkyang-nvi: looks good otherwise

krishung5: Right, I think we should set DCGM_VERSION as a global arg so that --build-arg can be passed through each build stage. Thanks for the catch!

CoderHam: You can define the DCGM arg with the group of args in L42 like 

```
ARG DCGM_VERSION=2.2.9
``` 

But inside the final stage for creating the sdk container you will need to add it again

```
ARG DCGM_VERSION
```

For example refer to TRITON_CORE_REPO_TAG

CoderHam (dismissed):
Minor change suggested by @jbkyang-nvi  else LGTM

CoderHam (approved):

PR #4093, opened 2022-03-22
Add work around to enable building PyTorch backend in CPU-only container

CoderHam:
- Use Triton min container to copy out cuda libraries, cudnn, nccl for CPU only build when PyTorch backend is enabled.

> This is needed because the pytorch NGC container has a build that is not designed for a CPU only environment.

### DIFF
--- a/build.py
+++ b/build.py
@@ -109,7 +109,8 @@ OVERRIDE_CORE_CMAKE_FLAGS = {}
 EXTRA_BACKEND_CMAKE_FLAGS = {}
 OVERRIDE_BACKEND_CMAKE_FLAGS = {}
 
-SCRIPT_DIR = os.path.dirname(os.path.abspath(getsourcefile(lambda:0)))
+SCRIPT_DIR = os.path.dirname(os.path.abspath(getsourcefile(lambda: 0)))
+
 
 def log(msg, force=False):
     if force or not FLAGS.quiet:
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -879,6 +888,19 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
...
+            # openmpi, nccl and cudnn libs are not available. We must copy them from the Triton min container ourselves.
+            df += '''
+COPY --from=min_container /usr/local/cuda /usr/local/cuda
+COPY --from=min_container /usr/lib/x86_64-linux-gnu/libnccl.so.2 /usr/lib/x86_64-linux-gnu/libnccl.so.2
+COPY --from=min_container /usr/lib/x86_64-linux-gnu/libcudnn.so.8 /usr/lib/x86_64-linux-gnu/libcudnn.so.8
+
+RUN apt-get update && \
+        apt-get install -y --no-install-recommends openmpi-bin
+
+ENV LD_LIBRARY_PATH /usr/local/cuda/targets/x86_64-linux/lib:${LD_LIBRARY_PATH}
```
deadeyegoodwin, line 902: This gets all the required libraries and headers into the build, but don't you need to copy the stubs and some other libraries into the final image so that Triton can actually run?

CoderHam: These are only needed during runtime. Moving to final image section.

deadeyegoodwin (approved):

build.py
```diff
@@ -1075,6 +1091,33 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
  && rm -f ${_CUDA_COMPAT_PATH}/lib
 '''
 
+    elif 'pytorch' in backends:
+        # The cpu-only build uses ubuntu as the base image, and so the cuda,
+        # openmpi, nccl and cudnn libs are not available. We must copy these
+        # pytorch dependencies from the Triton min container ourselves.
```
tanmayv25, line 1097: Also add a line about how the pytorch binary is not CPU only hence have cuda deps.

CoderHam: Already added in L1267 inside container_build()
```
    # For cpu-only image we need to copy some cuda libraries and dependencies
    # since we are using a PyTorch container that is not CPU-only
```

tanmayv25: Yes. Can you add a line here because we are copying the library here.

CoderHam: Are you suggest to _replicate_ these lines to the section under L1094 or _move_ those lines to this section?

tanmayv25: I think you might have had some reason to describe the situation in L1267. So, I am just saying that we must be more descriptive here and justify why we are copying cuda files when enable_gpu is false. Something like below helps:
```
Add dependencies for pytorch backend. Note: Even if the build is cpu-only, pytorch depends upon libraries like cuda and cudnn, which are not present in ubuntu base image. Hence must copy these pytorch dependencies from the Triton min container ourselves.
```

tanmayv25 (approved):

deadeyegoodwin (approved):

PR #4097, opened 2022-03-23
Fix L0_warmup

krishung5:


CoderHam (changes requested):
On second glance, the naming convention should still be needed for the sequence models.

Only the addsub models had the need for the naming question removed.

krishung5:
@CoderHam I see, but it seems like the libtorch sequence model is still using INPUT__0 as input name. https://github.com/triton-inference-server/server/blob/main/qa/common/gen_qa_sequence_models.py#L1208

CoderHam:
> On second glance, the naming convention should still be needed for the sequence models.

Yes that is what I said in the statement above. 
The special handling for the libtorch naming convention (inputName__index) should still be needed for the sequence models.

qa/L0_warmup/test.sh
```diff
@@ -65,7 +65,7 @@ for BACKEND in ${BACKENDS}; do
     cp -r /data/inferenceserver/${REPO_VERSION}/qa_model_repository/${BACKEND}_float32_float32_float32 models/. && \
     cp -r /data/inferenceserver/${REPO_VERSION}/qa_sequence_model_repository/${BACKEND}_sequence_int32 models/.
 
-    INPUT_PREFIX="INPUT" && [ "$BACKEND" == "libtorch" ] && INPUT_PREFIX="INPUT__"
+    INPUT_PREFIX="INPUT"
```
CoderHam, line 68: Since INPUT_PREFIX is being used for both `*_zero_1_object` [identity] models and `*_float32_float32_float32` [addsub] models you should create a new variable for IDENTITY_INPUT_PREFIX as
```
IDENTITY_INPUT_PREFIX ="INPUT" && [ "$BACKEND" == "libtorch" ] && IDENTITY_INPUT_PREFIX="INPUT__"
```
to handle this case.

This way the INPUT_PREFIX change is only for addsub models and stands correct.

CoderHam (changes requested):

qa/L0_warmup/test.sh
krishung5: Thanks for clarifying, working on it now.

CoderHam (approved):
Re-run CI if needed, trigger L0_warmup and make sure it passes before merging.

PR #4099, opened 2022-03-23
Fix inference header handling for Vertex and SageMaker

GuanLuo:


src/sagemaker_server.cc
```diff
@@ -34,8 +34,9 @@ TRITONSERVER_Error*
 SagemakerAPIServer::GetInferenceHeaderLength(
     evhtp_request_t* req, int32_t content_length, size_t* header_length)
 {
-  // Check mime type and set inference header length. If missing set to 0
-  *header_length = 0;
+  // Check mime type and set inference header length.
+  // Set to content length in case that it is not specified
+  *header_length = content_length;
```
jbkyang-nvi, line 39: is content length or header length guaranteed to be 0? I see some places it is default initialized if not populated

GuanLuo: Can you link the places where the value is not initialized? The content length is initialized [here ](https://github.com/triton-inference-server/server/blob/main/src/http_server.cc#L2602)before `GetInferenceHeader()` is called

jbkyang-nvi: I think just doing `int header_length` does not guarantee the value to be 0. https://en.cppreference.com/w/cpp/language/default_initialization

CoderHam (approved):
LGTM > Copyright updates.

tanmayv25 (approved):

src/sagemaker_server.cc
GuanLuo: That is true, consider this function as the initialization for `header_length`, so it is fine as long as `header_length` is not used before calling this function.

tanmayv25 (approved):

CoderHam (approved):

PR #4106, opened 2022-03-24
Add shm leak detection to the Python backend restart test

Tabrizian:


qa/L0_backend_python/restart/models/restart/1/model.py
```diff
@@ -0,0 +1,56 @@
...
+        # the stub process will be killed to trigger Python backend restart.
+        # After that this value will be read again to make sure that it matches
+        # before restart.
+
+        file_name = 'free_memory.txt'
+        current_free_memory = str(c_utils.shared_memory.free_memory())
+        if path.exists(file_name):
+            with open(file_name, 'r') as f:
+                expected_free_memory = f.read()
+                assert(expected_free_memory == current_free_memory)
```
dyastremsky, line 45: Could you add a message to the assert to provide the values of expected_free_memory and current_free_memory?

dyastremsky (changes requested):

qa/L0_backend_python/restart/models/restart/1/model.py
Tabrizian: Improved the assert error message.

dyastremsky: Awesome! Approving, though you might want to change the verbiage (nit).  "is equal to X is not equal to" can be a bit confusing.

dyastremsky (approved):

PR #4107, opened 2022-03-24
Cleanup cmake options to include only those needed for server

deadeyegoodwin:


dyastremsky (dismissed):

CoderHam (approved):

tanmayv25 (approved):

build.py
```diff
@@ -402,22 +402,9 @@ def core_cmake_args(components, backends, install_dir):
...
-        if be == 'tensorrt':
-            cargs += tensorrt_cmake_args()
-        if (be in CORE_BACKENDS) and (be in backends):
-            if be == 'ensemble':
-                pass
-            else:
-                fail('unknown core backend {}'.format(be))
+        cmake_core_enable('TRITON_ENABLE_ENSEMBLE', 'ensemble' in backends))
+    cargs.append(
+        cmake_core_enable('TRITON_ENABLE_TENSORRT', 'tensorrt' in backends))
```
GuanLuo, line 407: Don't we still need to add the backend flags to core for supporting old model config (only specifies "platform" but not "backend") https://github.com/triton-inference-server/core/blob/main/src/model_config_utils.cc#L792-L794

deadeyegoodwin: I just made that part of the core be always be included. If a model specifies and old platform that isn't actually available then it will fail because the backend won't be available.

deadeyegoodwin: The only reason TRITON_ENABLE_TENSORRT is there is because we need to know when to build caffe2plan or not.

PR #4110, opened 2022-03-25
Fix compose.py for cpu only image with pytorch backend

CoderHam:


dyastremsky (dismissed):

build.py
```diff
@@ -1037,7 +1037,11 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
 ENV LD_LIBRARY_PATH /opt/tritonserver/backends/onnxruntime:${LD_LIBRARY_PATH}
 '''
 
-    ort_dependencies = "libgomp1" if 'onnxruntime' in backends else ""
+    # libgomp is needed by both onnxruntime and pytorch backends
+    openmp_dependencies = ""
```
tanmayv25, line 1041: just declare a variable "backend_dependencies". Then keep adding the dependencies in this conditioned on whether onnxruntime, or pytorch is a part of backends. It will be more scalable this way.

tanmayv25 (changes requested):

build.py
CoderHam: Done.

tanmayv25 (dismissed):

GuanLuo (approved):

build.py
```diff
@@ -1037,10 +1037,14 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
 ENV LD_LIBRARY_PATH /opt/tritonserver/backends/onnxruntime:${LD_LIBRARY_PATH}
 '''
 
-    ort_dependencies = "libgomp1" if 'onnxruntime' in backends else ""
-    pytorch_dependencies = ""
+    # libgomp1 is needed by both onnxruntime and pytorch backends
```
tanmayv25, line 1040: I missed this. Can you move this comment just before L1042?

tanmayv25 (approved):

PR #4121, opened 2022-03-28
Add reshape tests for autocomplete

krishung5:


GuanLuo (commented):
Are the model files generated specifically for these tests or they are the QA models that we generate? If they are the QA models, you should modify `test.sh` to copy the model files from QA model directory to the test, so the models are always up-to-date and we don't need to update them separately.

krishung5:
@GuanLuo Thanks for the catch.

qa/L0_model_config/autofill_noplatform_success/tensorflow_graphdef/reshape_config_provided/expected
```diff
@@ -0,0 +1,58 @@
+name: "reshape_config_provided"
+platform: "tensorflow_graphdef"
```
tanmayv25, line 2: It is strange to provide "platform" in the config for autofill_**noplatform**_success subtests. You can remove the platform field from here and everywhere.

Ideally, you should create another directory named "autofill_success" and place these test cases over there. But I am fine with just removing platform fields from here. Specifying backend field should be sufficient.

tanmayv25 (changes requested):

qa/L0_model_config/autofill_noplatform_success/tensorflow_graphdef/reshape_config_provided/expected
krishung5: I might have understood something, please correct me if I'm wrong, but aren't we expecting the platform field in every expected files and only removing the platform filed from the config files?

tanmayv25: My bad! This was expected file. You are good!

tanmayv25 (approved):

qa/L0_model_config/test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2019-2021, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019-2022, NVIDIA CORPORATION. All rights reserved.
```
GuanLuo, line 2: Should be `Copyright 2019-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
`, otherwise the change looks good to me

GuanLuo (approved):

PR #4122, opened 2022-03-28
Add shared memory leak detection to Python backend tests

Tabrizian:
Adds shared memory leak detection to Python backend tests. The tests that involved growing the shared memory region do not use this utility.

Corresponding Python backend PR: https://github.com/triton-inference-server/python_backend/pull/130

Usage is shown in the `shm_util.py`.

### DIFF
--- a/qa/L0_backend_python/ensemble/ensemble_test.py
+++ b/qa/L0_backend_python/ensemble/ensemble_test.py
@@ -28,6 +28,7 @@ import sys
 sys.path.append("../../common")
 
 import test_util as tu
+import shm_util
 import tritonclient.http as httpclient
 from tritonclient.utils import *
 import numpy as np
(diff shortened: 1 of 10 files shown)


qa/L0_backend_python/python_unittest.py
```diff
@@ -48,6 +51,17 @@ def test_python_unittest(self):
             # Otherwise, it will return 0.
             self.assertEqual(output0, [1])
 
+    def test_python_unittest(self):
+        model_name = os.environ['MODEL_NAME']
+
+        # The shared memory size changes during these tests and as a result
```
tanmayv25, line 57: Can we provide an optional "threshold" to the probe? This way we can test whether the increment or decrement in the shared memory regions is within expected range.

Anyways you should elaborate more here. A little more about why is the memory change expected. Is it an increase or decrease?

tanmayv25: Also, does it keep increasing with every infer call? If it is only for once- initialization, then we can run the unittest again within probe this time after L60.

Tabrizian: > Can we provide an optional "threshold" to the probe? This way we can test whether the increment or decrement in the shared memory regions is within expected range.

The problem here is that the shared memory region might grow as a result of requiring more memory. Because of this when the shared memory probe is examining the memory usage, it might discover that the free memory has increased instead of being the same. Per your second suggestion, I called the tests that grow the region two times and did the shared memory probing in the second call.

tanmayv25: > Per your second suggestion, I called the tests that **grow the region two times** and did the shared memory probing in the second call.

If the test is growing the memory even in second invocation then wouldn't the test still fail along the similar reasoning?

Tabrizian: It wouldn't grow for the second time because now the region is grown and doesn't require additional memory. Let's say the inference requires 128MB of shared memory and the region is 64 MB. When the inference is called, the region is grown to 128MB. When the second inference happens, the region size is still 128 and there is no need for growing the region.

qa/common/shm_util.py
```diff
@@ -344,3 +348,49 @@ def register_add_either_shm_regions(inputs, outputs, shm_region_prefix,
...
+            time.sleep(5)
+            self._shm_region_free_sizes = []
+            for shm_monitor in self._shm_monitors:
+                self._shm_region_free_sizes.append(shm_monitor.free_memory())
+
+            return self
+
+        def __exit__(self, type, value, traceback):
+            # Wait so that any pending shared memory object is freed.
+            time.sleep(5)
```
tanmayv25, line 372: Are these sleeps needed only for BLS case? If yes, then they should be called out explicitly in those cases.
No need to slow down other tests with 10sec overhead right? 
+ can we somehow determine there are no pending shared memory operations?

Tabrizian: > Are these sleeps needed only for BLS case? If yes, then they should be called out explicitly in those cases.
No need to slow down other tests with 10sec overhead right?

The reason for the delay is that the pending shared memory frees, will be called when the backend returns from the ModelInstanceExecute and this is after the responses have been sent. I will put the delay only for the BLS tests.

tanmayv25 (approved):

qa/L0_backend_python/ensemble/ensemble_test.py
```diff
@@ -28,6 +28,7 @@
 sys.path.append("../../common")
 
 import test_util as tu
+import shm_util
```
CoderHam, line 31: Make Copyright updates here and where needed else LGTM

CoderHam (changes requested):

tanmayv25 (approved):

CoderHam (approved):

PR #4123, opened 2022-03-28
Enable PyTorch Models to Share Weights

dyastremsky:
This PR creates tests for an associated PyTorch backend change, which allows a Triton user to enable multiple instances of a model on the same device to share weights. This is turned off by default and can be enabled via a model config parameter, "ENABLE_WEIGHT_SHARING."

Enabling weight reuse can reduce memory usage of model loading and inference. It should not be used with models that maintain state (due to the reusing of weights).

Related backend code and documentation change: https://github.com/triton-inference-server/pytorch_backend/pull/54


qa/L0_libtorch_shared_weights/libtorch_shared_weights_test.py
```diff
@@ -0,0 +1,93 @@
...
+            # Get the result from the initiated asynchronous inference request.
+            # Note the call will block until the server responds.
+            results = requests[i].get_result()
+
+            output_data = results.as_numpy("OUTPUT__0")
+            self.assertIsNotNone(output_data,
+                                 "error: expected 'OUTPUT__0' to be found")
+            np.testing.assert_allclose(output_data, input_datas[i])
+
+        # Make sure the requests ran in parallel.
```
tanmayv25, line 71: How are we testing that requests ran in parallel? With an instance count of 4, I presume only 4 requests will run at a time in pytorch.

qa/L0_libtorch_shared_weights/test.sh
```diff
@@ -0,0 +1,172 @@
...
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+## Basic framework from other test.sh
+
+## Copy over libtorch non-sequence model
+## Make model have multiple instances on one device
+## Make non-sequence model have share_weights on
```
tanmayv25, line 31: I think the order is reverse. First the share_weight is on, then it is off.

qa/L0_libtorch_shared_weights/test.sh
```diff
@@ -0,0 +1,172 @@
...
+LOG_IDX=0
+
+# Prepare model repository
+rm -fr models
+mkdir models
+for i in models; do
+    cp -r $DATADIR//qa_identity_model_repository/libtorch_nobatch_zero_1_float32 models/.
+done
+
+for MC in `ls models/libtorch*/config.pbtxt`; do
```
tanmayv25, line 76: I think another combination to try would be kinds : CPU and GPU.

tanmayv25 (changes requested):

qa/L0_libtorch_shared_weights/test.sh
dyastremsky: Thanks for catching this. This was an outline for myself... I thought I removed it. 😅 Deleted.

qa/L0_libtorch_shared_weights/libtorch_shared_weights_test.py
dyastremsky: Good point. I read through the documentation and see now that requests for that model wait until an instance is free, if all the instances are taken. I fixed the comments. I also updated the instance count to 16 and concurrency to 128, since it's a small model (test takes a few seconds), to ensure everything works with greater parallelism.

qa/L0_libtorch_shared_weights/test.sh
dyastremsky: Added!

qa/L0_libtorch_shared_weights/test.sh
```diff
@@ -0,0 +1,204 @@
...
+    echo "instance_group [ { count: ${INSTANCE_CNT} kind: KIND_GPU}]" >> $MC
+done
+
+# SharedWeightsTest.test_pytorch_identity_model
```
tanmayv25, line 69: Instead of duplicating the code for gpu and cpu, you can run it in a loop. 
See how we do it in L0_infer:
https://github.com/triton-inference-server/server/blob/main/qa/L0_infer/test.sh#L132
https://github.com/triton-inference-server/server/blob/main/qa/L0_infer/test.sh#L238

dyastremsky: Thanks for the tip! Updated.

CoderHam (approved):

tanmayv25 (approved):

CoderHam (approved):

dyastremsky:
@joaopcm1996 It should not have an impact. Since the models are already trained, we're just sharing the constant (read-only) values.

PR #4125, opened 2022-03-29
Fix the model copies in L0_sequence_batcher test

tanmayv25:


qa/L0_sequence_batcher/test.sh
```diff
@@ -210,13 +210,11 @@ for BACKEND in $BACKENDS; do
...
     done
 
-    if [[ $BACKEND == "graphdef" ]]; then
-      MODELS="$MODELS $DATADIR/$FIXED_MODEL_REPOSITORY/${BACKEND}_sequence_graphdef_sequence_int32"
-    fi
-
     if [ "$ENSEMBLES" == "1" ]; then
       for DTYPE in $DTYPES; do
-        MODELS="$MODELS $DATADIR/qa_ensemble_model_repository/$FIXED_MODEL_REPOSITORY/*_${BACKEND}_sequence_${DTYPE}"
+        if [[ $DTYPE != "bool" ]]; then
```
Tabrizian, line 215: Why is this line is required? Is it because ensembles do not have bool version? Can you add a comment explaining this?

Can we use `set +e`, `set -e` so that the script fails if there are invalid copies?

tanmayv25: > Is it because ensembles do not have bool version?
Yes. See here: https://github.com/triton-inference-server/server/blob/main/qa/common/gen_qa_sequence_models.py#L1421-L1422

> Can we use set +e, set -e so that the script fails if there are invalid copies?
I have a ticket for this to revisit all our test scripts. I have used L0_sequence_batcher as an example.
DLIS-3614

CoderHam (dismissed):

krishung5 (approved):

Tabrizian (approved):

PR #4128, opened 2022-03-29
Update README and docs post-22.03 release

mc-nv:


build.py
```diff
@@ -88,7 +88,7 @@
 TRITON_VERSION_MAP = {
     '2.21.0dev': (
         '22.04dev',  # triton container
-        '22.02',  # upstream container
```
dzier, line 91: Don't do this yet. This is the next step, "Advance upstream containers" and requires a lot more work. We can do this one together. For this commit, leave it as 22.02.

dzier (changes requested):

deadeyegoodwin (dismissed):

build.py
mc-nv: @dzier  I'have reverted it back as you suggested

dzier (approved):

PR #4136, opened 2022-03-30
Allow http and grpc server bind to specific address

krishung5:


src/grpc_server.cc
```diff
@@ -4416,12 +4416,12 @@ GRPCServer::Create(
     const std::shared_ptr<TRITONSERVER_Server>& server,
     triton::server::TraceManager* trace_manager,
     const std::shared_ptr<SharedMemoryManager>& shm_manager, int32_t port,
-    bool use_ssl, const SslOptions& ssl_options, int infer_allocation_pool_size,
-    grpc_compression_level compression_level,
+    std::string grpc_addr, bool use_ssl, const SslOptions& ssl_options,
```
GuanLuo, line 4419: just call it `address` in grpc_server and http_server. At this level, there is no need to distinguish it.

src/http_server.h
```diff
@@ -89,7 +91,7 @@ class HTTPMetricsServer : public HTTPServer {
   explicit HTTPMetricsServer(
       const std::shared_ptr<TRITONSERVER_Server>& server, const int32_t port,
       const int thread_cnt)
-      : HTTPServer(port, thread_cnt), server_(server),
+      : HTTPServer(port, "0.0.0.0", thread_cnt), server_(server),
```
GuanLuo, line 94: I think metrics server should also accept the new address parameter

GuanLuo (changes requested):
There is a `CheckPortCollision()` function in main.cc which you will need to update as well. i.e. different endpoint can bind to the same port as long as they are listening to different address, so the collision should check both the address and port

src/main.cc
```diff
@@ -599,59 +599,63 @@ bool
...
   for (auto curr_it = ports.begin(); curr_it != ports.end(); ++curr_it) {
     // If the current service doesn't specify the allow port range for other
     // services, then we don't need to revisit the checked services
-    auto comparing_it = (std::get<2>(*curr_it)) ? ports.begin() : (curr_it + 1);
+    auto comparing_it = (std::get<3>(*curr_it)) ? ports.begin() : (curr_it + 1);
     for (; comparing_it != ports.end(); ++comparing_it) {
       if (comparing_it == curr_it) {
         continue;
       }
       // Set range and comparing service port is out of range
```
GuanLuo, line 641: I think you can just check if the addresses are the same here, if they are not the same, then there won't be collision and you can `continue;`. Otherwise, then you need to check if the port are colliding.

krishung5: Thanks for the feedback. Updated!

src/main.cc
```diff
@@ -591,59 +599,65 @@ bool
...
-                  << std::get<3>(*curr_it) << ", " << std::get<4>(*curr_it)
+                  << std::get<2>(*comparing_it) << ", allowed port range is ["
+                  << std::get<4>(*curr_it) << ", " << std::get<5>(*curr_it)
                   << "]" << std::endl;
         return true;
       }
-      if (std::get<1>(*curr_it) == std::get<1>(*comparing_it)) {
+      if (std::get<1>(*curr_it) != std::get<1>(*comparing_it)) {
+        continue;
+      }
```
GuanLuo, line 654: Move to line 640, the range check can be skipped as well if the addresses are different.

GuanLuo (changes requested):
Also need to add tests for this feature.

src/main.cc
krishung5: I see. Thanks for the catch. Updated.

src/main.cc
```diff
@@ -367,6 +373,8 @@ std::vector<Option> options_
        "Allow the server to listen for GRPC requests."},
       {OPTION_GRPC_PORT, "grpc-port", Option::ArgInt,
        "The port for the server to listen on for GRPC requests."},
+      {OPTION_GRPC_ADDR, "grpc-addr", Option::ArgStr,
```
GuanLuo, line 376: nit: `xxx-address`

krishung5: Should I also make changes to all the parameters/arguments?

src/http_server.cc
```diff
@@ -75,10 +75,10 @@ HTTPServer::Start()
     evhtp_enable_flag(htp_, EVHTP_FLAG_ENABLE_NODELAY);
     evhtp_set_gencb(htp_, HTTPServer::Dispatch, this);
     evhtp_use_threads_wexit(htp_, NULL, NULL, thread_cnt_, NULL);
-    if (evhtp_bind_socket(htp_, "0.0.0.0", port_, 1024) != 0) {
+    if (evhtp_bind_socket(htp_, http_address_.c_str(), port_, 1024) != 0) {
       return TRITONSERVER_ErrorNew(
           TRITONSERVER_ERROR_UNAVAILABLE,
-          (std::string("Port '0.0.0.0:") + std::to_string(port_) +
+          (std::string("Port '") + http_address_ + ":" + std::to_string(port_) +
```
CoderHam, line 81: Fix error message here. The port is just `port_`. A socket is a combination of port and IP address.

src/http_server.h
```diff
@@ -66,6 +67,7 @@ class HTTPServer {
   static void StopCallback(evutil_socket_t sock, short events, void* arg);
 
   int32_t port_;
+  std::string http_address_;
```
CoderHam, line 70: Change to just `address_`?

src/sagemaker_server.h
```diff
@@ -60,7 +60,8 @@ class SagemakerAPIServer : public HTTPAPIServer {
       triton::server::TraceManager* trace_manager,
       const std::shared_ptr<SharedMemoryManager>& shm_manager,
       const int32_t port, const int thread_cnt)
-      : HTTPAPIServer(server, trace_manager, shm_manager, port, thread_cnt),
+      : HTTPAPIServer(
```
CoderHam, line 63: Do not set default magic values here. These should be done in main.cc and commented to explain why

src/vertex_ai_server.cc
```diff
@@ -41,7 +41,8 @@ VertexAiAPIServer::VertexAiAPIServer(
     const std::shared_ptr<SharedMemoryManager>& shm_manager, const int32_t port,
     const int thread_cnt, const std::string& prediction_route,
     const std::string& health_route, const std::string& default_model_name)
-    : HTTPAPIServer(server, trace_manager, shm_manager, port, thread_cnt),
+    : HTTPAPIServer(
+          server, trace_manager, shm_manager, port, "0.0.0.0", thread_cnt),
```
CoderHam, line 45: Same here.

CoderHam (changes requested):

src/vertex_ai_server.cc
krishung5: Thanks for the review. Updated.

CoderHam (dismissed):

qa/L0_socket/test.sh
```diff
@@ -0,0 +1,295 @@
...
+        # when using different addresses, allow http/grpc port overlap with grpc/http explicit
+        if [ "$address" == "explicit" ]; then
+            if [ "$p" == "http" ]; then
+                SERVER_ARGS="--model-repository=$DATADIR $DIFF_EXPLICIT_ADDRESS --http-port 8007 --grpc-port 8007"
+            else
+                SERVER_ARGS="--model-repository=$DATADIR $DIFF_EXPLICIT_ADDRESS --grpc-port 8007 --http-port 8007"
+            fi
+            run_server_nowait
+            sleep 15
+            if [ "$SERVER_PID" == "0" ]; then
```
GuanLuo, line 171: Should send ready / live request to make sure the server is running.

GuanLuo (changes requested):
Make sure you also have the corresponding change in CI since the test name is changed

CoderHam:
@krishung5 you will need to update gitlab yaml for the change in naming for the L0_multiple_ports to avoid breaking the CI

krishung5:
@GuanLuo @CoderHam Thank you both for the heads up! Will update the CI as well.

qa/L0_socket/test.sh
krishung5: Updated

GuanLuo (approved):

PR #4141, opened 2022-03-31
Improve documentation for Jetson 22.03

CoderHam:
- Document segmentation fault in ORT backend due to pytorch backend
- Need to set LD_LIBRARY_PATH while loading pytorch models

Follow up:
```
$ ldd backends/pytorch/libtriton_pytorch.so
       ...
        libc10.so => not found
        libc10_cuda.so => not found
        ...
```

### DIFF
--- a/docs/jetson.md
+++ b/docs/jetson.md
@@ -179,12 +179,23 @@ The PyTorch runtime depenencies are the same as the build dependencies listed ab
 ### Usage
 
 **Note**: The PyTorch backend depends on libomp.so, which is not loaded automatically.
-If using the PyTorch backend in Triton, you need to set the LD_LIBRARY_PATH to allow
+If using the PyTorch backend in Triton, you must set the LD_LIBRARY_PATH to allow
 libomp.so to be loaded as needed before launching Triton.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/jetson.md
```diff
@@ -179,13 +179,24 @@ The PyTorch runtime depenencies are the same as the build dependencies listed ab
...
 ```
 
+When launching Triton, you must set the LD_LIBRARY_PATH to include the pytorch backend
+directory to avoid the error `unable to load shared library: libc10.so`.
+
+```
+LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/path/to/tritonserver/backends/pytorch
+```
+
+When shutting down Triton after loading Onnx Runtime models, you will notice a segmentation
```
tanmayv25, line 196: Why is there a segmentation fault in ORT backend because of a missing dependency of pytorch backend?

CoderHam: Yes. I believe I have fixed the segfault already and am trying a build with this. Unfortunately the CI runners for jetson are down so I am running the test manually to confirm the same

CoderHam:
Closing since I have resolved these issues

PR #4148, opened 2022-04-02
Fix shared memory monitor on windows

Tabrizian:


qa/common/shm_util.py
```diff
@@ -29,10 +29,10 @@
 import numpy as np
 import time
 from ctypes import *
+import platform
```
CoderHam, line 32: Is this needed?

CoderHam (changes requested):

qa/common/shm_util.py
Tabrizian: Fixed

CoderHam (approved):

GuanLuo (approved):

PR #4174, opened 2022-04-06
Update build.py to generate scripts containing build commands

deadeyegoodwin:


deadeyegoodwin:
We will not merge this until post-22.04

PR #4175, opened 2022-04-06
Set http address for metrics when http service is disable

krishung5:


src/main.cc
```diff
@@ -133,6 +133,10 @@ std::unique_ptr<triton::server::HTTPServer> metrics_service_;
 bool allow_metrics_ = true;
 int32_t metrics_port_ = 8002;
 float metrics_interval_ms_ = 2000;
+#ifndef TRITON_ENABLE_HTTP
+// Need to set http address for metrics when http service is disable
+std::string http_address_ = "0.0.0.0";
```
CoderHam, line 138: We are assuming the http address and metrics address is always the same? I guess that works fine but I would recommend adding a comment about the same.

CoderHam (changes requested):

src/main.cc
krishung5: Yes, right now we are using the same address for http and metrics. Comment added, please review. Please also let me know if we want to use a different address for metrics. I can make changes for that.

tanmayv25 (dismissed):

src/main.cc
CoderHam: Thanks. I think for now we can use the same http address for metrics. Let's worry about that only when we have future requests. 

For now make sure the docs / help indicates that http address is reused for metrics endpoint as well.

CoderHam (dismissed):

CoderHam (approved):

PR #4180, opened 2022-04-07
Add argument validation tests

Tabrizian:


qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class ArgumentValidationTest(unittest.TestCase):
+    def test_infer_request_args(self):
+        # Dummy arguments used in the tests.
+        inputs = [pb_utils.Tensor('INPUT0', np.asarray([1, 2], dtype=np.int32))]
+        model_name = 'my_model'
+        requested_output_names = ['my_output']
+
+        # inputs field validation
```
dyastremsky, line 39: Perhaps structure the comments a bit more for clarity? And add the specifics of each test.
For example:
Validating invalid inference request arguments return error
Inputs
Null input
_test_
Null input list
_test_

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
```diff
@@ -0,0 +1,132 @@
...
+        with self.assertRaises(BaseException) as e:
+            pb_utils.InferenceRequest(inputs=[None], model_name=model_name,
+                    requested_output_names=requested_output_names)
+
+        with self.assertRaises(BaseException) as e:
```
dyastremsky, line 44: Does error message checking not matter here? If there's an easy way to add that, it could be useful. That'd be more specific, so the exception isn't being returned for the wrong reason and we fail to actually check something.

If it'd add testing complexity, it could be not worth it, since the test is checking the desired behavior.

dyastremsky (changes requested):

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
tanmayv25: I would prefer a more specific error checking in this case even if it adds test complexity. The correct error messages are crucial for developer. 
Also, checking for specific error message would make the documentation of these test you requested above redundant.

GuanLuo: Is `with self.assertRaises(BaseException) as e:` similar to `try ... catch ...` that expects an exception to be raised? If we are checking exception, shouldn't we expect a more specific exception instead of `BaseException`? i.e. `InferenceServerException` in client testing.

Tabrizian: > Is with self.assertRaises(BaseException) as e: similar to try ... catch ... that expects an exception to be raised? If we are checking exception, shouldn't we expect a more specific exception instead of BaseException? i.e. InferenceServerException in client testing.

@GuanLuo I'll make this change.

@dyastremsky @tanmayv25 The errors are quite verbose so I'd rather not check for the exact error message here. I think this will make the test bloated with repetitive messages. I've shown an example below:

```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. c_python_backend_utils.InferenceRequest(request_id: str = '', correlation_id: int = 0, inputs: List[triton::backend::python::PbTensor], requested_output_names: List[str], model_name: str, model_version: int = -1, flags: int = 0)

Invoked with: kwargs: inputs=None, model_name='my_model', requested_output_names=['my_output']
```

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
Tabrizian: Improved the comment structure

tanmayv25 (approved):

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
dyastremsky: Makes sense. The tests are clear and the specific error checking is good, so approving. For future PRs, we could probably do a self.assertIn(str(e.exception), "incompatible constructor arguments").

Alternatively, if checking many error messages like in this test, it may be worthwhile to write a function like [this one.](https://stackoverflow.com/a/8673096), except with assertIn. Then, you can just add a few words from the error as a parameter for assertRaisesWithMessage.

dyastremsky (dismissed):

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
Tabrizian: > Alternatively, if checking many error messages like in this test, it may be worthwhile to write a function like [this one.](https://stackoverflow.com/a/8673096), except with assertIn. Then, you can just add a few words from the error as a parameter for assertRaisesWithMessage.

Sounds like a good suggestion that can apply to many of our unittests. Can you file a ticket for this?

dyastremsky: Sure, see here: https://jirasw.nvidia.com/browse/DLIS-3689

dyastremsky (approved):

CoderHam (approved):

PR #4181, opened 2022-04-07
Enable Triton for SageMaker MME mode

nskool:
[Currently work-in-progress, uses references to personal repo (github.com/nskool). Will be removed before PR is marked Ready for review]

1. Detect SM SME or MME mode via serve script. Accordingly, set model repository.
2. Detect path regex and parse input request
3. Adds implementation for SageMaker LOAD, LIST, GET, UNLOAD and INVOKE APIs as per: https://docs.aws.amazon.com/sagemaker/latest/dg/mms-container-apis.html
4. Add error-out if ensemble is used in MME mode
5. Add validation tests in L0_sagemaker/ folder for MME APIs.


src/sagemaker_server.cc
```diff
@@ -138,4 +283,285 @@ SagemakerAPIServer::Create(
...
+  std::string model_name = parse_map.at("model_name");
+
+  /* Error out if there's more than one subdir/version within
+   * supplied model repo, as ensemble in MME is not (currently)
+   * supported
+   */
+  DIR* dir;
+  struct dirent* ent;
+  int dir_count = 0;
+  if ((dir = opendir(repo_path.c_str())) != NULL) {
```
GuanLuo, line 462: This may not be portable to other platform like Windows? Should the CMakeLists.txt be updated to disallow building with SM endpoint on Windows?

src/sagemaker_server.cc
```diff
@@ -138,4 +283,285 @@ SagemakerAPIServer::Create(
...
+   * 3rd customer_subdir: optional_customer_subdir
+   */
+
+  std::string repo_parent_path, subdir, customer_subdir;
+  RE2::FullMatch(
+      repo_path, model_path_regex_, &repo_parent_path, &subdir,
+      &customer_subdir);
+
+  auto param = TRITONSERVER_ParameterNew(
+      subdir.c_str(), TRITONSERVER_PARAMETER_STRING, model_name.c_str());
```
GuanLuo, line 497: Make sure `TRITONSERVER_ParameterDelete`  is called after use

src/sagemaker_server.cc
```diff
@@ -138,4 +283,285 @@ SagemakerAPIServer::Create(
...
+  } else if (err != nullptr) {
+    EVBufferAddErrorJson(req->buffer_out, err);
+    evhtp_send_reply(req, EVHTP_RES_BADREQ);
+    TRITONSERVER_ErrorDelete(err);
+    return;
+  }
+
+  /* TODO: Check if we should use TRITONSERVER_ServerLoadModelWithParameters */
+  TRITONSERVER_Error* load_err =
+      nullptr; /* TODO: Check if we can re-use previous object*/
```
GuanLuo, line 530: Yes, `err` can be re-used since it will still be `nullptr` if control reaches this point

src/sagemaker_server.h
```diff
@@ -103,6 +125,9 @@ class SagemakerAPIServer : public HTTPAPIServer {
   const std::string model_version_str_;
 
   static const std::string binary_mime_type_;
+
+  // Maintain list of loaded models
+  std::unordered_map<std::string, std::string> sagemaker_models_list;
```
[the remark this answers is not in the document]
GuanLuo: Looks like there can be concurrent read / write to this object (concurrent calls to different endpoint APIs), should avoid race condition with mutex.

GuanLuo (commented):
Sorry for not reviewing it sooner, left some comments. Let me know when it is ready to go through CI

nskool:
> Sorry for not reviewing it sooner, left some comments. Let me know when it is ready to go through CI

Thanks for the review! I have made changes as per the comments, if they look good, could you enable the CI?

GuanLuo:
Just triggered the CI, will also take another look of the PR at the meanwhile

src/sagemaker_server.cc
GuanLuo: Looks like by default the SageMaker endpoint will not be built on Windows ([ref](https://github.com/triton-inference-server/server/blob/main/build.py#L1677)), but you may add an explicit check [here](https://github.com/triton-inference-server/server/blob/main/CMakeLists.txt#L99) to disallow specifying SageMaker endpoint when `WIN32` is true

GuanLuo (approved):
The CI is looking good, tag @CoderHam @tanmayv25 for awareness

PR #4192, opened 2022-04-08
Add testing for incorrect shape argument in PA

Tabrizian:


tanmayv25 (changes requested):
Doesn't tests like these are more suitable to be added as a unit test now with the doctest support?

https://github.com/triton-inference-server/client/commit/ae2e06097180cb86f88f52cb7178a8e2e6f04aa0

Tabrizian:
I agree that the unit test framework would be a good use case for this kinds of testing. However, I think the infrastructure is still not there yet. It looks like the doctest itself doesn't have mocking capabilities and we need to rely on other libraries for mocking which would be required if we wanted to add this test as a unittest.

tanmayv25 (approved):
@matthewkotila note that these tests makes a prime candidate for unit testing.

PR #4207, opened 2022-04-12
Fix L0_backend_python

Tabrizian:
Looks like in some cases there could be shared memory growth on the second call too.

### DIFF
--- a/qa/L0_backend_python/python_unittest.py
+++ b/qa/L0_backend_python/python_unittest.py
@@ -56,8 +56,9 @@ class PythonUnittest(tu.TestResultCollector):
 
         if model_name == 'bls' or model_name == 'bls_memory' or model_name == 'bls_memory_async':
             # For these tests, the memory region size will be grown. Because of
-            # this we need to use the shared memory probe only on the second
+            # this we need to use the shared memory probe only on the later
             # call so that the probe can detect the leak correctly.
+            self._run_unittest(model_name)
             self._run_unittest(model_name)
             with self._shm_leak_detector.Probe() as shm_probe:
                 self._run_unittest(model_name)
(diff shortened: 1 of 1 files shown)


tanmayv25:
> Looks like in some cases there could be shared memory growth on the second call too.

Is this expected? And why was this intermittent?
Based on this explanation the second call should already have all the memory extended right? 
https://github.com/triton-inference-server/server/pull/4122#discussion_r838024839

Tabrizian:
If I run the test more than 2 times, the growth doesn't happen anymore so I don't think there is a leak. But it is curious why memory region growing is happening on the second calling of the `bls` test. Looking into it.

Tabrizian:
This looks like an intermittent issue that requires further investigation. I have filed DLIS-3684. But I don't think there is any memory leak happening here because when I run the `bls` test 3-10 times with the shared memory detector probe, it doesn't detect any leak. I think it is likely because some variable is still in scope and hence it causes a difference between the amount of shared memory before and after the call.

qa/L0_backend_python/python_unittest.py
```diff
@@ -56,9 +56,10 @@ def test_python_unittest(self):
 
         if model_name == 'bls' or model_name == 'bls_memory' or model_name == 'bls_memory_async':
             # For these tests, the memory region size will be grown. Because of
-            # this we need to use the shared memory probe only on the second
+            # this we need to use the shared memory probe only on the later
             # call so that the probe can detect the leak correctly.
             self._run_unittest(model_name)
+            self._run_unittest(model_name)
```
tanmayv25, line 62: Does it make sense to add the jira ticket you create over L62?

tanmayv25 (approved):

qa/L0_backend_python/python_unittest.py
Tabrizian: Added.

tanmayv25 (approved):

PR #4226, opened 2022-04-15
Enable isolation for windows container build process. Moving away from commit hash reference and start using tag.

mc-nv:
**Dockerfile.win10.min**
---
Moving away from checking out the specific commit and start using [vcpkg:tag:2022.04.12](https://github.com/microsoft/vcpkg/tree/2022.04.12) repository tag reference. 

**build.py** 
---
During the new CI runner node to setup, I got an issue with the docker engine. Our [buyild.py](https://github.com/triton-inference-server/server/blob/main/build.py) Python script starts the docker build process inside docker container and on particular node we are getting issue with obtaining proper process from the kernel like.
```
hcsshim::CreateComputeSystem <CONTAINER HASH>: The requested resource is in use.
```
I was able to use flag `--isolation=process` in order to handle the issue.
This approach has a reference to official documentation:
[Microsoft : Virtualization / Containers on Windows / Windows Container Essentials / Isolation Modes](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/hyperv-container)
[Docker : Security](https://docs.docker.com/engine/security/)[](https://jirasw.nvidia.com/secure/RapidBoard.jspa?rapidView=28774&view=detail&selectedIssue=DLIS-3582#)

### DIFF
--- a/build.py
+++ b/build.py
@@ -1333,6 +1333,7 @@ def create_docker_build_script(script_name, container_install_dir,
 
         if target_platform() == 'windows':
             runargs += ['--memory', FLAGS.container_memory]
+            runargs += ['--isolation=process']
             runargs += [
                 '-v', '\\\\.\pipe\docker_engine:\\\\.\pipe\docker_engine'
             ]
--- a/Dockerfile.win10.min
+++ b/Dockerfile.win10.min
@@ -58,12 +58,8 @@ ARG VS_INSTALL_PATH_WP="C:\BuildTools"
 RUN powershell.exe Start-Process -FilePath vs_buildtools.exe -ArgumentList "--wait","--quiet","--norestart","--nocache","--installPath","%VS_INSTALL_PATH_WP%","--channelUri","C:\tmp\doesnotexist.chman","--add","Microsoft.VisualStudio.Workload.VCTools`;includeRecommended","--add","Microsoft.Component.MSBuild" -Wait -PassThru
 
 WORKDIR /
-# vcpkg 2021.05.12 does not contain b64 port, so checking out master branch here (commit a82c62d)
-# should be switched to next vcpkg tag when one is created
-# option --depth=1 should also be restored
-RUN git clone --single-branch -b master https://github.com/microsoft/vcpkg.git
+RUN git clone --single-branch -b 2022.04.12 --depth=1 https://github.com/microsoft/vcpkg.git
 WORKDIR /vcpkg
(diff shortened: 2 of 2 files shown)


build.py
```diff
@@ -1333,6 +1333,7 @@ def create_docker_build_script(script_name, container_install_dir,
 
         if target_platform() == 'windows':
             runargs += ['--memory', FLAGS.container_memory]
```
deadeyegoodwin, line 1335: There is another place in build.py where --memory is used.
Also, build.py has a --container-memory flag that is documented as setting the memory in the windows container. If we are not going to support that anymore then we need to remove the --container-memory flag.

deadeyegoodwin (changes requested):

build.py
mc-nv: @deadeyegoodwin  your advise is to extend that [baseargs](https://github.com/triton-inference-server/server/blob/main/build.py#L1303) list with `--isolation=process` argument? 

Regarding `--container-memory` flag I'm not capable to confirm that we are going to remove it.

deadeyegoodwin: I don't know what --isolation=process does. But if you are replacing --memory with it, then it seems like you would want to replace it everywhere. Right?
If after your changes we don't use --container-memory anymore, then we should remove it.

mc-nv: In my understanding docker is inherits LXC container and manage the process by [namespaces](https://man7.org/linux/man-pages/man7/namespaces.7.html) and [chroot](https://man7.org/linux/man-pages/man2/chroot.2.html) etc. Similar to LXC some process may compete for a process within docker engine and try to send specific instructions, to get network, file system etc , by sharing same socket we probably may get a situation when we braking that rule and we start seeing the situation where process trying to ocupate same resource in "namespace" unless isolation is enabled. 
I added some documentation in description for current [PR#4226](https://github.com/triton-inference-server/server/pull/4226)

---

This change won't remove `--container-memory` it will extend that call and appear as:
```
docker run -w /workspace/build --name tritonserver_builder --memory 8g 
--isolation=process -v \\.\pipe\docker_engine:\\.\pipe\docker_engine tritonserver_buildbase 
powershell.exe -noexit -File ./cmake_build.ps1 
```

deadeyegoodwin: I see, I misread the change, I thought you were replacing --memory with --isolation... but you are just adding it.

deadeyegoodwin (approved):

PR #4228, opened 2022-04-15
Add 'TF-TRT' and `ORT-TRT' keywords to docs for searchability

rmccorm4:
Having this keyword included should benefit users searching for Triton support of TF-TRT models. Currently it's only mentioned explicitly in a few CI test comments.


CoderHam (changes requested):
@rmccorm4 this is a good idea. Can you also repeat this for ORT-TRT (ONNX with TensorRT Optimization) ?

CoderHam (approved):

PR #4229, opened 2022-04-15
Fix minor typos introduced by https://github.com/triton-inference-ser…

CoderHam:
…ver/server/pull/4228

@ryanleary I missed this in my review. This should update the links for the TF-TRT and ORT-TRT sections since we modified their names.


rmccorm4 (approved):
Nice catch

PR #4230, opened 2022-04-16
Fix instance type to CPU for optional python model

CoderHam:
- Fixed jetson failure since GPU instances are not supported

Fixes L0_backend_python failure

### DIFF
--- a/qa/python_models/optional/config.pbtxt
+++ b/qa/python_models/optional/config.pbtxt
@@ -53,3 +53,10 @@ output [
     dims: [ 1 ]
   }
 ]
+
+instance_group [
+  {
+    count: 1
+    kind : KIND_CPU
+  }
+]


mc-nv (approved):

GuanLuo:
Why need to set CPU instance explicitly? @Tabrizian wouldn't Python backend create CPU instance if there is no supported GPU?

CoderHam:
The GPU support for the python backend refers to supporting the GPU tensors as I/O. This means that a model can run with CPU instance and still use GPU with python backend as long as the I/O for that model is via the CPU.

Tabrizian:
Using GPU/CPU instances doesn't change the behavior of Python backend at all. The instance settings will be passed to the model's initialize function and it is up to the user on how they want to handle this. The error is returned from the common backend utilities: https://github.com/triton-inference-server/core/blob/b338d65a97a9cfef9f68910f516556599f583057/src/backend_model_instance.cc#L220 Do we default to using the KIND_GPU if the user doesn't specify anything?

PR #4252, opened 2022-04-19
Cat SERVER_LOG when failing to grep it

rmccorm4:
Currently we cat the client log when failing to find something in server log output - this makes it hard to see what happened in CI test failure output.


nv-kmcgill53 (commented):
If it's useful to have both the client and the server log when troubleshooting then I would suggest leaving in the "$cat $MINIMAL_LOG" in. If that interaction is not useful for identifying what went wrong then the replacement of the client log with the server log is fine for me.

rmccorm4:
I think it's better to have more info than less when debugging, so I'll include both.

dyastremsky (approved):

PR #2, opened 2018-11-27
README links and text edits

tripti-singhal:
Added links to corresponding documentation in the feature list and minor text edits.

PR #4259, opened 2022-04-21
Add L0 memory growth test

jbkyang-nvi:


jbkyang-nvi:
based on @dyastremsky : https://github.com/triton-inference-server/server/pull/3827
https://github.com/triton-inference-server/server/compare/dyas-memory-growth...kyang-memory-growth
and @jack's PR: https://github.com/triton-inference-server/server/pull/4054

qa/L0_java_memory_growth/Simple.java
```diff
@@ -0,0 +1,847 @@
...
+
+        // Wait for the inference to complete.
+        TRITONSERVER_InferenceResponse completed_response = completed.get();
+        futures.remove(irequest);
+
+        FAIL_IF_ERR(
+            TRITONSERVER_InferenceResponseError(completed_response),
+            "response status");
+
+        Check(
```
dyastremsky, line 511: We should probably make Check only run based on a command line boolean. That way we can break up the test into a sanity test to make sure that the full Simple example has no memory leak (run for a small number of iterations) and a high-iteration memory leak test just for the inference. Right now, I suspect Check is likely taking up a lot of the time per inference, which I suspect we don't want. What do you think?

qa/L0_java_memory_growth/test.sh
```diff
@@ -0,0 +1,77 @@
...
+    /<artifactId>tensorrt-platform<\/artifactId>/ {d}
+}' $SAMPLES_REPO/pom.xml
+
+BASE_COMMAND="mvn clean compile -f $SAMPLES_REPO exec:java -Djavacpp.platform=linux-x86_64"
+source ../common/util.sh
+
+# Create local model repository
+rm -r models
+cp -r `pwd`/../L0_simple_ensemble/models .
+mkdir ${MODEL_REPO}/ensemble_add_sub_int32_int32_int32/1
```
dyastremsky, line 55: Let's either change the previous line to only copy the simple model from models, or make this line remove L0_simple_ensemble. It's not part of the test anymore, so we don't need to load it.

dyastremsky (commented):
Beautiful! This passed for 1M iterations for me with the memory usage bouncing around within a few decimal points of 17MB. Currently running it for 20M Iterations, which will likely run for a while.

Added two small comments. Let me know what you think. You can make the updates and I can review this ticket or I can do it and we can get someone else to review this test.

dyastremsky (approved):

dyastremsky:
Amazing work on the PR! Heads up that the test failed on 20M iterations due to a larger memory allocation range (it's saying a 7.7MB, 32% diff). There are two parts to this:

1. The memory oscillates in the 14-17MB range. The current 10% tolerance should be good, though we can increase this to 20% if we want to ensure there's no intermittence on this test.
2. Outliers. Out of 6,750 memory readings, 3 were higher (22-23MB). I'll look into whether DoubleSummaryStatistics can provide percentiles or get rid of outliers; else, whether there's a better way to track memory statistics.

dyastremsky:
> Amazing work on the PR! Heads up that the test failed on 20M iterations due to a larger memory allocation range (it's saying a 7.7MB, 32% diff). There are two parts to this:
> 
> 1. The memory oscillates in the 14-17MB range. The current 10% tolerance should be good, though we can increase this to 20% if we want to ensure there's no intermittence on this test.
> 2. Outliers. Out of 6,750 memory readings, 3 were higher (22-23MB). I'll look into whether DoubleSummaryStatistics can provide percentiles or get rid of outliers; if not, whether there's a better way to track memory statistics.

I don't see a way to fix #2 with DoubleSummaryStatistics. The ideal would be to use a percentile (e.g. comparing the 90th percentile value versus the median to get the difference), but that'd be hard to do without holding all our data in a data structure, which we don't want to do (especially for the long-running memory growth test). We need to figure out a way to identify and reject/ignore outliers.

PR #4280, opened 2022-04-25
Support model file override in HTTP/GRPC server. Add documentation. Add test

GuanLuo:


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2237,6 +2237,144 @@ def test_config_override(self):
...
+                        encoded_files={"file:1/model.onnx": encoded_file})
+                    self.assertTrue(
+                        False, "expected error on missing override config")
+                except InferenceServerException as ex:
+                    # [FIXME] Improve error reporting to mention missing config
+                    self.assertIn(
+                        "failed to load '{}', failed to poll from model repository"
+                        .format(model_name), ex.message())
+
+                # Sanity check on previous loaded version is still available
```
dyastremsky, line 2288: These comments are throwing me off a bit. So only one of the versions (3) should be available, right? After this, it looks like you load the version 1 and the original version is also available then... but 3 is not?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2237,6 +2237,144 @@ def test_config_override(self):
...
+                try:
+                    triton_client.load_model(
+                        model_name,
+                        config="""{{"backend":"{backend}" }}""".format(
+                            backend=base[1]),
+                        encoded_files={"file:1/model.onnx": encoded_file})
+                except Exception as ex:
+                    self.assertTrue(False, "unexpected error {}".format(ex))
+
+                # Sanity check on previous loaded version is still available
```
dyastremsky, line 2340: I think each of these comments may need to be differentiated. Like here, it looks like you're expecting the load to be successful, which is why 1 is available, right? You're not just checking that the previous versions are still available. Also, is version 3 no longer available because we only allow one version at a time? If so, I wonder if it's necessary to check all of them. If you do so for completeness, you may want to find a way to make it clear what you're specifically checking (since the code ends up being broader, comments may be more important).

src/http_server.cc
```diff
@@ -1305,7 +1305,20 @@ HTTPAPIServer::HandleRepositoryControl(
                        "unexpected error getting load model request buffers"));
         }
       }
-      std::vector<const TRITONSERVER_Parameter*> params;
+      // [FIXME] release parameter
```
rmccorm4, line 1308: is this comment still needed?

rmccorm4 (changes requested):
Also, can we add some kind of documention / motivation for this new b64 encoding feature in https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md or similar?

GuanLuo:
The b64 decoding added here is for internal use and not accessible to user, so I don't want to document it. If the user wants such feature, the proper way is to use the repository agent which allows highly customized file handling across the model lifecycle.

qa/L0_lifecycle/lifecycle_test.py
GuanLuo: The load above it fails (file override requires config to be provided as well), so the model availability hasn't changed and version 3 should still be ready.

qa/L0_lifecycle/lifecycle_test.py
GuanLuo: Yes, the comment should be updated. The reason why version 3 is unavailable is because the override model directory (a new and different model directory) will be used, where it only has version 1.

rmccorm4 (approved):

tanmayv25 (approved):

PR #4289, opened 2022-04-26
Move Java API Prerelease Into Main

dyastremsky:
Move the Java API from its own prerelease branch to the main branch.


qa/L0_java_simple_example/test.sh
```diff
@@ -0,0 +1,170 @@
...
+        RET=1
+    fi
+
+    # Run with FP32 datatype
+    sed -i 's/TYPE_INT32/TYPE_FP32/g' $MODEL_REPO/simple/config.pbtxt
+    sed -i 's/TYPE_INT32/TYPE_FP32/g' $MODEL_REPO/ensemble_add_sub_int32_int32_int32/config.pbtxt
+    $BASE_COMMAND -Dexec.args="-r $MODEL_REPO -v" >>$CLIENT_LOG 2>&1
+    if [ $? -ne 0 ]; then
+        echo -e "Failed to run: ${BASE_COMMAND} -Dexec.args=\"-r ${MODEL_REPO} -v\""
+        RET=1
```
rmccorm4, line 96: IIUC, it looks like you're editing the config.pbtxt file to replace INT32 -> FP32 for the [simple](https://github.com/triton-inference-server/server/tree/main/docs/examples/model_repository/simple) model, but the actual model file itself for simple expects INT32, so I'm not sure this will work. 

image

GuanLuo (dismissed):
Approving to unblock the merge. But should create follow-up ticket to enhance the Java test framework. Currently it feels like we are creating different Java applications for different uses, should be more structure and see if we can re-use the common setup code.

qa/L0_java_resnet/test.sh
```diff
@@ -40,14 +40,21 @@ fi
...
 
 # Create local model repository
-mkdir -p models/
-cp -r $DATADIR/perf_model_store/resnet50* models/
+mkdir -p ${MODEL_REPO}
+# TODO: fix build to support GPU only resnet50v1.5_fp16_savedmodel
+for BACKEND in _fp32_libtorch _fp32_onnx; do
+    cp -r $DATADIR/perf_model_store/resnet50${BACKEND} ${MODEL_REPO}/
+    echo ${MODEL_REPO}/resnet50${BACKEND}/config.pbtxt
+    sed -i "s/kind: KIND_GPU/kind: KIND_CPU/" ${MODEL_REPO}/resnet50${BACKEND}/config.pbtxt
```
GuanLuo, line 51: I thought the models can be deployed on GPU? This suggests that is not the case? If it can and you want the test can also be run on cpu-only environment, you can change it to KIND_AUTO and Triton should assign the proper instance kind for you

GuanLuo (dismissed):

GuanLuo (approved):

PR #4290, opened 2022-04-26
Add test for decoupled API in Python backend

Tabrizian:
Testing for the decoupled API: https://github.com/triton-inference-server/python_backend/pull/136


CoderHam:
@Tabrizian can you run these tests on the Jetson runner before merging?

Tabrizian:
Will verify on jetson before merging.

CoderHam:
No, you will need to run them manually. Lets sync up offline

qa/L0_backend_python/decoupled/decoupled_test.py
```diff
@@ -0,0 +1,145 @@
...
+                                      np_to_triton_dtype(input_data_1.dtype))
+            ]
+            inputs[0].set_data_from_numpy(input_data_0)
+            inputs[1].set_data_from_numpy(input_data_1)
+            client.async_stream_infer(model_name=model_name, inputs=inputs)
+            data_item = user_data._completed_requests.get()
+            if type(data_item) == InferenceServerException:
+                self.assertEqual(
+                    data_item.message(),
+                    "Python model 'decoupled_return_response_error_0' is using the decoupled mode and the execute function must return None.",
```
CoderHam, line 115: Format long line

qa/L0_decoupled/test.sh
```diff
@@ -48,78 +48,91 @@ rm -f *.log
 
 CLIENT_LOG=`pwd`/client.log
 DATADIR=/data/inferenceserver/${REPO_VERSION}/qa_model_repository
-MODELDIR=`pwd`/models
```
CoderHam, line 51: Update copyright

CoderHam (changes requested):
Minor changes else LGTM

qa/L0_decoupled/test.sh
```diff
@@ -48,78 +48,91 @@ rm -f *.log
...
+
+  run_server
+  if [ "$SERVER_PID" == "0" ]; then
+      echo -e "\n***\n*** Failed to start $SERVER\n***"
+      cat $SERVER_LOG
+      exit 1
+  fi
+  
+  for i in \
+              test_one_to_none \
```
tanmayv25, line 79: Doesn't it make sense to add shared memory leak detector check on these tests?

qa/L0_backend_python/decoupled/decoupled_test.py
```diff
@@ -0,0 +1,145 @@
...
+def callback(user_data, result, error):
+    if error:
+        user_data._completed_requests.put(error)
+    else:
+        user_data._completed_requests.put(result)
+
+
+class DecoupledTest(tu.TestResultCollector):
+
+    def test_decoupled_execute_error(self):
```
tanmayv25, line 56: We must be using shared memory leak detector in these tests right?

qa/L0_decoupled/test.sh
tanmayv25: I wonder what is the best way to detect memory leak is though.

Tabrizian: I can give it a try and if it took a lot of time. I'll do it separately.

Tabrizian: It looks like there is not any shared memory growth issues but the shared memory leak detector fails because of some underlying object that is in scope (perhaps by the response threads) that makes it fail to match the exact value. I logged the free shared memory values as the test ran and it doesn't look like there is any leak:
image

I think we might need to change our leak detection tests to be a bit more forgiving. @GuanLuo do you think it makes sense to add shared memory usage monitoring to the long running tests so that we can detect any unbounded growth?

tanmayv25 (approved):

CoderHam (approved):

tanmayv25 (approved):

PR #4291, opened 2022-04-26
Fix instance type to CPU for optional python model (#4230)

CoderHam:
- Fixed jetson failure since GPU instances are not supported


CoderHam:
Cherry pick to fix L0_backend_python failure on jetpack

mc-nv (approved):

PR #4294, opened 2022-04-26
Removing default memory limits for Windows container

mc-nv:


CoderHam (approved):

CoderHam:
@mc-nv would we want to document the same ?

deadeyegoodwin (approved):

mc-nv:
> 
It has a description in `--help` menu for `build.py`, but it doesn't explain that it's a non stable flag and windows container doesn't support shared memory according to [documentation](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/hyperv-container#piercing-the-isolation-boundary)

PR #4295, opened 2022-04-26
Add TYPE_BF16 scaffolding

rmccorm4:


qa/L0_backend_identity/identity_test.py
```diff
@@ -173,18 +173,23 @@
...
             ("identity_nobatch_int8", np.int8, [7]),
-            ("identity_bytes", object, [1, 1])):
+            ("identity_bytes", object, [1, 1]),
+            ("identity_bf16", "BF16", [1, 1])):
             # yapf: enable
             if np_dtype != object:
                 input_data = (16384 * np.random.randn(*shape)).astype(np_dtype)
+            # FIXME: Add proper bfloat16 support instead of fp16 WAR here
+            elif np_dtype == "BF16":
+                input_data = (16384 * np.random.randn(*shape)).astype(np.float16)
```
rmccorm4, line 183: @jbkyang-nvi this conversion for triton<->np type is a temporary WAR to test identity backend with existing L0_backend_identity. Once proper python client support for bf16 is added, you should tweak this.

GuanLuo: This WAR probably won't work without modifying the client library to specially handle BF16?

rmccorm4: I'll see if the WAR can solely be done on test side without dummy implementation in client

GuanLuo (approved):
Should turn identity_test.py to actual unit test when we have the chance.

qa/L0_backend_identity/identity_test.py
```diff
@@ -231,3 +263,8 @@
             if param2 != False:
                 print("error: expected 'param2' == False")
                 sys.exit(1)
+
+    # FIXME: Use identity_bf16 model in test above once proper python client
```
tanmayv25, line 267: It makes sense to include the ticket id here.

tanmayv25 (approved):

azraelkuan:
hi, when build with docker, the bf16 will raise a error:
image

tanmayv25:
@azraelkuan Are you using correct branch and using repo tags as per your requirement? You must use the consistent release specific tags for these repos to get passing  build. If building on main, try cleaning up your build artifacts and build from the scratch such that this change gets picked up from **core:main** https://github.com/triton-inference-server/core/pull/66/files.

azraelkuan:
@tanmayv25 hi, i am building on 22.04 branch.

PR #4307, opened 2022-04-29
Updating gpg keys; cmake installation streamlined

borisfom:
1. Fixed container build failure : Updated GPG keys according to: 
https://github.com/NVIDIA/cuda-repo-management/issues/4

2. Changed cmake install to pip-based one. In some base containers (notably upcoming CUDA 11.8), existing wget install was successful, but old cmake still in the path. Install via pip is simpler and works everywhere. 

Signed-off-by: Boris Fomitchev <bfomitchev@nvidia.com>

### DIFF
--- a/build.py
+++ b/build.py
@@ -821,7 +821,7 @@ ENV DCGM_VERSION {}
 # Install DCGM. Steps from https://developer.nvidia.com/dcgm#Downloads
 RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/sbsa/cuda-ubuntu2004.pin && \
     mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600 && \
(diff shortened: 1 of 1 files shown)


deadeyegoodwin:
Note this slack thread: https://nvidia.slack.com/archives/C8H1F7HAM/p1651252005561269
There are some changes going on with how cmake is installed in the base image. Using pip to install cmake instead of apt is at best a strange thing to do. Why does our current way of installing cmake with apt need to change just because using cuda 11.8? Why would using CUDA 11.8 have anything to do with how cmake is installed or used?

rmccorm4:
@borisfom any updates based on David's feedback above?

rmccorm4:
Closing due to lack of activity and the key-rotation issues have been fixed. New CUDA version related issues have likely been fixed by this point as well if there were any.

PR #4315, opened 2022-04-29
Add testing for PyTorch string I/O

CoderHam:


qa/L0_infer/infer_test.py
```diff
@@ -100,27 +100,31 @@ def _infer_exact_helper(tester,
...
-                    use_grpc=use_grpc,
-                    use_http_json_tensors=use_http_json_tensors,
-                    skip_request_id_check=skip_request_id_check,
-                    use_streaming=use_streaming,
-                    correlation_id=correlation_id,
-                    use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
-                    use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY)
+
+                # model that supports batching. Skip for libtorch string I/O
+                if pf == 'libtorch' and tu.validate_for_libtorch_model(
```
GuanLuo, line 105: Has been checked below (search `validate_for_libtorch_model` within the file)

GuanLuo: Should probably add a new case below to handle Torch string model explicitly, the Torch case just doesn't fit into this helper function.

CoderHam: We need to check again with the batch size since string I/O does not support batching.

While the torch case does not fit directly into the helper function it still avoid redundant code. I do not think taking it out into a different helper function is wise.

qa/common/gen_qa_sequence_models.py
```diff
@@ -1108,10 +1108,14 @@ def create_libtorch_modelfile(models_dir, model_version, max_batch, dtype,
...
     if not tu.validate_for_libtorch_model(dtype, dtype, dtype, shape, shape,
-                                          shape):
+                                          shape, max_batch):
         return
 
+    # Skip for String I/O
     torch_dtype = np_to_torch_dtype(dtype)
+    if torch_dtype is None:
+        return
+
```
GuanLuo, line 1118: You can check type and skip model / config generation at where these two functions get called in all the gen files, i.e. ~line 1402, `if FLAGS.libtorch and xxx:`

qa/L0_infer/infer_test.py
```diff
@@ -100,27 +100,31 @@ def _infer_exact_helper(tester,
...
-                    skip_request_id_check=skip_request_id_check,
-                    use_streaming=use_streaming,
-                    correlation_id=correlation_id,
-                    use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
-                    use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY)
+
+                # model that supports batching. Skip for libtorch string I/O
+                elif pf == 'libtorch' and tu.validate_for_libtorch_model(
+                        input_dtype, output0_dtype, output1_dtype, tensor_shape,
+                        tensor_shape, tensor_shape, bs):
```
GuanLuo, line 107: This evaluates to only run below for supported Torch model, all other backends will be skipped. Why you prefer changing the common utility instead of adding the Torch string infer ~line 200?

qa/L0_infer_variable/infer_variable_test.py
```diff
@@ -94,26 +94,30 @@ def _infer_exact_helper(tester,
...
-                    use_http=use_http,
-                    use_grpc=use_grpc,
-                    skip_request_id_check=skip_request_id_check,
-                    use_streaming=use_streaming,
-                    correlation_id=correlation_id,
-                    use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
-                    use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY)
+
+                # model that supports batching. Skip for libtorch string I/O
+                elif pf == 'libtorch' and tu.validate_for_libtorch_model(
```
GuanLuo, line 99: Same argument as above

qa/common/gen_qa_sequence_models.py
CoderHam: Done

qa/L0_infer/infer_test.py
CoderHam: Done

GuanLuo (approved):

PR #4316, opened 2022-04-29
Fix test failures on Jetpack

CoderHam:
Fix L0_perf_resnet and L0_backend_python failures

### DIFF
--- a/qa/L0_perf_resnet/test.sh
+++ b/qa/L0_perf_resnet/test.sh
@@ -67,9 +67,10 @@ INSTANCE_CNT=1
 CONCURRENCY=1
 
 # Disable TF-TRT test on Jetson due to Segfault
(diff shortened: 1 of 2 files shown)


Tabrizian (approved):

qa/L0_perf_resnet/test.sh
```diff
@@ -67,9 +67,10 @@ INSTANCE_CNT=1
 CONCURRENCY=1
 
 # Disable TF-TRT test on Jetson due to Segfault
+# Disable ORT-TRT test on Jetson due to support being disabled
```
Tabrizian, line 70: Why is the support disabled? Is there any plans to enable it in the future?

CoderHam: It was disabled in 22.03. We're waiting on Microsoft to confirm the same

PR #4318, opened 2022-04-29
Add test for ensemble config contains backend

GuanLuo:


tanmayv25 (approved):

qa/L0_model_config/autofill_noplatform/ensemble/has_backend/expected
```diff
@@ -0,0 +1 @@
+Ensemble model 'has_backend' must have platform type 'ensemble' and empty backend type
```
tanmayv25, line 1: add a new line here.

GuanLuo: The "expected" files don't have a new line, not sure if `grep` will still be correct after adding it.

tanmayv25: Is this true for all.  I wonder why I didn't come to this issue before.

tanmayv25 (approved):

PR #4320, opened 2022-05-02
Update NGC versions post-22.04 release

mc-nv:


deploy/gcp/values.yaml
```diff
@@ -27,7 +27,7 @@
 replicaCount: 1
 
 image:
-  imageName: nvcr.io/nvidia/tritonserver:22.02-py3
+  imageName: nvcr.io/nvidia/tritonserver:22.04-py3
```
CoderHam, line 30: Do we always advance this @dzier ?

Was this not advanced for 22.03 intentionally?

build.py
```diff
@@ -1121,8 +1121,9 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
 COPY --from=min_container /usr/lib/{libs_arch}-linux-gnu/libnccl.so.2 /usr/lib/{libs_arch}-linux-gnu/libnccl.so.2
 COPY --from=min_container /usr/lib/{libs_arch}-linux-gnu/libcudnn.so.8 /usr/lib/{libs_arch}-linux-gnu/libcudnn.so.8
 
+# patchelf is needed to add deps of libcublasLt.so.11 to libtorch_cuda.so
```
CoderHam, line 1124: Looks like the rebase was not done correctly

You seem to be missing some lines here: https://github.com/triton-inference-server/server/blob/r22.04/build.py#L996-L1001

Please double check if other changes have been missed as well.

CoderHam (changes requested):

deploy/gcp/values.yaml
mc-nv: The `values.yaml` is used by HELM chart and designed to pull proper image from repository during the object deployment in Kubernetes. I assume it must be up to date with our development version.

build.py
mc-nv: Hi @CoderHam  

The issue came up because this commit doesn't have correct parent version in `main` branch and during the commit it's conflicts as following: 
```
<<<<<<< HEAD
=======
    for noncore in NONCORE_BACKENDS:
        if noncore in backends:
            df += '''
COPY --chown=1000:1000 --from=tritonserver_build {0}/install/backends backends
'''.format(build_dir)
            break

    # This is required since libcublasLt.so is not present during the build
    # stage of the PyTorch backend
    if not FLAGS.enable_gpu and ('pytorch' in backends):
        df += '''
RUN patchelf --add-needed /usr/local/cuda/lib64/stubs/libcublasLt.so.11 backends/pytorch/libtorch_cuda.so
'''

    if len(repoagents) > 0:
        df += '''
COPY --chown=1000:1000 --from=tritonserver_build {0}/install/repoagents repoagents
'''.format(build_dir)

    mkdir(ddir)
>>>>>>> c213be65... Fix CPU only failure for pytorch backend (#4227)
```
But because we no longer using NONCORE_BACKENDS in our `build.py` it generates build failure.

deploy/gcp/values.yaml
CoderHam: Update the copyright for this file

CoderHam (changes requested):
LGTM. Check for a clean CI to make sure these changes work after resolving merge conflicts before merging into main

CoderHam (approved):

PR #4322, opened 2022-05-03
Add testing for python model api for auto-complete

krishung5:


Tabrizian (changes requested):
Can you add a test for using the auto_complete_config with Python backend's Execution environments as well? Wanted to make sure the execution environments are properly used too.

Also, can you add an additional test in L0_backend_python that tests if an error is raised during `auto_complete_config`, we will be properly returning the error to the user? We have similar tests in https://github.com/triton-inference-server/server/blob/main/qa/python_models/wrong_model/model.py for the execute function.

qa/L0_model_config/autofill_platform/python/mismatch_datatype/1/model.py
```diff
@@ -0,0 +1,180 @@
...
+          * model_repository: Model repository path
+          * model_version: Model version
+          * model_name: Model name
+        """
+
+        # You must parse model_config. JSON string is not parsed here
+        # self.model_config = model_config = json.loads(args['model_config'])
+        self.model_config = model_config = json.loads('{"name":"add_sub","platform":"","backend":"python","version_policy":{"latest":{"num_versions":1}},"max_batch_size":0,"input":[],"output":[{"name":"OUTPUT0","data_type":"TYPE_FP32","dims":[4],"label_filename":"","is_shape_tensor":false},{"name":"OUTPUT1","data_type":"TYPE_FP32","dims":[4],"label_filename":"","is_shape_tensor":false}],"batch_input":[],"batch_output":[],"optimization":{"priority":"PRIORITY_DEFAULT","input_pinned_memory":{"enable":true},"output_pinned_memory":{"enable":true},"gather_kernel_buffer_threshold":0,"eager_batching":false},"instance_group":[{"name":"add_sub_0","kind":"KIND_CPU","count":1,"gpus":[],"secondary_devices":[],"profile":[],"passive":false,"host_policy":""}],"default_model_filename":"","cc_model_filenames":{},"metric_tags":{},"parameters":{},"model_warmup":[]}')
```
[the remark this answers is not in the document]
krishung5: It's added by mistake. Will remove the line.

qa/L0_backend_python/examples/test.sh
```diff
@@ -240,6 +240,36 @@ if [ "$TEST_JETSON" == "0" ]; then
     wait $SERVER_PID
 fi
 
+# Example 4
```
tanmayv25, line 243: Instead of Example number just call it "Auto Complete Model Configuration Example"

qa/python_models/auto_complete_error/model.py
```diff
@@ -0,0 +1,59 @@
...
+    @staticmethod
+    def auto_complete_config(model_config):
+        """
+        The body of this model doesn't matter. The main purpose of this model is
+        to test correct handling of Python errors in the `auto_complete_config`
+        function.
+        """
+        auto_complete_model_config = pb_utils.ModelConfig()
+
+        input0 = {'name': 'INPUT0', 'data_type': 'TYPE_FP32', 'dims': [16]}
```
tanmayv25, line 43: For better testing, specially in case of incomplete input/output in L0_model_config above, why not add a model that just provides those missing pieces in  auto_complete_config?

qa/L0_model_config/test.sh
```diff
@@ -182,6 +182,19 @@ for modelpath in \
...
 done
 
+# Copy Python models into the test model repositories.
+for modelpath in \
+        autofill_platform/python/mismatch_datatype/1 \
+        autofill_platform/python/mismatch_dims/1 \
+        autofill_platform/python/unknown_input/1 \
+        autofill_platform/python/unknown_output/1 \
+        autofill_platform_success/python/empty_config/1 \
+        autofill_platform_success/python/incomplete_input/1 \
```
tanmayv25, line 192: I meant these two here. You can do something like:

incomplete_input : <incomplete_config> + <full input/ouptut config in auto_complete_config> == This will be like you have now.

incomplete_output:  <incomplete_config> + <just the missing config in auto_complete_config>

tanmayv25 (changes requested):

qa/L0_model_config/test.sh
krishung5: Please correct me if I'm wrong, do you mean that, for example, if INPUT0 is missing dims property, then in model.py we just provide dims property for INPUT0? But our API has a checking to make sure that users add all name, datatype and dims properties when setting the configuration, we cannot only provide a specific property for a input/output.

krishung5: If we provide only INPUT0 in config.pbtxt and INPUT1 in model.py, I don't think this will pass the logic for checking allowed input/output name here: https://github.com/triton-inference-server/python_backend/blob/krish-python/src/python.cc#L2506
Do we want to remove this checking?

tanmayv25: @krishung5  I see. Let's keep the check for now for consistency with other backends.  We can relax it in future if required.

@krishung5 @Tabrizian  Do you think it will be a value add for users to not specify every input in `auto_complete_config` model.py?
Unlike other models. the model.py execution will not be affected by extra inputs/outputs in request/response.

qa/L0_model_config/autofill_platform_success/python/empty_config/config.pbtxt
```diff
@@ -0,0 +1 @@
+backend: "python"
```
tanmayv25, line 1: You can truly have empty config.pbtxt with Iman's change here: https://github.com/triton-inference-server/python_backend/pull/138#discussion_r868257846

qa/python_models/auto_complete/config.pbtxt
```diff
@@ -0,0 +1,28 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "auto_complete"
+backend: "python"
```
tanmayv25, line 28: Remove the `backend: "python"` ?
refer: https://github.com/triton-inference-server/python_backend/pull/138#discussion_r868257846

qa/L0_backend_python/examples/test.sh
```diff
@@ -97,7 +97,7 @@ fi
 
 grep "PASS" $CLIENT_LOG
 if [ $? -ne 0 ]; then
-    echo -e "\n***\n*** Failed to verify pytorch example. \n***"
```
tanmayv25, line 100: This is for pytorch example. Please revert this change.

tanmayv25 (changes requested):

qa/L0_model_config/test.sh
krishung5: Do you mean allowing users to add some extra inputs/outputs that are not used in the model.py execution? So if they use `curl localhost:8000/v2/models/<model name>/config`, they will see the extra inputs/outputs in the configuration? I think this should be fine since it won't affect any request/response like you mentioned, as long as this won't confuse users at any point?

Tabrizian: I think after the discussion in [here](https://github.com/triton-inference-server/python_backend/pull/138#discussion_r868251707) it is allowed to have unknown inputs since it will be auto-completed by the backend.

Tabrizian (changes requested):
We would also need a test for returning None from `auto_complete_config` which is not allowed and an error should be returned.

qa/L0_model_config/autofill_noplatform/python/input_wrong_property/expected
```diff
@@ -0,0 +1 @@
+input 'INPUT1' contains property other than 'name', 'data_type' and 'dims'.
```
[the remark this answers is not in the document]
tanmayv25: Make similar changes for the output.

tanmayv25 (changes requested):

qa/L0_model_config/autofill_noplatform/python/conflicting_max_batch_size/expected
```diff
@@ -0,0 +1 @@
+configuration specified max_batch_size 2 but Python model specified max_batch_size 4
```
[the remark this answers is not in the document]
krishung5: Referring to the comment here: https://github.com/triton-inference-server/python_backend/pull/138#discussion_r870840585 I was under the impression that if max_batch_size has a confliction, we should throw an error. Should I change the logic so that it won't throw an error when the max_batch_size specified in config.pbtxt is less than the max_batch_size specified in model.py?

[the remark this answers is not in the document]
tanmayv25: Yes. I think the user might want to tune this value separately. And we must not throw error if batch_size is within max_batch_size provided by the model.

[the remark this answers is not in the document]
krishung5: I see. Will modify the logic, thanks.

tanmayv25 (approved):

qa/L0_backend_python/examples/test.sh
```diff
@@ -240,6 +240,41 @@ if [ "$TEST_JETSON" == "0" ]; then
     wait $SERVER_PID
 fi
 
+# Auto Complete Model Configuration Example
+CLIENT_LOG="./auto_complete_client.log"
+mkdir -p models/nobatch_auto_complete/1/
+mkdir -p models/batch_auto_complete/1/
+cp examples/auto_complete/nobatch_model.py models/nobatch_auto_complete/1/model.py
+cp examples/auto_complete/nobatch_config.pbtxt models/nobatch_auto_complete/config.pbtxt
```
Tabrizian, line 248: Remove config.pbtxt

qa/L0_backend_python/examples/test.sh
```diff
@@ -240,6 +240,41 @@ if [ "$TEST_JETSON" == "0" ]; then
...
+cp examples/auto_complete/batch_model.py models/batch_auto_complete/1/model.py
+cp examples/auto_complete/batch_config.pbtxt models/batch_auto_complete/config.pbtxt
```
Tabrizian, line 250: Remove config.pbtxt

qa/python_models/auto_complete/config.pbtxt
```diff
@@ -0,0 +1,27 @@
+# Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 1: Remove this file

Tabrizian:
Looks like the auto_complete for env model includes model configuration with the same inputs too. Does the model reload properly? I think we need to remove input/output description from here otherwise it will error out because of duplicate input/outputs. https://github.com/triton-inference-server/server/blob/main/qa/python_models/python_version/config.pbtxt#L30-L43

krishung5:
> Looks like the auto_complete for env model includes model configuration with the same input/output too. Does the model reload properly? I think we need to remove input/output description from here otherwise it will error out because of duplicate input/outputs. https://github.com/triton-inference-server/server/blob/main/qa/python_models/python_version/config.pbtxt#L30-L43

It only raises error when the properties mismatch. This doesn't error out because the input/output has the same properties. But I think we can still remove the input/output from the config file to have a clearer model configuration.

Tabrizian (approved):

krishung5:
Branch was rebased. Please re-approve, thanks!

Tabrizian (approved):

PR #4323, opened 2022-05-03
Moved TF Tests In L0_model_config From Success to Failure

nv-kmcgill53:
The following tests in L0_model_config were moved from success to known failure:
- bad_input_dims
- bad_input_type
- bad_output_dims
- bad_output_type
- too_many_inputs
- unknown_input
- unknown_output

The reason for this refactor is tensorflow now uses the config.pbtxt as hints during autoconfiguration rather than ignoring it altogether and creating a config from the model itself.


qa/L0_model_config/autofill_noplatform/tensorflow_savedmodel/too_many_inputs/expected
```diff
@@ -0,0 +1 @@
+Config file specifies too many inputs, 3, while loaded model 'too_many_inputs' specifies 2
```
tanmayv25, line 1: Some of these error messages will change with my requested change here: https://github.com/triton-inference-server/tensorflow_backend/pull/58
We would like to return consistent error message back to the user, whether or not using auto-complete logic.

qa/L0_model_config/autofill_noplatform_success/tensorflow_savedmodel/reshape_config_provided/expected.3
```diff
@@ -16,9 +16,12 @@ input {
 input {
   name: "INPUT1"
   data_type: TYPE_FP32
-  dims: 4
-  dims: 1
-  dims: 2
+  dims: 8
```
tanmayv25, line 19: This is good!

nv-kmcgill53: thanks!

qa/L0_model_config/autofill_noplatform/tensorflow_savedmodel/too_many_inputs/expected
nv-kmcgill53: Updated the expected messages after the changes made to the tensorflow backend.

tanmayv25:
Adding @GuanLuo  for visibility.
See as we are going to enable the autofiller by default, we will not be disregarding the input/output config provided to the model. We depend upon the input/output config for supporting various cases(unknown rank, ragged batching) which failed when using autofiller. Instead of ignoring incorrect config within config.pbtxt, we will throw above errors.
This is a change in the current behavior and we need the whatever config provided by the user to be correct so our deductions are correct.

GuanLuo (approved):
> We depend upon the input/output config for supporting various cases(unknown rank, ragged batching) which failed when using autofiller.

@tanmayv25 is there example of those tests before / after the change?

tanmayv25:
@nv-kmcgill53 Do we have these tests right now? I believe they were L0_tf_unknown_rank and something similar

nv-kmcgill53:
I'm trying to understand what that test might look like. I don't think we have the tests which @GuanLuo mentioned we should have. These would be good tests to have though, proving a config file throws an error when `--strict-model-config=-true` and succeeds when `--strict-model-config=false`. Then again, I may have misunderstood the ask here.

tanmayv25:
> The reason for this refactor is tensorflow now uses the config.pbtxt as hints during autoconfiguration rather than ignoring it altogether and creating a config from the model itself.

I think the ask here is with this refactor what tests are we trying to fix?

tanmayv25 (approved):

nv-kmcgill53:
I see. The test which is getting fixed with the associated backend changes is "tf contrib" and this PR is asserting the new behavior of the tf backend.

PR #4326, opened 2022-05-03
Fix L0_lifecycle test when autocompleteing models

nv-kmcgill53:
--strict-model-config=false is going to be the new default. This PR fixes a test which relies on --strict-model-config=true being set.

### DIFF
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -450,7 +450,7 @@ done
 rm models/graphdef_float32_float32_float32/config.pbtxt
 
 SERVER_ARGS="--model-repository=`pwd`/models --model-repository=`pwd`/models_0 \
-             --exit-on-error=false --exit-timeout-secs=5"
+             --exit-on-error=false --exit-timeout-secs=5 --strict-model-config=true"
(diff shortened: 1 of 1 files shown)


qa/L0_lifecycle/test.sh
```diff
@@ -450,7 +450,7 @@ done
 rm models/graphdef_float32_float32_float32/config.pbtxt
 
 SERVER_ARGS="--model-repository=`pwd`/models --model-repository=`pwd`/models_0 \
-             --exit-on-error=false --exit-timeout-secs=5"
+             --exit-on-error=false --exit-timeout-secs=5 --strict-model-config=true"
```
tanmayv25, line 453: Why auto-complete needs to be disabled for this test? If there is some credible reason behind it then we must add it as a comment on top of it.

nv-kmcgill53: Added comment in the test as to why --strict-model-config=true was added.

GuanLuo: The test was run with the assumption that `--strict-model-config=true` (previous default behavior), so we need to explicitly set it if we move the default to false

tanmayv25 (approved):

PR #4338, opened 2022-05-05
Improve GRPC load model with file override to accept file binary

GuanLuo:


tanmayv25 (dismissed):

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2252,7 +2252,7 @@ def test_file_override(self):
             # Prepare override file
             with open("models/{}/3/model.{}".format(model_name, base[0]),
                       'rb') as f:
-                encoded_file = base64.b64encode(f.read())
+                file_content = f.read()
```
tanmayv25, line 2255: Should add C++ client library tests as well.

src/http_server.cc
```diff
@@ -1353,8 +1354,24 @@ HTTPAPIServer::HandleRepositoryControl(
...
+                  m.c_str(), TRITONSERVER_PARAMETER_STRING, param_str);
+            } else if (m.rfind("file:", 0) == 0) {
+              // Decode base64
+              base64_decodestate s;
+              base64_init_decodestate(&s);
+
+              // The decoded can not be larger than the input...
+              binary_files.emplace_back(std::vector<char>(param_len + 1));
+              size_t decoded_size = base64_decode_block(
+                  param_str, param_len, binary_files.back().data(), &s);
```
rmccorm4, line 1370: Should this still call `base64_decode_block` ? I noticed in `lifecycle_test.py` changes you removed the b64 encoding :
```python
# before
encoded_file = base64.b64encode(f.read())
# after
file_content = f.read()
```

But I see the client library is doing some B64 encoding, so maybe that's OK

src/http_server.cc
```diff
@@ -1353,8 +1354,24 @@ HTTPAPIServer::HandleRepositoryControl(
...
+            if (m == "config") {
+              param = TRITONSERVER_ParameterNew(
+                  m.c_str(), TRITONSERVER_PARAMETER_STRING, param_str);
+            } else if (m.rfind("file:", 0) == 0) {
+              // Decode base64
+              base64_decodestate s;
+              base64_init_decodestate(&s);
+
+              // The decoded can not be larger than the input...
+              binary_files.emplace_back(std::vector<char>(param_len + 1));
```
rmccorm4, line 1368: Does a `binary_files` list of vectors need to be maintained here or used later? Seems like just using a temp vector for each iteration would do the same currently.

rmccorm4 (changes requested):

src/http_server.cc
GuanLuo: For HTTP the encoding is still required to transfer via JSON, the client side changes now handles the encoding implicitly for ease of use. If you are forming HTTP request outside the client library, the encoding will need to be added.

src/http_server.cc
GuanLuo: The decoded content is stored in `binary_files` and it needs to be valid until `LoadModelWithParameter` returns

src/http_server.cc
rmccorm4: > If you are forming HTTP request outside the client library, the encoding will need to be added.

Is this documented anywhere, even if just a code comment? IIRC you mentioned this was an internal feature, not expected for public users.

GuanLuo: It is documented in the [protocol](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_model_repository.md#load) that file content sent via HTTP load API must be base64 encoded, so the users should follow that if they are forming the HTTP on their own. Client library is wrapper over the protocol and it will take care of the encoding for the user.

rmccorm4 (approved):

tanmayv25 (approved):

PR #4355, opened 2022-05-09
fix typo in command test.md

jqueguiner:


GuanLuo:
Thanks for the contribution, but it appears that -v`host_path`:`docker_path` is also valid in docker. Are you seeing error returned with the current command?

jqueguiner:
my bad. sorry

PR #4360, opened 2022-05-10
Adding GPG fix

mc-nv:
This is temporary fix. 
Required because we are using public image for `pytorch/torchserve` which is quiet outdated and doesn't have nightly builds or container with proper fix. 

https://hub.docker.com/r/pytorch/torchserve/tags

### DIFF
--- a/qa/L0_perf_ts/test.sh
+++ b/qa/L0_perf_ts/test.sh
@@ -38,6 +38,11 @@ if [ ! -z "$TEST_REPO_ARCH" ]; then
     REPO_VERSION=${REPO_VERSION}_${TEST_REPO_ARCH}
 fi
 
+# TODO: DLIS-3777 following key update is required only while base image 
+#       is not updated accordingly
+apt-key del 7fa2af80
+apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/$(uname -m)/3bf863cc.pub
+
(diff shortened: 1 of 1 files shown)


qa/L0_perf_ts/test.sh
```diff
@@ -38,6 +38,9 @@ if [ ! -z "$TEST_REPO_ARCH" ]; then
     REPO_VERSION=${REPO_VERSION}_${TEST_REPO_ARCH}
 fi
 
+sudo apt-key del 7fa2af80
```
CoderHam, line 41: We should not need to use sudo here. Also add a comment why this is needed, and file a follow up ticket to remove this when possible.

CoderHam (changes requested):

qa/L0_perf_ts/test.sh
mc-nv: Thanks,I moved away `sudo` call on given statement. 
For the issue you mentioned, should it be public?

tanmayv25: Just mention `DLIS-<ticket-id>` in the comment. It will be fine.

mc-nv: Thank you, - proposed changes were added

CoderHam (approved):

tanmayv25 (approved):

PR #4370, opened 2022-05-12
Add copyrights to test data for Java resnet tests

jbkyang-nvi:


mc-nv:
@jbkyang-nvi  

Using `set +e` rather than `set -` causes these options to be turned off according to [documentation](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html). 

In this pull request it looks like you trying to use it in opposite, wondering if it accurate?

mc-nv (approved):

GuanLuo (approved):
Rename PR to drop r22.05

PR #4377, opened 2022-05-13
Add testing for GPU support in decoupled

Tabrizian:


qa/L0_backend_python/io/test.sh
```diff
@@ -45,44 +45,62 @@ rm -fr *.log ./models
...
-    cp ../../python_models/dlpack_io_identity/config.pbtxt ./models/$model_name/
-    (cd models/$model_name && \
-              sed -i "s/^name:.*/name: \"$model_name\"/" config.pbtxt)
-done
+TRIALS="default decoupled"
 
-mkdir -p models/ensemble_io/1/
-cp ../../python_models/ensemble_io/config.pbtxt ./models/ensemble_io
+for trial in $TRIALS; do
+    export TRIAL=$TRIAL
```
GuanLuo, line 51: `=$trial`?

Tabrizian: Good catch. Fixed.

qa/L0_backend_python/io/io_test.py
```diff
@@ -28,45 +28,74 @@
...
+                        client.async_stream_infer(model_name=model_name,
+                                                  inputs=inputs)
+                        result = user_data._completed_requests.get()
+                        output0 = result.as_numpy('OUTPUT0')
+                        self.assertIsNotNone(output0)
+                        self.assertTrue(np.all(output0 == input0))
+
+    def test_ensemble_io(self):
+        # Only run the shared memory leak detection with the default trial
+        if TRIAL == 'default':
```
tanmayv25, line 94: $TRIAL == 'default'

qa/L0_backend_python/io/test.sh
```diff
@@ -45,44 +45,62 @@ rm -fr *.log ./models
...
+    export TRIAL=$trial
+    rm -rf ./models
 
-run_server
-if [ "$SERVER_PID" == "0" ]; then
-    echo -e "\n***\n*** Failed to start $SERVER\n***"
-    cat $SERVER_LOG
-    RET=1
-fi
+    if [ $trial = "default" ]; then
```
tanmayv25, line 54: $trial == "default"

qa/L0_backend_python/io/test.sh
```diff
@@ -45,44 +45,62 @@ rm -fr *.log ./models
...
+            cp ../../python_models/dlpack_io_identity/model.py ./models/$model_name/1/
+            cp ../../python_models/dlpack_io_identity/config.pbtxt ./models/$model_name/
+            (cd models/$model_name && \
+                      sed -i "s/^name:.*/name: \"$model_name\"/" config.pbtxt)
+        done
+    else
+        for i in {1..3}; do
+            model_name=dlpack_io_identity_$i
+            mkdir -p models/$model_name/1/
+            cp ../../python_models/dlpack_io_identity_decoupled/model.py ./models/$model_name/1/
```
tanmayv25, line 67: Can you modify the dlpack_io_identity_decoupled model to actually send multiple responses back to the client? 
The number of responses can be determined by an additional input to the model.
I wonder how Triton will scale.

tanmayv25 (changes requested):

qa/L0_backend_python/io/io_test.py
Tabrizian: This is Python script with TRIAL defined above.

qa/L0_backend_python/io/test.sh
Tabrizian: Both of them work in bash.
https://unix.stackexchange.com/questions/382003/what-are-the-differences-between-and-in-conditional-expressions

qa/L0_backend_python/io/test.sh
Tabrizian: > Can you modify the dlpack_io_identity_decoupled model to actually send multiple responses back to the client?
Can you elaborate on what we are testing by sending multiple responses? I think we are already testing sending multiple responses in `L0_decoupled`. Do you want to test sending multiple GPU responses back?

tanmayv25: Exactly. It will help us in checking if the multiple responses are returned properly in GPU.

GuanLuo (dismissed):

qa/L0_backend_python/io/test.sh
Tabrizian: Added

tanmayv25 (dismissed):

tanmayv25 (approved):

PR #4383, opened 2022-05-17
Fix trace_timestamps_ memory growth in GRPC server

rmccorm4:


src/grpc_server.cc
```diff
@@ -2290,11 +2290,7 @@ class InferHandlerState {
   ~InferHandlerState()
   {
 #ifdef TRITON_ENABLE_TRACING
-    if (trace_ != nullptr) {
-      for (const auto& timestamp : trace_timestamps_) {
-        trace_->CaptureTimestamp(timestamp.first, timestamp.second);
-      }
-    }
```
rmccorm4, line 2297: This didn't call trace_.reset() or trace_timestamps_.clear() before as they would automatically be cleaned up, but I figured it was cleaner to make use of the same helper here anyways.

src/grpc_server.cc
```diff
@@ -2309,23 +2305,34 @@ class InferHandlerState {
...
+    ClearTraceTimestamps();
+#endif  // TRITON_ENABLE_TRACING
   }
 
   void Release()
   {
     context_ = nullptr;
+#ifdef TRITON_ENABLE_TRACING
+    ClearTraceTimestamps();
+#endif  // TRITON_ENABLE_TRACING
```
GuanLuo, line 2320: You can drop the `ifdef` here and above as you have done so in `ClearTraceTimestamps()`, this invocation should be optimized out if tracing is disabled.

tanmayv25 (dismissed):

GuanLuo (approved):

tanmayv25 (approved):

PR #4393, opened 2022-05-17
Add L0_libtorch_io_names test and update docs for new Naming Convention for PyTorch backend

CoderHam:
- Tests fall back to enforcing ordering for cases when index and model arguement naming convention for I/O is not met
- These models that would earlier fail to load, now load succesfully and the input/outputs assume the order they are specified in inside the model configuration file

### DIFF
--- a/docs/model_configuration.md
+++ b/docs/model_configuration.md
@@ -138,37 +138,47 @@ expected by the model.
 
 #### Special Conventions for PyTorch Backend
 
-***Naming Convention:*** Due to the absence of names for outputs in TorchScript
-models, the "name" attribute of outputs in the configuration must follow a
-specific naming convention i.e. "\<name\>__\<index\>". Where \<name\> can be
-any string and \<index\> refers to the position of the corresponding output.
-For inputs, the user has the choice to follow the above naming convention or use
-the names of input arguments to the forward function instead.
+**Naming Convention:** 
 
-This means that if there are two inputs and two outputs, the outputs must be named
-as: "OUTPUT__0" and "OUTPUT__1" such that "OUTPUT__0" refers to first output
-and "OUTPUT__1" refers to the second output.
+Due to the absence of sufficient metadata for inputs/outputs in TorchScript
+model files, the "name" attribute of inputs/outputs in the configuration must
+follow specific naming conventions. These are detailed below.
 
-Similarly, for the inputs, the first and second inputs would be named "INPUT__0"
-and "INPUT__1" respectively. Alternatively, if the forward function for the
-Torchscript model was defined as `forward(self, input0, input1)`, the first and
-second inputs could also be named "input0" and "input1" respectively. 
+1. [Only for Inputs] When the input is not a Dictionary of Tensors, the input
+names in the configuration file can mirror the names of the input arguements to
+the forward function in the model's definition.
 
-***Dictionary of Tensors as Input:*** The PyTorch Backend supports passing
-the inputs to the model in the form of a Dictionary of Tensors. This is only
-supported when there is a *single* input to the model of type Dictionary that
-contains a mapping from a string to a tensor. As an example, if there is a 
-model that expects the input of the form:
+For example, if the forward function for the Torchscript model was defined as
+`forward(self, input0, input1)`, the first and second inputs should be named
+"input0" and "input1" respectively. 
+
+2. `<name>__<index>`: Where \<name\> can be any string and \<index\> is an
+integer index that refers to the position of the corresponding input/output.
+
+This means that if there are two inputs and two outputs, the first and second
+inputs would could named "INPUT__0" and "INPUT__1" and the first and second
+outputs can be named "OUTPUT__0" and "OUTPUT__1" respectively.
+
+3. If all inputs (or outputs) do not follow the same naming convention, then we
+enforce strict ordering from the model configuration i.e. we assume the order of
+inputs (or outputs) in the configuartion is the true ordering of these inputs.
+
+***Dictionary of Tensors as Input:*** 
+
+The PyTorch backend supports passing of inputs to the model in the form of a
+Dictionary of Tensors. This is only supported when there is a *single* input to
+the model of type Dictionary that contains a mapping of string to tensor. As an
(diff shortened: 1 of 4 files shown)


qa/L0_libtorch_io_names/io_names_client.py
```diff
@@ -0,0 +1,95 @@
...
+        io_names = ["INPUT0", "INPUT1", "OUTPUT__0", "OUTPUT__1"]
+        self._infer_helper("libtorch_output_index", io_names)
+
+    def test_no_output_index(self):
+        io_names = ["INPUT0", "INPUT1", "OUTPUT0", "OUTPUT1"]
+        self._infer_helper("libtorch_no_output_index", io_names)
+
+    def test_no_arguements_no_output_index(self):
+        io_names = ["INPUTA", "INPUTB", "OUTPUTA", "OUTPUTB"]
+        self._infer_helper("libtorch_no_arguements_output_index", io_names)
```
GuanLuo, line 92: Add test for the mix of different rules.

qa/L0_libtorch_io_names/io_names_client.py
```diff
@@ -0,0 +1,95 @@
...
+
```
dyastremsky, line 93: Potential tests to add:
Test when one of inputs is valid argument and has index and the other does not (Can also do this with outputs; clarify/test intended behavior)
Test where one of two conditions is met (argument or index) to clarify/test intended behavior.

CoderHam: Thanks for the suggestions. Added these in

qa/L0_libtorch_io_names/io_names_client.py
CoderHam: Added the mix of rules as well

1. ["INPUTA", "INPUT__1", "OUTPUTA", "OUTPUT__1"]
2. ["INPUT0", "INPUTB", "OUTPUTA", "OUTPUT__1"]
3. ["INPUT0", "INPUT__1", "OUTPUT0", "OUTPUT__1"]
4.  ["INPUT1", "INPUT0", "OUTPUT__1", "OUTPUT__0"] (unordered)

dyastremsky (dismissed):

dyastremsky (approved):

qa/L0_libtorch_io_names/io_names_client.py
```diff
@@ -0,0 +1,119 @@
...
+    def test_mix_index(self):
+        io_names = ["INPUTA", "INPUT__1", "OUTPUTA", "OUTPUT__1"]
```
GuanLuo, line 100: Should we fail the model loading if we see only a subset inputs have index suffix? (same for output) If user specify index suffix in config, the intention is clear that the user is aware of the naming convention and would like to make sure the ordering is as intended. Say he specify the following by mistake: `"INPUT1", "INPUT__0"`, with the intended ordering to be `2nd model input, 1st model input`. It appears that the backend will treat this as config ordering and make it `1st model input, 2nd model input`

CoderHam: We do not fail. We allow users to specify model names and log the warning that we enforce strict ordering from the model config. This holds prior to my current changes i.e. users can now specify INPUT__0 twice in the model config and that model would be loaded successfully.

dyastremsky (dismissed):

dyastremsky (dismissed):

tanmayv25 (dismissed):

CoderHam:
@tanmayv25 @dyastremsky I updates the docs to be more clear about the config naming convention for I/O in the PyTorch backend

docs/model_configuration.md
```diff
@@ -138,37 +138,47 @@ expected by the model.
...
-model that expects the input of the form:
+For example, if the forward function for the Torchscript model was defined as
+`forward(self, input0, input1)`, the first and second inputs should be named
+"input0" and "input1" respectively. 
+
+2. `<name>__<index>`: Where \<name\> can be any string and \<index\> is an
+integer index that refers to the position of the corresponding input/output.
+
+This means that if there are two inputs and two outputs, the first and second
+inputs would could named "INPUT__0" and "INPUT__1" and the first and second
```
dyastremsky, line 159: "would could" duplicate word? Should also match the below line, I believe. I.e. they both can or must be named that.

docs/model_configuration.md
```diff
@@ -138,37 +138,47 @@ expected by the model.
...
+Due to the absence of sufficient metadata for inputs/outputs in TorchScript
+model files, the "name" attribute of inputs/outputs in the configuration must
+follow specific naming conventions. These are detailed below.
 
-Similarly, for the inputs, the first and second inputs would be named "INPUT__0"
-and "INPUT__1" respectively. Alternatively, if the forward function for the
-Torchscript model was defined as `forward(self, input0, input1)`, the first and
-second inputs could also be named "input0" and "input1" respectively. 
+1. [Only for Inputs] When the input is not a Dictionary of Tensors, the input
+names in the configuration file can mirror the names of the input arguements to
```
dyastremsky, line 148: Arguments, typo.

dyastremsky: Also, mixing of can and should. We need to be consistent about whether a user can do something or should/must do it.

docs/model_configuration.md
```diff
@@ -138,37 +138,47 @@ expected by the model.
...
 ```
 {'A': tensor1, 'B': tensor2}
 ```
 
-Then the input fields in the configuration do not need to follow the "\<name\>__\<index\>" 
-convention. Instead, the names of the inputs in this case must map to the string
-value 'key' for that specific tensor. In this case the inputs would be "A" and "B",
-where input "A" refers to value corresponding to tensor1 and "B" refers to the
-value corresponding to tensor2.
+The input names in the configuration in this case must not to follow the above
```
dyastremsky, line 177: "Must not follow" (remove to)

dyastremsky (changes requested):

dyastremsky (approved):

tanmayv25 (approved):

PR #4395, opened 2022-05-17
Autocomplete is Used by Triton Supported Backends

nv-kmcgill53:
- Updated L0_backend_config test to include backends which make use of the default-max-batch-size parameter.
- Updated the documentation to declare which backends make use of this feature.

### DIFF
--- a/docs/model_configuration.md
+++ b/docs/model_configuration.md
@@ -300,19 +300,24 @@ configuration](#minimal-model-configuration). You must still provide
 the optional portions of the model configuration by editing the
 config.pbtxt file.
 
-When a model is using the auto-complete feature, a default maximum batch size may 
-be set by using the `--backend-config=default-max-batch-size=<int>` command line argument. 
-This allows all models which are capable of batching and which make use of 
-[Auto Generated Model Configuration](#auto-generated-model-configuration) to have
-a default maximum batch size. This value is set to 2 by default. If the value is
-greater than 1, the [dynamic_batching](#dynamic-batcher) will be set for the model if
-no scheduler is provided in the configuration file. Backend developers may make use of this
-default-max-batch-size by obtaining it from the TRITONBACKEND_BackendConfig api. Currently, the following 
-backends which utilize these default batch values and enable dynamic batching in their
-generated model configurations are:
+### Default Max Batch Size and Dynamic Batcher
+
+When a model is using the auto-complete feature, a default maximum 
+batch size may be set by using the `--backend-config=default-max-batch-size=<int>` 
+command line argument. This allows all models which are capable of
+batching and which make use of [Auto Generated Model Configuration](#auto-generated-model-configuration)
+to have a default maximum batch size. This value is set to 2 by 
+default. Backend developers may make use of this default-max-batch-size
+by obtaining it from the TRITONBACKEND_BackendConfig api. Currently, the
+following backends which utilize these default batch values and turn on 
+dynamic batching in their generated model configurations are:
 
 1. [TensorFlow backend](https://github.com/triton-inference-server/tensorflow_backend)
 2. [Onnxruntime backend](https://github.com/triton-inference-server/onnxruntime_backend)
+
+If a value greater than 1 for the maximum batch size is set for the 
+model, the [dynamic_batching](#dynamic-batcher) config will be set
+if no scheduler is provided in the configuration file.
 
 
 ## Datatypes
--- a/docs/model_configuration.md
+++ b/docs/model_configuration.md
@@ -306,7 +306,7 @@ When a model is using the auto-complete feature, a default maximum
 batch size may be set by using the `--backend-config=default-max-batch-size=<int>` 
 command line argument. This allows all models which are capable of
 batching and which make use of [Auto Generated Model Configuration](#auto-generated-model-configuration)
-to have a default maximum batch size. This value is set to 2 by 
+to have a default maximum batch size. This value is set to 4 by 
 default. Backend developers may make use of this default-max-batch-size
 by obtaining it from the TRITONBACKEND_BackendConfig api. Currently, the
 following backends which utilize these default batch values and turn on 
--- a/docs/model_configuration.md
+++ b/docs/model_configuration.md
@@ -121,14 +121,6 @@ For models that do not support batching, or do not support batching in
 the specific ways described above, *max_batch_size* must be set to
 zero.
 
-When a model is using the auto-complete feature, a default maximum batch size may 
-be set by using the `--backend-config=default-max-batch-size=<int>` command line argument. 
-This allows all models which are capable of batching and which make use of 
-[Auto Generated Model configuration](#auto-generated-model-configuration) to have
-a default maximum batch size. This value is set to 4 by default. While none of the
-officially supported Triton backends implement this feature, backend developers 
-may make use of this value by obtaining it from the TRITONBACKEND_BackendConfig api.
-
 
 ### Inputs and Outputs
 
@@ -307,6 +299,21 @@ Triton only generates the [minimal portion of the model
 configuration](#minimal-model-configuration). You must still provide
 the optional portions of the model configuration by editing the
 config.pbtxt file.
+
+When a model is using the auto-complete feature, a default maximum batch size may 
+be set by using the `--backend-config=default-max-batch-size=<int>` command line argument. 
+This allows all models which are capable of batching and which make use of 
+[Auto Generated Model Configuration](#auto-generated-model-configuration) to have
+a default maximum batch size. This value is set to 2 by default. If the value is
+greater than 1, the [dynamic_batching](#dynamic-batcher) will be set for the model if
+no scheduler is provided in the configuration file. Backend developers may make use of this
+default-max-batch-size by obtaining it from the TRITONBACKEND_BackendConfig api. Currently, the following 
+backends which utilize these default batch values and enable dynamic batching in their
+generated model configurations are:
+
+1. [TensorFlow backend](https://github.com/triton-inference-server/tensorflow_backend)
+2. [Onnxruntime backend](https://github.com/triton-inference-server/onnxruntime_backend)
+
 
 ## Datatypes
 
--- a/qa/L0_backend_config/test.sh
+++ b/qa/L0_backend_config/test.sh
@@ -220,7 +220,7 @@ else
 
     # Assert the max-batch-size is 1 in the case batching is supported
     # in the model but not in the config.
-    MAX_BATCH_LOG_LINE=$(grep -a "\"max_batch_size\":1" $TRIAL.out)
+    MAX_BATCH_LOG_LINE=$(grep -a "\"max_batch_size\":0" $TRIAL.out)
     if [ "$MAX_BATCH_LOG_LINE" == "" ]; then
         echo "*** FAILED: Expected max batch size to be 1 but found: $MAX_BATCH_LOG_LINE\n"
         RET=1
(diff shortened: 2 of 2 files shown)


docs/model_configuration.md
```diff
@@ -125,9 +125,11 @@ When a model is using the auto-complete feature, a default maximum batch size ma
 be set by using the `--backend-config=default-max-batch-size=<int>` command line argument. 
 This allows all models which are capable of batching and which make use of 
 [Auto Generated Model configuration](#auto-generated-model-configuration) to have
-a default maximum batch size. This value is set to 4 by default. While none of the
-officially supported Triton backends implement this feature, backend developers 
-may make use of this value by obtaining it from the TRITONBACKEND_BackendConfig api.
+a default maximum batch size. This value is set to 2 by default. If the value is
+greater than 1, the dynamic batch scheduler will be set for the model if
```
tanmayv25, line 129: dynamic batch scheduler ->  dynamic_batching
add a link to dynamic_batcher.

qa/L0_backend_config/test.sh
```diff
@@ -52,6 +52,9 @@ rm -f $SERVER_LOG_BASE*
 
 COMMON_ARGS="--model-repository=`pwd`/models --strict-model-config=false --log-verbose=1 "
 
+#TODO: NEED TO CHECK OTHER BACKENDS WHICH MAKE USE OF THIS.
```
tanmayv25, line 55: No need of upper-casing.

docs/model_configuration.md
```diff
@@ -125,9 +125,11 @@ When a model is using the auto-complete feature, a default maximum batch size ma
...
+no scheduler is provided in the configuration file. Backends which support this
```
tanmayv25, line 130: Currently, the following backends utilizes these default batch values and enable dynamic batching in generated model configuration:

1. TensorFlow backend
2. OnnxRuntime backend

Add link to the backend repo ^

docs/model_configuration.md
tanmayv25, line 129: All the talk about enabling dynamic batching should  be a part of #auto-generated-model-configuration section and not here.

qa/L0_backend_config/test.sh
```diff
@@ -150,6 +152,136 @@ for ((i=0; i < ${#NEGATIVE_PARSE_ARGS[@]}; i++)); do
...
+    # in the model but not in the config.
+    DYNAMIC_BATCHING_LOG_LINE=$(grep "\"max_batch_size\": 1" $SERVER_LOG)
+    if [ "$DYNAMIC_BATCHING_LOG_LINE" == "" ]; then
+        echo "*** FAILED: Expected max batch size to be 1 but found: $DYNAMIC_BATCHING_LOG_LINE\n"
+        RET=1
+    fi
+
+    # Assert batching disabled    
+    if [ "$(grep -E '\"dynamic_batching\": \{}' $SERVER_LOG)" != "" ]; then
+        echo "*** FAILED: Expected dynamic batching found in configuration when none expected.\n"
```
tanmayv25, line 215: Expected -> Unexpected?

tanmayv25: Also, IMO a better test to look for presence of dynamic batching would be to run the server in --log-verbose=1 and look for messages from batcher thread:
`Starting dynamic-batcher thread`
more info: https://github.com/triton-inference-server/core/blob/main/src/dynamic_batch_scheduler.cc#L278-L290

This is just to ensure that the config change had an actual affect.

tanmayv25 (changes requested):

qa/L0_backend_config/test.sh
```diff
@@ -150,6 +152,136 @@ for ((i=0; i < ${#NEGATIVE_PARSE_ARGS[@]}; i++)); do
...
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "*** FAILED: Server failed to start $SERVER\n"
+    RET=1
+
+else
+    kill $SERVER_PID
+    wait $SERVER_PID
+
+    # Assert the max-batch-size is the command line value
+    DYNAMIC_BATCHING_LOG_LINE=$(grep "\"max_batch_size\": 5" $SERVER_LOG)
```
GuanLuo, line 178: Instead of checking server log, it would be more formal to use [model config protocol](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_model_configuration.md#httprest) to get the model config

qa/L0_backend_config/test.sh
```diff
@@ -150,6 +152,136 @@ for ((i=0; i < ${#NEGATIVE_PARSE_ARGS[@]}; i++)); do
...
+    echo -e "*** FAILED: Server failed to start $SERVER\n"
+    RET=1
+
+else
+    kill $SERVER_PID
+    wait $SERVER_PID
+
+    # Assert the max-batch-size is 1 in the case batching is supported
+    # in the model but not in the config.
+    DYNAMIC_BATCHING_LOG_LINE=$(grep "\"max_batch_size\": 1" $SERVER_LOG)
```
GuanLuo, line 207: Don't we use `max_batch_size: 0` to indicate the model doesn't support batching?

docs/model_configuration.md
nv-kmcgill53: I'm a little unclear whether you want me to split this documentation up or move the whole paragraph up.

qa/L0_backend_config/test.sh
nv-kmcgill53: took out this todo

qa/L0_backend_config/test.sh
nv-kmcgill53: This is a legacy thing. Previously when autocompleting we would set the max batch size to 1 and not set the dynamic batcher. I kept this in to be consistent with past behavior. In the case of a max batch size of 0 or 1, aren't they equivalent?

docs/model_configuration.md
nv-kmcgill53: Added link back to dynamic batcher

docs/model_configuration.md
nv-kmcgill53: Added the statement with links to the backends.

qa/L0_backend_config/test.sh
nv-kmcgill53: Fixed log messages.

qa/L0_backend_config/test.sh
nv-kmcgill53: Updated the tests to parse the returned model configuration rather than parsing the log.

docs/model_configuration.md
tanmayv25: I am talking about splitting this documentation. The things with dynamic batcher you are talking about makes more sense in auto complete config.

qa/L0_backend_config/test.sh
tanmayv25: Unfortunately, I think there is no way to differentiate whether max_batch_size = 0, was provided my the user or is default value from the protobuf.

If the user has not provided any config (input/output) and max_batch_size is read as 0, the auto-complete logic should check the model inputs/outputs and see if it has (-1) as first dimension or is marked with unknown_rank.  If yes, then it should enable batching with max_batch_size set to the default value.

But if they have provided input/output config, then we must look at whether the number of dims is one less than in the model. This will hint dynamic batching.

tanmayv25: > In the case of a max batch size of 0 or 1, aren't they equivalent?

They are different. In sense, the io config dims will be different.

docs/model_configuration.md
nv-kmcgill53: Updated documentation to split out the dynamic batching explanation.

tanmayv25 (dismissed):

docs/model_configuration.md
```diff
@@ -302,6 +294,26 @@ configuration](#minimal-model-configuration). You must still provide
 the optional portions of the model configuration by editing the
 config.pbtxt file.
 
+### Default Max Batch Size and Dynamic Batcher
+
+When a model is using the auto-complete feature, a default maximum 
+batch size may be set by using the `--backend-config=default-max-batch-size=<int>` 
+command line argument. This allows all models which are capable of
+batching and which make use of [Auto Generated Model Configuration](#auto-generated-model-configuration)
+to have a default maximum batch size. This value is set to 2 by
```
GuanLuo, line 303: Now the value is 4?

qa/L0_backend_config/test.sh
```diff
@@ -150,6 +150,165 @@ for ((i=0; i < ${#NEGATIVE_PARSE_ARGS[@]}; i++)); do
...
+
+SERVER_ARGS="--backend-config=tensorflow,default-max-batch-size=5 $COMMON_ARGS"
+SERVER_LOG=$SERVER_LOG_BASE.backend_config_tensorflow_batch_5.log
+run_server
+
+TRIAL=tensorflow_batching_on
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "*** FAILED: Server failed to start $SERVER\n"
+    RET=1
+
```
GuanLuo, line 184: Remove new line

qa/L0_backend_config/test.sh
GuanLuo: Yes on the protobuf default value and the auto-complete logic when nothing is provided. But note that the server runs with `--backend-config=tensorflow,default-max-batch-size=0` so the auto-completed config should have `max_batch_size: 0` and full shape as I/O shape even all first dimensions are dynamic axis.

tanmayv25: In other words, `max_batch_size=1` does qualifies the model as a batching with size 1. The user can only run inferences with size [1,16]
`max_batch_size=0` implies the model is not batching, single request with input shape: [N,16] is possible for `max_batch_size:0`, but not for `max_batch_size:1`. Where N>1.

nv-kmcgill53: Updated the tests to reflect this behaviour. Please see associated backends for changes as well (couldn't find how to link within a PR):
- tensorflow.cc:1296
- onnxruntime.cc:789

docs/model_configuration.md
nv-kmcgill53: updated to 4

qa/L0_backend_config/test.sh
nv-kmcgill53: removed.

nv-kmcgill53:
@GuanLuo Updated the tests to assert the max_batch_size is 0 when the default-max-batch-size is 0. Please see associated backends for changes as well:

- [tensorflow.cc:1296](https://github.com/triton-inference-server/tensorflow_backend/pull/60/files)
- [onnxruntime.cc:789](https://github.com/triton-inference-server/onnxruntime_backend/pull/120/files)

GuanLuo (approved):

GuanLuo (approved):

nv-kmcgill53:
Rebased on top of main

qa/L0_backend_config/test.sh
```diff
@@ -150,6 +150,164 @@ for ((i=0; i < ${#NEGATIVE_PARSE_ARGS[@]}; i++)); do
...
+else
+    save_model_config
+
+    # Assert the max-batch-size is 0 in the case batching is supported
+    # in the model but not in the config.
+    MAX_BATCH_LOG_LINE=$(grep -a "\"max_batch_size\":0" $TRIAL.out)
+    if [ "$MAX_BATCH_LOG_LINE" == "" ]; then
+        echo "*** FAILED: Expected max batch size to be 1 but found: $MAX_BATCH_LOG_LINE\n"
```
tanmayv25, line 224:
```suggestion
        echo "*** FAILED: Expected max batch size to be 0 but found: $MAX_BATCH_LOG_LINE\n"
```

tanmayv25 (changes requested):

tanmayv25 (approved):

PR #4401, opened 2022-05-18
Fix long-running Java memory growth test

dyastremsky:
The long-running Java memory growth test can fail due to intermittence in the way Java garbage collects. Given that this test runs at about ~15MB, a megabyte difference can result in test failure, since I shrunk the allowed growth in a previous test.

This PR aims to resolve intermittence in the test results by allowing memory to grow by 25% instead of 5%, allowing for variance due to Java's memory management while providing a small enough value that a leak would cause the test to fail over its runtime. I also added an additional constraint of ensuring the memory stays under 30MB, a value that allows for variance but will fail if memory needs for this test increase in the future.


qa/L0_java_memory_growth/Simple.java
```diff
@@ -394,6 +394,13 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
       } else {
         return false;
       }
+      if((memory_snapshots.get(index_90) / 1E6) >= 30){
```
tanmayv25, line 397: Not too familiar with java, but shouldn't these magic numbers governing constraints be a some kind of constants?

qa/L0_java_memory_growth/Simple.java
```diff
@@ -394,6 +394,13 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
...
+        System.out.println("Exceeded allowed memory (30MB), got " +
```
tanmayv25, line 398: The 30 MB should come from the constant variable. So that in future we can just change the constant value and all these messages would not need change.

qa/L0_java_memory_growth/Simple.java
```diff
@@ -834,7 +841,7 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
 
       Runnable runnable =
         () -> { 
-          if(ValidateMemoryGrowth(.05f)){
+          if(ValidateMemoryGrowth(.25f)){
```
tanmayv25, line 844: How about abstracting all these factors in a constraint object?

dyastremsky: Thanks for the suggestion. Do you mean a class to hold all the constants that gets passed into this function? I can also make these command line arguments (with default values) and just pass them into this function, if that works.

tanmayv25: Yes. The cli arguments works great! I was little confused with the Simple.java nomenclature.

Can we rename the Simple.java to memory_growth_test.java so that we are consistent with all other tests?

dyastremsky: Sure thing. Thanks for clarifying!

GuanLuo: 25% is a large tolerance. Although I understand that the base memory usage is small so 25% may end up be a small amount, we should keep the small percentage tolerance and use the fixed minimum delta added above as additional checking

qa/L0_java_memory_growth/Simple.java
```diff
@@ -394,6 +394,13 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
       } else {
         return false;
       }
```
GuanLuo, line 396: Don't return early, otherwise the added code will never be reached.

qa/L0_java_memory_growth/Simple.java
dyastremsky: I realized the print message needs to be fixed (decimal points), but the last run had a 15.2% memory "growth" due to the memory snapshots starting out around 17MB and later going to 15MB.

If we keep a 5% tolerance threshold, tests will sometimes fail. 25% seems like a better spot to mitigate intermittence, since the memory would grow way beyond that in 12 hours if there's a memory leak (as we've seen in the past). Or what threshold would you prefer? I can also make it so that the threshold is higher for the long-running test only.

GuanLuo: I was thinking that we will fail the test if the validation doesn't satisfy both % and delta constraints

dyastremsky: We can do that, but if we leave it at 5%, we should expect future failures. This PR fixes the error due to it allowing a higher percentage of variation (25%) to account for the non-deterministic garbage collection in Java.

Also, note that the new constraint I added is the maximum memory (total), but I can change it to the absolute delta if preferred.

qa/L0_java_memory_growth/Simple.java
dyastremsky: Thanks! Addressed.

qa/L0_java_memory_growth/Simple.java
dyastremsky: Fixed.

qa/L0_java_memory_growth/Simple.java
dyastremsky: Fixed.

qa/L0_java_memory_growth/test.sh
```diff
@@ -83,13 +85,15 @@ CLIENT_LOG="./client_$LOG_IDX.log"
...
+MAX_MEM_GROWTH=10
 if [ "$TRITON_PERF_LONG" == 1 ]; then
     # ~1 day
     ITERS=125000000
+    MAX_MEM_GROWTH=25
 fi
 
 echo -e "\nRunning Memory Growth Test, $ITERS Iterations\n"
-$BASE_COMMAND -Dexec.args="-r $MODEL_REPO -c -i $ITERS" >>$CLIENT_LOG 2>&1
+$BASE_COMMAND -Dexec.args="-r $MODEL_REPO -c -i $ITERS --max-growth $MAX_MEM_GROWTH" >>$CLIENT_LOG 2>&1
```
jbkyang-nvi, line 96: Can we set max heap size? I think this could help with the machine dependent growth behavior

qa/L0_java_memory_growth/test.sh
```diff
@@ -38,7 +38,8 @@ set -e
 
 MODEL_REPO=`pwd`/models
 SAMPLES_REPO=`pwd`/javacpp-presets/tritonserver/samples
-cp Simple.java $SAMPLES_REPO
+cp MemoryGrowthTest.java $SAMPLES_REPO/Simple.java
+sed -i 's/MemoryGrowthTest/Simple/g' $SAMPLES_REPO/Simple.java
```
GuanLuo, line 42: Why need to place it to simple repo as simple.java at the end? Can't it be compiled as is?

qa/L0_java_memory_growth/test.sh
```diff
@@ -83,13 +85,15 @@ CLIENT_LOG="./client_$LOG_IDX.log"
 
 # Longer-running memory growth test 
 ITERS=1000000
+MAX_MEM_GROWTH=10
 if [ "$TRITON_PERF_LONG" == 1 ]; then
     # ~1 day
     ITERS=125000000
+    MAX_MEM_GROWTH=25
```
GuanLuo, line 92: Add unit to variable name

qa/L0_java_memory_growth/MemoryGrowthTest.java
```diff
@@ -378,26 +386,36 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
...
+        System.out.println("Exceeded allowed memory growth (" +
+          (max_growth_allowed * 100) + "%)");
       }
+
+      if((memory_snapshots.get(index_max) / bytes_in_mb) >= max_mem_allowed){
+        passed = false;
+        System.out.println("Exceeded allowed memory (" + max_mem_allowed + 
+          "MB), got " + memory_allocation_delta_mb + "MB");
+      }
+      return passed;
```
GuanLuo, line 414: Just to confirm, so test passes only if `(< %) && (< delta)`? And should we further limit the delta allowed? From the test it seems to be much less than 30 MB even for long running setting

dyastremsky: The test only passes if the delta is (< %) and the maximum memory used is (<max_mem_allowed). We can reduce it further to 20MB. (It usually runs in the 15-17MB range). The 30MB was to provide more slack to prevent intermittence from Java's garbage collection.

qa/L0_java_memory_growth/test.sh
dyastremsky: I don't believe so. The result is simply the program crashing (out of memory error), not the memory being managed differently.

This was what happened last time, but I tested it again to make sure. When I set it to 64MB, it crashed due to the memory running out during runtime. When I set it to 512MB, the behavior did not change. If you'd prefer, I can set the memory higher, but it doesn't seem to change the memory management.

qa/L0_java_memory_growth/test.sh
dyastremsky: Maven relies on the pom.xml in that folder, and everything's already set up that way to make it easy.

I originally did this the same way as the sequence test. To help with clarity around this comment, I modified the test script to instead copy it as MemoryGrowthTest.java and update the pom.xml instead (more lines of code, but should be clearer). I'm trying to keep everything standardized between tests.

qa/L0_java_memory_growth/test.sh
dyastremsky: Thanks, addressed.

qa/L0_java_memory_growth/MemoryGrowthTest.java
```diff
@@ -69,14 +74,16 @@ protected static class DeleteDeallocator extends TRITONSERVER_Server implements
         System.err.println(msg);
       }
 
-      System.err.println("Usage: java " + Simple.class.getSimpleName() + " [options]");
+      System.err.println("Usage: java " + MemoryGrowthTest.class.getSimpleName() + " [options]");
```
GuanLuo, line 77: nit: `getSimpleName()`?

qa/L0_java_memory_growth/MemoryGrowthTest.java
GuanLuo: I see, the limit is on total memory usage.

qa/L0_java_memory_growth/test.sh
```diff
@@ -38,7 +38,9 @@ set -e
 
 MODEL_REPO=`pwd`/models
 SAMPLES_REPO=`pwd`/javacpp-presets/tritonserver/samples
-cp Simple.java $SAMPLES_REPO
+rm $SAMPLES_REPO/Simple.java
+cp MemoryGrowthTest.java $SAMPLES_REPO/MemoryGrowthTest.java
+sed -i 's/Simple/MemoryGrowthTest/g' $SAMPLES_REPO/pom.xml
```
GuanLuo, line 43: Does it make sense if we add `pom.xml` in this test folder, and it is dedicated to build the memory growth test? Then the pom modification below shouldn't be needed as well.

qa/L0_java_memory_growth/MemoryGrowthTest.java
dyastremsky: This is a Java function to get the class name (java.lang.Class.getSimpleName()), not part of our code. I can see how that'd be confusing, though I don't think this can be helped.

qa/L0_java_memory_growth/test.sh
dyastremsky: Good suggestion. The pom.xml and Java file are now located under the test subdirectory (to prevent Maven from searching the subdirectories for other build files and getting confused).

qa/L0_java_memory_growth/test.sh
dyastremsky: Tested the other garbage collection options and parameters you sent over. Ultimately, ended up setting XX:MaxGCPauseMillis to 40ms instead of the default of 200ms. The garbage collector should be collecting more frequently since it's aiming to pause for a shorter window each time.

This seems to be resulting in slightly less memory growth during test runs, but we'll see in time if that consistency continues in the long-running stress test.

GuanLuo (dismissed):
Is there a running CI on regular L0_java_memory_growth?

tanmayv25 (dismissed):

dyastremsky:
> Is there a running CI on regular L0_java_memory_growth?

https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/4896903

Thanks for checking. Will merge once it passes.

dyastremsky:
The test initially failed, as I the memory was 21MB in one run, so I increased the max memory allowed back to 30MB. I don't know if this had to do with the options I added to (hopefully) reduce memory growth variance or there's just variability, as some runs were lower and some higher.

Then, it failed again due to a race condition ("Memory growth test passed" was split up). Java's print line is not supposed to have race conditions. However, it seems that the Triton bindings might bypass the native Java locks in the way they print. There's not a way to call the server destructor early in Java that works 100% of the time or is recommended in Java. For ease, I added a 5-second sleep so that the garbage collector has time to see the server is no longer used and deallocate it (waiting for ServerIsLive, etc. prevents the garbage collector from marking the object for collection).

In summary, two new commits:
- Increasing the max memory back to 30MB. Several test runs have already been anywhere from 15-22MB, so this provides flexibility so that we're not constantly going back to this test to reconsider the constraints.
- Added a sleep to prevent race conditions due to the way the bindings interact with Java's native print.

If you're okay, please re-review/re-approve, @GuanLuo and @tanmayv25.

dyastremsky:
Rebased. Also reverted to the old way of using the pom.xml from JavaCPP in the most recent commit, but left the test named MemoryGrowthTest for clarity. I realized that we'd need to maintain the pom.xml instead of letting JavaCPP's release process do it if we created our own (e.g. updating the versions on every JavaCPP release).

The PR is ready for review.

tanmayv25 (changes requested):
Minor comments.

qa/L0_java_memory_growth/test/MemoryGrowthTest.java
```diff
@@ -660,11 +678,29 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
...
           case "-?":
             Usage(null);
             break;
+          case "--max-growth":
+            i++;
+            try {
+              max_growth_allowed = Integer.parseInt(args[i]) / 100.0f;
+            } catch (NumberFormatException e){
+              Usage(
+                  "--max-growth must be used to specify allowed memory growth (%)");
```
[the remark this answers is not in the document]
dyastremsky: Updated.

qa/L0_java_memory_growth/test/MemoryGrowthTest.java
```diff
@@ -833,8 +869,16 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
...
       Runnable runnable =
-        () -> { 
-          if(ValidateMemoryGrowth(.05f)){
+        () -> {
+          boolean passed = ValidateMemoryGrowth(max_growth_allowed, max_mem_allowed);
+          
+          try {
+            Thread.sleep(5000);
+          } catch (InterruptedException e){
+            System.out.println("Sleep interrupted.");
```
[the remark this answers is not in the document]
dyastremsky: Sure, done.

qa/L0_java_memory_growth/test/MemoryGrowthTest.java
```diff
@@ -660,11 +678,29 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
...
+                  "--max-growth must be used to specify allowed memory growth (%)");
+            }
+            break;
+          case "--max-memory":
+            i++;
+            try {
+              max_mem_allowed = Integer.parseInt(args[i]);
+            } catch (NumberFormatException e){
+              Usage(
+                  "--max-memory must be used to specify maximum allowed memory (MB)");
```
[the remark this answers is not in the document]
dyastremsky: Updated.

qa/L0_java_memory_growth/test/MemoryGrowthTest.java
```diff
@@ -833,8 +869,16 @@ output0, output1, new FloatPointer(input0_data), new FloatPointer(input1_data),
       }
 
       Runnable runnable =
-        () -> { 
-          if(ValidateMemoryGrowth(.05f)){
+        () -> {
+          boolean passed = ValidateMemoryGrowth(max_growth_allowed, max_mem_allowed);
+          
+          try {
+            Thread.sleep(5000);
```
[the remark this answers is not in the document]
dyastremsky: Good suggestion. Added a clarifying comment.

PR #4409, opened 2022-05-19
Fixed Tests to Align With New Tensorflow Batching Support

nv-kmcgill53:
- Fixed L0_model_config to have new error message. Before, there was validation being done during autocomplete; after, this error gets checked during the model validation step.
- Fixed L0_unknown_rank test to not check the shape of the output anymore. This test was old and had regressed without our knowledge since the test succeeded if the response did not raise an exception.
- Added new test_wrong_input to prove poorly shaped inputs from the request raise an exception.


qa/L0_model_config/autofill_noplatform/tensorflow_savedmodel/unknown_output/expected
```diff
@@ -1 +1 @@
-Invalid argument: unexpected inference output 'OUTPUT_UNKNOWN', allowed outputs are: OUTPUT0, OUTPUT1
\ No newline at end of file
+Invalid argument: Config file provided output name 'OUTPUT_UNKNOWN' but none found in loaded model 'unknown_output'.
```
tanmayv25, line 1: This message should be deferred to the validation at the model load right?
I think we fixed this in IO Refactor PR?

nv-kmcgill53: You're right, this leaked in.

tanmayv25 (approved):

tanmayv25 (approved):

tanmayv25 (approved):

PR #4413, opened 2022-05-20
fix error when deploying a model from mlflow

Rusteam:
Add `config_path` key to `copy_paths` to avoid `KeyError` when deploying from mlflow models in **onnx** flavor with `config.pbtxt`

### DIFF
--- a/deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
+++ b/deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
@@ -303,6 +303,11 @@ class TritonPlugin(BaseDeploymentClient):
                 elif file.name == 'config.pbtxt':
                     config_file = file.name
                     copy_paths['config_path'] = {}
+                elif file.suffix == '.txt' and file.stem != 'requirements':
+                    copy_paths[file.stem] = {
+                        'from': file,
+                        'to': triton_deployment_dir
+                    }
             if model_file is None:
                 for file in artifact_path.iterdir():
                     if file.suffix == '.onnx':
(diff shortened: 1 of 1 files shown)


GuanLuo:
Hi @Rusteam, thanks for the contribution. The change looks good but do you mind to sign the CLA as instructed [here](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) before we can merge this pull request?

Rusteam:
I don't mind. But I'm not a corporation, I'm an individual contributor. Is it fine?

Rusteam:
@GuanLuo I've sent a signed CLA to the email mentioned. Can you review and merge now?

I've also added two other commits: to fix ensemble deployment and `labels.txt` if present

GuanLuo:
@Rusteam sorry that I missed the earlier message, we have received your CLA and the change looks good to me. I would like to extend `L0_mlflow` test to cover this case where ONNX model is logged with additional files (i.e. config / labels). I wonder if you have any insight on logging ONNX flavor in this form, currently the test is using the MLFlow ONNX plugin to [log the model](https://github.com/triton-inference-server/server/blob/main/qa/L0_mlflow/plugin_test.py#L53-L55), but I don't think `mflow.onnx.log_model` can accept additional files

Rusteam:
You're right it does not. What I do is I log other files as artifacts right after logging the model, i.e.:
```
with mlflow.start_run():
    mlflow.onnx.log_model(model, path)
    mlflow.log_artifact("labels.txt", path)
    mlflow.log_artifact("config.pbtxt", path)
```

GuanLuo (approved):

GuanLuo:
Thanks for the hint, I have extended the test to cover the case with config in this PR https://github.com/triton-inference-server/server/pull/4469, feel free to leave any comments

tanmayv25 (approved):

PR #4417, opened 2022-05-20
Improve error message for shared memory manager

Tabrizian:


src/shared_memory_manager.cc
```diff
@@ -315,7 +315,8 @@ SharedMemoryManager::RegisterCUDASharedMemory(
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INVALID_ARG,
         std::string(
```
tanmayv25, line 317:
```
std::string(
            "failed to register CUDA shared memory region '" + name +
            "': " + TRITONSERVER_ErrorMessage(err))
            .c_str());
```
Conforms with most of our error messages.

tanmayv25: Note that this has removed `. error:` part.

tanmayv25 (approved):

PR #4420, opened 2022-05-20
Add triton_c_api to perf tests

rmccorm4:
- [x] resnet
- [x] nomodel
- [x] skip/check [unsupported combinations](https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md#non-supported-functionalities) with C API
  - [x] SHM / CUDA SHM
  - [x] Python backend requires SHM, so skip
- [ ] email
  - [x] add CAPI protocol to existing test, gitlab kibana summary
- [x] dashboards
  - [x] add visualizations, filter by protocol

### DIFF
--- a/qa/L0_perf_nomodel/run_test.sh
+++ b/qa/L0_perf_nomodel/run_test.sh
@@ -83,13 +83,10 @@ PERF_CLIENT_EXTRA_ARGS="$PERF_CLIENT_PERCENTILE_ARGS --shared-memory \"${SHARED_
 # Overload use of PERF_CLIENT_PROTOCOL for convenience with existing test and 
 # reporting structure, though "triton_c_api" is not strictly a "protocol".
 if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
-    # Using C API requires extra info to start in-process server
-    # NOTE: PA C API doesn't expose --backend-directory, but defaults to
-    #       ${TRITON_DIR}/backends
+    # Server will be run in-process with C API
     SERVICE_ARGS="--service-kind triton_c_api \
                   --triton-server-directory ${TRITON_DIR} \
                   --model-repository ${MODEL_REPO}"
-# Otherwise run as normal via HTTP/GRPC
 else
     SERVICE_ARGS="-i ${PERF_CLIENT_PROTOCOL}"
 fi
--- a/qa/L0_perf_nomodel/run_test.sh
+++ b/qa/L0_perf_nomodel/run_test.sh
@@ -139,13 +139,6 @@ for BACKEND in $BACKENDS; do
         REPO_DIR=./custom_models
     elif [ $BACKEND == "python" ]; then
         REPO_DIR=./python_models
-
-        # FIXME: If PA C API adds SHMEM support, remove this.
-        if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
-            echo "WARNING: Python backend requires Shared Memory and PA C API \
-                  does not support shared memory. Skipping this test."
-            continue
-        fi
     else
         REPO_DIR=$DATADIR/qa_identity_model_repository
     fi
--- a/qa/L0_perf_nomodel/run_test.sh
+++ b/qa/L0_perf_nomodel/run_test.sh
@@ -139,6 +139,13 @@ for BACKEND in $BACKENDS; do
         REPO_DIR=./custom_models
     elif [ $BACKEND == "python" ]; then
         REPO_DIR=./python_models
+
+        # FIXME: If PA C API adds SHMEM support, remove this.
+        if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
+            echo "WARNING: Python backend requires Shared Memory and PA C API \
+                  does not support shared memory. Skipping this test."
+            continue
+        fi
     else
         REPO_DIR=$DATADIR/qa_identity_model_repository
     fi
--- a/qa/L0_perf_nomodel/run_test.sh
+++ b/qa/L0_perf_nomodel/run_test.sh
@@ -47,7 +47,8 @@ TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
 ARCH=${ARCH:="x86_64"}
 SERVER=${TRITON_DIR}/bin/tritonserver
 BACKEND_DIR=${TRITON_DIR}/backends
-SERVER_ARGS="--model-repository=`pwd`/models --backend-directory=${BACKEND_DIR}"
+MODEL_REPO="${PWD}/models"
+SERVER_ARGS="--model-repository=${MODEL_REPO} --backend-directory=${BACKEND_DIR}"
 source ../common/util.sh
 
 # DATADIR is already set in environment variable for aarch64
@@ -78,6 +79,18 @@ PERF_CLIENT_PERCENTILE_ARGS="" &&
     (( ${PERF_CLIENT_PERCENTILE} != 0 )) &&
     PERF_CLIENT_PERCENTILE_ARGS="--percentile=${PERF_CLIENT_PERCENTILE}"
 PERF_CLIENT_EXTRA_ARGS="$PERF_CLIENT_PERCENTILE_ARGS --shared-memory \"${SHARED_MEMORY}\""
+
+# Overload use of PERF_CLIENT_PROTOCOL for convenience with existing test and 
+# reporting structure, though "triton_c_api" is not strictly a "protocol".
+if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
+    # Using C API requires extra info to start in-process server
+    SERVICE_ARGS="--service-kind triton_c_api \
+                  --triton-server-directory ${TRITON_DIR} \
+                  --model-repository ${MODEL_REPO}"
+# Otherwise run as normal via HTTP/GRPC
+else
+    SERVICE_ARGS="-i ${PERF_CLIENT_PROTOCOL}"
+fi
 
 #
 # Use "identity" model for all model types.
@@ -154,12 +167,15 @@ for BACKEND in $BACKENDS; do
                 echo "dynamic_batching { preferred_batch_size: [ ${DYNAMIC_BATCH} ] }" >> config.pbtxt)
     fi
 
-    SERVER_LOG="${RESULTDIR}/${NAME}.serverlog"
-    run_server
-    if [ $SERVER_PID == 0 ]; then
-        echo -e "\n***\n*** Failed to start $SERVER\n***"
-        cat $SERVER_LOG
-        exit 1
+    # Don't start separate server if testing perf with C API
+    if [[ "${PERF_CLIENT_PROTOCOL}" != "triton_c_api" ]]; then
+        SERVER_LOG="${RESULTDIR}/${NAME}.serverlog"
+        run_server
+        if [ $SERVER_PID == 0 ]; then
+            echo -e "\n***\n*** Failed to start $SERVER\n***"
+            cat $SERVER_LOG
+            exit 1
+        fi
     fi
 
     set +e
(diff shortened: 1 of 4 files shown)


qa/L0_perf_nomodel/run_test.sh
```diff
@@ -124,6 +139,13 @@ for BACKEND in $BACKENDS; do
         REPO_DIR=./custom_models
     elif [ $BACKEND == "python" ]; then
         REPO_DIR=./python_models
+
+        # FIXME: If PA C API adds SHMEM support, remove this.
+        if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
```
Tabrizian, line 144: I was wondering why Python backend is excluded here. PA C-API shared memory support is independent from Python backend's internal shared memory.

rmccorm4: I mis-associated some shm errors I was getting. Thanks for the catch.

qa/L0_perf_resnet/test.sh
```diff
@@ -137,6 +137,13 @@ done
 # Tests with optimization enabled models
 for MODEL_NAME in $OPTIMIZED_MODEL_NAMES; do
     for PROTOCOL in $PROTOCOLS; do
+        # NOTE: Skip TF-TRT for C API test, it is too inconsistent with
```
tanmayv25, line 140: Create a jira ticket to enable TF-TRT optimization in C API test. You can mention several solutions like offline TF-TRT optimization we were discussing.

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,20 +66,58 @@ fi
...
-$PERF_CLIENT -v -i ${PERF_CLIENT_PROTOCOL} -m $MODEL_NAME -p${MEASUREMENT_WINDOW} \
-                -b${STATIC_BATCH} --concurrency-range ${CONCURRENCY}
+# Overload use of PERF_CLIENT_PROTOCOL for convenience with existing test and 
+# reporting structure, though "triton_c_api" is not strictly a "protocol".
+# NOTE: tritonserver will be ran in-process with perf_analyzer
+if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
+    # Using C API requires extra info to start in-process server
+    # NOTE: PA C API doesn't expose --backend-directory, but defaults to
+    #       ${TRITON_DIR}/backends
+    # NOTE: PA C API doesn't expose --backend-config, but this is only used
```
tanmayv25, line 76: This comment is confusing. I don't think we need --backend-config tfserving grpc test.

tanmayv25: Shouldn't all tfserving benchmarking be in L0_perf_tfs?

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,20 +66,58 @@ fi
...
+    #       for tfserving grpc test currently.
+    SERVICE_ARGS="--service-kind triton_c_api \
+                  --triton-server-directory ${TRITON_DIR} \
+                  --model-repository ${MODEL_REPO}"
+# Otherwise run server as separate process as normal via HTTP/GRPC
```
tanmayv25, line 81: No need to add so many comments. The if-else logic is simple enough to follow. A lot of comments sometimes becomes difficult to maintain. Imagine in future if a new condition is added and we skip updating this comment which is already in if block.
If you think logic is being hidden then you can do something like:
```
if [[ "${PERF_CLIENT_PROTOCOL}" == "triton_c_api" ]]; then
....
else # PERF_CLIENT_PROTOCOL != "triton_c_api"
...
fi
```

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,20 +66,58 @@ fi
...
+    # Don't start separate server if testing perf with C API
+    SERVER_LOG="${NAME}.serverlog"
+    run_server
+    if (( $SERVER_PID == 0 )); then
+        echo -e "\n***\n*** Failed to start $SERVER\n***"
+        cat $SERVER_LOG
+        exit 1
+    fi
+
+    # NOTE: Running PA for warmup doesn't make sense for C API since it uses
```
tanmayv25, line 94: Move these lines below L98. And reword to may be something like this:

```
Note: Running extra PA for warmup doesn't make sense for C API since it uses in-process tritonserver which will exit along with this PA process.
```

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,20 +66,58 @@ fi
...
+                    ${SERVICE_ARGS}
+fi
 
-$PERF_CLIENT -v -i ${PERF_CLIENT_PROTOCOL} -m $MODEL_NAME -p${MEASUREMENT_WINDOW} \
+# Measure perf client results and write them to a file for reporting
+$PERF_CLIENT -v -m $MODEL_NAME -p${MEASUREMENT_WINDOW} \
                 -b${STATIC_BATCH} --concurrency-range ${CONCURRENCY} \
+                ${SERVICE_ARGS} \
                 -f ${NAME}.csv 2>&1 | tee ${NAME}.log
+# Check perf client succeeded
```
tanmayv25, line 109: Comment can be removed. In general, we must try to add comment only when the logic is not clear from the code.

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,20 +66,58 @@ fi
...
 if (( $? != 0 )); then
     RET=1
 fi
-curl localhost:8002/metrics -o ${NAME}.metrics >> ${NAME}.log 2>&1
-if (( $? != 0 )); then
-    RET=1
+
+# No metrics endpoint available when running with C API
+if [[ "${PERF_CLIENT_PROTOCOL}" != "triton_c_api" ]]; then
+    # NOTE: The metrics API output is not currently used
```
tanmayv25, line 116: May be remove this altogether? We must initiate discussion with the team if we can remove these metrics from `L0_perf_*`. I don't see much value in these.

qa/L0_perf_nomodel/test.sh
```diff
@@ -161,6 +187,13 @@ for idx in "${!TEST_NAMES[@]}"; do
     TEST_INSTANCE_COUNT=${TEST_INSTANCE_COUNTS[$idx]}
     TEST_CONCURRENCY=${TEST_CONCURRENCY[$idx]}
 
+    # FIXME: If PA C API adds SHMEM support, remove this.
+    if [[ "${BENCHMARK_TEST_SHARED_MEMORY}" != "none" ]] && \
+       [[ "${TEST_PROTOCOL}" == "triton_c_api" ]]; then
+      echo "WARNING: Perf Analyzer C API doesn't currently support shared memory, skipping."
```
tanmayv25, line 193: Perf Analyzer does not support shared memory I/O when benchmarking directly with Triton C API, skipping.

tanmayv25 (changes requested):

qa/L0_perf_resnet/run_test.sh
rmccorm4: [This](https://github.com/triton-inference-server/server/blob/2f60b85e37ed6578cb25623013364bebba104341/qa/L0_perf_resnet/test.sh#L212-L230) was what I was referring to, but I agree I have too many comments here. Cleaning them up, and removing mention of this.

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,22 +66,41 @@ fi
...
-$PERF_CLIENT -v -i ${PERF_CLIENT_PROTOCOL} -m $MODEL_NAME -p${MEASUREMENT_WINDOW} \
+# Measure perf client results and write them to a file for reporting
+$PERF_CLIENT -v -m $MODEL_NAME -p${MEASUREMENT_WINDOW} \
                 -b${STATIC_BATCH} --concurrency-range ${CONCURRENCY} \
+                ${SERVICE_ARGS} \
                 -f ${NAME}.csv 2>&1 | tee ${NAME}.log
 if (( $? != 0 )); then
     RET=1
 fi
-curl localhost:8002/metrics -o ${NAME}.metrics >> ${NAME}.log 2>&1
```
rmccorm4, line 87: @tanmayv25 @Tabrizian @GuanLuo What do you think of removing this? It does not appear to be used anywhere. The metrics aren't pushed to kibana and aren't read in [reporter.py](https://github.com/triton-inference-server/server/blob/main/qa/common/reporter.py)

rmccorm4: Based on Tanmay's comment: https://github.com/triton-inference-server/server/pull/4420#discussion_r881063460

qa/L0_perf_resnet/run_test.sh
```diff
@@ -73,22 +66,41 @@ fi
...
+    run_server
+    if (( $SERVER_PID == 0 )); then
+        echo -e "\n***\n*** Failed to start $SERVER\n***"
+        cat $SERVER_LOG
+        exit 1
+    fi
+
+    # Run the model once to warm up. Some frameworks do optimization on the first requests.
+    # Must warmup similar to actual run so that all instances are ready
+    # Note: Running extra PA for warmup doesn't make sense for C API since it
```
Tabrizian, line 89: Does the warmup cause significant difference in the performance of C-API models? If yes, maybe we can use the [server warmup](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#model-warmup) to make sure the perf numbers are accurate?

qa/L0_perf_resnet/test.sh
```diff
@@ -137,6 +137,13 @@ done
 # Tests with optimization enabled models
 for MODEL_NAME in $OPTIMIZED_MODEL_NAMES; do
     for PROTOCOL in $PROTOCOLS; do
+        # DLIS-3805: Skip TF-TRT for C API test, it is too inconsistent with
```
Tabrizian, line 140: Does using the server warmup fix this problem?

qa/L0_perf_resnet/run_test.sh
Tabrizian: I think it should be fine to remove them if it is not being used anywhere. The PA report already includes the server stats if we decide to use them. The only metrics that are not available is the GPU usage which I don't think we're tracking them anywhere.

qa/L0_perf_resnet/test.sh
rmccorm4: @tanmayv25 do you think it would be OK to add a `model_warmup` to each original model config directly that we copy over from model store, **instead** of doing something like this in the script? It would help with warmup for C API tests, and I don't think it would hurt any existing tests, so no need to add it conditionally.
```
if model == TFTRT:
  echo <model_warmup for "input"> >> config.pbtxt
if model == ORT_TRT:
  echo <model_warmup for "import/input:0"> >> config.pbtxt
if model == libtorch:
  echo <model_warmup for "INPUT__0" and NCHW dims> >> config.pbtxt
```

qa/L0_perf_resnet/run_test.sh
rmccorm4: Followup in other thread: https://github.com/triton-inference-server/server/pull/4420#discussion_r881866156

qa/L0_perf_resnet/test.sh
```diff
@@ -104,6 +104,22 @@ for MODEL_NAME in $OPTIMIZED_MODEL_NAMES; do
...
+        echo "    inputs {" >> ${CONFIG_PATH} && \
+        echo "        key: \"input\"" >> ${CONFIG_PATH} && \
+        echo "        value: {" >> ${CONFIG_PATH} && \
+        echo "            data_type: TYPE_FP32" >> ${CONFIG_PATH} && \
+        echo "            dims: [ 224, 224, 3 ]" >> ${CONFIG_PATH} && \
+        echo "            random_data: true" >> ${CONFIG_PATH} && \
+        echo "        }" >> ${CONFIG_PATH} && \
+        echo "    }" >> ${CONFIG_PATH} && \
+        echo "}]" >> ${CONFIG_PATH}
+    fi
```
rmccorm4, line 122: ref

tanmayv25 (approved):

rmccorm4:
Note the `L0_kibana_summary` will need some results pushed from main branch before it will generate the email correctly.

rmccorm4:
Note the `L0_kibana_summary` will need some results pushed from main branch before it will generate the email correctly.

PR #4424, opened 2022-05-24
Enable nvtx markers in pytorch backend

tanmayv25:
Added this as a part of work on a bug. Might do so for some other backends if required in future.
See linked PR here: https://github.com/triton-inference-server/pytorch_backend/pull/64

### DIFF
--- a/build.py
+++ b/build.py
@@ -627,6 +627,9 @@ def pytorch_cmake_args(images):
             cargs.append(
                 cmake_backend_enable('pytorch',
                                      'TRITON_PYTORCH_ENABLE_TORCHTRT', True))
+        cargs.append(
+            cmake_backend_enable('pytorch',
+                                 'TRITON_ENABLE_NVTX', FLAGS.enable_nvtx))
     return cargs
 
 


GuanLuo (approved):

dyastremsky (approved):

build.py
```diff
@@ -627,6 +627,9 @@ def pytorch_cmake_args(images):
             cargs.append(
                 cmake_backend_enable('pytorch',
                                      'TRITON_PYTORCH_ENABLE_TORCHTRT', True))
+        cargs.append(
+            cmake_backend_enable('pytorch',
+                                 'TRITON_ENABLE_NVTX', FLAGS.enable_nvtx))
```
rmccorm4, line 632: Does it make sense to just pass this for [every backend](https://github.com/triton-inference-server/server/blob/b6d1a6f664d687b7b86d8a17c22a6c5a3c86f078/build.py#L579-L580), and backend can accept/use it optionally?

tanmayv25: I think it is better for passing it to specific backends that support NVTX for now. I agree that adding it in common subset would be more scalable for future. But I would prefer to do when most of our backends are using nvtx markers.

rmccorm4 (approved):

PR #4427, opened 2022-05-24
Removing image layer

mc-nv:


GuanLuo (approved):
Need to be in main branch as well

PR #4428, opened 2022-05-24
Remove CUDA Binding Dependency for Java Binding Tests

dyastremsky:


dyastremsky:
@Tabrizian Previously, we provided a supported GPU example. However, that used the JavaCPP presets for CUDA and TensorRT bindings. We wanted to remove reliance on them, since they're not officially supported and users can choose how they want to allocate CUDA memory. Our main use case is CPU, so we're removing this dependency in our testing.

Tabrizian (approved):

PR #4431, opened 2022-05-25
Update windows release to use 22.03 base sw 

mc-nv:


Dockerfile.win10.min
```diff
@@ -92,10 +92,10 @@ ARG CUDA_PACKAGES="nvcc_${CUDA_MAJOR}.${CUDA_MINOR} \
                    visual_studio_integration_${CUDA_MAJOR}.${CUDA_MINOR}"
 ARG CUDA_INSTALL_ROOT_WP="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${CUDA_MAJOR}.${CUDA_MINOR}"
 
-ARG TENSORRT_VERSION=8.2.2.1
+ARG TENSORRT_VERSION=8.2.4.2
 ARG TENSORRT_ZIP=TensorRT-${TENSORRT_VERSION}.Windows10.x86_64.cuda-11.4.cudnn8.2.zip
```
rmccorm4, line 96: I know this change isn't in this PR, but just double checking that this TENSORRT_ZIP URL is fine to use with CUDNN 8.3 and CUDA 11.5 even though the link here is `*.cuda-11.4.cudnn8.2.*` ? Should the TRT ZIP be updated to CUDA 11.5 and CUDNN 8.3 if available?

rmccorm4 (approved):
LGTM other than one question

rmccorm4 (commented):
I know this change isn't in this PR, but just double checking that this TENSORRT_ZIP URL is fine to use with CUDNN 8.3 and CUDA 11.5 even though the link here is `*.cuda-11.4.cudnn8.2.*` ? Should the TRT ZIP be updated to CUDA 11.5 and CUDNN 8.3 if available?

Dockerfile.win10.min
mc-nv: Yes, it also bring my attention at the time I'm source the dependencies.
But you can review the link evaluation on below screenshot: 
image

rmccorm4: I see, LGTM - thanks!

rmccorm4 (dismissed):

tanmayv25 (approved):

GuanLuo (approved):

mc-nv:
Closing back to review, I was unable to bypass testing.

PR #4437, opened 2022-05-27
remove --build-dir   --cmake-dir  from container build

jenhaoyang:
remove --build-dir   --cmake-dir  from container build because is is invalid

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -328,7 +328,7 @@ and so you must enable them explicitly. The following build.py
 invocation builds all features and backends available on windows.
 
 ```bash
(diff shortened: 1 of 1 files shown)


GuanLuo:
Thanks for your contribution, can you help me understand the context of this change? `build.py` does recognize `--build-dir` and `--cmake-dir`, so I am not sure about what does the "invalid" refer to? I suppose the error is related to that the directory does not exist? In such case, the proper documentation fix should be pointing out that the directories should be a valid directory before specifying them in running `build.py`

rmccorm4:
Closing due to lack of activity. We can reopen this after addressing Guan's question above if needed.

PR #4441, opened 2022-05-28
fix scoop is not work

jenhaoyang:


GuanLuo:
See comment in https://github.com/triton-inference-server/server/pull/4442

PR #4442, opened 2022-05-28
scoop is not work

jenhaoyang:


rmccorm4:
Hi @jenhaoyang, 

Looks like you have a few similar PRs targeting multiple branches: [22.03](https://github.com/triton-inference-server/server/pull/4442), [22.01](https://github.com/triton-inference-server/server/pull/4441), [21.12](https://github.com/triton-inference-server/server/pull/4440)

1. Can you describe your use case in more detail?
2. Can you share the full errors you are getting and the commands you are running to reproduce these errors?
3. I'm not sure we would retroactively patch previous versions for this. There may be a better alternative or workaround. CC @mc-nv

PR #4455, opened 2022-06-02
Correct number of resource copies for R4

winstonhic:
According to "The available number of resource copies are, by default, the max across all model instances that list that resource."

"R4: 7" should change to "R4: 5"


tanmayv25 (approved):
Thanks for the fix!

PR #4461, opened 2022-06-03
Fix L0_model_config for new autocomplete functionality

nv-kmcgill53:
Updated the autofill_noplatform_success tests which now use the default-max-batch-size parameter and enable the dynamic batcher.

### DIFF
--- a/qa/L0_model_config/test.sh
+++ b/qa/L0_model_config/test.sh
@@ -233,146 +233,146 @@ rm -f $SERVER_LOG_BASE* $CLIENT_LOG
 RET=0
 
 # Run special test cases
-# for TARGET in `ls special_cases`; do
-#     SERVER_ARGS="--model-repository=`pwd`/models --strict-model-config=true"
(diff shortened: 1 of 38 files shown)


qa/L0_model_config/test.sh
```diff
@@ -233,146 +233,146 @@ rm -f $SERVER_LOG_BASE* $CLIENT_LOG
...
-            fi
-        done
-
-        if [ "$EXFOUND" == "0" ]; then
-            echo -e "*** FAILED: ${TARGET_DIR}" >> $CLIENT_LOG
-            RET=1
-        fi
-    fi
-done
+# for TARGET in `ls special_cases`; do
```
tanmayv25, line 236: Why are these test cases being commented out?

nv-kmcgill53: This was a mistake from isolating failures. Fixed.

tanmayv25 (approved):
@nv-kmcgill53  You would have to make similar changes for TensorRT models once you add dynamic batching in its auto-complete.

PR #4465, opened 2022-06-03
Fix L0_backend_python_jetson

krishung5:


Tabrizian (approved):
Feel free to merge whenever the CI passes.

qa/L0_backend_python/examples/test.sh
```diff
@@ -246,6 +246,13 @@ mkdir -p models/nobatch_auto_complete/1/
 mkdir -p models/batch_auto_complete/1/
 cp examples/auto_complete/nobatch_model.py models/nobatch_auto_complete/1/model.py
 cp examples/auto_complete/batch_model.py models/batch_auto_complete/1/model.py
+if [ "$TEST_JETSON" == "1" ]; then
+    echo -e 'name: "nobatch_auto_complete" \ninstance_group [{ kind: KIND_CPU }]' > \
+        models/nobatch_auto_complete/config.pbtxt
```
nv-kmcgill53, line 251: Is having to set the kind to CPU indicative that we are not auto completing correctly? Given no config file we should correctly determine that we should use the CPU rather than the GPU in this case.

krishung5: The auto-complete logic of Python backend doesn't determine using CPU when config file is missing. @Tabrizian Should we add this to our auto-complete logic?

krishung5: Updated: created a ticket for this [DLIS-3832](https://jirasw.nvidia.com/browse/DLIS-3832)

Tabrizian: Thanks for creating the ticket Kris. The auto-complete for model instance is done automatically from the core side and right now there is not much we can do from the backend.

PR #4466, opened 2022-06-03
Updated documentation regarding default CPU model instances

nv-kmcgill53:
- Updated documentation regarding default cpu instances 
- Added tests in L0_model_config.
- Updated L0_perf* tests to explicitly use `count:1 ` when `kind: KIND_CPU` is set.


qa/L0_model_config/autofill_noplatform_success/onnx/cpu_instance/1/model.onnx
```diff
@@ -0,0 +1,33 @@
...

+OUTPUT0
+
+var_1
+b

+OUTPUT1
+
+var_2
+B
```
GuanLuo, line 33: You can use identity model instead for the test, there is identity model that has only 1 input and output, which means we only need to provide one "expected" file (model only has one I/O permutation)

docs/model_configuration.md
```diff
@@ -570,7 +570,7 @@ GPU 0 and two execution instances on GPUs 1 and 2.
 The instance group setting is also used to enable execution of a model
 on the CPU. A model can be executed on the CPU even if there is a GPU
 available in the system. The following places two execution instances
-on the CPU.
+(the default number of cpu instances) on the CPU.
```
rmccorm4, line 573: Does this documentation still make sense if it only applies to a specific allowlist of backends ? i.e. pytorch/openvino wouldn't have 2 instances by default if I understand core PR correctly?

rmccorm4 (changes requested):

docs/model_configuration.md
rmccorm4: Maybe a statement somewhere on how "default number of cpu instances" will vary depending on the backend ?

nv-kmcgill53: Updated the documentation explaining the opt in nature of the default cpu instance count.

tanmayv25:
Instead of including the model files in the PR, can you use the pre-generated identity models as suggested by Guan instead?

qa/L0_model_config/autofill_noplatform_success/onnx/cpu_instance/1/model.onnx
nv-kmcgill53: Fixed

qa/L0_model_config/test.sh
```diff
@@ -239,6 +240,11 @@ mkdir -p autofill_noplatform_success/tensorrt/reshape_config_provided/1
 cp /data/inferenceserver/${REPO_VERSION}/qa_reshape_model_repository/plan_zero_4_float32/1/model.plan \
     autofill_noplatform_success/tensorrt/reshape_config_provided/1
 
+# Copy identity model into onnx test directories
+mkdir -p autofill_noplatform_success/onnx/cpu_instance/1
+cp -r /data/inferenceserver/${REPO_VERSION}/qa_identity_model_repository/onnx_zero_1_float16/1/model.onnx \
+    autofill_noplatform_success/onnx/cpu_instance/1
```
GuanLuo, line 246: Do the same for the TF savedmodel?

nv-kmcgill53: I might push back on this as all other savedmodels are being stored in their respective test dir. I think it would make the test directory less intuitive to have one test pull from the model repository and the rest don't. I can file a ticket for cleanup here if this is what we want moving forward?

GuanLuo (approved):

qa/L0_model_config/test.sh
GuanLuo: I see your point. Personally I would want to use model from model repository if possible, then if we need to understand the model, we can track back to the generation script. I guess you can still examine the model file directly to understand its structure, but you will need to use some tools to read in the model.

tanmayv25 (approved):

PR #4472, opened 2022-06-06
Fix L0_output_validation test

Tabrizian:
With the new PyTorch naming convention the model loads properly but the inference fails. Updated the error message check so that the test passes.

### DIFF
--- a/qa/L0_output_validation/lt_op_val_client.py
+++ b/qa/L0_output_validation/lt_op_val_client.py
@@ -65,7 +65,7 @@ class OutputValidationTest(tu.TestResultCollector):
         msg = response.json()["error"]
         self.assertTrue(
             msg.startswith(
-                "Request for unknown model: 'libtorch_name_1_float32' has no available versions"
+                "unexpected inference output 'OUTPUT__0' for model 'libtorch_name_1_float32'"
             ))
 
     # successful run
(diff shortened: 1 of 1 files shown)


GuanLuo (approved):

rmccorm4 (commented):
1. copyright update


2. (Not necessary right now to clean CI but questions in my head) Does this unit test still make sense after the [naming convention](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#special-conventions-for-pytorch-backend) update?
    - Is inference expected to fail?
    - Should the description comment above the test be updated accordingly?

qa/L0_output_validation/lt_op_val_client.py
```diff
@@ -65,7 +65,7 @@ def test_name(self):
         msg = response.json()["error"]
         self.assertTrue(
             msg.startswith(
-                "Request for unknown model: 'libtorch_name_1_float32' has no available versions"
+                "unexpected inference output 'OUTPUT__0' for model 'libtorch_name_1_float32'"
```
tanmayv25, line 68: As @rmccorm4  suggested, add a comment below L60 on how the naming convention violation is being handled.

Tabrizian: @rmccorm4:

1. Done ✔️ 
2. I have removed the test since it looks like it is no longer required after naming convention updated.

GuanLuo (approved):

PR #4473, opened 2022-06-06
Tests for setting dynamic batching scheduler during python autocomplete

nv-kmcgill53:
Python autocomplete now sets the dynamic_batching scheduler when the max_batch_size is greater than 1. This PR improves testing surrounding this feature.


tanmayv25 (dismissed):

krishung5 (approved):

nv-kmcgill53:
Updated tests to to conform to the revision of the python backend.

qa/L0_model_config/autofill_noplatform/python/conflicting_scheduler_sequence/expected
```diff
@@ -0,0 +1 @@
+Configuration already has scheduler 'sequence_batching' and will not auto-complete scheduler to dynamic_batching.
```
tanmayv25, line 1: configuration specified scheduling_choice as 'sequence_batching', but auto-complete-config function for model 'conflicting_scheduler_sequence' tries to set scheduling_choice as 'dynamic_batching'.

qa/L0_model_config/autofill_noplatform/python/dynamic_batching_no_op/expected
```diff
@@ -0,0 +1 @@
+Configuration already has scheduler 'dynamic_batching' and will not auto-complete scheduler to dynamic_batching.
```
tanmayv25, line 1: This error does not make sense. Why are we throwing error here if the configuration is already providing the scheduler choice as 'dynamic_batching'?

qa/L0_model_config/autofill_noplatform_success/python/conflicting_scheduler_ensemble/conflicting_scheduler_ensemble/expected
```diff
@@ -0,0 +1,45 @@
+name: "conflicting_scheduler_ensemble"
```
tanmayv25, line 1: I think we must expect an error here. Model trying to set scheduling_choice of 'dynamic_batching' to a model with 'ensemble_scheduling'

tanmayv25 (changes requested):
Make sure that all the failing cases are in `autofill_noplatform` and all the passing one are in 'autofill_noplatform_success'.

qa/L0_model_config/autofill_noplatform/python/dynamic_batching_no_op/expected
nv-kmcgill53: This is related to [this thread](https://github.com/triton-inference-server/python_backend/pull/165#discussion_r891619472) where the decision was made to throw an error in this case.  I will update this test in accordance with the linked thread.

qa/L0_model_config/autofill_noplatform_success/python/conflicting_scheduler_ensemble/conflicting_scheduler_ensemble/expected
nv-kmcgill53: The ensemble backend [doesn't call any autocomplete](https://github.com/triton-inference-server/core/blob/4092b56124ee3dece0eb9bce439b700e6d30ff45/src/ensemble_model.cc#L47) functions in the backend and so while the model config has the `ensemble_scheduling` set, the python backend won't call the autocomplete function of the ensemble model.  This test is to protect against any future changes in the ensemble backend where it may start to perform autocomplete functions.

qa/L0_model_config/autofill_noplatform/python/conflicting_scheduler_sequence/expected
nv-kmcgill53: To double check my understanding, you are suggesting a more comprehensive error statement, correct?

tanmayv25: correct. And this is in accordance with our other error statements for failing cases.

qa/L0_model_config/autofill_noplatform_success/python/conflicting_scheduler_ensemble/conflicting_scheduler_ensemble/expected
tanmayv25: That makes sense! I was coming up with the conditions that we check against.

qa/L0_model_config/autofill_noplatform/python/conflicting_scheduler_sequence/expected
nv-kmcgill53: fixed

qa/L0_model_config/autofill_noplatform/python/dynamic_batching_no_op/expected
nv-kmcgill53: moved to success directory

tanmayv25 (approved):

PR #4484, opened 2022-06-07
Prefix Logs with Request ID

dyastremsky:
Prefix logs with the request ID, where possible.

Related changes:
Backend: https://github.com/triton-inference-server/backend/pull/60
Core: https://github.com/triton-inference-server/core/pull/85

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2569,7 +2569,7 @@ class LifeCycleTest(tu.TestResultCollector):
         except InferenceServerException as ex:
             self.assertIn("in ensemble 'ensemble_zero_1_float32'", ex.message())
             self.assertIn(
-                "Server is stopping, scheduler for model has stopped accepting new inference requests'",
+                "Server is stopping, scheduler for model has stopped accepting new inference requests",
                 ex.message())
 
         # Wait until the results are available in user_data
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2567,9 +2567,9 @@ class LifeCycleTest(tu.TestResultCollector):
             self.assertTrue(False,
                             "expected error for new inference during shutdown")
         except InferenceServerException as ex:
+            self.assertIn("in ensemble 'ensemble_zero_1_float32'", ex.message())
             self.assertIn(
-                "in ensemble 'ensemble_zero_1_float32', [request id: <id_unknown>]" \
-                "Server is stopping, scheduler for model has stopped accepting new inference requests",
+                "Server is stopping, scheduler for model has stopped accepting new inference requests'",
                 ex.message())
 
         # Wait until the results are available in user_data
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2568,7 +2568,8 @@ class LifeCycleTest(tu.TestResultCollector):
                             "expected error for new inference during shutdown")
         except InferenceServerException as ex:
             self.assertIn(
-                "in ensemble 'ensemble_zero_1_float32', Server is stopping, scheduler for model has stopped accepting new inference requests",
+                "in ensemble 'ensemble_zero_1_float32', [request id: <id_unknown>]" \
+                "Server is stopping, scheduler for model has stopped accepting new inference requests",
                 ex.message())
 
         # Wait until the results are available in user_data
--- a/src/grpc_server.cc
+++ b/src/grpc_server.cc
@@ -3684,7 +3684,7 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
     LOG_TRITONSERVER_ERROR(
         TRITONSERVER_InferenceRequestId(irequest, &request_id),
         "unable to retrieve request ID string");
-    if ((request_id != nullptr) && (request_id[0] != '\0')) {
+    if ((request_id == nullptr) || (request_id[0] == '\0')) {
       request_id = "<id_unknown>";
     }
     if (err == nullptr) {
@@ -4112,7 +4112,7 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
     LOG_TRITONSERVER_ERROR(
         TRITONSERVER_InferenceRequestId(irequest, &request_id),
         "unable to retrieve request ID string");
-    if ((request_id != nullptr) && (request_id[0] != '\0')) {
+    if ((request_id == nullptr) || (request_id[0] == '\0')) {
       request_id = "<id_unknown>";
     }
     if (err == nullptr) {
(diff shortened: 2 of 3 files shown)


src/grpc_server.cc
```diff
@@ -3698,7 +3698,12 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
     // has initiated... completion callback will transition to
     // COMPLETE. If error go immediately to COMPLETE.
     if (err != nullptr) {
-      LOG_VERBOSE(1) << "Infer failed: " << TRITONSERVER_ErrorMessage(err);
+      const char* request_id;
+      LOG_TRITONSERVER_ERROR(
+          TRITONSERVER_InferenceRequestIdString(irequest, &request_id),
+          "unable to retrieve request ID string");
+      LOG_VERBOSE(1) << request_id
+                     << "Infer failed: " << TRITONSERVER_ErrorMessage(err);
```
rmccorm4, line 3706: I'm concerned `request_id` may spit out some garbage data or have other side effects being uninitialized here if RequestIdString call fails in any of these places:
```cpp
rmccormick@RMCCORMICK:~/triton/jira/3780-trt-reload/tritonserver$ cat test2.cpp 
#include <iostream>

int main()
{
  const char* request_id;
  std::cout << "request_id: [" << request_id << "] and some text" << std::endl;

  const char* request_id_init = "";
  std::cout << "request_id_init: [" << request_id_init << "] and some text" << std::endl;
  return 0;
}
rmccormick@RMCCORMICK:~/triton/jira/3780-trt-reload/tritonserver$ g++ test2.cpp 
rmccormick@RMCCORMICK:~/triton/jira/3780-trt-reload/tritonserver$ ./a.out 
request_id: [] and some text
request_id_init: [] and some text
```

rmccorm4 (changes requested):

src/grpc_server.cc
dyastremsky: Good catch, and thanks for the code example. I initialized to an empty string to keep it simple.

The API call doesn't seem to have a way to fail currently (the error logging was to be extra safe), so there shouldn't be a case where request_id is overridden by garbage data.

src/http_server.cc
```diff
@@ -2737,7 +2737,12 @@ HTTPAPIServer::HandleInfer(
   }
 
   if (err != nullptr) {
-    LOG_VERBOSE(1) << "Infer failed: " << TRITONSERVER_ErrorMessage(err);
+    const char* request_id = "";
```
GuanLuo, line 2740: Maybe the default can be "<id_unknown>"? If `TRITONSERVER_InferenceRequestId` ever returns error then we will print something instead of empty string

dyastremsky: Thanks! Updated.

src/http_server.cc
```diff
@@ -2899,7 +2904,7 @@ HTTPAPIServer::InferRequestClass::FinalizeResponse(
   triton::common::TritonJson::Value response_json(
       triton::common::TritonJson::ValueType::OBJECT);
 
-  const char* request_id;
+  const char* request_id = "";
```
GuanLuo, line 2907: Base on the check below (`request_id != nullptr`), should initialize to `nullptr`

GuanLuo (approved):
small change needed, otherwise looks good to me

src/http_server.cc
dyastremsky: Thanks, updated.

GuanLuo (dismissed):

CoderHam (approved):

rmccorm4 (approved):

src/grpc_server.cc
```diff
@@ -3679,6 +3679,14 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
           &state->alloc_payload_ /* response_allocator_userp */,
           InferResponseComplete, reinterpret_cast<void*>(state));
     }
+    // Get request ID for logging in case of error.
+    const char* request_id = nullptr;
+    LOG_TRITONSERVER_ERROR(
+        TRITONSERVER_InferenceRequestId(irequest, &request_id),
+        "unable to retrieve request ID string");
+    if ((request_id != nullptr) && (request_id[0] != '\0')) {
```
rmccorm4, line 3687: do we need these checks for null terminator at start / empty string? Was there an edge case?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2568,7 +2568,8 @@ def callback(user_data, result, error):
                             "expected error for new inference during shutdown")
         except InferenceServerException as ex:
             self.assertIn(
-                "in ensemble 'ensemble_zero_1_float32', Server is stopping, scheduler for model has stopped accepting new inference requests",
+                "in ensemble 'ensemble_zero_1_float32', [request id: <id_unknown>]" \
+                "Server is stopping, scheduler for model has stopped accepting new inference requests",
```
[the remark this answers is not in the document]
dyastremsky: Sure, thanks for catching this early.

dyastremsky:
I rebased this ticket.
In addition, a few new commits:
- added checks for null or empty strings (default to unknown request ID)
- updated testing to expect the request_id
- moved the request_id getter to before inference to prevent a segfault when there is a request error (server does cleanup that results in a segfault when trying to call Id() after a failed inference, even if it's before we call delete).

This ticket is now passing CI (awaiting final run addressing Ryan's comment about the assert statement, but otherwise passes).

src/http_server.cc
```diff
@@ -2910,7 +2910,7 @@ HTTPAPIServer::InferRequestClass::FinalizeResponse(
 
   const char* request_id = nullptr;
   RETURN_IF_ERR(TRITONSERVER_InferenceResponseId(response, &request_id));
-  if ((request_id != nullptr) && (request_id[0] != '\0')) {
+  if ((request_id == nullptr) || (request_id[0] == '\0')) {
```
rmccorm4, line 2913: I think this check should've remained `!=` in this spot, we don't need to add an empty or nullptr request_id to the response, right?

I would double check each of these request_id related if statements to see that they make sense in each spot in each of your PRs.

rmccorm4 (changes requested):

src/http_server.cc
dyastremsky: Thanks so much for your careful reviews. Everything should be good now.

src/http_server.cc
```diff
@@ -2737,7 +2745,8 @@ HTTPAPIServer::HandleInfer(
   }
 
   if (err != nullptr) {
-    LOG_VERBOSE(1) << "Infer failed: " << TRITONSERVER_ErrorMessage(err);
+    LOG_VERBOSE(1) << "[request id: " << request_id << "]"
```
rmccorm4, line 2748: Does it make more sense to use a common `GetRequestId()` for both servers in this PR too? See this comment: https://github.com/triton-inference-server/backend/pull/60/files#r912262759

rmccorm4: Same as other discussion, please add spaces after the "]" for consistency

rmccorm4: @dyastremsky

dyastremsky: Sure, thanks. Done.

dyastremsky: Sure, thanks. Done.

CoderHam (approved):

rmccorm4 (approved):

PR #4494, opened 2022-06-08
Add L0_backend_bls test

krishung5:


qa/L0_backend_bls/test.sh
```diff
@@ -0,0 +1,113 @@
...
+
+# Install torch
+pip3 uninstall -y torch
+pip3 install torch==1.9.0+cu111 -f https://download.pytorch.org/whl/torch_stable.html
+
+git clone https://github.com/triton-inference-server/bls_backend -b $BLS_BACKEND_REPO_TAG
+cp -r ./models ./all_models/
+cp -r ./bls_backend/models/* ./all_models/
+
+# Run the server without having required pytorch and add_sub models.
```
tanmayv25, line 48: Instead of separate server runs, how about launching server in explicit mode and try out these different test combinations by loading these dependent models one at a time?

tanmayv25 (changes requested):

qa/L0_backend_bls/test.sh
krishung5: Updated the test to use explicit mode.

qa/L0_backend_bls/test.sh
```diff
@@ -0,0 +1,165 @@
...
+TRITON_BACKEND_REPO_TAG=${TRITON_BACKEND_REPO_TAG:="main"}
+TRITON_CORE_REPO_TAG=${TRITON_CORE_REPO_TAG:="main"}
+TRITON_COMMON_REPO_TAG=${TRITON_COMMON_REPO_TAG:="main"}
+
+SERVER=/opt/tritonserver/bin/tritonserver
+source ../common/util.sh
+
+RET=0
+
+# Client build requires recent version of CMake (FetchContent required)
```
GuanLuo, line 39: Backend build

qa/L0_backend_bls/test.sh
```diff
@@ -0,0 +1,165 @@
...
+grep "PASS" $CLIENT_LOG
+if [ $? -ne 0 ]; then
+    echo -e "\n***\n*** bls_test.py FAILED. \n***"
+    cat $CLIENT_LOG
+    cat $SERVER_LOG
+    RET=1
+fi
+set -e
+
+# Send a request without model 'addsub_python'.
```
GuanLuo, line 98: I don't think we need extensive testing for a example backend, just check if it runs properly with correct set up

qa/L0_backend_bls/test.sh
krishung5: Thanks for the catch. Updated.

GuanLuo (approved):

Tabrizian (approved):

qa/L0_backend_bls/test.sh
krishung5: Update: Removed extensive testing. Please review.

tanmayv25 (approved):

PR #4500, opened 2022-06-08
Expose --model-load-thread-count for model load thread pool (no boost)

rmccorm4:
Allows users to define desired concurrency level based on number of models they expect to load concurrently via server CLI for model load thread pool in https://github.com/triton-inference-server/core/pull/83.

[common] Adds custom ThreadPool implementation in common: https://github.com/triton-inference-server/common/pull/60
[core] Use thread pool for loading models in model_repository_manager.cc: https://github.com/triton-inference-server/core/pull/86
[server] Expose num threads to server CLI: https://github.com/triton-inference-server/server/pull/4500


src/main.cc
```diff
@@ -1824,6 +1835,10 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
       TRITONSERVER_ServerOptionsSetBufferManagerThreadCount(
           loptions, std::max(0, buffer_manager_thread_count)),
       "setting buffer manager thread count");
+  FAIL_IF_ERR(
+      TRITONSERVER_ServerOptionsSetModelLoadThreadCount(
+          loptions, std::max(1u, model_load_thread_count)),
+      "setting model load thread count");
```
nv-kmcgill53, line 1841: This question is a duplicate of [this](https://github.com/triton-inference-server/core/pull/86#discussion_r892926372). Question: Since this is a user provided variable, should we set an upper bound or is that being too opinionated? The user could potentially set thread count to -1 and lock up the system by generating a large number of threads. What would be a reasonable max here, I'm thinking 2x or 3x of `std::thread::hardware_concurrency()`

GuanLuo (dismissed):

src/main.cc
```diff
@@ -1282,6 +1287,9 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
   int32_t repository_poll_secs = repository_poll_secs_;
   int64_t pinned_memory_pool_byte_size = 1 << 28;
   int32_t buffer_manager_thread_count = 0;
+  // hardware_concurrency() can return 0 if not implemented, so check it.
```
CoderHam, line 1290: "hardware_concurrency() returns 0 if not well defined or not computable"

Remove 'so check it'

CoderHam (changes requested):

CoderHam (approved):

tanmayv25 (approved):

PR #4510, opened 2022-06-13
Add libgtest-dev as system package

gvanhoy:
I could not get a debug build to work with build.py. This fixed it for me. Perhaps a sub-project or third-party module used to install this and now it does not.


rmccorm4:
Hi @gvanhoy,

Can you check if this is still an issue? I believe this would have been fixed by this PR: https://github.com/triton-inference-server/third_party/pull/19

PR #4514, opened 2022-06-14
Fix Trace Documentation Rendering

dyastremsky:
GitHub's markdown renders the trace_file examples as math, so the text is difficult to read. I wasn't able to find a way to escape the dollar signs, so I wrapped them as an in-line code block for readability.

### DIFF
--- a/docs/protocol/extension_trace.md
+++ b/docs/protocol/extension_trace.md
@@ -73,7 +73,7 @@ trace settings are defined:
 
 - "trace_file" : the file where the trace output will be saved. If
 "log frequency" is set, this will be the prefix of the files to save the
-trace output, resulting files in name "${trace_file}.0", "${trace_file}.1"...,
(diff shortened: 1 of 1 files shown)


docs/protocol/extension_trace.md
```diff
@@ -72,25 +72,25 @@ setting value, or an array of $string for some settings. Currently the following
 trace settings are defined:
 
 - "trace_file" : the file where the trace output will be saved. If
-"log frequency" is set, this will be the prefix of the files to save the
-trace output, resulting files in name "${trace_file}.0", "${trace_file}.1"...,
-see trace setting "log frequency" below for detail.
+"log_frequency" is set, this will be the prefix of the files to save the
+trace output, resulting files in name `${trace_file}.0", "${trace_file}.1", ...`,
```
jbkyang-nvi, line 76: I think you can just do
"\${trace_file}.0", "\${trace_file}.1"...,

jbkyang-nvi (approved):

PR #4518, opened 2022-06-17
Protect shared_memory_map_ from concurrent access

kthui:
Protect GetMemoryInfo and GetCUDAHandle with mutex to avoid race condition when the map information is being accessed.

### DIFF
--- a/src/shared_memory_manager.cc
+++ b/src/shared_memory_manager.cc
@@ -334,6 +334,9 @@ SharedMemoryManager::GetMemoryInfo(
     const std::string& name, size_t offset, void** shm_mapped_addr,
     TRITONSERVER_MemoryType* memory_type, int64_t* device_id)
 {
+  // protect shared_memory_map_ from concurrent access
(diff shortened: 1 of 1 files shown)


tanmayv25:
@CoderHam  should also take a look.

Tabrizian (approved):
@CoderHam please review these changes too. Thanks.

CoderHam (approved):
LGTM. I believe this may have performance implications. Have we tested the impact of these changes on the shm variants of the L0_perf tests?

What is the observed impact on throughput and latency?

kthui:
@CoderHam 
There is an observable impact on performance, for example:
```
L0_perf_nomodel, system, 16mb_latency_http_shm:
   Before: 
      Inferences/Second vs. Client p95 Batch Latency
      Concurrency: 1, throughput: 20.9407 infer/sec, latency 51349 usec
   After:
      Inferences/Second vs. Client p95 Batch Latency
      Concurrency: 1, throughput: 19.5506 infer/sec, latency 55139 usec
```

PR #4534, opened 2022-06-21
Remove nvidia-ml deps for multi_server

CoderHam:
- We do not need to link against the nvidia-ml library for multi_server


CoderHam:
cc @nv-kmcgill53 can you try to build a version of 22.02 on Jetson with this branch instead of r22.02?

PR #4543, opened 2022-06-23
Sagemaker invoke 507

nskool:
The changes in this commit [link](https://github.com/triton-inference-server/server/commit/2965840a637898c270f80e592715292f41fe49cd) enable SM endpoint to throw a custom '507' error when an OOM is encountered on the GPU. 

However, there are scenarios where the models maybe loaded, but the 'inference' request fails on GPU OOM. e.g.
```
RuntimeError: CUDA out of memory. Tried to allocate 2.00 MiB (GPU 0; 14.76 GiB total capacity; 11.55 GiB already allocated; 15.75 MiB free; 11.71 GiB reserved in total by PyTorch
```

This PR enables the inference request to throw a 507 on failure as well, so the platform can react accordingly.

### DIFF
--- a/src/sagemaker_server.cc
+++ b/src/sagemaker_server.cc
@@ -226,8 +226,8 @@ SagemakerAPIServer::Handle(evhtp_request_t* req)
             evhtp_send_reply(req, EVHTP_RES_NOTFOUND); /* 404*/
             return;
           }
-
-          HandleInfer(req, multi_model_name, model_version_str_);
+          LOG_VERBOSE(1) << "SM MME Custom Invoke Model" << std::endl;
+          SageMakerMMEHandleInfer(req, multi_model_name, model_version_str_);
           return;
         }
         if (action.empty()) {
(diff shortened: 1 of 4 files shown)


src/sagemaker_server.cc
```diff
@@ -334,6 +334,281 @@ SagemakerAPIServer::ParseSageMakerRequest(
...
+        "INFER_RESPONSE_COMPLETE", TraceManager::CaptureTimestamp());
+  }
+#endif  // TRITON_ENABLE_TRACING
+
+  if (err == nullptr) {
+    evthr_defer(infer_request->thread_, OKReplyCallback, infer_request);
+  } else {
+    EVBufferAddErrorJson(infer_request->req_->buffer_out, err);
+    TRITONSERVER_ErrorDelete(err);
+    if (SageMakerMMECheckOOMError(err) == true) {
```
GuanLuo, line 387: If true, should invoke `BADReplyCallback507` instead

nskool: Thanks, updated, it was a commit mistake. Could you run the CI?

GuanLuo (approved):

PR #4544, opened 2022-06-23
Fix L0_model_config

nv-kmcgill53:
Added test variant for the no_batch_hint for tensorflow autocomplete. Previously, the order of the I/Os mattered when determining the batch hint. This change proves the order doesn't matter when determining batch hinting.


tanmayv25:
What do you mean by "previously, the order of the I/Os mattered when determining the batch hint"?

nv-kmcgill53:
>What do you mean by "previously, the order of the I/Os mattered when determining the batch hint"?

Suppose we have two model configs for the same model, model_0.

Config A: 
```
output [
  {
    name: "OUTPUT0"
    data_type: TYPE_INT8
  },
  {
    name: "OUTPUT1"
    dims: [ -1, 16 ]
  }
]
```
Config B:
```
output [
  {
    name: "OUTPUT1"
    dims: [ -1, 16 ]
  },
  {
    name: "OUTPUT0"
    data_type: TYPE_INT8
  }
]
```

Previously Config A would not batch since the explicit hint at "no batching", OUTPUT1 came after the autocompleted output, OUTPUT0; but, Config B would incorrectly set batching to true as the explicit dimensional hint OUTPUT1 was processed before OUTPUT0. So Config B caused an error when it came to interpreting the dimension sizes. The error was finding 3 dims in OUTPUT1 when it expected only 2 for the underlying model.

The tensorflow_backend changes fix this error and the added variant protects regression.

tanmayv25 (approved):

PR #4552, opened 2022-06-27
 add pa ground truth

debermudez:
Adding a ground truth test to lock down PA behavior


qa/L0_perf_analyzer_ground_truth/test.sh
```diff
@@ -0,0 +1,142 @@
...
+        RET=1
+    fi
+    if [ $(cat ${CLIENT_LOG} |  grep "${ERROR_STRING}" | wc -l) -ne 0 ]; then
+        cat ${CLIENT_LOG}
+        echo -e "\n***\n*** Test Failed\n***"
+        RET=1
+    fi
+}
+
+function check_performance {
```
tgerdesnv, line 62: Does this function take args? If so, document

qa/L0_perf_analyzer_ground_truth/test.sh
```diff
@@ -0,0 +1,142 @@
...
+	echo "[PERMUTATION] Protocol=${protocol} Model=${model}"
+	echo "================================================================"
+
+        ${PERF_ANALYZER} -v -i ${protocol} -m ${model} -f ${OUTPUT_FILE} | tee ${CLIENT_LOG} 2>&1
+        check_perf_analyzer_error $?
+
+    done;
+done;
+
+check_performance ${OUTPUT_FILE} ${EXPECTED_RESULT} ${TOLERANCE}
```
tgerdesnv, line 125: part of this file seems to be set up to handle multiple tests/cases, and part of it doesn't.  It seems like this should be inside of the loop at line 121?

qa/L0_perf_analyzer_ground_truth/test.sh
```diff
@@ -0,0 +1,142 @@
...
+
+# Cleanup
+kill $SERVER_PID
+wait $SERVER_PID
+
+if [ $RET -eq 0 ]; then
+  echo -e "\n***\n*** Test Passed\n***"
+else
+  echo "=== START SERVER LOG ==="
+  cat ${SERVER_LOG}
```
tgerdesnv, line 137: should client log be printed here on failure as well?

qa/python_models/add_sub_ground_truth/model.py
```diff
@@ -0,0 +1,74 @@
...
+        time.sleep(0.01)
+        output0_dtype = self.output0_dtype
+        output1_dtype = self.output1_dtype
+
+        responses = []
+        for request in requests:
+            in_0 = pb_utils.get_input_tensor_by_name(request, "INPUT0")
+            in_1 = pb_utils.get_input_tensor_by_name(request, "INPUT1")
+            if in_0.as_numpy().dtype.type is np.bytes_ or in_0.as_numpy(
+            ).dtype == np.object_:
```
tgerdesnv, line 61: Is there overhead in this math? Our goal is to have the only overhead be the time.sleep function above. Any reason we don't just directly return the inputs into the outputs (or return zeros or whatever takes the least amount of overhead)?

qa/python_models/add_sub_ground_truth/model.py
```diff
@@ -0,0 +1,74 @@
...
+        self.output0_dtype = pb_utils.triton_string_to_numpy(
+            output0_config['data_type'])
+        self.output1_dtype = pb_utils.triton_string_to_numpy(
+            output1_config['data_type'])
+
+    def execute(self, requests):
+        """ This function is called on inference request.
+        """
+
+        time.sleep(0.01)
```
tgerdesnv, line 52: follow-on stories if not handled in this story:
- Ability to programmatically set the DELAY
- Ability to programmatically set the VARIANCE and use that to add noise to the delay
- test using c_api to minimize all overhead and just test PA
- add more tests using this infrastructure
  - High variance = no stabilization
  - very low delay
  - high delay 
  - dynamic batching in the model (+ higher concurrency)

qa/L0_perf_analyzer_ground_truth/test.sh
tgerdesnv: Also, if every line is going to just be python, why not add a 'check_performance.py' script instead?

tgerdesnv (changes requested):

qa/python_models/add_sub_ground_truth/model.py
debermudez: added tests

debermudez:
@GuanLuo I dont own the CI for this repo so I was trying to make as few changes there as possible. If you think that is the correct decision, I can do that or we can loop in @mc-nv to handle it.

mc-nv:
> @GuanLuo I dont own the CI for this repo so I was trying to make as few changes there as possible. If you think that is the correct decision, I can do that or we can loop in @mc-nv to handle it.

mc-nv:
> > @GuanLuo I dont own the CI for this repo so I was trying to make as few changes there as possible. If you think that is the correct decision, I can do that or we can loop in @mc-nv to handle it.

@debermudez  seems comment from @GuanLuo is talking about structure in current repo and location for a new files and  about user experience because from the CI perspective job will call out the instructions no matter the location is.

Tabrizian (commented):
Can we use an identity model instead of the add_sub model? You can find an example identity model here:

https://github.com/triton-inference-server/server/blob/main/qa/python_models/identity_fp32/model.py

debermudez:
@Tabrizian I updated the model for testing. It does not make a large functional difference in measurements but its good to strip out as much noise as possible.

Tabrizian (approved):

tgerdesnv (approved):

PR #4553, opened 2022-06-28
Increased timeout from 3600 to 4000

CoderHam:
Fixes failure for L0_infer on jetson due to slight increase in overall time to load all models


CoderHam:
@mc-nv In case you want to cherry-picked into 22.06

tanmayv25 (approved):

mc-nv (approved):

PR #4568, opened 2022-06-30
Fix collision test in L0_lifecycle: use separate server logs

GuanLuo:


tanmayv25 (approved):

dyastremsky (approved):

CoderHam (approved):

rmccorm4 (approved):
Just to add context if this PR gets looked at in the future to see why this change was made:

> Two servers are being run for the collision tests, and the server log will be mangled if they write to the same file, so use separate filenames for each pair of servers to get consistent logs for error/output checking.

PR #4569, opened 2022-06-30
Add negative test to check response cache fails with decoupled mode

rmccorm4:


qa/L0_response_cache/test.sh
```diff
@@ -42,6 +43,25 @@ if [ $? -ne 0 ]; then
...
+source ../common/util.sh
+run_server
+if [ "$SERVER_PID" != "0" ]; then
+    echo -e "\n***\n*** Failed: $SERVER started successfully when it was expected to fail\n***"
+    cat $SERVER_LOG
+    RET=1
+
+    kill $SERVER_PID
+    wait $SERVER_PID
+fi
```
nv-kmcgill53, line 63: Add a simple check to prove the server failed with "Response Cache does not currently support the 'decoupled' model transaction policy" and not some other error

nv-kmcgill53 (dismissed):

qa/L0_response_cache/test.sh
```diff
@@ -32,16 +32,46 @@ UNIT_TEST=./response_cache_test
...
 
+# UNIT TEST
 set +e
 export CUDA_VISIBLE_DEVICES=0
 LD_LIBRARY_PATH=/opt/tritonserver/lib:$LD_LIBRARY_PATH $UNIT_TEST >>$TEST_LOG 2>&1
 if [ $? -ne 0 ]; then
     cat $TEST_LOG
     echo -e "\n***\n*** Response Cache Unit Test Failed\n***"
-    RET=1
+    #RET=1
```
tanmayv25, line 42: I wonder why this line is commented?

tanmayv25 (approved):

PR #4592, opened 2022-07-05
[DO NOT MERGE] error codes 21.02

jbkyang-nvi:


src/servers/http_server.cc
```diff
@@ -1111,6 +1112,22 @@ class HTTPAPIServer : public HTTPServerImpl {
   re2::RE2 cudasharedmemory_regex_;
 };
 
+void HTTPAPIServer::HandleInferErrors(evhtp_request_t* req, TRITONSERVER_Error* err) {
+  const char* message = TRITONSERVER_ErrorMessage(err);
+  std::string error_string(message);
+  EVBufferAddErrorJson(req->buffer_out, err);
+  if (error_string.find("GPU error occured") == std::string::npos) {
```
GuanLuo, line 1119: `!=`?

GuanLuo: Should check with the customer on the exact error message to check

src/servers/http_server.cc
```diff
@@ -2325,13 +2342,10 @@ HTTPAPIServer::HandleInfer(
 
   if (err != nullptr) {
     LOG_VERBOSE(1) << "Infer failed: " << TRITONSERVER_ErrorMessage(err);
-    EVBufferAddErrorJson(req->buffer_out, err);
-    evhtp_send_reply(req, EVHTP_RES_BADREQ);
+    HandleInferErrors(req, err);
```
GuanLuo, line 2345: Control reaches here only if there is error before the model executes the request, so also need the check after `TRITONSERVER_InferenceResponseError()` where the error during execution will be reported to

src/servers/http_server.cc
```diff
@@ -2325,13 +2342,10 @@ HTTPAPIServer::HandleInfer(
 
   if (err != nullptr) {
     LOG_VERBOSE(1) << "Infer failed: " << TRITONSERVER_ErrorMessage(err);
```
rmccorm4, line 2344: You can move this into the helper too, and doesn't need to call TRITONSERVER_ErrorMessage again, you can re-use `message`.

rmccorm4 (commented):
Agree with Guan's comments as well.

src/servers/http_server.cc
jbkyang-nvi: I confirmed with customer

Dockerfile.sdk
```diff
@@ -126,8 +129,7 @@ RUN pip3 install --upgrade numpy pillow && \
     xargs pip3 install --upgrade
 
 # Install DCGM
-RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb && \
-    dpkg -i datacenter-gpu-manager_${DCGM_VERSION}_amd64.deb
+RUN apt-get update && apt-get install -y datacenter-gpu-manager=1:${DCGM_VERSION}
```
jbkyang-nvi, line 132: can ignore these changes. These are for building the container with the recent cuda changes.

src/servers/http_server.cc
```diff
@@ -2452,9 +2497,16 @@ HTTPAPIServer::InferRequestClass::InferResponseComplete(
...
-    TRITONSERVER_ErrorDelete(err);
-    evthr_defer(infer_request->thread_, BADReplyCallback, infer_request);
+    std::string msg(TRITONSERVER_ErrorMessage(err));
+    if (InferErrorIsInternalError(msg)) {
+      EVBufferAddErrorJson(infer_request->req_->buffer_out, err);
+      TRITONSERVER_ErrorDelete(err);
+      evthr_defer(infer_request->thread_, FAILReplyCallback, infer_request);
+    } else {
+      EVBufferAddErrorJson(infer_request->req_->buffer_out, err);
+      TRITONSERVER_ErrorDelete(err);
```
GuanLuo, line 2507: Seems like you can move this two lines out of the conditional:
```
EVBufferAddErrorJson...
TRITONSERVER_ErrorDelete...
if (InferErrorIsInternalError(msg)) {
   FAILReplyCallback
} else {
   BADReplyCallback
}
```

rmccorm4:
LGTM. 

Not approving to avoid accidental merge.

Dockerfile.sdk
jbkyang-nvi: @pragyaarora

rmccorm4:
Closing, not needed anymore after reviews.

PR #4602, opened 2022-07-07
Fix L0_dyna_sequence_batcher and L0_dyna_implicit_state

tanmayv25:
The tests are sensitive to the number of instances created as they change the manner requests get batched.

### DIFF
--- a/qa/L0_dyna_sequence_batcher/test.sh
+++ b/qa/L0_dyna_sequence_batcher/test.sh
@@ -69,11 +69,16 @@ rm -fr *.log *.serverlog
 
 # models
 rm -fr models && mkdir models
-cp -r ${DATADIR}/$MODEL_REPOSITORY/* models/.
+for MODEL in ${DATADIR}/$MODEL_REPOSITORY/* ; do
+    cp -r $MODEL models/. && \
+        (cd models/$(basename $MODEL) && \
+            sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 1/" config.pbtxt)
+done
 
 # Implicit state models for custom backend do not exist.
 if [ $IMPLICIT_STATE == "0" ]; then
     cp -r ../custom_models/custom_dyna_sequence_int32 models/.
+    sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 1/" models/custom_dyna_sequence_int32/config.pbtxt
     # Construct custom dyna_sequence_model with STRING sequence ID. Copy model and edit config.pbtxt
     cp -r models/custom_dyna_sequence_int32 models/custom_string_dyna_sequence_int32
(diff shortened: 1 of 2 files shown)


nv-kmcgill53:
What was the root cause of the batches being different?

tanmayv25:
One of the failing test was `test_multi_sequence`.  
With a single instance all the requests received when the inference is in progress gets batched together for when this instance is made available. However, with two instances, the second instance do not wait for the inference to complete. It directly forms the batch(mostly a smaller one) and goes into its own execution.

nv-kmcgill53 (approved):

qa/L0_dyna_sequence_batcher/test.sh
```diff
@@ -69,11 +69,16 @@ rm -fr *.log *.serverlog
 
 # models
 rm -fr models && mkdir models
-cp -r ${DATADIR}/$MODEL_REPOSITORY/* models/.
+for MODEL in ${DATADIR}/$MODEL_REPOSITORY/* ; do
+    cp -r $MODEL models/. && \
+        (cd models/$(basename $MODEL) && \
+            sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 1/" config.pbtxt)
```
Tabrizian, line 75: I thought we only increased the default instance count to 2 for TRT, ONNX, and TF. Are we setting the default instance count to two for all the backends?

nv-kmcgill53: We decided in the [PR](https://github.com/triton-inference-server/core/pull/82) to go for an opt in approach. The only backends which are opted in are tensorflow and onnxruntime.

Tabrizian: I see. I think I commented on the wrong place. I was wondering why we need to set instance count of 1 for custom backend models? Is it just to be safe?

qa/L0_dyna_sequence_batcher/test.sh
```diff
@@ -69,11 +69,16 @@ rm -fr *.log *.serverlog
...
+done
 
 # Implicit state models for custom backend do not exist.
 if [ $IMPLICIT_STATE == "0" ]; then
     cp -r ../custom_models/custom_dyna_sequence_int32 models/.
+    sed -i "s/kind: KIND_CPU/kind: KIND_CPU\\ncount: 1/" models/custom_dyna_sequence_int32/config.pbtxt
```
Tabrizian, line 81: I meant whether setting the instance count to 1 is required here or not.

tanmayv25: This has been added just to be safe. The motivation is derived from the L0_sequence_batcher.

Tabrizian (approved):

PR #4621, opened 2022-07-11
Cat server log on failure in tf_unknown_rank

rmccorm4:


mc-nv:
@rmccorm4 , @Tabrizian  we spoke about it and it can be enabled on a CI side.

rmccorm4:
> we spoke about it and it can be enabled on a CI side.

Sounds great, looking forward to the change for more generic debugging.

However, for running CI tests locally during development, I think this change will still be useful.

PR #4633, opened 2022-07-13
Fix expected byte size calculation

kthui:
When the data type is BYTE, `JsonBytesArrayByteSize` function is called to calculate the expected byte size. The calculation uses depth-first search (DFS) to finds each characters from the input, no matter how "deep" they are nested within the dimensions, and greedily sums the byte size along the search. The function calls itself iteratively to complete the DFS.

The original search function will initialize the byte size to 0 whenever it is called. This is incorrect as when the DFS switches branch, the byte size from the previous branch is discarded. The updated search function fixes the re-initialization bug.

For example, the original function calculates `['a', 'b']` correctly as there is no nested array, and also `[['a', 'b']]` correctly as there is no branch switch, but `[['a'], ['b']]` incorrectly when switching branch from `['a']` to `['b']`, which is fixed in this PR.

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -239,10 +239,13 @@ AllocEVBuffer(const size_t byte_size, evbuffer** evb, void** base)
   return nullptr;  // success
 }
 
+// Recursively adds to byte_size from multi dimensional data input
 TRITONSERVER_Error*
 JsonBytesArrayByteSize(
     triton::common::TritonJson::Value& tensor_data, size_t* byte_size)
 {
+  *byte_size = 0;
+
   for (size_t i = 0; i < tensor_data.ArraySize(); i++) {
     triton::common::TritonJson::Value el;
     RETURN_IF_ERR(tensor_data.At(i, &el));
@@ -251,7 +254,9 @@ JsonBytesArrayByteSize(
     TRITONSERVER_Error* assert_err =
         el.AssertType(triton::common::TritonJson::ValueType::ARRAY);
     if (assert_err == nullptr) {
-      RETURN_IF_ERR(JsonBytesArrayByteSize(el, byte_size));
+      size_t byte_size_;
+      RETURN_IF_ERR(JsonBytesArrayByteSize(el, &byte_size_));
+      *byte_size += byte_size_;
     } else {
       // Serialized data size is the length of the string itself plus
       // 4 bytes to record the string length.
@@ -2419,7 +2424,6 @@ HTTPAPIServer::EVBufferToInput(
               "Unable to parse 'data'");
 
           if (dtype == TRITONSERVER_TYPE_BYTES) {
-            byte_size = 0;  // initialize for its recursive computation
             RETURN_IF_ERR(JsonBytesArrayByteSize(tensor_data, &byte_size));
           } else {
             byte_size = element_cnt * TRITONSERVER_DataTypeByteSize(dtype);
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -2419,7 +2419,7 @@ HTTPAPIServer::EVBufferToInput(
               "Unable to parse 'data'");
 
           if (dtype == TRITONSERVER_TYPE_BYTES) {
-            byte_size = 0;
+            byte_size = 0;  // initialize for its recursive computation
             RETURN_IF_ERR(JsonBytesArrayByteSize(tensor_data, &byte_size));
           } else {
             byte_size = element_cnt * TRITONSERVER_DataTypeByteSize(dtype);
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -243,8 +243,6 @@ TRITONSERVER_Error*
 JsonBytesArrayByteSize(
     triton::common::TritonJson::Value& tensor_data, size_t* byte_size)
 {
-  *byte_size = 0;
-
   for (size_t i = 0; i < tensor_data.ArraySize(); i++) {
     triton::common::TritonJson::Value el;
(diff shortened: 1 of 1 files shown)


src/http_server.cc
```diff
@@ -2421,6 +2419,7 @@ HTTPAPIServer::EVBufferToInput(
               "Unable to parse 'data'");
 
           if (dtype == TRITONSERVER_TYPE_BYTES) {
+            byte_size = 0;
```
rmccorm4, line 2422: Should the be `*byte_size = 0` ? byte_size is of type `size_t*`

rmccorm4, line 2422: Can you add a comment on why this initialization is important here?

kthui: Yes, added.

kthui:
Rebased.

rmccorm4 (dismissed):

krishung5 (approved):

GuanLuo (dismissed):

src/http_server.cc
```diff
@@ -243,8 +243,6 @@ TRITONSERVER_Error*
 JsonBytesArrayByteSize(
```
tanmayv25, line 243: Can you add a short description of this function?
Because the name doesn't do the justice as it can recursively accumulate the byte size.

src/http_server.cc
```diff
@@ -243,8 +243,6 @@ TRITONSERVER_Error*
...
     triton::common::TritonJson::Value& tensor_data, size_t* byte_size)
 {
-  *byte_size = 0;
```
tanmayv25, line 246: Actual I would prefer to keep the function like this and accumulate byte_size for elements in a separate variable in L254

tanmayv25: In the end you can set byte_size to the accumulated byte size.

kthui: How about rename the recursive function to `FindJsonBytesArrayByteSizeHelper`, and wrap it with `FindJsonBytesArrayByteSize` that will initialize `*byte_size = 0;`?

LIke this:
```
TRITONSERVER_Error*
FindJsonBytesArrayByteSizeHelper(
    triton::common::TritonJson::Value& tensor_data, size_t* byte_size)
{
  for (size_t i = 0; i < tensor_data.ArraySize(); i++) {
    triton::common::TritonJson::Value el;
    RETURN_IF_ERR(tensor_data.At(i, &el));

    // Recurse if not last dimension...
    TRITONSERVER_Error* assert_err =
        el.AssertType(triton::common::TritonJson::ValueType::ARRAY);
    if (assert_err == nullptr) {
      RETURN_IF_ERR(JsonBytesArrayByteSize(el, byte_size));
    } else {
      // Serialized data size is the length of the string itself plus
      // 4 bytes to record the string length.
      const char* str;
      size_t len = 0;
      RETURN_MSG_IF_ERR(
          el.AsString(&str, &len), "Unable to parse JSON bytes array");
      *byte_size += len + sizeof(uint32_t);
    }
    TRITONSERVER_ErrorDelete(assert_err);
  }

  return nullptr;  // success
}

TRITONSERVER_Error*
FindJsonBytesArrayByteSize(
    triton::common::TritonJson::Value& tensor_data, size_t* byte_size)
{
  *byte_size = 0;
  return FindJsonBytesArrayByteSizeHelper(tensor_data, byte_size);
}

... skip lines ...
          if (dtype == TRITONSERVER_TYPE_BYTES) {
            RETURN_IF_ERR(FindJsonBytesArrayByteSize(tensor_data, &byte_size));
          } else {
            byte_size = element_cnt * TRITONSERVER_DataTypeByteSize(dtype);
          }
```

kthui: I see. Do you mean this? 
```
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -243,6 +243,8 @@ TRITONSERVER_Error*
 JsonBytesArrayByteSize(
     triton::common::TritonJson::Value& tensor_data, size_t* byte_size)
 {
+  *byte_size = 0;
+
   for (size_t i = 0; i < tensor_data.ArraySize(); i++) {
     triton::common::TritonJson::Value el;
     RETURN_IF_ERR(tensor_data.At(i, &el));
@@ -251,7 +253,9 @@ JsonBytesArrayByteSize(
     TRITONSERVER_Error* assert_err =
         el.AssertType(triton::common::TritonJson::ValueType::ARRAY);
     if (assert_err == nullptr) {
-      RETURN_IF_ERR(JsonBytesArrayByteSize(el, byte_size));
+      size_t byte_size_;
+      RETURN_IF_ERR(JsonBytesArrayByteSize(el, &byte_size_));
+      *byte_size += byte_size_;
     } else {
       // Serialized data size is the length of the string itself plus
       // 4 bytes to record the string length.
@@ -2419,7 +2423,6 @@ HTTPAPIServer::EVBufferToInput(
               "Unable to parse 'data'");
 
           if (dtype == TRITONSERVER_TYPE_BYTES) {
-            byte_size = 0;  // initialize for its recursive computation
             RETURN_IF_ERR(JsonBytesArrayByteSize(tensor_data, &byte_size));
           } else {
             byte_size = element_cnt * TRITONSERVER_DataTypeByteSize(dtype);
```

tanmayv25: Yes. Something like this.

kthui: Ok, updated.

rmccorm4 (approved):

GuanLuo:
@kthui I missed the part about testing, can you add test to L0_http to avoid regression?

PR #4634, opened 2022-07-13
Add an example to demonstrate shape tensor handling

tanmayv25:
Read about the shape tensors here: https://docs.nvidia.com/deeplearning/tensorrt/developer-guide/index.html#shape_tensor_io

And our test here: https://github.com/triton-inference-server/server/blob/main/qa/L0_trt_shape_tensors/test.sh


docs/model_configuration.md
```diff
@@ -454,16 +454,45 @@ value. For the above example, an inference request must provide inputs
 with the following shapes.
 
 ```
-  "input0": [ x, -1]
-  "input1": [ 1 ]
-  "output0": [ x, -1]
+  "input0": [ x, 1, 3]
+  "input1": [ 3 ]
```
GuanLuo, line 458: Why `input1` has shape [3] instead of [2] (config)

GuanLuo: Meaning config shape + 1 (if batching)?

docs/model_configuration.md
```diff
@@ -454,16 +454,45 @@ value. For the above example, an inference request must provide inputs
...
+```
+
+Assuming these requests get batched together would be delivered to the
+model as:
+
+
+```
+Batched Requests to model:
+input0: [[[1,2,3]], [[4,5,6]], [[7,8,9]], [[10,11,12]]] <== shape of this tensor [4,1,3]
+input1: [4, 4, 6] <== shape of this tensor [3]
```
GuanLuo, line 492: Should explain more about shape tensor is only for TensorRT and in what case (example?) it will be needed

docs/model_configuration.md
tanmayv25: That is because of the batching. Let me try to elaborate on that.

docs/model_configuration.md
tanmayv25: Sure!

dyastremsky (approved):

rmccorm4 (approved):

GuanLuo (approved):

PR #4635, opened 2022-07-13
Resolve disable-auto-complete-config and strict-model-config conflict

nv-kmcgill53:
This PR adds a warning on startup for the user if they are using the now deprecated `--strict-model-config` flag. Additionally, we add a warning that the `--disable-auto-complete-config` flag will be used instead of the `--strict-model-config` if both are provided and they contradict. In the event the two options are in agreement, then the user only gets the deprecation warning.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -1289,6 +1289,8 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
   std::string server_id("triton");
   std::set<std::string> model_repository_paths;
   bool exit_on_error = true;
+  bool disable_auto_complete_config = false;
+  bool strict_model_config_present = false;
   bool strict_model_config = false;
   bool strict_readiness = true;
   std::list<std::pair<int, uint64_t>> cuda_pools;
@@ -1445,9 +1447,11 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
         exit_on_error = ParseBoolOption(optarg);
         break;
       case OPTION_DISABLE_AUTO_COMPLETE_CONFIG:
-        strict_model_config = true;
+        disable_auto_complete_config = true;
         break;
       case OPTION_STRICT_MODEL_CONFIG:
+        std::cerr << "Warning: '--strict-model-config' has been deprecated! Please use '--disable-auto-complete-config' instead." << std::endl;
+        strict_model_config_present = true;
         strict_model_config = ParseBoolOption(optarg);
         break;
       case OPTION_STRICT_READINESS:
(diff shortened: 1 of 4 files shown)


src/main.cc
```diff
@@ -1445,9 +1447,13 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
         exit_on_error = ParseBoolOption(optarg);
         break;
       case OPTION_DISABLE_AUTO_COMPLETE_CONFIG:
-        strict_model_config = true;
+        disable_auto_complete_config = true;
         break;
       case OPTION_STRICT_MODEL_CONFIG:
+        std::cerr << "Warning: '--strict-model-config' has been deprecated! "
```
Tabrizian, line 1453: Use `LOG_MESSAGE(TRITONSERVER_LOG_WARN, "message")` instead of std::cerr.

src/main.cc
```diff
@@ -1781,6 +1787,18 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
     return false;
   }
 
+  // Check if there is a conflict between --disable-auto-complete-config
+  // and --strict-model-config
+  if (disable_auto_complete_config && strict_model_config_present) {
+    if (!strict_model_config) {
+      std::cerr
```
Tabrizian, line 1794: Same

src/main.cc
```diff
@@ -1781,6 +1787,18 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
...
+          << "Warning: Overriding deprecated '--strict-model-config' from "
+             "False to True in favor of '--disable-auto-complete-config'!"
+          << std::endl;
+    }
+    strict_model_config = true;
```
Tabrizian, line 1799: Do we need a warning for when the `strict_model_config=true` and `disable_auto_complete_config` is false?

Tabrizian (changes requested):

src/main.cc
nv-kmcgill53: In this case, the user gets the warning that they are using an deprecated CLI option from the switch statement. We should still allow them to use the `--strict-model-config` option to not break backward compatibility.

src/main.cc
nv-kmcgill53: LOG_MESSAGE hasn't been declared in this scope and I am not sure it should be since we allow logging to be turned on and off at compile time. I was copying this pattern from `OPTION_MODEL_CONTROL_MODE` or some of the other case statements.

src/main.cc
nv-kmcgill53: See other [comment](https://github.com/triton-inference-server/server/pull/4635/files#r920271597)

Tabrizian (approved):

tanmayv25 (approved):

PR #4640, opened 2022-07-14
Server changes for new logging extension

fpetrini15:


src/grpc_server.cc
```diff
@@ -1340,6 +1340,224 @@ CommonHandler::SetUpAllRequests()
...
+
+#ifdef TRITON_ENABLE_LOGGING
+    TRITONSERVER_Error* err = nullptr;
+    std::string log_file_path = LOG_OUT_FILE;
+    bool log_info_status = LOG_INFO_IS_ON;
+    bool log_warn_status = LOG_WARNING_IS_ON;
+    bool log_error_status = LOG_ERROR_IS_ON;
+    uint32_t verbose_level = LOG_VERBOSE_LEVEL;
+    triton::common::Logger::Format log_format_final = LOG_FORMAT;
+    std::string log_format_parse = LOG_FORMAT_STRING;
```
GuanLuo, line 1370: I don't think you need to declare these variables early, when you are preparing the response, you can just invoke the macro, i.e.: 
```
(*response->mutable_settings())["log_file"].set_string_param(LOG_OUT_FILE);
```

src/grpc_server.cc
```diff
@@ -1340,6 +1340,224 @@ CommonHandler::SetUpAllRequests()
...
+              inference::LogSettingsRequest_SettingValue::ParameterChoiceCase::
+                  kStringParam) {
+            err = TRITONSERVER_ErrorNew(
+                TRITONSERVER_ERROR_INVALID_ARG,
+                (std::string("expect string for '") + setting_name + "'")
+                    .c_str());
+            GOTO_IF_ERR(err, earlyexit);
+          } else {
+            // Set new settings in server then in core
+            log_file_path = it->second.string_param();
```
GuanLuo, line 1390: Change to `const std::string&` as you can declare the variable in place

src/grpc_server.cc
```diff
@@ -1340,6 +1340,224 @@ CommonHandler::SetUpAllRequests()
...
+            LOG_SET_OUT_FILE(log_file_path);
+            FAIL_IF_ERR(
+                TRITONSERVER_ServerOptionsSetLogOutFile(
+                    NULL, log_file_path.c_str()),
```
GuanLuo, line 1394: `nullptr`, should add comment here and below on passing `nullptr` is okay because we know the implementation detail that the update will be applied to global object so no pointer to object needs to be provided to this API call.

src/grpc_server.cc
```diff
@@ -1340,6 +1340,224 @@ CommonHandler::SetUpAllRequests()
...
+        "the server does not suppport dynamic logging");
+    GrpcStatusUtil::Create(status, err);
+    TRITONSERVER_ErrorDelete(err);
+#endif
+  };
+
+  new CommonCallData<
+      grpc::ServerAsyncResponseWriter<inference::LogSettingsResponse>,
+      inference::LogSettingsRequest, inference::LogSettingsResponse>(
+      "Dynamic_Logging", 0, OnRegisterLogging, OnExecuteLogging,
```
GuanLuo, line 1558: Drop `dynamic`

src/http_server.cc
```diff
@@ -1822,6 +1822,157 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
 #endif
 }
 
+void
+HTTPAPIServer::HandleLogging(evhtp_request_t* req)
```
GuanLuo, line 1826: Make similar changes as requested for GRPC

GuanLuo (changes requested):

qa/L0_logging/test.sh
```diff
@@ -0,0 +1,590 @@
...
+    if [ $? -ne 0 ]; then
+        RET=1
+    fi
+done
+
+set -e
+
+kill $SERVER_PID
+wait $SERVER_PID
+
```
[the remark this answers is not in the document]
fpetrini15: I have a check later that makes sure the file is populated with information, but I agree, it's not the strongest check. I think the cleanest way to approach this is to check in the python backend logging changes first, then modify this script based on a new python model that logs a variety of logs level. This way the script can check against their presence or lack thereof to ensure things are working correctly.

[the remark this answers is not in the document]
GuanLuo: I agree that with the Python backend change it would be easier to test if the backend logging is redirected correctly. Besides that, the endpoint logging / core logging redirection should be checked, and this can be done without waiting for the Python backend change. I think you will be able to generate all level of logging by running server in different setting.

qa/L0_logging/test.sh
```diff
@@ -0,0 +1,590 @@
...
+MODELSDIR=`pwd`/log_models
+
+SERVER=/opt/tritonserver/bin/tritonserver
+source ../common/util.sh
+
+rm -f *.log
+rm -fr $MODELSDIR && mkdir -p $MODELSDIR
+
+# set up simple and global_simple model using MODELBASE
```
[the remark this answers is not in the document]
fpetrini15: Removed the global model. Only using the simple model for testing.

qa/L0_logging/test.sh
```diff
@@ -0,0 +1,577 @@
...
+
+$SIMPLE_GRPC_CLIENT >> client_test_log_file.log 2>&1
+if [ $? -ne 0 ]; then
+    RET=1
+fi
+
+# Check redirection worked properly (server log has tolerance of 40 due to 
+# unavoidable onnx framework logging)
+verify_log_redirection "log_file.log" "inference_server_log_file.log" 50 40
+verify_log_redirection "other_log.log" "inference_server_log_file.log" 50 40
```
GuanLuo, line 224: Instead of counting lines, you can use `grep ^E[0-9][0-9][0-9][0-9].* file | wc -l` (from [here](https://github.com/triton-inference-server/server/blob/main/qa/L0_model_config/test.sh#L323)) to get number of lines reported for specific log level (error level in this example), which then you have finer control over what you are expecting.

docs/protocol/extension_logging.md
```diff
@@ -0,0 +1,199 @@
...
+setting value. Currently, the following log settings are defined:
+
+- "log_file" : a $string parameter defining the file where the log outputs will be saved. If not specified,
+log outputs will stream to the console.
+
+- "log_info" : a $boolean parameter that controls whether the Triton server logs INFO level messages. 
+
+- "log_warnings" : a $boolean parameter that controls whether the Triton server logs WARNING level messages. 
+
+- "log_errors" : a $boolean parameter that controls whether the Triton server logs ERROR level messages.
```
GuanLuo, line 78: Remove "s"?
```suggestion
- "log_warning" : a $boolean parameter that controls whether the Triton server logs WARNING level messages. 

- "log_error" : a $boolean parameter that controls whether the Triton server logs ERROR level messages. 
```

qa/L0_logging/test.sh
fpetrini15: I set the expected value to whatever current value it receives now, but in the future this may require maintenance.

src/grpc_server.cc
```diff
@@ -1340,6 +1340,226 @@ CommonHandler::SetUpAllRequests()
...
+                  kBoolParam) {
+            err = TRITONSERVER_ErrorNew(
+                TRITONSERVER_ERROR_INVALID_ARG,
+                (std::string("expect boolean for '") + setting_name + "'")
+                    .c_str());
+            GOTO_IF_ERR(err, earlyexit);
+          } else {
+            bool log_info_status = it->second.bool_param();
+            LOG_ENABLE_INFO(log_info_status);
+            FAIL_IF_ERR(
```
GuanLuo, line 1416: Don't `FAIL_IF_ERR`...

src/grpc_server.cc
```diff
@@ -1340,6 +1340,226 @@ CommonHandler::SetUpAllRequests()
...
+              std::string empty;
+              TRITONSERVER_ServerOptionsSetLogFile(nullptr, empty.c_str());
+              err = TRITONSERVER_ErrorNew(
+                  TRITONSERVER_ERROR_INVALID_ARG, ("Failed to open log file"));
+              GOTO_IF_ERR(err, earlyexit);
+            }
+            // Okay to pass nullptr because we know the update will be applied
+            // to the global object.
+            TRITONSERVER_ServerOptionsSetLogFile(
+                nullptr, log_file_path.c_str());
```
GuanLuo, line 1396: Check error status

src/grpc_server.cc
```diff
@@ -1340,6 +1340,226 @@ CommonHandler::SetUpAllRequests()
...
+                    .c_str());
+            GOTO_IF_ERR(err, earlyexit);
+          } else {
+            // Set new settings in server then in core
+            const std::string& log_file_path = it->second.string_param();
+            bool success = LOG_SET_OUT_FILE(log_file_path);
+            // On failure, close log file and revert to default "empty"
+            if (!success) {
+              std::string empty;
+              TRITONSERVER_ServerOptionsSetLogFile(nullptr, empty.c_str());
```
GuanLuo, line 1388: Don't change anything on error, reverting to previous state instead of back to default. i.e. if user have sucessfully set log file to "abc.log", later failure to set "def.log" shouldn't change the destination to standard error, otherwise more confusion is created.

src/grpc_server.cc
GuanLuo: Same for the other

GuanLuo (changes requested):

src/grpc_server.cc
fpetrini15: Kept duplicate check code because in the event that the server log settings fail and revert, it will still be aligned with the core log settings.

src/grpc_server.cc
fpetrini15: Previous state reversion now handled in logging.h, not here.

docs/protocol/extension_logging.md
```diff
@@ -0,0 +1,199 @@
...
+
+$log_setting = $string : $string | $boolean | $number
+```
+
+Each $log_setting JSON describes a “name”/”value” pair, where the “name” is
+the $string representation of the log setting and the “value” is a $string, $bool, or $number representation of the
+setting value. Currently, the following log settings are defined:
+
+- "log_file" : a $string parameter defining the file where the log outputs will be saved. If not specified,
+log outputs will stream to the console.
```
GuanLuo, line 72: "If not specified" -> "If empty string is specified". I think "not specified" just means that the user doesn't intend to change current setting.

GuanLuo (approved):

PR #4649, opened 2022-07-14
Fix L0_decoupled

Tabrizian:

PR #4650, opened 2022-07-14
Fix seqfault when calling TRITONSERVER_InferenceRequestId

krishung5:


src/grpc_server.cc
```diff
@@ -3681,9 +3681,11 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
     }
     // Get request ID for logging in case of error.
     const char* request_id = nullptr;
-    LOG_TRITONSERVER_ERROR(
-        TRITONSERVER_InferenceRequestId(irequest, &request_id),
-        "unable to retrieve request ID string");
+    if (irequest != nullptr) {
```
nv-kmcgill53, line 3684: Should this return if the `irequest` is null? The `request_id` becomes `"<id_unknown>"` but then we will attempt to use the null `irequest` on L3704  for TRITONSERVER_ServerInferAsync.

krishung5: I don't think TRITONSERVER_ServerInferAsync on L3704 will be triggered since the error is set on L3628 and we have this check on L3692

nv-kmcgill53 (approved):

nv-kmcgill53 (commented):
I think we should close this PR in favor of @Tabrizian's [PR](https://github.com/triton-inference-server/server/pull/4649)

krishung5:
Just saw that this has been fixed in Iman's PR. Closing this PR. Thanks Kyle.

PR #4654, opened 2022-07-15
Fix L0_model_config

nv-kmcgill53:
This PR fixes `strict_model_config` not being set to `true` when the `--disable-auto-complete-config` flag is used and the `--strict-model-config` flag is not present.


Tabrizian (approved):
We need to cherry-pick this change too, right?

PR #4656, opened 2022-07-15
fixed strict_model_config and disable-auto-complete-config

mc-nv:


nv-kmcgill53 (approved):

src/main.cc
```diff
@@ -1781,6 +1787,18 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
     return false;
   }
 
+  // Check if there is a conflict between --disable-auto-complete-config
+  // and --strict-model-config
+  if (disable_auto_complete_config) {
+    if (strict_model_config_present && !strict_model_config) {
+      std::cerr
+          << "Warning: Overriding deprecated '--strict-model-config' from "
+             "False to True in favor of '--disable-auto-complete-config'!"
```
dyastremsky, line 1796: Approving because I'm not sure whether these are from other fixes/PRs, but I think verbiage is a tad confusing here. It could be clearer if it was something along the lines of:
"Warning: Overriding deprecated '--strict-model-config'=True in favor of '--disable-auto-complete-config'."

nv-kmcgill53: These are two PRs merged for this cherry pick. I can fix this going forward.

dyastremsky: Got it, thanks Kyle!

Issue #4674, opened 2022-07-20
perf_analyzer


Vozf:
**Description**
While using perf_analyzer with json file with string\bytes input of the pipeline I encountered this error.

```
Thread [0] had error: in ensemble '', Failed to process the request(s) for model instance 'preprocess_0', message: error: unpack_from requires a buffer of at least 3774863619 bytes for unpacking 3774863615 bytes at offset 4 (actual buffer size is 1507101)

At:
  /opt/tritonserver/backends/python/triton_python_backend_utils.py(116): deserialize_bytes_tensor
```
I generated json as described here https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md
I used b64 encoding as was described for my bytes data.
Although it's not specified that in the beginning of b64 string there should be a 4 bytes with the lenght of data.
First I generated b64 with this code
```
{"b64": base64.b64encode((path.read_bytes())).decode()}
```
But you need to prepend the string size. This got me working
```
def get_binary_image_with_prepended_size(filename):
    img_size = os.path.getsize(filename)
    bts = bytearray(Path(filename).read_bytes())
    binary_file = img_size.to_bytes(4, byteorder='little') + bts

    return binary_file


bts = get_binary_image_with_prepended_size(path)
return {"b64": base64.b64encode(bts).decode()}

```
I believe https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md should mention that you need to prepend the data length in base64 string

Tabrizian:
Thank you for your documentation request. I have created a PR to improve the documentation.

Vozf:
Looks good

PR #4676, opened 2022-07-20
Add a note about string data type when using base64

Tabrizian:
Closes https://github.com/triton-inference-server/server/issues/4674

### DIFF
--- a/docs/perf_analyzer.md
+++ b/docs/perf_analyzer.md
@@ -402,6 +402,10 @@ The following is the example to provide contents as base64 string with explicit 
 }
 ```
 
+Note that for STRING type an element is represented by a 4-byte unsigned integer giving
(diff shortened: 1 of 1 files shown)


docs/perf_analyzer.md
```diff
@@ -402,6 +402,10 @@ The following is the example to provide contents as base64 string with explicit
 }
 ```
 
+Note that for STRING type an element is represented by a 4-byte unsigned integer giving
```
matthewkotila, line 405: Should an example be given for STRING type? Or is there already an example? Is there a difference between a base64 input and a string input?

Tabrizian: The example is provided right above this line. Base64 is an encoding of a bytes array.

matthewkotila (approved):

PR #4684, opened 2022-07-22
Add Tracing to Perf Analyzer GRPC/HTTP Client

dyastremsky:
Related Client PR: https://github.com/triton-inference-server/client/pull/136

### DIFF
--- a/qa/L0_perf_analyzer/test.sh
+++ b/qa/L0_perf_analyzer/test.sh
@@ -642,7 +642,7 @@ for PROTOCOL in grpc http; do
     $PERF_ANALYZER -v -i $PROTOCOL -m simple_savedmodel_sequence_object -p 2000 -t5 --sync --trace-file $TRACE_FILE \
     --trace-level TIMESTAMPS --trace-rate 1000 --trace-count 100 --log-frequency 10 \
     --input-data=$SEQ_JSONDATAFILE >$CLIENT_LOG 2>&1
-    if [ $? -eq 0 ]; then
+    if [ $? -ne 0 ]; then
         cat $CLIENT_LOG
         echo -e "\n***\n*** Test Failed\n***"
         RET=1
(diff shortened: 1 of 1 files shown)


qa/L0_perf_analyzer/test.sh
```diff
@@ -635,6 +635,27 @@ for PROTOCOL in grpc http; do
         set -e
     done
 
+    # Testing that trace logging works
+    set +e
+    TRACE_FILE="trace.json"
+    rm ${TRACE_FILE}*
+    $PERF_ANALYZER -v -i $PROTOCOL -m simple_savedmodel_sequence_object -p 2000 -t5 --sync --trace-file $TRACE_FILE \
+    --trace-level TIMESTAMPS --trace-rate 1000 --trace-count 100 --log-frequency 10 \
```
nv-kmcgill53, line 643: Should we also test `--trace-level TENSORS `? Or do we expect the same output regardless of the two options?

nv-kmcgill53 (changes requested):

qa/L0_perf_analyzer/test.sh
dyastremsky: We could. [L0_cmdline_trace](https://github.com/triton-inference-server/server/blob/main/qa/L0_cmdline_trace/test.sh) already tests that the HTTP/GRPC clients trace correctly across different permutations, so I was trying to avoid repeating those tests. My plan with testing was to ensure tracing gets enabled correctly when the trace command line options are passed into perf_analyzer (which feeds them into the HTTP/GRPC clients), but I can add additional tests if you think they'd be useful.

nv-kmcgill53: Looks like the test coverage is good. Thanks for pointing me to `L0_cmdline_trace`.

nv-kmcgill53 (approved):

PR #4693, opened 2022-07-25
Increase SERVER_TIMEOUT for L0_infer_valgrind

krishung5:
Related PRs:
common: https://github.com/triton-inference-server/common/pull/67
backend: https://github.com/triton-inference-server/backend/pull/67
tensorrt_backend: https://github.com/triton-inference-server/tensorrt_backend/pull/44

### DIFF
--- a/qa/L0_infer/infer_test.py
+++ b/qa/L0_infer/infer_test.py
@@ -64,6 +64,7 @@ class InferTest(tu.TestResultCollector):
                     output0_raw,
                     output1_raw,
                     swap,
+                    # 60 sec is the default value
                     network_timeout=60.0):
 
         def _infer_exact_helper(tester,
--- a/qa/L0_infer/infer_test.py
+++ b/qa/L0_infer/infer_test.py
@@ -41,6 +41,7 @@ TEST_SYSTEM_SHARED_MEMORY = bool(
 TEST_CUDA_SHARED_MEMORY = bool(int(os.environ.get('TEST_CUDA_SHARED_MEMORY',
                                                   0)))
 CPU_ONLY = (os.environ.get('TRITON_SERVER_CPU_ONLY') is not None)
+TEST_VALGRIND = bool(int(os.environ.get('TEST_VALGRIND', 0)))
 
 USE_GRPC = (os.environ.get('USE_GRPC', 1) != "0")
 USE_HTTP = (os.environ.get('USE_HTTP', 1) != "0")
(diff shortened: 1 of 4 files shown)


tanmayv25 (dismissed):

GuanLuo:
It shouldn't take this long (2 days) to load models, right? When did the test start timing out?

krishung5:
@GuanLuo I don't expect the test to takes 2 days long either. I set it to 2 days just in case it runs over the TIMEOUT and we need to restart it all over again. In the last run, the test got killed because there is a 8h timeout set in our CI pipeline configuration.

krishung5:
Removed onnx and python backends for valgrind tests. Added a possible memory leak introduced from [OpenVINO side](https://github.com/openvinotoolkit/openvino/issues/6593). Made some changes in [OpenVINO backend](https://github.com/triton-inference-server/openvino_backend/pull/40).

krishung5:
Rebased

qa/L0_infer/test.sh
```diff
@@ -63,6 +63,9 @@ if [ "$TEST_VALGRIND" -eq 1 ]; then
     LEAKCHECK_ARGS_BASE="--leak-check=full --show-leak-kinds=definite --max-threads=3000 --num-callers=20"
     SERVER_TIMEOUT=4000
     rm -f $LEAKCHECK_LOG_BASE*
+    # Remove onnx and python backends for now as the server hangs up when
```
tanmayv25, line 66: Did you create follow-up tickets to investigate this? If yes, then can you add the DLIS-<IDs> here?

krishung5: Added ticket number.

krishung5:
Although L0_infer_valgrind passes locally, it seems like running `test_class_bbb` test case with TF models on CI would fail due to timeout issue. Increased network_timeout for this case.

krishung5:
L0_infer_valgrind passes on CI: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/42488617

qa/L0_infer/infer_test.py
```diff
@@ -56,8 +57,14 @@
...
-    def _full_exact(self, input_dtype, output0_dtype, output1_dtype,
-                    output0_raw, output1_raw, swap):
+    def _full_exact(self,
+                    input_dtype,
+                    output0_dtype,
+                    output1_dtype,
+                    output0_raw,
+                    output1_raw,
+                    swap,
+                    network_timeout=60.0):
```
tanmayv25, line 67: Why are we adding the network timeout for this test?

krishung5: We only increase `network_timeout `for TensorFlow models for valgrind test. The default value of  `network_timeout ` is 60.0. (https://github.com/triton-inference-server/client/blob/main/src/python/library/tritonclient/http/__init__.py#L155)

tanmayv25 (approved):

PR #4704, opened 2022-07-27
Add test coverage to check if output tensor is on GPU

krishung5:
python_backend: https://github.com/triton-inference-server/python_backend/pull/179

### DIFF
--- a/qa/L0_backend_python/bls/test.sh
+++ b/qa/L0_backend_python/bls/test.sh
@@ -79,7 +79,9 @@ mkdir -p models/dlpack_identity/1/
 cp ../../python_models/dlpack_identity/model.py models/dlpack_identity/1/
 cp ../../python_models/dlpack_identity/config.pbtxt models/dlpack_identity
 
-cp -r ${DATADIR}/qa_sequence_implicit_model_repository/onnx_nobatch_sequence_int32/ ./models
+cp -r ${DATADIR}/qa_sequence_implicit_model_repository/onnx_nobatch_sequence_int32/ ./models && \
+    (cd models/onnx_nobatch_sequence_int32 && \
+        sed -i "s/kind: KIND_GPU/kind: KIND_CPU/" config.pbtxt)
 
 run_server
 if [ "$SERVER_PID" == "0" ]; then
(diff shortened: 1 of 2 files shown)


qa/L0_backend_python/bls/test.sh
```diff
@@ -79,7 +79,9 @@ mkdir -p models/dlpack_identity/1/
 cp ../../python_models/dlpack_identity/model.py models/dlpack_identity/1/
 cp ../../python_models/dlpack_identity/config.pbtxt models/dlpack_identity
 
-cp -r ${DATADIR}/qa_sequence_implicit_model_repository/onnx_nobatch_sequence_int32/ ./models
+cp -r ${DATADIR}/qa_sequence_implicit_model_repository/onnx_nobatch_sequence_int32/ ./models && \
+    (cd models/onnx_nobatch_sequence_int32 && \
+        sed -i "s/kind: KIND_GPU/kind: KIND_CPU/" config.pbtxt)
```
Tabrizian, line 84: I was wondering why is this change required? Also, can you add a test where you send an inference to an onnx model on GPU and check the output tensor provided by BLS to make sure it is on GPU?

krishung5: > I was wondering why is this change required?

I thought we cannot use numpy for GPU tensors as it shows [here](https://github.com/triton-inference-server/python_backend/blob/2b98842559055b4d2837dd972e2392683c846079/src/pb_tensor.cc#L405), that's why I changed it to KIND_CPU so that we can use `as_numpy` in the [unittest](https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/python_unittest.py#L48) script.

> Also, can you add a test where you send an inference to an onnx model on GPU and check the output tensor provided by BLS to make sure it is on GPU?

Thanks for the heads up, I will add a test for checking if the output tensor is on GPU.

krishung5: Update: Just found out that we can use dlpack() to replace numpy. I will modify the test.

krishung5: Reverted the changes on onnx model. Added checks to see if output tensor is on GPU when sending inference to onnx model.

Tabrizian (approved):

PR #4706, opened 2022-07-28
Add end-to-end config tuning doc

rmccorm4:
Context is to give users a clear path from taking their trained model to deploying it optimally/at-scale with Triton.

### DIFF
--- a/docs/performance_tuning.md
+++ b/docs/performance_tuning.md
@@ -179,11 +179,11 @@ To check if the model loaded successfully, we expect to see our model in a `READ
 ...
 I0802 18:11:47.100537 135 model_repository_manager.cc:1345] successfully loaded 'densenet_onnx' version 1
 ...
-+---------+---------+--------+
-| Model   | Version | Status |
-+---------+---------+--------+
++---------------+---------+--------+
+| Model         | Version | Status |
++---------------+---------+--------+
 | densenet_onnx | 1       | READY  |
-+---------+---------+--------+
++---------------+---------+--------+
 ...
 ```
 
--- a/docs/performance_tuning.md
+++ b/docs/performance_tuning.md
@@ -53,11 +53,8 @@ For those who wish to jump right in, skip to the [end-to-end example](#end-to-en
     Here is a simplified output for demonstration purposes:
 
     ```
-    # NOTE: "my_model" represents the model currently being served by 
-    #       Triton in the background from the previous step.
-
+    # NOTE: "my_model" represents a model currently being served by Triton
     $ perf_analyzer -m my_model
-
     ...
 
     Inferences/Second vs. Client Average Batch Latency
@@ -65,7 +62,7 @@ For those who wish to jump right in, skip to the [end-to-end example](#end-to-en
     ```
 
     - This gives us a sanity test that we are able to successfully form input requests and receive output responses to communicate with the model backend via Triton APIs. 
-    If Perf Analyzer fails to send requests and it is unclear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects. 
+    - If Perf Analyzer fails to send requests and it is unclear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects. 
     If the config is correct, check that the model runs successfully using its original framework directly. 
     If you don't have your own script or tool to do so, [Polygraphy](https://github.com/NVIDIA/TensorRT/tree/main/tools/Polygraphy) is a useful tool to run sample inferences on your model via various frameworks. 
     Currently, Polygraphy supports ONNXRuntime, TensorRT, and TensorFlow 1.x.
@@ -76,14 +73,16 @@ For those who wish to jump right in, skip to the [end-to-end example](#end-to-en
 
 3. How can I improve my model performance?
     - To further understand the best model configuration you can provide to Triton for your use case, Triton's [Model Analyzer](https://github.com/triton-inference-server/model_analyzer) tool can help. 
+    Model Analyzer can automatically or [manually](https://github.com/triton-inference-server/model_analyzer/blob/main/docs/config_search.md) search through config combinations to find the optimal triton configuration to meet your constraints.
     After running Model Analyzer to find the optimal configurations for your model/use case, you can transfer the generated config files to your [Model Repository](https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md). 
     Model Analyzer provides a [Quickstart](https://github.com/triton-inference-server/model_analyzer/blob/main/docs/quick_start.md) guide with some examples to walk through.
     - Upon serving the model with the newly optimized configuration file found by Model Analyzer and running Perf Analyzer again, you should expect to find better performance numbers in most cases compared to a default config.
-    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer since they don't apply to all models. 
+    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer's automatic search since they don't apply to all models.
     For instance, [backends](https://github.com/triton-inference-server/backend) can expose backend-specific configuration options that can be tuned as well. 
     The [ONNXRuntime Backend](https://github.com/triton-inference-server/onnxruntime_backend), for example, has several [parameters](https://github.com/triton-inference-server/onnxruntime_backend#model-config-options) that affect the level of parallelization when executing inference on a model. 
     These backend-specific options may be worth investigating if the defaults are not providing sufficient performance. 
-    - For manual tuning of your model configuration, there is further information in our [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md) docs.
+    To tune custom sets of parameters, Model Analyzer supports [Manual Configuration Search](https://github.com/triton-inference-server/model_analyzer/blob/main/docs/config_search.md).
+    - To learn more about further optimizations for your model configuration, see the [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md) docs.
 
 ### Other Areas of Interest
 
--- a/docs/walkthrough.md
+++ b/docs/walkthrough.md
@@ -28,11 +28,11 @@
 
 # From Model to Deployment
 
-Given a trained model, how do I take it and deploy it at-scale with an
+Given a trained model, how do I deploy it at-scale with an
 optimal configuration using Triton Inference Server? This document is
 here to help answer that.
 
-For those who like a high level overview, below is the common flow for most use cases. For those who wish to jump in, skip to an [end-to-end example](#end-to-end-example).
+For those who like a high level overview, below is the common flow for most use cases. For those who wish to jump right in, skip to the [end-to-end example](#end-to-end-example).
 
 1. Is my model compatible with Triton?
     - If your model falls under one of Triton's [supported backends](https://github.com/triton-inference-server/backend), then we can simply try to deploy the model as described in the [Quickstart](https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md) guide. For the ONNXRuntime, TensorFlow SavedModel, and TensorRT backends, the model configuration can be completely inferred from the model using Triton's [AutoComplete](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#auto-generated-model-configuration) feature, so no `config.pbtxt` file is required. Additionally, by enabling verbose logging via `--log-verbose=1` - you can see the complete config that Triton sees internally in the server log output. For other backends, refer to the [Minimal Model Configuration](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#minimal-model-configuration) required to get started.
@@ -55,18 +55,18 @@ For those who like a high level overview, below is the common flow for most use 
     ```
 
     - This gives us a sanity test that we are able to successfully form input requests and receive output responses to communicate with the model backend via Triton APIs. If for some reason Perf Analyzer fails to send requests and it is not clear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects, and that you can perform inference directly using your model's original framework.
-    - The defintion of "performing well" is subject to change for each use case. There are many variables that can be tweaked just within your Triton configuration (`config.pbtxt`) to obtain different results.
+    - The defintion of "performing well" is subject to change for each use case. Some common metrics are throughput, latency, and GPU utilization. There are many variables that can be tweaked just within your Triton configuration (`config.pbtxt`) to obtain different results.
     - As your model, config, or use case evolves, [Perf Analyzer](https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md) is a great tool to quickly verify model functionality and performance.
 
 3. How can I improve my model performance?
     - To further understand the best model configuration you can provide to Triton for your use case, Triton's [Model Analyzer](https://github.com/triton-inference-server/model_analyzer) tool can help. After running Model Analyzer to find the optimal configurations for your model/use case, you can transfer the generated config files to your [Model Repository](https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md). Model Analyzer provides a [Quickstart](https://github.com/triton-inference-server/model_analyzer/blob/main/docs/quick_start.md) guide with some examples to walk through.
     - Upon serving the model with the newly optimized configuration file found by Model Analyzer and running Perf Analyzer again, you should expect to find better performance numbers in many cases compared to a default config.
-    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer since they don't apply to all models. For instance, [backends](https://github.com/triton-inference-server/backend) can expose backend-specific configuration options that can be tuned as well. The [ONNXRuntime Backend](https://github.com/triton-inference-server/onnxruntime_backend) for example has several [parameters](https://github.com/triton-inference-server/onnxruntime_backend#model-config-options) that affect the level of parallelization used when executing inference on a model and may be worth investigating if the defaults are not providing sufficient performance. 
+    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer since they don't apply to all models. For instance, [backends](https://github.com/triton-inference-server/backend) can expose backend-specific configuration options that can be tuned as well. The [ONNXRuntime Backend](https://github.com/triton-inference-server/onnxruntime_backend), for example, has several [parameters](https://github.com/triton-inference-server/onnxruntime_backend#model-config-options) that affect the level of parallelization when executing inference on a model. These backend-specific options may be worth investigating if the defaults are not providing sufficient performance. 
     - For manual tuning of your model configuration, there is further information in our [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md) docs.
 
 ## Other Areas of Interest
 
-1. My model performs slowly when it is first loaded by Triton (cold-start penalty)
+1. My model performs slowly when it is first loaded by Triton (cold-start penalty), what do I do?
     - Triton exposes the ability to run [ModelWarmup](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#model-warmup) requests when first loading the model to ensure that the model is sufficiently initialized before being marked "READY" for inference.
 
 2. Why doesn't my model perform significantly faster on GPU?
@@ -76,6 +76,8 @@ For those who like a high level overview, below is the common flow for most use 
     - If none of the above can help get sufficient GPU-accelerated performance for your model, the model may simply be better designed for CPU execution and the [OpenVINO Backend](https://github.com/triton-inference-server/openvino_backend) may help further optimize your CPU execution.
 
 ## End-to-end Example
+
+> NOTE: If you have never worked with Triton before, you may be interested in first checking out the [Quickstart](https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md) example. Some basic understanding of Triton may be useful for the following section, but this example is meant to be straightforward enough without prior experience.
 
 Let's take an ONNX model as our example since ONNX is designed to be a format that can be [easily exported](https://github.com/onnx/tutorials#converting-to-onnx-format) from most other frameworks.
 
@@ -127,6 +129,9 @@ docker run -d --gpus=all --network=host -v $PWD:/mnt --name triton-sdk nvcr.io/n
 
 
 4. Serve the model with Triton
+
+To serve our model, we will use the `server` container that we already started which comes pre-installed with a `tritonserver` binary.
+
 ```bash
 # Enter server container interactively
 docker exec -ti triton-server bash
@@ -139,9 +144,11 @@ tritonserver --model-repository=/mnt/models
 
 5. Verify the model can run inference
 
-In a separate shell, we use Perf Analyzer to sanity check that we can run inference on our model and get a baseline for the kind of performance we expect from this model.
+To verify our model can perform inference, we will use the `triton-sdk` (client) container that we already started which comes with `perf_analyzer`  pre-installed.
 
-In the example below, Perf Analyzer is sending requests to models served on the same machine (`localhost` from the server container via `--network=host`). However, you may also test models being served remotely by setting the `-u` flag, such as `perf_analyzer -m alexnet -u <IP>:<PORT>`.
+In a separate shell, we use Perf Analyzer to sanity check that we can run inference and get a baseline for the kind of performance we expect from this model.
+
+In the example below, Perf Analyzer is sending requests to models served on the same machine (`localhost` from the server container via `--network=host`). However, you may also test models being served remotely at some `<IP>:<PORT>`  by setting the `-u` flag, such as `perf_analyzer -m alexnet -u 127.0.0.1:8000`.
 
 ```bash
 # Enter SDK container interactively
@@ -161,6 +168,12 @@ Concurrency: 4, throughput: 965.21 infer/sec, latency 4142 usec
 ```
 
 6. Run Model Analyzer to find the best configurations for our model
+
+While Model Analyzer comes pre-installed in the SDK container and supports various modes of connecting to a Triton server,
+for simplicity we will use install Model Analyzer in our `server` container to use the `--triton-launch-mode=local` flag.
+To learn more about other methods of connecting Model Analyzer to a running Triton Server, see the `--triton-launch-mode`
+Model Analyzer flag.
+
 ```bash
 # Enter server container interactively
 docker exec -ti triton-server bash
@@ -199,7 +212,7 @@ Example Model Analyzer output summary:
 | alexnet_config_2 | 0 | Enabled | 3/GPU | 44.403 | 1563.16 | 5117 | 55.1 |
 | alexnet_config_default | 0 | Disabled | 1/GPU | 8.274 | 1001.77 | 3786 | 24.6 |
 
-In the table above, we see that setting our GPU [Instance Count](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#instance-groups) to 5 allows us to achieve the highest throughput and lowest latency for this example. Note that our example "alexnet" model has a fixed batch-size that is explicitly specified in the first dimension of the Input/Output `dims`, therefore the `max_batch_size` is set to 0 as described [here](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#maximum-batch-size). For models that support dynamic batch size, Model Analyzer would also tune the `max_batch_size` parameter.
+In the table above, we see that setting our GPU [Instance Count](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#instance-groups) to 5 allows us to achieve the highest throughput and lowest latency for this example. Note that our this "alexnet" model has a fixed batch-size that is explicitly specified in the first dimension of the Input/Output `dims`, therefore the `max_batch_size` parameter is set to 0 as described [here](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#maximum-batch-size). For models that support dynamic batch size, Model Analyzer would also tune the `max_batch_size` parameter.
 
 7. Extract optimal config from Model Analyzer results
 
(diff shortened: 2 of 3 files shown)


rmccorm4:
First draft, expecting a lot of feedback.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+-->
+
+# From Model to Deployment
+
+Given a trained model, how do I take it and deploy it at-scale with an
+optimal configuration using Triton Inference Server? This document is
+here to help answer that.
+
+There are a few steps in this process, and there is no single answer for
+everything, but here is the common flow:
```
nv-kmcgill53, line 36: I would add a link to the end-to-end example with something like:

> For those who like a high level overview, below is the common flow for most use cases. For those who wish to jump in, skip to an [end-to-end example](End-to-end Example).

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+    - For manual tuning of your model configuration, there is further information in our [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md) docs.
+
+## Other Areas of Interest
+
+1. My model performs slowly when it is first loaded by Triton (cold-start penalty)
+    - Triton exposes the ability to run [ModelWarmup](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#model-warmup) requests when first loading the model to ensure that the model is sufficiently initialized before being marked "READY" for inference.
+
+2. Why doesn't my model perform significantly faster on GPU?
+    - Most official backends supported by Triton are optimized for GPU inference and should perform well on GPU out of the box.
+    - However, Triton expose options to further optimize your model for GPU execution for you. Triton's [Framework Specific Optimizations](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md#framework-specific-optimization) goes into further detail on this topic.
```
nv-kmcgill53, line 73: >expose

Change to "exposes".

>  However, Triton expose options to further optimize your model for GPU execution for you.

Even with the above change, this sentence reads a bit awkward. Perhaps, "Triton exposes options for you to optimize your model further on the GPU." Drop the "However," bullet points don't need adverbs as you are declaring a list rather than splitting paragraphs into points.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+    - Lastly, complete conversion of your model to a backend fully optimized for GPU inference such as [TensorRT](https://developer.nvidia.com/tensorrt) may provide even better results. You may find more Triton-specific details about TensorRT in the [TensorRT Backend](https://github.com/triton-inference-server/tensorrt_backend).
```
nv-kmcgill53, line 74: Remove "Lastly," as noted in the above comment.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+  --model-repository=/mnt/models \
+  --profile-models=alexnet \
+  --triton-launch-mode=local \
+  --output-model-repository-path=results
+
+# Summarize the profiling results
+model-analyzer analyze --analysis-models=alexnet
+```
+
+Example Model Analyzer output summary:
```
nv-kmcgill53, line 181: I think just a bit more discussion about the chart you are displaying to the user. Sure we see that we are choosing the best throughput, but why not choose the default config when it has the lowest latency? Motivate the user as to why they should read the whole chart and not just the throughput.

I like that you chose the alexnet model to get some good numbers here.

nv-kmcgill53 (changes requested):
On the whole I think you did a great job. I think the walkthrough, especially the end-to-end example will help new users make sense of the triton ecosystem.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+    - If your model does not come directly from a supported backend, but is executable through Python; you may be interested in writing a simple python script for running your model through the [Python Backend](https://github.com/triton-inference-server/python_backend).
+    - Otherwise, you can create a [Custom Backend](https://github.com/triton-inference-server/backend/blob/main/examples/README.md) for executing your model.
+
+2. Does my model perform well?
```
matthewkotila, line 43: What value/actions does this section provide? It seems like none, since the next step (Model Analyzer) actually searches for the optimal performance/config.

Same comment for [here](https://github.com/triton-inference-server/server/pull/4706/files#diff-2baaf382502aa8c67478520691f377834f23808a4e116e0b4ee824413e639383R137).

rmccorm4: Hm, this section is a bit bare, thanks for pointing that out. This PA snippet in particular is to give the reader a visual idea of what PA is immediately without having to click into the other docs/links wondering "what is Perf Analyzer?"

Also, I want to tweak this to more clear that it's a sanity test that you can successfully run inference on your served model (if not, take a step back here and check your config, etc.) as model analyzer will fail

As for the latter section using PA, I think the baseline perf number from a quick PA run help to make Model Analyzer shine a little bit, showing that we can find a config for you that boosts your baseline performance significantly. This is also a gentle guide to show users how to check any arbitrary config quickly - maybe they want to add some `optimization` or `parameters` to the `config.pbtxt` and Model Analyzer can't search/optimize and they want a rough idea of the impact.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+    - Assuming you were able to deploy your model on Triton, the next step is to benchmark the performance of your model. Triton's [Perf Analyzer](https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md) tool specifically fits this purpose. Here is a simplified output for demonstration purposes:
```
dyastremsky, line 44: Nit: Replace deploy with load? If they were able to deploy it, they wouldn't need this guide.

dyastremsky: It might also be worthwhile to state here or above that they can follow the Quickstart guide for starting the Triton server and client container (if you don't want to repeat the commands here). Maybe a description of each container.* For the next step, they should do it from the client container.

*Example: The server container runs the server, with their models loaded. The client container allows them to use our clients or their own to run inference on those models. Perf_analyzer is pre-installed in that container.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+
+    $ perf_analyzer -m my_model
+
+    ...
+
+    Inferences/Second vs. Client Average Batch Latency
+    Concurrency: 1, throughput: 482.8 infer/sec, latency 12613 usec
+    ```
+
+    - The defintion of "performing well" is subject to change for each use case. There are many variables that can be tweaked just within your Triton configuration (`config.pbtxt`) to obtain different results.
```
dyastremsky, line 58: Provide an example? Two common use cases is to get their throughput (number of inference requests completed per second) to a certain target or reduce their latency.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+    - Upon serving the model with the newly optimized configuration file found by Model Analyzer and running Perf Analyzer again, you should expect to find better performance numbers in many cases compared to a default config.
+    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer since they don't apply to all models. For instance, [backends](https://github.com/triton-inference-server/backend) can expose backend-specific configuration options that can be tuned as well. The [ONNXRuntime Backend](https://github.com/triton-inference-server/onnxruntime_backend) for example has several [parameters](https://github.com/triton-inference-server/onnxruntime_backend#model-config-options) that affect the level of parallelization used when executing inference on a model and may be worth investigating if the defaults are not providing sufficient performance.
```
dyastremsky, line 63: Nit: "...Backend, for example, has several..."

Nit: Sentence is a bit long for readability. "On a model. They may be worth investigating..." could be an easy split.

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer since they don't apply to all models. For instance, [backends](https://github.com/triton-inference-server/backend) can expose backend-specific configuration options that can be tuned as well. The [ONNXRuntime Backend](https://github.com/triton-inference-server/onnxruntime_backend) for example has several [parameters](https://github.com/triton-inference-server/onnxruntime_backend#model-config-options) that affect the level of parallelization used when executing inference on a model and may be worth investigating if the defaults are not providing sufficient performance. 
+    - For manual tuning of your model configuration, there is further information in our [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md) docs.
+
+## Other Areas of Interest
+
+1. My model performs slowly when it is first loaded by Triton (cold-start penalty)
```
dyastremsky, line 68: Nit: for consistency, this should ideally be a question. Even if not, you'd want to end with punctuation (period, question mark, etc.).

docs/walkthrough.md
dyastremsky: I actually see that you really clearly, comprehensively run through the end-to-end steps at the bottom of this document, so you can probably ignore the container descriptions unless you want to add them there (I think it could still be useful, especially so users understand what the sdk container is vs. the server one).

docs/walkthrough.md
```diff
@@ -0,0 +1,204 @@
...
+
+2. Why doesn't my model perform significantly faster on GPU?
+    - Most official backends supported by Triton are optimized for GPU inference and should perform well on GPU out of the box.
+    - However, Triton expose options to further optimize your model for GPU execution for you. Triton's [Framework Specific Optimizations](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md#framework-specific-optimization) goes into further detail on this topic.
+    - Lastly, complete conversion of your model to a backend fully optimized for GPU inference such as [TensorRT](https://developer.nvidia.com/tensorrt) may provide even better results. You may find more Triton-specific details about TensorRT in the [TensorRT Backend](https://github.com/triton-inference-server/tensorrt_backend).
+    - If none of the above can help get sufficient GPU-accelerated performance for your model, the model may simply be better designed for CPU execution and the [OpenVINO Backend](https://github.com/triton-inference-server/openvino_backend) may help further optimize your CPU execution.
+
+## End-to-end Example
```
dyastremsky, line 77: You did a really solid job here. I think this will be quite powerful for users.

It might be worthwhile to outline in the start of the doc that you'll cover Triton at a high-level first, then provide a step-by-step guide for deploying their model at the end. That'll get their interest and avoid any frustration from trying to deploy from the top section.

docs/walkthrough.md
matthewkotila: Thanks for clarifying.

I agree that using PA as a sanity test is a value-add--this should be added to the text. I also agree that mentioning PA here to help market PA is a value-add. I also agree that showing baseline numbers with PA could help illustrate the improvements from MA (although MA does show performance numbers from a "default" run, too), which is a value-add--this could also be added to the text.

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+For those who like a high level overview, below is the common flow for most use cases. For those who wish to jump right in, skip to the [end-to-end example](#end-to-end-example).
+
+1. Is my model compatible with Triton?
+    - If your model falls under one of Triton's [supported backends](https://github.com/triton-inference-server/backend), then we can simply try to deploy the model as described in the [Quickstart](https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md) guide. For the ONNXRuntime, TensorFlow SavedModel, and TensorRT backends, the model configuration can be completely inferred from the model using Triton's [AutoComplete](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#auto-generated-model-configuration) feature, so no `config.pbtxt` file is required. Additionally, by enabling verbose logging via `--log-verbose=1` - you can see the complete config that Triton sees internally in the server log output. For other backends, refer to the [Minimal Model Configuration](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#minimal-model-configuration) required to get started.
```
dyastremsky, line 38: Nit: the dash before "you can see the complete config" should be replaced with a comma (better flow, grammatically correct).

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+    - If your model does not come directly from a supported backend, but is executable through Python; you may be interested in writing a simple python script for running your model through the [Python Backend](https://github.com/triton-inference-server/python_backend).
```
dyastremsky, line 39: Nit: Semicolon here is similar to above comment (breaks the flow, grammatically incorrect). Semicolons break apart sentences that can stand on their own, but this is one longer sentence. Lots of ways you can do this. Simple one is replacing with a comma. You could also break this into two shorter sentences, e.g. "Your model may not run on a supported backend but still be executable through Python. In that case, you could write a simple Python script for running your model through the Python backend."

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+    $ perf_analyzer -m my_model
+
+    ...
+
+    Inferences/Second vs. Client Average Batch Latency
+    Concurrency: 1, throughput: 482.8 infer/sec, latency 12613 usec
+    ```
+
+    - This gives us a sanity test that we are able to successfully form input requests and receive output responses to communicate with the model backend via Triton APIs. If for some reason Perf Analyzer fails to send requests and it is not clear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects, and that you can perform inference directly using your model's original framework.
+    - The defintion of "performing well" is subject to change for each use case. Some common metrics are throughput, latency, and GPU utilization. There are many variables that can be tweaked just within your Triton configuration (`config.pbtxt`) to obtain different results.
```
dyastremsky, line 58: Nit: definition

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+
+    $ perf_analyzer -m my_model
+
+    ...
+
+    Inferences/Second vs. Client Average Batch Latency
+    Concurrency: 1, throughput: 482.8 infer/sec, latency 12613 usec
+    ```
+
+    - This gives us a sanity test that we are able to successfully form input requests and receive output responses to communicate with the model backend via Triton APIs. If for some reason Perf Analyzer fails to send requests and it is not clear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects, and that you can perform inference directly using your model's original framework.
```
dyastremsky, line 57: Nit: Sentence runs on a bit. Maybe break into a couple of sentences like "If Perf Analyzer fails to send requests and it is unclear from the error how to proceed... expects. Also, check that your model runs successfully on its original framework."

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+for simplicity we will use install Model Analyzer in our `server` container to use the `--triton-launch-mode=local` flag.
+To learn more about other methods of connecting Model Analyzer to a running Triton Server, see the `--triton-launch-mode`
+Model Analyzer flag.
+
+```bash
+# Enter server container interactively
+docker exec -ti triton-server bash
+
+# Stop existing tritonserver process if still running
+# because model-analyzer will start its own server
```
dyastremsky, line 182: Nit: consistency of spelling (comment below says "model analyzer", one later in the document also says "model-analyzer")

dyastremsky (approved):

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+For those who like a high level overview, below is the common flow for most use cases. For those who wish to jump right in, skip to the [end-to-end example](#end-to-end-example).
+
+1. Is my model compatible with Triton?
+    - If your model falls under one of Triton's [supported backends](https://github.com/triton-inference-server/backend), then we can simply try to deploy the model as described in the [Quickstart](https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md) guide. For the ONNXRuntime, TensorFlow SavedModel, and TensorRT backends, the model configuration can be completely inferred from the model using Triton's [AutoComplete](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#auto-generated-model-configuration) feature, so no `config.pbtxt` file is required. Additionally, by enabling verbose logging via `--log-verbose=1`, you can see the complete config that Triton sees internally in the server log output. For other backends, refer to the [Minimal Model Configuration](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#minimal-model-configuration) required to get started.
```
tanmayv25, line 38: minimal model configuration can be inferred.... ..., so `config.pbtxt` is not essentially required.

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# From Model to Deployment
+
+Given a trained model, how do I deploy it at-scale with an
+optimal configuration using Triton Inference Server? This document is
+here to help answer that.
+
+For those who like a high level overview, below is the common flow for most use cases. For those who wish to jump right in, skip to the [end-to-end example](#end-to-end-example).
```
tanmayv25, line 35: Try to break a single long line to multiple lines.

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+    - If your model does not come directly from a supported backend, but is executable through Python, you may be interested in writing a simple python script for running your model through the [Python Backend](https://github.com/triton-inference-server/python_backend).
```
tanmayv25, line 39: The way it is written it looks like we are recommending using python backend. Python backend is easier to write but suffers from performance issue. I would recommend not using if, otherwise construct and point out the alternatives. We can talk about how C++ custom backend can be more performant.

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+
+    $ perf_analyzer -m my_model
+
+    ...
+
+    Inferences/Second vs. Client Average Batch Latency
+    Concurrency: 1, throughput: 482.8 infer/sec, latency 12613 usec
+    ```
+
+    - This gives us a sanity test that we are able to successfully form input requests and receive output responses to communicate with the model backend via Triton APIs. If Perf Analyzer fails to send requests and it is unclear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects. If the config is correct, check that the model runs successfully using its original framework directly. If you don't have your own script or tool to do so, [Polygraphy](https://github.com/NVIDIA/TensorRT/tree/main/tools/Polygraphy) is a useful tool to run sample inferences on a model via various frameworks.
```
tanmayv25, line 57: Polygraphy does not support all the model frameworks. I think it only supports TensorRT and OnnxRuntime? It would be useful to point that out.

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+    - For manual tuning of your model configuration, there is further information in our [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md) docs.
+
+## Other Areas of Interest
+
+1. My model performs slowly when it is first loaded by Triton (cold-start penalty), what do I do?
+    - Triton exposes the ability to run [ModelWarmup](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#model-warmup) requests when first loading the model to ensure that the model is sufficiently initialized before being marked "READY" for inference.
```
tanmayv25, line 70: *sufficiently warmed up before

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+|---|---|---|---|---|---|---|---|
+| alexnet_config_4 | 0 | Enabled | 5/GPU | 22.333 | 1675.98 | 6437 | 31.7 |
+| alexnet_config_3 | 0 | Enabled | 4/GPU | 43.26 | 1612.95 | 5809 | 53.4 |
+| alexnet_config_2 | 0 | Enabled | 3/GPU | 44.403 | 1563.16 | 5117 | 55.1 |
+| alexnet_config_default | 0 | Disabled | 1/GPU | 8.274 | 1001.77 | 3786 | 24.6 |
+
+In the table above, we see that setting our GPU [Instance Count](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#instance-groups) to 5 allows us to achieve the highest throughput and lowest latency for this example. Note that our this "alexnet" model has a fixed batch-size that is explicitly specified in the first dimension of the Input/Output `dims`, therefore the `max_batch_size` parameter is set to 0 as described [here](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#maximum-batch-size). For models that support dynamic batch size, Model Analyzer would also tune the `max_batch_size` parameter.
```
tanmayv25, line 215: You should talk here about how these configs are system specific. For a smaller GPU instance_count of 5 may be counter-productive.

tanmayv25, line 215: > For models that support dynamic batch size,

How does model analyzer detect the support for dynamic batch size?

docs/walkthrough.md
```diff
@@ -0,0 +1,231 @@
...
+# (optional) Backup our original config.pbtxt (if any) to another directory
+cp /mnt/models/alexnet/config.pbtxt /tmp/original_config.pbtxt
+
+# Copy over the optimal config.pbtxt from Model Analyzer results to our model repository
+cp ./results/alexnet_config_4/config.pbtxt /mnt/models/alexnet/
+```
+
+Now that we have an optimized Model Configuration, we are ready to take our model to deployment. For further manual tuning, read the [Model Configuration](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md) and [Optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md#) docs to learn more about Triton's complete set of capabilities.
+
+In this example, we happened to get both the highest throughput and lowest latency from the same configuration, but in some cases this is a tradeoff that must be made. Certain models or configurations may achieve a higher throughput but also incur a higher latency in return. It is worthwhile to fully inspect the reports generated by Model Analyzer to ensure your model performance meets your requirements.
```
nv-kmcgill53, line 231: I like this concluding paragraph!

nv-kmcgill53 (dismissed):

dyastremsky (dismissed):

docs/walkthrough.md
rmccorm4: I'm not sure what to put here currently, but I think it is not necessarily something a new user needs to know here - it should be clear that it was tuned from the MA results showing different batch sizes or not.

I addressed the rest of your feedback though.

Tabrizian (commented):
You also need to provide a link to this guide from the user guide: 

https://github.com/triton-inference-server/server/tree/main/docs#user-guide

docs/walkthrough.md
```diff
@@ -0,0 +1,261 @@
...
+model-analyzer analyze --analysis-models=alexnet
+```
+
+Example Model Analyzer output summary:
+
+> In 41 measurements across 6 configurations, `alexnet_config_4` provides the best throughput: 1676 infer/sec.
+>
+> **This is a 67% gain over the default configuration (1002 infer/sec), under the given constraints.**
+
+| Model Config Name | Max Batch Size | Dynamic Batching | Instance Count | p99 Latency (ms) | Throughput (infer/sec) | Max GPU Memory Usage (MB) | Average GPU Utilization (%) |
```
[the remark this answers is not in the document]
rmccorm4: @Tabrizian is this actively being worked on? High GPU util is a common user requirement/metric, so I think it's a big value add to prospective users knowing that MA reports it.

docs/walkthrough.md
```diff
@@ -0,0 +1,261 @@
...
+    - If none of the above can help get sufficient GPU-accelerated performance for your model, the model may simply be better designed for CPU execution and the [OpenVINO Backend](https://github.com/triton-inference-server/openvino_backend) may help further optimize your CPU execution.
+
+## End-to-end Example
+
+> NOTE: If you have never worked with Triton before, you may be interested in first checking out the [Quickstart](https://github.com/triton-inference-server/server/blob/main/docs/quickstart.md) example. 
+Some basic understanding of Triton may be useful for the following section, but this example is meant to be straightforward enough without prior experience.
+
+Let's take an ONNX model as our example since ONNX is designed to be a format that can be [easily exported](https://github.com/onnx/tutorials#converting-to-onnx-format) from most other frameworks.
+
+1. Download our example Alexnet model from the [ONNX Model Zoo](https://github.com/onnx/models/blob/main/vision/classification/alexnet/model/bvlcalexnet-12.onnx).
```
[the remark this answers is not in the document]
rmccorm4: I will use our `densenet_onnx` example model instead

docs/walkthrough.md
```diff
@@ -0,0 +1,261 @@
...
+
+```bash
+# Create model repository
+mkdir -p ./models/alexnet/1
+
+# Download model and place it in model repository
+wget -O models/alexnet/1/model.onnx https://github.com/onnx/models/blob/main/vision/classification/alexnet/model/bvlcalexnet-12.onnx?raw=true
+```
```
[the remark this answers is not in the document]
rmccorm4: I will use our `densenet_onnx` example model instead

rmccorm4:
Probably final rounds of review now.

jbkyang-nvi (dismissed):

dyastremsky (dismissed):

krishung5 (dismissed):

docs/walkthrough.md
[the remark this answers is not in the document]
Tabrizian: Yes, this is being worked on but it may not be in MA/PA until 22.09.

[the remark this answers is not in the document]
rmccorm4: I think it is more enticing to the user to see it, and this is the actual output of MA currently (at least referring to the table). I am happy to update this doc with better numbers once this bug has been fixed. I don't think the current numbers here are suspicious enough to immediately raise questions, and users will get this in their output if they try it themselves.

Tabrizian (approved):

docs/performance_tuning.md
```diff
@@ -0,0 +1,285 @@
...
+    The Python Backend provides a simple interface to execute requests through a generic python script, but may not be as performant as a Custom C++ Backend. 
+    Depending on your use case, the Python Backend performance may be a sufficient tradeoff for the simplicity of implementation.
+
+2. Can I run inference on my served model?
+    - Assuming you were able to load your model on Triton, the next step is to verify that we can run inference requests and get a baseline performance benchmark of your model. 
+    Triton's [Perf Analyzer](https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md) tool specifically fits this purpose. 
+    Here is a simplified output for demonstration purposes:
+
+    ```
+    # NOTE: "my_model" represents the model currently being served by
```
tgerdesnv, line 56: What is the 'previous step' in this context?

docs/performance_tuning.md
```diff
@@ -0,0 +1,285 @@
...
+    Concurrency: 1, throughput: 482.8 infer/sec, latency 12613 usec
+    ```
+
+    - This gives us a sanity test that we are able to successfully form input requests and receive output responses to communicate with the model backend via Triton APIs. 
+    If Perf Analyzer fails to send requests and it is unclear from the error how to proceed, then you may want to sanity check that your model `config.pbtxt` inputs/outputs match what the model expects. 
+    If the config is correct, check that the model runs successfully using its original framework directly. 
+    If you don't have your own script or tool to do so, [Polygraphy](https://github.com/NVIDIA/TensorRT/tree/main/tools/Polygraphy) is a useful tool to run sample inferences on your model via various frameworks. 
+    Currently, Polygraphy supports ONNXRuntime, TensorRT, and TensorFlow 1.x.
+    - The definition of "performing well" is subject to change for each use case. 
+    Some common metrics are throughput, latency, and GPU utilization.
```
tgerdesnv, line 73: Note that PA doesn't currently provide GPU utilization, although it will soon (22.09?)

docs/performance_tuning.md
```diff
@@ -0,0 +1,285 @@
...
+    If the config is correct, check that the model runs successfully using its original framework directly. 
+    If you don't have your own script or tool to do so, [Polygraphy](https://github.com/NVIDIA/TensorRT/tree/main/tools/Polygraphy) is a useful tool to run sample inferences on your model via various frameworks. 
+    Currently, Polygraphy supports ONNXRuntime, TensorRT, and TensorFlow 1.x.
+    - The definition of "performing well" is subject to change for each use case. 
+    Some common metrics are throughput, latency, and GPU utilization. 
+    There are many variables that can be tweaked just within your model configuration (`config.pbtxt`) to obtain different results.
+    - As your model, config, or use case evolves, [Perf Analyzer](https://github.com/triton-inference-server/server/blob/main/docs/perf_analyzer.md) is a great tool to quickly verify model functionality and performance.
+
+3. How can I improve my model performance?
+    - To further understand the best model configuration you can provide to Triton for your use case, Triton's [Model Analyzer](https://github.com/triton-inference-server/model_analyzer) tool can help.
```
tgerdesnv, line 78: One more sentence after this to give a high level overview of what MA does would be helpful.  Something like "Model Analyzer can automatically or manually search through config combinations to find the optimal triton configuration to meet your constraints.

docs/performance_tuning.md
```diff
@@ -0,0 +1,285 @@
...
+
+3. How can I improve my model performance?
+    - To further understand the best model configuration you can provide to Triton for your use case, Triton's [Model Analyzer](https://github.com/triton-inference-server/model_analyzer) tool can help. 
+    After running Model Analyzer to find the optimal configurations for your model/use case, you can transfer the generated config files to your [Model Repository](https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md). 
+    Model Analyzer provides a [Quickstart](https://github.com/triton-inference-server/model_analyzer/blob/main/docs/quick_start.md) guide with some examples to walk through.
+    - Upon serving the model with the newly optimized configuration file found by Model Analyzer and running Perf Analyzer again, you should expect to find better performance numbers in most cases compared to a default config.
+    - Some parameters that can be tuned for a model may not be exposed to Model Analyzer since they don't apply to all models.
```
tgerdesnv, line 82: Are these parameters set somewhere other than the config.pbtxt? MA can manually run through anything in config.pbtxt if you tell it to.

tgerdesnv (changes requested):

docs/performance_tuning.md
rmccorm4: You're right, it's a bit confusing here - I'll remove that.

docs/performance_tuning.md
rmccorm4: Added reference and links to manual config search

tgerdesnv (approved):

nv-kmcgill53 (approved):

dyastremsky (approved):

PR #4707, opened 2022-07-28
Add cmdline option to set GPU limit for loading

GuanLuo:


src/main.cc
```diff
@@ -618,13 +619,20 @@ std::vector<Option> options_
...
-        "Currently supported settings are 'numa-node', 'cpu-cores'. Note that "
-        "'numa-node' setting will affect pinned memory pool behavior, see "
-        "--pinned-memory-pool for more detail."
+    OPTION_MODEL_LOAD_GPU_LIMIT, "model-load-gpu-limit",
+        "<device_id>:<fraction>",
+        "Specify the limit on GPU memory usage in fraction. If model loading "
+        "on "
+        "the device is requested and the current memory usage exceeds the "
+        "limit,"
+        "the load will be rejected. If not specified, the limit will not set."
```
rmccorm4, line 635:
```suggestion
        "limit,"
        "the load will be rejected. If not specified, the limit will not be set."
```

src/main.cc
```diff
@@ -618,13 +619,20 @@ std::vector<Option> options_
...
   {
-    OPTION_HOST_POLICY, "host-policy", "<string>,<string>=<string>",
-        "Specify a host policy setting associated with a policy name. The "
-        "format of this flag is --host-policy=<policy_name>,<setting>=<value>. "
-        "Currently supported settings are 'numa-node', 'cpu-cores'. Note that "
-        "'numa-node' setting will affect pinned memory pool behavior, see "
-        "--pinned-memory-pool for more detail."
+    OPTION_MODEL_LOAD_GPU_LIMIT, "model-load-gpu-limit",
+        "<device_id>:<fraction>",
+        "Specify the limit on GPU memory usage in fraction. If model loading "
```
rmccorm4, line 631:
```suggestion
        "<device_id>:<fraction>",
        "Specify the limit on GPU memory usage as a fraction. If model loading "
```

src/main.cc
```diff
@@ -1099,8 +1107,9 @@ ParseFloatOption(const std::string arg)
 }
 #endif
 
+template <>
 double
-ParseDoubleOption(const std::string arg)
+ParseOption(const std::string& arg)
```
rmccorm4, line 1112: Won't `ParseOption<int>(optarg)` still return a double?

src/main.cc
```diff
@@ -1702,6 +1712,9 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
       case OPTION_HOST_POLICY:
         host_policies.push_back(ParseHostPolicyOption(optarg));
         break;
+      case OPTION_MODEL_LOAD_GPU_LIMIT:
+        load_gpu_limit.emplace(ParsePairOption<int, double>(optarg, ":"));
```
rmccorm4, line 1716: Should we add some kind of check for `device_id < 0` ?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2596,6 +2596,35 @@ def callback(user_data, result, error):
...
+        # After the first load, the memory consumption should have exceeded
+        # the specified limit, load will fail
+        try:
+            triton_client = grpcclient.InferenceServerClient("localhost:8001",
+                                                             verbose=True)
+            triton_client.load_model(model_name + "_2")
+            self.assertTrue(False, "expected error for loading model")
+        except Exception as ex:
+            self.assertIn("memory limit set for GPU 0 has exceeded",
+                          ex.message())
```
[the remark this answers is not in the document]
GuanLuo: No. The device placement is still depending on model config, so users will hit this only if they ask to load on that particular GPU (or they don't and Triton by default load the copies on each of the visible GPU and still hit this for at least one GPU and fail the whole load)

src/main.cc
GuanLuo: There is already a template specialization for `int` above, so it should get resolved to that implementation.

src/main.cc
```diff
@@ -1927,6 +1939,24 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
...
+      exit(1);
+    } else if ((limit.second < 0.0) || (limit.second > 1.0)) {
+      std::cerr << "--model-load-gpu-limit option expects limit fraction to be "
+                   "in range [0.0, 1.0], Got "
+                << limit.second << std::endl;
+      exit(1);
+    }
+    static std::string key_prefix = "model-load-gpu-limit-device-";
+    FAIL_IF_ERR(
+        TRITONSERVER_ServerOptionsSetBackendConfig(
```
Tabrizian, line 1955: Why are we using the `TRITONSERVER_ServerOptionsSetBackendConfig` field? This looks like a ServerOptions and not a backend field.

qa/L0_lifecycle/test.sh
```diff
@@ -1636,8 +1636,8 @@ mkdir models
 cp -r ../python_models/cuda_memory_consumer models/cuda_memory_consumer_1 && \
     cp -r ../python_models/cuda_memory_consumer models/cuda_memory_consumer_2
 
-# Run server to stop model loading if > 30% of GPU 0 memeory is used
-SERVER_ARGS="--model-repository=`pwd`/models --model-control-mode=explicit --model-load-gpu-limit 0:0.3"
+# Run server to stop model loading if > 60% of GPU 0 memeory is used
```
rmccorm4, line 1639:
```suggestion
# Run server to stop model loading if > 60% of GPU 0 memory is used
```

qa/L0_lifecycle/test.sh
```diff
@@ -1636,8 +1636,8 @@ mkdir models
...
+SERVER_ARGS="--model-repository=`pwd`/models --model-control-mode=explicit --model-load-gpu-limit 0:0.6"
```
rmccorm4, line 1640: Maybe add simple negative tests for the checks you added expecting server to fail?

i.e.
```
SERVER_ARGS="--model-repository=`pwd`/models --model-control-mode=explicit --model-load-gpu-limit 0:<invalid_fraction>
```
and
```
SERVER_ARGS="--model-repository=`pwd`/models --model-control-mode=explicit --model-load-gpu-limit <negative_device_id>:0.6
```

src/main.cc
GuanLuo: This may not the intended usage as we didn't document it, but config for "empty backend" are used to pass in global configs which will be accessed by model related components (i.e. we pass `backend-directory` this way). It is probably more consistent if I add new API in a similar way to how we set backend directory

rmccorm4 (approved):

rmccorm4 (approved):
Nice changes and tests :rocket:

PR #4710, opened 2022-07-29
Add test case for config.pbtxt check for models not supporting auto-c…

tanmayv25:
…omplete

### DIFF
--- a/qa/L0_model_config/test.sh
+++ b/qa/L0_model_config/test.sh
@@ -226,6 +226,11 @@ done
 mkdir -p special_cases/invalid_platform/1
 cp -r /data/inferenceserver/${REPO_VERSION}/qa_model_repository/savedmodel_float32_float32_float32/1/model.savedmodel \
     special_cases/invalid_platform/1/
+# Note that graphdef models don't support auto-complete-config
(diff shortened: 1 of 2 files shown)


qa/L0_model_config/test.sh
```diff
@@ -297,7 +302,14 @@ done
 
 # Run special test cases
 for TARGET in `ls special_cases`; do
-    SERVER_ARGS="--model-repository=`pwd`/models --strict-model-config=true"
+    case $TARGET in
+        "invalid_platform")
```
rmccorm4, line 306: should this be `noautofill_noconfig` ?

tanmayv25: It is little counter-intuitive. noautofill in this model means that the underlying model doesn't support auto-complete. But to test this special scenario we need the server to run in default mode(with auto-complete enabled).

The scenario being tested here is that for models not supporting auto-complete we throw error if there is no config.pbtxt file provided.

I am open to suggestion for better nomenclature.

Tabrizian (approved):

rmccorm4 (approved):

GuanLuo (approved):

Issue #4720, opened 2022-08-01
Trace summary script doesn't correctly handle splitting string values


brightsparc:
**Description**
The trace summary python file doesn't correctly support splitting strings due to the the `get_numpy_array` only expect floats:

```
...
Data Flow:
	==========================================================
	Name:   titanic_ensemble
	Version:1
	QUEUE_INPUT:
Traceback (most recent call last):
  File "trace_summary.py", line 426, in <module>
    summarize_dataflow(trace_data)
  File "trace_summary.py", line 301, in summarize_dataflow
    print_dataflow_tensor(dataflow_tree_map, traces, depth[0], step=0)
  File "trace_summary.py", line 322, in print_dataflow_tensor
    print_tensor_by_id(parent_id, traces, depth, step)
  File "trace_summary.py", line 349, in print_tensor_by_id
    get_numpy_array(trace["tensor"])))
  File "trace_summary.py", line 394, in get_numpy_array
    array = np.array(list(value), dtype=dtype)
ValueError: could not convert string to float: '"3"'
```

**Triton Information**
What version of Triton are you using?

nvcr.io/nvidia/tritonserver:22.05-pyt-python-py3
Are you using the Triton container or did you build it yourself?

triton container
Are you using the Triton container or did you build it yourself?

**To Reproduce**
I created a ensemble file that has a string input

```
name: "titanic_ensemble"
platform: "ensemble"
max_batch_size: 0 # Dynamic batch not supported
input [
  {
    name: "Pclass"
    data_type: TYPE_STRING
    dims: [ 1 ]
  },
...
```

See also #4636 with more detail on the full example

The PR #4717 to fix writing out correctly scape strings is still pending, so I've used the following to clean up trace file:

```
sed 's/:\"\"/:\"\\"/g;s/\"\",/\\"\",/g' trace.json
```

Then. I was able to get the `trace_summary.py` file to run which hit the above error.

**Expected behavior**
I would expect that the `get_numpy_array` function would handle strings correctly.  Suggest using a csv reader which is able to handle the comma delimiter within quoted values correct eg:

```
import csv
next(csv.reader(['"3", "a, b", "C"'], skipinitialspace=True))
```

PR #4721, opened 2022-08-01
fix: Update to support parsing string tensor data

brightsparc:
Fixes: #4720 

Added support for splitting strings in `get_numpy_array` using python csv import, example 

```
>>> import csv
>>> next(csv.reader(['"3", "a, b", "C"'], skipinitialspace=True))
['3', 'a, b', 'C']
```

### DIFF
--- a/qa/common/trace_summary.py
+++ b/qa/common/trace_summary.py
@@ -27,8 +27,8 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import argparse
+import csv
 import json
-import sys
 import numpy as np
 
 FLAGS = None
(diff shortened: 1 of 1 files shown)


tanmayv25:
@brightsparc  Can you complete and send the [CLA](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla)?

qa/common/trace_summary.py
```diff
@@ -389,7 +389,10 @@ def find_first_id_with_tensor(traces):
 
 def get_numpy_array(tensor):
     dtype = TRITON_TYPE_TO_NUMPY[tensor["dtype"]]
-    value = map(float, tensor["data"].split(","))
+    if dtype == np.object_:
+        value = next(csv.reader([tensor["data"]], skipinitialspace=True))
```
rmccorm4, line 393: Would something like `value = map(str, tensor["data"].split(","))` work here instead? This would align more closely with the existing code.

Calling `next` just once to retrieve a single element from an iterator seems unintuitive here. Can `csv.reader(...)` yield an iterator for multiple elements for example?

rmccorm4: Or even just `value = tensor["data"].split(",")` ?

brightsparc: Using split doesn’t handle the scenario where the comma delimiter is inside a quoted string which is why I am using the `CSV`

brightsparc:
> Can you complete and send the [CLA](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla)?

Thanks, I have sent the CLA through via my employer Predibase.

brightsparc:
Any progress on landing this fix?

tanmayv25 (approved):

tanmayv25:
Thanks for your contribution!

PR #4723, opened 2022-08-01
Fix error handling for TritonJson APIs

krishung5:


src/http_server.cc
```diff
@@ -515,7 +516,7 @@ WriteDataToJson(
           output_name, byte_size, sizeof(uint8_t) * element_count));
       const uint8_t* cbase = reinterpret_cast<const uint8_t*>(base);
       for (size_t e = 0; e < element_count; ++e) {
-        data_json->AppendUInt(cbase[e]);
+        RETURN_IF_ERR(data_json->AppendUInt(cbase[e]));
```
nv-kmcgill53, line 519: Maybe more of a question for @GuanLuo, is there a reason why in server this macro is `RETURN_IF_ERR` and not `RETURN_IF_ERROR`?

rmccorm4 (approved):
Will probably want to run NIGHTLY tests for these changes in case we're breaking tests that were silently failing before.

src/http_server.cc
tanmayv25: I don't think there is a strong reasoning behind this. It might just be because the way it was defined before: https://github.com/triton-inference-server/server/blob/main/src/common.h#L48

tanmayv25 (approved):

GuanLuo (approved):

PR #4729, opened 2022-08-02
Updating scripts to accept default variables

mc-nv:


qa/common/gen_qa_model_repository
```diff
@@ -64,29 +64,29 @@ TENSORRT_IMAGE=${TENSORRT_IMAGE:=nvcr.io/nvidia/tensorrt:$TRITON_VERSION-py3}
...
-HOST_VARIMPLICITSEQDESTDIR=/tmp/$TRITON_VERSION/qa_variable_sequence_implicit_model_repository
-HOST_INITIALSTATEIMPLICITSEQDESTDIR=/tmp/$TRITON_VERSION/qa_sequence_initial_state_implicit_model_repository
-HOST_VARINITIALSTATEIMPLICITSEQDESTDIR=/tmp/$TRITON_VERSION/qa_variable_sequence_initial_state_implicit_model_repository
-HOST_TORCHTRTDESTDIR=/tmp/$TRITON_VERSION/torchtrt_model_store
+HOST_BUILD_DIR=${HOST_BUILD_DIR:=/tmp}
+HOST_SRCDIR=$HOST_BUILD_DIR/gen_srcdir
+HOST_DESTDIR=$HOST_BUILD_DIR/$TRITON_VERSION/qa_model_repository
+HOST_VARDESTDIR=$HOST_BUILD_DIR/$TRITON_VERSION/qa_variable_model_repository
+HOST_IDENTITYDESTDIR=$HOST_BUILD_DIR/$TRITON_VERSION/qa_identity_model_repository
+HOST_SIGDEFDESTDIR=$HOST_BUILD_DIR/$TRITON_VERSION/qa_tf_tag_sigdef_repository
```
tanmayv25, line 72: I think you have missed `qa_tf_parameters_repository`.

qa/common/gen_xavier_trt_models
```diff
@@ -50,13 +50,13 @@
 ##
 ############################################################################
 
-TRITON_VERSION=22.07
+TRITON_VERSION=${TRITON_VERSION:=22.07}
 CUDA_DEVICE=0
 
-REPO_ARCH="xavier"
```
tanmayv25, line 56: We don't need `*_xavier` in the model directory name? Isn't that useful in differentiating the model stores for different platforms?

tanmayv25 (changes requested):

qa/common/gen_qa_model_repository
mc-nv: I have restored it

Tabrizian (commented):
I was wondering does this PR fix the nested directory issue?

qa/common/gen_xavier_trt_models
mc-nv: This script designed to generate models, so that folder is required if you trying to differentiate the local directories. 
I can restore it back, but I removed it to keep some consistency. <br>
Important is the directory that we passing to the `test.sh` script.

mc-nv:
> I was wondering does this PR fix the nested directory issue?
This PR will change our source for the large models, it means we can stop using those links, and literally by removing target directory my guess that we will remove possibility to create nested links

GuanLuo (approved):

PR #4732, opened 2022-08-03
Allow multiple servers to bind on the same http/grpc port

krishung5:
- [ ] Implementation
- [ ] Testing


qa/L0_socket/client.py
```diff
@@ -0,0 +1,139 @@
...
+            threading.Thread(target=infer,
+                             args=(FLAGS.protocol, FLAGS.url, FLAGS.verbose)))
+
+    for t in threads:
+        t.start()
+
+    for t in threads:
+        t.join()
+
+    sys.exit(EXIT_CODE_)
```
GuanLuo, line 139: Use `unittest` instead..

GuanLuo: or just reuse existing client

qa/common/util.sh
```diff
@@ -392,3 +392,45 @@ function check_test_results () {
...
+
+    if [ -z "$SERVER_LD_PRELOAD" ]; then
+        echo "=== Running $SERVER $SERVER_ARGS"
+    else
+        echo "=== Running LD_PRELOAD=$SERVER_LD_PRELOAD $SERVER $SERVER_ARGS"
+    fi
+
+    LD_PRELOAD=$SERVER_LD_PRELOAD $SERVER $SERVER1_ARGS > $SERVER1_LOG 2>&1 &
+    SERVER1_PID=$!
+    sleep 10
```
GuanLuo, line 421: why sleep

qa/common/util.sh
```diff
@@ -392,3 +392,45 @@ function check_test_results () {
...
+    fi
+
+    if [ ! -f "$SERVER" ]; then
+        echo "=== $SERVER does not exist"
+        return
+    fi
+
+    SERVER1_PID=0
+    SERVER2_PID=0
+    SERVER3_PID=0
```
GuanLuo, line 411: can you make this helper function dynamic

qa/common/util.sh
```diff
@@ -392,3 +392,45 @@ function check_test_results () {
...
+
+    if [ -z "$SERVER_LD_PRELOAD" ]; then
+        echo "=== Running $SERVER $SERVER_ARGS"
+    else
+        echo "=== Running LD_PRELOAD=$SERVER_LD_PRELOAD $SERVER $SERVER_ARGS"
```
GuanLuo, line 416: Not printing out message correctly

src/grpc_server.cc
```diff
@@ -4462,12 +4462,13 @@ GRPCServer::GRPCServer(
     const std::shared_ptr<TRITONSERVER_Server>& server,
     triton::server::TraceManager* trace_manager,
     const std::shared_ptr<SharedMemoryManager>& shm_manager,
-    const std::string& server_addr, bool use_ssl, const SslOptions& ssl_options,
-    const int infer_allocation_pool_size,
+    const std::string& server_addr, const bool is_port_reused, bool use_ssl,
```
GuanLuo, line 4465: change "is_port_reused" to "reuse_port" everywhere

qa/common/util.sh
krishung5: Thanks for the catch. Fixed

qa/common/util.sh
krishung5: Done

qa/common/util.sh
krishung5: We needed to make sure the first server is launched successfully and the other two servers fail. We don't need sleep anymore after refactoring the helper function.

qa/L0_socket/client.py
krishung5: Reused our existing clients. Removed this script.

GuanLuo (approved):

PR #4740, opened 2022-08-04
Backend Autocompletion of Custom Backends

dyastremsky:
Related PRs
Core: https://github.com/triton-inference-server/core/pull/111


docs/model_configuration.md
```diff
@@ -296,8 +296,9 @@ and [Onnxruntime](https://github.com/triton-inference-server/onnxruntime_backend
 backends as examples of how to acheive this. Currently, only
 [inputs, outputs](#inputs-and-outputs), [max_batch_size](#maximum-batch-size)
 and [dynamic batching](#dynamic-batcher) settings can be populated by
-backend. You would still need to provide a config.pbtxt with
-`backend` field for these custom backends.
+backend. For custom backends, your config.pbtxt file must
+include a `backend` field or your model name must be in the
+form `<model_name>.<backend_name>`.
```
nv-kmcgill53, line 301: >`<model_name>.<backend_name>`.

Should be "model.<backend_name>" the way you have coded it.

qa/L0_model_config/autofill_noplatform_success/custom/no_backend.identity/config.pbtxt
```diff
@@ -0,0 +1,41 @@
...
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
nv-kmcgill53, line 25: Do we need the copyright on all other configs? I don't think so. cc @tanmayv25

nv-kmcgill53 (changes requested):

docs/model_configuration.md
dyastremsky: We might be talking about two different things here. To make it [consistent with what we expect for the other frameworks](https://github.com/triton-inference-server/server/blob/main/docs/model_repository.md#model-files), we'd want the model file to be named model.<backend>. However, the model name is pulled from the model directory unless the user creates a model mapping to override it (directory->model_name).

So, for example, a model `add_sub` would be in the subdirectory `add_sub` in the model repository. But in `model_repository/add_sub/1`, there would be a model file named `model.<backend>`.

Would you prefer it done or documented a different way? We don't need to set the default model filename for a custom backend, but I thought it would be a nice touch to be consistent with the other backends, so users can use that field if they so desire.

nv-kmcgill53: I think that you're right, I was thinking of something different. Thank you for clarifying, I think the way you have it documented should be fine.

nv-kmcgill53 (approved):

qa/L0_model_config/autofill_noplatform_success/custom/no_backend.identity/config.pbtxt
dyastremsky: I think we typically put it. I must have used one of the configs in the model folder (e.g. custom_models in the qa folder or the identity configs from an internal directory), so I included the copyright.

I'd rather avoid delaying the merging of this PR, so I removed it for now or consistency. We can either have a separate PR to add the copyrights in or leave as is.

PR #4759, opened 2022-08-11
Add L0_response_stats test

Tabrizian:


qa/L0_response_stats/response_stats_test.py
```diff
@@ -0,0 +1,157 @@
...
+        outputs.append(grpcclient.InferRequestedOutput('OUT'))
+
+        self._client.async_stream_infer(model_name=self._model_name,
+                                        inputs=inputs,
+                                        outputs=outputs)
+        self._wait_until_responses_complete(number_of_responses)
+
+    def _check_success_duration(self, duration, count):
+        self.assertEqual(duration['count'], count)
+        self.assertGreater(duration['ns'], 0)
```
tanmayv25, line 92: Does it make sense to add some custom delays in square backend that should be treated as compute_infer in generating responses?
These delays can be provided as backend parameters in the config.
We can then set it to that value instead of 0.

qa/L0_response_stats/response_stats_test.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _check_fail_duration(self, duration):
```
tanmayv25, line 94: We should test with a failing inference too. We can provide the expected count as another parameter.

tanmayv25 (changes requested):

qa/L0_response_stats/response_stats_test.py
kthui: Added custom delays.

tanmayv25 (approved):

rmccorm4 (approved):

PR #4770, opened 2022-08-16
Fix MacOS ARM docker building

alxmamaev:
This little fix allows to build `tritonserver` docker image on MacOS with M-series chip. 
Actually in MacOS with arm chip `arch` command will return `arm64`, while on linux its returns `aarch64`, it's broke *miniconda* installing, because they have no install script for `Linux-arm64`.

### DIFF
--- a/build.py
+++ b/build.py
@@ -841,7 +841,9 @@ RUN curl -o /tmp/cuda-keyring.deb \
 
 def install_miniconda(conda_version, target_machine):
     if target_machine == "arm64":
-        # Arm architecture on MacOS named arm64
+        # This branch used for the case when linux container builds on MacOS with ARM chip
+        # macos arm arch names "arm64" when in linux it's names "aarch64".
+        # So we just replace the architecture to able find right conda version for Linux 
         target_machine = "aarch64"
     if conda_version == '':
         fail(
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -840,6 +840,9 @@ def install_dcgm_libraries(dcgm_version, target_machine):
 
 
 def install_miniconda(conda_version, target_machine):
+    if target_machine == "arm64":
+        # Arm architecture on MacOS named arm64
+        target_machine = "aarch64"
```
Tabrizian, line 845: Conda does have a MacOS-arm64 build too. The reason that we need to change this line is because you are running Linux container on Mac? If so, can you update the comment to mention this.

Tabrizian (commented):
Thank you for your contribution.

build.py
alxmamaev: @Tabrizian yes, running linux container on mac, ok I will update it now

Tabrizian (approved):
Can you sign the CLA as instructed here: https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla

Thank you!

alxmamaev:
@Tabrizian Done

PR #4794, opened 2022-08-19
Fix inferentia gen_pytorch_model

kthui:
Related PR: https://github.com/triton-inference-server/python_backend/pull/182


jbkyang-nvi (approved):

tanmayv25 (dismissed):

GuanLuo:
Should add comment on the import, otherwise it may accidentally be removed because LGTM reports unused?

tanmayv25:
Interestingly it still generates the alert. Is there a way to suppress the alert altogether?

kthui:
> Interestingly it still generates the alert. Is there a way to suppress the alert altogether?

https://github.com/github/codeql/issues/3293
I think it will suppress once it is merged.

tanmayv25 (approved):

jbkyang-nvi (approved):

PR #4830, opened 2022-08-30
Add OpenVINO autocomplete tests

kthui:
Core PR: https://github.com/triton-inference-server/core/pull/118
OpenVINO backend PR: https://github.com/triton-inference-server/openvino_backend/pull/44


GuanLuo:
Are the model files the same? If so we can keep one copy and let `test.sh` to prepare the model directories.

kthui:
Yes. The models are grouped.

kthui:
Rebased.

tanmayv25:
@kthui  Why are we putting model files in the repo? Why not use models from the model generator scripts? If there are few models that are missing, I would advice improving our gen scripts.

kthui:
Add documentation about the model files in the repo, and adjusted the openvino dynamic_batch case to account for parsing default_max_batch_size.

tanmayv25 (approved):

PR #4835, opened 2022-08-31
Add environment variable in SM entrypoint to specify TF version

nskool:
SM customers cannot specify TF version since Triton only supports switching via cmdline, related issue: https://github.com/aws/deep-learning-containers/issues/2176

Adding this environment variable to allow specifying TF version as default when starting Triton via the serve script.

### DIFF
--- a/docker/sagemaker/serve
+++ b/docker/sagemaker/serve
@@ -70,6 +70,9 @@ fi
 if [ -n "$SAGEMAKER_TRITON_SHM_GROWTH_BYTE_SIZE" ]; then
     SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --backend-config=python,shm-growth-byte-size=${SAGEMAKER_TRITON_SHM_GROWTH_BYTE_SIZE}"
 fi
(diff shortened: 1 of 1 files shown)


GuanLuo (commented):
Triton recently switched the TF lib default to TF 2, and I was told that it causes some confusion on SM side. Do you want to add `--backend-config=tensorflow,version=1` the default setting if user doesn't specify the TF version env var?

nskool:
> Triton recently switched the TF lib default to TF 2, and I was told that it causes some confusion on SM side. Do you want to add `--backend-config=tensorflow,version=1` the default setting if user doesn't specify the TF version env var?

I think it's better to follow the experience that users come to experience from Triton on SageMaker. If a certain version of Triton has a default as TF2 (e.g. starting 22.08), so users may expect that to be true on SM as well. It makes sense to follow the default as Triton sets it, and allow users to explicitly set a specific version later.

GuanLuo (approved):

joostwestra:
Thanks for the patch. Think this helps many people.

PR #4852, opened 2022-09-08
Fix L0_perf_resnet script. run_server perform "set -e" as side effect which can cause unexpected exit

GuanLuo:


qa/L0_perf_resnet/run_test.sh
```diff
@@ -89,11 +87,13 @@ else
     # Must warmup similar to actual run so that all instances are ready
     # Note: Running extra PA for warmup doesn't make sense for C API since it
     # uses in-process tritonserver which will exit along with this PA process.
+    set +e
```
mc-nv, line 90: I may be wrong but to enable immediate exit on error we bash syntax is `set -e` and to disable it `set +e`...

GuanLuo: this command is allowed to fail due to instability from warmup and that is why this is run to warm up the model. We don't want to exit the script as it is allowed to fail

mc-nv (approved):

PR #4873, opened 2022-09-13
Fix L0_custom_ops failure

tanmayv25:
@GuanLuo  We should relnote this work-around in our release docs.

### DIFF
--- a/qa/L0_custom_ops/test.sh
+++ b/qa/L0_custom_ops/test.sh
@@ -111,7 +111,10 @@ LD_LIBRARY_PATH=/opt/tritonserver/backends/pytorch:$LD_LIBRARY_PATH
 
 # Pytorch
 SERVER_ARGS="--model-repository=/data/inferenceserver/${REPO_VERSION}/qa_custom_ops/libtorch_custom_ops"
-SERVER_LD_PRELOAD="/data/inferenceserver/${REPO_VERSION}/qa_custom_ops/libtorch_custom_ops/libtorch_modulo/custom_modulo.so"
+# FIXME: Pre-loading the python library system to satisfy the symbol definitions
+# as the custom op library is built with different python version within
+# pytorch container. See DLIS-4152.
+SERVER_LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libpython3.8.so.1:/data/inferenceserver/${REPO_VERSION}/qa_custom_ops/libtorch_custom_ops/libtorch_modulo/custom_modulo.so"
 run_server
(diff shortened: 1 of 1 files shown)


mc-nv (approved):

rmccorm4 (approved):

qa/L0_custom_ops/test.sh
```diff
@@ -111,7 +111,10 @@ LD_LIBRARY_PATH=/opt/tritonserver/backends/pytorch:$LD_LIBRARY_PATH
 
 # Pytorch
 SERVER_ARGS="--model-repository=/data/inferenceserver/${REPO_VERSION}/qa_custom_ops/libtorch_custom_ops"
-SERVER_LD_PRELOAD="/data/inferenceserver/${REPO_VERSION}/qa_custom_ops/libtorch_custom_ops/libtorch_modulo/custom_modulo.so"
+# FIXME: Pre-loading the python library system to satisfy the symbol definitions
+# as the custom op library is built with different python version within
+# pytorch container. See DLIS-4152.
+SERVER_LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libpython3.8.so.1:/data/inferenceserver/${REPO_VERSION}/qa_custom_ops/libtorch_custom_ops/libtorch_modulo/custom_modulo.so"
```
Tabrizian, line 117: Do we need to bring this up with the DLFW team? I thought all the Python versions are aligned between different containers. I was curious how does this fix the issue? Is`/usr/lib/x86_64-linux-gnu/libpython3.8.so.1` the same version as the Python version used in the PyTorch container?

tanmayv25: Yes. The change here is more like a workaround. I will initiate a conversation. I have filed a follow-up ticket to look into it with more details. I am not sure about the exact reason as well. 
> Is/usr/lib/x86_64-linux-gnu/libpython3.8.so.1 the same version as the Python version used in the PyTorch container?

It appears the custom_modulo.so does not link properly with python library on our container and runs into undefined symbols issues.  Pre-loading the python library ensures the symbols are available for the custom_modulo.so.

Tabrizian: > It appears the custom_modulo.so does not link properly with python library on our container and runs into undefined symbols issues. Pre-loading the python library ensures the symbols are available for the custom_modulo.so.

Maybe only the linking process has changed from 22.08 -> 22.09 and the Python version difference always existed?

Tabrizian (approved):

PR #4900, opened 2022-09-20
Initialize Local Request ID to Empty String

dyastremsky:
Initialize local request ID to empty string. This is to prevent null pointers from being accessed.

### DIFF
--- a/src/grpc_server.cc
+++ b/src/grpc_server.cc
@@ -3896,7 +3896,7 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
           "unable to retrieve request ID string");
     }
 
-    if ((request_id == "") || (request_id[0] == '\0')) {
+    if ((!strcmp(request_id, "")) || (request_id[0] == '\0')) {
       request_id = "<id_unknown>";
     }
     if (err == nullptr) {
@@ -4324,7 +4324,7 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
     LOG_TRITONSERVER_ERROR(
         TRITONSERVER_InferenceRequestId(irequest, &request_id),
         "unable to retrieve request ID string");
-    if ((request_id == "") || (request_id[0] == '\0')) {
+    if (!strcmp(request_id, "") || (request_id[0] == '\0')) {
       request_id = "<id_unknown>";
     }
     if (err == nullptr) {
--- a/src/grpc_server.cc
+++ b/src/grpc_server.cc
@@ -3889,14 +3889,14 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
           InferResponseComplete, reinterpret_cast<void*>(state));
     }
     // Get request ID for logging in case of error.
-    const char* request_id = nullptr;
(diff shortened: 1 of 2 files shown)


src/grpc_server.cc
```diff
@@ -3889,14 +3889,14 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
-    const char* request_id = nullptr;
+    const char* request_id = "";
     if (irequest != nullptr) {
       LOG_TRITONSERVER_ERROR(
           TRITONSERVER_InferenceRequestId(irequest, &request_id),
           "unable to retrieve request ID string");
     }
 
-    if ((request_id == nullptr) || (request_id[0] == '\0')) {
+    if ((!strcmp(request_id, "")) || (request_id[0] == '\0')) {
```
rmccorm4, line 3899: For simplicity and readability, can we just do this instead?

```suggestion
    if (!std::strncmp(request_id, "", 1)) {
```

See godbolt example for equivalency, null terminator check is redundant: https://gcc.godbolt.org/z/3EaqbE64e

src/grpc_server.cc
```diff
@@ -3889,14 +3889,14 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
-    const char* request_id = nullptr;
+    const char* request_id = "";
     if (irequest != nullptr) {
       LOG_TRITONSERVER_ERROR(
           TRITONSERVER_InferenceRequestId(irequest, &request_id),
           "unable to retrieve request ID string");
     }
 
-    if ((request_id == nullptr) || (request_id[0] == '\0')) {
+    if ((!strncmp(request_id, "", 1))) {
```
rmccorm4, line 3899: extra parens
```suggestion
    if (!strncmp(request_id, "", 1)) {
```

src/http_server.cc
```diff
@@ -2883,7 +2883,7 @@ HTTPAPIServer::HandleInfer(
       LOG_TRITONSERVER_ERROR(
           TRITONSERVER_InferenceRequestId(irequest, &request_id),
           "unable to retrieve request ID string");
-      if ((request_id == nullptr) || (request_id[0] == '\0')) {
+      if ((!strncmp(request_id, "", 1))) {
```
rmccorm4, line 2886: extra parens
```suggestion
      if (!strncmp(request_id, "", 1)) {
```

src/http_server.cc
```diff
@@ -3065,9 +3065,9 @@ HTTPAPIServer::InferRequestClass::FinalizeResponse(
   triton::common::TritonJson::Value response_json(
       triton::common::TritonJson::ValueType::OBJECT);
 
-  const char* request_id = nullptr;
+  const char* request_id = "";
   RETURN_IF_ERR(TRITONSERVER_InferenceResponseId(response, &request_id));
-  if ((request_id != nullptr) && (request_id[0] != '\0')) {
+  if ((strncmp(request_id, "", 1))) {
```
rmccorm4, line 3070: extra parens
```suggestion
  if (strncmp(request_id, "", 1)) {
```

rmccorm4 (approved):

PR #4943, opened 2022-10-04
Update NGC versions post-22.09 release

mc-nv:


build.py
```diff
@@ -67,10 +67,12 @@
 # incorrectly load the other version of the openvino libraries.
 #
 TRITON_VERSION_MAP = {
+<<<<<<< HEAD
```
tanmayv25, line 70: This shouldn't be here.

tanmayv25 (changes requested):

build.py
mc-nv: I have removed the value from there

dyastremsky (approved):

PR #4946, opened 2022-10-05
Add TensorRT uint8 models and tests

kthui:


qa/common/gen_qa_models.py
```diff
@@ -493,12 +495,29 @@ def create_plan_dynamic_rf_modelfile(models_dir, max_batch, model_version,
 
     in0 = network.add_input("INPUT0", trt_input_dtype, input_with_batchsize)
     in1 = network.add_input("INPUT1", trt_input_dtype, input_with_batchsize)
+
+    # uint8 must be converted before any operation
```
tanmayv25, line 499: Can you elaborate more about why we are casting uint8 into float32?
Such as TRT identity layer does not support uint8 datatype yet.

tanmayv25 (changes requested):

tanmayv25:
We should also create add_sub models for uint8. You can cast uint8 into fp32 and feed it to add and sub layers if they don't support uint8 i/o too. Our goal should be to enable testing for uint8 in all the L0_infer variants at the very least.

qa/common/gen_qa_models.py
kthui: Yes. Added "TRT uint8 cannot be used to represent quantized floating-point value yet".

kthui:
Expanded model generation and added new models to L0_infer.

tanmayv25 (approved):

Issue #4416, opened 2022-05-20
Allow loading/unloading of specific version for a given model


mmischitelli:
**Is your feature request related to a problem? Please describe.**
I find the model control mode "explicit" to be very interesting. Although, for my usecase, it pretty much useless as we have few models consisting of tens or even hundred of versions. Each model is basically updated weekly and previous versions must be kept available.

**Describe the solution you'd like**
A new overload to the LoadModel method, accepting not only the model name, but also the version.

**Describe alternatives you've considered**
An additional API, such as LoadVersion, which still sports two arguments: model name and version.

**Additional context**
My aim here is basically keeping the memory usage under control. I don't want to load all the versions of a given model in memory as it would clog to death my client, as well as requiring my server to have terabytes of Ram.

dyastremsky:
I've filed a ticket for this feature request.

I don't know your specific use case, but you may also be able to use model names in the meantime (e.g. model100_2022_05_23 or whatever makes sense for you). Another option is leveraging the [specific version policy](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#version-policy). We've added this to our queue to review.

mmischitelli:
Thanks!

Unfortunately we cannot "flatten" the model/version hierarchy as we have some strict requirements that would make things way too complicated. For instance, we plot the inference results on a chart: if the function's name is the same we can esily do comparisons, as the version is just a parameter.

The specific version policy instead is kinda "static". Our solution provides users with the backend to run simulations on their own, eventually comparing data from different years. Since we produce model versions weekly, it would result in way too much memory being occupied for no real reason (except for being able to serve all the versions available).

If there was support for an on-demand load/unload of versions, we would then be able to implement some kind of LRU policy for managing the available resources.

Lingkar:
@mmischitelli combining the specific version policy with the control mode POLL is quite fast in updating the available models, right?

@dyastremsky My specific use-case will be controlling the model versioning through the API within the KServe modelmesh environment. I will have to find out if either workaround will work. Where can I keep an eye on the progress of the feature request ticket? Would you happen to have an indication of the timeline for such a feature? Thanks for the information!

dyastremsky:
@Lingkar The ticket was filed internally. There is no timeline yet for this feature request. Your best bet is to keep an eye on this GitHub issue, as I'm leaving it open as we look into this feature request.

mmischitelli:
[...]

PR #4948, opened 2022-10-05
Check if the format of override model config has extra fields

krishung5:
Loading model with override model config is working properly. The [issue](https://github.com/triton-inference-server/server/issues/4416) was that the config that the user used has some extra fields but no error message was shown.
Fix the error checking for invalid override config in Core repo: https://github.com/triton-inference-server/core/pull/124
Client repo: https://github.com/triton-inference-server/client/pull/170

### DIFF
--- a/src/grpc_server.cc
+++ b/src/grpc_server.cc
@@ -2037,6 +2037,17 @@ CommonHandler::SetUpAllRequests()
           std::vector<const TRITONSERVER_Parameter*> const_params;
           for (const auto& param_proto : request.parameters()) {
             if (param_proto.first == "config") {
+              if ((param_proto.second.string_param().find("config") !=
+                   std::string::npos) ||
+                  (param_proto.second.string_param().find("parameters")) !=
+                      std::string::npos) {
+                err = TRITONSERVER_ErrorNew(
+                    TRITONSERVER_ERROR_INVALID_ARG,
+                    "invalid value for config: please remove the extra "
+                    "'parameters' or 'config' field from the override config.");
+                break;
+              }
+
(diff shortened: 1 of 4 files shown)


src/grpc_server.cc
```diff
@@ -2037,6 +2037,17 @@ CommonHandler::SetUpAllRequests()
           std::vector<const TRITONSERVER_Parameter*> const_params;
           for (const auto& param_proto : request.parameters()) {
             if (param_proto.first == "config") {
+              if ((param_proto.second.string_param().find("config") !=
+                   std::string::npos) ||
+                  (param_proto.second.string_param().find("parameters")) !=
+                      std::string::npos) {
```
GuanLuo, line 2043: I don't think this is the proper way to check it, what if the user nests "some_other_field"? This check is going to miss that. The question seems to be why the model manager fails to recognize the config is ill-formed and reject the model loading.

krishung5: Seems like I didn't think it thoroughly enough. I have fixed the error checking for the ill-formed config in the core repo. Reverted the previous changes. Please review.

dyastremsky (dismissed):
Looks good! Great work on these tickets.

krishung5:
Updated test scripts according to new changes.

dyastremsky (dismissed):

krishung5:
@dyastremsky There was some typo, my bad. Please re-approve, thank you!
Passed on the [CI](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/6114332).

dyastremsky (approved):

PR #4982, opened 2022-10-14
Using alternative managed symbolic links  for library paths

mc-nv:
**OBSERVATIONS:**
Triton server uses `build.py` to generate `Dockerfile` for CPU only builds.
One of the steps for it is the `COPY` library dependencies.
As of now those library dependencies are using version specific path instead of using CUDA defined symlinks with [alternatives](https://linux.die.net/man/8/alternatives) .

**HYPOTHESIS**
Change hardcoded version values to the defined alternatives.

**BENEFITS:**
Triton server build became less version agnostic


mc-nv:
By mistake I squash the commits and some additional conversation can be found here: 
https://github.com/triton-inference-server/server/commit/d6151d2c3784c43f067923558bc97dfee491c2dc

rmccorm4 (approved):

PR #4991, opened 2022-10-18
Update the limitation of multiple servers binding to the same http/grp…

krishung5:
…c port

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -402,7 +402,7 @@ std::vector<Option> options_
       {OPTION_REUSE_HTTP_PORT, "reuse-http-port", Option::ArgBool,
        "Allow multiple servers to listen on the same HTTP port when every "
        "server has this option set. If you plan to use this option as a way to "
-       "load-balance between different triton servers, the same model "
+       "load balance between different Triton servers, the same model "
        "repository or set of models must be used for every server."},
       {OPTION_HTTP_ADDRESS, "http-address", Option::ArgStr,
        "The address for the http server to binds to."},
@@ -417,7 +417,7 @@ std::vector<Option> options_
       {OPTION_REUSE_GRPC_PORT, "reuse-grpc-port", Option::ArgBool,
        "Allow multiple servers to listen on the same GRPC port when every "
        "server has this option set. If you plan to use this option as a way to "
-       "load-balance between different triton servers, the same model "
+       "load balance between different Triton servers, the same model "
        "repository or set of models must be used for every server."},
       {OPTION_GRPC_ADDRESS, "grpc-address", Option::ArgStr,
        "The address for the grpc server to binds to."},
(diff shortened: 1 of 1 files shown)


dyastremsky:
Are we sure this is the best place for this documentation? I think users usually check our documentation (e.g. .MD files) rather than the command line arguments help screen. Especially since it's quite long. If we want users to know about and understand these options, it might make sense to also include them in documentation.

krishung5:
@dyastremsky I might be wrong but I couldn't find any documentation regarding anything about http/grpc port except in `main.cc`. It seems like we ask users to use[ `tritonserver --help`](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/inference_protocols.md#grpc-options) to see those options that are not in the MD files. That's why I only updated `main.cc`, but I'm willing to add those options to the document if this makes more sense.

src/main.cc
```diff
@@ -401,7 +401,9 @@ std::vector<Option> options_
        "The port for the server to listen on for HTTP requests."},
       {OPTION_REUSE_HTTP_PORT, "reuse-http-port", Option::ArgBool,
        "Allow multiple servers to listen on the same HTTP port when every "
-       "server has this option set."},
+       "server has this option set. The same set of models/same model "
```
Tabrizian, line 404: "If you plan to use this option as a way to load-balance between different triton servers, the same model repository or set of models must be used for every server."


> Note that this feature only supports stateless models.

I think it might be better to remove this sentence since the customers may figure out a way to control which server gets the requests and address this limitation.

dyastremsky:
> @dyastremsky I might be wrong but I couldn't find any documentation regarding anything about http/grpc port except in `main.cc`. It seems like we ask users to use[ `tritonserver --help`](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/inference_protocols.md#grpc-options) to see those options that are not in the MD files. That's why I only updated `main.cc`, but I'm willing to add those options to the document if this makes more sense.

I see, okay. Sounds like we're being consistent here then.

src/main.cc
krishung5: I see. Updated the document, thanks for the comment!

src/main.cc
```diff
@@ -401,7 +401,9 @@ std::vector<Option> options_
        "The port for the server to listen on for HTTP requests."},
       {OPTION_REUSE_HTTP_PORT, "reuse-http-port", Option::ArgBool,
        "Allow multiple servers to listen on the same HTTP port when every "
-       "server has this option set."},
+       "server has this option set. If you plan to use this option as a way to "
+       "load-balance between different triton servers, the same model "
```
dyastremsky, line 405: Here and below:
- Triton should be capitalized.
- Load balance should not have a dash.

krishung5: Thanks for the catch! Updated.

dyastremsky (approved):

Tabrizian (approved):

PR #5005, opened 2022-10-21
Add tolerance to cpu metrics test

rmccorm4:
The CPU utilization metric has occasionally been failing the L0_metrics test by having duplicate values. This can be expected to happen if the server is either not doing much work or doing some consistent work. I propose adding some tolerance for number of duplicate values for the utilization and memory metrics. They will still fail the test if consistently returning the same value for more than the defined tolerance (50%).

### DIFF
--- a/qa/L0_metrics/test.sh
+++ b/qa/L0_metrics/test.sh
@@ -169,22 +169,22 @@ WAIT_INTERVAL_SECS=2.0
 for metric in ${CPU_METRICS}; do
     echo "\n=== Checking Metric: ${metric} ===\n"
     prev_value=`curl -s localhost:8002/metrics | grep ${metric} | grep -v "HELP\|TYPE" | awk '{print $2}'`
-    echo "  Initial ${metric} value: ${prev_value}"
 
     num_not_updated=0
+    num_not_updated_threshold=3
     for (( i = 0; i < $num_iterations; ++i )); do
       sleep $WAIT_INTERVAL_SECS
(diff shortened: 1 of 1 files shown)


qa/L0_metrics/test.sh
```diff
@@ -171,23 +171,31 @@ for metric in ${CPU_METRICS}; do
...
       current_value=`curl -s localhost:8002/metrics | grep ${metric} | grep -v "HELP\|TYPE" | awk '{print $2}'`
       echo "  ${metric} value: ${current_value}"
       if [ $current_value == $prev_value ]; then
+        num_not_updated=$((num_not_updated+1))
+      fi
+      prev_value=$current_value
+    done
+
+    # Give CPU metrics some tolerance to not update
+    # An alternative may be to run some busy work in the background
```
tanmayv25, line 186: You should describe the new logic in L165-L166. Add the alternative for busy work as a TODO and create a jira ticket for the same.

rmccorm4: I wasn't sure if there was any benefit to introducing busy work or not, moreso just offering a future alternative if this tolerance wasn't OK - so I didn't want to add a JIRA ticket for something we may never do. Do you think it's worth adding?

tanmayv25: If the test keeps failing then it would be a nice add to ensure the test robustness.

rmccorm4: added ticket, added threshold variable, updated comments

tanmayv25 (approved):

PR #5006, opened 2022-10-22
Add checks for TRT uint8 support

kthui:


qa/common/gen_qa_models.py
```diff
@@ -512,10 +512,10 @@ def create_plan_dynamic_rf_modelfile(models_dir, max_batch, model_version,
     out1 = sub if not swap else add
 
     # uint8 conversion after operations
-    if trt_output0_dtype == trt.uint8:
+    if output0_dtype == np.uint8:
         out0 = network.add_identity(out0.get_output(0))
         out0.set_output_type(0, trt.uint8)
```
GuanLuo, line 517: Won't there still be the chance of referring `trt.uint8` here? Would be safer to have `hasattr(trt, "uint8")` check and skip uint8 all together if not supported?

nv-kmcgill53: I will second this comment. In the context of uint8, we should skip over this completely in Jetson until this data type is supported.

kthui: Yes, it is possible if the caller did not check with the added `support_trt_uint8()`, and skip all uint8 generations.
I think we can unify the check by always calling `support_trt_uint8()`, and have it removed when TRT uint8 is fully supported.

tanmayv25 (approved):

GuanLuo (approved):

PR #5011, opened 2022-10-25
fix GKE testing error

mengdong:


mc-nv:
I'm going to decline your change, as Triton team storing/uses these files as an examples . 
We don't encourage our customers stay with the given values, you can adjust configuration in way which you may found suitable for your infrastructure needs.

mc-nv:
> I'm going to decline your change, as Triton team storing/uses these files as an examples . We don't encourage our customers stay with the given values, you can adjust configuration in way which you may found suitable for your infrastructure needs.

@mengdong  - I see you made a changes to the `values.yml` and `schema.yml` may I know what is the reason of such a change. 

`values.yml` can be easily overwritten by user same as `schema.yml` 

I also see a pitfall for those who will use these values as default and decide to store models in same location by overriding them over the time...

mengdong:
Thanks @mc-nv , could we keep this open for a bit of time before we fix everything, this is just temp workaround

PR #5019, opened 2022-10-28
Increase the L0_backend_python client_timeout

Tabrizian:


GuanLuo (approved):
Do you know why more time is needed?

Tabrizian:
@GuanLuo Some of the tests copy large tensors (around 1GB). I think in some systems this might take longer which is the reason why the test fails.

PR #5030, opened 2022-11-02
Enable request rate manager test for C-API

Tabrizian:


matthewkotila:
What's the backstory for this? Was there another PR that implemented request rate mode in C API mode recently? If so, can you link it?

Tabrizian:
@matthewkotila The request rate was already supported. I just enabled the test. @jbkyang-nvi mentioned that there were some segfaults and that's why the test was disabled. We resolved a couple of segfaults in the C-API which seems to have resolved the request rate segfault too.

matthewkotila (approved):

PR #5032, opened 2022-11-02
Add concurrent loading speedup test

kthui:
Add a case to L0_lifecycle to ensure a speedup is observed when two models are loading in parallel.
Core PR: https://github.com/triton-inference-server/core/pull/137

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2625,6 +2625,39 @@ class LifeCycleTest(tu.TestResultCollector):
         except Exception as ex:
             self.assertTrue(False, "unexpected error {}".format(ex))
 
+    def test_concurrent_load_speedup(self):
+        model_names = ["identity_zero_1_int32_1", "identity_zero_1_int32_2"]
+        start_time = time.time()
+        try:
+            triton_client = grpcclient.InferenceServerClient("localhost:8001",
+                                                             verbose=True)
+            threads = []
+            for model_name in model_names:
+                threads.append(
+                    threading.Thread(target=triton_client.load_model,
+                                     args=(model_name,)))
+            for thread in threads:
+                thread.start()
+            for thread in threads:
+                thread.join()
+        except Exception as ex:
+            self.assertTrue(False, "unexpected error {}".format(ex))
+        end_time = time.time()
+        loading_time = end_time - start_time
+        # Each of the two models has a minimum loading delay of 10 seconds
+        # Speedup is observed when the concurrent loading time < 20 seconds but
+        # use a tighter bound of 15 seconds
+        self.assertLess(loading_time, 15.0,
+                        "Concurrent loading speedup not observed")
+        # Concurrent loading time cannot be < 10 seconds
+        self.assertGreaterEqual(loading_time, 10.0,
+                                "Invalid concurrent loading time")
+        # Make sure the models are loaded
+        self.assertTrue(triton_client.is_server_live())
+        self.assertTrue(triton_client.is_server_ready())
+        for model_name in model_names:
+            self.assertTrue(triton_client.is_model_ready(model_name))
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2626,22 +2626,24 @@ class LifeCycleTest(tu.TestResultCollector):
             self.assertTrue(False, "unexpected error {}".format(ex))
 
     def test_concurrent_load_speedup(self):
-        model_names = ["identity_zero_1_int32_1", "identity_zero_1_int32_2"]
-        start_time = time.time()
+        # Initialize client
         try:
             triton_client = grpcclient.InferenceServerClient("localhost:8001",
                                                              verbose=True)
-            threads = []
-            for model_name in model_names:
-                threads.append(
-                    threading.Thread(target=triton_client.load_model,
-                                     args=(model_name,)))
-            for thread in threads:
-                thread.start()
-            for thread in threads:
-                thread.join()
         except Exception as ex:
             self.assertTrue(False, "unexpected error {}".format(ex))
+        # Load both models concurrently
+        model_names = ["identity_zero_1_int32_1", "identity_zero_1_int32_2"]
+        threads = []
+        for model_name in model_names:
+            threads.append(
+                threading.Thread(target=triton_client.load_model,
+                                 args=(model_name,)))
+        start_time = time.time()
+        for thread in threads:
+            thread.start()
+        for thread in threads:
+            thread.join()
         end_time = time.time()
         loading_time = end_time - start_time
         # Each of the two models has a minimum loading delay of 10 seconds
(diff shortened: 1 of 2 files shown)


qa/L0_lifecycle/slow_load_repo/identity_slow_load/config.pbtxt
```diff
@@ -0,0 +1,49 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "identity_slow_load"
+backend: "identity"
```
GuanLuo, line 28: Avoid using / extending identity backend if possible, can't observe the speed up with two Python model?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2625,6 +2625,32 @@ def test_load_gpu_limit(self):
...
+                thread.start()
+            for thread in threads:
+                thread.join()
+        except Exception as ex:
+            self.assertTrue(False, "unexpected error {}".format(ex))
+        end_time = time.time()
+        loading_time = end_time - start_time
+        # Each of the two models has a minimum loading delay of 10 seconds
+        # Speedup is observed when the concurrent loading time < 20 seconds
+        self.assertTrue(loading_time < 20.0,
```
GuanLuo, line 2649: set a tighter bound, like 15

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2625,6 +2625,32 @@ def test_load_gpu_limit(self):
         except Exception as ex:
             self.assertTrue(False, "unexpected error {}".format(ex))
 
+    def test_concurrent_load_speedup(self):
```
GuanLuo, line 2628: And you need more tests on concurrent load / unload combination of the same model, and also for ensemble where the operations are on different models but there is collision in the underlying graph

qa/L0_lifecycle/slow_load_repo/identity_slow_load/config.pbtxt
kthui: I can't observe the speed up with two Python models:
```
I1103 18:30:44.668903 3610 model_lifecycle.cc:459] loading: python_slow_load_1:1
I1103 18:30:44.669170 3610 model_lifecycle.cc:459] loading: python_slow_load_2:1
I1103 18:30:45.940105 3610 python_be.cc:1856] TRITONBACKEND_ModelInstanceInitialize: python_slow_load_1 (GPU device 0)
I1103 18:30:56.127921 3610 python_be.cc:1856] TRITONBACKEND_ModelInstanceInitialize: python_slow_load_2 (GPU device 0)
I1103 18:30:56.127991 3610 model_lifecycle.cc:694] successfully loaded 'python_slow_load_1' version 1
I1103 18:31:06.266695 3610 model_lifecycle.cc:694] successfully loaded 'python_slow_load_2' version 1
```
(Each Python model will delay for 10 seconds during initialization)
I guess the Python is using the same interpreter for both models. If so, then the GIL will remove any parallelism which explains the result.

However, the speed up is observed with two Identity models:
```
I1103 18:31:58.640473 4141 model_lifecycle.cc:459] loading: identity_slow_load_2:1
I1103 18:31:58.640715 4141 model_lifecycle.cc:459] loading: identity_slow_load_1:1
...skip...
I1103 18:31:58.642488 4141 identity.cc:232] Creation delay is set to: 10
I1103 18:31:58.642509 4141 identity.cc:232] Creation delay is set to: 10
I1103 18:32:08.645463 4141 identity.cc:662] TRITONBACKEND_ModelInstanceInitialize: identity_slow_load_1 (GPU device 0)
I1103 18:32:08.646140 4141 model_lifecycle.cc:694] successfully loaded 'identity_slow_load_1' version 1
I1103 18:32:08.646241 4141 identity.cc:662] TRITONBACKEND_ModelInstanceInitialize: identity_slow_load_2 (GPU device 0)
I1103 18:32:08.646658 4141 model_lifecycle.cc:694] successfully loaded 'identity_slow_load_2' version 1
```

GuanLuo: @Tabrizian don't we use separate processes to work around the GIL?

qa/L0_lifecycle/lifecycle_test.py
kthui: Updated

qa/L0_lifecycle/slow_load_repo/identity_slow_load/config.pbtxt
kthui: Since the L0_lifecycle already use one identity model, maybe we can stick with identity models instead?

kthui:
Rebased. Changed models to only use identity.

qa/L0_lifecycle/slow_load_repo/identity_slow_load/config.pbtxt
Tabrizian: Yes, we do use separate processes to work around GIL. Are we sure that separate `TRITONBACKEND_ModelInstanceInitialize` are called for each Python backend model instance?

qa/L0_lifecycle/lifecycle_test.py
kthui: Added concurrent load/unload of the same model, and unload dependency while an ensemble model is loading.

qa/L0_lifecycle/slow_load_repo/identity_slow_load/config.pbtxt
GuanLuo: > Since the L0_lifecycle already use one identity model, maybe we can stick with identity models instead?

Given the situation, we will have to use identity backend for now. If possible, I will avoid adding any more use of identity backend, it is used in some tests is a legacy usage (don't have Python backend back then). The reason is that C/C++ backend is not flexible for testing where the models often need to execute in certain way and all those configurations are stuff into single backend file, while Python backend model is more flexible and clear in the way that you can define different model in dedicated model file. What's more, since there is dedicated model file, it is easier to look into the file and understand how exactly the model operates

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2625,6 +2625,95 @@ def test_load_gpu_limit(self):
...
+                                                             verbose=True)
+        except Exception as ex:
+            self.assertTrue(False, "unexpected error {}".format(ex))
+        # Load identity_zero_1_int32 and unload it while it is loading
+        # The unload operation should have no effect
+        threads = [
+            threading.Thread(target=triton_client.load_model,
+                             args=("identity_zero_1_int32",)),
+            threading.Thread(target=triton_client.unload_model,
+                             args=("identity_zero_1_int32",))
```
GuanLuo, line 2676: check if error message is returned for failed request

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2625,6 +2625,95 @@ def test_load_gpu_limit(self):
...
+        for thread in threads:
+            thread.start()
+            time.sleep(2)  # wait between load and unload
+        for thread in threads:
+            thread.join()
+        self.assertTrue(triton_client.is_server_live())
+        self.assertTrue(triton_client.is_server_ready())
+        self.assertTrue(triton_client.is_model_ready("identity_zero_1_int32"))
+        # Load ensemble_zero_1_float32 and unload its dependency while it is loading
+        # The unload operation should have no effect
```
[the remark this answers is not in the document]
GuanLuo: And what about directly loading two ensembles that has the same depending model. And if they will fail, should the requests be failing?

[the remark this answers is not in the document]
kthui: Both cases above have two conflicting load requests, the request that comes later (but before the earlier request can finish) will fail.
The user should not issue concurrent load requests that have overlapping models.

qa/L0_lifecycle/lifecycle_test.py
kthui: Yes. Added.

qa/L0_lifecycle/lifecycle_test.py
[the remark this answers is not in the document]
GuanLuo: It's okay to fail the request and return error for now, but it is really not a good user experience that whether a model is ensemble is hidden from user: imagine the models are instead named "image_segmentation", "resnet", user sees them as two independent models (we want them to see model unanimously) and naturally thought concurrent load should be fine.

My original idea was that we don't fail a request that has conflict but hold on to the requests and proceed the requests in chronological order when "in-transit" is unmarked.

Let's just keep it simple and return error to avoid additional code change at the last moment, but keep in mind the other approach which may be requested in the future.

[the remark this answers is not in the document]
kthui: Sure

kthui:
Added test for loading the same model concurrently.
Rebased.

GuanLuo (approved):

PR #5036, opened 2022-11-03
Add test for TRT data dependent shape output. Add generation script for TRT data-dependent shape model

GuanLuo:


qa/L0_trt_data_dependent_shape/trt_data_dependent_shape_test.py
```diff
@@ -0,0 +1,89 @@
...
+    def setUp(self):
+        self.triton_client = client.InferenceServerClient(
+            "localhost:8000", verbose=True)
+
+    def test_fixed(self):
+        model_name = "plan_nobatch_nonzero_fixed"
+        input_np = np.arange(16, dtype=np.int32).reshape((4, 4))
+        expected_output_np = np.nonzero(input_np)
+
+        # Use shared memory to bypass the shape check in client library, because
```
tanmayv25, line 47: The comment mention that we must use shared_memory to bypass some shape checks but it looks like you are using standard numpy arrays below to set the inputs.

qa/L0_trt_data_dependent_shape/trt_data_dependent_shape_test.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_dynamic(self):
+        model_name = "plan_nobatch_nonzero_dynamic"
+        input_data = []
+        for i in range(20*16):
+            input_data.append(i if (i % 2) == 0 else 0)
+        input_np = np.array(input_data, dtype=np.int32).reshape((20, 16))
+        expected_output_np = np.nonzero(input_np)
+
+        # Use shared memory to bypass the shape check in client library, because
```
tanmayv25, line 71: Same here. I don't see shared memory in use here.

qa/L0_trt_data_dependent_shape/trt_data_dependent_shape_test.py
GuanLuo: Yes, stale comment from the reference test file. Removed.

tanmayv25 (approved):

PR #5038, opened 2022-11-05
Update Go Test

dyastremsky:
Update Go test to generate Go stubs and packages from the client repository. This makes it so that we have one source of truth (the client script).

This PR updates the testing so that the changes in [this client PR](github.com/triton-inference-server/client/pull/138) do not break the L0_simple_go_client test.

### DIFF
--- a/qa/L0_simple_go_client/test.sh
+++ b/qa/L0_simple_go_client/test.sh
@@ -60,8 +60,9 @@ git clone --single-branch --depth=1 -b $TRITON_COMMON_REPO_TAG \
 bash gen_go_stubs.sh
 popd
 
+# Copy to the GOPATH, where Go expects to find packages.
 PACKAGE_PATH="${GOPATH}/src/github.com/triton-inference-server"
-rm -r ${PACKAGE_PATH}/client
+rm -rf ${PACKAGE_PATH}/client
 mkdir -p ${PACKAGE_PATH}
 cp -r client $PACKAGE_PATH
 
--- a/qa/L0_simple_go_client/test.sh
+++ b/qa/L0_simple_go_client/test.sh
@@ -50,6 +50,7 @@ RET=0
 
 # Generate Go stubs.
 rm -fr client common
+## TODO: Remove branch below after CI passes
 git clone -b dyas-go-fix https://github.com/triton-inference-server/client.git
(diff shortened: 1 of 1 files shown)


dyastremsky:
@rmccorm4, do you happen to know where the grpc_simple_client.go is actually coming from? I'm not seeing where it's being pulled from, but it's an older version (test is breaking, because it's expecting the old package name).

rmccorm4:
Looks like it's copied here: https://github.com/triton-inference-server/server/blob/7ed727f00d953f4e68d7a1227b3fcc081cc0bdd9/Dockerfile.QA#L282

rmccorm4 (commented):
FYI rather than update the commands being used in this script, (sourcing package name, tweaking `protoc` command, etc.), my original comment was moreso asking if we can simply do something like:
```bash
# test.sh
rm -fr client
git clone -b dyas-go-fix https://github.com/triton-inference-server/client.git
bash client/src/grpc_generated/go/gen_go_stubs.sh
```
since `gen_go_stubs.sh` appears to be doing a similar `protoc ... ` to generate the stubs.

Then we can remove all of this below logic from the `test.sh` and only have one place (`gen_go_stubs.sh`) to maintain the `protoc` and go package related commands:
```
### Remove all this from `test.sh` and simply call `gen_go_stubs.sh` instead for this logic if possible/equivalent

# Fix to allow global stubs import
sed -i "s/.\/$PACKAGE/$PACKAGE/g" $SIMPLE_GO_CLIENT

PACKAGE_PATH="${GOPATH}/src"
mkdir -p ${PACKAGE_PATH}

# ...

# Requires protoc and protoc-gen-go plugin: https://github.com/golang/protobuf#installation
# Use "M" arguments since go_package is not specified in .proto files.
# As mentioned here: https://developers.google.com/protocol-buffers/docs/reference/go-generated#package)
protoc -I core --go_out=plugins=grpc:${PACKAGE_PATH} --go_opt=Mgrpc_service.proto=./${PACKAGE} \
    --go_opt=Mmodel_config.proto=./${PACKAGE} core/*.proto
```

dyastremsky:
Thanks for clarifying, Ryan! Updated this PR to use that approach.

rmccorm4 (approved):
LGTM if CI passes :rocket:

PR #5056, opened 2022-11-09
Add test for older OS modification time

kthui:
Refer to the ticket DLIS-4310 for details.
Core PR: https://github.com/triton-inference-server/core/pull/139

### DIFF
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -1297,48 +1297,61 @@ for protocol in grpc http; do
     if [[ $protocol == "grpc" ]]; then
        export TRITONSERVER_USE_GRPC=1
     fi
-    rm -fr models simple_float32_float32_float32
-    mkdir models
-    # Prepare two models of different platforms, but with the same name
-    cp -r $DATADIR/qa_model_repository/plan_float32_float32_float32 models/simple_float32_float32_float32
-    sed -i "s/plan_float32_float32_float32/simple_float32_float32_float32/" models/simple_float32_float32_float32/config.pbtxt
-    cp -r $DATADIR/qa_model_repository/libtorch_float32_float32_float32 simple_float32_float32_float32
-    sed -i "s/libtorch_float32_float32_float32/simple_float32_float32_float32/" simple_float32_float32_float32/config.pbtxt
 
-    SERVER_ARGS="--model-repository=`pwd`/models --model-control-mode=explicit \
-                 --load-model=simple_float32_float32_float32 \
-                 --exit-timeout-secs=5"
-    SERVER_LOG="./inference_server_$LOG_IDX.log"
-    run_server
-    if [ "$SERVER_PID" == "0" ]; then
-        echo -e "\n***\n*** Failed to start $SERVER\n***"
-        cat $SERVER_LOG
-        exit 1
-    fi
+    for prep_order in normal reverse; do
+        rm -fr models simple_float32_float32_float32
+        mkdir models
+        # Prepare two models of different platforms, but with the same name
+        if [[ $prep_order == "normal" ]]; then
+            # Prepare the TRT model first, then the pytorch model
+            cp -r $DATADIR/qa_model_repository/plan_float32_float32_float32 models/simple_float32_float32_float32
+            sed -i "s/plan_float32_float32_float32/simple_float32_float32_float32/" models/simple_float32_float32_float32/config.pbtxt
+            cp -r $DATADIR/qa_model_repository/libtorch_float32_float32_float32 simple_float32_float32_float32
+            sed -i "s/libtorch_float32_float32_float32/simple_float32_float32_float32/" simple_float32_float32_float32/config.pbtxt
+        else
+            # Prepare the pytorch model first, then the TRT model
+            # When the replacement model is older, the reload must commence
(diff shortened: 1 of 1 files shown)


tanmayv25:
@kthui  Can you use [os.utime](https://docs.python.org/3/library/os.html#os.utime) to manually change the timestamp metadata within subtest?

kthui:
> @kthui Can you use [os.utime](https://docs.python.org/3/library/os.html#os.utime) to manually change the timestamp metadata within subtest?

I think the OS interpret modification differently than we do. The copy action itself will change the status time, but not modification time, because the file binary is the same from the OS's perspective. We can change triton modification time = max(os modification time, os status time). Then this will fix the problem, without having to change the metadata.

qa/L0_lifecycle/test.sh
```diff
@@ -1297,48 +1297,61 @@ for protocol in grpc http; do
...
+            cp -r $DATADIR/qa_model_repository/libtorch_float32_float32_float32 simple_float32_float32_float32
+            sed -i "s/libtorch_float32_float32_float32/simple_float32_float32_float32/" simple_float32_float32_float32/config.pbtxt
+        else
+            # Prepare the pytorch model first, then the TRT model
+            # When the replacement model is older, the reload must commence
+            cp -r $DATADIR/qa_model_repository/libtorch_float32_float32_float32 simple_float32_float32_float32
+            sed -i "s/libtorch_float32_float32_float32/simple_float32_float32_float32/" simple_float32_float32_float32/config.pbtxt
+            cp -r $DATADIR/qa_model_repository/plan_float32_float32_float32 models/simple_float32_float32_float32
+            sed -i "s/plan_float32_float32_float32/simple_float32_float32_float32/" models/simple_float32_float32_float32/config.pbtxt
+        fi
```
GuanLuo, line 1318: Can you add some explanation on why these two orders are specifically tested? Basically to summarize your finding so reader knows why it is done other than what is done?

qa/L0_lifecycle/test.sh
```diff
@@ -1297,48 +1297,61 @@ for protocol in grpc http; do
...
+            sed -i "s/plan_float32_float32_float32/simple_float32_float32_float32/" models/simple_float32_float32_float32/config.pbtxt
+            cp -r $DATADIR/qa_model_repository/libtorch_float32_float32_float32 simple_float32_float32_float32
+            sed -i "s/libtorch_float32_float32_float32/simple_float32_float32_float32/" simple_float32_float32_float32/config.pbtxt
+        else
+            # Prepare the pytorch model first, then the TRT model
+            # When the replacement model is older, the reload must commence
+            cp -r $DATADIR/qa_model_repository/libtorch_float32_float32_float32 simple_float32_float32_float32
+            sed -i "s/libtorch_float32_float32_float32/simple_float32_float32_float32/" simple_float32_float32_float32/config.pbtxt
+            cp -r $DATADIR/qa_model_repository/plan_float32_float32_float32 models/simple_float32_float32_float32
+            sed -i "s/plan_float32_float32_float32/simple_float32_float32_float32/" models/simple_float32_float32_float32/config.pbtxt
```
GuanLuo, line 1317: And you actually want to add a wait in between, right? Isn't there still be the that the operations can be too fast which results in the same timestamps for both files? Or that is not really the case.

kthui: Same timestamp for both files is ok, because the copy operation will take place much later, which is more than enough for the timestamps to differ.

qa/L0_lifecycle/test.sh
kthui: Yes. It is added in front of the loop that iterates "normal" and "reverse" order.

qa/L0_lifecycle/test.sh
```diff
@@ -1297,48 +1297,67 @@ for protocol in grpc http; do
...
-    if [ "$SERVER_PID" == "0" ]; then
-        echo -e "\n***\n*** Failed to start $SERVER\n***"
-        cat $SERVER_LOG
-        exit 1
-    fi
+    # The OS file system is more granular when determining modification time, a
+    # file is modified when its content is changed in place, and a file is
+    # updated when it is copied or moved. With Triton, any operation that
+    # changes a file is modification. Thus, preparing the models backward will
+    # test when a replacement model is haing an earlier or equal modification
```
GuanLuo, line 1305:
```suggestion
    # test when a replacement model is having an earlier or equal modification
```

qa/L0_lifecycle/test.sh
```diff
@@ -1297,48 +1297,67 @@ for protocol in grpc http; do
...
-    SERVER_LOG="./inference_server_$LOG_IDX.log"
-    run_server
-    if [ "$SERVER_PID" == "0" ]; then
-        echo -e "\n***\n*** Failed to start $SERVER\n***"
-        cat $SERVER_LOG
-        exit 1
-    fi
+    # The OS file system is more granular when determining modification time, a
+    # file is modified when its content is changed in place, and a file is
+    # updated when it is copied or moved. With Triton, any operation that
```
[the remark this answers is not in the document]
kthui: Yes. The statement is updated.

qa/L0_lifecycle/test.sh
kthui: Fixed.

GuanLuo (approved):

PR #5059, opened 2022-11-10
Build and link Triton with tcmalloc

tanmayv25:


zhaohb:
Have you compared the performance of tcmalloc and mimalloc? Which one is better？

PR #5061, opened 2022-11-10
Revert "Using alternative managed symbolic links  for library paths (…

mc-nv:
…#4982)"

This reverts commit f036cb07465df9f03e07126c75896040bd92dc4a.


rmccorm4 (approved):
Maybe we can revisit a more generic solution in the future, even if just copying via wildcard to get the necessary versioned files if that works

Tabrizian (approved):

Issue #4981, opened 2022-10-13
Docker build fail on windows 


RandySheriffH:
**Description**
Build docker image failed on "Dockerfile.win10.min"

**Triton Information**
main 

**To Reproduce**
`docker build -t win10-py3-min2 -f Dockerfile.win10.min .`

Got:

**Step 21/61** : RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
 ---> Running in 4e03abdf511e
Computing installation plan...
A suitable version of cmake was not found (required v3.22.2). Downloading portable cmake v3.22.2...
Downloading cmake...
  https://github.com/Kitware/CMake/releases/download/v3.22.2/cmake-3.22.2-windows-i386.zip -> C:\vcpkg\downloads\cmake-3.22.2-windows-i386.zip
Extracting cmake...
A suitable version of 7zip_msi was not found (required v21.7.0). Downloading portable 7zip_msi v21.7.0...
Downloading 7zip_msi...
  https://www.7-zip.org/a/7z2107-x64.msi -> C:\vcpkg\downloads\7z2107-x64.msi
Extracting 7zip_msi...
The following packages will be built and installed:
    b64[core]:x64-windows -> 2.0.0.1
  * boost-array[core]:x64-windows -> 1.78.0
  * boost-assert[core]:x64-windows -> 1.78.0
  * boost-build[core]:x64-windows -> 1.78.0#1
  * boost-config[core]:x64-windows -> 1.78.0
  * boost-container[core]:x64-windows -> 1.78.0
  * boost-container-hash[core]:x64-windows -> 1.78.0
  * boost-core[core]:x64-windows -> 1.78.0
  * boost-detail[core]:x64-windows -> 1.78.0
  * boost-integer[core]:x64-windows -> 1.78.0
    boost-interprocess[core]:x64-windows -> 1.78.0
  * boost-intrusive[core]:x64-windows -> 1.78.0
  * boost-modular-build-helper[core]:x64-windows -> 1.78.0#3
  * boost-move[core]:x64-windows -> 1.78.0
  * boost-predef[core]:x64-windows -> 1.78.0
  * boost-preprocessor[core]:x64-windows -> 1.78.0
  * boost-smart-ptr[core]:x64-windows -> 1.78.0
[...]

PR #5085, opened 2022-11-18
Update Windows dockerfile

krishung5:
Cherry-picked Misha's commit since the updated windows base container should resolve the docker build issue: https://github.com/triton-inference-server/server/issues/4981


mc-nv (dismissed):

krishung5:
@mc-nv Found some path issues with the previous commit and also reverted the version of CUDNN. We can update the version once it's out moving forward.

mc-nv (approved):

PR #5096, opened 2022-11-21
Set range to mlflow package version

mc-nv:


rmccorm4 (approved):

rmccorm4:
FYI it looks like we also install the package in the test script itself [here](https://github.com/triton-inference-server/server/blob/4fddb4333d45f9ff1106be2b91d83fdca57770cb/qa/L0_mlflow/test.sh#L37), but it ultimately gets overridden with the correct version from the setup.py specified in this PR.

PR #5104, opened 2022-11-23
Update NGC versions post-22.11 release

mc-nv:


Tabrizian (approved):

rmccorm4 (approved):

Dockerfile.win10.min
```diff
@@ -100,9 +100,8 @@ ARG TENSORRT_ZIP="TensorRT-${TENSORRT_VERSION}.Windows10.x86_64.cuda-11.8.cudnn8
 
 LABEL TENSORRT_VERSION="${TENSORRT_VERSION}"
 
-ARG CUDNN_VERSION=8.3.2.44
-ARG CUDNN_DIR=cudnn-windows-x86_64-${CUDNN_VERSION}_cuda11.5-archive
-ARG CUDNN_ZIP=${CUDNN_DIR}.zip
+ARG CUDNN_VERSION=8.7.0.80
+ARG CUDNN_ZIP=cudnn-${CUDNN_VERSION}.zip
```
krishung5, line 104: I think we need to keep make the var `CUDNN_DIR` to be the full file name like it was before, and set `ARG CUDNN_ZIP=${CUDNN_DIR}.zip` so that the name of `CUDNN_ZIP` would be the same as what users download. `ARG CUDNN_ZIP=cudnn-${CUDNN_VERSION}.zip` only works on our windows runners since we pull it from the artifactory.

Issue #2657, opened 2021-03-23
Add support for STSAssumeRoleWebIdentityCredentialsProvider for S3 repositories


DXist:
**Is your feature request related to a problem? Please describe.**
I run Triton on EKS Kubernetes cluster in AWS. I configured an IAM role to access an S3 bucket with models and configuration. Then I associated a K8S serviceaccount to this IAM role as described in https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

Triton pod is configured with the serviceaccount and has valid AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE env variables set by a mutation webhook.

The problem: Triton can't get temporary AWS credentials using AWS STS AssumeRoleWithWebIdentity API operation.

**Describe the solution you'd like**
Triton uses SDK provider to get temporary credentials - https://github.com/aws/aws-sdk-cpp/blob/2be13177875e944151132d90305dbc46e80bf8e3/aws-cpp-sdk-core/source/auth/STSCredentialsProvider.cpp#L32

**Describe alternatives you've considered**
As a workaround I run `aws sts assume-role-with-web-identity` and save temporary credentials in default AWS profile.
Can't use it for poll model control mode since credentials are expired after timeout.

bryantbiggs:
xref #8090

PR #5112, opened 2022-11-24
Provide an ability to use temporary AWS credentials in the AWS Deployment

smcbn:
This should solve #2657.

The solution for this is to remove the secrets within the aws helm chart:
https://github.com/triton-inference-server/server/blob/main/deploy/aws/templates/secrets.yaml
https://github.com/triton-inference-server/server/blob/main/deploy/aws/templates/deployment.yaml#L64-L78

Doing so means that Triton will not use specific credentials to authenticate (https://github.com/triton-inference-server/core/blob/01f1d365d740a49268227ca5dd86cf493ba9cc89/src/filesystem.cc#L1566)

The AWS SDK will then check for numerous environment variables (https://docs.aws.amazon.com/sdk-for-cpp/v1/developer-guide/credentials.html#credother) and if you specify a serviceAccountName for the deployment you will automatically get these into the pod for you by EKS.

I've tested this on an EKS cluster and Triton has successfully pulled the models from the secure s3 bucket.

### DIFF
--- a/deploy/aws/README.md
+++ b/deploy/aws/README.md
@@ -107,6 +107,7 @@ $ aws s3 cp --recursive docs/examples/model_repository s3://triton-inference-ser
 ```
 
 ### AWS Model Repository
+#### Explicit credentials
 To load the model from the AWS S3, you need to convert the following AWS credentials in the base64 format and add it to the values.yaml
 
 ```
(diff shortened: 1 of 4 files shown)


dyastremsky:
Thanks for submtiting this fix! Have you submitted a signed contributor license agreement (CLA) yet? If not, please do so here: https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla

PR #5117, opened 2022-11-28
Updated example to fix deprecation issues.

tanayvarshney:


docs/examples/stable_diffusion/README.md
```diff
@@ -64,7 +65,7 @@ docker run -it --gpus all -p 8888:8888 -v ${PWD}:/mount nvcr.io/nvidia/pytorch:y
 
 pip install transformers ftfy scipy
```
Tabrizian, line 66: Can you fix the versions of other libraries too to ensure the example doesn't break in the future?

tanayvarshney: I don't really use direct functions, just models from the other dependencies, so we should be fine.

Tabrizian (approved):

PR #5119, opened 2022-12-01
Use new Cache C APIs

rmccorm4:
This PR introduces new CLI flags to support the new TRITONCACHE APIs, and updates the caching related documentation. See below for more details on the functional changes.

Corresponding core changes: https://github.com/triton-inference-server/core/pull/157

### --cache-config
- `--cache-config` CLI arg that follows the same format as `--backend-config` arg. However, cache-agnostic (no name provided) configs are not currently supported like they are for backends. Example:
> tritonserver --model-repository models  --cache-config local,size=1048576

- If neither `--cache-config` or `--resposne-cache-byte-size` are provided (no cache configs), then no caches will be created in Triton core and the table printout will show that the cache is not enabled:
```
+----------------------------------+--------------------------------------------------------------------+
| Option                           | Value                                                              |
+----------------------------------+--------------------------------------------------------------------+
| server_id                        | triton                                                             |
| ...
| cache_enabled                    | 0                                                                  |
+----------------------------------+--------------------------------------------------------------------+
```
- If cache configs are provided for exactly one cache type (ex: oneof "local" OR "redis" is provided, such as `--cache-config=local,key=value ...` or `--cache-config=redis,key=value ...`), then the cache will be considered "enabled" in the example table below, and triton will initialize a cache of that type with the provided config:
```
+----------------------------------+--------------------------------------------------------------------+
| Option                           | Value                                                              |
+----------------------------------+--------------------------------------------------------------------+
| server_id                        | triton                                                             |
| ...
| cache_enabled                    | 1                                                                  |
+----------------------------------+--------------------------------------------------------------------+
```
- If cache configs are provided for multiple types (ex: "local" AND "redis" both provided, such as `--cache-config=local,key1=value1 --cache-config=redis,key2=value2` ...), that will fail with something like the following error:
```
I0114 03:32:43.952672 2629989 server.cc:275] No server context available. Exiting immediately.
error: creating server: Invalid argument - found multiple cache configurations, but only a single cache is currently supported
```
### --response-cache-byte-size

- `--response-cache-byte-size 1048576` will fallback to be the equivalent of `--cache-config local,size=1048576`, but issue a warning that it defaults to `local` cache type. This flag won't be compatible with the `redis` cache type. Example:
```
$ tritonserver --model-repository models  --response-cache-byte-size=12345
Warning: '--response-cache-byte-size' has been deprecated! This will default to the 'local' cache implementation with the provided byte size for its config. Please use '--cache-config' instead. The equivalent --cache-config CLI args would be: '--cache-config=local,size=12345'
```
- If both `--cache-config` and `--response-cache-byte-size` are provided, an error will be thrown about incompatible CLI args similar to `--strict-model-config` and `--disable-auto-complete`. Example:
```
$ tritonserver --model-repository models  --cache-config local,size=12345 --response-cache-byte-size 67890
...
Error: Incompatible flags --response-cache-byte-size and --cache-config both provided. Please provide one or the other.
```

### --cache-dir

Similar to `--backend-dir` and `--repoagent-dir` and adds the global search path for cache shared libraries. It defaults to `/opt/tritonserver/caches`, and you would expect to find default implementations in the container at `${CACHE_DIR}/{local,redis}` or `/opt/tritonserver/caches/{local,redis}`

### TODO
- [ ] release container setup
    - [ ] ship server container with built cache lib.so file


Dockerfile.QA
```diff
@@ -70,6 +69,13 @@ RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/nul
     apt-get install -y --no-install-recommends \
       cmake-data=3.21.1-0kitware1ubuntu20.04.1 cmake=3.21.1-0kitware1ubuntu20.04.1
 
+# Install boost version >= 1.78 for boost::span
+# Current libboost-dev apt packages are < 1.78, so install from tar.gz
+RUN wget -O ${TRITONTMP_DIR}/boost.tar.gz \
+        https://boostorg.jfrog.io/artifactory/main/release/1.80.0/source/boost_1_80_0.tar.gz && \
+    (cd ${TRITONTMP_DIR} && tar xzf boost.tar.gz) && \
+    mv ${TRITONTMP_DIR}/boost_1_80_0/boost /usr/include/boost
+
```
rmccorm4, line 78: TODO: check if this is required in QA container. May only be needed as a build time dependency in `build.py`

build.py
```diff
@@ -873,6 +872,13 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
 RUN pip3 install --upgrade pip && \
     pip3 install --upgrade wheel setuptools docker
 
+# Install boost version >= 1.78 for boost::span
+# Current libboost-dev apt packages are < 1.78, so install from tar.gz
+RUN wget -O /tmp/boost.tar.gz \
+        https://boostorg.jfrog.io/artifactory/main/release/1.80.0/source/boost_1_80_0.tar.gz && \
+    (cd /tmp && tar xzf boost.tar.gz) && \
+    mv /tmp/boost_1_80_0/boost /usr/include/boost
```
Tabrizian, line 880: You can also install boost library as a part of core repo CMake: https://github.com/triton-inference-server/python_backend/blob/54c693e5eac685ee343a3d11d4d37494e92286a3/CMakeLists.txt#L96-L105

src/main.cc
```diff
@@ -1725,7 +1730,16 @@ Parse(TRITONSERVER_ServerOptions** server_options, int argc, char** argv)
         cuda_pools.push_back(ParsePairOption<int, uint64_t>(optarg, ":"));
         break;
       case OPTION_RESPONSE_CACHE_BYTE_SIZE:
-        response_cache_byte_size = (uint64_t)ParseLongLongOption(optarg);
+        std::cerr
+            << "Warning: '--response-cache-byte-size' has been deprecated! "
```
Tabrizian, line 1734: Should we return an error here or make some changes to make it backward compatible?

GuanLuo: Agree, should still set it if user specifies it. Internally set the value in cache config if not provided and use local cache

build.py
rmccorm4: I tried copying what you did in Python backend, but ran into some issues with building. Build doesn't seem to find the boost install with `add_depencies(triton-core boostorg)` nor `find_package` after pulling boost 1.80 in through `ExternalProject_Add`. 

I'll leave this as-is for now in server repo to get working builds and focus on the functional tasks. I'll circle back to moving this to core repo build later.

Dockerfile.QA
```diff
@@ -70,6 +69,13 @@ RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/nul
     apt-get install -y --no-install-recommends \
       cmake-data=3.21.1-0kitware1ubuntu20.04.1 cmake=3.21.1-0kitware1ubuntu20.04.1
 
+# Install boost version >= 1.78 for boost::span
+# Current libboost-dev apt packages are < 1.78, so install from tar.gz
+RUN wget -O ${TRITONTMP_DIR}/boost.tar.gz \
+        https://boostorg.jfrog.io/artifactory/main/release/1.80.0/source/boost_1_80_0.tar.gz && \
+    (cd ${TRITONTMP_DIR} && tar xzf boost.tar.gz) && \
+    mv ${TRITONTMP_DIR}/boost_1_80_0/boost /usr/include/boost
```
Tabrizian, line 77: I wonder why we installed boost in both QA and build container but not in the actual linux container? Is boost not needed in the actual linux container and is only required for QA and build

docs/user_guide/model_configuration.md
```diff
@@ -1027,18 +1027,15 @@ for examples on specifying different variants of warmup samples.
...
+  enable: true
 }
 ```
 
-See the [Response
-Cache](response_cache.md)
-and [ModelConfig
-protobuf](https://github.com/triton-inference-server/common/blob/main/protobuf/model_config.proto).
-docs for more information.
+In addition to enabling the cache in the model config, a `--cache-config` must
```
Tabrizian, line 1038: Is it possible to have a default value for local cache to improve the UX and remove one additional step?

docs/user_guide/response_cache.md
```diff
@@ -51,39 +51,131 @@ records that result in the cache so that subsequent inference requests can
...
+[model configuration docs](model_configuration.md#response-cache).
+
+### Cache Implementations
+
+As of the 23.02 release, Triton has a set of TRITONCACHE APIs that are used to
+communicate with a cache implementation of the user's choice. Currently, 
+Triton comes with two cache implementations out of the box: `local` or `redis`.
+These implementations can be found in 
+[tritonserver release containers ](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/tritonserver)
+>= 23.02 in the `/opt/tritonserver/caches` directory.
```
Tabrizian, line 103: This line renders strangely in Github MD.

docs/user_guide/response_cache.md
```diff
@@ -51,39 +51,131 @@ records that result in the cache so that subsequent inference requests can
...
+
+With these new APIs, `tritonserver` exposes a new `--cache-config` CLI flag
+that gives the user flexible customization of which cache implementation
+to use, and how to configure it. Similar the the `--backend-config` flag,
```
Tabrizian, line 107: the the -> to the

Tabrizian (commented):
Few comments, otherwise LGTM!

docs/user_guide/model_configuration.md
rmccorm4: The old behavior was to opt-in on server-side via a non-zero `--response-cache-byte-size` arg, so I tried to mimic that with the `--cache-config` arg rather than require some extra `--enable/disable-cache` flag for server-wide control.

I'd rather keep the same behavior for now, and consider a default server-side choice if there was user complaint. If we wanted to remove this step, we'd need a bit more logic to determine if there is at least one model being served with `response_cache { enable: true }` in its config to allocating a chunk of memory for the cache unnecessarily.

Dockerfile.QA
rmccorm4: We don't have `/usr/include/boost` in the server container, I imagine because it's bundled at compile time.

Looks like we have boost in the QA container because we actually do a couple builds, ex: building python backend for triton-shm-monitor test. However, this newer boost version isn't required for that build, so I think this change is not needed in QA contianer. I'm checking on that now.

Tabrizian: Python backend fetches boost as a part of its own CMake so I think it doesn't need to install it manually.

docs/user_guide/model_configuration.md
Tabrizian: Makes sense. Resolving.

Tabrizian (approved):

PR #5125, opened 2022-12-02
Add LD_PRELOAD for sequence batcher/infer tests

Tabrizian:


krishung5 (approved):

qa/L0_infer/test.sh
```diff
@@ -100,6 +100,13 @@ else
     TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
     SERVER=${TRITON_DIR}/bin/tritonserver
     BACKEND_DIR=${TRITON_DIR}/backends
+
+    # PyTorch on SBSA requires libgomp to be loaded first. See the following
+    # GitHub issue for more information:
+    # https://github.com/pytorch/pytorch/issues/2575
+    if [ $TEST_REPO_ARCH = "arm_a100" ]; then
```
M2Chornyi, line 107: @Tabrizian  
I think there should be `==` sign for condition. 
Also wondering if we can compare against `HOSTTYPE` or `uname -m` result instead of using NV value?

mc-nv, line 107: @Tabrizian  
I think there should be `==` sign for condition. 
Also wondering if we can compare against `HOSTTYPE` or `uname -m` result instead of using NV value?

Tabrizian: Fixed the variable name. Both `=` and `==` will compare the equality. See below for more info:

https://unix.stackexchange.com/a/72042/322680

mc-nv (approved):

krishung5 (approved):

PR #5137, opened 2022-12-05
Improve the documentation for oldest startegy

Tabrizian:


rmccorm4 (commented):
Just my 2 cents, but using code blocks instead (`CONTROL_SEQUENCE_START`, `fp32_false_true`, etc.) will also render underscores correctly and also help signify that it is some kind of special keyword to the reader. 

I don't know if this part is true or not, but may also help when searching for references to this keyword.

ex: this might only show code now and not include any doc files that reference it because of the included backslash. Github search may act similarly.
```
$ grep -R "CONTROL_SEQUENCE_START" .
```

This is mostly personal preference though, I'm fine with this change too.

GuanLuo:
The rendered page looks fine without this change. Can you point out how it was broken?

Tabrizian:
My editor was showing some red markers for these underscores. Looks like it would only an issue if there are two underscores which translates to underline. I have reverted the underscore changes.

rmccorm4 (approved):

docs/user_guide/architecture.md
```diff
@@ -536,11 +536,13 @@ all inference requests in a sequence are routed to the same model
...
-each inference request in the batch belongs to. The
-CONTROL_SEQUENCE_READY control is typically not needed because all
-inferences in the batch will always be ready for inference.
+inferences together. With this strategy the model may need to use the
+CONTROL_SEQUENCE_CORRID control so that it knows which sequence each inference
+request in the batch belongs to. This is typically not required if the model is
+using [implicit state management](#implicit-state-management), since the
+sequence batcher will manage the state associated with each sequence. The
+CONTROL_SEQUENCE_READY control is typically not needed because all inferences
+in the batch will always be ready for inference.
```
GuanLuo, line 545: I think "will always be ready for inference" can be elaborated a bit more, the wording doesn't have to be exact but something like "..., comparing to the case of direct sequence batcher where a batch slot has to reserved for the same sequence which may not be ready in the batch if the request hasn't arrived"

rmccorm4:
@Tabrizian I think this branch/change got mixed up with #5218

Tabrizian:
@rmccorm4 Yeah, I pushed to the wrong branch 😢 . I'll fix this.

PR #5143, opened 2022-12-07
Add mismatch key vs name gpu io test

kthui:
The current L0_tf_gpu_io test relies on well created TF savedmodels, which the I/O tensors key and name will always match, but this is not the case in the real world. The added model and test is derived from the same "qa_identity_model_repository/savedmodel_nobatch_zero_1_float32" but deliberately made the key and name to not match for testing the change.

Related PR: https://github.com/triton-inference-server/tensorflow_backend/pull/84


kthui:
Run this new test without updating TF backend: 
```
+-----------------------+---------+--------------------------------------------------------------------------------------------------------------------------+
| Model                 | Version | Status                                                                                                                   |
+-----------------------+---------+--------------------------------------------------------------------------------------------------------------------------+
| mismatch_key_name_def | 1       | READY                                                                                                                    |
| mismatch_key_name_gpu | 1       | UNAVAILABLE: Internal: Tensor INPUT0_key:0, specified in either feed_devices or fetch_devices was not found in the Graph |
+-----------------------+---------+--------------------------------------------------------------------------------------------------------------------------+
```

qa/L0_tf_gpu_io/test.sh
```diff
@@ -42,6 +42,7 @@ export CUDA_VISIBLE_DEVICES=0
 
 CLIENT=../clients/perf_client
 BACKENDS=${BACKENDS:="graphdef savedmodel"}
+MODEL_TYPES="$BACKENDS mismatch_key_name"
```
GuanLuo, line 45: Should just make it an additional case in this test so you don't need to have the additional type check and if-else below

qa/L0_tf_gpu_io/test.sh
```diff
@@ -123,11 +143,11 @@ for BACKEND in $BACKENDS; do
...
-    echo -e "\n${BACKEND}\n************"
-    cat ${BACKEND}.log
-    echo -e "\n${BACKEND} with GPU I/O\n************"
-    cat ${BACKEND}.gpu.log
+for MODEL_TYPE in $MODEL_TYPES; do
+    echo -e "\n${MODEL_TYPE}\n************"
+    cat ${MODEL_TYPE}.log
+    echo -e "\n${MODEL_TYPE} with GPU I/O\n************"
+    cat ${MODEL_TYPE}.gpu.log
 done
```
GuanLuo, line 151: Shouldn't always print the log, only print if test case failed

qa/L0_tf_gpu_io/mismatch_key_name/config.pbtxt
```diff
@@ -0,0 +1,44 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR 
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "mismatch_key_name"
```
GuanLuo, line 27: I would name it differently, IIUC, it is still an valid model, just that its signature def expose the I/O in names different from the tensor names in the TF graph.

I think the model created from `gen_tag_sigdef.py` has this characteristic so you should be able to re-use that for this test case. CC @jbkyang-nvi

qa/L0_tf_gpu_io/test.sh
kthui: Sure.

qa/L0_tf_gpu_io/mismatch_key_name/config.pbtxt
kthui: Yes. Updated.

kthui:
With the sig_tag0 model, when the test is run without the TF change, the server will fail to start: 
```
+----------+---------+---------------------------------------------------------------------------------------------------------------------+
| Model    | Version | Status                                                                                                              |
+----------+---------+---------------------------------------------------------------------------------------------------------------------+
| sig_tag0 | 1       | UNAVAILABLE: Internal: Tensor INPUT:0, specified in either feed_devices or fetch_devices was not found in the Graph |
+----------+---------+---------------------------------------------------------------------------------------------------------------------+
```

kthui:
CC @nv-kmcgill53, PR for TF gpu_io MakeCallable fix.

qa/L0_tf_gpu_io/test.sh
```diff
@@ -123,12 +125,21 @@ for BACKEND in $BACKENDS; do
...
+SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=1"
+SERVER_LOG="sig_tag0.serverlog"
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+kill $SERVER_PID
+wait $SERVER_PID
```
GuanLuo, line 142: Send request to test if the model is working.

GuanLuo: And check the response

kthui: Yes. Updated.

kthui:
Rebased.

GuanLuo (approved):

PR #5177, opened 2022-12-15
Allow User-Defined Batching Strategies

dyastremsky:
This PR aims to create tests for Triton's new custom batching feature. The test will use two example custom batching strategies provided in the backend examples folder. The test will start up the server with two instances of the model, then send 12 requests to it. It will check to see if the correct batching strategy was used.

The feature allows the user to pass in the custom batching feature via the model config. If not done so, it will check for a file called "batchstrategy.so" in the model version, model, and backend directories (in that order).

The test checks for the following cases:

1. Single batching is correctly loaded via the backend directory and succeeds.
2. Volume batching is correctly loaded via the backend directory and succeeds.
3. Volume batching is correctly loaded via the model directory and succeeds (with single batching ignored in the backend directory).
4. Volume batching is correctly loaded via the model version directory and succeeds (with single batching ignored in the backend directory).
5. Volume batching is correctly loaded via the model config and succeeds (with single batching ignored in the backend directory).

Core PR: https://github.com/triton-inference-server/core/pull/143/files
Backend PR: https://github.com/triton-inference-server/backend/pull/75


docs/user_guide/model_configuration.md
```diff
@@ -930,6 +930,14 @@ allow the queue to be configured so that individual requests are
 rejected or deferred if their time in the queue exceeds a specified
 timeout.
 
+#### Custom Batching
+
+You can set custom batching rules that work _in addition to_ the default behavior of the dynamic batcher. To do so, you would implement three functions in [tritonbackend.h](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonbackend.h) and create a shared library. These functions are `TRITONBACKEND_ModelBatchIncludeRequest`, `TRITONBACKEND_ModelBatchInitialize`, and `TRITONBACKEND_ModelBatchFinalize`.
```
nv-kmcgill53, line 935: NIT: I believe we have a max width of 100 chars per line with the exception of links. Please fix.

dyastremsky: Thanks! I didn't know what the hard rule was, and I don't believe our formatter works on .md files. Good to know. Fixed. :)

qa/L0_batch_custom/batch_custom_test.py
```diff
@@ -0,0 +1,272 @@
...
+from collections.abc import Iterable
+import tritonclient.grpc as grpcclient
+
+# By default, find tritonserver on "localhost", but can be overridden
+# with TRITONSERVER_IPADDR envvar
+_tritonserver_ipaddr = os.environ.get('TRITONSERVER_IPADDR', 'localhost')
+
+TEST_SYSTEM_SHARED_MEMORY = bool(
+    int(os.environ.get('TEST_SYSTEM_SHARED_MEMORY', 0)))
+TEST_CUDA_SHARED_MEMORY = bool(int(os.environ.get('TEST_CUDA_SHARED_MEMORY',
```
GuanLuo, line 48: Don't need the handling for shared memory / client variants, only keep the part to test the new functionality

nv-kmcgill53 (approved):

qa/L0_batch_custom/batch_custom_test.py
dyastremsky: Thank you for the feedback. Removed the variants.

GuanLuo (approved):

PR #5184, opened 2022-12-19
Fix np.object_ deprecations

Tabrizian:


krishung5 (dismissed):

qa/L0_backend_python/python_test.py
```diff
@@ -297,7 +297,7 @@ def test_unicode(self):
         model_name = "string"
         shape = [1]
 
-        for i in range(3):
+        for i in range(2):
```
tanmayv25, line 300: I am wondering why are we reducing the iteration counts?

krishung5: I think the change of the iteration counts here is to match the responses from the model. The model was sending three responses for each dytpes, and now we remove the `np.object_` data type so we reduce the iteration counts to 2. @Tabrizian Would it make sense to add some comments here for this?

Tabrizian: I have added a comment.

krishung5 (approved):

tanmayv25 (approved):

PR #5186, opened 2022-12-20
Add skip-pull flag to allow local build

jbkyang-nvi:
Triton has backends that are not shipped with the official containers such as
[FasterTransformer Backend](https://github.com/triton-inference-server/fastertransformer_backend).
This PR allows `compose.py` to use local images so users can further decrease the size of the images
 with these more customized backends.

Use case: 
`python3 compose.py --backend pytorch --backend fastertransformer --image full,triton_with_ft:22.11 --image gpu-min,nvcr.io/nvidia/tritonserver:22.11-py3-min --image min,nvcr.io/nvidia/tritonserver:22.11-py3-min --skip-pull`

### DIFF
--- a/compose.py
+++ b/compose.py
@@ -163,7 +163,7 @@ def get_container_version_if_not_specified():
     log('using container version {}'.format(FLAGS.container_version))
 
 
-def create_argmap(images):
+def create_argmap(images, skip_pull):
     # Extract information from upstream build and create map other functions can
     # use
     full_docker_image = images["full"]
@@ -177,13 +177,14 @@ def create_argmap(images):
     log("pulling container:{}".format(full_docker_image))
     p = subprocess.run(['docker', 'pull', full_docker_image])
     fail_if(
-        p.returncode != 0,
+        p.returncode != 0 and not skip_pull,
         'docker pull container {} failed, {}'.format(full_docker_image,
                                                      p.stderr))
     if enable_gpu:
         pm = subprocess.run(['docker', 'pull', min_docker_image])
         fail_if(
-            pm.returncode != 0, 'docker pull container {} failed, {}'.format(
+            pm.returncode != 0 and not skip_pull,
+            'docker pull container {} failed, {}'.format(
                 min_docker_image, pm.stderr))
         pm_path = subprocess.run(base_run_args + [
             '{{range $index, $value := .Config.Env}}{{$value}} {{end}}',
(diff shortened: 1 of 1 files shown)


compose.py
```diff
@@ -177,13 +177,14 @@ def create_argmap(images):
     log("pulling container:{}".format(full_docker_image))
     p = subprocess.run(['docker', 'pull', full_docker_image])
```
GuanLuo, line 178: Why not actually skip the pull?
```suggestion
    if not skip_pull:
        p = subprocess.run(['docker', 'pull', full_docker_image])
        ...
```

jbkyang-nvi: I feel like it looks cleaner like this since there's minimal logic block change. Note the gpu_min one will have 2 separate code blocks

GuanLuo: But then it does pull the image, if there is such an image on remote, wouldn't it overwrite the local image and cause confusion to the user?

jbkyang-nvi: you're right... changed 👍

compose.py
```diff
@@ -174,18 +174,19 @@ def create_argmap(images, skip_pull):
...
+        log("pulling container:{}".format(full_docker_image))
+        p = subprocess.run(['docker', 'pull', full_docker_image])
         fail_if(
-            pm.returncode != 0 and not skip_pull,
-            'docker pull container {} failed, {}'.format(
-                min_docker_image, pm.stderr))
+            p.returncode != 0, 'docker pull container {} failed, {}'.format(
+                full_docker_image, p.stderr))
+    if enable_gpu:
+        if (not skip_pull):
```
Tabrizian, line 184: nit: remove parentheses

compose.py
```diff
@@ -174,18 +174,19 @@ def create_argmap(images, skip_pull):
...
     # first pull docker images
-    log("pulling container:{}".format(full_docker_image))
-    p = subprocess.run(['docker', 'pull', full_docker_image])
-    fail_if(
-        p.returncode != 0 and not skip_pull,
-        'docker pull container {} failed, {}'.format(full_docker_image,
-                                                     p.stderr))
-    if enable_gpu:
-        pm = subprocess.run(['docker', 'pull', min_docker_image])
+    if (not skip_pull):
```
Tabrizian, line 177: nit: remove parentheses

GuanLuo (dismissed):

Tabrizian (approved):

PR #5204, opened 2022-12-28
Document custom batching cache functions

dyastremsky:
This PR builds on #5177. For custom batching where the user defines additional constraints for the dynamic batcher, this PR allows the user to call two new optional functions that initialize and finalize once. This allows them to create a cached object that can be read for every batch without needing to be re-initialized.

An example use case is where the user needs to read a value from the model config for batching, so they would only want to read it from the config once and save it to use with their custom batch-handling logic. See volume_batching.cc in this PR as an example.

The server PR documents these two new functions.

Backend: https://github.com/triton-inference-server/backend/pull/76
Core: https://github.com/triton-inference-server/core/pull/149

### DIFF
--- a/docs/user_guide/model_configuration.md
+++ b/docs/user_guide/model_configuration.md
@@ -934,15 +934,15 @@ timeout.
 
 You can set custom batching rules that work _in addition to_ the behavior of the dynamic batcher.
 To do so, you would implement five functions in [tritonbackend.h](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonbackend.h) 
-and create a shared library. These functions are:
+and create a shared library. These functions are described below.
 
 | Function | Description| 
 | :--          |   :--           |
 | TRITONBACKEND_ModelBatchIncludeRequest | Determines whether a request should be included in the current batch |
 | TRITONBACKEND_ModelBatchInitialize | Initializes record-keeping data structure for a new batch |
-| TRITONBACKEND_ModelBatchFinalize | Frees memory of record-keeping data structure |
+| TRITONBACKEND_ModelBatchFinalize | Frees memory of record-keeping data structure after batch is formed |
 | TRITONBACKEND_ModelBatcherInitialize | Initializes read-only data structure for use with all batches |
-| TRITONBACKEND_ModelBatcherFinalize | Frees read-only batcher data structure |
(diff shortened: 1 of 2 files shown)


docs/user_guide/model_configuration.md
```diff
@@ -932,10 +932,13 @@ timeout.
...
-batcher. To do so, you would implement three functions in [tritonbackend.h](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonbackend.h)
-and create a shared library. These functions are `TRITONBACKEND_ModelBatchIncludeRequest`,
-`TRITONBACKEND_ModelBatchInitialize`, and `TRITONBACKEND_ModelBatchFinalize`.
+You can set custom batching rules that work _in addition to_ the default behavior of the dynamic 
+batcher. To do so, you would implement three functions in [tritonbackend.h](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonbackend.h) 
+and create a shared library. These functions are `TRITONBACKEND_ModelBatchIncludeRequest`, 
+`TRITONBACKEND_ModelBatchInitialize`, and `TRITONBACKEND_ModelBatchFinalize`. 
+You can also implement  two optional functions in your shared library: 
+`TRITONBACKEND_ModelBatchCacheInitialize` and `TRITONBACKEND_ModelBatchCacheFinalize` to store 
+read-only information that is initialized once and used across batches.
```
nv-kmcgill53, line 941: I would suggest making this into a table. I've included a "description" column but I think this is optional.
```suggestion
read-only information that is initialized once and used across batches.

| Function | Required | Description| 
| :--          |   :--         |   :--           |
| TRITONBACKEND_ModelBatchIncludeRequest | Yes | ... |
| TRITONBACKEND_ModelBatchInitialize | Yes | ... |
| TRITONBACKEND_ModelBatchFinalize | Yes | ... |
| TRITONBACKEND_ModelBatchCacheInitialize | No | ... |
| TRITONBACKEND_ModelBatchCacheFinalize | No | ... |
```

nv-kmcgill53 (changes requested):

docs/user_guide/model_configuration.md
dyastremsky: Great suggestion! Updated.

dyastremsky:
To try to get this in quickly, I'm going to merge this into the main PR branch. All feedback should have been addressed. :) If anyone has additional feedback, please comment it in the main dyas-custom-batching PR.

PR #5226, opened 2023-01-06
Increase PA test stability threshold

matthewkotila:
We always get errors saying test failed to stabilize, but it's not necessary to stabilize to verify what the test wants to verify, so, if we increase stability threshold to 100, we skip the stabilization check, which was unnecessary anyway.


debermudez:
seems reasonable to me. defering approval to iman

Tabrizian (approved):

PR #5244, opened 2023-01-12
Mchornyi r23.01 libcufft

mc-nv:


build.py
```diff
@@ -769,7 +769,7 @@ def install_dcgm_libraries(dcgm_version, target_machine):
 RUN curl -o /tmp/cuda-keyring.deb \
     https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/sbsa/cuda-keyring_1.0-1_all.deb \
     && apt install /tmp/cuda-keyring.deb && rm /tmp/cuda-keyring.deb && \
-    apt-get update && apt-get install -y datacenter-gpu-manager=1:{}
```
Tabrizian, line 772: Is this the correct place to add this library? It is under `install_dcgm_libraries`.

mc-nv: It's a good question.
I don't know for how long we will have this solution "ON". 
But with understanding that this step has place for "x86" and "sbsa" build and that the previous build (not tests) was successful without `libcufft-11-8` installation I assuming to reuse this instruction. 
I can modify statement  in order to give us more clarity on it:
```
apt-get update && apt-get install -y datacenter-gpu-manager=1:{} && \
apt-get install -y  libcufft-11-8 # only to cover 23.01 release
```

Tabrizian: How about just adding it in? https://github.com/triton-inference-server/server/blob/f63c5783c0bf1f71a8b6a83ae91f85320a53064a/build.py#L891

tanmayv25: I agree with Iman's reasoning. Adding the libcufft installation alongwith dcgm appears to be confusing.

mc-nv: Sure, I have updated the `build.py` with suggested change.

Tabrizian (approved):

PR #5257, opened 2023-01-16
StartupProbe for K8s-onprem added and documented

okdimok:
In Kubernetes 1.19 `startupProbe` was introduced in Beta to solve the following issue:

> Slow starting containers are difficult to address with the current status of health probes: they are either killed before being up, or could be left deadlocked during a very long time before being killed.
>This proposal adds a new probe called `startupProbe` that holds off all the other probes until the pod has finished its startup. In the case of a slow-starting pod, it could poll on a relatively short period with a high `failureThreshold`. Once it is satisfied, the other probes can start.

Since a Triton container having multiple heavy models can be a good example of a slow-starting container, I've added the `startupProbe` to the helm deployment template with some reasonable values and documented the probes in the proper README.

For the details on the probes, see [the K8s documentation on probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) and the [feature page of the startup probe](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/950-liveness-probe-holdoff/README.md).

### DIFF
--- a/deploy/k8s-onprem/README.md
+++ b/deploy/k8s-onprem/README.md
@@ -236,10 +236,11 @@ $ helm install example -f config.yaml .
 
 ## Probe Configuration
 
-In `templates/deployment.yaml` is configurations for `livenessProbe`, `readinessProbe` and `startupProbe` for the Triton server container. 
-By default, Triton loads all the models before starting the HTTP server to respond to the probes. The process can take several minutes, depending on the models sizes. 
-If it is not completed in `startupProbe.failureThreshold * startupProbe.periodSeconds` seconds then Kubernetes considers this as a pod failure and restarts it, ending up with an infinite loop of restarting pods, so make sure to sufficiently set these values for your use case.
-The liveliness and readiness probes are being sent only after the first success of a startup probe. 
+In `templates/deployment.yaml` is configurations for `livenessProbe`, `readinessProbe` and `startupProbe` for the Triton server container.
+By default, Triton loads all the models before starting the HTTP server to respond to the probes. The process can take several minutes, depending on the models sizes.
+If it is not completed in `startupProbe.failureThreshold * startupProbe.periodSeconds` seconds then Kubernetes considers this as a pod failure and restarts it,
+ending up with an infinite loop of restarting pods, so make sure to sufficiently set these values for your use case.
+The liveliness and readiness probes are being sent only after the first success of a startup probe.
 
 For more details, see the [Kubernetes probe documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) and the [feature page of the startup probe](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/950-liveness-probe-holdoff/README.md).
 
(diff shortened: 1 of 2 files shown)


deploy/k8s-onprem/templates/deployment.yaml
```diff
@@ -79,12 +79,25 @@ spec:
             - containerPort: 8002
               name: metrics
           livenessProbe:
+            initialDelaySeconds: 15
+            failureThreshold: 3
+            periodSeconds: 10
```
okdimok, line 84: These are already the defaults. Adding them verbose, so that people could customize the values easier.

deploy/k8s-onprem/README.md
```diff
@@ -234,6 +234,8 @@ EOF
 $ helm install example -f config.yaml .
 ```
 
+In the configuration itself, pay attention to liveliness, readiness and startup probes of the Trion container itself, configured in the `templates/deployment.yaml`. By default, Triton loads all the models, before starting the HTTP server to respond to the probes. The process can take several minutes, depending on the models sizes. If it is not completed in  `startupProbe.failureThreshold * startupProbe.periodSeconds` seconds then Kubernetes considers this as a pod failure, and restarts it, ending up with an infinite loop of restarting pods. So, make sure to sufficiently increase these values. The liveliness and readiness probes are being sent ony after the first success of a startup probe. For details, see [the K8s documentation on probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) and the [feature page of the startup probe](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/950-liveness-probe-holdoff/README.md).
```
rmccorm4, line 237:
```suggestion
## Probe Configuration

In `templates/deployment.yaml` is configurations for `livenessProbe`, `readinessProbe` and `startupProbe` for the Triton server container. 
By default, Triton loads all the models before starting the HTTP server to respond to the probes. The process can take several minutes, depending on the models sizes. 
If it is not completed in `startupProbe.failureThreshold * startupProbe.periodSeconds` seconds then Kubernetes considers this as a pod failure and restarts it, ending up with an infinite loop of restarting pods, so make sure to sufficiently set these values for your use case.
The liveliness and readiness probes are being sent only after the first success of a startup probe. 

For more details, see the [Kubernetes probe documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) and the [feature page of the startup probe](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/950-liveness-probe-holdoff/README.md).
```

rmccorm4:
Hi @okdimok, thanks for the contribution and well documented notes along with it!

Can you share a snippet showing a successful deployment with the above config? 

Otherwise it looks good to me. CC @nealvaidya @mengdong for your thoughts as well.

nealvaidya:
Thanks for this update! It looks good. Did you consider including options in the `values.yaml` file for these probe timings? It would be easier for the user to change the values if they were there so we may want to add them, but I don't feel very strongly about it

mengdong:
looks good!

dyastremsky (approved):
Fantastic work here! A customer was able to start with this set-up [here](https://github.com/triton-inference-server/server/issues/6469), which can serve as an example of a successful deployment. If we would like to move the values to `values.yaml`, we can do so in a follow-up ticket.

dyastremsky (approved):

PR #5261, opened 2023-01-18
Update cache metrics doc for implementation-specific metrics

rmccorm4:
Remove global poll cache metrics from doc, add description of cache-implementation-specific metrics

Corresponding core PR to remove polling cache metrics: https://github.com/triton-inference-server/core/pull/161


docs/user_guide/metrics.md
```diff
@@ -133,36 +133,40 @@ If building Triton locally, the `TRITON_ENABLE_METRICS_CPU` CMake build flag can
...
-On cache misses, "Cache Miss Lookup Time" indicates the time spent looking up
-the request hash and "Cache Miss Insertion Time" indicates the time spent
-inserting the computed output tensor data into the cache. Otherwise, "Compute
-Input Time" /  "Compute Time" / "Compute Output Time" will be recorded as usual.
+On cache misses, "Cache Miss Time" indicates the time spent looking up
+the request hash and inserting the computed output tensor data into the cache. 
+Otherwise, "Compute Input Time" /  "Compute Time" / "Compute Output Time" will
+be recorded as usual.
 
 |Category      |Metric          |Metric Name |Description                            |Granularity|Frequency    |
```
dyastremsky, line 164: It looks like you no longer use the category column, so you may want to get rid of it.

docs/user_guide/metrics.md
```diff
@@ -133,36 +133,40 @@ If building Triton locally, the `TRITON_ENABLE_METRICS_CPU` CMake build flag can
...
-Compute latency metrics in the [Inference Request Metrics table](#inference-request-metrics) above are calculated for the
-time spent in model inference backends. If the response cache is enabled for a
-given model (see [Response Cache](response_cache.md)
+Cache metrics can be reported in two ways:
+
+1. A base set of cache metrics will be reported
+by Triton directly, such as the cache hit/miss counts and durations described
+below.
+
+2. As of 23.02, additional cache metrics may be reported depending on the
```
dyastremsky, line 142: Do you need the 23.02?

rmccorm4: I wanted to make it clear to users when the shift from old to new behavior happened. If you think it reads poorly, I can remove it - but I think it's something I've seen in some github issues a few times where users were confused when exactly certain features were added or removed

docs/user_guide/metrics.md
rmccorm4: Good catch, fixed the column

docs/user_guide/metrics.md
```diff
@@ -133,36 +133,40 @@ If building Triton locally, the `TRITON_ENABLE_METRICS_CPU` CMake build flag can
...
-|Utilization   |Total Cache Utilization |`nv_cache_util` |Total Response Cache utilization rate (0.0 - 1.0) |Server-wide |Per interval |
-|Count         |Total Cache Entry Count |`nv_cache_num_entries` |Total number of responses stored in response cache across all models |Server-wide |Per interval |
-|              |Total Cache Lookup Count |`nv_cache_num_lookups` |Total number of response cache lookups done by Triton across all models |Server-wide |Per interval |
-|              |Total Cache Hit Count |`nv_cache_num_hits` |Total number of response cache hits across all models |Server-wide |Per interval |
-|              |Total Cache Miss Count |`nv_cache_num_misses` |Total number of response cache misses across all models |Server-wide |Per interval |
-|              |Total Cache Eviction Count |`nv_cache_num_evictions` |Total number of response cache evictions across all models |Server-wide |Per interval |
-|              |Cache Hit Count |`nv_cache_num_hits_per_model` |Number of response cache hits per model |Per model |Per request |
+|Count         |Cache Hit Count |`nv_cache_num_hits_per_model` |Number of response cache hits per model |Per model |Per request |
```
Tabrizian, line 166: Does it mean that we will no longer report "Cache Utilization"?

docs/user_guide/metrics.md
```diff
@@ -133,36 +133,40 @@ If building Triton locally, the `TRITON_ENABLE_METRICS_CPU` CMake build flag can
...
-given model (see [Response Cache](response_cache.md)
+Cache metrics can be reported in two ways:
+
+1. A base set of cache metrics will be reported
+by Triton directly, such as the cache hit/miss counts and durations described
+below.
+
+2. As of 23.02, additional cache metrics may be reported depending on the 
+[cache implementation](response_cache.md#cache-implementations) 
+being used through Triton's [Metrics API](#custom-metrics).
```
Tabrizian, line 144: So, if they use custom metrics with the same metric name and type that we use in the local cache, all the tooling that uses these values (e.g., perf analyzer) will report correct numbers?

docs/user_guide/metrics.md
Tabrizian: I think it is good practice to add the version number with new features. I have seen similar pattern in Python documentation. https://docs.python.org/3/library/functions.html#aiter

dyastremsky: Interesting, thank you both for sharing. Maybe we should make this more ubiqutious then.

dyastremsky (approved):

docs/user_guide/metrics.md
rmccorm4: Cache Utilization will be moved to `local` cache implementation and added via C API

docs/user_guide/metrics.md
rmccorm4: The minimal set that remains in Triton-core and in this doc (per-model cache_hit_count/duration, cache_miss_count/duration) is used by perf analyzer.

As with any existing metrics, it's possible for a user to update their values with the metrics API - though they should not do this. I mention a couple concerns / potential solutions around this in [DLIS-4072](https://jirasw.nvidia.com/browse/DLIS-4072), but for now it's essentially up to the user to not pick duplicate names.

Dockerfile.QA
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 1: Accidentally bumped copyright date without any changes in other [PR](https://github.com/triton-inference-server/server/pull/5119/files#diff-0c3eb9bfef58682e83232ffa9140d7cab2bbcc40488da3dc0f6dabb474296f2f), so fixing here

Tabrizian (approved):

PR #5264, opened 2023-01-19
new instruction for GKE marketplace app

mengdong:
Provide the fix since GKE depreciate the istio on GKE so we manually install. 

Also change the github repo to a static GCS public bucket with example models before we fix the bert pipeline which generate TRT engine for different version of Triton.


kthui (approved):

kthui:
Thanks for the update @mengdong.

CC @GuanLuo @rmccorm4 make sure the instruction update looks good to you too.

chuckcoder:
Could you add documentation about how to create service account used in `--service-account=gke-test@k80-exploration.iam.gserviceaccount.com` option? Specifically what are the roles and permissions necessary for gke-test@k80-exploration.iam.gserviceaccount.com`

PR #5280, opened 2023-01-26
Add cache to build pipeline, update cache unit test

rmccorm4:


compose.py
```diff
@@ -130,6 +130,18 @@ def add_requested_repoagents(ddir, dockerfile_name, repoagents):
     with open(os.path.join(ddir, dockerfile_name), "a") as dfile:
         dfile.write(df)
 
+def add_requested_caches(ddir, dockerfile_name, caches):
```
Tabrizian, line 133: Do we test cache/repoagent functionality of `compose.py` script anywhere?

rmccorm4: It looks like it is included in all the *-compose tests via the env vars `REPOAGENT` and `CACHE`. I will manually run them and see the behavior.

rmccorm4: With cache added: It passes the compose tests, shows up in the resultant container, and can be loaded:
```
root@rmccormick-dt:/opt/tritonserver# ls caches/local/libtritoncache_local.so 
caches/local/libtritoncache_local.so

root@rmccormick-dt:/opt/tritonserver# tritonserver --model-repository models --cache-config local,size=1024                
I0127 22:47:00.481844 205 cache_manager.cc:55] Creating TritonCache with name: 'local', libpath: '/opt/tritonserver/caches/local/libtritoncache_local.so', cache_config: '{"size":"1024"}'
```

Tabrizian (approved):

PR #5281, opened 2023-01-26
Improve fastertransformer build in build.py

jbkyang-nvi:
ref: https://github.com/triton-inference-server/fastertransformer_backend/blob/main/docker/create_dockerfile_and_build.py#L117
and 
https://github.com/triton-inference-server/fastertransformer_backend/blob/main/docker/Dockerfile

### DIFF
--- a/build.py
+++ b/build.py
@@ -1093,8 +1093,9 @@ ENV TCMALLOC_RELEASE_RATE 200
 '''.format(gpu_enabled=gpu_enabled, backend_dependencies=backend_dependencies)
 
     if ('fastertransformer' in backends):
+        be = "fastertransformer"
         import importlib.util, requests
-        url = 'https://raw.githubusercontent.com/triton-inference-server/fastertransformer_backend/main/docker/create_dockerfile_and_build.py'
+        url = 'https://raw.githubusercontent.com/triton-inference-server/fastertransformer_backend/{}/docker/create_dockerfile_and_build.py'.format(backends[be])
         response = requests.get(url)
         spec = importlib.util.spec_from_loader('fastertransformer_buildscript',
                                                loader=None,
--- a/build.py
+++ b/build.py
@@ -765,7 +765,9 @@ def fastertransformer_cmake_args():
     print(
         "Warning: Fastertransformer backend is not officially supported part of Triton."
     )
-    return [cmake_backend_arg(be, 'CMAKE_EXPORT_COMPILE_COMMANDS', None, 1)]
+    cargs = [cmake_backend_arg('fastertransformer', 'CMAKE_EXPORT_COMPILE_COMMANDS', None, 1)]
+    cargs.append(cmake_backend_arg('fastertransformer', 'ENABLE_FP8', None, 'OFF'))
+    return cargs
 
 
 def install_dcgm_libraries(dcgm_version, target_machine):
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -755,6 +761,15 @@ def armnn_tflite_cmake_args():
     ]
 
 
+def fastertransformer_cmake_args():
+    print(
+        "Warning: Fastertransformer backend is not officially supported part of Triton."
```
dyastremsky, line 766: Nit: Wording. "not an officially supported part" or "not officially supported." I also believe FasterTransformer uses CamelCase for their capitalization (i.e. "FasterTransformer").

build.py
```diff
@@ -1079,6 +1099,38 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
...
+        run_subprocess(op, msg)
+
+        import create_dockerfile_and_build
+        df += create_dockerfile_and_build.create_postbuild(
+            is_multistage_build=False)
+
+        # clean up
+        op = ['rm', '-r', 'create_dockerfile_and_build.py']
+        msg = 'Failed to remove directory {}'.format(dir_name)
+        run_subprocess(op, msg)
```
GuanLuo, line 1132: So the fastertransformer backend is built at the last docker stage? Based on the comment it should not contain the backends:
```
Production stage: Create container with just inference server executable
```
Why can't fastertransformer follow our guideline to encapsulate the build in CMake so it can utilize the same `backend_build` step in `build.py`?

jbkyang-nvi: @GuanLuo we do build it with the same `backend_build` with `def fastertransformer_cmake_args():`. I'm simply running the few commands required in the post-process phase:
```
     df = '''
ENV WORKSPACE /workspace
WORKDIR /workspace       
'''
    df += '''
ENV NCCL_LAUNCH_MODE=GROUP
RUN sed -i 's/#X11UseLocalhost yes/X11UseLocalhost no/g' /etc/ssh/sshd_config \\
    && mkdir /var/run/sshd -p
    '''
```

jbkyang-nvi: https://github.com/triton-inference-server/fastertransformer_backend/blob/main/docker/create_dockerfile_and_build.py

tanmayv25: I was wondering if we can prevent this additional clone process. How about copying out create_dockerfile_and_build.py file from the build step of FasterTransformer backend?

GuanLuo: Thanks for the clarification. Given the intention, what if you directly fetch the file content (HTTP GET versioned/path/to/create_dockerfile_and_build.py) and [construct Python module in place](https://stackoverflow.com/questions/5362771/how-to-load-a-module-from-code-in-a-string)? Then there is no exchange outside the script process

build.py
```diff
@@ -111,6 +111,12 @@ def fail_if(p, msg):
         sys.exit(1)
 
 
+def run_subprocess(command_list, msg):
```
tanmayv25, line 114: This can be removed?

build.py
```diff
@@ -755,6 +761,17 @@ def armnn_tflite_cmake_args():
     ]
 
 
+def fastertransformer_cmake_args():
+    print("Warning: FasterTransformer backend is not officially supported.")
+    cargs = [
+        cmake_backend_arg('fastertransformer', 'CMAKE_EXPORT_COMPILE_COMMANDS',
+                          None, 1)
+    ]
+    cargs.append(
```
dyastremsky, line 770: It might be cleaner to initialize with this arg rather than append it after. Example [here](https://github.com/triton-inference-server/server/blob/0e677f6cc6d2ffe6fa4f21ce72b14da698436726/build.py#L584-L589).

build.py
```diff
@@ -1079,6 +1092,18 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
 ENV TCMALLOC_RELEASE_RATE 200
 '''.format(gpu_enabled=gpu_enabled, backend_dependencies=backend_dependencies)
 
+    if ('fastertransformer' in backends):
+        import importlib.util, requests
+        url = 'https://raw.githubusercontent.com/triton-inference-server/fastertransformer_backend/main/docker/create_dockerfile_and_build.py'
```
GuanLuo, line 1097: change `main` to the branch specified when enabling FT build

dyastremsky (approved):

tanmayv25 (approved):

PR #5282, opened 2023-01-27
Refactor L0_tf_gpu_io to remove perf_client dependency

dyastremsky:
Perf_client was being used to generate requests in l0_tf_gpu_io. This PR decouples the test from perf_client. Instead one request is sent to Triton server then logs are examined, speeding up this test and removing for the dependency on perf_client.

### DIFF
--- a/qa/L0_tf_gpu_io/test.sh
+++ b/qa/L0_tf_gpu_io/test.sh
@@ -42,7 +42,6 @@ export CUDA_VISIBLE_DEVICES=0
 
 TF_TEST=tf_gpu_io_test.py
 BACKENDS=${BACKENDS:="graphdef savedmodel"}
-TENSOR_SIZE=16384
 
 DATADIR=/data/inferenceserver/${REPO_VERSION}
 
--- a/qa/L0_tf_gpu_io/test.sh
+++ b/qa/L0_tf_gpu_io/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2019-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019-2023, NVIDIA CORPORATION. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -40,7 +40,7 @@ fi
 
 export CUDA_VISIBLE_DEVICES=0
 
-CLIENT=../clients/perf_client
+TF_TEST=tf_gpu_io_test.py
 BACKENDS=${BACKENDS:="graphdef savedmodel"}
 TENSOR_SIZE=16384
 
@@ -78,8 +78,7 @@ for BACKEND in $BACKENDS; do
 
     set +e
 
-    $CLIENT -m${MODEL_NAME}_def --shape INPUT0:${TENSOR_SIZE} \
-                >> ${BACKEND}.sanity.log 2>&1
+    python $TF_TEST TfGpuIoTest.test_${MODEL_NAME}_def >> ${BACKEND}.sanity.log 2>&1
     if (( $? != 0 )); then
         cat ${BACKEND}.sanity.log
         RET=1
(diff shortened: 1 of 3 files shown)


qa/L0_tf_gpu_io/test.sh
```diff
@@ -104,21 +102,6 @@ for BACKEND in $BACKENDS; do
...
-        RET=1
-    fi
-
-    $CLIENT -m${MODEL_NAME}_gpu --shape INPUT0:${TENSOR_SIZE} \
-            >> ${BACKEND}.gpu.log 2>&1
-    if (( $? != 0 )); then
-        cat ${BACKEND}.gpu.log
-        RET=1
-    fi
-
```
rmccorm4, line 121: Are these not needed?

qa/L0_tf_gpu_io/test.sh
```diff
@@ -40,7 +40,7 @@ fi
 
 export CUDA_VISIBLE_DEVICES=0
 
-CLIENT=../clients/perf_client
+TF_TEST=tf_gpu_io_test.py
 BACKENDS=${BACKENDS:="graphdef savedmodel"}
 TENSOR_SIZE=16384
```
rmccorm4, line 45: This should either be removed or read in as an env var in the python test

qa/L0_tf_gpu_io/tf_gpu_io_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def test_sig_tag0(self):
+        self._test_helper("sig_tag0", False, "INPUT", "OUTPUT", [
+            16,
+        ])
+
+    def test_graphdef_zero_1_float32_def(self):
+        self._test_helper("graphdef_zero_1_float32_def", True, "INPUT0",
+                          "OUTPUT0", [
+                              16384,
```
rmccorm4, line 70: Set `16384` as some global variable (or read the env var from test.sh) and use that everywhere instead of a magic number

dyastremsky: Thanks, Ryan! Added a global variable.

qa/L0_tf_gpu_io/test.sh
dyastremsky: Good catch. Removed.

qa/L0_tf_gpu_io/test.sh
dyastremsky: I don't believe so. We don't do anything with the latency results.

This seems to run perf_analyzer a second time to get the results and save them as a log. However, this seems to be redundant given that it's already run once, so I'd rather remove this second run to preserve CI resources (and the whole point of this ticket is to remove the need for perf_analyzer).

rmccorm4 (approved):

qa/L0_tf_gpu_io/tf_gpu_io_test.py
```diff
@@ -0,0 +1,95 @@
...
+            inputs.append(httpclient.InferInput(input_name, shape, "FP32"))
+            if batching_enabled:
+                input_data = np.ones(shape=shape, dtype=np.float32)
+            else:
+                input_data = np.ones(shape=shape, dtype=np.float32)
+            inputs[0].set_data_from_numpy(input_data, binary_data=True)
+
+            outputs.append(
+                httpclient.InferRequestedOutput(output_name, binary_data=True))
+            results = triton_client.infer(model_name, inputs, outputs=outputs)
```
tanmayv25, line 60: Can you enhance this test to validate the output results from the inference?

dyastremsky: Sure, done!

qa/common/infer_util.py
```diff
@@ -896,12 +903,16 @@ def infer_zero(tester,
         platform = metadata.platform
 
     for io_num in range(io_cnt):
-        if platform == "pytorch_libtorch":
-            input_name = "INPUT__{}".format(io_num)
-            output_name = "OUTPUT__{}".format(io_num)
+        if override_input_names:
+            input_name = override_input_names[io_num]
+            output_name = override_output_names[io_num]
```
tanmayv25, line 908: Should this be under condition override_output_names ?

dyastremsky: Great point. Fixed.

tanmayv25 (approved):

rmccorm4 (approved):

PR #5298, opened 2023-02-02
Improvements to SM server and entrypoint

nskool:
1. Enable verbose logging by default, customers can still disable it by specifying `SAGEMAKER_TRITON_LOG_VERBOSE`:`false`.
2. Added more strings for OOM checker to detect, in order to return 507
3. Changed /ping health check to rely on 'live', instead of 'ready' to handle SageMaker corner cases where it expects model server to react with '200' for /ping even if model load has failed separately for some other models, whereas has succeeded for others and are ready to be served. (difference of 'live'/'ready' is found in kserve description - https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/rest_predict_v2.yaml#L10-L26)

### DIFF
--- a/docker/sagemaker/serve
+++ b/docker/sagemaker/serve
@@ -52,8 +52,11 @@ fi
 if [ -n "$SAGEMAKER_TRITON_THREAD_COUNT" ]; then
     SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --sagemaker-thread-count=${SAGEMAKER_TRITON_THREAD_COUNT}"
 fi
+# Enable verbose logging by default. If env variable is specified, use value from env variable
 if [ -n "$SAGEMAKER_TRITON_LOG_VERBOSE" ]; then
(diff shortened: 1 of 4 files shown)


GuanLuo (approved):

nskool:
Thanks @GuanLuo for taking a look, can we run the CI to check if the tests pass?

GuanLuo:
@nskool Thanks for the PR, I just ran the CI and this [case](https://github.com/triton-inference-server/server/blob/main/qa/L0_sagemaker/test.sh#L341-L367) in L0_sagemaker failed, I think this is expected based on your change for 3., can you update the test case to address the new behavior?

nskool:
@GuanLuo Thank you for running the CI, I've fixed the test. Could you please run again?

GuanLuo:
Test passed, merging

GuanLuo (approved):

PR #5314, opened 2023-02-04
Update variable reference name

mc-nv:


qa/common/gen_qa_custom_ops
```diff
@@ -42,7 +42,7 @@ NVIDIA_UPSTREAM_VERSION=${NVIDIA_UPSTREAM_VERSION:=$TRITON_VERSION}
 TENSORFLOW_IMAGE=${TENSORFLOW_IMAGE:=nvcr.io/nvidia/tensorflow:$NVIDIA_UPSTREAM_VERSION-tf2-py3}
 PYTORCH_IMAGE=${PYTORCH_IMAGE:=nvcr.io/nvidia/pytorch:$NVIDIA_UPSTREAM_VERSION-py3}
 
-CUDA_DEVICE=${RUNNER_ID:=0}
+CUDA_DEVICE=${NV_GPU:=0}
```
Tabrizian, line 45: Is the variable name, renamed in the upstream?

Tabrizian (approved):

qa/common/gen_qa_custom_ops
mc-nv: it's CI  internal only and previously always been `0` but now it can vary, and no longer accurate to use `RUNNER_ID`.

GuanLuo: Is this why sometimes the model is generated against GPU with different compute capability?

mc-nv: It can be a reason if you have multiple devices on the board and passing wrong ID of it

Issue #5309, opened 2023-02-03
Onnx runtime build error for 23.01


nskool:
**Description**
1. I'm trying to build the Triton container as follows, where `RELEASE_TAG=r23.01`
```
./build.py --enable-logging --enable-stats --enable-tracing --enable-metrics --enable-gpu-metrics --enable-gpu --no-container-interactive --endpoint=http --endpoint=grpc --endpoint=sagemaker --repo-tag=common:$RELEASE_TAG --repo-tag=core:$RELEASE_TAG --repo-tag=backend:$RELEASE_TAG --repo-tag=thirdparty:$RELEASE_TAG --backend=ensemble:$RELEASE_TAG --backend=tensorrt:$RELEASE_TAG --backend=identity:$RELEASE_TAG --backend=repeat:$RELEASE_TAG --backend=square:$RELEASE_TAG --backend=onnxruntime:$RELEASE_TAG --backend=pytorch:$RELEASE_TAG --backend=tensorflow1:$RELEASE_TAG --backend=tensorflow2:$RELEASE_TAG --backend=python:$RELEASE_TAG --backend=dali:$RELEASE_TAG --backend=fil:$RELEASE_TAG --backend=fastertransformer:main --repoagent=checksum:$RELEASE_TAG
```
Error detected:
```
[ 70%] Building CXX object CMakeFiles/onnxruntime_providers.dir/workspace/onnxruntime/onnxruntime/core/providers/cpu/math/cumsum.cc.o
/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/type_traits(566): error: "cuda" is ambiguous
/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/type_traits(566): error: too many arguments for alias template "cuda::std::__4::_BoolConstant"
/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/type_traits(568): error: expected a ";"
/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/type_traits(575): error: "cuda" is ambiguous
/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/type_traits(575): error: too many arguments for alias template "cuda::std::__4::_BoolConstant"
/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/type_traits(577): error: expected a ";"
[...]

PR #5340, opened 2023-02-09
Adding default base container for ORT backend build

tanmayv25:
This will use 22.12 container for ORT backend build by default so the user doesn't have to explicitly provide it as an option in build.py.

See the discussion in this issue here: https://github.com/triton-inference-server/server/issues/5309

### DIFF
--- a/build.py
+++ b/build.py
@@ -631,16 +631,25 @@ def onnxruntime_cmake_args(images, library_paths):
                                  'TRITON_ENABLE_ONNXRUNTIME_OPENVINO', False)
         ]
     else:
+        ort_image = None
+        if ('base' in images) and ('ort-base' in images):
+            ort_image = images['ort-base']
+        elif ('base' in images):
+            # If ort-base is not specified, we default to
+            # tensorrt:22.12-py3 base container for ort backend
+            # build with cuda 11.8 library as ORT does
+            # not support cuda 12 in 23.01 containers.
+            ort_image = 'nvcr.io/nvidia/tensorrt:22.12-py3'
         if target_platform() == 'windows':
-            if 'ort-base' in images:
+            if ort_image is not None:
                 cargs.append(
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -631,16 +631,25 @@ def onnxruntime_cmake_args(images, library_paths):
...
     else:
+        ort_image = None
+        if ('base' in images) and ('ort-base' in images):
+            ort_image = images['ort-base']
+        elif ('base' in images):
+            # If ort-base is not specified, we default to
+            # tensorrt:22.12-py3 base container for ort backend
+            # build with cuda 11.8 library as ORT does
+            # not support cuda 12 in 23.01 containers.
+            ort_image = 'nvcr.io/nvidia/tensorrt:22.12-py3'
```
Tabrizian, line 642: It is strange that we need tensorrt:22.12 when we built with the 22.12 min container in our CI. Do you happen to have the error log that you saw when the min container was used?

tanmayv25: I think I have lost it. I am trying again with the min container.

tanmayv25: tritonserver:22.12-py3-min is working. I might not have set the containers properly before.

build.py
```diff
@@ -67,7 +67,7 @@
 # incorrectly load the other version of the openvino libraries.
 #
 TRITON_VERSION_MAP = {
-    '2.30.0': (
+    '2.30.0dev': (
```
Tabrizian, line 70: Is dev required? I believe dev is only required on `main` branch.

tanmayv25: I will remove this commit once CI passes and before merging the PR. https://github.com/triton-inference-server/server/pull/5340/commits/641096b761f7cbe4689aa08adb5dd4f69ca7657b
The 23.01-devel gitlab CI expects dev.

Tabrizian (approved):

Tabrizian (approved):

PR #5347, opened 2023-02-13
Add missing flag in quickstart guide

dyastremsky:
Add networking (--net=host) flag to Docker server container to allow the sdk container to communicate with it.

### DIFF
--- a/docs/getting_started/quickstart.md
+++ b/docs/getting_started/quickstart.md
@@ -1,5 +1,5 @@
 <!--
-# Copyright (c) 2018-2020, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2018-2023, NVIDIA CORPORATION. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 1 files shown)


docs/getting_started/quickstart.md
```diff
@@ -68,7 +68,8 @@ for Docker to recognize the GPU(s). The --gpus=1 flag indicates that 1
 system GPU should be made available to Triton for inferencing.
 
 ```
-$ docker run --gpus=1 --rm -p8000:8000 -p8001:8001 -p8002:8002 -v/full/path/to/docs/examples/model_repository:/models nvcr.io/nvidia/tritonserver:<xx.yy>-py3 tritonserver --model-repository=/models
+$ docker run --gpus=1 --rm -p8000:8000 -p8001:8001 -p8002:8002 -v/full/path/to/docs/examples/model_repository:/models --net=host
```
rmccorm4, line 71: The `-p8000:8000 ...` args actually publish these ports on the host, so `--net=host` is redundant here (and exposes everything). If the SDK container can't talk to server container, it's because the SDK container would need the `--net=host` flag (or to futz with docker networking).

rmccorm4 (changes requested):

docs/getting_started/quickstart.md
rmccorm4: So either this isn't needed here, or you should remove the `-p` port args here in favor of `--net=host` if we want simplicity

dyastremsky: Oh, didn't know that. Thanks! Closing this out.

PR #5360, opened 2023-02-14
Add parallel model loading speedup test

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/165

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2674,8 +2674,8 @@ class LifeCycleTest(tu.TestResultCollector):
             thread_1 = pool.submit(triton_client.load_model, "identity_model")
             time.sleep(2)  # wait between loads
             # Switch the model file to python backend
-            shutil.rmtree("models/identity_model")
-            shutil.move("identity_model", "models")
+            shutil.move("models", "models_v1")
+            shutil.move("models_v2", "models")
             # Second load should be blocked until the first completes
             thread_2 = pool.submit(triton_client.load_model, "identity_model")
             # Both loads should succeed
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2668,20 +2668,26 @@ class LifeCycleTest(tu.TestResultCollector):
                                                              verbose=True)
         except Exception as ex:
             self.assertTrue(False, "unexpected error {}".format(ex))
-        # Load same model concurrently
+        # Load same named model concurrently
         with concurrent.futures.ThreadPoolExecutor() as pool:
-            thread_1 = pool.submit(triton_client.load_model,
-                                   "identity_zero_1_int32")
+            # First load an 10 seconds delayed identity backend model
+            thread_1 = pool.submit(triton_client.load_model, "identity_model")
             time.sleep(2)  # wait between loads
-            thread_2 = pool.submit(triton_client.load_model,
-                                   "identity_zero_1_int32")
-            # Second load should wait until the first completes
+            # Switch the model file to python backend
+            shutil.rmtree("models/identity_model")
+            shutil.move("identity_model", "models")
+            # Second load should be blocked until the first completes
+            thread_2 = pool.submit(triton_client.load_model, "identity_model")
             # Both loads should succeed
             thread_1.result()
             thread_2.result()
+        # Check the model is ready
         self.assertTrue(triton_client.is_server_live())
         self.assertTrue(triton_client.is_server_ready())
-        self.assertTrue(triton_client.is_model_ready("identity_zero_1_int32"))
+        self.assertTrue(triton_client.is_model_ready("identity_model"))
+        # Check the finally loaded model is the second one
+        model_metadata = triton_client.get_model_metadata("identity_model")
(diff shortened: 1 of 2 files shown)


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2625,6 +2626,112 @@ def test_load_gpu_limit(self):
...
+            triton_client = grpcclient.InferenceServerClient("localhost:8001",
+                                                             verbose=True)
+        except Exception as ex:
+            self.assertTrue(False, "unexpected error {}".format(ex))
+        # Load same model concurrently
+        with concurrent.futures.ThreadPoolExecutor() as pool:
+            thread_1 = pool.submit(triton_client.load_model,
+                                   "identity_zero_1_int32")
+            time.sleep(2)  # wait between loads
+            thread_2 = pool.submit(triton_client.load_model,
```
GuanLuo, line 2676: Should change model file so it produce different inference results, then perform second load. In this way we can check if the proper model is loaded and available.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2625,6 +2626,112 @@ def test_load_gpu_limit(self):
...
+            time.sleep(2)  # wait between load and unload
+            unload_thread = pool.submit(triton_client.unload_model,
+                                        "identity_zero_1_int32")
+            load_thread.result()
+            unload_thread.result()
+        self.assertTrue(triton_client.is_server_live())
+        self.assertTrue(triton_client.is_server_ready())
+        self.assertFalse(triton_client.is_model_ready("identity_zero_1_int32"))
+        # Load ensemble_zero_1_float32 and unload its dependency while loading
+        # The unload operation should wait until the load is completed
```
GuanLuo, line 2707: This is not checked in the test. And is this the requirement? i.e. without knowing implementation detail, should the test expect the unload to be blocked until the previous load return, or it just expect the model is marked unavailable after both operation completed (return order not enforced)

kthui: > without knowing implementation detail, should the test expect the unload to be blocked until the previous load return

Yes. When load/unload requests are having conflict, any ongoing load/unload will block any incoming load/unload until the ongoing load/unload returns.

However, if there are >2 requests, i.e. the 1st is ongoing and the 2nd and 3rd are blocked, it is undefined whether the 2nd or 3rd request will be the next ongoing request after the 1st request returns. Thus, if the 1st is load, the 2nd is unload and 3rd is load. There are two possible execution orders: `load -> unload -> load` or `load -> load -> unload`, which the eventual state is unknown in advance. Therefore, if the test is going to cover >2 concurrent load/unload requests, the eventual model state is valid on both load or unload state.

qa/L0_lifecycle/lifecycle_test.py
kthui: They are the exact same model which there is only one model file. This is to show the same model can be loaded concurrently, and both should succeed.

Do you mean put the same model name on two different namespaces?

kthui:
Rebased.

kthui:
Rebased.

kthui:
Rebased.

qa/L0_lifecycle/lifecycle_test.py
tanmayv25: @kthui The second model load even though using the same model name should replace the model. I think you are missing that testing. I would suggest that you have 2 model files model_file_1 and model_file_2 in this test case..

Start with model_A with model_file_1, start loading model_A, replace model_file_1 with model_file_2 for model_A itself, then send request to load model_A again. 

You should be testing that finally loaded model_A is holding model_file_2.

kthui: Thanks for clarifying.
Update commit: [Update concurrent test on same model name diff model file](https://github.com/triton-inference-server/server/pull/5360/commits/3825dd2a1a81da664afe73c36f9d20c151465e0c)

kthui:
Rebased.

kthui:
Rebased.

qa/L0_lifecycle/test.sh
```diff
@@ -1705,6 +1711,117 @@ set -e
 kill $SERVER_PID
 wait $SERVER_PID
 
+LOG_IDX=$((LOG_IDX+1))
+
+# LifeCycleTest.test_concurrent_load_speedup
+rm -rf models
+mkdir models
+cp -r identity_zero_1_int32 models && mkdir -p models/identity_zero_1_int32/1
```
dyastremsky, line 1719: This is a bit difficult to read. Could you save `identity_zero_1_int32` as a variable to make it easier to see what's happening?

dyastremsky: Same for other variables that could simplify things below, like `python_identity_fp32`.

kthui: Update commit: 
[Move test model name into variable](https://github.com/triton-inference-server/server/pull/5360/commits/771d2217df343eff3d55c886bebb99cf10a3d3a2)

dyastremsky (approved):

tanmayv25 (approved):

GuanLuo (approved):

PR #5362, opened 2023-02-14
Expose option to enable model namespacing. Add test

GuanLuo:


kthui (dismissed):

qa/L0_model_namespacing/test_dynamic_resolution/addsub_repo/composing_model/1/model.py
```diff
@@ -0,0 +1,72 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import numpy as np
+import json
+import triton_python_backend_utils as pb_utils
+
+
+class TritonPythonModel:
```
tanmayv25, line 32: Instead of repeating the model.py in each directory, how about keeping the addsub and subadd models in L0_model_namespacing? The test.sh can then copy and setup these models into the corresponding composing_model directory?

GuanLuo:
CC @nskool the tests are example usage with the model namespacing feature.

qa/L0_model_namespacing/test.py
```diff
@@ -0,0 +1,328 @@
...
+        # error check
+        try:
+            self.client_.infer("composing_model", self.inputs_)
+            self.assertTrue(False,
+                            "expected error for inferring ambiguous named model")
+        except InferenceServerException as ex:
+            self.assertIn("ambiguity", ex.message())
+
+
+class ModelNamespaceExplicit(tu.TestResultCollector):
```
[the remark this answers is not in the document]
GuanLuo: Extracted some, but I don't think we should extract more to make the test workflow harder to follow (jumping around the code) for avoid duplication.

src/main.cc
```diff
@@ -648,14 +649,16 @@ std::vector<Option> options_
...
-  }
+      {OPTION_MODEL_LOAD_GPU_LIMIT, "model-load-gpu-limit",
+       "<device_id>:<fraction>",
+       "Specify the limit on GPU memory usage as a fraction. If model loading "
+       "on the device is requested and the current memory usage exceeds the "
+       "limit, the load will be rejected. If not specified, the limit will "
+       "not be set."
+      },
+      {OPTION_MODEL_NAMESPACING, "model-namespacing", Option::ArgBool,
+       "Whether model namespacing is enable or not. If true, model with the "
```
dyastremsky, line 660: Nit: "If true, model**s** with the..."

dyastremsky (approved):

kthui (approved):

qa/L0_model_namespacing/test_no_duplication/subadd_repo/composing_subadd/1/model.py
```diff
@@ -0,0 +1,6 @@
+import sys
+import os
+
+# load pre-defined QA model
+sys.path.append(os.environ["TRITON_QA_PYTHON_MODEL_DIR"])
+from python_subadd import *
```
rmccorm4, line 6: Nice code-reuse 🧠

rmccorm4 (approved):

qa/L0_model_namespacing/test.py
```diff
@@ -0,0 +1,330 @@
...
+    def test_duplication(self):
+        # Enable model namspacing on repositories that each repo has one
+        # ensemble and it requires an composing model ('composing_model') that
+        # exists in both repos.
+        # Expect all models are visible, the ensemble will pick up the correct
+        # model even the composing model can't be inferred individually.
+        self.check_health()
+
+        # infer check
+        for model in ["simple_addsub",]:
```
Tabrizian, line 129: nit: I think comma is not required here

Tabrizian (approved):
I was wondering how model namespacing would work with model file override? Does the namespace apply to the model whose files are being overridden or the namespace would be the temporary path where the model files are stored?

tanmayv25 (approved):

GuanLuo:
> I was wondering how model namespacing would work with model file override? Does the namespace apply to the model whose files are being overridden or the namespace would be the temporary path where the model files are stored?

Current implementation will assign empty ("") as the namespace, from empty namespace, it is probably not obvious that the model is loaded from different source (other than physical storage), but I don't think it should associate with the (storage) namespace for grouping purpose

PR #5375, opened 2023-02-16
Changing 'aarch64' configuration for model generation

mc-nv:


qa/common/gen_qa_model_repository
```diff
@@ -525,3 +523,66 @@ if [ $? -ne 0 ]; then
     echo -e "Failed"
     exit 1
 fi
+
+else # [[ "aarch64" != $(uname -m) ]]
```
nv-kmcgill53, line 527: Just to check my understanding. We only generate `tensorrt` models on aarch64 with this change? Will we use the other backend models created on x86_64 when testing on aarch64?

mc-nv: Yes it's correct observation for now it's only TensorRT models

nv-kmcgill53 (dismissed):

qa/common/gen_qa_model_repository
Tabrizian, line 527: Is this to add support for jetson in the model gen scripts?

mc-nv: Yes it's added to generate models directly on Jetson device. <br>
Jetson NVIDIA doesn't ship all the required images for `aarch64` architecture, so we workaround it by using Jetson hardware instead.

jbkyang-nvi: Can you maybe add a comment on that here? Also nit I think we have been doing if else statements in bash like this:
```
if [ something ]; then
   ...
else
   ...
fi
```
https://github.com/triton-inference-server/server/blob/main/qa/L0_infer/test.sh#L71-L75

mc-nv: It actually has this structure but to be able see it you need to load entire file

qa/python_models/init_args/model.py
```diff
@@ -42,13 +39,10 @@ def check_init_args(args):
         'model_instance_device_id':
             '0',
         'model_repository':
-            '/opt/tritonserver/qa/L0_backend_python/models/init_args',
+            os.getenv("TRITON_DIR", "/opt/tironserver") + '/qa/L0_backend_python/models/init_args',
```
krishung5, line 42: Where is the env `TRITON_DIR` being set? Is it in the gitlab yaml file?

mc-nv: The directory value is defined in our `test.sh` script:
https://github.com/triton-inference-server/server/blob/mchornyi-r23.02-jp/qa/L0_backend_python/test.sh#L42

krishung5: For testing on jetson, I was wondering where the `TRITON_DIR` is defined for the correct path on jetson runners? i.e. for jetson runners, the `TRITON_DIR` would be `/raid/home/gitlab-runner/builds/FeEZGYLm/0/dl/dgx/tritonserver/tritonserver`

qa/common/gen_qa_model_repository
jbkyang-nvi: In that case, is the indentation incorrect then? 
```
if [ $? -ne 0 ]; then
    echo -e "Failed"
    exit 1
fi

else # [[ "aarch64" != $(uname -m) ]]
```

is not following the convention

mc-nv: It didn't follow the convention earlier either, and reason for it that wea are using `cat >> file.name<<EOF` construction to create script entries. 
By start using  the  proposed convention  for `if/else` we have to bear in mind that the indentation will be applied to the autogenerated scripts.

qa/python_models/init_args/model.py
mc-nv: @krishung5  please ping me tomorrow we have to address this comment by referring to the internal repository.

krishung5: Synced offline with Misha. Resolving.

Tabrizian (approved):

qa/common/gen_qa_model_repository
Tabrizian: Actually I still don't fully understand how this works? The if condition used to be only for skipping openvino models on ARM SBSA but looks like with your changes we will be skipping ONNX and PyTorch too? Is my understanding correct?

Tabrizian (changes requested):

Tabrizian (approved):

qa/common/gen_qa_model_repository
rmccorm4: I think your understanding is correct Iman, it looks like after this change we are skipping everything except the final section on aarch64. 

@mc-nv could we make this easier to read by grouping the sections into functions:
```bash
function gen_openvino() {
...
}
function gen_onnx() {
...
}
...
### MAIN ###
if [[ x86 ]]
  gen_openvino()
  gen_onnx()
  gen_tf()
  ...
else   # aarch64
  # skip Openvino on aarch64
  gen_onnx()
  gen_tf(
  ...  
fi
```

rmccorm4 (changes requested):
See my comment here: https://github.com/triton-inference-server/server/pull/5375#discussion_r1117530347

qa/common/gen_qa_model_repository
GuanLuo: > reason for it that wea are using cat >> file.name<<EOF construction to create script entries.

The indentation will be added to the generated script, but will it make the script no longer work as intended?

rmccorm4: I think its fine to exclude the indentation if we move the sections to functions. If we want the indentation for readability, but not to include it in the resulting file, we can pipe through some `sed -e "s/^\s*//"` or similar to remove leading whitespaces on each line before writing to target file

qa/common/gen_qa_model_repository
```diff
@@ -525,3 +523,66 @@ if [ $? -ne 0 ]; then
...
+chmod -R 777 $HOST_DYNASEQIMPLICITDESTDIR
+python3 $HOST_SRCDIR/gen_qa_ragged_models.py --tensorrt --models_dir=$HOST_RAGGEDDESTDIR
+chmod -R 777 $HOST_RAGGEDDESTDIR
+python3 $HOST_SRCDIR/gen_qa_trt_format_models.py --models_dir=$HOST_FORMATDESTDIR
+chmod -R 777 $HOST_FORMATDESTDIR
+python3 $HOST_SRCDIR/gen_qa_trt_data_dependent_shape.py --models_dir=$HOST_DATADEPENDENTDIR
+chmod -R 777 $HOST_DATADEPENDENTDIR
+EOF
+    # Executing TRTSCRIPT
+    bash -ex $HOST_SRCDIR/$TRTSCRIPT
```
GuanLuo, line 581: Is this largely the same generation step for both arm and x86? Does it make sense to put the conditional to where the script is executed? i.e.
```
if arm:
    run script on bare-metal;
else:
    run in docker
```
Other enhancement can be that we condition on whether models of certain framework will be generated, i.e. at the top of script:
```
if arm:
    GEN_TRT_MODEL=true;
    GEN_XXX=false;
else:
    ....
```
Then in actual preparation of script and generation, there will be:
```
if GEN_XXX:
    generate_script;
    run_script in different style;
```

rmccorm4 (approved):
Approving to not block release as this is mostly cosmetic, however Misha will look into script cleanup next sprint.

PR #5397, opened 2023-02-22
Add GRPC option for restricted protocol access

GuanLuo:


Tabrizian (dismissed):

docs/customization_guide/inference_protocols.md
```diff
@@ -114,6 +114,55 @@ These options can be used to configure the KeepAlive settings:
 
 For client-side documentation, see [Client-Side GRPC KeepAlive](https://github.com/triton-inference-server/client/blob/main/README.md#grpc-keepalive).
 
+### Limit Endpoint Access
```
tanmayv25, line 117: Can we provide these settings in a json config file instead of CLI options? The optiion might look difficult to comprehend with protocol type, protocols(which are already quite a few and might grow in the future), port (and may be address if the user wants different address for this service).

GuanLuo: I am inclined to have separate config files but I don't think that is in favorite in a containerized setting, i.e. configure and start Triton in single command line without need to maintain / mount additional files. CC @Tabrizian @rmccorm4 as it is what I get out from cache dicussion.

Tabrizian: Maybe we could have a global configuration file that let's you express all the CLI flags (similar to MA) so that you can more easily describe nested settings? The CLI flags will be given more priority if it is described in both CLI and JSON config file.

GuanLuo: But we do want to keep the ability to specify via CLI alone, right? config file will be an addition instead of replacement?

Tabrizian: Yes, it will be an additional point of configuration.

docs/customization_guide/inference_protocols.md
```diff
@@ -114,6 +114,54 @@ These options can be used to configure the KeepAlive settings:
...
+
+In some use cases, Triton users may want to restrict the access of the protocols on a given endpoint.
+For example, there can be need for two separate protocol groups that one exposes standard inference
+protocols for user access, while the other one exposes other extension protocols for administration
+usage and should not be accessible by non-admin user.
+
+The following option can be specified to declare an restricted protocol group:
+
+```
+--grpc-restricted-protocol=<protocol_1>,<protocol_2>,...:<restricted-header>=<restricted-key>
```
whoisj, line 127: would the HTTP API be affected as well?

assuming it would not be. How would you expect the HTTP API to be affected?

trying to wrap my head around `--grpc-restricted-protocol=<protocol_1>,<protocol_2>`. 

could this be `--grpc-restricted-protocol=grpc,http`? If yes, then why the prefix `--grpc-` and not use `--restricted-protocol` as the option name?

docs/customization_guide/inference_protocols.md
```diff
@@ -114,6 +114,54 @@ These options can be used to configure the KeepAlive settings:
...
+* `restricted-key` : The value of the header to be matched in order to preceed
+the process of the specified protocols.
+
+#### Example
+
+To start server with the two GRPC endpoints in use case described above, the following command line arguments can be set to start "standard inference" endpoint on port 8002 and "administration" endpoint with health check on port 8003:
+
+```
+tritonserver \
+--endpoint-config=infer,type=grpc --endpoint-config=infer,grpc-port=8002 --endpoint-config=infer,protocols=health,metadata,inference \
```
whoisj, line 160: this set of options is quite different from the `--grpc-restricted-protocol=` options above.

how are they reconciled?

GuanLuo: I missed the update of example section, so the current way is that `--grpc-restricted-protocol` will be recognized instead of `--endpoint-config`, it is going to just enforce additional check on the specified protocols (different from HTTP/GRPC, it is how we call the "endpoint APIs").

So this example will be, where only KServe protocol will not be restricted, others need to have proper header "<admin-key> : <admin-value>". And there is still only one GRPC endpoint running to accept all GRPC requests
```
tritonserver --grpc-restricted-protocol=shared-memory,model-config,model-repository,statistics,trace:<admin-key>=<admin-value>
```

docs/customization_guide/inference_protocols.md
whoisj: configuration files are fine in containers as well. we use them heavily in TMS

qa/L0_grpc/python_unit_test.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python
+# Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 2: 2023

docs/customization_guide/inference_protocols.md
```diff
@@ -114,6 +114,55 @@ These options can be used to configure the KeepAlive settings:
...
+
+#### Example
+
+To start server with a subset of protocols to be restricted in use case
+described above, the following command line arguments can be set to accept
+"standard inference" request without additional header and the rest of the
+protocols with `<admin-key>=<admin-value>` specified in header:
+
+```
+tritonserver --grpc-restricted-protocol=shared-memory,model-config,model-repository,statistics,trace:<admin-key>=<admin-value> ...
```
Tabrizian, line 162: Since this is a header that Triton parses maybe we should not give the ability to specify the key to the user or a prefix like `triton-api-key` will be always attached to the key?

docs/protocol/README.md
```diff
@@ -36,17 +36,25 @@ plus several extensions that are defined in the following documents:
...
+Note that some extensions introduce new fields onto the inference protocols,
+and the other extensions define new protocols that Triton follows, please refer
+to the extension documents for detail.
+
 For the GRPC protocol, the [protobuf
 specification](https://github.com/triton-inference-server/common/blob/main/protobuf/grpc_service.proto)
 is also available. In addition, you can find the GRPC health checking protocol protobuf
 specification [here](https://github.com/triton-inference-server/common/blob/main/protobuf/health.proto).
+
+Users may configure the Triton endpoints, which implement the protocols, to
```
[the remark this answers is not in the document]
GuanLuo: not really authorization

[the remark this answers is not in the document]
Tabrizian: How about this:

"You can restrict access on any GRPC/HTTP endpoint that by requiring a specific header key-value pair. Please refer to ... for more information."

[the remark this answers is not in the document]
GuanLuo: I will change the wording for a bit, but I will keep the part about network settings because the pointing section actually cover endpoint configuration.

Tabrizian (approved):

docs/protocol/README.md
```diff
@@ -36,17 +36,25 @@ plus several extensions that are defined in the following documents:
 
 - [Binary tensor data extension](./extension_binary_data.md)
 - [Classification extension](./extension_classification.md)
-- [Model configuration extension](./extension_model_configuration.md)
-- [Model repository extension](./extension_model_repository.md)
 - [Schedule policy extension](./extension_schedule_policy.md)
 - [Sequence extension](./extension_sequence.md)
 - [Shared-memory extension](./extension_shared_memory.md)
+- [Model configuration extension](./extension_model_configuration.md)
```
dyastremsky, line 42: Is there a reason to move this lower? It looks like this list was alphabetic except for the last two: logging/parameters (which we should hopefully list alphabetically later too for consistency).

GuanLuo: It's moved so that the infer-related extensions are "grouped", given this parameter extension should be moved up. I think grouping by functionality is better than by alphabet. Alphabetic ordering works better when the reader knows what they are looking for.

src/command_line_parser.cc
```diff
@@ -1527,20 +1566,57 @@ TritonParser::ParseBackendConfigOption(const std::string& arg)
...
+TritonParser::ParseGrpcRestrictedProtocolOption(const std::string& arg)
 {
-  // Format is "<policy_name>,<setting>=<value>"
-  int delim_name = arg.find(",");
-  int delim_setting = arg.find("=", delim_name + 1);
+  try {
+    return ParseGenericConfigOption(arg, ":", "=");
+  }
+  catch (const ParseException& pe) {
+    // catch and throw exception with option specific message
```
dyastremsky, line 1575: Nit: optional. I also don't know if this comment is necessary/helpful. Your code is self-explanatory.

src/command_line_parser.cc
```diff
@@ -1527,20 +1566,57 @@ TritonParser::ParseBackendConfigOption(const std::string& arg)
...
-  // Check for 2 semicolons
-  if ((delim_name < 0) || (delim_setting < 0)) {
+std::tuple<std::string, std::string, std::string>
+TritonParser::ParseHostPolicyOption(const std::string& arg)
+{
+  try {
+    return ParseGenericConfigOption(arg, ",", "=");
+  }
+  catch (const ParseException& pe) {
+    // catch and throw exception with option specific message
```
dyastremsky, line 1593: Same comment as above (nit: optional; comment may not be necessary).

src/grpc_server.cc
```diff
@@ -4906,18 +5064,41 @@ Server::Server(
   model_infer_cq_ = builder_.AddCompletionQueue();
   model_stream_infer_cq_ = builder_.AddCompletionQueue();
 
+  // Read and set protocol-wise restriction
```
dyastremsky, line 5067: Nit: protocol-wide. May also want to clean up this comment for clarity.

docs/protocol/README.md
dyastremsky: I see. Thanks for explaining, Guan.

src/command_line_parser.cc
GuanLuo: I am going to keep this comment to emphasis that is re-throwing the exception for better messaging.

qa/L0_grpc/python_unit_test.py
```diff
@@ -0,0 +1,154 @@
...
+            grpcclient.InferInput('INPUT0', [1, 16], "INT32"),
+            grpcclient.InferInput('INPUT1', [1, 16], "INT32")
+        ]
+        inputs[0].set_data_from_numpy(np.ones(shape=(1, 16), dtype=np.int32))
+        inputs[1].set_data_from_numpy(np.ones(shape=(1, 16), dtype=np.int32))
+
+        # This test only care if the request goes through
+        with self.assertRaisesRegex(InferenceServerException,
+                                    "This protocol is restricted"):
+            results = self.client_.infer(model_name=self.model_name_,
```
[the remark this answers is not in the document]
dyastremsky: Do you want to get rid of `results` and just call the infer function like is done below? Not necessary to approve this PR, but could make it clearer that the result is unused.

dyastremsky (approved):

src/command_line_parser.cc
GuanLuo: "option" refers to this command line option

dyastremsky: Thanks for clarifying. For the future, if you hyphenate `option-specific`, that could make it clearer.

PR #5423, opened 2023-02-27
Add timed wait during UNLOAD while the model becomes UNAVAILABLE in SageMaker

nskool:
This PR updates the behavior of SageMaker UNLOAD function to use the repository index and verify that the model has been completely UNLOADED to the best of Triton's ability i.e. the model _state_ is `UNAVAILABLE` and the _reason_ is `unloaded`.

This (at least in case of python backend) ensures that the function returns only after the associated python workers for the models have been killed.

### DIFF
--- a/src/sagemaker_server.cc
+++ b/src/sagemaker_server.cc
@@ -681,8 +681,8 @@ SagemakerAPIServer::SageMakerMMEUnloadModelCheckStatus(const char* model_name)
     index_json.MemberAsString("reason", &reason, &reason_len);
 
     if (strcmp(name, model_name) == 0) {
-      if (strcmp(state, "UNAVAILABLE") == 0 &&
-          strcmp(reason, "unloaded") == 0) {
+      if (strcmp(state, UNLOAD_EXPECTED_STATE_.c_str()) == 0 &&
+          strcmp(reason, UNLOAD_EXPECTED_REASON_.c_str()) == 0) {
         is_model_unavailable = true;
       }
     }
@@ -694,35 +694,11 @@ void
 SagemakerAPIServer::SageMakerMMEUnloadModel(
     evhtp_request_t* req, const char* model_name)
 {
-  std::lock_guard<std::mutex> lock(mutex_);
-
   if (sagemaker_models_list_.find(model_name) == sagemaker_models_list_.end()) {
     LOG_VERBOSE(1) << "Model " << model_name << " is not loaded." << std::endl;
     evhtp_send_reply(req, EVHTP_RES_NOTFOUND); /* 404*/
     return;
   }
-
-  /* Always unload dependents as well - this is required to unload dependents in
-   * ensemble */
-  triton::common::TritonJson::Value request_parameters(
-      triton::common::TritonJson::ValueType::OBJECT);
-  triton::common::TritonJson::Value unload_parameter(
-      request_parameters, triton::common::TritonJson::ValueType::OBJECT);
-
-  unload_parameter.AddBool("unload_dependents", true);
-  request_parameters.Add("parameters", std::move(unload_parameter));
-
-  const char* buffer;
-  size_t byte_size;
-
-  triton::common::TritonJson::WriteBuffer json_buffer_;
-  json_buffer_.Clear();
-  request_parameters.Write(&json_buffer_);
-
-  byte_size = json_buffer_.Size();
-  buffer = json_buffer_.Base();
-
-  evbuffer_add(req->buffer_in, buffer, byte_size);
 
   /* Extract targetModel to log the associated archive */
   const char* targetModel =
@@ -730,28 +706,45 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
 
   LOG_INFO << "Unloading SageMaker TargetModel: " << targetModel << std::endl;
 
-  HandleRepositoryControl(req, "", model_name, "unload");
+  /* Always unload dependents as well - this is required to unload dependents in
+   * ensemble */
+  TRITONSERVER_Error* unload_err = nullptr;
+  unload_err =
+      TRITONSERVER_ServerUnloadModelAndDependents(server_.get(), model_name);
+
+  if (unload_err != nullptr) {
+    EVBufferAddErrorJson(req->buffer_out, unload_err);
+    evhtp_send_reply(req, EVHTP_RES_BADREQ);
+
+    LOG_ERROR
+        << "Error when unloading SagMaker Model with dependents for model: "
+        << model_name << std::endl;
+
+    TRITONSERVER_ErrorDelete(unload_err);
+  }
 
   /*Note: Model status check is repo-specific and therefore must be run before
    * unregistering the repo, else the model information is lost*/
   bool is_model_unavailable = false;
   uint32_t unload_time_in_secs = 0;
 
-  LOG_VERBOSE(1) << "Using Model Repository Index during UNLOAD to check for "
-                    "status of model: "
-                 << model_name;
-
-  /* Wait for the model to be completely unloaded. SageMaker waits a maximum of
-  360 seconds for the UNLOAD request to timeout. Setting a limit of 350 seconds
-  for Triton unload.*/
-  while (!is_model_unavailable && unload_time_in_secs < UNLOAD_TIMEOUT_SECS_) {
-    is_model_unavailable = SageMakerMMEUnloadModelCheckStatus(model_name);
-    sleep(1);
-    unload_time_in_secs += 1;
+  /* Wait for the model to be completely unloaded. SageMaker waits a maximum
+  of 360 seconds for the UNLOAD request to timeout. Setting a limit of 350
+  seconds for Triton unload. This should be run only if above UNLOAD call has
+  succeeded.*/
+  if (unload_err == nullptr) {
+    LOG_VERBOSE(1) << "Using Model Repository Index during UNLOAD to check for "
+                      "status of model: "
+                   << model_name;
+    while (!is_model_unavailable &&
+           unload_time_in_secs < UNLOAD_TIMEOUT_SECS_) {
+      is_model_unavailable = SageMakerMMEUnloadModelCheckStatus(model_name);
+      sleep(1);
+      unload_time_in_secs += 1;
+    }
+    LOG_INFO << "UNLOAD for model " << model_name << " took "
+             << unload_time_in_secs << " seconds.";
   }
-
-  LOG_VERBOSE(1) << "UNLOAD for model " << model_name << " took "
-                 << unload_time_in_secs << " seconds." << std::endl;
 
   if (is_model_unavailable == false &&
       unload_time_in_secs >= UNLOAD_TIMEOUT_SECS_) {
@@ -763,7 +756,7 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
 
   std::string repo_parent_path = sagemaker_models_list_.at(model_name);
 
-  TRITONSERVER_Error* unload_err = TRITONSERVER_ServerUnregisterModelRepository(
+  unload_err = TRITONSERVER_ServerUnregisterModelRepository(
       server_.get(), repo_parent_path.c_str());
 
   if (unload_err != nullptr) {
@@ -772,8 +765,11 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
     LOG_ERROR << "Unable to unregister model repository for path: "
               << repo_parent_path << std::endl;
     TRITONSERVER_ErrorDelete(unload_err);
+  } else {
+    evhtp_send_reply(req, EVHTP_RES_OK);
   }
 
+  std::lock_guard<std::mutex> lock(mutex_);
   sagemaker_models_list_.erase(model_name);
 }
 
@@ -957,8 +953,8 @@ SagemakerAPIServer::SageMakerMMELoadModel(
       if (config_fstream.is_open()) {
         ensemble_config_content << config_fstream.rdbuf();
       } else {
-        continue;  // A valid config.pbtxt does not exit at this path, or cannot
-                   // be read
+        continue;  // A valid config.pbtxt does not exit at this path, or
+                   // cannot be read
       }
 
       /* Compare matched string with `platform: "ensemble"` or
@@ -1086,5 +1082,4 @@ SagemakerAPIServer::SageMakerMMELoadModel(
 
   return;
 }
-
 }}  // namespace triton::server
--- a/src/sagemaker_server.cc
+++ b/src/sagemaker_server.cc
@@ -655,8 +655,6 @@ SagemakerAPIServer::SageMakerMMECheckUnloadedModelIsUnavailable(
   TRITONSERVER_MessageSerializeToJson(
       server_model_index_message_, &index_buffer, &index_byte_size);
 
-  LOG_VERBOSE(1) << "Printing entire model index repository" << index_buffer;
-
   /* Read into json buffer*/
   triton::common::TritonJson::Value server_model_index_json;
   server_model_index_json.Parse(index_buffer, index_byte_size);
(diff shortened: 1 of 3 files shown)


src/sagemaker_server.cc
```diff
@@ -670,6 +732,35 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
...
+                    "status of model: "
+                 << model_name;
+
+  /* Wait for the model to be completely unloaded. SageMaker waits a maximum of
+  360 seconds for the UNLOAD request to timeout. Setting a limit of 350 seconds
+  for Triton unload.*/
+  while (!is_model_unavailable && unload_time_in_secs < UNLOAD_TIMEOUT_SECS_) {
+    is_model_unavailable = SageMakerMMEUnloadModelCheckStatus(model_name);
+    sleep(1);
+    unload_time_in_secs += 1;
```
rohithkrn, line 750: I believe this does not account for time taken for this `SageMakerMMEUnloadModelCheckStatus(model_name)` function call. Is it too small to be neglected?

nskool: That's correct, in SM-Triton, there's one model repo per model, so the model repository index is expected to contain only one model (except in case of ensembles, where it maybe approx ~5-8), so the call is expected to return within few milliseconds.

rohithkrn: IMHO, we can measure time for that as well and report a more accurate number. It may not matter much in normal cases but if something weird happens in that function, we will be logging incorrect unload time.

davidthomas426: Related issue: `sleep` is not exact, either. 

Rather than doing `unload_time_in_secs += 1`, we should be measuring elapsed time since start of loop after each sleep. This would solve both problems.

src/sagemaker_server.cc
```diff
@@ -628,6 +628,68 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
   }
 }
 
+bool
+SagemakerAPIServer::SageMakerMMEUnloadModelCheckStatus(const char* model_name)
```
davidthomas426, line 632: I don't love the name if it just returns a `bool`. Hard to tell without looking at the code what the bool represents. Does `true` mean it is unloaded? Does `true` mean the model is available, so not yet unloaded? I have no idea since I have not read the implementation code yet.

src/sagemaker_server.cc
```diff
@@ -628,6 +628,68 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
+
+  bool is_model_unavailable = false;
+
+
+  for (size_t id = 0; id < server_model_index_json.ArraySize(); ++id) {
+    triton::common::TritonJson::Value index_json;
+    server_model_index_json.IndexAsObject(id, &index_json);
+
+    index_json.MemberAsString("name", &name, &name_len);
+    index_json.MemberAsString("version", &version, &version_len);
```
davidthomas426, line 679: Do you need to match on a particular version? Or with sagemaker API, is there only ever one version for a given model name?

src/sagemaker_server.cc
```diff
@@ -628,6 +628,68 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
+    index_json.MemberAsString("state", &state, &state_len);
+    index_json.MemberAsString("reason", &reason, &reason_len);
```
davidthomas426, line 681: Are we guaranteed that these calls will succeed? We're ignoring a return value that could be success or error, according to https://github.com/triton-inference-server/common/blob/main/include/triton/common/triton_json.h#L799-L840

src/sagemaker_server.cc
```diff
@@ -670,6 +732,35 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
...
+  LOG_VERBOSE(1) << "Using Model Repository Index during UNLOAD to check for "
+                    "status of model: "
+                 << model_name;
+
+  /* Wait for the model to be completely unloaded. SageMaker waits a maximum of
+  360 seconds for the UNLOAD request to timeout. Setting a limit of 350 seconds
+  for Triton unload.*/
+  while (!is_model_unavailable && unload_time_in_secs < UNLOAD_TIMEOUT_SECS_) {
+    is_model_unavailable = SageMakerMMEUnloadModelCheckStatus(model_name);
+    sleep(1);
```
davidthomas426: The sleep amount could also be a constant, like `UNLOAD_TIMEOUT_SECS_` is. Could also use `std::this_thread::sleep_for(std::chrono::milliseconds(UNLOAD_SLEEP_MILLISECONDS_));` instead of `sleep` to get more granularity, if desired (especially if you frequently need more exactly one sleep but it's usually done much more quickly than one second).

davidthomas426: Is it ok to block the thread with `sleep` like this? Or do we need to do some kind of shenanigans with event loop to check again after a timeout without blocking the thread?

src/sagemaker_server.cc
```diff
@@ -772,8 +765,11 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
     LOG_ERROR << "Unable to unregister model repository for path: "
               << repo_parent_path << std::endl;
     TRITONSERVER_ErrorDelete(unload_err);
+  } else {
+    evhtp_send_reply(req, EVHTP_RES_OK);
   }
 
+  std::lock_guard<std::mutex> lock(mutex_);
```
davidthomas426, line 772: Just a comment, but it can be nice to name a mutex based on what it is intended to protect. `models_list_mutex_` would make it clearer what it is needed for.

src/sagemaker_server.cc
```diff
@@ -764,11 +793,11 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
     evhtp_send_reply(req, EVHTP_RES_BADREQ);
     LOG_ERROR << "Unable to unregister model repository for path: "
               << repo_parent_path << std::endl;
-    TRITONSERVER_ErrorDelete(unload_err);
   } else {
     evhtp_send_reply(req, EVHTP_RES_OK);
   }
 
+  TRITONSERVER_ErrorDelete(unload_err);
```
davidthomas426, line 800: I think this is a segfault waiting to happen, since you're returning `nullptr` at the end of the `TRITONSERVER_ServerUnregisterModelRepository`.

`TRITONSERVER_ErrorDelete` does unconditional `delete` on the argument: https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L712-L717

davidthomas426 (changes requested):

src/sagemaker_server.cc
GuanLuo: I think delete on `nullptr` will not do any harm, but I see the `unload_err` is reused for multiple API calls and there is chance of memory leak, i.e. `unload_err` may be set in line 743 and the pointer is overwritten in line 779

src/sagemaker_server.cc
```diff
@@ -628,61 +628,168 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
+  for (size_t id = 0; id < server_model_index_json.ArraySize(); ++id) {
+    triton::common::TritonJson::Value index_json;
+    server_model_index_json.IndexAsObject(id, &index_json);
+
+    RETURN_IF_ERR(index_json.MemberAsString("name", &name, &name_len));
+
+    if (std::string(name) == std::string(model_name)) {
+      RETURN_IF_ERR(index_json.MemberAsString("state", &state, &state_len));
+
+      if (std::string(state) == UNLOAD_EXPECTED_STATE_) {
```
GuanLuo, line 678: I think you should do the opposite checking here, i.e. set `is_model_unavailable = true;` at the beginning and mark it false if you have seen the model NOT in expected state and reason. Because the index will report model with multiple versions in different entries, say there is model "A" with version 1, 2, depending on the timing, you may receive the following model index:
```
{
  name: "A",
  version: "1",
  state: UNLOADING,
  reason: ""
},
{
  name: "A",
  version: "2",
  state: UNAVAILABLE,
  reason: "unloaded"
}
```

src/sagemaker_server.cc
davidthomas426: You're right, sorry. delete on nullptr is actually fine. From https://en.cppreference.com/w/cpp/language/delete:

> If expression evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer.

src/sagemaker_server.cc
nskool: Thank you for comments, I've addressed this by using elapsed time calculation.

src/sagemaker_server.cc
nskool: Makes sense, I have changed the function name, and am no longer using a return type of bool

src/sagemaker_server.cc
nskool: That's correct, we expect only one version of for a model name

src/sagemaker_server.cc
nskool: Good point - I've updated the return type to an error type instead, that will be logged in case of failure to parse.

src/sagemaker_server.cc
nskool: Thanks, used the `std::this_thread::sleep_for(
          std::chrono::milliseconds(UNLOAD_SLEEP_MILLISECONDS_))` to sleep for a shorter duration

src/sagemaker_server.cc
nskool: Thanks, done

src/sagemaker_server.cc
```diff
@@ -725,8 +734,8 @@ SagemakerAPIServer::SageMakerMMEUnloadModel(
 
   /*Note: Model status check is repo-specific and therefore must be run before
    * unregistering the repo, else the model information is lost*/
-  bool is_model_unavailable = false;
-  uint32_t unload_time_in_secs = 0;
+  bool* is_model_unavailable = new bool(false);
```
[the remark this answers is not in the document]
nskool: Thanks, done

src/sagemaker_server.cc
nskool: Updated the code to not re-use the unload_err pointer. 

Regarding the multiple version scenario - SM currently expects only one model version in a single model repository, and this logic currently holds fine on testing. I might consider updating this in a future PR, thank you for the suggestion. Even so, I've added a log statement that also includes the version number.

src/sagemaker_server.cc
nskool: Updated the code to not re-use the unload_err pointer

src/sagemaker_server.cc
GuanLuo: > so the model repository index is expected to contain only one model.

Note that index API poll all registered model repos, so the performance is actually proportioned to all "visible" models. Although that is probably still neglectable with respect to the unload time.

src/sagemaker_server.cc
```diff
@@ -628,69 +628,184 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
 }
 
+TRITONSERVER_Error*
+SagemakerAPIServer::SageMakerMMECheckUnloadedModelIsUnavailable(
+    const char* model_name, bool* is_model_unavailable)
+{
+  /* Use the RepositoryIndex API to check if the model state has become
+  UNAVAILABLE i.e. model is no longer in the 'in-the-process-of' being
+  UNLOADED. Consequently, the reason field should be 'unloaded'.*/
+  TRITONSERVER_Message* server_model_index_message_ = nullptr;
```
GuanLuo, line 638: remove `_` since it is local

src/sagemaker_server.cc
```diff
@@ -628,69 +628,184 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
-  const char* buffer;
-  size_t byte_size;
+    LOG_ERROR
+        << "Error when unloading SagMaker Model with dependents for model: "
+        << model_name << std::endl;
 
-  triton::common::TritonJson::WriteBuffer json_buffer_;
-  json_buffer_.Clear();
-  request_parameters.Write(&json_buffer_);
+    TRITONSERVER_ErrorDelete(unload_err);
```
GuanLuo, line 733: should it return early if unload fails?

src/sagemaker_server.cc
```diff
@@ -628,69 +628,184 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
+                                .count();
+    }
+    LOG_INFO << "UNLOAD for model " << model_name << " completed in "
+             << unload_time_in_secs << " seconds.";
+    TRITONSERVER_ErrorDelete(unload_err);
+  }
 
-  HandleRepositoryControl(req, "", model_name, "unload");
+  if (is_model_unavailable == false &&
+      unload_time_in_secs >= UNLOAD_TIMEOUT_SECS_) {
```
GuanLuo, line 775: style:
```suggestion
  if ((is_model_unavailable == false) &&
      (unload_time_in_secs >= UNLOAD_TIMEOUT_SECS_)) {
```

GuanLuo (commented):
minor comment

src/sagemaker_server.cc
```diff
@@ -628,69 +628,184 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
-  unload_parameter.AddBool("unload_dependents", true);
-  request_parameters.Add("parameters", std::move(unload_parameter));
+  if (unload_err != nullptr) {
+    EVBufferAddErrorJson(req->buffer_out, unload_err);
+    evhtp_send_reply(req, EVHTP_RES_BADREQ);
 
-  const char* buffer;
-  size_t byte_size;
+    LOG_ERROR
+        << "Error when unloading SagMaker Model with dependents for model: "
```
davidthomas426, line 730: typo: SagMaker

src/sagemaker_server.cc
```diff
@@ -628,69 +628,184 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
+  UNAVAILABLE i.e. model is no longer in the 'in-the-process-of' being
+  UNLOADED. Consequently, the reason field should be 'unloaded'.*/
+  TRITONSERVER_Message* server_model_index_message_ = nullptr;
+  uint32_t ready_flag = 0;  // value of 1 should be set if only the 'ready'
+                            // models are required from the index. In this case,
+                            // we need all models.
+  TRITONSERVER_ServerModelIndex(
+      server_.get(), ready_flag, &server_model_index_message_);
+
+  std::shared_ptr<TRITONSERVER_Message> managed_msg(
```
davidthomas426, line 645: Generally I'd recommend using `unique_ptr` rather than `shared_ptr` unless there's a specific reason `shared_ptr` is needed (and there's not, here).

src/sagemaker_server.cc
nskool: Done, thanks

src/sagemaker_server.cc
nskool: Makes sense, added, thanks!

src/sagemaker_server.cc
nskool: Done, thanks

src/sagemaker_server.cc
```diff
@@ -628,69 +628,184 @@ SagemakerAPIServer::SageMakerMMEHandleInfer(
...
-  size_t byte_size;
+    LOG_ERROR
+        << "Error when unloading SagMaker Model with dependents for model: "
+        << model_name << std::endl;
 
-  triton::common::TritonJson::WriteBuffer json_buffer_;
-  json_buffer_.Clear();
-  request_parameters.Write(&json_buffer_);
+    TRITONSERVER_ErrorDelete(unload_err);
+  }
```
davidthomas426, line 735: Isn't it a problem that this block falls through to the rest of the function?

src/sagemaker_server.cc
nskool: Thanks, fixed

src/sagemaker_server.cc
nskool: Thanks, this was addressed as part of the previous update, so now the function should return early

src/sagemaker_server.cc
nskool: My concern was whether this is required due to multiple evhtp threads calling on the unload method. I can check more on this and modify at a later point in another PR, noted it...

@GuanLuo do you have any suggestion?

davidthomas426 (approved):

src/sagemaker_server.cc
GuanLuo: The returned message is going to be unique to the caller, so there is no share of ownership here. `unique_ptr` will be better, except that syntax-wise it is a bit messy to attach custom deleter (in C++11).

PR #5428, opened 2023-02-27
GRPC endpoint clean up. Add missing protocol in doc

GuanLuo:


src/grpc_server.cc
```diff
@@ -61,7 +61,7 @@
 
 #define REGISTER_GRPC_INFER_THREAD_COUNT 2
 
-namespace triton { namespace server {
+namespace triton { namespace server { namespace grpc {
```
dyastremsky, line 64: Nice job cleaning this code up!

What is the purpose/benefit of the grpc namespace? I don't see where it's changing anything beyond requiring the addition of "::" before each grpc namespacing.

GuanLuo: I missed the part of dropping GRPC prefix from the objects, but the idea is to group correlated objects in the form of namespacing. Otherwise, we need to do the grouping via prefixing. If you look across all endpoints, you will see that all endpoint live under triton::server namespace, and to distinguish the names, there is prefix with the endpoint types (i.e. `HttpXXX`, `GrpcXXX`).

Another thing is that if you are modifying within GRPC endpoint, you can refer to objects without the mention of "GRPC", the endpoint type is implied in this context.

dyastremsky (dismissed):

tanmayv25:
> I missed the part of dropping GRPC prefix from the objects, but the idea is to group correlated objects in the form of namespacing. 

This is still pending.

dyastremsky (approved):

rmccorm4 (commented):
Nice refactoring 🎉

Tabrizian (approved):

rmccorm4 (approved):

PR #5437, opened 2023-03-01
Update NGC version post-23.02 release

mc-nv:


build.py
```diff
@@ -67,9 +67,9 @@
 # incorrectly load the other version of the openvino libraries.
 #
 TRITON_VERSION_MAP = {
-    '2.31.0dev': (
-        '23.02dev',  # triton container
```
Tabrizian, line 71: Should this be 23.03dev?

PR #5443, opened 2023-03-02
Extract command line parser into separate file

GuanLuo:


src/command_line_parser.cc
```diff
@@ -0,0 +1,1528 @@
...
+  }
+  if ((arg == "false") || (arg == "off") || (arg == "0")) {
+    return false;
+  }
+
+  throw ParseException("invalid value for bool option: " + arg);
+}
+
+// Template specialization for ParsePairOption
+// [FIXME] replace ParseXXXOPtion with these
```
dyastremsky, line 57: This feels a bit messy to read through. Could you separate the ParseXXXOption functions away from the template functions section and have a header comment for each (e.g. explaining how/when they are used, since both are being compiled into the code).

src/command_line_parser.cc
```diff
@@ -0,0 +1,1528 @@
...
+#endif
+
+template <>
+double
+ParseOption(const std::string& arg)
+{
+  return std::stod(arg);
+}
+
+// Condition here merely to avoid compilation error, this function will
```
dyastremsky, line 102: Can you explain this further? What is the compilation error and why do we need this? If we need to keep this, ideally we need the logic included in the comment.

src/command_line_parser.cc
```diff
@@ -0,0 +1,1528 @@
...
+  return std::stoll(arg);
+}
+
+#if 0
+float
+ParseFloatOption(const std::string arg)
+{
+  return std::stof(arg);
+}
+#endif
```
rmccorm4, line 93: Is this dead code?

src/command_line_parser.cc
rmccorm4, line 102: Optional if you agree, but I believe this file could be a good candidate for starting to break out some ifdefs into `constexpr bool`s like [rapids_triton does](https://github.com/rapidsai/rapids-triton/blob/4a522cfacf9d1792376056c30d63657ee34b9d91/cpp/include/rapids_triton/build_control.hpp#L24-L28). Then all `ifdefs` in this source can be replaced with `if (logging_enabled) { }` and it shouldn't complain about unused functions/variables even when these are disabled.

GuanLuo: This is going to be a different approach to address the problem, more on runtime check (and compiler optimization) rather than compile time selection. I am not going to do this in this PR and it will need a discussion before we agree to move towards this way.

rmccorm4: Sounds good.

GuanLuo: it is mentioned in the second half of the comment, not the exact compilation error code though

src/command_line_parser.h
```diff
@@ -0,0 +1,317 @@
...
+  bool allow_vertex_ai_{false};
+  std::string vertex_ai_address_{"0.0.0.0"};
+  int32_t vertex_ai_port_{8080};
+  // The number of threads to initialize for the Vertex AI HTTP front-end.
+  int vertex_ai_thread_cnt_{8};
+  std::string vertex_ai_default_model_{};
+#endif  // TRITON_ENABLE_VERTEX_AI
+
+  // [FIXME] who should call this function?
+  void CheckPortCollision();
```
rmccorm4, line 263: Is this comment still valid?

rmccorm4 (dismissed):
LGTM, just one question above

src/command_line_parser.h
GuanLuo: This will probably be a helper function in `main.cc`, once the parser is broken down, the collision checking will be done after all endpoint parsing is done so can't be part of any option struct

rmccorm4 (approved):

PR #5455, opened 2023-03-06
Lower the tensor size to avoid intermittent issue

krishung5:
The tensor byte size is still at least 1GB in order to trigger the out of memory error.


Tabrizian (dismissed):

krishung5:
@Tabrizian I went ahead and added some unittest cases for the response iterator. Please let me know if there is other corner cases that I missed. Thanks!

qa/python_models/bls/model.py
```diff
@@ -553,6 +555,86 @@ def test_timeout(self):
...
+
+            infer_response = next(infer_responses)
+            self.assertFalse(infer_response.has_error())
+            output0 = pb_utils.get_output_tensor_by_name(infer_response, 'OUT')
+            self.assertIsNotNone(output0)
+            self.assertEqual(4, output0.as_numpy())
+
+            # The iterator now should only have 1 remaining response.
+            infer_responses = self._test_response_iterator_square(
+                1, response_num, infer_responses)
```
Tabrizian, line 635: I think we also need to have a test for the case that the iterator is not fully used and we just delete it. 

```
delete inference_responses
```

I think it might be a bit difficult to confirm whether the responses have all been retrieved.

Tabrizian (approved):

qa/python_models/bls_memory/model.py
```diff
@@ -56,7 +56,7 @@ def _send_identity_tensor(self, size, is_decoupled):
         return input0_np, infer_response
 
     def test_bls_out_of_memory(self):
-        tensor_size = 1024 * 1024 * 1024
```
Tabrizian, line 59: Is it still going to be more than 1GB after changing it to 256?

krishung5: I believe the unit is in float so it's 4*256*1024*1024 bytes which is still larger than 1GB.

Tabrizian (approved):

PR #5193, opened 2022-12-22
fix: usinginfo->byte_size_ after delete info

zimo-geek:
fix: using info->byte_size_ after delete info

PR #5462, opened 2023-03-07
Fix invalid use of info after deletion

rmccorm4:
Finishing what was started in https://github.com/triton-inference-server/server/pull/5193 after lack of activity


oandreeva-nv (approved):
LGTM!

PR #5468, opened 2023-03-08
Update the tracing in frontends to use TraceReportActivity

Tabrizian:


rmccorm4 (commented):
Looks good to me, but I'll let Guan comment in case I'm not aware of some trace related impact. Or if we need any additional API fields to avoid some v2/v3 API down the line.

krishung5:
Would we want to have some example for the usage of adding custom info to the trace in the backend?

src/tracer.cc
```diff
@@ -287,24 +287,12 @@ void
...
-    }
-    *ss << "{\"id\":" << trace_id_ << ",\"timestamps\":["
-        << "{\"name\":\"" << name << "\",\"ns\":" << timestamp_ns << "}]}";
-  }
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceTraceSetName(trace_, name.c_str()),
+      "set trace name");
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceTraceReportActivity(trace_, timestamp_ns),
+      "report trace activity");
```
GuanLuo, line 295: I don't think this function requires to go through the C API for tracing, it is the helper function to be use by the frontend (which already knows the concrete instance of the trace manager), if the reason of change is to reuse code, it should be able to call the `TraceActivity()` instead

PR #5502, opened 2023-03-14
Add onnx and python models back to L0_infer_valgrind test

krishung5:
Python models: Doesn't cause server hanging, only has OOM issue when loading all models at once.
Onnx models: It takes about 12 hours to load all the models. At some point the server stuck for a while when this [line](https://github.com/microsoft/onnxruntime/blob/rel-1.13.1/include/onnxruntime/core/common/spin_pause.h#L22) is called, but it will resume eventually. Reduce the instance count and the number of models being loaded help with the case.
Hence, modify to test with python and onnx models in separate runs.

### DIFF
--- a/qa/L0_infer/test.sh
+++ b/qa/L0_infer/test.sh
@@ -63,6 +63,7 @@ if [ "$TEST_VALGRIND" -eq 1 ]; then
     LEAKCHECK_ARGS_BASE="--leak-check=full --show-leak-kinds=definite --max-threads=3000 --num-callers=20"
     SERVER_TIMEOUT=4000
     rm -f $LEAKCHECK_LOG_BASE*
+    BACKENDS="python"
 fi
 
 if [ "$TEST_SYSTEM_SHARED_MEMORY" -eq 1 ] || [ "$TEST_CUDA_SHARED_MEMORY" -eq 1 ]; then
--- a/qa/L0_infer/test.sh
+++ b/qa/L0_infer/test.sh
@@ -63,8 +63,8 @@ if [ "$TEST_VALGRIND" -eq 1 ]; then
     LEAKCHECK_ARGS_BASE="--leak-check=full --show-leak-kinds=definite --max-threads=3000 --num-callers=20"
     SERVER_TIMEOUT=4000
     rm -f $LEAKCHECK_LOG_BASE*
-    # Remove 'python' and 'python_dlpack' from BACKENDS. Need to run python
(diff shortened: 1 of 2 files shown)


qa/L0_infer/test.sh
```diff
@@ -331,6 +342,102 @@ for TARGET in cpu gpu; do
...
+    fi
+
+    for TARGET in cpu gpu; do
+      rm -fr *models
+      generate_model_repository
+      mkdir nobatch_models
+      mv ./models/*nobatch_* ./nobatch_models/.
+      cp -fr ./models/nop_* ./nobatch_models/.
+
+      for BATCHING in batch nobatch; do
```
tanmayv25, line 375: BATCHING -> BATCHING_MODE

qa/L0_infer/test.sh
```diff
@@ -331,6 +342,102 @@ for TARGET in cpu gpu; do
...
+  TESTING_BACKENDS="python python_dlpack onnx"
+  if [ "$TEST_JETSON" == "0" ]; then
+    if [[ "aarch64" != $(uname -m) ]] ; then
+        pip3 install torch==1.13.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+    else
+        pip3 install torch==1.13.0 -f https://download.pytorch.org/whl/torch_stable.html
+    fi
+  fi
+
+  for BACKENDS in $TESTING_BACKENDS; do
```
tanmayv25, line 360: BACKENDS -> FW 

For consistency.

qa/L0_infer/test.sh
krishung5: Renamed.

qa/L0_infer/test.sh
krishung5: The `BACKENDS` env is exported and used in [infer_test.py](https://github.com/triton-inference-server/server/pull/5502/files#diff-59676af454541b48dd8e7c83879c5f50c8994c0e3161d9801f05b23b9b7e3e94L50). It's set to all backends [here](https://github.com/triton-inference-server/server/pull/5502/files#diff-66b8024d0503d76eac74493835a3d7a6e192903c3daca39f00c54fa96e9120b9R130) in the `L0_infer--base` test. I think it would make sense to use `BACKENDS`  so that we can export the env directly. Let me know your thought on this.

tanmayv25 (approved):

PR #5505, opened 2023-03-15
Fixing issue in ResolveBackendConfigs

oandreeva-nv:
This PR adds tests to check edge cases for backend configuration resolution.
Current logic for backend config resolution sorts default configs and specific configs lexicographically. Then it iterates through them  and prioritizes backend specific config over the default one.
 We would like to make sure that in the following situation (example)
 ```
          specific_config = {"a":4}
           default_config = {"b":5,"c":6,"d":7}
```
"c" and "d" are not lost.  

And in the following situation
 (example)
 ```
          specific_config = {"x":4, "y":5, "z":9}
           default_config = {"b":5,"c":6,"d":7}
```
 "y" and "z" are not lost.

### DIFF
--- a/qa/L0_backend_config/test.sh
+++ b/qa/L0_backend_config/test.sh
@@ -350,7 +350,7 @@ fi
 # defaults are properly overriden.
 # One of defaultconfigs is `min-compute-capability`. This test
 # checks if it is properlly overriden.
-MIN_COMPUTE_CAPABILITY="XX"
+MIN_COMPUTE_CAPABILITY=XX
 SERVER_ARGS="--backend-config=onnxruntime,min-compute-capability=$MIN_COMPUTE_CAPABILITY $COMMON_ARGS"
 SERVER_LOG=$SERVER_LOG_BASE.global_configs.log
 run_server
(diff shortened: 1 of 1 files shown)


GuanLuo (approved):

tanmayv25 (dismissed):

oandreeva-nv:
@GuanLuo  Let me know if new tests are ok, waiting for CI to run.

GuanLuo:
Why the test needs to be changed? Shouldn't the core simplification result in the same behavior?

oandreeva-nv:
@GuanLuo The former logic involved sorting. Global configs and Specific configs were first sorted lexicographically. Then we iterated over global configs and specific configs and populated final config with settings: 
```
Part 1:
if (global_setting < specific_setting) -> final_config.push_back(global_setting) ; increase global_config index
else if (specific_setting > global_setting) -> final_config.push_back(specific_setting); increase specific_config index
else ->   final_config.push_back(specific_setting) increase both
Part 2:
finally if global_config or specific_config still have entries left, add these entries to final_config 
```
The initial issue was in the last part, when not all entries were added to final config in the end. Thus tests were designed so that to test edge cases related to iteration over 2 sorted arrays: 1)if more than 1 entry is left in global config, Part 2 adds all of them to final config, and 2) vice versa.

Since updated logic does not depend on sorting any longer and we iterate first over the whole global config (also default config), then over the whole specific config, the former tests did not make sense. New tests make sure, that specific configs are added to final configs in full, and any default parameters that need to be overriden, are overriden.

GuanLuo (dismissed):
Need to update copyright year at the top

GuanLuo (approved):

dyastremsky:
Core PR associated with this one (for posterity): https://github.com/triton-inference-server/core/pull/175

PR #5511, opened 2023-03-15
Updating Jetson  model generation

mc-nv:


qa/common/gen_qa_model_repository
```diff
@@ -523,66 +525,3 @@ if [ $? -ne 0 ]; then
     echo -e "Failed"
     exit 1
 fi
-
-else # [[ "aarch64" != $(uname -m) ]]
```
jbkyang-nvi, line 527: why are we deleting this part?

mc-nv: This is a revert to unblock a QA

nv-kmcgill53 (approved):

PR #5521, opened 2023-03-17
Fix memory leak in http server

Tabrizian:


GuanLuo (approved):

GuanLuo (approved):
I think it is not preferred to try-and-cast the parameter until the operation success. Should look into exposing the rapid-json type method in `triton_json.h` and use that for checking type of the parameter value

Tabrizian:
@GuanLuo will address this in a follow up PR.

PR #5531, opened 2023-03-21
Update Dockerfile.win10.min

mc-nv:


Dockerfile.win10.min
```diff
@@ -95,12 +92,11 @@ ARG CUDA_INSTALL_ROOT_WP="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${
 
 WORKDIR /
 
-ARG TENSORRT_VERSION=8.5.1.7
-ARG TENSORRT_ZIP="TensorRT-${TENSORRT_VERSION}.Windows10.x86_64.cuda-11.8.cudnn8.6.zip"
```
krishung5, line 99: Is `TENSORRT_ZIP` deleted by accident? I'm seeing `TENSORRT_ZIP` being used below.

krishung5 (changes requested):

Dockerfile.win10.min
```diff
@@ -66,10 +66,7 @@ RUN vcpkg.exe update
 RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
 RUN vcpkg.exe integrate install
 
-# Install a recent version of cmake. The version installed with VS is
```
dyastremsky, line 69: We may want to keep this comment. Unless there is a reason to delete it.

mc-nv: I don't believe we need that comment anymore: 
1. ONNXRUNTIME requires 3.24.1 version as minimal as of 1.14.1 release 
2. Comment is irrelevant to me, CMAKE and VSCode are different frameworks where CMAKE actually is using VSCode compiler to build targets according to generated instructions.

Dockerfile.win10.min
mc-nv: Yes, it was removed by mistake

dyastremsky (approved):

krishung5 (approved):

PR #5538, opened 2023-03-22
Update codeql run frequency

dyastremsky:
Increase codeql static analysis run frequency. Run daily on Sunday-Friday, 6pm PST. (Interpreted in UTC: 1am Monday-Sunday).

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -33,7 +33,7 @@ on:
     # The branches below must be a subset of the branches above
     branches: [ 'main' ]
   schedule:
-    - cron: '44 10 * * 1'
+    - cron: '0 1 * * 1-6'
 
 jobs:
   analyze:


dyastremsky:
> Can you explain the rationale? Are there any costs with running more frequently? Too many spam alerts? And why cron style over per pull request like before?

Mostly for consistency. I was looking at GitHub's template that specifies doing it 5x per week and updated the rest to do something similar for our workdays. Reviewing codesql's [documentation](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#configuring-frequency), it looks you're right that scheduled jobs are in addition to the events.

`Scanning code on a schedule informs you about the latest vulnerabilities and errors that GitHub, security researchers, and the community discover, even when developers aren't actively maintaining the repository.`

This costs us nothing as public repositories get unlimited minutes of GitHub Actions for free. We wouldn't get more alerts, the alerts would just be refreshed more often. However, this may be more than we need. If that's the case, then I'd still want to be consistent across all repos. If you'd prefer to keep a weekly job, let me know... I may put you down as a reviewer to update the already-merged PRs, if that's okay with you.

rmccorm4 (approved):

PR #5550, opened 2023-03-24
Add multi-gpu, multi-stream testing for dlpack tensors

oandreeva-nv:
Updating dlpack tests to verify updated functions work as expected.

I've also removed some bool-related unnecessary code:  `pytorch_dtypes` doesn't have `torch.bool` as a member, so I removed comparison on former lines 103-111. I've also updated info regarding PyTorch's DLPack bool support.

Currently, PyTorch does not support DLPack bool, numpy only added DLPack bool support 2 months ago and it is not in any of their latest releases yet. Thus, I only test DLPack bool internally (i.e. PbTensor.from_dlpack(PbTensor)) , to make sure no internal issues raises.

### DIFF
--- a/qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
+++ b/qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
@@ -145,14 +145,14 @@ class ArgumentValidationTest(unittest.TestCase):
             pb_utils.Tensor("OUTPUT0", None)
 
         # Test None as dlpack capsule
-        with self.assertRaises(TypeError) as e:
+        with self.assertRaises(pb_utils.TritonModelException) as e:
             pb_utils.Tensor.from_dlpack("OUTPUT0", None)
 
-        # Test empty string as model name (from_dlpack)
-        with self.assertRaises(TypeError) as e:
+        # Test empty string as tensor name (from_dlpack)
+        with self.assertRaises(pb_utils.TritonModelException) as e:
             pb_utils.Tensor.from_dlpack("", None)
 
-        # Test empty string as model name
+        # Test empty string as tensor name
         with self.assertRaises(TypeError) as e:
             pb_utils.Tensor("", None)
 
--- a/qa/L0_backend_python/unittest/test.sh
+++ b/qa/L0_backend_python/unittest/test.sh
@@ -43,6 +43,9 @@ pip3 uninstall -y torch
 pip3 install torch==1.13.0+cu117 -f https://download.pytorch.org/whl/torch_stable.html
 pip3 install tensorflow
 
+# Install CuPy for testing non_blocking compute streams
+pip3 install cupy-cuda12x
+
 rm -fr *.log ./models
 
 mkdir -p models/dlpack_test/1/
--- a/qa/L0_backend_python/unittest/test.sh
+++ b/qa/L0_backend_python/unittest/test.sh
@@ -32,6 +32,7 @@ CLIENT_LOG="./client.log"
 EXPECTED_NUM_TESTS="1"
 TEST_RESULT_FILE='test_results.txt'
 SERVER_LOG="./inference_server.log"
+export CUDA_VISIBLE_DEVICES=0,1,2,3
 
 RET=0
 rm -fr *.log ./models
@@ -87,4 +88,5 @@ else
     echo -e "\n***\n*** Unittest test PASSED. \n***"
 fi
 
+export CUDA_VISIBLE_DEVICES=0
 exit $RET
--- a/qa/python_models/dlpack_test/model.py
+++ b/qa/python_models/dlpack_test/model.py
@@ -24,6 +24,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import cupy as cp
 import numpy as np
 import unittest
 import torch
@@ -46,15 +47,15 @@ class PBTensorTest(unittest.TestCase):
             pb_tensor = pb_utils.Tensor.from_dlpack('test_tensor',
                                                     dlpack_tensor)
             self.assertTrue(
-                np.all(pb_tensor.as_numpy() == pytorch_tensor.numpy()))
+                np.array_equal(pb_tensor.as_numpy(), pytorch_tensor.numpy()))
 
             # Convert the tensor back to DLPack and ensure that both tensors are
             # the same
             pytorch_tensor_dlpack = from_dlpack(pb_tensor.to_dlpack())
-            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
+            self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))
 
-            self.assertTrue(
-                pytorch_tensor.type() == pytorch_tensor_dlpack.type())
+            self.assertEqual(
+                pytorch_tensor.type(), pytorch_tensor_dlpack.type())
 
             # Now let's check that upgraded DLPack implementation also
             # works as expected, i.e. from_dlpack should work with 
@@ -62,16 +63,16 @@ class PBTensorTest(unittest.TestCase):
 
             pb_tensor_upgraded = pb_utils.Tensor.from_dlpack('test_tensor',
                                                     pytorch_tensor)
-            self.assertTrue(
-                np.all(pb_tensor_upgraded.as_numpy() == pytorch_tensor.numpy()))
+            self.assertTrue(np.array_equal(
+                pb_tensor_upgraded.as_numpy(), pytorch_tensor.numpy()))
 
             # Here we check that `pb_tensor` as a producer, properly
             # invokes `__dlpack__` and `__dlpack_device__`
             pytorch_tensor_dlpack = from_dlpack(pb_tensor_upgraded)
-            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
+            self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))
 
-            self.assertTrue(
-                pytorch_tensor.type() == pytorch_tensor_dlpack.type())  
+            self.assertEqual(
+                pytorch_tensor.type(), pytorch_tensor_dlpack.type())  
 
     def test_non_contiguous_error(self):
         pytorch_tensor = torch.rand([20, 30], dtype=torch.float16)
@@ -118,18 +119,18 @@ class PBTensorTest(unittest.TestCase):
             # Convert the tensor back to DLPack and ensure that both tensors are
             # the same
             pytorch_tensor_dlpack = from_dlpack(pb_tensor.to_dlpack())
-            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
-            self.assertTrue(
-                pytorch_tensor.type() == pytorch_tensor_dlpack.type())
+            self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))
+            self.assertEqual(
+                pytorch_tensor.type(), pytorch_tensor_dlpack.type())
             
             # Now we make sure that updated DLPack implementation works
             # with GPU as well
             pb_tensor = pb_utils.Tensor.from_dlpack('test_tensor',
                                                     pytorch_tensor)
             pytorch_tensor_dlpack = from_dlpack(pb_tensor)
-            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
-            self.assertTrue(
-                pytorch_tensor.type() == pytorch_tensor_dlpack.type())
+            self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))
+            self.assertEqual(
+                pytorch_tensor.type(), pytorch_tensor_dlpack.type())
             
 
     def test_dlpack_gpu_numpy(self):
@@ -141,7 +142,8 @@ class PBTensorTest(unittest.TestCase):
         # Make sure that `__dlpack_device__` works as expected
         self.assertFalse(pb_tensor.is_cpu())
         self.assertTrue(pytorch_tensor.is_cuda)
-        self.assertEqual(pb_tensor.__dlpack_device__(), pytorch_tensor.__dlpack_device__())
+        self.assertEqual(
+            pb_tensor.__dlpack_device__(), pytorch_tensor.__dlpack_device__())
     
         with self.assertRaises(Exception) as e:
             pb_tensor.as_numpy()
@@ -156,11 +158,13 @@ class PBTensorTest(unittest.TestCase):
                                     device='cpu') * 100
         pb_tensor = pb_utils.Tensor.from_dlpack('tensor', pytorch_tensor)
         numpy_tensor_dlpack = np.from_dlpack(pb_tensor)
-        self.assertTrue(np.all(numpy_tensor_dlpack == pytorch_tensor.numpy()))
+        self.assertTrue(np.array_equal(
+            numpy_tensor_dlpack, pytorch_tensor.numpy()))
         # Make sure that `__dlpack_device__` works as expected
         self.assertTrue(pb_tensor.is_cpu())
         self.assertFalse(pytorch_tensor.is_cuda)
-        self.assertEqual(pb_tensor.__dlpack_device__(), pytorch_tensor.__dlpack_device__())
+        self.assertEqual(
+            pb_tensor.__dlpack_device__(), pytorch_tensor.__dlpack_device__())
     
     def test_bool_datatype(self):
         # [FIXME] pass bool_array directly to `pb_utils.Tensor.from_dlpack`,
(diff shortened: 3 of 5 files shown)


qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,50 @@ def test_dlpack_gpu_tensors(self):
...
+        # with numpy
+        pytorch_tensor = torch.rand([100], dtype=torch.float16,
+                                    device='cpu') * 100
+        pb_tensor = pb_utils.Tensor.from_dlpack('tensor', pytorch_tensor)
+        numpy_tensor_dlpack = np.from_dlpack(pb_tensor)
+        self.assertTrue(np.all(numpy_tensor_dlpack == pytorch_tensor.numpy()))
+        # Make sure that `__dlpack_device__` works as expected
+        self.assertTrue(pb_tensor.__dlpack_device__() == pytorch_tensor.__dlpack_device__())
+    
+    def test_pdtensor_bool_internal_support(self):
```
Tabrizian, line 161: Rename the function to "test_bool_datatype".

qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,50 @@ def test_dlpack_gpu_tensors(self):
...
+        bool_array = np.asarray([False, True])
+        bool_tensor = pb_utils.Tensor('tensor', bool_array)
+        bool_tensor_dlpack = pb_utils.Tensor.from_dlpack('tensor', bool_tensor)
```
Tabrizian, line 164: Would it be possible to test passing the numpy object directly in `from_dlpack`? For example,

pb_utils.Tensor.from_dlpack('tensor', bool_array)?

qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,50 @@ def test_dlpack_gpu_tensors(self):
...
+    def test_dlpack_cpu_numpy(self):
+        # Check compatibiity of PbTensor DLPack implementation
+        # with numpy
+        pytorch_tensor = torch.rand([100], dtype=torch.float16,
+                                    device='cpu') * 100
+        pb_tensor = pb_utils.Tensor.from_dlpack('tensor', pytorch_tensor)
+        numpy_tensor_dlpack = np.from_dlpack(pb_tensor)
+        self.assertTrue(np.all(numpy_tensor_dlpack == pytorch_tensor.numpy()))
+        # Make sure that `__dlpack_device__` works as expected
+        self.assertTrue(pb_tensor.__dlpack_device__() == pytorch_tensor.__dlpack_device__())
```
Tabrizian, line 159: I think we can also add another assert for `pb_tenosr.is_cpu()` and `pytorch_tensor.is_cuda()` to make sure.

qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,50 @@ def test_dlpack_gpu_tensors(self):
...
+            
 
     def test_dlpack_gpu_numpy(self):
         # DLPack tesnors that are in GPU cannot be converted to NumPy
         pytorch_tensor = torch.rand([100], dtype=torch.float16,
                                     device='cuda') * 100
         pb_tensor = pb_utils.Tensor.from_dlpack('tensor',
                                                 to_dlpack(pytorch_tensor))
+        # Make sure that `__dlpack_device__` works as expected
+        self.assertTrue(pb_tensor.__dlpack_device__() == pytorch_tensor.__dlpack_device__())
```
Tabrizian, line 142: Same as below.

qa/python_models/dlpack_test/model.py
oandreeva-nv: Not yet. Released Numpy libraries are not supporting bool types by their implementation of DLPack. So for now, if you directly pass `bool_array` to `from_dlpack`, it will cause `TypeError: DLPack only supports signed/unsigned integers, float and complex dtypes.`  Bool support for DLPack was added 2 months ago to master numpy brunch and one of future releases should support that. I can add FIXME to change it later.

qa/python_models/dlpack_test/model.py
oandreeva-nv: renamed in latest commit

qa/python_models/dlpack_test/model.py
oandreeva-nv: added in the latest commit

qa/python_models/dlpack_test/model.py
oandreeva-nv: added in the latest commit

qa/python_models/dlpack_test/model.py
Tabrizian: I see. Thanks.

Tabrizian (dismissed):

qa/python_models/dlpack_test/model.py
```diff
@@ -56,6 +56,23 @@ def test_pytorch_dlpack(self):
...
+
+            pb_tensor_upgraded = pb_utils.Tensor.from_dlpack('test_tensor',
+                                                    pytorch_tensor)
+            self.assertTrue(
+                np.all(pb_tensor_upgraded.as_numpy() == pytorch_tensor.numpy()))
+
+            # Here we check that `pb_tensor` as a producer, properly
+            # invokes `__dlpack__` and `__dlpack_device__`
+            pytorch_tensor_dlpack = from_dlpack(pb_tensor_upgraded)
+            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
```
dyastremsky, line 71: Are you able to make this a bit more easy-to-read with something like `self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))`? This also should skip the computation if the tensors are not the same shape.

qa/python_models/dlpack_test/model.py
```diff
@@ -56,6 +56,23 @@
...
+
+            self.assertTrue(
+                pytorch_tensor.type() == pytorch_tensor_dlpack.type())
```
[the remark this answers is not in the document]
dyastremsky: Please address this and the other notes when you get a chance.

qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,56 @@ def test_dlpack_gpu_tensors(self):
...
-                    pytorch_tensor.type() == pytorch_tensor_dlpack.type())
+            self.assertTrue(
+                pytorch_tensor.type() == pytorch_tensor_dlpack.type())
+            
+            # Now we make sure that updated DLPack implementation works
+            # with GPU as well
+            pb_tensor = pb_utils.Tensor.from_dlpack('test_tensor',
+                                                    pytorch_tensor)
+            pytorch_tensor_dlpack = from_dlpack(pb_tensor)
+            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
```
dyastremsky, line 130: Same comment as above. `self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))` I know all of these coding styles were present before your ticket (e.g. the one on line 121), but cleaning them up here for going forward would be good.

qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,56 @@ def test_dlpack_gpu_tensors(self):
...
+        self.assertFalse(pytorch_tensor.is_cuda)
+        self.assertEqual(pb_tensor.__dlpack_device__(), pytorch_tensor.__dlpack_device__())
+    
+    def test_bool_datatype(self):
+        # [FIXME] pass bool_array directly to `pb_utils.Tensor.from_dlpack`,
+        # when numpy release supports DLPack bool type
+        bool_array = np.asarray([False, True])
+        bool_tensor = pb_utils.Tensor('tensor', bool_array)
+        bool_tensor_dlpack = pb_utils.Tensor.from_dlpack('tensor', bool_tensor)
+        self.assertTrue(np.all(bool_array == bool_tensor_dlpack.as_numpy()))
```
dyastremsky, line 171: For this one, would you be able to do something like `self.assertTrue(np.array_equal(bool_array, bool_tensor_dlpack.as_numpy()))`?

Aside from readability, this also has some extra checks. Specifically, not crashing if one of the arrays is a diferent size and not returning true if one of the arrays is empty.

qa/python_models/dlpack_test/model.py
```diff
@@ -100,27 +119,56 @@ def test_dlpack_gpu_tensors(self):
...
             'Tensor is stored in GPU and cannot be converted to NumPy.')
+    
+    def test_dlpack_cpu_numpy(self):
+        # Check compatibiity of PbTensor DLPack implementation
+        # with numpy
+        pytorch_tensor = torch.rand([100], dtype=torch.float16,
+                                    device='cpu') * 100
+        pb_tensor = pb_utils.Tensor.from_dlpack('tensor', pytorch_tensor)
+        numpy_tensor_dlpack = np.from_dlpack(pb_tensor)
+        self.assertTrue(np.all(numpy_tensor_dlpack == pytorch_tensor.numpy()))
```
dyastremsky, line 159: If applicable, would you be able to do something like self.assertTrue(np.array_equal(numpy_tensor_dlpack, pytorch_tensor.numpy()))?

dyastremsky (commented):
Great work on this ticket! This is going to be very helpful for users.

Added some comments, mostly on the asserts.

Now that we've got code scanning enabled, picking up that we use a lot of " a == b" in our asserts, which we'll probably want to move away from as a team for readability, correctness, and efficiency. If you've got a chance to address these and agree with the comments, please do. At the least, it'd be good to address the ones that came from the static scanner, otherwise those alerts will still be listed in our Security tab waiting to be addressed.

qa/L0_backend_python/unittest/test.sh
```diff
@@ -32,6 +32,7 @@ CLIENT_LOG="./client.log"
 EXPECTED_NUM_TESTS="1"
 TEST_RESULT_FILE='test_results.txt'
 SERVER_LOG="./inference_server.log"
+export CUDA_VISIBLE_DEVICES=0,1,2,3
```
dyastremsky, line 35: Will this work with all devices? Have you confirmed the L0_backend_python_jetson test would still pass?

oandreeva-nv: Pipeline: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/8146001
I've added changes for L0_backend_python to request 4 GPUs for testing, other tests set `CUDA_VISIBLE_DEVICES=0` , for this test I make visible all 4 gpus, and then in the end of the test change to  `CUDA_VISIBLE_DEVICES=0`. L0_backend_python--base green in the pipeline

oandreeva-nv: verifying on Jetson

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
```diff
@@ -145,14 +145,14 @@ def test_tensor_args(self):
...
 
         # Test None as dlpack capsule
-        with self.assertRaises(TypeError) as e:
+        with self.assertRaises(pb_utils.TritonModelException) as e:
             pb_utils.Tensor.from_dlpack("OUTPUT0", None)
 
-        # Test empty string as model name (from_dlpack)
-        with self.assertRaises(TypeError) as e:
+        # Test empty string as tensor name (from_dlpack)
+        with self.assertRaises(pb_utils.TritonModelException) as e:
```
nnshah1, line 152: do we need to test with a proper dlpack (not None) but invalid name?

qa/python_models/dlpack_test/model.py
```diff
@@ -46,15 +47,32 @@ def test_pytorch_dlpack(self):
...
                                                     dlpack_tensor)
             self.assertTrue(
-                np.all(pb_tensor.as_numpy() == pytorch_tensor.numpy()))
+                np.array_equal(pb_tensor.as_numpy(), pytorch_tensor.numpy()))
 
             # Convert the tensor back to DLPack and ensure that both tensors are
             # the same
             pytorch_tensor_dlpack = from_dlpack(pb_tensor.to_dlpack())
-            self.assertTrue(torch.all(pytorch_tensor_dlpack == pytorch_tensor))
+            self.assertTrue(torch.equal(pytorch_tensor_dlpack, pytorch_tensor))
```
nnshah1, line 55: question: Is the goal to ensure the underlying memory is the same (i.e. zero copy) or that the values are the same?

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+        pytorch_tensor = torch.rand([100], dtype=torch.float16,
+                                    device='cpu') * 100
+        pb_tensor = pb_utils.Tensor.from_dlpack('tensor', pytorch_tensor)
+        numpy_tensor_dlpack = np.from_dlpack(pb_tensor)
+        self.assertTrue(np.array_equal(
+            numpy_tensor_dlpack, pytorch_tensor.numpy()))
+        # Make sure that `__dlpack_device__` works as expected
+        self.assertTrue(pb_tensor.is_cpu())
+        self.assertFalse(pytorch_tensor.is_cuda)
+        self.assertEqual(
```
nnshah1, line 166: do we need to try ```pb_tensor.as_numpy()```

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+        with non_blocking_stream:
+            matrix_a = cp.random.rand(size,size)
+            res = cp.matmul(matrix_a,matrix_a)
+            for _ in range(1000):
+                res = cp.matmul(res,matrix_a)
+            cupy_tensor += cp.array([2,2,2,2])
+        
+        pb_tensor = pb_utils.Tensor.from_dlpack('tensor', cupy_tensor)
+        # Verify that non-blocking stream has no pending jobs left
+        self.assertTrue(non_blocking_stream.done)
```
nnshah1, line 212: do we need to wait explicity - or that happends automatically with the `with` statement?

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+        expected_dlpack_device = (2, 1) #  DLDeviceType::kDLCUDA, device_id 1
+        with cp.cuda.Device(1):
+            cupy_tensor = cp.array([0,0,0,0])
+            matrix_a = cp.random.rand(size,size)
+            res = cp.matmul(matrix_a,matrix_a)
+            for _ in range(1000):
+                res = cp.matmul(res,matrix_a)
+            cupy_tensor += cp.array([2,2,2,2])
+        with cp.cuda.Device(0):
+            pb_tensor = pb_utils.Tensor.from_dlpack('tensor', cupy_tensor)
```
nnshah1, line 232: question:  this is moving from device 0 to device 1?

qa/python_models/dlpack_test/model.py
oandreeva-nv: here we ensuring that values are the same

qa/python_models/dlpack_test/model.py
oandreeva-nv: with `with` statement we schedule work on external stream. `matmuls` take its time to do the work, I use it as a way to occupy the stream. Alternatively, I can write a kernel in pycuda, which makes the stream sleep, but I think `matmuls` do the same trick. In L210, we take a cupy tensor, and cupy's tensor's `__dlpack__` method should sync with our default stream through a cuda event - wait event scheme: https://github.com/cupy/cupy/blob/cbc69127e10c0188c17439a47c1ad58a9d32b2e0/cupy/_core/core.pyx#L320-L325

so after we consumed capsule, external stream should be done with computation. L212 asserts that.

qa/python_models/dlpack_test/model.py
oandreeva-nv: I wouldn't use the word `moving`, but rather scheduling work on GPU 1 and GPU 0

qa/L0_backend_python/argument_validation/models/argument_validation/1/model.py
oandreeva-nv: I'll check that

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+    def test_cuda_multi_stream(self):
+        # Test that external stream syncs with the default
+        # and pb_tensor has proper data
+        s1 = torch.cuda.Stream()
+        size = 5000
+        pytorch_tensor = torch.tensor([0,0,0,0], device='cuda')
+        expected_output = torch.tensor([2,2,2,2], device='cuda')
+        with torch.cuda.stream(s1):
+            matrix_a = torch.randn(size,size, device='cuda')
+            res = torch.matmul(matrix_a,matrix_a)
```
Tabrizian, line 187: Is the code formatted using format.py? There should be a space between matrix_a and matrix_a.

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+            for _ in range(1000):
+                res = torch.matmul(res,matrix_a)
+            pytorch_tensor += torch.tensor([2,2,2,2], device='cuda')
+        
+        pb_tensor = pb_utils.Tensor.from_dlpack('tensor', pytorch_tensor)
```
Tabrizian, line 192: does this work properly if we use `pb_utils.Tensor.from_dlpack('tensor', to_dlpack(pytorch_tensor))`?

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+        self.assertTrue(cp.array_equal(cupy_tensor_dlpack, expected_output))
+        self.assertFalse(pb_tensor.is_cpu())
+        self.assertEqual(
+            pb_tensor.__dlpack_device__(), cupy_tensor.__dlpack_device__())
+        
+    def test_cuda_multi_gpu(self):
+        # Test that pb_tensor on different device
+        size = 5000
+        expected_output = cp.array([2,2,2,2])
+        expected_dlpack_device = (2, 1) #  DLDeviceType::kDLCUDA, device_id 1
```
Tabrizian, line 223: small nit: put the comment on top of the line.

qa/python_models/dlpack_test/model.py
```diff
@@ -99,28 +119,172 @@ def test_dlpack_gpu_tensors(self):
...
+        with cp.cuda.Device(1):
+            cupy_tensor = cp.array([0,0,0,0])
+            matrix_a = cp.random.rand(size,size)
+            res = cp.matmul(matrix_a,matrix_a)
+            for _ in range(1000):
+                res = cp.matmul(res,matrix_a)
```
Tabrizian, line 229: Does Cupy block on this line or it schedules the work asynchrounsly?

qa/L0_backend_python/unittest/test.sh
Tabrizian: This test is excluded from jetson since GPU tensors are not supported on jetson devices.

Tabrizian (commented):
Nice testing!

qa/python_models/dlpack_test/model.py
oandreeva-nv: Thanks for catching this, I'll re-format the code

qa/python_models/dlpack_test/model.py
oandreeva-nv: I'll check that

qa/python_models/dlpack_test/model.py
oandreeva-nv: I am not sure what you mean by `block`. I believe the scheduling is always sequential, since it happens on CPU.

This test schedules work on 2 different GPUs, since I don't create separate streams, the work is done on the default stream. I am checking that, when I am trying to consume capsule on GPU 0, ` pb_utils.Tensor.from_dlpack` switches to device 1 and waits for the work to finish on the device (see [here](https://github.com/triton-inference-server/python_backend/pull/223/files#diff-a6a4001d0fb912c1e37fd5d1ebac688a8e793c6b4f657d0e3c6d7de03282daffR371)). Basically the only way to do that (as I see it) is through the Device Sync. In this case, we don't care how the work is scheduled and at what time Device 0 requested the capsule, we still wait for the work to finish. This is a very strict way to do that, which I discussed [here](https://github.com/triton-inference-server/python_backend/pull/223#issuecomment-1531839008). It may happen that we will wait for the work to finish, when it is unnecessary. I've suggested one more addition to the code (a flag to make device sync controllable by the user), I would like to discuss that before though.

Tabrizian: I meant for example if you call `cp.matmul` is it going to schedule the `matmul` on a separate stream and immediately return or wait until the `matmul` is complete.

Tabrizian: I meant for example if you call `cp.matmul` is it going to schedule the `matmul` on a separate stream and immediately return or wait until the `matmul` is complete.

Tabrizian (approved):
small nit: "Add multi-gpu, multi-stream testing for dlpack tensors"

PR #5557, opened 2023-03-27
Adding repo tag to torch.hub.load to fix compatibility issues

oandreeva-nv:
To make sure `ResNet50` model is loaded and compatible with `torchvision=0.14`, that we are using in tests, torch hub needs a proper tag.

### DIFF
--- a/qa/python_models/torchvision/resnet50/model.py
+++ b/qa/python_models/torchvision/resnet50/model.py
@@ -37,7 +37,13 @@ class TritonPythonModel:
         This function initializes pre-trained ResNet50 model.
         """
         self.device = 'cuda' if args["model_instance_kind"] == "GPU" else 'cpu'
-        self.model = torch.hub.load("pytorch/vision:v0.14.1", "resnet50", weights="IMAGENET1K_V2")\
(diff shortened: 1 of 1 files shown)


tanmayv25 (dismissed):

tanmayv25:
What happens when there is an incompatibility in versions? 
Should we document somewhere in the code about this caveat?

oandreeva-nv:
@tanmayv25 In our tests we install torch and torchvision of specific versions (at least those I worked with). If no tag specified, `torch.hub.load` will use the `main` / `master` branch. This may bring (and did with python model instance kind test example) some incompatibility issues, e.g. lack of some attributes etc. By specifying tag, we make sure that compatible version of ResNet50 is loaded.

Do we have a guidance for what torch version to use? I judged based on the fast that torch of version 1.13 was installing and the latest compatible torchvision version is 0.14, that was compatible with that version of torch.

tanmayv25:
I understand. Thanks for the explanation. I would recommend we document this reason for constraint above L40.

Tabrizian (dismissed):

PR #5575, opened 2023-03-31
Add support for forwarding HTTP/GRPC headers

Tabrizian:
- [x] Refactor the GRPC frontend
- [x] Use table printer for printing grpc keep alive options
- [x] Add HTTP/GRPC headers
- [x] Add testing


src/command_line_parser.h
```diff
@@ -127,6 +126,7 @@ struct TritonServerParameters {
       std::max(2u, 2 * std::thread::hardware_concurrency())};
   std::map<int, double> load_gpu_limit_;
 
+
```
dyastremsky, line 129: Nit: accidental extra line?

dyastremsky (commented):
Great changes! A few small comments. Will conduct a full review next week.

src/http_server.h
```diff
@@ -95,7 +97,9 @@ class HTTPMetricsServer : public HTTPServer {
   explicit HTTPMetricsServer(
       const std::shared_ptr<TRITONSERVER_Server>& server, const int32_t port,
       std::string address, const int thread_cnt)
-      : HTTPServer(port, false /* reuse_port */, address, thread_cnt),
+      : HTTPServer(
+            port, false /* reuse_port */, address,
+            "" /* header forward pattern*/, thread_cnt),
```
rmccorm4, line 102:
```suggestion
            "" /* header_forward_pattern */, thread_cnt),
```

src/http_server.cc
```diff
@@ -2765,6 +2765,38 @@ HTTPAPIServer::EVBufferToRawInput(
...
+ForEachHeader(evhtp_header_t* header, void* arg)
+{
+  HeaderSearchPayload* header_search_payload =
+      reinterpret_cast<HeaderSearchPayload*>(arg);
+
+  TRITONSERVER_InferenceRequest* irequest = header_search_payload->request_;
+  const char* pattern = header_search_payload->regex_;
+
+  std::regex base_regex(pattern);
+  if (std::regex_search(header->key, base_regex)) {
```
rmccorm4, line 2784: As I understand it, this will do regex matching on every header in every request. I imagine this could be quite expensive on the critical path. 

Have you done any kind of ballpark benchmarking on what latency this adds? If the latency is significant, would something like a `string.startsWith` be better(and enforcing some prefix/format for forwarded headers)?

src/http_server.cc
```diff
@@ -2765,6 +2765,38 @@ HTTPAPIServer::EVBufferToRawInput(
...
+int
+ForEachHeader(evhtp_header_t* header, void* arg)
+{
+  HeaderSearchPayload* header_search_payload =
+      reinterpret_cast<HeaderSearchPayload*>(arg);
+
+  TRITONSERVER_InferenceRequest* irequest = header_search_payload->request_;
+  const char* pattern = header_search_payload->regex_;
+
+  std::regex base_regex(pattern);
```
rmccorm4, line 2783: Similar to other latency/perf comment, I think this line is expensive (compiling regex). If the regex pattern is the same everytime on server startup (since CLI arg is passed at startup), I think it may be significantly more performant to compile the regex once upfront, and then re-use it.

src/grpc/infer_handler.cc
```diff
@@ -0,0 +1,970 @@
...
+  if (err == nullptr) {
+    err = SetInferenceRequestMetadata(irequest, request);
+  }
+
+  if (err == nullptr && header_forward_pattern_ != "") {
+    const auto& metadata = state->context_->ctx_->client_metadata();
+    for (const auto& pair : metadata) {
+      auto& key = pair.first;
+      auto& value = pair.second;
+      std::regex base_regex(header_forward_pattern_);
```
rmccorm4, line 803: Same as other comment, I think this line is expensive and can be improved by constructing the `std::regex` once and re-using that `std::regex` object

src/http_server.cc
Tabrizian: It is for pre-compiled regex.

Tabrizian: It is for pre-compiled regex.

rmccorm4: What's the cost of forwarding all headers to parameters? Is there a need to only forward specific ones?

qa/L0_parameters/test.sh
```diff
@@ -47,6 +47,10 @@ SERVER_ARGS="--model-repository=model_repository --exit-timeout-secs=120"
 SERVER_LOG="./inference_server.log"
 source ../common/util.sh
 
+# A parameter used by `parameters_test.py` that controls whether the script will
+# test for inclusion of headers in parameters or not.
+export TEST_HEADER=0
```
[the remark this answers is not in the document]
Tabrizian: I think the idea behind unit tests is that the tests should be able to run automatically without external arguments:

https://stackoverflow.com/questions/11380413/python-unittest-passing-arguments

As you can see in the thread above, it is a bit tricky and can interfere with the unittest argument parsing. I think that's why we are not using CLI arguments in this test or other tests.

src/http_server.cc
Tabrizian: Depends on the number of headers and their values. For Python backend, we have to a copy of all of them to pass it through shared memory.

src/http_server.cc
```diff
@@ -2901,6 +2934,18 @@ HTTPAPIServer::HandleInfer(
             infer_request.get());
       }
     }
+    if (err == nullptr && header_forward_pattern_ != "") {
```
GuanLuo, line 2937:
```suggestion
    if (err == nullptr && (!header_forward_pattern_.empty())) {
```

src/http_server.cc
```diff
@@ -2901,6 +2934,18 @@ HTTPAPIServer::HandleInfer(
...
+      HeaderSearchPayload header_search_payload;
+      header_search_payload.error_ = nullptr;
+      header_search_payload.regex_ = header_forward_pattern_.c_str();
```
GuanLuo, line 2940: Use constructor for the struct? Then you can have
```
HeaderSearchPayload(const std::string& regex, TritonRequest* request)
```
And the struct can directly store `const std::string& regex_;`

src/grpc/stream_infer_handler.cc
```diff
@@ -163,6 +236,24 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
+          std::string param_key = std::string(key.begin(), key.end());
+          std::string param_value = std::string(value.begin(), value.end());
+          err = TRITONSERVER_InferenceRequestSetStringParameter(
+              irequest, param_key.c_str(), param_value.c_str());
+          if (err != nullptr) {
+            break;
+          }
+        }
+      }
+    }
```
GuanLuo, line 255: Can be a helper function in InferHandler?

src/command_line_parser.h
Tabrizian: Fixed.

src/http_server.cc
```diff
@@ -2765,6 +2765,38 @@ HTTPAPIServer::EVBufferToRawInput(
...
+  const char* pattern = header_search_payload->regex_;
+
+  std::regex base_regex(pattern);
+  if (std::regex_search(header->key, base_regex)) {
+    std::cout << "HTTP Regex matched" << std::endl;
+    std::cout << "Header is " << (header->key) << std::endl;
+    header_search_payload->error_ =
+        TRITONSERVER_InferenceRequestSetStringParameter(
+            irequest, header->key, header->val);
+    std::cout << "parameter was set" << std::endl;
```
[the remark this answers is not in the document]
Tabrizian: Fixed.

src/http_server.cc
```diff
@@ -2765,6 +2765,38 @@ HTTPAPIServer::EVBufferToRawInput(
...
+
+    if (header_search_payload->error_ != nullptr) {
```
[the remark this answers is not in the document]
Tabrizian: The error will be deleted later. See http_server:2994

src/http_server.h
Tabrizian: Fixed.

src/http_server.cc
Tabrizian: Changed to prefix string matching 👍

src/http_server.cc
Tabrizian: Changed to prefix string matching 👍

src/grpc/infer_handler.cc
Tabrizian: Changed to prefix string matching 👍

src/http_server.cc
Tabrizian: Fixed.

src/grpc/stream_infer_handler.cc
Tabrizian: Fixed.

qa/L0_parameters/test.sh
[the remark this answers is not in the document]
dyastremsky: Makes sense, thanks for explaining the rationale.

src/http_server.cc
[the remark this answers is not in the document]
rmccorm4: I see, thanks!

src/http_server.cc
```diff
@@ -2765,6 +2765,41 @@ HTTPAPIServer::EVBufferToRawInput(
...
+ForEachHeader(evhtp_header_t* header, void* arg)
+{
+  HeaderSearchPayload* header_search_payload =
+      reinterpret_cast<HeaderSearchPayload*>(arg);
+
+  TRITONSERVER_InferenceRequest* request = header_search_payload->request_;
+  const std::string& prefix = header_search_payload->prefix_;
+
+  // If the prefix is '*' forward all the headers.
+  if (prefix == "*" || (std::string(header->key).find(prefix) == 0)) {
```
rmccorm4, line 2790: TBD if `re2` is suitable per offline conversation, but just a nit: `find(prefix)` won't actually enforce that this is a prefix. It may match in the middle of the string. I don't think that's an issue necessarily, just pointing it out if we want to document the behavior explicitly, or just change to startsWith() or similar.

Tabrizian: C++ doesn't have `startswith`.

> find(prefix) won't actually enforce that this is a prefix.

find by itself won't guarantee that but I made sure that the position returned by `find` is equal to zero (i.e., it is a prefix).

rmccorm4: Ah I see, misread. Thanks!

qa/L0_parameters/test.sh
```diff
@@ -73,7 +73,28 @@ kill $SERVER_PID
...
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+set -e
+
+kill $SERVER_PID
+wait $SERVER_PID
+
+export TEST_HEADER=all
```
GuanLuo, line 96: If only used by the unit test script, perhaps put it as scoped environment variable:
```
TEST_HEADER=all python3 $TEST_SCRIPT_PY >$CLIENT_LOG 2>&1
```
You may also consider to represent them to be different test cases:
```
class CommonTestUtils(tu.TestResultCollector):
    ...

class ForwardPrefixTest(CommonTestUtils):
    ...

class ForwardAllTest(CommonTestUtils):
    ...
```
And invoke the desired test case directly:
```
python3 $TEST_SCRIPT_PY ForwardAllTest >$CLIENT_LOG 2>&1
```

tanmayv25:
@Tabrizian Can we create directory for other server implementations and move the files there? `http`, `sagemaker`, `vertex_ai`? No need to refactor `http` just yet but lets organize the files in a consistent way?

qa/L0_parameters/parameters_test.py
```diff
@@ -76,9 +77,29 @@ async def asyncSetUp(self):
...
+            # Only these headers should be forwarded to the model.
+            # The value of these headers will be ignored.
+            self.expected_http_headers = {
+                'Content-Length': '',
+                'Accept-Encoding': '',
+                'Host': '',
+                'Content-Type': '',
+                'User-Agent': '',
+                'Inference-Header-Content-Length': '',
+                'Accept': ''
```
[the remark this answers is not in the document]
Tabrizian: I removed this test. It was only added for the `pattern=*` which was removed as a result of moving back to regex matching.

qa/L0_parameters/test.sh
Tabrizian: This test is removed.

Tabrizian:
@tanmayv25 I can create a folder for frontends and put all the server implementations there. The reason I created a folder for `grpc` was that there were many files so I wanted to keep them in one place. For other server implementations, we are only using a single file so I think it should be fine for now? We can move them to their own folder whenever they are broken into multiple files.

tanmayv25:
I understand. That's fine for now. But I would prefer having consistent logical hierarchy among the source files/directories. All endpoints implementation in their own directories.

qa/L0_parameters/test.sh
```diff
@@ -62,12 +67,48 @@ if [ $? -ne 0 ]; then
...
+        echo -e "\n***\n*** Test Result Verification Failed\n***"
+        RET=1
+    fi
 fi
 set -e
 
 kill $SERVER_PID
 wait $SERVER_PID
 
+export TEST_HEADER=1
```
tanmayv25, line 83: To avoid code duplication, can we loop through the values of TEST_HEADER 0 and 1. 
If TEST_HEADER is 1 then set the extra SERVER_ARGS.

src/command_line_parser.cc
```diff
@@ -1060,6 +1072,10 @@ TritonParser::Parse(int argc, char** argv)
         break;
       case OPTION_HTTP_ADDRESS:
         lparams.http_address_ = optarg;
+        break;
+      case OPTION_HTTP_HEADER_FORWARD_PREFIX:
+        lparams.http_forward_header_pattern_ = optarg;
```
tanmayv25, line 1077: Can we create lhttp_options struct like for grpc_option below in L1207?

qa/L0_parameters/parameters_test.py
```diff
@@ -41,6 +41,8 @@
 import unittest
 import queue
 from functools import partial
+import os
+TEST_HEADER = os.environ.get('TEST_HEADER')
 
 
 class InferenceParametersTest(IsolatedAsyncioTestCase):
```
tanmayv25, line 48: We should probably add similar tests for our C++ client libraries.

src/command_line_parser.cc
```diff
@@ -380,6 +382,11 @@ std::vector<Option> TritonParser::recognized_options_
        "Allow the server to listen for HTTP requests."},
       {OPTION_HTTP_PORT, "http-port", Option::ArgInt,
        "The port for the server to listen on for HTTP requests."},
+      {OPTION_HTTP_HEADER_FORWARD_PREFIX, "http-header-forward-prefix",
```
rmccorm4, line 385: same as grpc flag comment, revert back to pattern/regex arg name and description

rmccorm4 (commented):
LGTM other than renaming CLI args/descriptions and Tanmay's comment

src/command_line_parser.cc
Tabrizian: I think it can be but it requires changes in vertex and others. I have created a ticket for this refactor: https://jirasw.nvidia.com/browse/DLIS-4758

qa/L0_parameters/parameters_test.py
Tabrizian: Added a TODO: https://jirasw.nvidia.com/browse/DLIS-4673

rmccorm4 (approved):
LGTM with the follow-up tickets

tanmayv25 (dismissed):

qa/L0_parameters/parameters_test.py
```diff
@@ -148,7 +177,8 @@ def test_http_async_parameter(self):
         for parameters in parameter_list:
             result = self.http.async_infer(model_name='parameter',
                                            inputs=inputs,
-                                           parameters=parameters).get_result()
+                                           parameters=parameters,
+                                           headers=self.headers).get_result()
```
GuanLuo, line 181: Just realize that you are testing both parameter and header together, does it make sense to break into two test cases?
```
def test_xxx_parameter(self):
    ...
def test_xxx_parameter_in_header(self):
    ...
```

Tabrizian: I think it makes sense but we have 8 unit tests. I would have to create header variant for all of them. I think this results in more code duplication so I'd rather to keep the current structure.

GuanLuo (approved):

PR #5581, opened 2023-03-31
Add metrics config CLI

rmccorm4:
Preliminary changes to expose config settings. These will be used to opt-in and toggle future metrics features.

Below are just rough ideas of things that would be configured based on user feedback, not the exact set of args we will necessarily support. We can roll out functionality in stages based on feedback and priorities.

Example (settings/values TBD): 
```
# Use summaries instead of counters for all built-in latency related metrics
tritonserver --metrics-config summaries=all 
# other alternatives for the same thing
tritonserver --metrics-config summaries=true
tritonserver --metrics-config summaries=*   (similar to --load-model support)

# specify custom quantiles for summary metrics
tritonserver \
  --metrics-config summaries=all
  --metrics-config summary_quantiles=0.5, 0.75, 0.9, 0.95, 0.99, 0.999

# only use summaries for compute metrics, counters otherwise, etc.
tritonserver --metrics-config summaries=compute

# specify histogram buckets across all histograms. possible ability to override per-model in model configs
tritonserver \
  --metrics-config histograms=all 
  --metrics-config histogram_buckets=0.01, 0.1, 1, 10, 100, 1000

# specify metric name prefixes, ex: omniverse
tritonserver \
  --metrics-config metrics-prefix="my_product_prefix_"

# specify additional labels, ex: omniverse
tritonserver \
  --metrics-config metrics-labels="key:value"
```

Core PR: https://github.com/triton-inference-server/core/pull/181

### DIFF
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -887,7 +887,7 @@ TritonServerParameters::BuildTritonServerOptions()
         TRITONSERVER_ServerOptionsSetMetricsConfig(
             loptions, std::get<0>(mcs).c_str(), std::get<1>(mcs).c_str(),
             std::get<2>(mcs).c_str()),
-        "setting metrics configurtion");
+        "setting metrics configuration");
   }
 
 #endif  // TRITON_ENABLE_METRICS
@@ -915,7 +915,7 @@ TritonServerParameters::BuildTritonServerOptions()
           ParseException,
           TRITONSERVER_ServerOptionsSetCacheConfig(
               loptions, cache_name.c_str(), json_config_str.c_str()),
-          "setting cache configurtion");
+          "setting cache configuration");
     }
   }
 
@@ -930,7 +930,7 @@ TritonServerParameters::BuildTritonServerOptions()
         TRITONSERVER_ServerOptionsSetBackendConfig(
             loptions, std::get<0>(bcs).c_str(), std::get<1>(bcs).c_str(),
             std::get<2>(bcs).c_str()),
-        "setting backend configurtion");
+        "setting backend configuration");
   }
   for (const auto& limit : load_gpu_limit_) {
     THROW_IF_ERR(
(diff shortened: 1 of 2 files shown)


Tabrizian (approved):

src/command_line_parser.cc
```diff
@@ -1443,6 +1461,43 @@ TritonParser::Usage()
...
 }
 
+std::tuple<std::string, std::string, std::string>
+TritonParser::ParseMetricsConfigOption(const std::string& arg)
+{
+  // Format is "<setting>=<value>" for generic configs/settings
+  int delim_name = arg.find(",");
+  int delim_setting = arg.find("=", delim_name + 1);
+
+  // No name-specific configs currently supported, though it may be in
```
tanmayv25, line 1471: When would be name of the configuration group be useful?

rmccorm4: I don't have any specific ideas right now, but wanted to keep the format available in case. 

One general idea I can see coming in the near future is supporting OpenTelemetry vs Prometheus directly. In the event we keep old behavior and add OTLP on top of that, maybe you would configure each like:
```
--metrics-config otlp,key=value
--metrics-config prometheus,key=value
```

Or maybe groups of metrics could be configured differently:
```
--metrics-config compute,summaries=true
--metrics-config cache,histograms=true
```

Configuration groups aren't sorted out by any means, but I can see them having value as there have been a wide range of metrics related requests recently.

tanmayv25 (approved):

src/command_line_parser.cc
```diff
@@ -507,6 +508,10 @@ std::vector<Option> TritonParser::recognized_options_
       {OPTION_METRICS_INTERVAL_MS, "metrics-interval-ms", Option::ArgFloat,
        "Metrics will be collected once every <metrics-interval-ms> "
        "milliseconds. Default is 2000 milliseconds."},
+      {OPTION_METRICS_CONFIG, "metrics-config", "<string>=<string>",
+       "Specify a metrics-specific configuration setting. The format of this "
+       "flag is --metrics-config=<setting>=<value>. It can be specified "
```
nv-kmcgill53, line 513: How will the CLI setting look going forward? Will `--metrics-config=<setting>=<value>` set global configs when the name-specific metrics are enabled? I want to make sure the API stays consistent in this case.

rmccorm4: Same as backend-config and cache-config:
- `--metrics-config setting=value` will set global configs for `name=""` (only supported format currently)
- `--metrics-config name,setting=value` will set name-specific configs for `name=name` if eventually added back in, but currently is rejected by CLI

src/command_line_parser.cc
tanmayv25: @oandreeva-nv  for viz if openTelemetery Tracing settings
we must have cli options like:

```
--tracing-config <name>,<setting key>=<setting value>
```

PR #5597, opened 2023-04-05
Fix L0_backend_python lifecycle test

krishung5:
Fixed the check for the error message as the returned error message starts with some error code, `[400] Failed to process the request(s)...`, which failed the check `str(e.exception).startswith`


Tabrizian (approved):

tanmayv25 (approved):

rmccorm4 (approved):
FYI there is a built-in `assertIn` instead of `assertTrue`, but this is fine.

PR #5600, opened 2023-04-05
Generate qa models using tf2 image

krishung5:
This PR updated the model generation scripts to make it compatible with TensorFlow 2


tanmayv25 (approved):

tanmayv25:
LGTM but we might be able to avoid tf.compat.v1 since we are using the models only with TF2.

krishung5:
@tanmayv25 I agree. We can generate the models using `Keras` for TF2 to replace `placeholder`. I gave it a try and had some trouble when trying to store the model as `graphdef` format. Given the short timeframe, I didn't want to spend too much time on that so we could start testing the newly generated models. I have filed an enhancement ticket and maybe we can revisit later.

PR #5625, opened 2023-04-12
Define NOMINMAX when compiling triton server

tanmayv25:


dyastremsky (approved):
Great work! Approving. It'd be good to remove the other NOMINMAX definitions in server. And apply the same fix to other repos and remove their NOMINMAX definitions.

nv-kmcgill53 (approved):

Issue #5610, opened 2023-04-07
TYPE_STRING datatype is throwing error while using tritonserver/Python backend/HTTP endpoint for Linux-s390x


pradghos:
**Description**
`TYPE_STRING` is throwing error while using tritonserver/Python backend via HTTP endpoint.

**Triton Information**
What version of Triton are you using?
`23.01`

Are you using the Triton container or did you build it yourself?
We have built it for `Linux-s390x` (Big Endian)

**To Reproduce**
Steps to reproduce the behavior.

- Build and install tritonserver and python_backend on Linux-s390x.
- Start tritonserver with simple model https://github.com/triton-inference-server/server/tree/main/qa/python_models/string
where input datatype `TYPE_STRING`. model returns the same string which has been sent as input via HTTP endpoint. 

Payload:
```python
import requests
import numpy as np
import time
import argparse


CLI = argparse.ArgumentParser()
CLI.add_argument("--model", type=str, default="string")
args = CLI.parse_args()
jason_data = {
  "inputs" : [
    {
      "name" : "INPUT0",
      "datatype" : "BYTES",
      "shape" : [1],
      "data" : ["abc"]
    }
  ],
  "outputs" : [
    {
      "name" : "OUTPUT0"
    }
  ]
}
inference_endpoint = "http://localhost:8000/v2/models/"+args.model+"/infer"
response = requests.post(inference_endpoint, json=jason_data)
print(response.text)
``` 

Output: 
```python
{"error":"Failed to process the request(s) for model instance 'string_0', message: error: unpack_from requires a buffer of at least 8 bytes for unpacking 4 bytes at offset 4 (actual buffer size is 7)\n\nAt:\n  /opt/tritonserver/backends/python/triton_python_backend_utils.py(115): deserialize_bytes_tensor\n"}
```
**Expected behavior**
It should return the same string as output. 
i.e. 
```python
{"model_name":"string","model_version":"1","outputs":[{"name":"OUTPUT0","datatype":"BYTES","shape":[1],"data":["abc"]}]}
```

rmccorm4:
Hi @pradghos,

I believe special padding/handling is required when using the REST API directly with STRING/BYTES types or binary data: https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_binary_data.md

[...]

PR #5626, opened 2023-04-12
Fix for TYPE_STRING datatype error on big endian system

pradghos:
- Fixing https://github.com/triton-inference-server/server/issues/5610
- This fix is to ensure that string length should getting padded always in little endian format while preparing for input data.

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -409,8 +409,13 @@ ReadDataFromJsonHelper(
                 TRITONSERVER_ERROR_INTERNAL,
                 "Shape does not match true shape of 'data' field");
           }
+#if defined(__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
+          size_t cstr_len = __builtin_bswap64(len);
+#else
+          size_t cstr_len = len;
+#endif
           memcpy(
-              base + *counter, reinterpret_cast<char*>(&len), sizeof(uint32_t));
+              base + *counter, reinterpret_cast<char*>(&cstr_len), sizeof(uint32_t));
           std::copy(cstr, cstr + len, base + *counter + sizeof(uint32_t));
           *counter += len + sizeof(uint32_t);
           break;
@@ -645,10 +650,16 @@ WriteDataToJson(
                   .c_str());
         }
 
-        const size_t len = *(reinterpret_cast<const uint32_t*>(cbase + offset));
+        const uint32_t len = *(reinterpret_cast< const uint32_t*>(cbase + offset));
+
+#if defined(__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
+        uint32_t str_len = __builtin_bswap32(len);
+#else
+        uint32_t str_len = len;
+#endif
         offset += sizeof(uint32_t);
 
-        if ((offset + len) > byte_size) {
+        if ((offset + str_len) > byte_size) {
           return TRITONSERVER_ErrorNew(
               TRITONSERVER_ERROR_INTERNAL,
               std::string(
@@ -660,8 +671,8 @@ WriteDataToJson(
         // Can use stringref because 'base' buffer is not deleted
         // until response is deleted and that happens after this json
         // is serialized.
-        RETURN_IF_ERR(data_json->AppendStringRef(cbase + offset, len));
-        offset += len;
+        RETURN_IF_ERR(data_json->AppendStringRef(cbase + offset, str_len));
+        offset += str_len;
       }
       break;
     }


pradghos:
Hi @rmccorm4 @dyastremsky @Tabrizian, Really appreciate your review/suggestions for this PR. Thanks a lot !

pradghos:
Please do note that https://github.com/triton-inference-server/server/blob/main/Triton-CCLA-v1.pdf has been submitted already. Thanks !

pradghos:
Hi All, It would be great to know your review comments/approval for this PR. Thanks in advance !

pradghos:
@dyastremsky  Thanks for the response ! Issue can be easily re-produced in Linux-s390x(big endian) with Triton Inference server with Python Backend (with or without tritonclient). 

As I have mentioned in scenario(issue#5610), when user is sending Json object (non binary data) with TYPE_STRING datatype, and string length is prefixed to payload(data) at server side.  Fix is to ensure that string length is always in little endian format that is what python backend would expect while calling `serialize_byte_tensor` and `deserialize_bytes_tensor` (ref: "/opt/tritonserver/backends/python/triton_python_backend_utils.py"). 

Fix will not impact `binary_data=True` (Ref :https://github.com/triton-inference-server/server/issues/5610#issuecomment-1503264010) and it will continue to work as before. Since, tritonclient is already padding string length in little endian format at client side itself. 

However, directly using REST API for Jason object with datatype TYPE_STRING is important use case and It is working fine for Little endian platform. This patch will make sure the functionality works fine in big-endian platform also. 

Please let me know if I need to test any suggested scenario. Thanks again !

pradghos:
Please let me know if you have any further comments / suggestions for this PR. Thank you !

GuanLuo:
Sorry for the late response, I think this change is reasonable because it defines how the prepended bytes should be interpreted to get the data content length. We only defined that a 4-byte integer of the length must be prepended, but the endian is not well-defined.

I think fixing it to always be little-endian and explicitly mention it in the byte type documentation is the correct direction forward, and we do need to apply the same re-ordering to other backends to make sure all components are aligned. @Tabrizian @tanmayv25 @nnshah1 , do you have any concern?

nnshah1:
Question: @Tabrizian , Won't the python backend have trouble with all number types serialized on a big endian system (not just string / bytes length)?

Also - will other backends need to be updated? If python backend is the only one incorrectly deserializing / serializing on big endian systems - then perhaps the change needs to be in the python backend (is python backend the only one that interprets it as little endian independent of the host?)

GuanLuo:
@nnshah1 I wouldn't worry much on the endianness of other types, because for other should happen within the same machine which implies the endianness is consistent.

As I state that the issue is that we define the length integer is included in the binary (/ serialized) data and didn't further define the endianness of the integer and leave ambiguity on how to serialize / deserialize the data. This PR is to address where the ambiguity hits: Triton server uses the endianness that the system uses and Python backend / Triton client uses little endianness.

To eliminate the ambiguity, we should be explicit on what endian the integer should be serialized into. I think the fact that we allow client to send raw binary data directly requires us to do so rather than letting the code follows the system's endianness, the client can be on little endian system and send request to server on big endian system, and vice versa.

pradghos:
@Tabrizian: I think reported issue can be reproduced only when triton server is running on big endian system with the test scenario (python backend, TYPE_STRING, binary_data=False). In case of `binary_data=False`, string length is fecthed and prefixed to payload(data) at server side. Thanks!

pradghos:
@Tabrizian @GuanLuo @nnshah1 Please let me know if you have any other comments for this PR or can we get this PR merged. Thanks!

nnshah1:
> @Tabrizian @GuanLuo @nnshah1 Please let me know if you have any other comments for this PR or can we get this PR merged. Thanks!

@pradghos Based on further investigation and analysis of the original issue details you provided - we're looking into all the changes required to support big endian servers / clients. I'm currently testing updates. For the specific issue of python backend and TYPE_STRING - current plan is to modify the backend to support the native machine endianness to match behavior of other backends. I'll open a few PRs based on this later this week. As there will be other changes needed to target GRPC / Binary encoding I believe we can do this in stages to resolve this issue first. Thanks for your patience.

pradghos:
> > @Tabrizian @GuanLuo @nnshah1 Please let me know if you have any other comments for this PR or can we get this PR merged. Thanks!
> 
> @pradghos Based on further investigation and analysis of the original issue details you provided - we're looking into all the changes required to support big endian servers / clients. I'm currently testing updates. For the specific issue of python backend and TYPE_STRING - current plan is to modify the backend to support the native machine endianness to match behavior of other backends. I'll open a few PRs based on this later this week. As there will be other changes needed to target GRPC / Binary encoding I believe we can do this in stages to resolve this issue first. Thanks for your patience.

@nnshah1 Any further update on this issue ? Thank you very much !

nnshah1:
> > > @Tabrizian @GuanLuo @nnshah1 Please let me know if you have any other comments for this PR or can we get this PR merged. Thanks!
> > 
> > 
> > @pradghos Based on further investigation and analysis of the original issue details you provided - we're looking into all the changes required to support big endian servers / clients. I'm currently testing updates. For the specific issue of python backend and TYPE_STRING - current plan is to modify the backend to support the native machine endianness to match behavior of other backends. I'll open a few PRs based on this later this week. As there will be other changes needed to target GRPC / Binary encoding I believe we can do this in stages to resolve this issue first. Thanks for your patience.
> 
> @nnshah1 Any further update on this issue ? Thank you very much !

@pradghos 

We decided on the following design / clarifications:

1) All data transmitted as binary (i.e. not using JSON or grpc typed fields) would be transmitted in little endian
2) All data transmitted between the frontend and the core and backends would be in host endian 
3) The frontend serve is responsible for converting from little endian to host endian where required

We have started a branch with changes for the server and the python backend with unit tests:

1) https://github.com/triton-inference-server/server/pull/6043/files
2) https://github.com/triton-inference-server/python_backend/pull/269/files

While the changes passed local end to end testing - due to other urgent priorities we were not able to put a full end to end test in CI / CD and thus haven't merged the changes in yet.

Any feedback on the direction is welcome and any help on developing a way for end to end testing is welcome as well - current thought is to use qemu docker emulation to emujlate a big endian host.

potula-chandra:
@nnshah1 Thank you very much for the support and new design changes. We will have a look at the changes and share our feedback very soon. Thank you !

pradghos:
@nnshah1: Sorry for the delayed response and this is great progress indeed. Just curious to know what CI might work for you. there is nice blog of different CI support on IBM Z builds. https://developer.ibm.com/blogs/developer-resources-for-building-your-open-source-app-for-linux-on-ibm-z-and-linuxone/. Please let me know and looking forward to collaborate further on this topic. Thanks!

pradghos:
> @nnshah1: Sorry for the delayed response and this is great progress indeed. Just curious to know what CI might work for you. there is nice blog of different CI support on IBM Z builds. https://developer.ibm.com/blogs/developer-resources-for-building-your-open-source-app-for-linux-on-ibm-z-and-linuxone/. Please let me know and looking forward to collaborate further on this topic. Thanks!

@nnshah1 Please let me know your thought on this. Thanks!

Kanupriyagoyal:
Hi @dyastremsky @nnshah1 
**FYI**
After Inclusion of the patches
```
user@0a38189eb5ce:~$ /opt/tritonserver/bin/tritonserver 
Failed to build Triton option:
error: creating server options:  -
``` 
Triton server is failing to start.
After that we haven't include the unit testcases file (endian_conversion_test.cc) and src/test/CmakeList.txt changes from the patch, able to run the triton server.
```I0105 14:37:20.674579 1 server.cc:631] 
-------------------------+
| Backend  | Path                                                      | Config                                                                                                                                                        |
+----------+-----------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------+
| python   | /opt/tritonserver/backends/python/libtriton_python.so     | {"cmdline":{"auto-complete-config":"true","backend-directory":"/opt/tritonserver/backends","min-compute-capability":"0.000000","default-max-batch-size":"4"}} |
I0105 14:37:20.674605 1 server.cc:674] 
+-----------+---------+--------+
| Model     | Version | Status |
+-----------+---------+--------+
| model_1 | 1       | READY  |
```
But still when we pass the **DATATYPE** as **STRING** as "**big_endian**" Failed with below message:
```FAILED : model_infer 
Error Message:  Failed to process the request(s) for model instance 'model_equals_b_bytes_0', message: error: unpack_from requires a buffer of at least 50331652 bytes for unpacking 50331648 bytes at offset 4 (actual buffer size is 7)

At:
  /opt/tritonserver/backends/python/triton_python_backend_utils.py(117): deserialize_bytes_tensor
 
Error Code:  StatusCode.INTERNAL \Debug Message:  UNKNOWN:Error received from peer  {grpc_message:"Failed to process the request(s) for model instance \'model_equals_b_bytes_0\', message: error: unpack_from requires a buffer of at least 50331652 bytes for unpacking 50331648 bytes at offset 4 (actual buffer size is 7)\n\nAt:\n  /opt/tritonserver/backends/python/triton_python_backend_utils.py(117): deserialize_bytes_tensor\n", grpc_status:13, created_time:"2024-01-24T09:34:26.548123464-05:00"}
```

**Triton Information**
What version of Triton are you using?
r23.07

Irrespective of the server instance big_endian or little_endian should be able to infer for STRING datatype. https://github.com/triton-inference-server/server/issues/5811
https://github.com/triton-inference-server/server/pull/5626#issuecomment-1598839165

PR #5630, opened 2023-04-12
Modify build.py to include tests in backend repo into CI

GuanLuo:


build.py
```diff
@@ -1714,6 +1714,9 @@ def cibase_build(cmake_script, repo_dir, cmake_dir, build_dir, install_dir,
         cmake_script.cp(
             os.path.join(ort_install_dir, 'test', 'custom_op_test.onnx'),
             os.path.join(ci_dir, 'qa', 'L0_custom_ops'))
+        # [WIP] other way than wildcard?
+        backend_tests = os.path.join(build_dir, 'onnxruntime', 'test', "*")
+        cmake_script.cpdir(backend_tests, os.path.join(ci_dir, 'qa'))
```
rmccorm4, line 1719: Can we rename test to L0_initializer_as_input for consistency with other tests?

GuanLuo: We can but I will want to explore something different since we are adding "external" tests. The way that we are using `L0_` prefix doesn't add much information except that "this is a test folder", which has been indicated in many other ways, so I want to make the naming concise and add it back if it is necessary.

The "backend_name -> test_name" hierarchy is captured through pipeline configuration (child pipeline). Ideally such grouping should be captured in the QA image but I keep it flatten as the test utilities interact with each other with relative path and assumed flat structure in QA folder..

rmccorm4: Agreed that L0 doesn't add much info, I'm open to changing or refactoring that moving forward. However I do like the distinction that the L0_* are tests and anything else in the qa/ folder (common, models, etc.) are utilities. 

If we want to remove or rename the L0 part, I think we should either put the tests into their own folder, or put the non-tests into their own folder to keep it grouped well.

Either way, don't need to block the merge/cherry-pick on this. I just want to keep the tests and utilities logically separated in some way as we move forward

tanmayv25: We can discuss more about the test structure layout. I find L0_python_backend tests better. I also believe as most of our tests are run only nightly, or weekly. It would make sense to have L0_*(each commit), L1_*(nightly) and L2_*(weekly).

tanmayv25 (approved):

PR #5637, opened 2023-04-13
Opentelemetry trace support for Triton tracing

oandreeva-nv:
Adding the limited support of Opentelemetry APIs for tracing in triton.

Major Changes:
 - Build with Opentelemetry. Currently there are issues with Windows, so code excludes opentelemetry build and use on Windows
 - TraceManager::Trace has new Opentelemetry related attributes: exporter, processor, provider and span. Exporter and Processor are in charge of exporting collected traces to either default or specified endpoint. This PR adds support for only  OTLP HTTP exporter and SimpleSpanProcessor, which sends every collected span to the endpoint. `provider` and `span`, are in charge of collecting and generating traces.
 - CLI for tracing: new flag is introduced `--trace-config=<<mode>,<setting>=<value>>
    ```
    Specifies global or trace mode specific configuration setting.
	The format of this flag is --tracing-config=<trace
	mode>,<setting>=<value>. Where <trace mode> is either "triton" or "opentelemetry".
	The default is "triton". To specify global trace settings (level,
	rate, count, or mode), the format would be
	--tracing-config=<setting>=<value>. For "triton", the server will use the Triton's trace APIs.
	For "opentelemetry", the server will use the OpenTelemetry's APIs to
	generate, collect and export traces for individual inference
	requests.
```
- Docs are still in process, something may be added
- Tests are on the way
- Opentelemetry tracing does not support bls models and Tensor level tracing

### DIFF
--- a/docs/user_guide/trace.md
+++ b/docs/user_guide/trace.md
@@ -30,12 +30,56 @@
 
 Triton includes that capability to generate a detailed trace for
 individual inference requests. Tracing is enable by command-line
-arguments when running the tritonserver executable. For example,
+arguments when running the tritonserver executable.
+
+Starting in 23.04 release, Triton has one flag `--trace-config` to specify 
+global or trace mode specific configuration setting. The format of this flag 
+is `--tracing-config=<trace mode>,<setting>=<value>`, where `<trace mode>` 
+is either `triton` or `opentelemetry`. By default, trace mode is set to `triton`,
+and the server will use the Triton's trace APIs. For `opentelemetry`, 
+the server will use the [OpenTelemetry's APIs](#opentelemetry-trace-support) to generate, collect and export 
+traces for individual inference requests.
+
+To specify global trace settings (level, rate, count, or mode), 
+the format is `--tracing-config=<setting>=<value>`.
+
+An example ussage, which invokes Triton's trace APIs:
 
 ```
-$ tritonserver --trace-file=/tmp/trace.json --trace-rate=100 \
-    --trace-level=TIMESTAMPS --trace-log-frequency=50 --trace-count=100 ...
+$ tritonserver --trace-config=triton,file=/tmp/trace.json \
+    --trace-config=triton,log-frequency=50 --trace-config=rate=100 \
+    --trace-config=level=TIMESTAMPS --trace-config=triton,log-frequency=50 \
+    --trace-config=count=100 ...
 ```
+
+## Trace settings
+### Global settings
+The following table shows available global trace settings to pass to `--trace-config`
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `rate`      | 1000          | Specifies the sampling rate. The same as deprecated `--trace-rate`. In the above example every 100-th inference request will be traced.|
+| `level`     | OFF           | Indicates the level of trace detail that should be collected and may be specified multiple times to trace multiple informations. The same as deprecated `--trace-level`|
+| `count`     | -1            | Specifies the remaining number of traces to be collected. In the example Triton will stop tracing requests after 100 traces are collected. The same as deprecated `--trace-count`. |
+| `mode`      | triton        | Specifies which trace APIs to use for collecting traces. |
+
+### Triton trace APIs settings
+
+The following table shows available Triton trace APIs settings for `--trace-config=triton,<setting>=<value>`.
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `file`      |     -         | Indicates where the trace output should be written. The same as deprecated `--trace-file`.|
+| `log-frequency`| 0          | Specifies the rate that the traces are written to file. In the example Triton will log to file for every 50 traces collected. The same as deprecated `--trace-log-frequency`|
+
+In addition to configure trace settings for Triton trace APIs 
+in command line arguments, The user may
+modify the trace setting when Triton server
+is running via the trace APIs, more information can be found in [trace
+protocol](../protocol/extension_trace.md). This option is currently not supported,
+when trace mode is set to `opentelemetry`.
+
+**Note**: the following flags are **depricated**:
 
 The `--trace-file` option indicates where the trace output should be
 written. The `--trace-rate` option specifies the sampling rate. In
@@ -48,11 +92,6 @@ file for every 50 traces collected. The `--trace-count` option specifies the
 remaining number of traces to be collected. In this example Triton will stop
 tracing more requests after 100 traces are collected.  Use the `--help` option
 to get more information.
-
-In addition to configure trace settings in command line arguments, The user may
-modify the trace setting when Triton server
-is running via the trace APIs, more information can be found in [trace
-protocol](../protocol/extension_trace.md).
 
 ## Supported Trace Level Option
 
@@ -317,3 +356,21 @@ The meaning of the trace timestamps is:
     scheduling.
 
   * BACKEND_OUTPUT: The tensor in the response of a backend.
+
+## Opentelemetry trace support
+
+Starting with 23.04 release triton provides an option to generate and export traces, 
+using [Opentelemetry APIs and SDKs](https://opentelemetry.io/). Currently this option is
+only supported for Ubuntu based systems and does not support tracing for bls models.
+
+To specify Opentelemetry as a tracing APIs, specify --trace-config flag as follows:
+
+```
+$ tritonserver --trace-config=mode=opentelemetry ...
+```
+At the moment, triton supports HTTP Exporter, provided by Opentelemetry. The endpoint 
+can be specified as follows:
+```
+$ tritonserver --trace-config=mode=opentelemetry --trace-config=opentelemetry,url=<your endpoint> ...
+```
+By default, opentelemetry uses `0.0.0.0:4318/v2/traces` as an endpoint for HTTP exporter.
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -76,6 +76,17 @@ if(${TRITON_ENABLE_HTTP} OR ${TRITON_ENABLE_METRICS} OR
   message(STATUS "Using libevent ${Libevent_VERSION}")
 endif()
 
+# OpenTelemetry
+#
+if(${TRITON_ENABLE_TRACING})
+  find_package(CURL CONFIG REQUIRED)
+  message(STATUS "Using curl ${CURL_VERSION}")
+  find_package(nlohmann_json CONFIG REQUIRED)
+  message(STATUS "Using nlohmann_json for otel")
+  find_package(opentelemetry-cpp CONFIG REQUIRED)
+  message(STATUS "Using opentelemetry-cpp ${opentelemetry-cpp_VERSION}")
+endif()
+
 # re2
 #
 find_library(RE2_LIBRARY NAMES re2)
@@ -226,9 +237,15 @@ if(${TRITON_ENABLE_STATS})
 endif() # TRITON_ENABLE_STATS
 
 if(${TRITON_ENABLE_TRACING})
+  
   target_compile_definitions(
     main
     PRIVATE TRITON_ENABLE_TRACING=1
+  )
+  target_include_directories(
+    main
+    PRIVATE
+      ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
   )
 endif() # TRITON_ENABLE_TRACING
 
@@ -281,6 +298,7 @@ if(${TRITON_ENABLE_GRPC})
     main
     PRIVATE
       $<TARGET_PROPERTY:gRPC::grpc,INTERFACE_INCLUDE_DIRECTORIES>
+      ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
   )
 
   target_compile_definitions(
@@ -370,8 +388,12 @@ if(${TRITON_ENABLE_HTTP}
 
   target_include_directories(
     http-endpoint-library
-    PRIVATE $<TARGET_PROPERTY:libevhtp::evhtp,INTERFACE_INCLUDE_DIRECTORIES>
+    PRIVATE 
+      $<TARGET_PROPERTY:libevhtp::evhtp,INTERFACE_INCLUDE_DIRECTORIES>
+      ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+      #tracing-library
   )
+  
 
   if(${TRITON_ENABLE_GPU})
     target_compile_definitions(
@@ -478,6 +500,16 @@ if(${TRITON_ENABLE_TRACING})
     tracing-library EXCLUDE_FROM_ALL
     tracer.cc tracer.h
   )
+  
+  target_include_directories(
+    tracing-library 
+    PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+  )
+
+  target_link_libraries(
+    tracing-library
+    PRIVATE 
+    ${OPENTELEMETRY_CPP_LIBRARIES})
 
   target_link_libraries(
     tracing-library
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -298,7 +298,6 @@ if(${TRITON_ENABLE_GRPC})
     main
     PRIVATE
       $<TARGET_PROPERTY:gRPC::grpc,INTERFACE_INCLUDE_DIRECTORIES>
-      ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
   )
 
   target_compile_definitions(
@@ -391,7 +390,6 @@ if(${TRITON_ENABLE_HTTP}
     PRIVATE 
       $<TARGET_PROPERTY:libevhtp::evhtp,INTERFACE_INCLUDE_DIRECTORIES>
       ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
-      #tracing-library
   )
   
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -78,13 +78,15 @@ endif()
 
 # OpenTelemetry
 #
-if(${TRITON_ENABLE_TRACING})
-  find_package(CURL CONFIG REQUIRED)
-  message(STATUS "Using curl ${CURL_VERSION}")
-  find_package(nlohmann_json CONFIG REQUIRED)
-  message(STATUS "Using nlohmann_json for otel")
-  find_package(opentelemetry-cpp CONFIG REQUIRED)
-  message(STATUS "Using opentelemetry-cpp ${opentelemetry-cpp_VERSION}")
+if (NOT WIN32)
+  if(${TRITON_ENABLE_TRACING})
+    find_package(CURL CONFIG REQUIRED)
+    message(STATUS "Using curl ${CURL_VERSION}")
+    find_package(nlohmann_json CONFIG REQUIRED)
+    message(STATUS "Using nlohmann_json for otel")
+    find_package(opentelemetry-cpp CONFIG REQUIRED)
+    message(STATUS "Using opentelemetry-cpp ${opentelemetry-cpp_VERSION}")
+  endif()
 endif()
 
 # re2
@@ -237,16 +239,18 @@ if(${TRITON_ENABLE_STATS})
 endif() # TRITON_ENABLE_STATS
 
 if(${TRITON_ENABLE_TRACING})
-  
   target_compile_definitions(
     main
     PRIVATE TRITON_ENABLE_TRACING=1
   )
-  target_include_directories(
-    main
-    PRIVATE
-      ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
-  )
+# FIXME: remove, when Windows support is added for Opentelemetry
+  if (NOT WIN32)
+    target_include_directories(
+      main
+      PRIVATE
+        ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+    )
+  endif()
 endif() # TRITON_ENABLE_TRACING
 
 if(${TRITON_ENABLE_NVTX})
@@ -387,11 +391,17 @@ if(${TRITON_ENABLE_HTTP}
 
   target_include_directories(
     http-endpoint-library
-    PRIVATE 
-      $<TARGET_PROPERTY:libevhtp::evhtp,INTERFACE_INCLUDE_DIRECTORIES>
-      ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+    PRIVATE $<TARGET_PROPERTY:libevhtp::evhtp,INTERFACE_INCLUDE_DIRECTORIES>
   )
   
+  # FIXME when Triton support of Opentelemetry is available on Windows
+  # add ${OPENTELEMETRY_CPP_INCLUDE_DIRS} to above target_include_directories
+  if (NOT WIN32)
+    target_include_directories(
+      http-endpoint-library
+      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+    )
+  endif()
 
   if(${TRITON_ENABLE_GPU})
     target_compile_definitions(
@@ -499,15 +509,17 @@ if(${TRITON_ENABLE_TRACING})
     tracer.cc tracer.h
   )
   
-  target_include_directories(
-    tracing-library 
-    PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
-  )
+  if (NOT WIN32)
+    target_include_directories(
+      tracing-library 
+      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+    )
 
-  target_link_libraries(
-    tracing-library
-    PRIVATE 
-    ${OPENTELEMETRY_CPP_LIBRARIES})
+    target_link_libraries(
+      tracing-library
+      PRIVATE 
+      ${OPENTELEMETRY_CPP_LIBRARIES})
+  endif()
 
   target_link_libraries(
     tracing-library
--- a/src/command_line_parser.cc
+++ b/src/command_line_parser.cc
@@ -90,7 +90,8 @@ getopt_long(
 #include "triton/common/triton_json.h"
 
 
-namespace triton { namespace server {
+namespace triton {
+namespace server {
 
 // [FIXME] expose following parse helpers for other type of parser
 namespace {
@@ -547,13 +548,13 @@ std::vector<Option> TritonParser::recognized_options_
        "when Triton collects the 100-th trace, it logs the traces to file "
        "<trace-file>.0, and when it collects the 200-th trace, it logs the "
        "101-th to the 200-th traces to file <trace-file>.1. Default is 0."},
-       {OPTION_TRACE_MODE, "trace-mode", Option::ArgStr, 
-        "Specify the mode for tracing. Options are \"triton\" "
-       "and \"opentelemetry\". The default is \"triton\". For \"triton\", " 
+      {OPTION_TRACE_MODE, "trace-mode", Option::ArgStr,
+       "Specify the mode for tracing. Options are \"triton\" "
+       "and \"opentelemetry\". The default is \"triton\". For \"triton\", "
        "the server will use the Triton's trace APIs. For \"opentelemetry\", "
        "the server will ose the OpenTelemetry's APIs to generate, collect and "
        "export traces for individual inference requests."},
-       {OPTION_TRACE_CONFIG, "trace-config", "<string>,<string>=<string>",
+      {OPTION_TRACE_CONFIG, "trace-config", "<string>,<string>=<string>",
        "Specify a framework-specific configuration setting. The format of this "
        "flag is --tracing-config=<framework_name>,<setting>=<value>. Where "
        "<framework_name> is one of the available frameworks. Currently, the "
@@ -987,6 +988,10 @@ TritonParser::Parse(int argc, char** argv)
   bool cache_config_present{false};
 #ifdef TRITON_ENABLE_TRACING
   bool explicit_disable_trace{false};
+  // bool trace_filepath_present{false};
+  bool trace_level_present{false};
+  bool trace_rate_present{false};
+  bool trace_count_present{false};
 #endif  // TRITON_ENABLE_TRACING
 
 #ifdef TRITON_ENABLE_GRPC
@@ -1257,9 +1262,21 @@ TritonParser::Parse(int argc, char** argv)
 
 #ifdef TRITON_ENABLE_TRACING
       case OPTION_TRACE_FILEPATH:
+        std::cerr << "Warning: '--trace-file' has been deprecated and will be"
+                     " removed in future releases. Please use "
+                     "'--trace-config=triton,trace-file=<filepath> instead."
+                  << std::endl;
+        // trace_filepath_present = true;
+        // lparams.trace_config_map_["triton"].push_back({"trace-file",
+        // optarg});
         lparams.trace_filepath_ = optarg;
         break;
       case OPTION_TRACE_LEVEL: {
+        std::cerr << "Warning: '--trace-level' has been deprecated and will be"
+                     " removed in future releases. Please use "
+                     "'--trace-config=trace-level= instead."
+                  << std::endl;
+        trace_level_present = true;
         auto parsed_level = ParseTraceLevelOption(optarg);
         explicit_disable_trace |=
             (parsed_level == TRITONSERVER_TRACE_LEVEL_DISABLED);
@@ -1267,10 +1284,21 @@ TritonParser::Parse(int argc, char** argv)
             lparams.trace_level_ | parsed_level);
         break;
       }
-      case OPTION_TRACE_RATE:
+      case OPTION_TRACE_RATE: 
+        std::cerr << "Warning: '--trace-rate' has been deprecated and will be"
+                     " removed in future releases. Please use "
+                     "'--trace-config=trace-rate=<trace rate> instead."
+                  << std::endl;
+        trace_rate_present = true;
         lparams.trace_rate_ = ParseOption<int>(optarg);
         break;
+      
       case OPTION_TRACE_COUNT:
+        std::cerr << "Warning: '--trace-count' has been deprecated and will be"
+                     " removed in future releases. Please use "
+                     "'--trace-config=trace-count=<trace count> instead."
+                  << std::endl;
+        trace_count_present = true;
         lparams.trace_count_ = ParseOption<int>(optarg);
         break;
       case OPTION_TRACE_LOG_FREQUENCY:
@@ -1279,10 +1307,16 @@ TritonParser::Parse(int argc, char** argv)
       case OPTION_TRACE_MODE:
         lparams.trace_mode_ = ParseTraceModeOption(optarg);
         break;
-      case OPTION_TRACE_CONFIG:
-        lparams.trace_config_settings_.push_back(
-            ParseTraceConfigOption(optarg));
+      case OPTION_TRACE_CONFIG: {
+        auto trace_config_setting = ParseTraceConfigOption(optarg);
+        triton::server::TraceConfig& tc =
+            lparams
+                .trace_config_map_[std::get<0>(trace_config_setting).c_str()];
+        tc.push_back(std::make_pair(
+            std::get<1>(trace_config_setting).c_str(),
+            std::get<2>(trace_config_setting).c_str()));
         break;
+      }
 #endif  // TRITON_ENABLE_TRACING
 
       case OPTION_POLL_REPO_SECS:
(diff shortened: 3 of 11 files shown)

### CODE REVIEW

File: docs/user_guide/trace.md [hunk 1]
```diff
@@ -30,13 +30,57 @@
...
 Triton includes that capability to generate a detailed trace for
 individual inference requests. Tracing is enable by command-line
-arguments when running the tritonserver executable. For example,
+arguments when running the tritonserver executable.
+
+Starting in 23.04 release, Triton has one flag `--trace-config` to specify 
+global or trace mode specific configuration setting. The format of this flag 
+is `--tracing-config=<trace mode>,<setting>=<value>`, where `<trace mode>` 
+is either `triton` or `opentelemetry`. By default, trace mode is set to `triton`,
+and the server will use the Triton's trace APIs. For `opentelemetry`,
```
rmccorm4, line 39:
```suggestion
is `--trace-config=<mode>,<setting>=<value>`, where `<mode>` 
is either `triton` or `opentelemetry`. By default, the trace mode is set to `triton`,
and the server will use Triton's trace APIs. For `opentelemetry` mode, 
```

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
+Starting in 23.04 release, Triton has one flag `--trace-config` to specify 
+global or trace mode specific configuration setting. The format of this flag 
+is `--tracing-config=<trace mode>,<setting>=<value>`, where `<trace mode>` 
+is either `triton` or `opentelemetry`. By default, trace mode is set to `triton`,
+and the server will use the Triton's trace APIs. For `opentelemetry`, 
+the server will use the [OpenTelemetry's APIs](#opentelemetry-trace-support) to generate, collect and export 
+traces for individual inference requests.
+
+To specify global trace settings (level, rate, count, or mode), 
+the format is `--tracing-config=<setting>=<value>`.
```
rmccorm4, line 44:
```suggestion
the format is `--trace-config=<setting>=<value>`.
```

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
+| `count`     | -1            | Specifies the remaining number of traces to be collected. In the example Triton will stop tracing requests after 100 traces are collected. The same as deprecated `--trace-count`. |
+| `mode`      | triton        | Specifies which trace APIs to use for collecting traces. |
+
+### Triton trace APIs settings
+
+The following table shows available Triton trace APIs settings for `--trace-config=triton,<setting>=<value>`.
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `file`      |     -         | Indicates where the trace output should be written. The same as deprecated `--trace-file`.|
```
rmccorm4, line 72: Default file value is `-` ? Should it be an empty string or some `/tmp` file?

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
+## Trace settings
+### Global settings
+The following table shows available global trace settings to pass to `--trace-config`
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `rate`      | 1000          | Specifies the sampling rate. The same as deprecated `--trace-rate`. In the above example every 100-th inference request will be traced.|
+| `level`     | OFF           | Indicates the level of trace detail that should be collected and may be specified multiple times to trace multiple informations. The same as deprecated `--trace-level`|
+| `count`     | -1            | Specifies the remaining number of traces to be collected. In the example Triton will stop tracing requests after 100 traces are collected. The same as deprecated `--trace-count`. |
+| `mode`      | triton        | Specifies which trace APIs to use for collecting traces. |
```
rmccorm4, line 64:
```suggestion
| `mode`      | triton        | Specifies which trace APIs to use for collecting traces. The choices are "triton" or "opentelemetry". |
```

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
+
+### Triton trace APIs settings
+
+The following table shows available Triton trace APIs settings for `--trace-config=triton,<setting>=<value>`.
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `file`      |     -         | Indicates where the trace output should be written. The same as deprecated `--trace-file`.|
+| `log-frequency`| 0          | Specifies the rate that the traces are written to file. In the example Triton will log to file for every 50 traces collected. The same as deprecated `--trace-log-frequency`|
```
rmccorm4, line 73:
```suggestion
| `log-frequency`| 0          | Specifies the rate that the traces are written to file. For example, a value of 50 specifies that Triton will log to file for every 50 traces collected. The same as deprecated `--trace-log-frequency`|
```

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
+    --trace-config=count=100 ...
 ```
 
+## Trace settings
+### Global settings
+The following table shows available global trace settings to pass to `--trace-config`
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `rate`      | 1000          | Specifies the sampling rate. The same as deprecated `--trace-rate`. In the above example every 100-th inference request will be traced.|
```
rmccorm4, line 61:
```suggestion
| `rate`      | 1000          | Specifies the sampling rate. The same as deprecated `--trace-rate`. For example, a value of 1000 specifies that every 1000-th inference request will be traced. |
```

docs/user_guide/trace.md
```diff
@@ -317,3 +356,21 @@ The meaning of the trace timestamps is:
     scheduling.
 
   * BACKEND_OUTPUT: The tensor in the response of a backend.
+
+## Opentelemetry trace support
+
+Starting with 23.04 release triton provides an option to generate and export traces, 
+using [Opentelemetry APIs and SDKs](https://opentelemetry.io/). Currently this option is
+only supported for Ubuntu based systems and does not support tracing for bls models.
```
rmccorm4, line 364:
```suggestion
Starting with the 23.04 release, Triton provides an option to generate and export traces 
using [Opentelemetry APIs and SDKs](https://opentelemetry.io/). Currently this option is
only supported for Ubuntu based systems and does not support tracing for 
[BLS](https://github.com/triton-inference-server/python_backend/tree/main#business-logic-scripting) models.
```

docs/user_guide/trace.md
```diff
@@ -317,3 +356,21 @@ The meaning of the trace timestamps is:
...
+
+To specify Opentelemetry as a tracing APIs, specify --trace-config flag as follows:
```
rmccorm4, line 366:
```suggestion
To specify Opentelemetry mode for tracing, specify the `--trace-config` flag as follows:
```

docs/user_guide/trace.md
```diff
@@ -317,3 +356,21 @@ The meaning of the trace timestamps is:
...
+
+```
+$ tritonserver --trace-config=mode=opentelemetry ...
+```
+At the moment, triton supports HTTP Exporter, provided by Opentelemetry. The endpoint
```
rmccorm4, line 371: Add hyperlink to HTTP Exporter

docs/user_guide/trace.md
```diff
@@ -317,3 +356,21 @@ The meaning of the trace timestamps is:
...
+
+To specify Opentelemetry as a tracing APIs, specify --trace-config flag as follows:
+
+```
+$ tritonserver --trace-config=mode=opentelemetry ...
+```
+At the moment, triton supports HTTP Exporter, provided by Opentelemetry. The endpoint 
+can be specified as follows:
+```
+$ tritonserver --trace-config=mode=opentelemetry --trace-config=opentelemetry,url=<your endpoint> ...
```
rmccorm4, line 374:
```suggestion
$ tritonserver --trace-config=mode=opentelemetry --trace-config=opentelemetry,url=<endpoint> ...
```

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
+The following table shows available global trace settings to pass to `--trace-config`
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `rate`      | 1000          | Specifies the sampling rate. The same as deprecated `--trace-rate`. In the above example every 100-th inference request will be traced.|
+| `level`     | OFF           | Indicates the level of trace detail that should be collected and may be specified multiple times to trace multiple informations. The same as deprecated `--trace-level`|
+| `count`     | -1            | Specifies the remaining number of traces to be collected. In the example Triton will stop tracing requests after 100 traces are collected. The same as deprecated `--trace-count`. |
+| `mode`      | triton        | Specifies which trace APIs to use for collecting traces. |
+
+### Triton trace APIs settings
```
rmccorm4, line 66: Can you add a similar `OpenTelemetry Trace API Settings` table that describes `url` and other relevant fields?

rmccorm4, line 66: Titlecase for headers
```suggestion
### Triton Trace APIs Settings
```

docs/user_guide/trace.md
```diff
@@ -30,13 +30,57 @@
...
-$ tritonserver --trace-file=/tmp/trace.json --trace-rate=100 \
-    --trace-level=TIMESTAMPS --trace-log-frequency=50 --trace-count=100 ...
+$ tritonserver --trace-config=triton,file=/tmp/trace.json \
+    --trace-config=triton,log-frequency=50 --trace-config=rate=100 \
+    --trace-config=level=TIMESTAMPS --trace-config=triton,log-frequency=50 \
+    --trace-config=count=100 ...
 ```
 
+## Trace settings
+### Global settings
```
rmccorm4, line 56:
```suggestion
## Trace Settings
### Global Settings
```

src/http_server.cc
```diff
@@ -1819,6 +1820,9 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
       req,
       trace_response.AddString("log_frequency", std::to_string(log_frequency)));
   HTTP_RESPOND_IF_ERR(req, trace_response.AddString("trace_file", filepath));
+  HTTP_RESPOND_IF_ERR(
+      req, trace_response.AddString(
+               "trace_mode", TRITONSERVER_InferenceTraceModeString(mode)));
```
GuanLuo, line 1825: Wouldn’t the “mode” be the property of the trace manager created? I don’t think it requires Triton core changes, and the change in core repo doesn’t do anything special with respect to the mode given.

oandreeva-nv: I thought that it would make sense to put `mode` into core now, since later implementations may touch the core as well. For example, `infer_trace.h` has `Report` to report trace activity. I thought it would make sense to later incorporate opentelemetry's span updates there. Plus, we would also need to add `bls` support for opentelemetry. That being said, I understand that the more discussions are need to happen around the implementation, but I believe it make sense to keep `mode` changes in the core repo for now. Let me know if I am wrong.

rmccorm4: I think the core changes in `tritonserver.h` etc. are not necessary right now. If we want them later, we can add them. But if we add them now and they're not needed, we have to keep it in there for backwards compatibility. So I think it's better to not add it until we really need it, for any C APIs.

rmccorm4: I don't think `TRITONSERVER_InferenceTraceLevelString` is currently used in any of our repos. It may be used by custom in-process applications written by users, but it's likely not a critical thing to have until requested.

Similarly, it may make more sense to pivot towards some `TRITONSERVER_InferenceTraceConfig()` with the move to `--trace-config` rather than exposing a new API for each field, but again probably not needed until requested.

oandreeva-nv: Got it, I'll incorporate changes regarding the `mode`

nv-kmcgill53 (changes requested):

docs/user_guide/trace.md
oandreeva-nv: It's empty string. I'll update it. Forgot to change it here

docs/user_guide/trace.md
oandreeva-nv: It is on its way.

src/command_line_parser.cc
```diff
@@ -1418,6 +1481,79 @@ TritonParser::Parse(int argc, char** argv)
...
+        }
+        lparams.trace_log_frequency_ = ParseOption<int>(mode_setting.second);
+      }
+    }
+
+    if (lparams.trace_filepath_.empty()) {
+      throw ParseException(
+          "Error: file path must be provided for trace mode='triton' "
+          "Please set file path with --trace-config=triton,file=<filepath>");
+    }
```
rmccorm4, line 1554: I think this logic needs to be checked, because the default trace mode is Triton, whether user passes CLI args or not. If they don't pass any trace CLI args (ex: tritonserver --model-repository models`), trace mode will be triton and this check will fail because no file was given.

src/tracer.h
```diff
@@ -125,6 +149,37 @@ class TraceManager {
...
+
+    std::unique_ptr<otel_trace_sdk::SpanProcessor> processor_;
+
+    std::shared_ptr<otel_trace_api::TracerProvider> provider_;
+
+    // Starts a span with the provided timestamp
+    void InitSpan(opentelemetry::common::SystemTimestamp timestamp_ns);
+
+    // Ends the started span and cleans up tracer
+    void CleanTracer();
```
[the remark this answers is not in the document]
oandreeva-nv: Ending span is not the only thing that `CleanTracer()` is doing. I'm adding `InitTracer` in the new revision, so there will bbe logic in `InitTracer` and `CleanTracer`

src/tracer.h
```diff
@@ -34,10 +34,26 @@
...
+namespace otlp = opentelemetry::exporter::otlp;
+namespace otel_trace_sdk = opentelemetry::sdk::trace;
+namespace otel_trace_api = opentelemetry::trace;
+#endif
+
 #include "triton/core/tritonserver.h"
 
 namespace triton { namespace server {
 
+using TraceConfig = std::vector<std::pair<std::string, std::string>>;
```
tanmayv25, line 54: Instead of a vector of string -> string pair, how about a class/struct holding all the config settings?

src/tracer.cc
```diff
@@ -33,34 +33,44 @@
...
+namespace otel_trace_api = opentelemetry::trace;
+namespace otel_common = opentelemetry::common;
+#endif
 
 namespace triton { namespace server {
 
 TRITONSERVER_Error*
 TraceManager::Create(
     TraceManager** manager, const TRITONSERVER_InferenceTraceLevel level,
     const uint32_t rate, const int32_t count, const uint32_t log_frequency,
```
tanmayv25, line 48: Shouldn't settings like rate, count, log_frequency and filepath be a part of config_map?
Here and everywhere below.

src/tracer.cc
```diff
@@ -33,34 +33,44 @@
...
 
 TraceManager::TraceManager(
     const TRITONSERVER_InferenceTraceLevel level, const uint32_t rate,
     const int32_t count, const uint32_t log_frequency,
-    const std::string& filepath)
+    const std::string& filepath, const TRITONSERVER_InferenceTraceMode mode,
+    const TraceConfigMap& config_map)
 {
   std::shared_ptr<TraceFile> file(new TraceFile(filepath));
   global_default_.reset(new TraceSetting(
```
tanmayv25, line 68: For all these default values can you simplify them by
false /* variable name */, false /*variable name*/ etc.

It will be easier on eyes.

src/tracer.cc
```diff
@@ -279,34 +312,75 @@ TraceManager::SampleTrace(const std::string& model_name)
...
+#ifndef _WIN32
+    } else if (setting_->mode_ == TRITONSERVER_TRACE_MODE_OPENTELEMETRY) {
+      otel_common::SystemTimestamp otel_timestamp{
+          (time_offset_ + std::chrono::nanoseconds{timestamp_ns})};
+      if (trace_span_ == nullptr) {
+        this->InitSpan(otel_timestamp);
+      }
+      trace_span_->AddEvent(
+          name, otel_timestamp, {{"triton.steady_timestamp_ns", timestamp_ns}});
+#endif
```
tanmayv25, line 356: Should write a else block for these conditionals. And generate a warning message for unsupported trace_mode.

src/tracer.cc
```diff
@@ -374,22 +451,42 @@ TraceManager::TraceActivity(
...
-    }
-    *ss << "},";
+      if (parent_id != 0) {
+        *ss << ",\"parent_id\":" << parent_id;
+      }
+      *ss << "},";
+#ifndef _WIN32
+    } else if (ts->setting_->mode_ == TRITONSERVER_TRACE_MODE_OPENTELEMETRY) {
+      ts->trace_span_->SetAttribute("model_name", model_name);
+      ts->trace_span_->SetAttribute("model_version", model_version);
```
tanmayv25, line 469: Why not include request_id and parent_ids as an attribute here?

src/command_line_parser.cc
oandreeva-nv: Agree, realized that later.

src/tracer.cc
oandreeva-nv: They should, but I think it would be better to adjust it in a separate PR, since it is a significant change. For now the logic is these args are set or overrided  during the post parsing validation in `command_line_parser.cc`.

src/tracer.cc
oandreeva-nv: updated

src/tracer.cc
oandreeva-nv: updated

src/tracer.h
oandreeva-nv: It's a good idea and can be introduced in a separate PR

oandreeva-nv:
I've added tests. + I don't add  `trace_mode` to a http/grpc responce of `GET v2[/models/${MODEL_NAME}]/trace/setting` , because a lot of tests at the moment (server and client) rely on hardcoded expected setting, which does not include `trace_mode`. This is better to be changed in a separate PR.

src/command_line_parser.cc
```diff
@@ -1730,6 +1794,172 @@ TritonParser::ParseTraceLevelOption(std::string arg)
...
+      arg.substr(delim_name + 1, delim_setting - delim_name - 1);
+  std::string value_string = arg.substr(delim_setting + 1);
+
+  if (setting_string.empty() || value_string.empty()) {
+    std::stringstream ss;
+    ss << "--trace-config option format is '<backend "
+          "name>,<setting>=<value>'. Got "
+       << arg << std::endl;
+    throw ParseException(ss.str());
+  }
```
rmccorm4, line 1852: 1. `--tracing-config` should be `--trace-config`
2. Update mentions of "framework" and "backend" to trace mode or similar where relevant

src/tracer.cc
```diff
@@ -33,34 +33,50 @@
...
+      false /*filepath_specified*/, false /*mode_specified*/,
+      false /*config_map_specified*/));
   global_setting_.reset(new TraceSetting(
-      level, rate, count, log_frequency, file, false, false, false, false,
-      false));
+      level, rate, count, log_frequency, file, mode, config_map,
+      false /*level_specified*/, false /*rate_specified*/,
+      false /*count_specified*/, false /*log_frequency_specified*/,
+      false /*filepath_specified*/, false /*mode_specified*/,
+      false /*config_map_specified*/));
```
rmccorm4, line 79: Ideally this will be encapsulated in a config struct or similar in a follow-up PR

oandreeva-nv: Yes, I've discussed that with @tanmayv25  and created a ticket for this

src/tracer.cc
```diff
@@ -338,18 +457,21 @@ TraceManager::TraceActivity(
...
-    if (ts->streams_.find(id) == ts->streams_.end()) {
-      std::unique_ptr<std::stringstream> stream(new std::stringstream());
-      ss = stream.get();
-      ts->streams_.emplace(id, std::move(stream));
-    } else {
-      ss = ts->streams_[id].get();
-      // If the string stream is not newly created, add "," as there is
-      // already content in the string stream
-      *ss << ",";
+    if (ts->setting_->mode_ == TRACE_MODE_TRITON) {
```
rmccorm4, line 464: I think its OK for now to get into release, but I think `TraceActivity` and `TraceTensorActivity()` should be refactored with helper functions, something like below. I find the current function pretty hard to parse through, lots of `if`s and `ifdef` to parse:
```
    if (ts->setting_->mode_ == TRACE_MODE_TRITON) {
      TraceActivityTriton()
    } 

    else if (OPENTELEMETRY) {
      TraceActivityOpenTelemetry();
    }
    
    // etc.
```

src/tracer.h
```diff
@@ -72,6 +97,12 @@ class TraceManager {
 
     bool clear_filepath_;
     const std::string* filepath_;
+
+    bool clear_mode_;
+    const InferenceTraceMode* mode_;
+
+    bool clear_config_map_;
+    const TraceConfigMap* config_map_;
```
rmccorm4, line 105: I don't think these new `clear_*` vars are being used. Do they need to be? I believe this is part of the trace extension in http/grpc servers.

If not, I would remove them for now until the feature is added.

src/tracer.cc
```diff
@@ -145,6 +163,16 @@ TraceManager::UpdateTraceSettingInternal(
...
+  const bool mode_specified =
+      (new_setting.clear_mode_ ? false
+                               : (((current_setting != nullptr) &&
+                                   current_setting->mode_specified_) ||
+                                  (new_setting.mode_ != nullptr)));
+  const bool config_map_specified =
+      (new_setting.config_map_ ? false
+                               : (((current_setting != nullptr) &&
+                                   current_setting->config_map_specified_) ||
+                                  (new_setting.config_map_ != nullptr)));
```
rmccorm4, line 175: Is there some logic off here?

I imagine this would have some symmetry like either:
```
const bool mode_specified = (new_setting.clear_mode_ ...)  // clear var
const bool config_map_specified = (new_setting.clear_config_map_ ...)  // clear var
```
or
```
const bool mode_specified = (new_setting.mode_ ...)  // actual var
const bool config_map_specified = (new_setting.config_map_ ...)  // actual var
```

but currently it is
```
const bool mode_specified = (new_setting.clear_mode_ ...)  // clear var
const bool config_map_specified = (new_setting.config_map_ ...)  // actual var
```

Side note, I don't think the new `clear_` vars are being used as mentioned in my other comment: https://github.com/triton-inference-server/server/pull/5637/files#r1167337834

So if they're not used in this PR, it may simplify the amount of code to review if you remove those parts

qa/L0_trace/test.sh
```diff
@@ -34,6 +34,10 @@ CLIENT_LOG="client.log"
 TEST_RESULT_FILE="test_results.txt"
 EXPECTED_NUM_TESTS="6"
 
+TRACE_COLLECTOR=socket_echo_server.py
```
tanmayv25, line 37: OTEL_TRACE_COLLECTOR? 
otel_socket_echo_server?

qa/L0_trace/test.sh
```diff
@@ -607,11 +611,122 @@ set -e
...
+    RET=1
+fi
+
+# Send grpc request and collect trace
+$SIMPLE_GRPC_CLIENT >> client_update.log 2>&1
+if [ $? -ne 0 ]; then
+    RET=1
+fi
+# To make sure log gets all data
+sleep 3
```
tanmayv25, line 692: This is very time sensitive  implementation. Our tests must not introduce such behavior. Can you add a comment here to re-design the implementation?
There should be some form of synchronization between tritonclient issuing request and socket_echo_server.

We should ideally have a single python unit test which starts the socket echo in one thread and issue requests in the main thread, then wait for the event with socket echo before checking for the collected trace.

qa/L0_cmdline_trace/test.sh
```diff
@@ -576,4 +570,59 @@ else
 fi
 
 
+# check deprecation warnings
```
rmccorm4, line 573: Can be follow-up, but add some negative tests and other config setting tests. Some off the top of my head:

1. Both triton and opentelemetry modes cannot both be specified at same time (I think?)
2. Can't set tensor level for opentelemetry mode
3. Assert default mode is Triton when not specifying any mode

src/tracer.cc
oandreeva-nv: thanks for catching this.

qa/L0_trace/test.sh
oandreeva-nv: Added FIXME in the beginning of the test

tanmayv25 (approved):

docs/user_guide/trace.md
```diff
@@ -30,13 +30,58 @@
...
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `file`      | empty string  | Indicates where the trace output should be written. The same as deprecated `--trace-file`.|
+| `log-frequency`| 0          | Specifies the rate that the traces are written to file. For example, a value of 50 specifies that Triton will log to file for every 50 traces collected. The same as deprecated `--trace-log-frequency`|
+
+In addition to configure trace settings for Triton trace APIs 
+in command line arguments, users may modify the trace setting when Triton server
+is running via the trace APIs, more information can be found in [trace
+protocol](../protocol/extension_trace.md). This option is currently not supported,
+when trace mode is set to `opentelemetry`.
```
Tabrizian, line 81:
```suggestion
In addition to the trace configuration settings in the command line, you can
modify the trace configuration using the [trace
protocol](../protocol/extension_trace.md). This option is currently not supported,
when trace mode is set to `opentelemetry`.
```

docs/user_guide/trace.md
```diff
@@ -317,3 +357,37 @@ The meaning of the trace timestamps is:
...
+Triton provides an option to generate and export traces 
+using [Opentelemetry APIs and SDKs](https://opentelemetry.io/). 
+
+To specify Opentelemetry mode for tracing, specify the `--trace-config` flag as follows:
+
+```
+$ tritonserver --trace-config=mode=opentelemetry --trace-config=opentelemetry,url=<endpoint> ...
+```
+
+### Opentelemetry trace APIs settings
```
Tabrizian, line 372: "Opentelemetry" -> "OpenTelemetry" here and everywhere.

qa/L0_trace/trace_collector.py
```diff
@@ -0,0 +1,66 @@
...
+        trace = ''
+        connection, client_address = sock.accept()
+        with connection:
+            with open(file_path, "a") as sys.stdout:
+                chunk = connection.recv(4096)
+                if not chunk:
+                    break
+                connection.sendall(chunk)
+                trace = chunk.decode()
+                print(trace)
```
Tabrizian, line 66: It is fine for the current PR but I think you can replace this Python script with a one line netcat command. (https://linux.die.net/man/1/nc see the data transfer section).

src/grpc/CMakeLists.txt
```diff
@@ -76,10 +76,19 @@ target_link_libraries(
...
+# FIXME when Triton support of Opentelemetry is available on Windows
+# add ${OPENTELEMETRY_CPP_INCLUDE_DIRS} to above target_include_directories
+# JIRA DLIS-4786
+if (NOT WIN32)
+  target_include_directories(
+    grpc-endpoint-library
+    PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+  )
+endif()
+
```
Tabrizian, line 91: Only include if tracing is enabled.

src/tracer.cc
```diff
@@ -338,18 +439,21 @@ TraceManager::TraceActivity(
   auto ts =
       reinterpret_cast<std::shared_ptr<TraceManager::Trace>*>(userp)->get();
 
+
```
Tabrizian, line 442: Remove extra space

src/tracer.cc
```diff
@@ -279,34 +300,101 @@ TraceManager::SampleTrace(const std::string& model_name)
...
   }
+  this->exporter_ = otlp::OtlpHttpExporterFactory::Create(opts);
+  this->processor_ = otel_trace_sdk::SimpleSpanProcessorFactory::Create(
+      std::move(this->exporter_));
+  this->provider_ = otel_trace_sdk::TracerProviderFactory::Create(
+      std::move(this->processor_));
 }
 
+void
+TraceManager::Trace::InitSpan(otel_common::SystemTimestamp timestamp_ns)
```
Tabrizian, line 379: Use const& for arguments.

src/tracer.cc
```diff
@@ -279,34 +300,101 @@ TraceManager::SampleTrace(const std::string& model_name)
...
+        }
+      }
+      *ss << "{\"id\":" << trace_id_ << ",\"timestamps\":["
+          << "{\"name\":\"" << name << "\",\"ns\":" << timestamp_ns << "}]}";
+    } else if (setting_->mode_ == TRACE_MODE_OPENTELEMETRY) {
+#ifndef _WIN32
+      otel_common::SystemTimestamp otel_timestamp{
+          (time_offset_ + std::chrono::nanoseconds{timestamp_ns})};
+      if (trace_span_ == nullptr) {
+        this->InitSpan(otel_timestamp);
```
Tabrizian, line 343: Remove this here and everywhere.

src/tracer.cc
```diff
@@ -279,34 +300,101 @@ TraceManager::SampleTrace(const std::string& model_name)
...
+      }
+      trace_span_->AddEvent(
+          name, otel_timestamp, {{"triton.steady_timestamp_ns", timestamp_ns}});
```
Tabrizian, line 346: Perhaps remove `steady_`?

src/CMakeLists.txt
```diff
@@ -372,6 +393,16 @@ if(${TRITON_ENABLE_HTTP}
     http-endpoint-library
     PRIVATE $<TARGET_PROPERTY:libevhtp::evhtp,INTERFACE_INCLUDE_DIRECTORIES>
   )
+  
+  # FIXME when Triton support of Opentelemetry is available on Windows
+  # add ${OPENTELEMETRY_CPP_INCLUDE_DIRS} to above target_include_directories
+  # JIRA DLIS-4786
+  if (NOT WIN32)
```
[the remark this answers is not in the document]
oandreeva-nv: if (NOT WIN32) was added everywhere where  OpenTelemetry was used

src/command_line_parser.cc
```diff
@@ -1444,7 +1508,7 @@ TritonParser::Parse(int argc, char** argv)
   }
   lparams.enable_cache_ = (cache_size_present || cache_config_present);
   return {lparams, {}};
-}
+}  // namespace server
```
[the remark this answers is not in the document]
oandreeva-nv: I think it may have been added with `clang`

qa/L0_trace/trace_collector.py
oandreeva-nv: I tried `netcat`, it wasn't installed in the `devel` container.

src/tracer.cc
oandreeva-nv: That is a reference to the steady clock

docs/user_guide/trace.md
```diff
@@ -30,13 +30,58 @@
...
+    --trace-config=triton,file=/tmp/trace.json \
+    --trace-config=triton,log-frequency=50 \
+    --trace-config=rate=100 \
+    --trace-config=level=TIMESTAMPS \
+    --trace-config=count=100 ...
 ```
 
+## Trace Settings
+### Global Settings
+The following table shows available global trace settings to pass to `--trace-config`
```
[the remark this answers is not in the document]
tanmayv25: Having all the flags as part of --trace-config makes sense as even though global, they are configuration of tracer.
I don't think the first equal sign is required. It can be interpreted correctly even with
`--trace-config rate=100`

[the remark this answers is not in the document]
tanmayv25: We can probably update the documentation of the cli examples below.

src/command_line_parser.cc
```diff
@@ -525,26 +526,44 @@ std::vector<Option> TritonParser::recognized_options_
...
        "<trace-file>.0, and when it collects the 200-th trace, it logs the "
        "101-th to the 200-th traces to file <trace-file>.1. Default is 0."},
+      {OPTION_TRACE_CONFIG, "trace-config", "<string>,<string>=<string>",
+       "Specify global or trace mode specific configuration setting. "
+       "The format of this flag is --trace-config="
+       "<mode>,<setting>=<value>. "
+       "Where <mode> is either \"triton\" or \"opentelemetry\". "
+       "The default is \"triton\". To specify global trace settings "
+       "(level, rate, count, or mode), the format would be "
+       "--trace-config=<setting>=<value>. For \"triton\" mode, the server will "
```
tanmayv25, line 562: remove extra equals from here as well.

tanmayv25 (commented):
LGTM ... @Tabrizian  @GuanLuo  If we can decide upon whether or not to deprecate the global tracer options in favor of specifying them as --tracer-config. @rmccorm4 's cli usage updates does clean-up our cli options help message quite a bit. Hence, it won't be not much of an issue to not deprecate the global settings just yet. 

But logically it looks nice to have everything within `--trace-config <>`

src/command_line_parser.cc
oandreeva-nv: Removed equal signs in `command_line_parser.cc` after all `--trace-config`

docs/user_guide/trace.md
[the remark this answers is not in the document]
oandreeva-nv: I think it make sense to move to one flag for same reasons as @tanmayv25 mentioned. With back to back `=` not required, it became eye-pleasant, and by eliminating separate global flags, users don't need to remember different flag names just to specify trace configuration.

It is also easier from user's side to check how many flags is set. If they have a large number of flags specified, by searching for `--trace-config` they can highlight all trace-related flags and count them for example

docs/user_guide/trace.md
oandreeva-nv: Updated

src/grpc/CMakeLists.txt
oandreeva-nv: Updated

src/tracer.cc
oandreeva-nv: Removed

src/tracer.cc
oandreeva-nv: Updated

src/tracer.cc
oandreeva-nv: Updated

src/tracer.h
oandreeva-nv: removed

src/CMakeLists.txt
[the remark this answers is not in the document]
oandreeva-nv: `if (NOT WIN32) ` is used together with `TRITON_ENABLE_TRACE`, most of the time not in the same conditional. If `if (NOT WIN32) ` present alone, then it is inside `TRITON_ENABLE_TRACE` check, which happens earlier

docs/user_guide/trace.md
oandreeva-nv: updated

src/tracer.cc
Tabrizian: I think it is an implementation detail that should be hidden from the user. Is there any way to make sure these timestamps have the same starting point?

docs/user_guide/trace.md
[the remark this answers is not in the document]
Tabrizian: I think it should be fine to use `--trace-config`.

docs/user_guide/trace.md
```diff
@@ -30,13 +30,58 @@
...
+
+| Setting     | Default Value |  Description                                                                                                                           |
+|:------------|:--------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `rate`      | 1000          | Specifies the sampling rate. The same as deprecated `--trace-rate`. For example, a value of 1000 specifies that every 1000-th inference request will be traced. |
```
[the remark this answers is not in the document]
Tabrizian: It seems like it is rendered incorrectly now: 
image

qa/L0_trace/trace_collector.py
Tabrizian: If you want to use it, you need to manually install it in the devel container by adding `apt-get install netcat`.

src/tracer.cc
oandreeva-nv: This way OpenTelemetry's spans contain the same info as our current traces (Triton trace API). 
 OpenTelemetry SDK relies on system's clock for event timestamps. Triton Tracing records timestamps using steady_clock. This is a monotonic clock, i.e. time is always moving forward. It is not related to wall clock time (for example, it can be time since last reboot). These 2 clocks are different clocks, and steady clock is usually used for more precise computations. However, so that traces are displayed in the trace collector correctly, system's clock is used. To sum up, since our current traces report steady timestamp only, I also include it in pan's info

docs/user_guide/trace.md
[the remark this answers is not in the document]
oandreeva-nv: I thought the line separation can still be clear. Meaning next line in the description belongs to the setting, specified earlier.

qa/L0_trace/trace_collector.py
oandreeva-nv: That was on my mind, but this was out of scope for this PR. I can add it as a separate PR later.

src/tracer.cc
Tabrizian: Also, I'm curious how this would work with some OTEL backends like prometheus and Jaeger. Have we done an end-to-end integration with these tools to see how the visualization would look like?

oandreeva-nv: @tanmayv25 we've discussed this at some point. Are we ok to remove triton's timestamp from trace's info?

@Tabrizian I've tested it with Jaeger, because this was the tool, that the customer is planning to use.  I showed it to Tanmay, it looks as expected. The introduction of `time_offset_` was driven by Jaeger's expectation for OpenTelemetry's timestamp to put trace on time axis.

docs/user_guide/trace.md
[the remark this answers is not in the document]
oandreeva-nv: Updating table style

src/tracer.cc
tanmayv25: We can remove triton's timestamp from OTEL trace for now. We can document the difference in behavior of the two trace modes.

docs/user_guide/trace.md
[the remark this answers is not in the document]
Tabrizian: Looks great! Thanks resolving.

nv-kmcgill53 (dismissed):

src/tracer.cc
oandreeva-nv: Removed steady timestamp from OTel spans.

docs/user_guide/trace.md
```diff
@@ -317,3 +426,56 @@ The meaning of the trace timestamps is:
...
+    <th>Description</th>
+  </tr>
+  </thead>
+  <tbody>
+    <tr>
+    <td><code>url</code></td>
+    <td><code>0.0.0.0:4318/v1/traces</code></td>
+    <td>
+      <code>host:port</code> to which the receiver is going to receive 
+      trace data.
```
rmccorm4, line 461: nit: The user isn't supposed to specify the route `/v1/traces/` part, are they? If not, make the default value in the table for `url` `0.0.0.0:4318` and note in the description that `/v1/traces` will be called at this `url`

src/tracer.cc
oandreeva-nv: @Tabrizian and @tanmayv25 I've removed steady timestamp and added  [entry](https://github.com/triton-inference-server/server/blob/197595f13d8c40aa0f4fb5b6a66ad08a540bf407/docs/user_guide/trace.md#differences-in-trace-contents-from-tritons-trace-output) to docs

docs/user_guide/trace.md
oandreeva-nv: I think it would be better to report default value the same as OpenTelemetry has for HTTP exporter:
https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.8.3/exporters/otlp/src/otlp_environment.cc#L139

I can add clarification that `/v1/traces` is not strongly required, for example?

oandreeva-nv: Also link to OTLP README, which I would like to follow for the default values: https://github.com/open-telemetry/opentelemetry-cpp/tree/v1.8.3/exporters/otlp#configuration-options--otlp-http-exporter-

Let me know what you think.

oandreeva-nv: `/v1/traces` won't be added to provided url - it wasn't added in my test, so I just collected on some port without `/v1/traces`. So if `url` is provided, it will be used as is. `/v1/traces` is added if environment variable `OTEL_EXPORTER_OTLP_ENDPOINT` is set: https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.8.3/exporters/otlp/src/otlp_environment.cc#L153

tanmayv25 (approved):

rmccorm4 (approved):

Tabrizian (approved):

PR #5642, opened 2023-04-14
Add and document memory usage in statistic protocol

GuanLuo:


docs/protocol/extension_statistics.md
```diff
@@ -119,6 +120,14 @@ $model_stat =
   due to different batch size (for example, larger batches typically
   take longer to compute).
 
+- "memory_usage" : The memory usage detected during model loading, which may be
```
tanmayv25, line 123: I understand why it makes sense to report the memory usage in per model statistics. I was wondering whether we should include per model GPU/CPU consumption in our CPU/GPU metrics. 
https://github.com/triton-inference-server/server/blob/main/docs/user_guide/metrics.md#gpu-metrics

The downside is that for triton serving large number of models, this might blow-up.

tanmayv25 (commented):
It seems we are always tracking the memory_stats for the model. From what I understood, this might affect the actual model load times. Shouldn't per model memory tracking be optional?

GuanLuo:
> From what I understood, this might affect the actual model load times.

I don't see noticeable difference even in inference case, making it optional require a larger change to propagating the configuration, let me know if you think it is still better to be configurable.

docs/protocol/extension_statistics.md
GuanLuo: I think the memory usage at model level is not as dynamic, I think it belongs to model metadata or model config as if it is part of the model properties, but I don't think the protocol is open to adding new field and thus put in statistics.

tanmayv25:
> I don't see noticeable difference even in inference case.

Were you experimenting with large enough models?

docs/protocol/extension_statistics.md
rmccorm4, line 123: Might not be in scope of this ticket, but I'm curious how this interacts with something like ONNXRuntime with TRT EP. I believe ORT will create new TRT engines for new shapes in inference requests at runtime. Would it be misleading if this reports `A` for the model's memory usage, but is actually taking up `B >> A` after a few requests?

docs/protocol/extension_statistics.md
```diff
@@ -217,6 +226,22 @@ $duration_stat =
 
 - “ns” : The total duration for the statistic in nanoseconds.
 
+```
+$memory_usage = 
+{
+  "type" : $string,
+  "id" : $number,
+  "byte_size" : $number
+}
```
Tabrizian, line 235: Is `memory_usage` an array of `type`, `id`, `byte_size` values? Can a model report a mix of these values?

docs/protocol/extension_statistics.md
```diff
@@ -325,7 +350,16 @@ message ModelStatistics
   // executed in the model. The batch statistics indicate how many actual
   // model executions were performed and show differences due to different
   // batch size (for example, larger batches typically take longer to compute).
-  InferBatchStatistics batch_stats = 7;
+  repeated InferBatchStatistics batch_stats = 7;
```
dyastremsky, line 353: Why does this need to be repeated? Will this affect the tests that check this (e.g. L0_batcher).

GuanLuo: The field itself is repeated, just the doc here is not matching the service definition. It is repeated because each entry captures the stats for a batch size, so there will be multiple entries for multiple different batch sizes.

docs/protocol/extension_statistics.md
GuanLuo: Yes. For example, if model is deployed on every GPU on the system, then there will be entries matching the number of GPUs.

docs/protocol/extension_statistics.md
dyastremsky: Got it, thanks for explaining.

dyastremsky (approved):

qa/L0_device_memory_tracker/test.py
```diff
@@ -0,0 +1,88 @@
...
+        nvidia_smi.nvmlInit()
+        self.gpu_handle_ = nvidia_smi.nvmlDeviceGetHandleByIndex(0)
+        self.client_ = tritonclient.InferenceServerClient(url="localhost:8000")
+
+    def tearDown(self):
+        nvidia_smi.nvmlShutdown()
+
+    def report_used_gpu_memory(self):
+        info = nvidia_smi.nvmlDeviceGetMemoryInfo(self.gpu_handle_)
+        return info.used
```
rmccorm4, line 48: I think this portion will be sensitive to other processes using GPU, so will probably want to use exclusive GPU template/tags for this test.

qa/L0_device_memory_tracker/test.py
```diff
@@ -0,0 +1,88 @@
...
+    def verify_recorded_usage(self, model_stat):
+        recorded_gpu_usage = 0
+        for usage in model_stat["memory_usage"]:
+            if usage["type"] == "GPU":
+                recorded_gpu_usage += usage["byte_size"]
+        # unload and verify recorded usage
+        before_total_usage = self.report_used_gpu_memory()
+        self.client_.unload_model(model_stat["name"])
+        # unload return after puttting model to unload process,
+        # wait to be finished
```
rmccorm4, line 63:
```suggestion
        # unload can return before model is actually unloaded,
        # wait to be finished
```

qa/L0_device_memory_tracker/test.py
```diff
@@ -0,0 +1,88 @@
...
+
+import nvidia_smi
+
+
+class MemoryUsageTest(unittest.TestCase):
+
+    def setUp(self):
+        nvidia_smi.nvmlInit()
+        self.gpu_handle_ = nvidia_smi.nvmlDeviceGetHandleByIndex(0)
+        self.client_ = tritonclient.InferenceServerClient(url="localhost:8000")
```
rmccorm4, line 41: Can we test both HTTP and GRPC clients?

Tabrizian (commented):
Should we add a flag to only enable this feature when requested? I guess we should be fine since it is only used during model loading?

GuanLuo:
> Should we add a flag to only enable this feature when requested? I guess we should be fine with always enabling it since it is only used during model loading?

@Tabrizian it can be enabled via backend config cmdline args. I don't think enabling it on demand is desired as it just complicated the initialization

qa/L0_device_memory_tracker/test.py
GuanLuo: We can but that is not related to the code under test, so I don't want to complicate the test with that.

GuanLuo: Actually you are right, it did change the frontend service, modified test for both network protocols.

tanmayv25 (dismissed):

src/grpc/grpc_server.cc
```diff
@@ -1183,6 +1183,37 @@ CommonHandler::RegisterModelStatistics()
...
+            size_t type_len;
+            err = usage.MemberAsString("type", &type, &type_len);
+            GOTO_IF_ERR(err, earlyexit);
+            memory_usage->set_type(std::string(type, type_len));
+          }
+          {
+            uint64_t id;
+            err = usage.MemberAsUInt("id", &id);
+            GOTO_IF_ERR(err, earlyexit);
+            memory_usage->set_id(ucnt);
```
nnshah1, line 1208:
```suggestion
            memory_usage->set_id(id);
```
Can we use MemberAsInt() instead of Uint? And int64_t as in the message format?

nnshah1 (approved):

tanmayv25 (approved):

build.py
```diff
@@ -598,6 +598,16 @@ def backend_cmake_args(images, components, be, install_dir, library_paths):
     cargs.append(
         cmake_backend_enable(be, 'TRITON_ENABLE_METRICS', FLAGS.enable_metrics))
 
+    if (target_platform() == 'windows') and (not FLAGS.no_container_build):
+        print(
+            "Warning: Detected docker build is used for Windows, backend utility 'device memory tracker' will be disabled due to missing library in CUDA Windows docker image."
```
rmccorm4, line 603: Is windows containers build still disabled? Looks like you added CUPTI to windows image [here](https://github.com/triton-inference-server/server/pull/5642/commits/8db5267b1e28d8505acc870fc4eefc48a5b4bd13)?

qa/L0_device_memory_tracker/test.sh
```diff
@@ -0,0 +1,128 @@
...
+fi
+set -e
+
+# Set multiple instances on selected model to test instance-wise collection
+# and accumulation.
+echo "instance_group [{ count: 2; kind: KIND_GPU }]" >> models/resnet152_plan/config.pbtxt
+echo "instance_group [{ count: 2; kind: KIND_GPU }]" >> models/densenet/config.pbtxt
+
+# testing use nvidia-smi for Python to validate the reported usage
+pip install nvidia-ml-py3
```
rmccorm4, line 98: Both appear to be officially published by NVIDIA, but just wanted to point out there appears to be a more [actively maintained](https://pypi.org/project/pynvml/11.5.0/#history) package [pynvml](https://github.com/gpuopenanalytics/pynvml) that appears to do the same things. 

According to their [github](https://github.com/gpuopenanalytics/pynvml):
> As of version 11.0.0, the NVML-wrappers used in pynvml are identical to those published through [nvidia-ml-py](https://pypi.org/project/nvidia-ml-py/).

rmccorm4 (approved):

rmccorm4 (approved):
LGTM, just minor comments

qa/L0_device_memory_tracker/test.sh
rmccorm4: > Not sure why there are two packages though.

UPDATE: I reached out to the maintainer of the `pynvml` package (Rick Zamora). Looks like `nvidia-ml-py` is the right choide :slightly_smiling_face: 

He said:
> Hey Ryan, you are best off using nvidia-ml-py. I’ll try to add a new note to the README of pynvml to make this suggestion more explicit. The pynvml project was effectively a last-resort solution during a time when the NVML team went a long time without publishing any python bindings (and now the project is effectively a mirror to support projects that still rely on pynvml).

nnshah1 (dismissed):

tanmayv25 (approved):

PR #5650, opened 2023-04-17
Reduce default --model-load-thread-count to 4

rmccorm4:
On systems with large core counts, this default is too large. In fact, it is causing issues on OCI nodes hitting resource limits.

4 should be a comfortable default for the average use case or simple ensemble cases (preprocess, infer, postprocess) - and users who knowingly need high load concurrency can increase via CLI flag. 

Corresponding core PR: https://github.com/triton-inference-server/core/pull/193

We should rel note this change in 23.05 as well. CC @nv-kmcgill53


Tabrizian (dismissed):

dyastremsky (approved):
Looks good. Can you update the PR description to fix the link to be to [the core PR](https://github.com/triton-inference-server/core/pull/193) instead of this server PR?

kthui (approved):

PR #5652, opened 2023-04-17
Remove unused imports

dyastremsky:
This PR removes unused imports from the Python files in the server repo.


qa/L0_model_namespacing/test_duplication/addsub_repo/composing_model/1/model.py
```diff
@@ -3,4 +3,3 @@
 
 # load pre-defined QA model
 sys.path.append(os.environ["TRITON_QA_PYTHON_MODEL_DIR"])
-from python_addsub import *
```
Tabrizian, line 6: I think this line should be kept here and everywhere. cc @GuanLuo

dyastremsky: Good catch! Removed.

qa/L0_model_queue/model_queue_test.py
```diff
@@ -36,7 +36,6 @@
 import infer_util as iu
 import test_util as tu
 from tritonclientutils import InferenceServerException
-from ctypes import *
```
Tabrizian, line 39: Is this also not used? It would be good to make sure all the modified tests are passing to make sure nothing breaks.

Tabrizian (dismissed):

qa/L0_model_queue/model_queue_test.py
dyastremsky: Removed these and rerunning CI with the changes to ensure all still passes.

We may want to be very strict with including these global imports in the future. They pollute the namespace and add a level of ambiguity. They're getting identified by the linters for that reason, but for someone reading the code too, it's hard to trace the code. I know this is old code, just sharing for awareness so others on the team also know to keep an eye out for this.

I spent a little time trying to figure out where ctypes is used in this file and shm_utils, even going into the ctypes documentation, but it doesn't seem to be. I'm leaving the previous pipeline running too to see if tests pass without the import.

Tabrizian (approved):

qa/L0_model_queue/model_queue_test.py
Tabrizian: I agree. Looks like importing everything from a package is considered bad practice in Python.

PR #5655, opened 2023-04-18
Update log file extension

mc-nv:


rmccorm4 (commented):
LGTM, but looks like at least one more here: https://github.com/triton-inference-server/server/blob/8db214419a9d061ee666664711b78d61666bfb29/qa/L0_long_running_stress/test.sh#L141

qa/L0_dyna_sequence_batcher/test.sh
```diff
@@ -104,7 +104,7 @@ for i in \
         test_simple_sequence \
         test_length1_sequence \
          ; do
-    SERVER_LOG="./$i.serverlog"
+    SERVER_LOG="./$i.server.log"
```
tanmayv25, line 107: We can remove serverlog from Line 68. Similarly for other files.

tanmayv25 (changes requested):
L0_model_control_stress on gitlab side has similar nomenclature which can be updated. 
We can also remove serverlog extension from our check_copyright.py script.

qa/L0_dyna_sequence_batcher/test.sh
tanmayv25: `rm -fr *.log *.serverlog ` becomes `rm -fr *.log *`

mc-nv: adressed

tanmayv25 (approved):

PR #5657, opened 2023-04-18
Update Dockerfile.win10.min structure

mc-nv:


mc-nv:
Not certain if we have to merge it in release branch, may be better to use `main` branch for this change.

tanmayv25:
@mc-nv  Did we use the updated `Dockerfile.win10.min` for r23.04 windows min container and r23.04 pipeline? If yes, then we must merge these changes to r23.04 branch. If not, then they should only go to main branch.

tanmayv25 (approved):

nv-kmcgill53 (approved):

PR #5658, opened 2023-04-19
Minor fix for running "mlflow deployments create -t triton --flavor triton ..."

yeahdongcn:
UnboundLocalError: local variable 'meta_dict' referenced before assignment

The above error shows in listing models in Triton model repository

### DIFF
--- a/deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
+++ b/deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
@@ -193,6 +193,8 @@ class TritonPlugin(BaseDeploymentClient):
                         )['Body'].read().decode('utf-8'))
                 elif os.path.isfile(mlflow_meta_path):
                     meta_dict = self._get_mlflow_meta_dict(d['name'])
+                else:
+                    continue
 
                 d['triton_model_path'] = meta_dict['triton_model_path']
                 d['mlflow_model_uri'] = meta_dict['mlflow_model_uri']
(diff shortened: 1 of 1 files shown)


deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
```diff
@@ -193,6 +193,8 @@ def list_deployments(self):
                         )['Body'].read().decode('utf-8'))
                 elif os.path.isfile(mlflow_meta_path):
                     meta_dict = self._get_mlflow_meta_dict(d['name'])
+                else:
+                    continue
```
GuanLuo, line 197: Can you share an example that the "reference before assignment" error occurs? I wonder if this change will shadow the real cause of the issue.

This code block is executed when the model is ready in Triton server, the error seems to be happening only if the Triton server is also used for other deployments, i.e. some models being served is not added via MLFlow plugin and thus they don't have `meta_dict` stored in model folder.

If that is the case, then this change makes sense to filter only the MLFlow deployed models.

yeahdongcn: In my case, besides the models published from MLFlow, there are some other models in my model repository (host path). These models don't contain `meta.yaml` so `meta_dict` won't be declared in the context.

GuanLuo: That makes sense. Then it looks fine that the model not deployed via MLFlow should be agnostic to the plugin user.

GuanLuo (approved):
LGTM. Will add test around this.

Issue #5666, opened 2023-04-20
Install mlflow-triton plugin in mlflow container (ghcr.io/mlflow/mlflow:latest) results error in processing aiohttp


yeahdongcn:
**Description**
Install mlflow-triton plugin in mlflow container (ghcr.io/mlflow/mlflow:latest) results error in processing aiohttp.

```bash
Searching for aiohttp>=3.8.1
Reading https://pypi.org/simple/aiohttp/
Downloading https://files.pythonhosted.org/packages/f0/f2/703dba52c7620199ea0ec8ea9a4a2f06203b4893b94f60240c2c10225043/aiohttp-4.0.0a1.tar.gz#sha256=b5036133c1ba77ed5a70208d2a021a90b76fdf8bf523ae33dae46d4f4380d86f
Best match: aiohttp 4.0.0a1
Processing aiohttp-4.0.0a1.tar.gz
Writing /tmp/easy_install-2afuxs6x/aiohttp-4.0.0a1/setup.cfg
Running aiohttp-4.0.0a1/setup.py -q bdist_egg --dist-dir /tmp/easy_install-2afuxs6x/aiohttp-4.0.0a1/egg-dist-tmp-26rpablz
...
```

This should be related to https://github.com/aio-libs/aiohttp/issues/6898 as ghcr.io/mlflow/mlflow:latest uses python 3.10.

**Triton Information**
What version of Triton are you using?
main.

Are you using the Triton container or did you build it yourself?
Build by myself.

**To Reproduce**
Steps to reproduce the behavior.
docker run -d -p 5000:5000 -v $PWD:/plugin ghcr.io/mlflow/mlflow mlflow server --host=0.0.0.0 ($PWD is deploy/mlflow-triton-plugin)
docker exec -w /plugin -t <<contaienr_id>> python setup.py install

Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).

**Expected behavior**
No error reports

PR #5667, opened 2023-04-20
Pin aiohttp to 3.8.4 to avoid installing error in Python 3.10

yeahdongcn:
Fix: #5666

### DIFF
--- a/deploy/mlflow-triton-plugin/setup.py
+++ b/deploy/mlflow-triton-plugin/setup.py
@@ -32,6 +32,6 @@ setup(
     long_description=open("README.md").read(),
     long_description_content_type="text/markdown",
     packages=find_packages(),
-    install_requires=["mlflow>=2.2.1,<3.0", "tritonclient[all]", "boto3"],
+    install_requires=["mlflow>=2.2.1,<3.0", "tritonclient[all]", "boto3", "aiohttp==3.8.4"],
     entry_points={"mlflow.deployments": "triton=mlflow_triton.deployments"},
 )


deploy/mlflow-triton-plugin/setup.py
```diff
@@ -32,6 +32,6 @@
     long_description=open("README.md").read(),
     long_description_content_type="text/markdown",
     packages=find_packages(),
-    install_requires=["mlflow>=2.2.1,<3.0", "tritonclient[all]", "boto3"],
+    install_requires=["mlflow>=2.2.1,<3.0", "tritonclient[all]", "boto3", "aiohttp==3.8.4"],
```
kthui, line 35: The `aiohttp` requirement is set in the client [requirements_http.txt](https://github.com/triton-inference-server/client/blob/main/src/python/library/requirements/requirements_http.txt#L28). Could you make the change over there?

I think it would be better to allow a small range of acceptable version, like `aiohttp>=3.8.1,<4.0.0`.

yeahdongcn: Please see the new one (https://github.com/triton-inference-server/client/pull/297) on `client` side. Let's close this one.

Can a new version of tritonclient be released to pypi soon? I'd like to use `mlflow-triton` in Python 3.10. Thanks.

yeahdongcn:
> Hi @yeahdongcn, thanks for your contribution. Could you sign and email us the [Contributor License Agreement (CLA)](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) so we can use your PR?

Yes, I've signed the CLA in https://github.com/triton-inference-server/pytorch_backend/pull/91

PR #5669, opened 2023-04-20
Add testing for Python custom metrics API

krishung5:
Python Backend: https://github.com/triton-inference-server/python_backend/pull/235

### DIFF
--- a/qa/python_models/custom_metrics/config.pbtxt
+++ b/qa/python_models/custom_metrics/config.pbtxt
@@ -35,4 +35,9 @@ output [
   }
 ]
 
-instance_group [{ kind: KIND_CPU }]
+instance_group [
+  { 
+    count: 3
+    kind: KIND_CPU 
+  }
+]
--- a/qa/python_models/custom_metrics/config.pbtxt
+++ b/qa/python_models/custom_metrics/config.pbtxt
@@ -1,0 +1,38 @@
+# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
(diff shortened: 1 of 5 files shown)


qa/python_models/custom_metrics/model.py
```diff
@@ -0,0 +1,215 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import numpy as np
+import unittest
+import triton_python_backend_utils as pb_utils
+import requests
+
+
+class PBBLSMemoryTest(unittest.TestCase):
```
Tabrizian, line 33: I think the test needs to be renamed.

Tabrizian (commented):
One minor comment, otherwise looks great!

qa/python_models/custom_metrics/model.py
krishung5: Good catch! Fixed.

qa/python_models/custom_metrics/model.py
```diff
@@ -0,0 +1,215 @@
...
+                description=
+                "test metric family with same name but different kind",
+                kind=pb_utils.GAUGE)
+            self.assertIsNone(metric_family2)
+
+        self.assertIsNotNone(metric_family1)
+
+    def test_dup_metric_family_diff_description(self):
+        # Test that a duplicate metric family name will still return the
+        # original metric family even if the description is changed
```
oandreeva-nv, line 138: Do we report this in logs? Something like: "Warning: duplicate metric family name is used, ignoring" or this is unnecessary? And do we want to log that?

oandreeva-nv:
Nice work! Had just one question.

qa/python_models/custom_metrics/model.py
rmccorm4: I don't think [MetricFamily](https://github.com/triton-inference-server/core/blob/ecbe0230d5770481d23c5f1f984ac6fac651e1df/src/metric_family.cc#L38-L62) objects are currently aware of each other, and prometheus doesn't throw any error on a (dupe name + different description) currently. 

We could probably add some static class map of name -> description if we wanted to catch this, but for now I think its OK as is.

Tabrizian (dismissed):

qa/python_models/custom_metrics/config.pbtxt
```diff
@@ -0,0 +1,38 @@
...
+
+output [
+  {
+    name: "OUTPUT0"
+    data_type: TYPE_FP32
+    dims: [ 16 ]
+  }
+]
+
+instance_group [{ kind: KIND_CPU }]
```
rmccorm4, line 38: As a sanity check, can you see if everything runs OK with an instance count > 1? Since each model instance should be calling the same MetricFamily/Metric constructors/etc. - this should also test the duplicate/reference behavior.

krishung5: Nice thoughts! Increased the instance count.

Tabrizian (approved):

Tabrizian (approved):

rmccorm4 (approved):

PR #5670, opened 2023-04-20
Give sufficient time to get inference statistics for sequence tests

dyastremsky:
Since responses are sent before inference statistics are reported, there is the risk that statistics are checked before they are ready. This may be the cause of L0_sequence_batcher failing infrequently with statistics missing one request, with messages like:

```
Traceback (most recent call last):
  File "sequence_batcher_test.py", line 577, in test_no_sequence_end
    self.check_status(model_name, {1: 4 * (idx + 1)},
AssertionError: 7 != 8 : expected model-execution-count 8 for batch size 1, got 7
```

Only in the case that sequence batcher is going to fail due to not having enough requests ready, this change will provide up to 10 seconds in .5 second increments for the requests to be ready.

### DIFF
--- a/qa/L0_sequence_batcher/test.sh
+++ b/qa/L0_sequence_batcher/test.sh
@@ -112,7 +112,7 @@ else
     fi
 fi
 
-SERVER_ARGS_EXTRA="--backend-directory=${BACKEND_DIR} --backend-config=tensorflow,version=${TF_VERSION}"
+SERVER_ARGS_EXTRA="--backend-directory=${BACKEND_DIR} --backend-config=tensorflow,version=${TF_VERSION} --log-verbose=1"
 
 source ../common/util.sh
 
--- a/qa/common/sequence_util.py
+++ b/qa/common/sequence_util.py
@@ -986,9 +986,9 @@ class SequenceBatcherTestUtil(tu.TestResultCollector):
     def check_status(self, model_name, batch_exec, exec_cnt, infer_cnt):
         start_time = time.time()
         # There is a time window between when responses are returned and statistics are updated.
-        # To prevent intermittent test failure during that window, wait up to 5 seconds for the
+        # To prevent intermittent test failure during that window, wait up to 10 seconds for the
         # inference statistics to be ready.
-        num_tries = 10
+        num_tries = 20
         for i in range(num_tries):
             stats = self.triton_client_.get_inference_statistics(
                 model_name, "1")
--- a/qa/common/sequence_util.py
+++ b/qa/common/sequence_util.py
@@ -984,8 +984,17 @@ class SequenceBatcherTestUtil(tu.TestResultCollector):
                              _max_sequence_idle_ms * 1000)  # 5 secs
 
     def check_status(self, model_name, batch_exec, exec_cnt, infer_cnt):
-        stats = self.triton_client_.get_inference_statistics(model_name, "1")
-        self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
+        start_time = time.time()
+        # Wait up to 10 seconds for the inference statistics to be ready
+        while (time.time() - start_time) < 10:
+            stats = self.triton_client_.get_inference_statistics(
+                model_name, "1")
+            self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
+            actual_exec_cnt = stats.model_stats[0].execution_count
+            if actual_exec_cnt == exec_cnt:
+                break
+            time.sleep(0.5)
+
         self.assertEqual(stats.model_stats[0].name, model_name,
                          "expect model stats for model {}".format(model_name))
         self.assertEqual(
--- a/qa/common/sequence_util.py
+++ b/qa/common/sequence_util.py
@@ -988,7 +988,7 @@ class SequenceBatcherTestUtil(tu.TestResultCollector):
         # There is a time window between when responses are returned and statistics are updated.
         # To prevent intermittent test failure during that window, wait up to 10 seconds for the
         # inference statistics to be ready.
-        num_tries = 20
+        num_tries = 10
         for i in range(num_tries):
             stats = self.triton_client_.get_inference_statistics(
                 model_name, "1")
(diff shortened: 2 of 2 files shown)


dyastremsky:
> I'd be surprised if the window between returning response and reporting statistics is really small enough for this to happen often.
> 
> On top of that, if it is just a timing issue, why is it so consistently off by 1? ex: `0 != 1`, `7 != 8`, `15 != 16` - that feels suspicious to me like more of a logic issue or internal race condition. Is there any reason why it's not off by 2 or 3 sometimes?
> 
> However, if it was in fact a logic/race condition issue (ex: the stats are just getting set incorrectly) and not a timing issue, then I suppose the test would still periodically fail even with this PR's change.
> 
> 1. I'd rather some more definitive proof via some backtraces or prints and timestamps in what backend is reporting vs what core is receiving, etc. -- but if you can't reproduce locally then I'm not sure we can do much. Maybe we can add a LOG_VERBOSE (3) around some suspected areas and run the test with that log level until it fails again? What do you think @GuanLuo
> 2. Separately, if this sleep/retry change goes through, can you add some printing/logging from the python test indicating each retry @dyastremsky ? Would be some additional information in the log that may help in the future, ex: if it is actually passing because of a retry or not.

Even if it does not resolve this issue, it seems prudent to have some sort of waiting to ensure the statistics are ready. I added the logging [here](https://github.com/triton-inference-server/server/pull/5670/commits/b295c1edf7a3a68ba7abe37cbc5cab4b583054b8). It's true that in the runs I've seen, it's always an off-by-one error and I wouldn't expect it to be as consistent if this was a timing issue. But this seemed like a good change to start with..

It's possible it would reproduce locally and it just hasn't hit it after a dozen runs yet. This test is a bit challenging as its runtime is just over an hour and it seems to hit this case <5% of runs. An additional option is enabling verbose logs so that if it fails again, we'll have verbose logs.

GuanLuo:
I agree that having the timestamp can be useful to understand the root cause of the issue.

Without further information, however, I think off-by-1 is more an indicator of timing issue than logical error. The client (test) will proceed once all responses are received and the batch stats will only be updated after all responses are sent on the server side, which is why 1-off may be observed. That being said, it is less likely that the network time for getting stats is less than the time for updating it, unless it is running on resource constrained system and the server thread is preempted.

qa/common/sequence_util.py
```diff
@@ -984,8 +984,19 @@ def check_setup(self, model_name):
...
+        start_time = time.time()
+        # Wait up to 10 seconds for the inference statistics to be ready
+        while (time.time() - start_time) < 10:
+            stats = self.triton_client_.get_inference_statistics(
+                model_name, "1")
+            self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
+            actual_exec_cnt = stats.model_stats[0].execution_count
+            if actual_exec_cnt == exec_cnt:
+                break
+            print("Waiting: expect {} executions, got {}".format(exec_cnt,
```
rmccorm4, line 996: Can you add an attempt number and some kind of "ERROR/WARNING" or similar so it's easier to spot in the logs?

qa/common/sequence_util.py
```diff
@@ -984,8 +984,19 @@ def check_setup(self, model_name):
                              _max_sequence_idle_ms * 1000)  # 5 secs
 
     def check_status(self, model_name, batch_exec, exec_cnt, infer_cnt):
-        stats = self.triton_client_.get_inference_statistics(model_name, "1")
-        self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
+        start_time = time.time()
+        # Wait up to 10 seconds for the inference statistics to be ready
```
rmccorm4, line 988: Add more details in comment about movitivation, ex: time window between response return and statistics update / test failure intermittence

rmccorm4 (commented):
LGTM, minor comments for future bookkeeping

qa/L0_sequence_batcher/test.sh
```diff
@@ -112,7 +112,7 @@ else
     fi
 fi
 
-SERVER_ARGS_EXTRA="--backend-directory=${BACKEND_DIR} --backend-config=tensorflow,version=${TF_VERSION}"
+SERVER_ARGS_EXTRA="--backend-directory=${BACKEND_DIR} --backend-config=tensorflow,version=${TF_VERSION} --log-verbose=1"
```
dyastremsky, line 115: Updating verbose logging so that we have more information if this fails again. SERVER_ARGS_EXTRA is used every time, so putting this here seemed to make more sense than putting it in five places.

qa/common/sequence_util.py
dyastremsky: Good idea. Updated.

qa/common/sequence_util.py
dyastremsky: Sure, done!

qa/common/sequence_util.py
```diff
@@ -985,16 +985,19 @@ def check_setup(self, model_name):
...
             stats = self.triton_client_.get_inference_statistics(
                 model_name, "1")
             self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
             actual_exec_cnt = stats.model_stats[0].execution_count
             if actual_exec_cnt == exec_cnt:
                 break
-            print("Waiting: expect {} executions, got {}".format(exec_cnt,
-                                                          actual_exec_cnt))
+            print("WARNING: expect {} executions, got {} (attempt {})".format(
+                exec_cnt, actual_exec_cnt, loop_count))
```
rmccorm4, line 1000: increment loop count? Alternatively just use a fixed number of iterations:
```
num_tries = 10
for i in range(num_tries):
  ...
  sleep(0.5)
```

dyastremsky: Nice catch, switched to using the loop count incrementer.

qa/common/sequence_util.py
```diff
@@ -986,18 +986,18 @@ def check_setup(self, model_name):
     def check_status(self, model_name, batch_exec, exec_cnt, infer_cnt):
         start_time = time.time()
         # There is a time window between when responses are returned and statistics are updated.
-        # To prevent intermittent test failure during that window, wait up to 10 seconds for the
+        # To prevent intermittent test failure during that window, wait up to 5 seconds for the
```
rmccorm4, line 989: You can keep it as 10 seconds by increasing sleep time. Might as well have more time for it to pass if we're doing the workaround.

dyastremsky: Sure, updated.

rmccorm4 (approved):

PR #5672, opened 2023-04-20
Make model dir for L0_perf_nomodel

dyastremsky:
The test relies on this testing directory. If it does not yet exist, create it.


dyastremsky:
Unnecessary. This is not why the current CI is failing, and this change already exists in the Docker version (where it's needed).

PR #5697, opened 2023-04-25
Update README and add RELEASE notes for 23.04

mc-nv:


Tabrizian (approved):

RELEASE.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Traced models in PyTorch seem to create overflows when int8 tensor values are 
+  transformed to int32 on the GPU. 
+
+  Refer to https://github.com/pytorch/pytorch/issues/66930 for more information.
+
+* Triton cannot retrieve GPU metrics with MIG-enabled GPU devices (A100 and A30).
+
+* Triton metrics might not work if the host machine is running a separate DCGM 
+  agent on bare-metal or in a container.
```
nv-kmcgill53, line 109: Add a note saying,

"PyTorch backend for Jetson uses the 23.03 release of PyTorch." 

@tanmayv25 to confirm verbage

mc-nv: I don't think that we need to add this info here.
I will mention in release notes that we are using proper Torch version
https://docs.nvidia.com/deeplearning/frameworks/install-pytorch-jetson-platform-release-notes/pytorch-jetson-rel.html

PR #5699, opened 2023-04-25
Give sufficient time to get inference statistics for batcher tests

dyastremsky:
L0_batcher occasionally fails with a similar error to that seen in L0_sequence_batcher seen [here](https://github.com/triton-inference-server/server/pull/5670). Proposing a similar fix to ensure the test does not check statistics before they are ready.


rmccorm4 (approved):
If we expand this to more tests, we should consider re-using some utility function instead

PR #5706, opened 2023-04-26
Add testing for Triton Client Plugin API

Tabrizian:
Client PR: https://github.com/triton-inference-server/client/pull/301


Dockerfile.QA
```diff
@@ -99,7 +99,6 @@ RUN mkdir -p qa/common && \
     cp -r docs/examples/model_repository/simple qa/L0_cuda_shared_memory/models/. && \
     mkdir qa/L0_client_java/models && \
     cp -r docs/examples/model_repository/simple qa/L0_client_java/models && \
-    mkdir qa/L0_grpc/models && \
     cp -r docs/examples/model_repository/simple qa/L0_grpc/models && \
```
nnshah1, line 102: Not sure - but this seems like it might still be needed?

Tabrizian: Good catch! Fixed.

qa/L0_grpc/grpc_client_plugin_test.py
```diff
@@ -0,0 +1,118 @@
...
+        inputs = prepare_infer_inputs({})
+        self._client.unregister_plugin()
+        response = self._client.infer(model_name=model, inputs=inputs)
+        test_success = response.as_numpy('TEST_SUCCESS')
+        self.assertEqual(test_success, True)
+
+    def tearDown(self):
+        self._client.close()
+
+
```
GuanLuo, line 116: Should have test to register multiple plugins and only the latest one will be used. Also may be a good idea to have API to return the registered plugin or None from the client that the user may check which plugin instance is registered.

qa/L0_http/http_client_plugin_test.py
```diff
@@ -0,0 +1,169 @@
...
+    def setUp(self):
+        self._headers = {'MY-KEY': 'MY-VALUE'}
+        self._plugin = TestPlugin(self._headers)
+        self._client = tritonhttpclient.InferenceServerClient(
+            url='localhost:8001')
+
+        # Use magic mock for the client stub
+        self._client._client_stub = MagicMock()
+
+    def test_server_is_live(self):
```
GuanLuo, line 123: Why we should test "server is live" here?

Tabrizian: The test is added to test a `GET` and `POST` method. I can add a comment to clarify.

GuanLuo (approved):

PR #5719, opened 2023-04-29
Check TensorRT error message more granularly

kthui:
It is observed that TensorRT might update their internal error messages. This PR is to enable checking TensorRT error message only on its critical parts, rather than trying to match the entire TensorRT error message.

### DIFF
--- a/qa/L0_trt_error_propagation/trt_error_propagation_test.py
+++ b/qa/L0_trt_error_propagation/trt_error_propagation_test.py
@@ -40,12 +40,23 @@ class TestTrtErrorPropagation(unittest.TestCase):
         with self.assertRaises(InferenceServerException) as cm:
             self.__triton.load_model("invalid_plan_file")
         err_msg = str(cm.exception)
-        self.assertIn("Internal: unable to create TensorRT engine", err_msg,
-                      "Caught an unexpected exception")
+        # "[err_msg]" -> "[triton_err_msg]: [backend_err_msg]: [trt_err_msg]"
+        backend_err_msg = "Internal: unable to create TensorRT engine: "
         self.assertIn(
-            "Error Code 4: Internal Error (Engine deserialization failed.)",
-            err_msg,
-            "Detailed error message not propagated back to triton client")
+            backend_err_msg, err_msg,
+            "Cannot find the expected error message from TensorRT backend")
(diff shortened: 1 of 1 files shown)


qa/L0_trt_error_propagation/trt_error_propagation_test.py
```diff
@@ -40,12 +40,23 @@ def test_invalid_trt_model(self):
         with self.assertRaises(InferenceServerException) as cm:
             self.__triton.load_model("invalid_plan_file")
         err_msg = str(cm.exception)
-        self.assertIn("Internal: unable to create TensorRT engine", err_msg,
-                      "Caught an unexpected exception")
+        # "[err_msg]" -> "[triton_err_msg]: [backend_err_msg]: [trt_err_msg]"
+        backend_err_msg = "Internal: unable to create TensorRT engine: "
```
tanmayv25, line 44: Can you use prefix terms like `expected_` and `got_` to make logic more readable?

kthui: Yes. Update commit: [Clarify source of error messages](https://github.com/triton-inference-server/server/pull/5719/commits/f4cbf55bb578b2529c7460aedae0c8cc2d1d414d)

kthui:
Rebased.

dyastremsky (dismissed):

dyastremsky (commented):
Great work on cleaning this up! Do we need to be testing everything that we're testing with the changes? I feel like we may be testing more than we need to be.

We can leave it if you think we're testing what we need to be here, but it seems like the tokenization isn't needed. Right now, you expect the error message to be A -> B -> C. So you test B, then you test that the part before B has and the part after has C. It might be more intuitive to just check if all the strings are in the message. If the ordering matters for testing, then you could do something along the lines of the below.

```for expected_msg in expected_msgs:

assertIn(expected_msg, msg)

_, msg = msg.split(expected_msg)```

kthui:
Good idea! Updated: [Consolidate tests for message parts](https://github.com/triton-inference-server/server/pull/5719/commits/999ed802225aad9e0f44e2d1317c37d107facbe5)

dyastremsky (approved):
Great work! This looks so clean.

PR #5723, opened 2023-05-01
Update note on SageMaker MME support for ensemble

casassg:
Small note update in Sagemaker integration. Note explains that Ensemble model is not supported. However code does parse for ensemble correctly and seems to support it as far as the code goes.

Note caused some confusion which led to us having to parse more code to understand what was the behaviour of sagemaker MME when using triton for ensemble models. 

I wrote some update on note to clarify actual code behaviour.


GuanLuo (approved):
Thanks for your contribution, the inline documentation update looks good to me.

PR #5724, opened 2023-05-01
Enhance tests in L0_model_update

kthui:
Changes: 
- Add test(s) for model instance name update. [Add model instance name update test](https://github.com/triton-inference-server/server/pull/5724/commits/a14ad288a8b4b5d529e22c8f4c6c9ce4953c90fa)
- Any changes to a file on disk should be fully written to disk before the modification function returns. [Add gap for timestamp to update](https://github.com/triton-inference-server/server/pull/5724/commits/e336d037ab93b9db2a41d3f0052cac0a02f849d8) (Passed 100+ iterations)
- For supported test cases, the test should cover starting the server with rate limiter off. [Extend supported test on rate limit off](https://github.com/triton-inference-server/server/pull/5724/commits/2fc3a18ce0bbc1c3913e8a933b98aee2bf947966)
- Add some tests with dynamic batching on. [Add some tests with dynamic batching](https://github.com/triton-inference-server/server/pull/5724/commits/97426939ebd44c6a0ff05388ca566c53209330e3)

### DIFF
--- a/qa/L0_model_update/test.sh
+++ b/qa/L0_model_update/test.sh
@@ -53,33 +53,42 @@ function setup_models() {
         mv models/model_init_del/model.py models/model_init_del/1
 }
 
-SERVER_ARGS="--model-repository=models --model-control-mode=explicit --rate-limit=execution_count --log-verbose=2"
-SERVER_LOG="./server.log"
-setup_models
-run_server
-if [ "$SERVER_PID" == "0" ]; then
-    echo -e "\n***\n*** Failed to start $SERVER\n***"
-    cat $SERVER_LOG
-    exit 1
-fi
+# Test model instance update with and without rate limiting enabled
+for RATE_LIMIT_MODE in "off" "execution_count"; do
 
-RET=0
+    export RATE_LIMIT_MODE=$RATE_LIMIT_MODE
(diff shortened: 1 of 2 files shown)


qa/L0_model_update/test.sh
```diff
@@ -53,33 +53,42 @@ function setup_models() {
...
+        echo -e "\n***\n*** Test FAILED\n***"
+        RET=1
+    fi
+    set -e
+
+    kill $SERVER_PID
+    wait $SERVER_PID
+
+    if [ $RET -ne 0 ]; then
+        exit $RET
```
tanmayv25, line 87: Why are we exiting on the first failure? 
Shouldn't we still execute "execution_count" mode even if the "off" mode failed?

kthui: > Why are we exiting on the first failure?

This is to make finding the location of the failure easier on the CI, because it is not covered by other tests.

 > Shouldn't we still execute "execution_count" mode even if the "off" mode failed?
 
 Yes, we can continue execute if "off" mode failed.

kthui: Updated: [Continue test if off mode failed](https://github.com/triton-inference-server/server/pull/5724/commits/e63f7d40f01650a44f522caa0865820e040b8c1c)

tanmayv25 (approved):

PR #5728, opened 2023-05-02
Fix L0_backend_python on Jetson 

dyastremsky:
- Removed shm monitor for Jetson devices, since Jetson devices may clean up backend memory growth slower and result in false positives (e.g. in L0_backend_python_jetson)
- Switched the print order of L0_backend_python to avoid printing the SERVER_LOG last in situations where it may be confusing

### DIFF
--- a/qa/common/shm_util.py
+++ b/qa/common/shm_util.py
@@ -35,6 +35,7 @@ from tritonclient.utils import *
 # By default, find tritonserver on "localhost", but can be overridden
 # with TRITONSERVER_IPADDR envvar
 _tritonserver_ipaddr = os.environ.get('TRITONSERVER_IPADDR', 'localhost')
+_test_jetson = bool(int(os.environ.get('TEST_JETSON', 0)))
 
 
 def _range_repr_dtype(dtype):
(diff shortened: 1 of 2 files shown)


Tabrizian (dismissed):

qa/common/shm_util.py
```diff
@@ -388,4 +395,9 @@ def __init__(self, prefix='triton_python_backend_shm_region'):
                     triton_shm_monitor.SharedMemoryManager(shm_region))
 
     def Probe(self):
-        return self.ShmLeakProbe(self._shm_monitors)
+        # Jetson cleanup takes too long and results in false positives.
+        # Do not use the shared memory check on Jetson.
```
nv-kmcgill53, line 399: I personally would like to see this get re-enabled at some point as shared memory can be platform dependent. Can we put a `FIXME` and reference a ticket to track this?

qa/common/shm_util.py
```diff
@@ -379,6 +384,8 @@ def __exit__(self, type, value, traceback):
             assert not shm_leak_detected, "Shared memory leak detected."
 
     def __init__(self, prefix='triton_python_backend_shm_region'):
+        if TEST_JETSON:
```
nv-kmcgill53, line 387: Should this be `_test_jetson` instead?

qa/common/shm_util.py
dyastremsky: Good suggestion, added.

qa/common/shm_util.py
dyastremsky: Good catch, fixed.

nv-kmcgill53 (approved):

PR #5739, opened 2023-05-05
Add testing for basic auth plugin for HTTP/gRPC clients

Tabrizian:
Client: https://github.com/triton-inference-server/client/pull/310


nnshah1 (approved):
there's a fair bit of repeated code / patterns here - but not sure if there is a good way to consolidate - maybe something for the future.

krishung5 (changes requested):
I think we need to address the duplicated and unused imports raised by CodeQL. LGTM otherwise.

krishung5 (approved):

PR #5756, opened 2023-05-09
Fix error handle bug in ModelStreamInferHandler::Process

kisisjrlly:
In the original process, if the error is not null, TRITONSERVER_ The InferenceRequestId will still be executed, causing triton server to crash


nnshah1:
@kisisjrlly 

Thanks for this pull request - we found the same issue and applied a similar fix in (only matching the corresponding logic in infer_handler):

https://github.com/triton-inference-server/server/pull/5842 

Appreciate the contribution!

nnshah1:
Thanks again - fix applied - closing this as duplicate.

PR #5775, opened 2023-05-11
Fix L0_custom_ops

nv-kmcgill53:
Fix L0_custom_ops for ubuntu 22.04.

We have upgraded to python 3.10 from python 3.8. the .so file needed to be updated.


rmccorm4 (approved):

rmccorm4:
@nv-kmcgill53 can you add the same fix to L0_storage_custom_ops in gitlab?

PR #5776, opened 2023-05-11
Fix L0_http_fuzz

nv-kmcgill53:
Adding back in pip install of boofuzz library into the devel container. 

Git blame on the main branch says this was added 2 years ago. @mc-nv is there a reason why this was deleted before?

### DIFF
--- a/qa/L0_http_fuzz/test.sh
+++ b/qa/L0_http_fuzz/test.sh
@@ -55,6 +55,47 @@ SERVER=/opt/tritonserver/bin/tritonserver
 SERVER_ARGS="--model-repository=$DATADIR"
 source ../common/util.sh
 
+# Remove this once foobuzz and tornado packages upgrade to work with python 3.10
+# This test tests the server's ability to handle poor input and not the compatibility 
+# with python 3.10. Python 3.8 is ok to use here.
+function_install_python38() {
+    apt-get update
+    apt-get remove -y python3
+    wget https://www.python.org/ftp/python/3.8.16/Python-3.8.16.tar.xz
+
+    # md5sum is not secure. Only use for sanity check
+    MD5SUM_PYTHON38=$(md5sum Python-3.8.16.tar.xz | awk '{ print $1 }' -)
+    CORRECT_MD5SUM_PYTHON38=621ac153586a3152e2ab7d3a8614df9a
+    if [ "$MD5SUM_PYTHON38" != "$CORRECT_MD5SUM_PYTHON38" ]; then
+        echo "md5sum of downloaded Python-3.8.16.tar.xz does not match! $MD5SUM_PYTHON38 != $CORRECT_MD5SUM_PYTHON38"
+        RET=1
+    fi 
+
+    # check the file size as well
+    FILE_SIZE_PYTHON38=$(ls -l Python-3.8.16.tar.xz | awk '{ print $5 }' -)
+    CORRECT_FILE_SIZE_PYTHON38=19046724
+    if [ "$FILE_SIZE_PYTHON38" != "$CORRECT_FILE_SIZE_PYTHON38" ]; then
+        echo "file size is not correct! $FILE_SIZE_PYTHON38 != $ $CORRECT_FILE_SIZE_PYTHON38" 
+        RET=1
+    fi
+    echo "Validated md5sum and file size of Python-3.8.16.tar.xz"
+
+    # Unpack pythonn and install 
+    tar -xf Python-3.8.16.tar.xz
+    cd Python-3.8.16
+    apt-get install -y libsqlite3-dev libffi-dev
+    ./configure --enable-loadable-sqlite-extensions
+    make 
+    make install
+
+    # Install test script dependencies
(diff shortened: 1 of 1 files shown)


rmccorm4:
Why don't we install this in `L0_http_fuzz/test.sh` instead? I think we install most other test-specific packages that way.

nv-kmcgill53:
>Why don't we install this in L0_http_fuzz/test.sh instead? 

Can do. It was installed into the QA container before, so I was putting back. No other reason. I can move this into `L0_http_fuzz/test.sh`

Dockerfile.QA
```diff
@@ -339,6 +339,11 @@ RUN pip3 install --upgrade wheel setuptools && \
     pip3 install --upgrade numpy pillow attrdict future grpcio requests gsutil \
                            awscli six grpcio-channelz prettytable virtualenv
 
+# L0_http_fuzz is hitting similar issue with boofuzz with latest version (0.4.0):
+# https://github.com/jtpereyda/boofuzz/issues/529
+# Hence, fixing the boofuzz version to 0.3.0
+RUN pip3 install 'boofuzz==0.3.0'
```
mc-nv, line 345: It causing en issue because it's trying to install `blinker` version `1.6` which previously installed with `apt install software-properties-common`  package that has version `1.4`. Unfortunately `pip` utility unable to upgraded OS installed packages.

nv-kmcgill53:
Updated this PR to install python3.8 instead of attempting to resolve python 3.10 issues with `boofuzz` dependency. The purpose of this test is to give Tritonserver bad inputs and ensure it handles them appropriately; thus, this test is python version agnostic.

qa/L0_http_fuzz/test.sh
```diff
@@ -55,6 +55,47 @@ SERVER=/opt/tritonserver/bin/tritonserver
...
+    # check the file size as well
+    FILE_SIZE_PYTHON38=$(ls -l Python-3.8.16.tar.xz | awk '{ print $5 }' -)
+    CORRECT_FILE_SIZE_PYTHON38=19046724
+    if [ "$FILE_SIZE_PYTHON38" != "$CORRECT_FILE_SIZE_PYTHON38" ]; then
+        echo "file size is not correct! $FILE_SIZE_PYTHON38 != $ $CORRECT_FILE_SIZE_PYTHON38" 
+        RET=1
+    fi
+    echo "Validated md5sum and file size of Python-3.8.16.tar.xz"
+
+    # Unpack pythonn and install
```
nnshah1, line 83:
```suggestion
    # Unpack python and install 
```

qa/L0_http_fuzz/test.sh
```diff
@@ -55,6 +55,47 @@ SERVER=/opt/tritonserver/bin/tritonserver
...
+    fi
+    echo "Validated md5sum and file size of Python-3.8.16.tar.xz"
+
+    # Unpack pythonn and install 
+    tar -xf Python-3.8.16.tar.xz
+    cd Python-3.8.16
+    apt-get install -y libsqlite3-dev libffi-dev
+    ./configure --enable-loadable-sqlite-extensions
+    make 
+    make install
```
Tabrizian, line 89: Instead of compiling Python from source, you can also just create a conda environment. we have some scripts to make it very easy: https://github.com/triton-inference-server/server/blob/725655e5168c1a83c34f9dbcfd9aa91f3854696d/qa/L0_backend_python/env/test.sh#L88

qa/L0_http_fuzz/test.sh
```diff
@@ -55,6 +55,47 @@ SERVER=/opt/tritonserver/bin/tritonserver
...
+
+    # Install test script dependencies
+    pip3 install --upgrade wheel setuptools boofuzz==0.3.0 numpy pillow attrdict future grpcio requests gsutil \
+                            awscli six grpcio-channelz prettytable virtualenv
+}
+WORKING_DIR=`pwd`
```
nnshah1, line 95: Small suggestion - maybe this is `pushd`, `popd` in the function itself - keeping the function self contained with regards to directories.

nnshah1 (dismissed):

qa/L0_http_fuzz/test.sh
nv-kmcgill53: I'm not very familiar with conda. Does this create the environment local to the bash instance or does it install it into the filesystem?

```
create_conda_env() {
  python_version=$1
  env_name=$2
  conda create -n $env_name python=$python_version -y
  conda activate $env_name
  conda install conda-pack -y
}
```

Tabrizian: It installs it in the filesystem.

Tabrizian (approved):

qa/L0_http_fuzz/test.sh
nv-kmcgill53: Just had to read up on this. Will use this in the future. With [@Tabrizian's suggestion](https://github.com/triton-inference-server/server/pull/5776#discussion_r1196787856) I don't think this will be needed. Will wait for the CI to finish though.

rmccorm4 (approved):

PR #5795, opened 2023-05-16
Fix L0_memory_growth

rmccorm4:
1. reduce MAX_ALLOWED_ALLOC to 100MB to be more strict for bounded tests, and generous for unbounded tests. Many failures were due to the unbounded test only reaching 110-130MB of usage, not > 150MB. Also, the bounded tests are typically only using 20-70MB depending on framework, so I think it's valuable to make the bounds tighter to catch growths/regressions.
2. allow unstable measurement from PA: same old here
3. improve logging for future triage: try to add memory usage to job log for every test so it's easier to see at a glance what each framework is roughly using if this test has to be modified in the future.

### DIFF
--- a/qa/L0_memory_growth/test.sh
+++ b/qa/L0_memory_growth/test.sh
@@ -92,8 +92,12 @@ else
 fi
 
 # Threshold memory growth in MB
-MAX_ALLOWED_ALLOC="150"
-export MAX_ALLOWED_ALLOC
+# NOTES:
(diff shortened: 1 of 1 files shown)


nv-kmcgill53 (approved):

nv-kmcgill53:
LGTM. It looks like the test could take some time to execute, but assuming the test passes in the CI, this looks fine.

nnshah1 (approved):
LGTM

PR #5805, opened 2023-05-17
Fix L0_mlflow

nv-kmcgill53:
This PR fixes mlflow failing when installing python test dependencies. The `python3-blinker` default installed version is `1.4` but the python dependencies need version `1.6` installed. Removing the default installed version of blinker which comes with Ubuntu 22.04 allows pip to install the necessary dependencies.

### DIFF
--- a/qa/L0_mlflow/test.sh
+++ b/qa/L0_mlflow/test.sh
@@ -31,6 +31,16 @@ source ../common/util.sh
 
 rm -fr *.log *.json
 
+# The default version of python 3.10.6 included in
+# Ubuntu 22.04 installs blinker 1.4. This doesn't 
+# work with the awscli which we try to install. 
+# Uninstalling blinker and allowing pip to install blinker 1.6 
+# fixes this issue. The alternative to this is to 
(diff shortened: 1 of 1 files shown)


qa/L0_mlflow/test.sh
```diff
@@ -31,8 +31,22 @@ source ../common/util.sh
...
+# install a higher version of python which uses blinker 1.6,
+# but it is unknown whether this test should rely on 
+# the default installation of python.
+apt remove -y python3-blinker
+
 RET=0
 
+# Uninstall the python 3.10 version of blinker. This 
+# causes other dependencies to not install
+apt-get remove python3-blinker
```
mc-nv, line 48: Is it necessary to keep same instruction repetitive ?

nv-kmcgill53: This was a `git pull` issue.

mc-nv (approved):

Tabrizian (approved):

GuanLuo (approved):

PR #5806, opened 2023-05-17
Fix client script for L0_libtorch_inference_mode

krishung5:
The client script used in the test was not set properly.


mc-nv (approved):

Tabrizian (approved):
Great find!

PR #5807, opened 2023-05-17
Fix L0_mlflow

nv-kmcgill53:
Issue here was that `apt remove` was terminating the script early because it couldn't find `python3-blinker` in the repository.

### DIFF
--- a/qa/L0_mlflow/test.sh
+++ b/qa/L0_mlflow/test.sh
@@ -39,6 +39,7 @@ rm -fr *.log *.json
 # install a higher version of python which uses blinker 1.6,
 # but it is unknown whether this test should rely on 
 # the default installation of python.
+apt-get update
 apt remove -y python3-blinker
 
 RET=0


rmccorm4 (commented):
Did this fix it locally? I didn't think an apt update was needed to be able to remove an already installed package. I thought it just updated the indexes/sources used to discover new repositories/packages (or new versions of currently installed)

nv-kmcgill53:
> Did this fix it locally? I didn't think an apt update was needed to be able to remove an already installed package. I thought it just updated the indexes/sources used to discover new repositories/packages (or new versions of currently installed)

It fixed it in the pipeline. The issue is that it returns a non-zero exit code and terminates the test script. You are correct that if you `exec` into the container and run the test, the error will print but the test will continue and succeed. This is something to do with the `docker run` command which I don't fully understand.

rmccorm4:
Hm, I'm not sure if that's related to docker run vs exec. When you exec and say it prints error and passes, how are you running the test?

`bash test.sh` or `bash -ex test.sh` (command in CI) ?

I believe -e specifies fail fast/early on error.

nv-kmcgill53:
I was running the test as `bash test.sh` rather than `bash -ex test.sh`. I think I don't want to change the docker run command for just this test though.

rmccorm4:
Spoke offline - I don't think this PR is needed and can likely be closed after verifying L0_mlflow 23.05 looks good after https://github.com/triton-inference-server/server/pull/5805 was merged.

PR #5808, opened 2023-05-17
Add MLFlow test for already loaded models (#5658). Update copyright year

GuanLuo:
CC @yeahdongcn

### DIFF
--- a/qa/L0_mlflow/test.sh
+++ b/qa/L0_mlflow/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


qa/L0_mlflow/test.sh
```diff
@@ -63,8 +63,12 @@ EOF
 
 rm -rf ./models
 mkdir -p ./models
+# Put some models in model repository to make sure MLFlow plugin would ignore
+# model that is not registered via MLFlow
+cp -r ./mlflow-triton-plugin/examples/onnx_float32_int32_int32 ./models/existing_model
+
 SERVER=/opt/tritonserver/bin/tritonserver
-SERVER_ARGS="--model-repository=./models --strict-model-config=false --model-control-mode=explicit"
+SERVER_ARGS="--model-repository=./models --strict-model-config=false --model-control-mode=explicit --load-model=*"
```
yeahdongcn, line 71: Just curious why `--load-model=*` is needed here.

```python
resp = self.triton_client.get_model_repository_index()
```

The above client call with default parameters could list all models.

Tabrizian (approved):

PR #5831, opened 2023-05-22
Add test for sequence model instance update

kthui:
Core PR: https://github.com/triton-inference-server/core/pull/207


docs/user_guide/model_management.md
```diff
@@ -225,11 +224,14 @@ request is received under
...
-guarentee any remaining request(s) from the in-flight sequence(s) will be routed
-to the same model instance for processing. It is currently the responsibility of
-the user to ensure any in-flight sequence(s) is complete before updating a
-sequence model.
+* If a sequence model is *reloaded* with in-flight sequence(s) (i.e. changes to
+the model file), Triton does not guarentee any remaining request(s) from the
+in-flight sequence(s) will be routed to the same model instance for processing.
+It is currently the responsibility of the user to ensure any in-flight
+sequence(s) are completed before reloading a sequence model.
+  * If a sequence model is *updated* (i.e. increasing/decreasing the instance
```
tanmayv25, line 232: nit: I think we don't need to mention the case of increasing the instance count.

qa/python_models/model_init_del/util.py
```diff
@@ -127,10 +127,31 @@ def update_instance_group(instance_group_str):
...
+            txt, post_match = txt.split("sequence_batching {")
+            if sequence_batching_str != "":
+                txt += "sequence_batching {\n"
+                txt += sequence_batching_str
+                txt += "\n}  # end sequence_batching\n"
+            txt += post_match.split("\n}  # end sequence_batching\n")[1]
+        elif sequence_batching_str != "":
+            txt += "\nsequence_batching {\n"
+            txt += sequence_batching_str
+            txt += "\n}  # end sequence_batching\n"
```
GuanLuo, line 154: Why you put the utils that is test specific to a model directory rather than just within the test directory?

qa/L0_model_update/instance_update_test.py
```diff
@@ -44,9 +45,21 @@ class TestInstanceUpdate(unittest.TestCase):
     __model_name = "model_init_del"
 
     def setUp(self):
-        # Initialize client
+        self.__reset_model()
         self.__triton = grpcclient.InferenceServerClient("localhost:8001")
 
+    def __reset_model(self):
```
GuanLuo, line 51: @Tabrizian is double underscore part of our code style?

qa/L0_model_update/instance_update_test.py
```diff
@@ -457,15 +467,116 @@ def test_instance_resource_decrease(self):
                 # explicit limit of 10 is set.
                 self.assertNotIn("Resource: R1\t Count: 3", f.read())
 
-    # Test for instance update on direct sequence scheduling
-    @unittest.skip("Sequence will not continue after update [FIXME: DLIS-4820]")
-    def test_instance_update_on_direct_sequence_scheduling(self):
-        pass
+    # Test sequence scheduler update
+    def test_sequence_instance_update(self):
```
GuanLuo, line 471: Add high level description on how the test operates

qa/python_models/model_init_del/util.py
kthui: This utils is model specific, so it is placed in the model directory.

qa/L0_model_update/instance_update_test.py
kthui: Yes. Added more descriptions on the scheduler update test, on [Add more documentaion for scheduler update test](https://github.com/triton-inference-server/server/pull/5831/commits/c070b68e8c4a4884f0cbc9c556e0bd66684c1885) commit.

qa/L0_model_update/instance_update_test.py
Tabrizian: It is not. If you want to declare a private method, you can just use a single underscore. Double underscores are usually reserved for internal Python methods (e.g., `__init__`)

kthui: I see, so we do not use https://docs.python.org/3.10/tutorial/classes.html#tut-private

kthui: Updated: [Use _ as private](https://github.com/triton-inference-server/server/pull/5831/commits/10277054f99663f9f16293872953678d27d8fffb)

qa/L0_model_update/instance_update_test.py
```diff
@@ -457,15 +467,115 @@ def test_instance_resource_decrease(self):
...
-    # Test for instance update on direct sequence scheduling
-    @unittest.skip("Sequence will not continue after update [FIXME: DLIS-4820]")
-    def test_instance_update_on_direct_sequence_scheduling(self):
-        pass
-
-    # Test for instance update on oldest sequence scheduling
-    @unittest.skip("Sequence will not continue after update [FIXME: DLIS-4820]")
-    def test_instance_update_on_oldest_sequence_scheduling(self):
-        pass
+    # Test instance update for direct and oldest scheduler
```
GuanLuo, line 470: This comment doesn't help me understand what the test setup is and what is the expected result from performing the test. To understand the test, I still need to look at the detail of the test.

And you are testing 4 test case ([direct, oldest] X [basic, ongoing]) in this test, please separate them into individual tests unless the current form is the minimal unit of the testing. Otherwise, it requires additional effort to distinguish which case is failing when this "collected" test fails.

GuanLuo (commented):
I think there are too many redirections in your test, for example, to understand one of your test cases, I will look at `_check_count` as the name indicates it is where the validation happens. It doesn't really tell what "kind" ("initialize" / "finalize") means, so I need to check `get_count` (in a different file in a different directory). `get_count` reads some value from a file, and that is another redirection that I need to look at the `model.py` for what is being written to the file and when will a value be written to a file.

qa/python_models/model_init_del/config.pbtxt
```diff
@@ -49,4 +49,4 @@ instance_group [
     count: 1
     kind: KIND_CPU
   }
-]
+]  # end instance_group
```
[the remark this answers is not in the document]
kthui: A marker for the utility function to easily find the end of the instance group config.

qa/L0_model_update/instance_update_test.py
kthui: I think we do not need to break it into 4 cases, because the test will show exactly which line that failed, which will give us a very good hint onto what went wrong. The only issue I had in the past is not knowing the line that failed represent whether the direct batcher or oldest batcher, so I separated it into 2 cases - direct and oldest.

Update commits: [Separate direct and oldest to different test cases](https://github.com/triton-inference-server/server/pull/5831/commits/c9d7b5f3f3059add7cea34a997a45bac27c84525) and [Separate nested tests in a loop into multiple test cases](https://github.com/triton-inference-server/server/pull/5831/commits/dcb55f0c047a0bc7c88f1fde7693bcfc3a01b89b)

kthui: Further split the cases for cleanness: [Refactor scheduler update test](https://github.com/triton-inference-server/server/pull/5831/commits/0268918bfbf613fdba943ea5dde34d3bd62a5534)

qa/L0_model_update/instance_update_test.py
```diff
@@ -43,18 +43,35 @@
     set_delay,
     update_instance_group,
     update_model_file,
+    update_sequence_batching,
 )
 from tritonclient.utils import InferenceServerException
 
+# The server is not restarted across different unit test cases. If a case would
+# to fail, all subsequent cases could also fail, because the model state may not
```
GuanLuo, line 51: "case" do you mean the test case (`def test_xxx`)? If so, shouldn't you have implement proper setup / teardown method for the test so each test is independent from the result of the other tests?

kthui: I added a `teardown` method to unload the model if a test case would to fail: [Add best effort to reset model state after a single test case failure](https://github.com/triton-inference-server/server/pull/5831/commits/99f59355474384719266c03686d62cb7816553be).

For simple failures, like an instance finalize count does not match, this will properly reset the model state to allow the next test case to function:
```
***
*** Failed model instance update test on rate limit mode off
***
+ cat instance_update_test.rate_limit_off.log
F...
======================================================================
FAIL: test_cpu_instance_update (__main__.TestInstanceUpdate)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/tritonserver/qa/L0_model_update/instance_update_test.py", line 186, in test_cpu_instance_update
    self.assertTrue(False, "Artificial failure")
AssertionError: False is not true : Artificial failure

----------------------------------------------------------------------
Ran 4 tests in 64.633s

FAILED (failures=1)
```
The first case failed, but the following cases passed.

If the server would to get stuck for some reasons, then this is insufficient and requires the server to be force stopped and restarted by the parent script. However, the goal of this test is to not restart the server for all the tests, so if restarting is the only fix, then it should be considered as an entire test failure.

PR #5878, opened 2023-05-31
Add response parameters for streaming GRPC inference to enhance decoupled support

rmccorm4:
Feature
- Add `response.parameters["triton_final_response"]` on streaming GRPC inference whenever a response has `TRITONSERVER_RESPONSE_COMPLETE_FINAL` flag set
- Add opt-in `request.parameters["triton_enable_empty_final_response"]` to tell server to generate an empty response (with model name/version, request ID, and parameters) and send it back to the client when a `nullptr` response is received with FINAL flag
    - a.k.a when backend calls `TRITONBACKEND_ResponseFactorySendFlags(nullptr, TRITONSERVER_RESPONSE_COMPLETE_FINAL)`
- Add `StateParameters` struct to set bool fields when doing first passthrough of request parameter parsing after state initialization
- Add common helpers for Join/Contains - this extends to http server and others easier if adding future "triton_" reserved params
- Move `ostream<<` `Step` operator to be usable by other helper methods after recent grpc folder restructure. The LOG_VERBOSE was outputting the enum number in files outside of `grpc_server.cc` after the restructure.

Tests: 
- Updated L0_decoupled test

Docs
- TBD - will probably update extension_parameters doc and decoupled doc

Client PR: https://github.com/triton-inference-server/client/pull/333


rmccorm4:
@nnshah1 General look and feel of server-side feature is all here. Few notes since design doc conversation:


1. For simplicity, I decided to keep scope to consistent behavior for streaming case, regardless of decoupled or not. i.e. the relevant response parameters get set no matter how you're using GRPC streaming. This means that grpc non-streaming, and http cases don't currently return any response parameters, but could be updated in future if needed. This is going off your offline message:
> BTW - if it becomes more confusing than worth it to describe the 'final response' parameter for http / non streaming - I'd advocate back to remove it. I see "streaming" itself as an extension now and we document the behavior just as we have a streaminferresponse. The only thing I didn't think made sense was for it to be model dependent (decoupled / not decoupled, etc.) - if its consistent within streaming - good for me.

2. I went with "flags_only" (`request.parameters["triton_enable_flags_only_response"]` and `response.parameters["triton_flags_only_response"]`) over "parameters_only". I think this is more intuitive to Triton users familiar with the intended use case of when a model/backend uses TRITONBACKEND_ResponseFactorySendFlags - i.e. the only reason this null response case comes up. Another plus is that "flags_only" is a bit shorter than "parameters_only" for code brevity. If you still want to change this naming convention, I'm happy to - just mentioning why I chose it.

Let me know if you have any thoughts or changes on these points.

src/grpc/stream_infer_handler.cc
```diff
@@ -540,57 +544,104 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
+      LOG_ERROR << "expected the response allocator to have added the response";
+    }
+  }
+
+  // Set response parameters to indicate certain behaviors to client
+  if (response) {
+    response->mutable_infer_response()->set_id(request_id);
+    auto& params = *(response->mutable_infer_response()->mutable_parameters());
+    params["triton_final_response"].set_bool_param(state->complete_);
+    params["triton_flags_only_response"].set_bool_param(flags_only);
```
Tabrizian, line 629: Do we need to explicitly communicate to the client that this is a flags only response? Is there a way for the client to implicitly realize that it is a flags only response? For example, by not having any tensors in the output.

Tabrizian:
> My two cents: unpack flags as parameters (we do this in other scenarios such as sequence) - use "params_only" as the request parameter. @Tabrizian , @tanmayv25 - thoughts here?

I think `params_only` is a bit confusing since parameters apply to both request and responses and are used outside of decoupled models. I think something along the lines of `flags_only` is more consistent with the server/backend API. I see your point that familiarity with backend/server APIs is not required when working with gRPC APIs but I think `params_only` is also too generic for decoupled models.

nnshah1:
> > My two cents: unpack flags as parameters (we do this in other scenarios such as sequence) - use "params_only" as the request parameter. @Tabrizian , @tanmayv25 - thoughts here?
> 
> I think `params_only` is a bit confusing since parameters apply to both request and responses and can be used outside of decoupled models. I think something along the lines of `flags_only` is more consistent with the server/backend API. I see your point that familiarity with backend/server APIs is not required when working with gRPC APIs but I think `params_only` is also too generic for decoupled models use-case here.

But when we describe it, "flags_only" == "response with only parameters and request id". I think its not specific to type of model - only the streaming inference protocol. I think it becomes important to decouple it from the implementation -

src/grpc/stream_infer_handler.cc
rmccorm4: This was brought up in the design doc, Guan had concerns with client differentiating between a response with no outputs/tensors being legitimate, or some hidden bug/error in server. Similarly, I don't think it's impossible for a model to have no outputs.

Do we have a contract/document somewhere saying that a model must have a non-zero number of outputs? I know it's certainly possible to return a response with no outputs because we don't verify it in core currently, but I don't know if it's "legal" in terms of what we expect.

src/grpc/stream_infer_handler.cc
```diff
@@ -540,57 +544,104 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
+                     << "Sending flags-only response";
+      response->mutable_infer_response()->Clear();
+    } else {
+      LOG_ERROR << "expected the response allocator to have added the response";
+    }
+  }
+
+  // Set response parameters to indicate certain behaviors to client
+  if (response) {
+    response->mutable_infer_response()->set_id(request_id);
```
nnshah1, line 626: If I follow correctly above in the error case we would pass along an empty request_id - but in this case we update it to "<<id unknown>>"

I think we would want to keep it unset. For logging I would potentially update to <<id not set>> or <<id empty>> instead of unknown - as unknown implies its set but not known ....

src/grpc/infer_handler.cc
tanmayv25, line 1: Didn't we decide to include the triton_final_response parameter for the ModelInfer as well? Similar case for the HTTP endpoint,

tanmayv25: Marking as resolved because of the explanation here: https://github.com/triton-inference-server/server/pull/5878#issuecomment-1569426115

src/grpc/stream_infer_handler.cc
tanmayv25: It is valid to have responses without any output tensor data. Some requests can just be intended to update the state of the model. We can have optional outputs in the responses.

tanmayv25:
May be we should steer clear of terms "params_only" and "flags_only" altogether to avoid confusion. How about the following?

```
triton_enable_flags_only_response -> triton_enable_final_response
triton_flags_only_response -> triton_is_final_response
```

src/grpc/stream_infer_handler.cc
```diff
@@ -540,57 +544,104 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
+  // Decoupled backends can return a null response via
+  // TRITONBACKEND_ResponseFactorySendFlags. By default, these null
+  // "empty" responses are not sent back to the client. Clients can
+  // opt-in to receiving these empty responses via request parameters.
+  // NOTE: The complete flag is the only flag used for this case at this time.
+  bool empty = (!iresponse && state->is_decoupled_ && state->complete_);
+  bool enable_empty = false;
+  if (empty) {
+    // Check request for opt-in parameter
+    const auto& request_params = state->request_.parameters();
```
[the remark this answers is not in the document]
rmccorm4: For efficiency, how about adding `state` as an arg to [SetInferenceRequestMetadata](https://github.com/triton-inference-server/server/blob/b9e8125eb6fa5a46c0dd3e7010b2a4294f0c04b9/src/grpc/infer_handler.cc#L217) and setting `state->enable_empty_response_` while request parameters are already being parsed there, rather than parsing twice?

Both `state` and `SetInferenceRequestMetadata` are used by both InferHandler and StreamInferHandler, so the `state->enable_empty_response_` wouldn't having meaning for the non-stream infer handler, but this is also true for `is_decoupled_` today.

Something like this:
```cpp
// Already parsing request parameters here when initializing state
TRITONSERVER_Error*
SetInferenceRequestMetadata(
    TRITONSERVER_InferenceRequest* inference_request,
    const inference::ModelInferRequest& request,
    State* state
)
{
...
  for (auto param : request.parameters()) {
    ...
    } else if (param.first.rfind("triton_", 0) == 0) {
      RETURN_IF_ERR(SetStateFromParameter(state, param));
    }
    ...
}
```
and add a new helper:
```cpp
TRITONSERVER_Error*
SetStateFromParameter(
  State* state,
  const inference::InferParameter& param)
{
  const auto& key = param.first;
  const auto& value = param.second;
  if (key == "triton_enable_empty_response") {
    if (value.parameter_choice_case() !=
          inference::InferParameter::ParameterChoiceCase::kBoolParam) {
      return TRITONSERVER_ErrorNew(
         "invalid value type for 'triton_enable_empty_response' parameter, expected bool_param.");
    }

    // Set state here for use in response complete callback
    state->enable_empty_response_ = value.bool_param();
  } else {
      return TRITONSERVER_ErrorNew(
          (std::string(
               "parameter keys starting with 'triton_' are reserved for Triton "
               "usage. Only the following keys starting with 'triton_' are "
               "allowed: ") +
           Join(TRITON_RESERVED_REQUEST_PARAMS, " "))
              .c_str());
  } 

  return nullptr;  // success
}
```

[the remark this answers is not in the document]
rmccorm4: It's a bit messy in likely having to move `SetInferenceRequestMetadata` into the InferHandler class, similar to [ForwardHeadersAsParameters](https://github.com/triton-inference-server/server/blob/ececc9fb45efd2eab2c9cc9418f80f55dacae9d7/src/grpc/infer_handler.h#L944-L945), and possibly the implementation into the header as well with the templated state types like [here](https://github.com/triton-inference-server/server/blob/ececc9fb45efd2eab2c9cc9418f80f55dacae9d7/src/grpc/infer_handler.h#L1060-L1087), which I don't love.

@nnshah1 Do you have any alternative ideas on a good spot to move this logic without too much mess?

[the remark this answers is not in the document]
nnshah1: unfortunately I think the best option is a new method of stream_infer_handler - we could put decoupled and ```enable_empty_response``` setting into a new method: SetStateOptions

Basically right at stream_infer_handler 200.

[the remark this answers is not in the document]
nnshah1: unfortunately I think the best option is a new method of stream_infer_handler - we could put decoupled and ```enable_empty_response``` setting into a new method: SetStateOptions

Basically right at stream_infer_handler 200. 

> May be we should steer clear of terms "params_only" and "flags_only" altogether to avoid confusion. How about the following?
> 
> ```
> triton_enable_flags_only_response -> triton_enable_final_response
> triton_flags_only_response -> triton_is_final_response
> ```

Summarizing both comments (allowing responses without tensors without specific indication, and providing specific naming to avoid confusion). We would have the following:

``` request.parameters["triton_enable_empty_final_response"]
    response.parameters["triton_final_response"]
```
Empty / non empty would be up to the client to handle appropriately. I think this is the clearest naming for the scenario at hand - probably if we need to be more general in the future we can plan for it then.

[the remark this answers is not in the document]
nnshah1: unfortunately I think the best option is a new method of stream_infer_handler - we could put decoupled and ```enable_empty_response``` setting into a new method: SetStateOptions

Basically right at stream_infer_handler 200. 

> May be we should steer clear of terms "params_only" and "flags_only" altogether to avoid confusion. How about the following?
> 
> ```
> triton_enable_flags_only_response -> triton_enable_final_response
> triton_flags_only_response -> triton_is_final_response
> ```

Summarizing both of @tanmayv25 comments (allowing responses without tensors without specific indication, and providing specific naming to avoid confusion). We would have the following:

``` request.parameters["triton_enable_empty_final_response"]
    response.parameters["triton_final_response"]
```
Empty / non empty would be up to the client to handle appropriately. I think this is the clearest naming for the scenario at hand - probably if we need to be more general in the future we can plan for it then.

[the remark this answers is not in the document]
nnshah1: > May be we should steer clear of terms "params_only" and "flags_only" altogether to avoid confusion. How about the following?
> 
> ```
> triton_enable_flags_only_response -> triton_enable_final_response
> triton_flags_only_response -> triton_is_final_response
> ```

Summarizing both of @Tcomments (allowing responses without tensors without specific indication, and providing specific naming to avoid confusion). We would have the following:

``` request.parameters["triton_enable_empty_final_response"]
    response.parameters["triton_final_response"]
```
Empty / non empty would be up to the client to handle appropriately. I think this is the clearest naming for the scenario at hand - probably if we need to be more general in the future we can plan for it then.

rmccorm4:
@nnshah1 @tanmayv25 @Tabrizian Feature should generally match what we all agreed upon offline. Requesting re-review to verify the code looks like it's doing what we agreed on.

I will follow-up with docs updates to describe the new feature/behavior. probably in https://github.com/triton-inference-server/server/blob/main/docs/user_guide/decoupled_models.md and https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_parameters.md.

If you'd like to see docs added to other places, let me know.

qa/L0_decoupled/decoupled_test.py
```diff
@@ -166,50 +230,51 @@ def _decoupled_infer(self,
...
-                                             order_sequence[i][j])
-                        this_data = result_list[j][1].as_numpy('OUT')
-                        self.assertEqual(len(this_data), 1)
-                        self.assertEqual(this_data[0], expected_data)
-                        this_idx = result_list[j][1].as_numpy('IDX')
-                        self.assertEqual(len(this_idx), 1)
-                        self.assertEqual(this_idx[0], j)
-                        expected_data += 1
+        for infer_helper in [self._stream_infer, self._stream_infer_with_params]:
+            try:
```
rmccorm4, line 236: this section is just indented under new for loop

Tabrizian (approved):

src/grpc/infer_handler.cc
```diff
@@ -37,6 +37,37 @@ NextUniqueId()
 
 namespace triton { namespace server { namespace grpc {
 
+// Define operator in triton::server::grpc namespace for use in all helpers
```
nnshah1, line 40: question: since we don't have this type of comment for others - would it be safe to remove?

src/grpc/infer_handler.cc
```diff
@@ -216,7 +247,7 @@ InferResponseStart(TRITONSERVER_ResponseAllocator* allocator, void* userp)
 TRITONSERVER_Error*
 SetInferenceRequestMetadata(
     TRITONSERVER_InferenceRequest* inference_request,
-    const inference::ModelInferRequest& request)
+    const inference::ModelInferRequest& request, StateParameters* state_params)
```
nnshah1, line 250: do we follow the pattern of using pointers for in/out parameters universally - or could we use a reference here (avoid the null check later)

src/grpc/infer_handler.cc
rmccorm4: I added it because it got lost in the last restructure of these files, but I don't mind removing it - I doubt it would stop a similar thing in the future anyways.

src/grpc/infer_handler.cc
rmccorm4: Probably majority percentage is pointers for in/out, but reference are sometimes used, I'm happy to use a reference in smaller cases like here. In larger chunks, I think it can get harder to read / spot smells, as the mutable references may look like local variables at a quick glance.

src/grpc/infer_handler.cc
```diff
@@ -340,6 +367,42 @@ SetInferenceRequestMetadata(
...
+    if (value.parameter_choice_case() !=
+        inference::InferParameter::ParameterChoiceCase::kBoolParam) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INVALID_ARG,
+          (std::string("invalid value type for '") + key +
+           std::string("' parameter, expected bool_param."))
+              .c_str());
+    }
+    state_params->enable_empty_final_response_ = value.bool_param();
+  } else if (!Contains(TRITON_RESERVED_REQUEST_PARAMS, key)) {
```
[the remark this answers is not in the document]
rmccorm4: Inverting check makes sense

src/grpc/infer_handler.cc
nnshah1: your choice - I like to avoid the null check if we can for internal apis - but more of a clarification on my part

nnshah1 (approved):

Tabrizian (dismissed):

qa/L0_decoupled/decoupled_test.py
```diff
@@ -137,8 +137,8 @@ def _stream_infer_with_params(self, request_count, request_delay, _,
                     if response.parameters.get("triton_final_response"):
                       completed_requests += 1
 
-                    # Detect flags-only response, no need to process it
-                    if not response.parameters.get("triton_empty_response"):
+                    # Detect empty response, only process response if non-empty
+                    if response.outputs:
```
GuanLuo, line 141: `is None`?

rmccorm4: It's correct as-is, I clarified the comment. It's equivalent to checking that `response.outputs is not None` - i.e. non-empty response.

GuanLuo: I meant `is not None`, it's more for style, similar to `if (ptr)` v.s. `if (ptr != nullptr)`.

rmccorm4: `if response.outputs is not None` will return True for an empty list `[]`. 

`if response.outputs` will return True for non-empty list, it will check for "trueness" across types, like a list having length > 0.

GuanLuo: When will each of the cases occurs? If there are two different representations, they should have different meaning (even only slightly) to the user then?

rmccorm4: Discussed offline. Pythonic way is to just check "truthyness"/"falsiness". If we will assign different meanings to `[ ]` vs. `None` in the future, we can change it. If we want to enforce more strict type checking, we can do so in the client library in the future (ex: `assert isinstance(response.outputs, list)` etc.) - but the pythonic UX is to just check `if response.outputs` in this context.

Tabrizian (approved):

nnshah1 (approved):

PR #5885, opened 2023-06-01
Add test on rate limiter max resource decrease update

kthui:
Add the test to decrease the rate limiter max resource limit, by updating instances to a lower resource count. If the rate limiter sees a instance is being removed that specifies a higher resource count than the max resource, then the "Should not print this" warning is printed, and the test script is also updated to capture the phrase "Should not print this" and report test failure if found.

Core PR: https://github.com/triton-inference-server/core/pull/208

### DIFF
--- a/qa/L0_model_update/instance_update_test.py
+++ b/qa/L0_model_update/instance_update_test.py
@@ -338,10 +338,10 @@ class TestInstanceUpdate(unittest.TestCase):
         # Unload model
         self.__unload_model()
 
-    # Test instance resource requirement increase
+    # Test instance resource requirement update
     @unittest.skipUnless(os.environ["RATE_LIMIT_MODE"] == "execution_count",
                          "Rate limiter precondition not met for this test")
-    def test_instance_resource_increase(self):
+    def test_instance_resource_update(self):
         # Load model
         self.__load_model(
             1,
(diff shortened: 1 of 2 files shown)


qa/L0_model_update/instance_update_test.py
```diff
@@ -311,10 +311,10 @@ def test_infer_while_updating(self):
         # Unload model
         self.__unload_model()
 
-    # Test instance resource requirement increase
+    # Test instance resource requirement update
     @unittest.skipUnless(os.environ["RATE_LIMIT_MODE"] == "execution_count",
                          "Rate limiter precondition not met for this test")
-    def test_instance_resource_increase(self):
+    def test_instance_resource_update(self):
         # Load model
```
tanmayv25, line 318: Can you describe the steps in the sub-test?

qa/L0_model_update/instance_update_test.py
tanmayv25, line 1: You should also add test cases where the resource limits are set from the server cli options. With these options and instance updates, check if the resource counts are appropriately updated - they should not be updated.

Triton server in verbose mode prints the computed resource max counts. You can probably add checks on these logs to see if the max limits are correctly updated.

qa/L0_model_update/instance_update_test.py
kthui: Yes. I will update as I update the tests on https://github.com/triton-inference-server/server/pull/5885#discussion_r1213864265

qa/L0_model_update/instance_update_test.py
kthui: Yes. Added an explicit resource round into the L0_model_update test, it makes sure all the applicable cases will pass with set explicit resource.

In addition, a new case exceeding the set explicit resource is added, which the model should not be unloaded or stuck on inference, when a request to update an instance above the explicit resource limit.

A new resource decrease case will ensure the resource is actually decreased, by checking the server output.

PR #5910, opened 2023-06-07
Allow changing ping behavior based on env variable in SageMaker and entrypoint updates

nskool:
* /ping relies on the ‘live’ ping check of the Triton model server i.e. return 200 if the model server is running irrespective of whether any model has been loaded, or whether any loaded model is ready to serve requests. 
* This change was made for the following reason:
    * In MME mode, if one of the models fails to load for any reason, the /ping health-check will not return 200, thereby preventing other future models to be loaded. This breaks the MME feature where customers expect models to be available independent of each other.

However, /ping need not rely on 'live' for SageMaker's default single-model endpoint mode. Hence, changing the default behavior in SME to use 'ready', and for MME to continue to use 'live'. When required, users can provide env variable `SAGEMAKER_TRITON_PING_MODE` to change this behavior.

Additionally, allow GRPC and metrics endpoint on SageMaker, for those customers who are willing to run a custom solution.

### DIFF
--- a/docker/sagemaker/serve
+++ b/docker/sagemaker/serve
@@ -73,6 +73,22 @@ fi
 if [ -n "$SAGEMAKER_SAFE_PORT_RANGE" ]; then
     SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --sagemaker-safe-port-range=${SAGEMAKER_SAFE_PORT_RANGE}"
 fi
+if [ -n $SAGEMAKER_TRITON_ALLOW_GRPC ]; then
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --allow-grpc=${SAGEMAKER_TRITON_ALLOW_GRPC}"
+else
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --allow-grpc=false"
+fi
+if [ -n $SAGEMAKER_TRITON_ALLOW_METRICS ]; then
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --allow-metrics=${SAGEMAKER_TRITON_ALLOW_METRICS}"
+else
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --allow-metrics=false"
+fi
+if [ -n "$SAGEMAKER_TRITON_METRICS_PORT" ]; then
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --metrics-port=${SAGEMAKER_TRITON_METRICS_PORT}"
+fi
+if [ -n "$SAGEMAKER_TRITON_GRPC_PORT" ]; then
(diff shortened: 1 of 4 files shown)


chen3933:
LGTM

GuanLuo (approved):
LGTM, will kick off CI

nskool:
> LGTM, will kick off CI

Hi @GuanLuo - I just rebased and updated the PR, could you please review again and trigger CI?

docker/sagemaker/serve
```diff
@@ -134,4 +150,4 @@ elif [ "${is_mme_mode}" = false ]; then
     SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --load-model=${SAGEMAKER_TRITON_DEFAULT_MODEL_NAME}"
 fi
 
-tritonserver --allow-sagemaker=true --allow-grpc=false --allow-http=false --allow-metrics=false --model-control-mode=explicit $SAGEMAKER_ARGS
+tritonserver --allow-sagemaker=true --allow-grpc=true --allow-http=false --allow-metrics=true --model-control-mode=explicit $SAGEMAKER_ARGS
```
GuanLuo, line 153: You need to expose way to configure the grpc and metrics port because the default value may not comply with SageMaker safe range if specified, this [test case](https://github.com/triton-inference-server/server/blob/main/qa/L0_sagemaker/test.sh#L229-L243) is an example that `serve` script will fail because default ports are not in safe range.

nskool: @GuanLuo can you review again, thank you

docker/sagemaker/serve
```diff
@@ -60,6 +73,22 @@ fi
 if [ -n "$SAGEMAKER_SAFE_PORT_RANGE" ]; then
     SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --sagemaker-safe-port-range=${SAGEMAKER_SAFE_PORT_RANGE}"
 fi
+if [ -n $SAGEMAKER_TRITON_ALLOW_GRPC ]; then
```
GuanLuo, line 76:
```suggestion
if [ -n "$SAGEMAKER_TRITON_ALLOW_GRPC" ]; then
```

docker/sagemaker/serve
```diff
@@ -60,6 +73,22 @@ fi
...
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --allow-grpc=${SAGEMAKER_TRITON_ALLOW_GRPC}"
+else
+    SAGEMAKER_ARGS="${SAGEMAKER_ARGS} --allow-grpc=false"
+fi
+if [ -n $SAGEMAKER_TRITON_ALLOW_METRICS ]; then
```
GuanLuo, line 81:
```suggestion
if [ -n "$SAGEMAKER_TRITON_ALLOW_METRICS" ]; then
```

GuanLuo (changes requested):
Test passed once the suggested change is addressed

GuanLuo (approved):

PR #5916, opened 2023-06-08
Add Redis cache build, tests, and docs

rmccorm4:
Add docs on Redis cache implementation, and update L0_response_cache tests to test Redis cache as well.

**NOTE:** The CI tests don't do anything comprehensive like starting or restarting multiple tritons or restarting Redis like I showed in my demo - they mostly test the same behaviors as existing `local` cache implementation. 

For these more comprehensive kinds of tests specific to Redis, I'd like to add some more end-to-end like testing with some Python clients/unittests in the style of our more recent CI tests in the future, **but not in time for this release**.

Core PR: https://github.com/triton-inference-server/core/pull/217

### DIFF
--- a/docs/user_guide/response_cache.md
+++ b/docs/user_guide/response_cache.md
@@ -153,7 +153,7 @@ For Triton-specific `redis` cache implementation details/configuration, see the
 
 #### Custom Cache
 
-With the new the TRITONCACHE API interface, it is now possible for
+With the TRITONCACHE API interface, it is now possible for
 users to implement their own cache to suit any use-case specific needs.
 To see the required interface that must be implemented by a cache
 developer, see the 
--- a/docs/user_guide/response_cache.md
+++ b/docs/user_guide/response_cache.md
@@ -101,7 +101,8 @@ that are used to communicate with a cache implementation of the user's choice.
 
 A cache implementation is a shared library that implements the required
 TRITONCACHE APIs and is dynamically loaded on server startup, if enabled. 
-For tags `>=23.03`, 
+
+Triton's most recent
 [tritonserver release containers](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/tritonserver)
 come with the following cache implementations out of the box:
 - [local](https://github.com/triton-inference-server/local_cache): `/opt/tritonserver/caches/local/libtritoncache_local.so`
--- a/docs/user_guide/response_cache.md
+++ b/docs/user_guide/response_cache.md
@@ -105,6 +105,7 @@ For tags `>=23.03`,
 [tritonserver release containers](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/tritonserver)
 come with the following cache implementations out of the box:
 - [local](https://github.com/triton-inference-server/local_cache): `/opt/tritonserver/caches/local/libtritoncache_local.so`
+- [redis](https://github.com/triton-inference-server/redis_cache): `/opt/tritonserver/caches/redis/libtritoncache_redis.so`
 
 With these TRITONCACHE APIs, `tritonserver` exposes a new `--cache-config` 
 CLI flag that gives the user flexible customization of which cache implementation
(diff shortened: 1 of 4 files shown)


qa/common/util.sh
```diff
@@ -66,7 +66,7 @@ function wait_for_server_ready() {
 
     local wait_secs=$wait_time_secs
     until test $wait_secs -eq 0 ; do
-        if ! kill -0 $spid; then
+        if ! kill -0 $spid > /dev/null 2>&1; then
```
rmccorm4, line 69: Note these changes in `qa/common/util.sh` are to mute log spam around calling `kill` on server pid when it doesn't start successfully.

docs/user_guide/response_cache.md
```diff
@@ -124,18 +125,44 @@ When `--cache-config local,size=SIZE` is specified with a non-zero `SIZE`,
...
+case, and Triton's `redis` cache will simply act as a client of your Redis 
+deployment. The [Redis docs](https://redis.io/docs/) should be consulted for 
+questions and details about configuring the Redis server.
+
+For Triton-specific `redis` cache implementation details/configuration, see the
+[redis cache implementation](https://github.com/triton-inference-server/redis_cache).
+
 #### Custom Cache
 
 With the new the TRITONCACHE API interface, it is now possible for
```
tanmayv25, line 156: Can we drop "the new " ?

qa/L0_response_cache/test.sh
```diff
@@ -29,12 +29,54 @@ RET=0
...
+# Only localhost supported in this test for now, but in future could make
+# use of a persistent remote redis server, or similarly use --replicaof arg.
+export TRITON_REDIS_HOST="localhost"
+export TRITON_REDIS_PORT="6379"
 
 rm -fr *.log
 
-# UNIT TEST
+function install_redis() {
+  ## Install redis if not already installed
```
tanmayv25, line 42: Why not install redis directly in our qa container? It would help keeping all the test dependencies in one place.

qa/L0_response_cache/test.sh
```diff
@@ -29,12 +29,54 @@ RET=0
...
+    echo "Redis successfully started in background"
+  else
+    echo -e "\n***\n*** Failed: Redis server did not start successfully\n***"
+    RET=1
+  fi
+}
+
+function stop_redis() {
+  echo "Stopping Redis server..."
+  redis-cli -h "${TRITON_REDIS_HOST}" -p "${TRITON_REDIS_PORT}" shutdown > /dev/null 2>&1 || true
```
tanmayv25, line 64: Is this call synchronous?

qa/L0_response_cache/test.sh
rmccorm4: I'd like to keep it local to the test for now to avoid bloating the container - I think it gets harder to track which packages belong to which tests, for cleaning up unused packages in the future. If redis expands to other tests in the future, I would probably move it to QA container install at that point.

qa/L0_response_cache/test.sh
rmccorm4: I believe it does from the docs: https://redis.io/commands/shutdown/

> Synchronously saves the database(s) to disk and shuts down the Redis server.

But reached out to Redis folks for confirmation. It shuts down very quickly, so probably hard to tell without filling it a lot of data

docs/user_guide/response_cache.md
rmccorm4: done

dyastremsky (dismissed):

Tabrizian (dismissed):

qa/L0_response_cache/test.sh
GuanLuo: > I'd like to keep it local to the test for now to avoid bloating the container ... If redis expands to other tests in the future

If the test should be self-contained and only testing redis specific functionality, should it be moved to redis cache repo? You probably want to group those tests like for the backends.

rmccorm4: That's a good idea, I'll look into moving the tests into cache implementation repos when I add add more e2e tests with python `unittest` and clients, but not going to restructure it for 23.06.

rmccorm4:
Added some simple tests to verify a password-protected redis server will work as well.

dyastremsky (approved):

PR #5922, opened 2023-06-08
Ensure HPCX dependencies found in container

dyastremsky:
Prepend HPCX dependencies to LD_LIBRARY_PATH search path so that the correct version is found by PyTorch.

These already exist in the install repo in CPU-only builds, so they are found there. For the GPU build, these need to be on the search paths so that libtorch.so finds them before the versions installed by apt for other packages.

PyTorch: https://github.com/triton-inference-server/pytorch_backend/pull/113

### DIFF
--- a/build.py
+++ b/build.py
@@ -1214,6 +1214,13 @@ COPY --from=min_container /usr/local/cuda-12.1/targets/{cuda_arch}-linux/lib/lib
 COPY --from=min_container /usr/local/cuda-12.1/targets/{cuda_arch}-linux/lib/libnvToolsExt.so.1 /usr/local/cuda/targets/{cuda_arch}-linux/lib/.
 COPY --from=min_container /usr/local/cuda-12.1/targets/{cuda_arch}-linux/lib/libnvJitLink.so.12 /usr/local/cuda/targets/{cuda_arch}-linux/lib/.
 
+RUN mkdir -p /opt/hpcx/ucc/lib/ /opt/hpcx/ucx/lib/
+COPY --from=min_container /opt/hpcx/ucc/lib/libucc.so.1 /opt/hpcx/ucc/lib/libucc.so.1
+COPY --from=min_container /opt/hpcx/ucx/lib/libucm.so.0 /opt/hpcx/ucx/lib/libucm.so.0
+COPY --from=min_container /opt/hpcx/ucx/lib/libucp.so.0 /opt/hpcx/ucx/lib/libucp.so.0
+COPY --from=min_container /opt/hpcx/ucx/lib/libucs.so.0 /opt/hpcx/ucx/lib/libucs.so.0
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -1102,6 +1102,12 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
     if 'onnxruntime' in backends:
         df += '''
 ENV LD_LIBRARY_PATH /opt/tritonserver/backends/onnxruntime:${LD_LIBRARY_PATH}
+'''
+
+    # Necessary for libtorch.so to find correct HPCX libraries
+    if enable_gpu and ('pytorch' in backends):
+        df += '''
+ENV LD_LIBRARY_PATH /opt/hpcx/ucx/lib/:${LD_LIBRARY_PATH}
```
Tabrizian, line 1110: I was wondering whether opencv is ok with this order. I think this means that opencv would pick up the newer hpcx version.

dyastremsky:
@Tabrizian I made the changes you requested in the related PR (https://github.com/triton-inference-server/pytorch_backend/pull/113). The changes seem to be passing so far. With these, we no longer need the PyTorch backend changes.

Would you be able to review this again?

build.py
```diff
@@ -1102,6 +1102,16 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
...
         df += '''
 ENV LD_LIBRARY_PATH /opt/tritonserver/backends/onnxruntime:${LD_LIBRARY_PATH}
+'''
+
+    # Necessary for libtorch.so to find correct HPCX libraries
+    if ('pytorch' in backends):
+        if not enable_gpu:
+            df += '''
+ENV LD_LIBRARY_PATH /opt/hpcx/ucc/lib/:${LD_LIBRARY_PATH}
+'''
```
Tabrizian, line 1112: Can we remove this part? Otherwise looks good.

dyastremsky: We cannot, or the CPU-only image will not be able to find libucc.so.1 (in the /opt/hpcx/ucc/lib folder). With the other version of these changes, it wasn't necessary since the library was added to the backend directory, which was on the RPATH. Since this is now copying over to the /opt/hpcx/ folder, the CPU-only build needs to know where to find library or it will have the previous failure to load.

If easier to read, I can combine this with the other LD_LIBRARY_PATH. I broke it up like this though, because I didn't want to modify the LD_LIBRARY_PATH in GPU images since we don't have to.

Tabrizian: Why do we only include ucc but not ucx. in cpu-only build. Is that a typo?

Tabrizian: That's why I'm a bit confused.

dyastremsky: We include both. There's no else statement. For the CPU-only build, we need to both the ucc and ucx on the path. For the GPU-only build, we only need the ucx path (since the ucc is found automatically).

Tabrizian (approved):

PR #5928, opened 2023-06-09
Enabling nested spans for trace mode OpenTelemetry

oandreeva-nv:
~Core side changes: https://github.com/triton-inference-server/core/pull/222~

~Mainly changes code to work with core-related changes and adds context.~

~Testing: currently check that there are 4 child spans (2 for http and 2 for grpc), I will add one more test, that makes sure that children's span ids correspond to parrent's span id. Open to more suggestions on testing~

Adding nested spans to OpenTelemetry tracing


oandreeva-nv:
Working on the design doc

src/tracer.cc
```diff
@@ -337,13 +342,14 @@ TraceManager::Trace::CaptureTimestamp(
           << "{\"name\":\"" << name << "\",\"ns\":" << timestamp_ns << "}]}";
     } else if (setting_->mode_ == TRACE_MODE_OPENTELEMETRY) {
 #ifndef _WIN32
-      otel_common::SystemTimestamp otel_timestamp{
-          (time_offset_ + std::chrono::nanoseconds{timestamp_ns})};
-      if (trace_span_ == nullptr) {
-        InitSpan(otel_timestamp);
+      if (otel_context_.HasKey(kRootSpan)) {
```
GuanLuo, line 345: constructed `Trace` object may not have root span?

src/tracer.cc
```diff
@@ -440,6 +641,17 @@ TraceManager::TraceActivity(
...
+
+  if (ts->setting_->mode_ == TRACE_MODE_OPENTELEMETRY) {
+#ifndef _WIN32
+    ts->ReportToOpenTelemetry(trace, activity, timestamp_ns);
+#else
+    LOG_ERROR << "Unsupported trace mode: "
+              << TraceManager::InferenceTraceModeString(mode_);
+#endif
+    return;
+  }
```
GuanLuo, line 653: For future improvement, we can derive subclass from `Trace` for different mode, i.e. `OpenTelemetryTrace`, `TritonTrace` and they implements a `Report()` function which is going to be `ReportToOpenTelemetry` or the `stringstream` appending below

src/tracer.cc
```diff
@@ -369,30 +375,225 @@ TraceManager::Trace::InitTracer(
...
+          std::chrono::steady_clock::now().time_since_epoch())
+          .count();
+  auto otel_start_timestamp = otel_common::SystemTimestamp{
+      time_offset_ + std::chrono::nanoseconds{steady_timestamp_ns}};
+  auto root_span = InitSpan(
+      "InferRequest", otel_start_timestamp, steady_timestamp_ns,
+      true /*is_root_span*/);
+  // Initializing OTel context and storring "InferRequest" span as a root span
+  // to keep it alive for the duration of the request.
+  otel_context_ = opentelemetry::context::Context({kRootSpan, root_span});
```
GuanLuo, line 393: Should root span key also be appended with the request id?

oandreeva-nv: For a standalone model traces and ensembles, there can be only one root span "InferRequest" with the key "root_span", which shows timestamps of when request was received/send and when inference was complete. For now,  simple key without request id is sufficient. In future, it may change

src/tracer.cc
oandreeva-nv: Thanks for this idea! I will add this to the refactoring ticket

src/tracer.cc
oandreeva-nv: No, it should have root span, since it was added during the initialization. I will remove these extra checks

src/tracer.h
```diff
@@ -174,22 +183,50 @@ class TraceManager {
...
+    // it starts a new request or compute span. For the request span it
+    // adds some triton related attributes, and adds this span to
+    // `otel_context_`. Alternativelly, if activity is
+    // `TRITONSERVER_TRACE_REQUEST_END` or `TRITONSERVER_TRACE_COMPUTE_END`,
+    // it ends the coresponding span.
+    void ReportToOpenTelemetry(
+        TRITONSERVER_InferenceTrace* trace,
+        TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns);
+
+    // Returns the span key, for which the activity belongs
```
tanmayv25, line 227: Using span key and span name interchangeably appears to be confusing. Let's finalize to use one of the options.

src/tracer.h
```diff
@@ -174,22 +183,50 @@ class TraceManager {
...
+
+    // Initializes Opentelemetry exporter, processor, provider and context
     void InitTracer(const TraceConfigMap& config_map);
 
     // Starts a span with the provided timestamp
-    void InitSpan(const opentelemetry::common::SystemTimestamp& timestamp_ns);
-
-    // Ends the started span and cleans up tracer
-    void EndSpan();
+    opentelemetry::nostd::shared_ptr<otel_trace_api::Span> InitSpan(
```
tanmayv25, line 199: `InitSpan -> StartSpan`

Why are we passing two timestamps to this function?

src/tracer.cc
```diff
@@ -337,13 +340,12 @@ TraceManager::Trace::CaptureTimestamp(
...
     } else if (setting_->mode_ == TRACE_MODE_OPENTELEMETRY) {
 #ifndef _WIN32
-      otel_common::SystemTimestamp otel_timestamp{
-          (time_offset_ + std::chrono::nanoseconds{timestamp_ns})};
-      if (trace_span_ == nullptr) {
-        InitSpan(otel_timestamp);
-      }
-      trace_span_->AddEvent(
-          name, otel_timestamp);
+      auto root_span = opentelemetry::nostd::get<
```
tanmayv25, line 343: As root_span is being used at multiple places, does it make sense to convert into class variable within Trace?

src/tracer.h
```diff
@@ -174,22 +183,50 @@ class TraceManager {
...
+        opentelemetry::nostd::shared_ptr<otel_trace_api::Span> span,
+        const uint64_t& raw_timestamp_ns);
+
+    // Reports TRITONSERVER_InferenceTraceActivity as event to
+    // the currently active span. If activity is an instance of
+    // `TRITONSERVER_TRACE_REQUEST_START` or
+    // `TRITONSERVER_TRACE_COMPUTE_START`,
+    // it starts a new request or compute span. For the request span it
+    // adds some triton related attributes, and adds this span to
+    // `otel_context_`. Alternativelly, if activity is
```
tanmayv25, line 220: *Alternatively

src/tracer.h
```diff
@@ -174,22 +183,50 @@ class TraceManager {
...
+    // `TRITONSERVER_TRACE_REQUEST_END` or `TRITONSERVER_TRACE_COMPUTE_END`,
+    // it ends the coresponding span.
```
tanmayv25, line 222: *corresponding

src/tracer.cc
```diff
@@ -369,30 +371,225 @@ TraceManager::Trace::InitTracer(
...
+        span->AddEvent(
+            TRITONSERVER_InferenceTraceActivityString(activity),
+            this->time_offset_ + std::chrono::nanoseconds{timestamp_ns});
+      }
+      break;
+    }
+    case TRITONSERVER_TRACE_COMPUTE_END:
+    case TRITONSERVER_TRACE_REQUEST_END: {
+      if (this->otel_context_.HasKey(current_span_key)) {
+        auto span = opentelemetry::nostd::get<
```
tanmayv25, line 538: L537-L543 are identical to L525-L531. 
Can you create a new function for AddEvent? 
You can include an option in this function to whether or not to call EndSpan.

src/tracer.cc
```diff
@@ -369,30 +371,225 @@ TraceManager::Trace::InitTracer(
...
+  if (current_span_key.empty()) {
+    return;
+  }
+  uint64_t id;
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceTraceId(trace, &id), "getting trace id");
+  current_span_key = current_span_key + std::to_string(id);
+  switch (activity) {
+    case TRITONSERVER_TRACE_REQUEST_START: {
+      const char* model_name;
```
tanmayv25, line 458: Create new function for the case blocks. 
It is a good practice to have shorter case blocks. Makes the code more readable.

src/tracer.h
```diff
@@ -174,22 +183,50 @@ class TraceManager {
...
+    void ReportToOpenTelemetry(
+        TRITONSERVER_InferenceTrace* trace,
+        TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns);
+
+    // Returns the span key, for which the activity belongs
+    std::string GetSpanNameForActivity(
```
tanmayv25, line 228: Shouldn't this be private? Can you run a brisk check through and make public function private that are not used outside this class.

src/tracer.h
oandreeva-nv: One represents the system time to place the span on the timeline, the second is the steady timestamp, for a more precise calculation for span's duration

src/tracer.h
oandreeva-nv: Span key and span name are different. Span name is what you will see in for example jaeger gui, when analyzing traces. Span key is what is used to store span entities in the context. Span's name is most likely will be different from span's key. I'll scan through to make sure everything is clear.

tanmayv25: Thanks for the clarification.

src/tracer.h
tanmayv25: We should add more documentation about these function parameters...

src/tracer.cc
oandreeva-nv: This way we will still need to store it in the OpenTelemetry context. We've discussed the task of exposing OpenTelemetry tracer and Context to user through python API, so that the user could create custom spans. Root span can be stored in / accessed from the OTel context.

I don't think that we should store the root span in multiple places, but I am open to discussion.

src/tracer.cc
oandreeva-nv: I refactored the code.

src/tracer.cc
oandreeva-nv: I've decided to get rid of switch-case altogether

src/tracer.cc
```diff
@@ -374,25 +369,192 @@ TraceManager::Trace::InitTracer(
...
+  uint64_t id;
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceTraceId(trace, &id), "getting trace id");
+
+  auto current_span_key = GetSpanKeyForActivity(activity, id);
+  if (current_span_key.empty()) {
+    return;
+  }
+
+  MaybeStartSpan(current_span_key, trace, activity, timestamp_ns, id);
```
tanmayv25, line 450: The "Maybe" logic is not completely captured in this function. 

I would propose creating a new wrapper function for calling an AddEvent including L450-L459. Based on the activity field, the function can create a span or not.

src/tracer.cc
```diff
@@ -374,25 +369,192 @@ TraceManager::Trace::InitTracer(
...
+  }
+}
+
+void
+TraceManager::Trace::MaybeStartSpan(
+    std::string span_key, TRITONSERVER_InferenceTrace* trace,
+    TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns,
+    uint64_t trace_id)
+{
+  if (activity != TRITONSERVER_TRACE_REQUEST_START &&
```
tanmayv25, line 514: Idealy, we should try to keep the related gating logic together. Currently, it is being exposed to the caller function when ending spans. This should get resolved by the wrapper function above.

src/tracer.h
```diff
@@ -174,26 +215,84 @@ class TraceManager {
...
+    /// OpenTelemetry context.
+    void EndSpan(std::string span_key);
+
+    /// Ends the provided span at specified steady timestamp.
+    ///
+    /// \param span_key Span's key to retrieve the corresponding span from the
+    /// OpenTelemetry context.
+    /// \param raw_timestamp_ns Steady timestamp to use as
+    /// `EndSpanOptions::end_steady_time`.
+    void EndSpan(std::string span_key, const uint64_t& raw_timestamp_ns);
```
tanmayv25, line 256: I am assuming no one would want to EndSpan with `raw_timestamp_ns=0`? 
We might be able to use same function signature -  if raw_timestamp_ns is 0, then we can record the current time.

src/tracer.h
```diff
@@ -157,7 +167,38 @@ class TraceManager {
...
+    /// \param activity  Trace activity.
+    /// \param timestamp_ns Steady timestamp, which is used to calculate
+    /// OpenTelemetry SystemTimestamp to display span on a timeline, and
+    /// OpenTelemetry SteadyTimestamp to calculate the duration on the span
+    /// with better precision.
+    void ReportToOpenTelemetry(
+        TRITONSERVER_InferenceTrace* trace,
+        TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns);
+
+   private:
```
GuanLuo, line 201: I think that `Trace` is complicated enough to be declare outside `TraceManager` and promote to proper class

src/tracer.cc
```diff
@@ -374,25 +369,192 @@ TraceManager::Trace::InitTracer(
       otel_trace_sdk::SimpleSpanProcessorFactory::Create(std::move(exporter_));
   provider_ =
       otel_trace_sdk::TracerProviderFactory::Create(std::move(processor_));
+  auto steady_timestamp_ns =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(
+          std::chrono::steady_clock::now().time_since_epoch())
+          .count();
+  auto root_span =
+      StartSpan("InferRequest", steady_timestamp_ns, true /*is_root_span*/);
+  // Initializing OTel context and storring "InferRequest" span as a root span
```
GuanLuo, line 378:
```suggestion
  // Initializing OTel context and storing "InferRequest" span as a root span
```

src/tracer.cc
```diff
@@ -374,25 +369,192 @@ TraceManager::Trace::InitTracer(
...
+  // to keep it alive for the duration of the request.
+  otel_context_ = opentelemetry::context::Context({kRootSpan, root_span});
 }
 
-void
-TraceManager::Trace::InitSpan(const otel_common::SystemTimestamp& timestamp_ns)
+opentelemetry::nostd::shared_ptr<otel_trace_api::Span>
+TraceManager::Trace::StartSpan(
+    std::string display_name, const uint64_t& raw_timestamp_ns,
+    bool is_root_span, std::string parent_span_key)
```
GuanLuo, line 386: Can `is_root_span` be deduced from `parent_span_key.empty()`?

GuanLuo (approved):

src/tracer.h
tanmayv25: Ignore this. It looks good as is.

src/tracer.h
oandreeva-nv: I'll move it during refactoring in a separate PR

src/tracer.cc
```diff
@@ -374,24 +369,203 @@ TraceManager::Trace::InitTracer(
...
+  uint64_t parent_id;
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceTraceParentId(trace, &parent_id),
+      "getting trace parent id");
+  std::string parent_span_key =
+      (parent_id == 0 && activity == TRITONSERVER_TRACE_REQUEST_START)
+          ? kRootSpan
+      : (activity == TRITONSERVER_TRACE_REQUEST_START)
+          ? kRequestSpan + std::to_string(parent_id)
+          : kRequestSpan + std::to_string(trace_id);
```
GuanLuo, line 408: using `if ... else ...` would be more clear than ternary op?

GuanLuo (dismissed):

tanmayv25 (dismissed):

tanmayv25 (approved):

GuanLuo (approved):

PR #5936, opened 2023-06-12
Add docs on decoupled final response feature

rmccorm4:


dyastremsky (dismissed):

docs/user_guide/decoupled_models.md
```diff
@@ -87,10 +87,100 @@ exactly one response per request. Even standard ModelInfer RPC in the GRPC endpo
...
 how the gRPC streaming can be used to infer decoupled models.
 
 If using [Triton's in-process C API](../customization_guide/inference_protocols.md#in-process-triton-server-api),
 your application should be cognizant that the callback function you registered with 
 `TRITONSERVER_InferenceRequestSetResponseCallback` can be invoked any number of times,
 each time with a new response. You can take a look at [grpc_server.cc](https://github.com/triton-inference-server/server/blob/main/src/grpc_server.cc)
+
+### Knowing When a Decoupled Inference Request is Complete
+
+A request is considered complete when a response containing the
```
nnshah1, line 100: At a high level - I don't think we need a full description of the different possibilities here. Could we leave it as "a request is considered complete when a response containing the ... is received from a model backend". 

And then split into two sections, "Client applications using the C API can retrieve this via the ...", "Client applications using the streaming GRPC protocol can retrieve this using response parameters. Decoupled models may not send a response for each request but may send a response after receiving multiple requests. In these cases GRPC client can opt-in to receive an empty final response for each request. By default, empty final responses are not sent to save network traffic."

I believe the implementation for a backend should already be covered as that isn't changing - so the existing documentation on sending flags / setting the response complete - should be sufficient - and we shouldn't need to describe how a backend indicates the end - only how a client can know.

docs/user_guide/decoupled_models.md
```diff
@@ -87,10 +87,100 @@ exactly one response per request. Even standard ModelInfer RPC in the GRPC endpo
...
+
+However, there are cases where a user may want to write a client that can
+generically handle any model, without knowing implementation details about it.
+Similarly, there are cases where the number of responses a model will send
+is unknown beforehand, so the client may need a programmatic way to know when
+the final response for a given request has been received. A common case for
+this may be where for a language model that has a `1` request -> `N` response
+structure. 
+
+To handle this case, Triton exposes a boolean `"triton_final_response"` response
```
nnshah1, line 142: need to add to the extension_parameters doc.

Also - maybe a line on the sequence extension doc (we reference grpc streaming mode there).

I think streaming mode should probably be its own extension doc - and ideally the new documentation would actually live there - as the changes are only for GRPC streaming mode - no changes for C API or for backend in actuality -

docs/user_guide/decoupled_models.md
```diff
@@ -87,10 +87,100 @@ exactly one response per request. Even standard ModelInfer RPC in the GRPC endpo
...
+#### TRITONBACKEND_ResponseFactorySendFlags 
+
+When a final response is sent via
+`TRITONBACKEND_ResponseFactorySendFlags(factory, TRITONSERVER_RESPONSE_COMPLETE_FINAL)`,
+no response is sent to the frontend by the backend/model, so nothing will be sent
+back to the client by default. For the client to receive a response containing
+this "final" signal via the `"triton_final_response"` parameter, the client
+will have to opt-in through the client library.
+
+To opt-in through the Python client library, the `enable_empty_final_response` arg
```
Tabrizian, line 168: Need to mention that this is only available for gRPC.

Tabrizian (commented):
@rmccorm4 would be great if we can finalize this PR by EOD so that I can include a link to this in the release notes.

docs/user_guide/decoupled_models.md
rmccorm4: 1. Added mention of the reserved params in parameters doc.
2. Took a look at sequence doc, but not sure where it makes sense to add anything at the moment.
3. Agreed streaming extension could be its own doc, but will defer to a future PR. Maybe along with KServe discussions.

Lmk what you think.

docs/user_guide/decoupled_models.md
rmccorm4: done

docs/user_guide/decoupled_models.md
rmccorm4: done

Tabrizian (dismissed):

docs/user_guide/decoupled_models.md
nnshah1: Sounds good

nnshah1 (approved):

rmccorm4:
Noticed small [typo](https://github.com/triton-inference-server/server/pull/5936/commits/cee930e30ef88930a599999c9cc2f9cacc328e9d) at the end, should be good now.

Tabrizian (approved):
We need to cherry-pick it to 23.06 too.

PR #5939, opened 2023-06-12
Fix count check in L0_batcher

dyastremsky:
Unlike the sequence batcher's check_status function, L0_batcher's check_status expects the execution count to be provided as a set. This allows one of the tests to provide two execution values. As a result, it gets checked via AssertIn. To be consistent in the stability check and allow it to pass when the batch is ready, L0_batcher needs to check that the actual execution count is in the expected execution count.

- Use "in" check rather than equality check for consistency
- Correct error message when execution count fails

### DIFF
--- a/qa/L0_batcher/batcher_test.py
+++ b/qa/L0_batcher/batcher_test.py
@@ -214,7 +214,7 @@ class BatcherTest(tu.TestResultCollector):
                 model_name, "1")
             self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
             actual_exec_cnt = stats.model_stats[0].execution_count
-            if actual_exec_cnt == exec_count:
+            if actual_exec_cnt in exec_count:
                 break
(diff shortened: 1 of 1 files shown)


tanmayv25:
So this change only allows the exec_count iterations break out early for this subtest: `test_max_queue_delay_only_non_default`

tanmayv25:
It doesn't seem to be fixing any bugs right?

dyastremsky:
> It doesn't seem to be fixing any CI failures right?

That's correct. Ryan noticed that for all of the tests using check_status in L0_batcher, it doesn't break out early because it's looking to see if they are equal and they never are. For example, even for the sub-tests providing one values, 2 != (2,). If we instead use in to be consistent with the assertIn later on, then 2 is in (2,), so it will break out. The result is that there's no change to CI failures, but we no longer wait the full 10 seconds each time when the batch is ready.

rmccorm4 (approved):

PR #5945, opened 2023-06-13
Add redis config and use local logfile for redis server

rmccorm4:
Add redis server logs to help debug future redis cache issues


qa/L0_response_cache/test.sh
```diff
@@ -47,7 +47,7 @@ function install_redis() {
 
 function start_redis() {
   # Run redis server in background
-  redis-server --daemonize yes --port "${TRITON_REDIS_PORT}"
+  redis-server redis.conf --daemonize yes --port "${TRITON_REDIS_PORT}"
```
slorello89, line 50: I think the .conf file and command line options are mutually exclusive - so you'll either need to echo the port to the end of the conf file or use the loglevel/logfile cli arguments as below
```suggestion
  redis-server --loglevel notice --logfile ./redis-server.log --daemonize yes --port "${TRITON_REDIS_PORT}"
```

rmccorm4: It appeared to work locally.

From these docs I see:
> Note that internally this generates an in-memory temporary config file (possibly concatenating the config file passed by the user, if any) where arguments are translated into the format of redis.conf.

Do you know why it says _possibly_ ?

dyastremsky (approved):

Tabrizian (approved):

PR #5946, opened 2023-06-13
Fix LD_LIBRARY_PATH for PyTorch backend

Tabrizian:


dyastremsky (approved):

dyastremsky:
From conversation with Iman, for posterity: the previous changes that prepended these directories to the path did so in build.py. That works when using build.py to build the container. However, the QA containers need these directories on the path as well, which is what Iman's PR here does.

Dockerfile.QA
```diff
@@ -371,3 +371,6 @@ ENV LD_LIBRARY_PATH /opt/tritonserver/qa/clients:${LD_LIBRARY_PATH}
 
 # DLIS-3631: Needed to run Perf Analyzer CI tests correctly
 ENV LD_LIBRARY_PATH /opt/hpcx/ompi/lib:${LD_LIBRARY_PATH}
+
+# Required for PyTorch to pickup the correct HPCX libraries
```
nnshah1, line 375: should we just combine this with the previous line and add a comment?

Tabrizian: I think it would be a bit more difficult to comment on which sections are added for PA/PyTorch.

PR #5962, opened 2023-06-20
adding tests for locale

nnshah1:


qa/python_models/python_version/model.py
```diff
@@ -54,6 +54,7 @@ def initialize(self, args):
         print(
             f'Python version is {sys.version_info.major}.{sys.version_info.minor}, NumPy version is {np.version.version}, and Tensorflow version is {tensorflow.__version__}',
             flush=True)
+        print(f'Locale is {locale.getlocale()}', flush=True)
```
rmccorm4, line 57: Do you need an `import locale` up top?

rmccorm4, line 57: Also need to update copyright 2023

qa/L0_backend_python/env/test.sh
```diff
@@ -123,6 +123,42 @@ for EXPECTED_VERSION_STRING in "$PY36_VERSION_STRING" "$PY37_VERSION_STRING" "$P
         RET=1
     fi
 done
+grep "Locale is \(None, None\)" $SERVER_LOG
```
rmccorm4, line 126: Instead of adding a comment in the CMake file per my original comment, can you instead add some brief description for why we're testing this here?

Something like this:
```
# Test locale in python stub processes
# NOTE: In certain pybind versions, the locale settings may not be propagated from parent to
#       stub processes correctly. See https://github.com/triton-inference-server/python_backend/pull/260.
```

Tabrizian (approved):

rmccorm4 (dismissed):

PR #5963, opened 2023-06-20
Ensure L0_batch_input requests received in order

dyastremsky:
L0_batch_input has been flaky in passing. It looks like the reason is that the responses rely on the requests being in order. Otherwise, for example, accumulation values will differ (since each request has a different volume and the tests assume that the request volumes will be received in the same order as requests sent).

This PR switches this test to use the GRPC client's streaming capability to ensure requests are received in order.

### DIFF
--- a/qa/L0_batch_input/batch_input_test.py
+++ b/qa/L0_batch_input/batch_input_test.py
@@ -62,10 +62,11 @@ class BatchInputTest(tu.TestResultCollector):
         model_name = "ragged_io"
 
         # The model is identity model
-        inputs = []
+        test_inputs = []
         for value in [2, 4, 1, 3]:
-            inputs.append([grpcclient.InferInput('INPUT0', [1, value], "FP32")])
-            inputs[-1][0].set_data_from_numpy(
+            test_inputs.append(
+                [grpcclient.InferInput('INPUT0', [1, value], "FP32")])
+            test_inputs[-1][0].set_data_from_numpy(
                 np.full([1, value], value, np.float32))
 
         user_data = queue.Queue()
@@ -77,11 +78,12 @@ class BatchInputTest(tu.TestResultCollector):
 
         async_requests = []
         try:
-            for inputs in inputs:
+            # Use
+            for input in test_inputs:
                 # Asynchronous inference call.
                 async_requests.append(
                     self.client.async_stream_infer(model_name=model_name,
-                                                   inputs=inputs,
+                                                   inputs=input,
                                                    outputs=outputs))
 
             expected_value_list = [[v] * v for v in [2, 4, 1, 3]]
@@ -115,11 +117,11 @@ class BatchInputTest(tu.TestResultCollector):
 
         async_requests = []
         try:
-            for inputs in self.inputs:
+            for input in self.inputs:
                 # Asynchronous inference call.
                 async_requests.append(
                     self.client.async_stream_infer(model_name=model_name,
-                                                   inputs=inputs,
+                                                   inputs=input,
                                                    outputs=outputs))
 
             value_lists = [[v] * v for v in [2, 4, 1, 3]]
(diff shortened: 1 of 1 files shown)


qa/L0_batch_input/batch_input_test.py
```diff
@@ -243,40 +274,42 @@ def test_max_element_count_as_shape(self):
...
+            callback=partial(self.client_callback, user_data))
 
         output_name = 'BATCH_OUTPUT'
-        outputs = [tritonhttpclient.InferRequestedOutput(output_name)]
+        outputs = [grpcclient.InferRequestedOutput(output_name)]
 
         async_requests = []
         try:
-            for inputs in self.inputs:
+            for inputs in inputs:
```
oandreeva-nv, line 301: for input(?) in inputs ?

qa/L0_batch_input/batch_input_test.py
```diff
@@ -307,23 +339,26 @@ def test_batch_item_shape(self):
...
+            callback=partial(self.client_callback, user_data))
 
         output_name = 'BATCH_OUTPUT'
-        outputs = [tritonhttpclient.InferRequestedOutput(output_name)]
+        outputs = [grpcclient.InferRequestedOutput(output_name)]
 
         async_requests = []
         try:
-            for request_inputs in inputs:
+            for inputs in inputs:
```
oandreeva-nv, line 351: ditto

qa/L0_batch_input/batch_input_test.py
```diff
@@ -154,27 +173,31 @@ def test_element_count(self):
...
+        self.client.start_stream(
+            callback=partial(self.client_callback, user_data))
 
         output_name = 'BATCH_AND_SIZE_OUTPUT'
-        outputs = [tritonhttpclient.InferRequestedOutput(output_name)]
+        outputs = [grpcclient.InferRequestedOutput(output_name)]
 
         async_requests = []
         try:
             for inputs in self.inputs:
```
oandreeva-nv, line 189: same here: could you please clarify why self.inputs here was not replaced as in other places?

oandreeva-nv:
Thanks for adding a fix for this! If you could add some clarification in comments in the code why at some places we are using `inputs` and in some `self.inputs`, I believe that would be of a great benefit to the clarity of this test.

qa/L0_batch_input/batch_input_test.py
dyastremsky: Done!

GuanLuo (approved):

dyastremsky:
@oandreeva-nv Thank you for your comments. In the original design of this test, 5 of the tests were using the same inputs, which were set before each test by the test setUp() function. 3 were using different inputs, so they either modified the global self.inputs variable or created a local one. To avoid using a global variable which seemed non-ideal, I modified them all to use the local variable if they were using different inputs.

Based on your feedback, I decided to refactor the test to have a set_inputs() function called at the start of each test. This avoids code duplication while making what's happening clearer to the reader, I think. Let me know if you're happy with those changes.

oandreeva-nv (dismissed):
LGTM! Thank you for a nice refactoring and for clarifications.

qa/L0_batch_input/batch_input_test.py
```diff
@@ -58,17 +50,26 @@ def callback(user_data, result, error):
...
+    def set_inputs(self, shapes, input_name):
+        self.dtype_ = np.float32
+        self.inputs = []
+        if len(shapes[0]) == 1:
+            for shape in shapes:
+                self.inputs.append(
+                    [grpcclient.InferInput(input_name, [1, shape[0]], "FP32")])
+                self.inputs[-1][0].set_data_from_numpy(
+                    np.full([1, shape[0]], shape[0], np.float32))
+        else:
```
GuanLuo, line 62: For future-proof, adding new test using this helper may result in unexpected input shape if not looking into the function definition.

Maybe better if breaking it into two set_input functions, i.e. `set_input` and `set_input_for_batch_item`

dyastremsky: Good idea. Done!

GuanLuo (approved):

PR #5976, opened 2023-06-22
Add test that >1000 model files can be loaded in S3

dyastremsky:
Now that >1000 model files can be loaded, add a test to verify this can be done.

Results of test before changes (model.py is not loaded due to not being one of first 1000 files in ascending order):
image
image

Results of test after changes:
image
image

Core: https://github.com/triton-inference-server/core/pull/225

### DIFF
--- a/qa/L0_storage_S3/test.sh
+++ b/qa/L0_storage_S3/test.sh
@@ -82,7 +82,7 @@ RET=0
 # Test 3 Scenarios:
 # 1. Only AWS ENV vars (Without aws configure)
 # 2. AWS ENV vars + dummy values in aws configure [ENV vars have higher priority]
-# 3. Only aws configured (Without AWS ENV vars)
+# 3. Only AWS configured (Without AWS ENV vars)
 for ENV_VAR in "env" "env_dummy" "config"; do
     SERVER_LOG=$SERVER_LOG_BASE.$ENV_VAR.log
     CLIENT_LOG=$CLIENT_LOG_BASE.$ENV_VAR.log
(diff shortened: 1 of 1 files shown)


qa/L0_storage_S3/test.sh
```diff
@@ -416,6 +417,41 @@ aws configure set default.region $AWS_DEFAULT_REGION && \
...
+# Clean up bucket contents
+aws s3 rm "${BUCKET_URL_SLASH}" --recursive --include "*"
+
+# Test case where S3 folder has >1000 files
+rm -rf models
+
+mkdir -p models/model/1
+echo $'class TritonPythonModel:\n\tdef execute(self):\n\t\tpass' > ./models/model/1/model.py
+
+for i in {1..1000}; do
```
GuanLuo, line 429: Can you set the end to be a bit larger so it is obvious that the model file will be in "2nd" page?

qa/L0_storage_S3/test.sh
```diff
@@ -416,6 +417,41 @@ aws configure set default.region $AWS_DEFAULT_REGION && \
...
+    touch models/model/0${i}.txt
```
kthui, line 430: I wonder would it be better if the `./models/model/1/model.py` script could check whether the 1000+ text files are present, maybe during the optional `def initialize(self, args)`?

qa/L0_storage_S3/test.sh
dyastremsky: Sure thing, done!

kthui (approved):

GuanLuo (approved):

PR #5979, opened 2023-06-22
Auto-formatting

dyastremsky:


Tabrizian (commented):
I think we need to update `CONTRIBUTING.md` with the updated version of clang as well.

Tabrizian (approved):

PR #5980, opened 2023-06-23
Add test for Python BLS model loading API

krishung5:
Python backend: https://github.com/triton-inference-server/python_backend/pull/262

### DIFF
--- a/qa/python_models/bls_model_loading/model.py
+++ b/qa/python_models/bls_model_loading/model.py
@@ -38,8 +38,8 @@ class PBBLSModelLoadingTest(unittest.TestCase):
         pb_utils.unload_model(self.model_name)
 
     def test_load_unload_model(self):
-        self.assertFalse(pb_utils.is_model_ready(self.model_name))
-        pb_utils.load_model(self.model_name)
+        self.assertFalse(pb_utils.is_model_ready(model_name=self.model_name))
+        pb_utils.load_model(model_name=self.model_name)
         self.assertTrue(pb_utils.is_model_ready(self.model_name))
         pb_utils.unload_model(self.model_name)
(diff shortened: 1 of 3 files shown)


qa/python_models/bls_model_loading/model.py
```diff
@@ -0,0 +1,119 @@
...
+        override_name = "override_model"
+        config = "{\"backend\":\"onnxruntime\"}"
+        with open('models/onnx_int32_int32_int32/3/model.onnx', 'rb') as file:
+            data = file.read()
+        files = {"file:1/model.onnx": data}
+
+        # Request to load the model with override file, should fail without
+        # providing override config.
+        with self.assertRaises(pb_utils.TritonModelException):
+            pb_utils.load_model(self.model_name, "", files)
```
Tabrizian, line 81: I was wondering why would it fail if config is not provided?

Tabrizian (commented):
Do we need to add some testing during finalize as well?

One use-case that I can think of is the user loads the dependent models in `initialize` and unloads them in finalize for BLS models.

krishung5:
> Do we need to add some testing during finalize as well?
> 
> One use-case that I can think of is the user loads the dependent models in `initialize` and unloads them in finalize for BLS models.

It seems like during `finalize` the server object of the model instance will not in the ready state, so the model loading API won't be working in `finalize`. I think we need to document this limitation as well.

qa/python_models/bls_model_loading/model.py
krishung5: This was from the core logic: https://github.com/triton-inference-server/core/blob/51674037a179fa661ae53b42c8d2bfcbf64a425a/src/model_repository_manager.cc#L145

Tabrizian (approved):

PR #5990, opened 2023-06-26
Refactor build.py CPU-only Linux libs for readability

dyastremsky:
- Split CPU-only Linux libs into own function to make if-else statement clearer

### DIFF
--- a/build.py
+++ b/build.py
@@ -1188,57 +1188,8 @@ RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
  && ldconfig \
  && rm -f ${_CUDA_COMPAT_PATH}/lib
 '''
-
     else:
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -1281,6 +1232,58 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
     return df
 
 
+def add_cpu_libs_to_linux_dockerfile(dockerfile, backends, target_machine):
```
Tabrizian, line 1235: can we make this part of the `BuildScript` class?

Tabrizian: You're right. I think what you previously had made more sense. I thought `dockerfile_prepare_container_linux` is part of the `BuildScript`.

I don't think the current PR would work since you can only call `add_cpu_libs_to_linux_dockerfile` using an instance of build script.

dyastremsky: Got it. I hadn't run CI with the updates yet, because I wanted to get your thoughts first. It had previously passed. Reverted the commits to get the PR back to the previous state.

PR #5995, opened 2023-06-27
Update `main` post-23.06 release

mc-nv:


build.py
```diff
@@ -1189,7 +1189,55 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
  && rm -f ${_CUDA_COMPAT_PATH}/lib
 '''
     else:
-        add_cpu_libs_to_linux_dockerfile(df, backends, target_machine)
+        libs_arch = 'aarch64' if target_machine == 'aarch64' else 'x86_64'
```
dyastremsky, line 1192: Thanks for doing the hard job of updating main based on the 23.06 commits. This code chunk isn't needed though. The line that these commit seem to be deleting right above (add_cpu_libs_to_linux_dockerfile) was refactored [here](https://github.com/triton-inference-server/server/pull/5990), which is why it's no longer in server. So the current code, without these additions (and the removal of add_cpu_libs_to_linux dockerfile above), is correct.

dyastremsky (changes requested):
PR looks good to me aside from needing to revert the changes to the else block in build.py's `dockerfile_prepare_container_linux` function.

build.py
mc-nv: I have dropped that commit for the PR.

Tabrizian (dismissed):

build.py
dyastremsky: Thanks, looks good!

dyastremsky (dismissed):

Tabrizian (approved):

PR #6000, opened 2023-06-28
Return updated dockerfile in add_cpu_libs

the-david-oy:
Python passes strings as immutable objects, so the new string needs to be returned in this function.

### DIFF
--- a/build.py
+++ b/build.py
@@ -1189,7 +1189,7 @@ RUN ln -sf ${_CUDA_COMPAT_PATH}/lib.real ${_CUDA_COMPAT_PATH}/lib \
  && rm -f ${_CUDA_COMPAT_PATH}/lib
 '''
     else:
-        df += add_cpu_libs_to_linux_dockerfile(df, backends, target_machine)
+        df += add_cpu_libs_to_linux_dockerfile(backends, target_machine)
 
     # Add dependencies needed for python backend
     if 'python' in backends:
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -1189,7 +1189,7 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
  && rm -f ${_CUDA_COMPAT_PATH}/lib
 '''
     else:
-        add_cpu_libs_to_linux_dockerfile(df, backends, target_machine)
+        df += add_cpu_libs_to_linux_dockerfile(df, backends, target_machine)
```
Tabrizian, line 1192: I think this would add the libraries twice.

dyastremsky: Thanks for raising this point It looks like strings don't get modified if passed as an argument: https://www.geeksforgeeks.org/is-python-call-by-reference-or-call-by-value/#

That's why the libraries weren't actually getting added in the CI. Removed that parameter to make it clearer and faster. Let me know if that works for you.

Tabrizian (approved):
Looks good. Not sure why Java changes are showing up in this branch.

dyastremsky:
> Looks good. Not sure why Java changes are showing up in this branch.

Thanks, Iman! It happened when the target branch changed to `mchornyi-main` instead of `main`. Hopefully, they should disappear when `mchornyi-main` is merged into `main`.

PR #6005, opened 2023-06-29
Allow uncompressed conda execution enviroments

jbkyang-nvi:
Add test for https://github.com/triton-inference-server/python_backend/pull/266

### DIFF
--- a/qa/L0_backend_python/env/test.sh
+++ b/qa/L0_backend_python/env/test.sh
@@ -30,7 +30,7 @@ source ../common.sh
 source ../../common/util.sh
 
 SERVER=/opt/tritonserver/bin/tritonserver
-BASE_SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=1 --strict-model-config=false"
+BASE_SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=1 --disable-auto-complete-config"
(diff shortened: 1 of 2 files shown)


qa/L0_backend_python/env/test.sh
```diff
@@ -63,6 +63,26 @@ cp ../../python_models/python_version/model.py ./models/python_3_7/1/
...
 conda deactivate
 
+# Use python-3-7 without conda pack
+path_to_conda_pack="$PWD/python-3-7"
+create_conda_env_with_specified_path "3.7" $path_to_conda_pack 
+conda install numpy=1.20.1 -y
+conda install tensorflow=2.1.0 -y
+conda install -c conda-forge libstdcxx-ng=12 -y
+
+PY37_VERSION_STRING="Python version is 3.7, NumPy version is 1.20.1, and Tensorflow version is 2.1.0"
```
dyastremsky, line 73: I'm not super familiar with this test, but it looks like this is redefining the previous PY37_VERSION_STRING with the same value before it's tested below. Is that intentional? Also, does this test fail before your changes in `python_backend`?

jbkyang-nvi: If the test failed, the server would not have been able to run at all. I'll change the version string

dyastremsky (approved):

dyastremsky (dismissed):

dyastremsky (approved):

PR #6016, opened 2023-07-03
Add testing for implicit state for PyTorch backend

Tabrizian:
Add testing for https://github.com/triton-inference-server/pytorch_backend/pull/103

- [x] Update the documentation
- [x] Disable string tests for max_batch_size > 0 tests
- [x] Add testing for string implicit state
- [x] Add testing for other data types

### DIFF
--- a/qa/common/gen_qa_implicit_models.py
+++ b/qa/common/gen_qa_implicit_models.py
@@ -28,7 +28,7 @@ import argparse
 import os
 import numpy as np
 import gen_ensemble_model_utils as emu
-from typing import List
+from typing import List, Tuple
 
 FLAGS = None
 np_dtype_string = np.dtype(object)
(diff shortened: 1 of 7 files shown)


GuanLuo (approved):

qa/common/gen_qa_implicit_models.py
```diff
@@ -381,6 +408,327 @@ def create_onnx_modelfile(models_dir, model_version, max_batch, dtype, shape,
         create_onnx_modelfile_with_initial_state(models_dir, model_version,
                                                  max_batch, dtype, shape)
 
+    if initial_state is None:
+        create_onnx_modelfile_wo_initial_state(models_dir, model_version,
+                                               max_batch, dtype, shape)
+    else:
+        # This model assumes that the initial state contains correct data
+        create_onnx_modelfile_with_initial_state(models_dir, model_version,
```
krishung5, line 416: I was wondering why we need to add this part. It looks like the same as the above.

krishung5 (commented):
nit: The copyright for some of the files need to updated.

qa/common/gen_qa_implicit_models.py
Tabrizian: Good catch, Fixed!

krishung5 (approved):

PR #6017, opened 2023-07-04
Allow non-decoupled model to send response and FINAL flag separately

GuanLuo:
Follow up on https://github.com/triton-inference-server/core/pull/229
For custom backend, one may call send response in the following style, even for "non-decoupled" model
```
TRITONBACKEND_ResponseSend(response, 0, nullptr /* success */);
TRITONBACKEND_ResponseFactorySendFlags(factory, TRITONSERVER_RESPONSE_COMPLETE_FINAL);
```
This yields single response of the request from the client's perspective and would expect a successful inference through HTTP / non-streaming GRPC as it is "non-decoupled". The previous implementation is restricted that it simply assume decoupled use case when response complete callback is invoked multiple time. The change is to relax the restriction and collapse the above into single response to the client.

### DIFF
--- a/src/grpc/infer_handler.cc
+++ b/src/grpc/infer_handler.cc
@@ -1,4 +1,4 @@
-// Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
--- a/src/grpc/infer_handler.cc
+++ b/src/grpc/infer_handler.cc
@@ -992,11 +992,6 @@ ModelInferHandler::InferResponseComplete(
     state->context_->EraseInflightState(state);
   }
 
-#ifdef TRITON_ENABLE_TRACING
-  state->trace_timestamps_.emplace_back(std::make_pair(
-      "INFER_RESPONSE_COMPLETE", TraceManager::CaptureTimestamp()));
-#endif  // TRITON_ENABLE_TRACING
-
   // If gRPC Stream is cancelled then no need of forming and returning
   // a response.
   if (state->IsGrpcContextCancelled()) {
@@ -1063,6 +1058,11 @@ ModelInferHandler::InferResponseComplete(
   if ((flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) == 0) {
     return;
   }
+
+#ifdef TRITON_ENABLE_TRACING
+  state->trace_timestamps_.emplace_back(std::make_pair(
+      "INFER_RESPONSE_COMPLETE", TraceManager::CaptureTimestamp()));
+#endif  // TRITON_ENABLE_TRACING
 
 #ifdef TRITON_ENABLE_TRACING
   state->trace_timestamps_.emplace_back(
--- a/src/grpc/infer_handler.cc
+++ b/src/grpc/infer_handler.cc
@@ -978,19 +978,19 @@ ModelInferHandler::InferResponseComplete(
   // notification.
   std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
 
-  // Increment the callback index
-  state->cb_count_++;
+  // Increment the callback index if received valid 'iresponse'
+  if (iresponse != nullptr) {
+    state->cb_count_++;
+  }
 
   LOG_VERBOSE(1) << "ModelInferHandler::InferResponseComplete, "
                  << state->unique_id_ << " step " << state->step_;
 
-  // Defer to the callback with the final response
-  if ((flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) == 0) {
-    LOG_ERROR << "[INTERNAL] ModelInfer received a response without FINAL flag";
-    return;
+  // Allow sending 1 response and final flag separately, only mark
+  // non-inflight when seeing final flag
+  if (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) {
+    state->context_->EraseInflightState(state);
   }
-
-  state->context_->EraseInflightState(state);
 
 #ifdef TRITON_ENABLE_TRACING
   state->trace_timestamps_.emplace_back(std::make_pair(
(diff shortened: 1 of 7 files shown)


rmccorm4:
@GuanLuo do we still plan to implement this to help unify decoupled/non-decoupled logic in backends like RIVA/TRTLLM etc?

src/http_server.cc
```diff
@@ -3126,17 +3120,23 @@ HTTPAPIServer::InferRequestClass::InferResponseComplete(
   }
 #endif  // TRITON_ENABLE_TRACING
 
+  LOG_TRITONSERVER_ERROR(
+      TRITONSERVER_InferenceResponseDelete(response),
+      "deleting inference response");
+
+  // Defer sending the response until FINAL flag is seen or
+  // there is error
+  if ((err == nullptr) && (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) == 0) {
```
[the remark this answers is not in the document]
GuanLuo: I made the change to not send error until FINAL flag is seen, otherwise there can be issue on the backend trying to call the response send function again after the `userp` has been released. Currently the only assumption is that the backend will not call the function again once final flag is sent.

src/http_server.cc
```diff
@@ -3729,25 +3729,19 @@ HTTPAPIServer::InferRequestClass::InferResponseComplete(
...
+  if (infer_request->response_count_ != 1) {
     err = TRITONSERVER_ErrorNew(
-        TRITONSERVER_ERROR_INTERNAL, "received an unexpected null response");
-  } else {
+        TRITONSERVER_ERROR_INTERNAL,
+        std::string(
+            "expected a single response, got " +
+            std::to_string(infer_request->response_count_))
+            .c_str());
+  } else if (response != nullptr) {
```
rmccorm4, line 3744: For the case:
```
TRITONBACKEND_ResponseSend(response, 0, nullptr /* success */);
TRITONBACKEND_ResponseFactorySendFlags(factory, TRITONSERVER_RESPONSE_COMPLETE_FINAL);
```

Am I correctly understanding that the "INFER_RESPONSE_COMPLETE" trace would get captured twice? I think for opentelemetry this is an "append" operation rather than an "overwrite". CC @oandreeva-nv

oandreeva-nv: it is an "append" operation. Need to verify this logic and will get back

oandreeva-nv: Seems like it, both of them call for `response_fn` callback, which captures "INFER_RESPONSE_COMPLETE" timestamp. I don't think this is that big of a deal for this PR. If we want to address it, file a jira ticket.

oandreeva-nv: Alternatively, this behavior can be described in docs

oandreeva-nv: Actually, IIUC 
`TRITONBACKEND_ResponseSend(response, 0, nullptr /* success */);` sets `flags` to 0, so
`flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL` will be 0, and per [line](https://github.com/triton-inference-server/server/pull/6017/files#diff-2d773636d85b98207338699c3cc2b9fc7d31f38e45131ca127dca73f45d4bbb2R3759), timestap capture should be skipped. Is this understanding correct?

GuanLuo: Yes, in this case we expect only one response so we could just record the response complete time when we see final signal

src/grpc/infer_handler.cc
```diff
@@ -978,20 +978,20 @@ ModelInferHandler::InferResponseComplete(
...
-    return;
+  // Allow sending 1 response and final flag separately, only mark
+  // non-inflight when seeing final flag
+  if (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) {
+    state->context_->EraseInflightState(state);
   }
 
-  state->context_->EraseInflightState(state);
-
 #ifdef TRITON_ENABLE_TRACING
```
oandreeva-nv, line 995: I have a feeling that `INFER_RESPONSE_COMPLETE` multiple capture issue will be happening here, since `deferring the response send` is happening later.

GuanLuo: Delayed

src/grpc/infer_handler.cc
```diff
@@ -1053,21 +1046,31 @@ ModelInferHandler::InferResponseComplete(
...
+
+#ifdef TRITON_ENABLE_TRACING
+  state->trace_timestamps_.emplace_back(std::make_pair(
+      "INFER_RESPONSE_COMPLETE", TraceManager::CaptureTimestamp()));
+#endif  // TRITON_ENABLE_TRACING
+
 #ifdef TRITON_ENABLE_TRACING
   state->trace_timestamps_.emplace_back(
       std::make_pair("GRPC_SEND_START", TraceManager::CaptureTimestamp()));
 #endif  // TRITON_ENABLE_TRACING
```
oandreeva-nv, line 1070:
```suggestion
#ifdef TRITON_ENABLE_TRACING
  state->trace_timestamps_.emplace_back(std::make_pair(
      "INFER_RESPONSE_COMPLETE", TraceManager::CaptureTimestamp()));
  state->trace_timestamps_.emplace_back(
      std::make_pair("GRPC_SEND_START", TraceManager::CaptureTimestamp()));
#endif  // TRITON_ENABLE_TRACING
```

oandreeva-nv: But we might want to capture `"INFER_RESPONSE_COMPLETE"`  where it was in the code before

GuanLuo: Perhaps, I am not sure as I just notice that the timing of the `RESPONSE_COMPLETE` capture was different in GRPC v.s. in HTTP. We can either keep the timing in previous implementation or unify it in both endpoints.

In HTTP, it is captured after the HTTP response has been fully prepared and right before Triton invoke the HTTP library to send the response. Whereas in GRPC, the timestamp is captured when the callback is invoked (before preparing the GRPC response).

oandreeva-nv (approved):
TMLG

src/grpc/infer_handler.cc
oandreeva-nv: resolved offline

rmccorm4 (approved):
I think it would be good to have some kind of test that asserts the traces come out correctly to avoid future breakage, but won't block the PR on it

PR #6018, opened 2023-07-04
Document instance naming on reuse

kthui:
core PR: https://github.com/triton-inference-server/core/pull/231

~~The `test_gpu_cpu_instance_update` test is removed, because adding and removing CPU and GPU instances impact the instance name, see the core PR for the naming rules, unless the instance names are explicitly controlled. The two primary goals of the removed `test_gpu_cpu_instance_update` test is to:~~
- ~~Ensure a model with more than one instance kind can be updated, which is covered by the new `test_instance_name_update`.~~
- ~~If the instances are simply shuffled on the config, no instance creation or destruction is performed, which is also covered by the new `test_instance_name_update`.~~

~~The `test_instance_signature` test is removed, because its primary goal is to ensure an instance is reused or not reused only based on the instance configuration, ignoring the name. Now, the name is taken into account, so a new test `test_instance_same_name` is added, which tests for the correct instance reused on instance configuration, with the instance name always explicitly fixed.~~

### DIFF
--- a/docs/user_guide/model_management.md
+++ b/docs/user_guide/model_management.md
@@ -224,6 +224,8 @@ request is received under
 'config.pbtxt' is modified in place. The swap file is not part of the model
 configuration, so its presence in the model directory may be detected as a new file
 and cause the model to fully reload when only an update is expected.
+  * The model instance reused might carry a different name than the one provided
+on the 'config.pbtxt', but the instances are equivalent beside their names.
 
 * If a sequence model is updated with in-flight sequence(s), Triton does not
 guarentee any remaining request(s) from the in-flight sequence(s) will be routed
--- a/docs/user_guide/model_management.md
+++ b/docs/user_guide/model_management.md
@@ -224,7 +224,7 @@ request is received under
 'config.pbtxt' is modified in place. The swap file is not part of the model
 configuration, so its presence in the model directory may be detected as a new file
 and cause the model to fully reload when only an update is expected.
-  * The model instance reused might carry a different name than the one provided
+  * When a model configuration is reloaded existing model instances will be checked against the new configuration and be reused if possible. If a model instance is reused it will retain its original name.
 on the 'config.pbtxt', but the instances are equivalent beside their names.
 
 * If a sequence model is updated with in-flight sequence(s), Triton does not
--- a/docs/user_guide/model_management.md
+++ b/docs/user_guide/model_management.md
@@ -224,8 +224,9 @@ request is received under
 'config.pbtxt' is modified in place. The swap file is not part of the model
 configuration, so its presence in the model directory may be detected as a new file
 and cause the model to fully reload when only an update is expected.
-  * When a model configuration is reloaded existing model instances will be checked against the new configuration and be reused if possible. If a model instance is reused it will retain its original name.
-on the 'config.pbtxt', but the instances are equivalent beside their names.
+  * When a model configuration is reloaded existing model instances will be
+checked against the new configuration and be reused if possible. If a model
+instance is reused it will retain its original name.
 
 * If a sequence model is updated with in-flight sequence(s), Triton does not
 guarentee any remaining request(s) from the in-flight sequence(s) will be routed
--- a/qa/L0_model_update/instance_update_test.py
+++ b/qa/L0_model_update/instance_update_test.py
@@ -186,6 +186,27 @@ class TestInstanceUpdate(unittest.TestCase):
         self.__update_instance_count(3, 0, "{\ncount: 7\nkind: KIND_GPU\n}")
         self.__unload_model()
 
+    # Test add/remove multiple CPU/GPU instances at a time
+    def test_gpu_cpu_instance_update(self):
+        # Load model with 1 GPU instance and 2 CPU instance
+        self.__load_model(
+            3,
+            "{\ncount: 2\nkind: KIND_CPU\n},\n{\ncount: 1\nkind: KIND_GPU\n}")
+        # Add 2 GPU instance and remove 1 CPU instance
+        self.__update_instance_count(
+            2, 1,
+            "{\ncount: 1\nkind: KIND_CPU\n},\n{\ncount: 3\nkind: KIND_GPU\n}")
+        # Shuffle the instances
+        self.__update_instance_count(
+            0, 0,
+            "{\ncount: 3\nkind: KIND_GPU\n},\n{\ncount: 1\nkind: KIND_CPU\n}")
+        # Remove 1 GPU instance and add 1 CPU instance
+        self.__update_instance_count(
+            1, 1,
+            "{\ncount: 2\nkind: KIND_GPU\n},\n{\ncount: 2\nkind: KIND_CPU\n}")
+        # Unload model
+        self.__unload_model()
+
     # Test model instance name update
     def test_instance_name_update(self):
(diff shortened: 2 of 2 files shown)


docs/user_guide/model_management.md
```diff
@@ -224,6 +224,8 @@ request is received under
 'config.pbtxt' is modified in place. The swap file is not part of the model
 configuration, so its presence in the model directory may be detected as a new file
 and cause the model to fully reload when only an update is expected.
+  * The model instance reused might carry a different name than the one provided
```
nnshah1, line 227:
```suggestion
  * When a model configuration is reloaded existing model instances will be checked against the new configuration and be reused if possible. If a model instance is reused it will retain its original name.
```

rmccorm4 (commented):
> If a model instance is reused it will retain its original name.

Is this true? IIRC the current behavior involved popping instances from the back in some reverse order and just renaming them. And current core PR just normalizes the names to always include the index.

kthui:
> > If a model instance is reused it will retain its original name.
> 
> Is this true? IIRC the current behavior involved popping instances from the back in some reverse order and just renaming them. And current core PR just normalizes the names to always include the index.

I should have described it more clearly. The instance cannot be renamed after it is created, but the selection algorithm for an instance to reuse does not honor the name of the instance that it is picking. For example:
```
Existing instances:   model_1_0   model_1_1   model_1_2
```
The user reduced the instance group count from 3 to 2, ideally, `model_1_0` and `model_1_1` should be instances remaining after the update. Since all instances are no different to each other beside their names, the algorithm may pick any two existing instance to be reused. The current behavior will pick `model_1_2` and `model_1_1` as the requested (or ideal) instance `model_1_0` and `model_1_1`, which:
```
New instances:   model_1_2   model_1_1
```
It looks like the instance `model_1_0` is renamed to `model_1_2`, but `model_1_2` is actually picked instead of ideally picking `model_1_0` at its place. (Note: nothing is renamed above, only which instance is picked and where it is placed on the list).

This is (ok?), if the instances are named by default. If the user would to flatten a instance group, for example:
```
{
  {kind: CPU, count: 1, name: "my_name_1"},
  {kind: CPU, count: 1, name: "my_name_2"},
  {kind: CPU, count: 1, name: "my_name_3"}
}
```
and then remove `my_name_3`:
```
{
  {kind: CPU, count: 1, name: "my_name_1"},
  {kind: CPU, count: 1, name: "my_name_2"}
}
```
Since the instance group name is not honored during reuse, the user will see `my_name_3   my_name_2` instead of the expected `my_name_1   my_name_2`. (Note: nothing is renamed, but `my_name_2` and `my_name_3` are picked for the reuse, instead of `my_name_1` and `my_name_2`)

Then, a solution is to add the name into the instance signature, so the group with the correct name will be picked. This works perfectly with user specified names, but not with default names. Maybe we should include names in instance matching if the name is specified by the user. If the name is not specified by the user, we should match with other equivalent instances that do not have a name specified by the user. In this case, maybe we can simply stop filling-in default names and leave those unspecified names empty. This naturally distinguished those that have names specified or not (`!name.empty()` vs `name.empty()`), so the instance matching algorithm can simply take the name into account without any worries. At the end, a name will be requested (i.e. `LOG_ERROR << instance.Name()`). If the name is not empty, the name is returned, If the name is empty, a default name can be assembled immediately.

Do we want to implement the solution on this ticket? Or we can add note on the limitation, and enhance it on a different ticket?

docs/user_guide/model_management.md
```diff
@@ -224,6 +224,9 @@ request is received under
 'config.pbtxt' is modified in place. The swap file is not part of the model
 configuration, so its presence in the model directory may be detected as a new file
 and cause the model to fully reload when only an update is expected.
+  * When a model configuration is reloaded existing model instances will be
+checked against the new configuration and be reused if possible. If a model
+instance is reused it will retain its original name.
```
rmccorm4, line 229: > If a model instance is reused it will retain its original name.
>
>> Is this true? IIRC the current behavior involved popping instances from the back in some reverse order and just renaming them. And current core PR just normalizes the names to always include the index.
>>
>>> I should have described it more clearly. The instance cannot be renamed after it is created, but the selection algorithm for an instance to reuse does not honor the name of the instance that it is picking. For example:
>>>
>>> Existing instances:   model_1_0   model_1_1   model_1_2
The user reduced the instance group count from 3 to 2, ideally, model_1_0 and model_1_1 should be instances remaining after the update. Since all instances are no different to each other beside their names, the algorithm may pick any two existing instance to be reused. The current behavior will pick model_1_2 and model_1_1 as the requested (or ideal) instance model_1_0 and model_1_1, which:
>>> 
>>> New instances:   model_1_2   model_1_1
It looks like the instance model_1_0 is renamed to model_1_2, but model_1_2 is actually picked instead of ideally picking model_1_0 at its place. (Note: nothing is renamed above, only which instance is picked and where it is placed on the list).

>>> This is (ok?), if the instances are named by default. If the user would to flatten a instance group, for example:
>>>{
>>>  {kind: CPU, count: 1, name: "my_name_1"},
>>>  {kind: CPU, count: 1, name: "my_name_2"},
>>>  {kind: CPU, count: 1, name: "my_name_3"}
>>>}
>>> and then remove my_name_3:
>>>
>>>{
>>>  {kind: CPU, count: 1, name: "my_name_1"},
>>>  {kind: CPU, count: 1, name: "my_name_2"}
>>>}
>>>
>>> Since the instance group name is not honored during reuse, the user will see my_name_3   my_name_2 instead of the expected my_name_1   my_name_2. (Note: nothing is renamed, but my_name_2 and my_name_3 are picked for the reuse, instead of my_name_1 and my_name_2)
>>>
>>> Then, a solution is to add the name into the instance signature, so the group with the correct name will be picked. This works perfectly with user specified names, but not with default names. Maybe we should include names in instance matching if the name is specified by the user. If the name is not specified by the user, we should match with other equivalent instances that do not have a name specified by the user. In this case, maybe we can simply stop filling-in default names and leave those unspecified names empty. This naturally distinguished those that have names specified or not (!name.empty() vs name.empty()), so the instance matching algorithm can simply take the name into account without any worries. At the end, a name will be requested (i.e. LOG_ERROR << instance.Name()). If the name is not empty, the name is returned, If the name is empty, a default name can be assembled immediately.
>>>
>>> Do we want to implement the solution on this ticket? Or we can add note on the limitation, and enhance it on a different ticket?

Moving comments so far to thread, my bad.

rmccorm4: Personally, I think this is not important to document right now. If we're planning some reworks or reconsiderations of naming behavior as discussed in [core PR](https://github.com/triton-inference-server/core/pull/231), I think we can just close this PR and leave out this comment for now and update the documentation when we make some behavioral changes. 

What do you think @GuanLuo @kthui ?

GuanLuo: Use hyperlink... I am leaning towards that this documentation is not necessary as well: what additional information that the user will gain from reading it? When the instance is "reused" (I am not sure about the use of this word as well), is the reader expecting the instance name will be different and thus we now need to state that the name is the same?

kthui: I agree the documentation does not help the user reading it. I will close this PR.

kthui:
Closing this PR as the documentation change is unnecessary at this point.

PR #6033, opened 2023-07-06
Add test to check the output memory type for onnx models

krishung5:
onnxruntime_backend: https://github.com/triton-inference-server/onnxruntime_backend/pull/202

### DIFF
--- a/qa/L0_backend_python/python_unittest.py
+++ b/qa/L0_backend_python/python_unittest.py
@@ -26,7 +26,7 @@
 
 import sys
 
-sys.path.append("../../common")
+sys.path.append("/opt/tritonserver/qa/common")
 
 import test_util as tu
 import shm_util
--- a/qa/L0_backend_python/python_unittest.py
+++ b/qa/L0_backend_python/python_unittest.py
@@ -26,7 +26,7 @@
 
 import sys
 
-sys.path.append("/opt/tritonserver/qa/common")
+sys.path.append("../../common")
 
 import test_util as tu
 import shm_util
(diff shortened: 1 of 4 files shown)


qa/L0_backend_python/python_unittest.py
```diff
@@ -26,7 +26,7 @@
 
 import sys
 
-sys.path.append("../../common")
+sys.path.append("/opt/tritonserver/qa/common")
```
Tabrizian, line 29: I think with this change we can remove this from `L0_dlpack_multi_gpu`. https://github.com/triton-inference-server/server/blob/fd96f23dd195e1eaa94284c1c572c5366b91d3aa/qa/L0_dlpack_multi_gpu/test.sh#L56

Tabrizian (approved):

qa/python_models/bls_onnx_warmup/model.py
```diff
@@ -0,0 +1,87 @@
...
+        expected_output_0 = input0.as_numpy() - input1.as_numpy()
+        expected_output_1 = input0.as_numpy() + input1.as_numpy()
+
+        output0 = from_dlpack(
+            output0.to_dlpack()).to('cpu').cpu().detach().numpy()
+        output1 = from_dlpack(
+            output1.to_dlpack()).to('cpu').cpu().detach().numpy()
+
+        self.assertTrue(np.all(output0 == expected_output_0))
+        self.assertTrue(np.all(output1 == expected_output_1))
```
GuanLuo, line 70: Why mixing the test case with the model file, how does it get invoked?

krishung5: I'm using Python model to run the unit test. It will be triggered by the `python_unittest.py` script, which is copied from L0_backend_python.

GuanLuo (approved):

PR #6043, opened 2023-07-11
Support for Big Endian Byte Order

nnshah1:
Updates to reflect:

1) All tensors transferred between core and backends are in host byte order
2) All binary tensors transferred over the network between client and server are in little endian byte order
3) gRPC and HTTP Server translate binary tensors into host byte order
4) Tensors transferred via shared memory are assumed in host byte order and no translation is performed


build.py
```diff
@@ -1179,6 +1181,15 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu,
         df += fastertransformer_buildscript.create_postbuild(
             is_multistage_build=False)
 
+    if target_machine == "s390x":
```
nnshah1, line 1184: @nv-kmcgill53 this is mainly for E2E testing as we do not publish such an image. Might be a better way to handle this however.

src/common.cc
```diff
@@ -111,4 +121,337 @@ Join(const std::vector<std::string>& vec, const std::string& delim)
...
+/// \param[in, out] base pointer to array of data type elements to convert
+/// \param[in] byte_size size of array in bytes
+///
+template <typename T>
+void
+SwapEndian(char* base, size_t byte_size)
+{
+  T* value = reinterpret_cast<T*>(base);
+  for (size_t offset = 0; offset < byte_size; offset += sizeof(T), ++value) {
+    *value = bswap(*value);
```
nv-kmcgill53, line 243: IMO we should look into using SIMD for these functions as they will be on the critical inference path. AVX2 shipped in 2013 which would be nice to target. PowerPC looks like it has [support for SIMD](https://stackoverflow.com/questions/66195879/simd-extensions-on-power-compiler-flags-and-processor-support) as well.

src/test/endian_conversion_test.cc
```diff
@@ -0,0 +1,179 @@
...
+{
+  const std::array<uint8_t, 10> original_values = {
+      0xde, 0xad, 0xde, 0xad, 0xde, 0xad, 0xde, 0xad, 0xde, 0xad};
+
+
+  ConversionTest(TRITONSERVER_TYPE_UINT8, original_values, original_values);
+}
+
+
+}}  // namespace triton::server
```
nv-kmcgill53, line 172: I would like to see tests for floating pint as well to make sure the mantissa bits are read correctly on a big-endian machine.

nv-kmcgill53 (changes requested):
First pass. I still need to review the partial array code.

build.py
nv-kmcgill53: I need to think more about this. Thanks for pointing this out.

src/common.cc
nnshah1: As this code will only be compiled on architectures where the native byte order is big endian - do you think that is something we could add as a follow on optimization? My thinking is to provide initial functionality / get feedback on the initial approach.

src/common.cc
```diff
@@ -111,4 +121,337 @@ Join(const std::vector<std::string>& vec, const std::string& delim)
...
+
+/// Returns uint16_t input with byte order swapped
+///
+/// \param[in] input 
+/// \return input with byte order swapped
+///
+uint16_t
+bswap(const uint16_t input)
+{
+  return __builtin_bswap16(input);
```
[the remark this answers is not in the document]
nnshah1: If I follow the thread correctly, on windows we will want to use:

```
_byteswap_ushort
```

instead of the GCC built ins - will update.

src/common.cc
rmccorm4: I like Kyle's perf recommendations, but I think they could be medium/high effort for low impact given the rarity at which we encounter big endian machines. I would advocate leaving it as a follow-up and waiting for reports of slowness.

src/common.h
```diff
@@ -169,4 +178,84 @@ bool Contains(const std::vector<std::string>& vec, const std::string& str);
...
+
+#define HostToLittleEndian(datatype, base, byte_size) \
+  do {                                                \
+  } while (false)
+#define LittleEndianToHost(datatype, base, byte_size, ...) \
+  do {                                                     \
+  } while (false)
+#define GetDataTypeForRawInput(server, model_name, model_version) \
+  do {                                                            \
+  } while (false)
```
Tabrizian, line 256: I was wondering why we need a macro for the little endian case but we use a function for big endian case? I probably need to dig more into the implementation.

src/grpc/infer_handler.cc
```diff
@@ -497,10 +498,7 @@ InferGRPCToInput(
             auto& serialized = serialized_data->back();
             serialized.reserve(
                 io.contents().int_contents_size() * elem_byte_size);
-            for (const auto& element : io.contents().int_contents()) {
-              // Assuming the system is little-endian, picking the
-              // least significant byte of 32-bit integer as a
-              // int8 element
+            for (const int8_t element : io.contents().int_contents()) {
```
Tabrizian, line 501: small nit: can we use "const auto"? Are we not using reference because the reference is larger than the actual data so making a copy of reference is slower than actually copying the data?

build.py
```diff
@@ -1337,6 +1348,9 @@ def create_build_dockerfiles(container_build_dir, images, backends, repoagents,
         base_image = images['base']
     elif target_platform() == 'windows':
         base_image = 'mcr.microsoft.com/dotnet/framework/sdk:4.8'
+    elif target_machine() == 's390x':
+        fail_if (FLAGS.enable_gpu, "s390x for testing big endian support only. GPU support not enabled.")
+        base_image = 's390x/ubuntu:22.04'
```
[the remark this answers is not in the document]
nnshah1: good idea! will check

src/common.h
nnshah1: Essentially wanted to completely eliminate these in the little endian case. For systems that are little endian already there is nothing to do. Using this to reduce the ifdefs at the place of usage. 

if the host type is already LittleEndian - then the macro HostToLittleEndian still does what is intended - but is a no-op.

Happy to discuss other ways.

src/grpc/infer_handler.cc
nnshah1: const auto  will use `int32_t` and then the append below only works correctly for little-endian (it copies the first byte of the 4 byte `int32_t`).

coercing to int8_t forces the compiler to cast to int8_t correctly for the native architecture and then we can copy it independent of big or little endian (i.e. it gets the correct byte either first or last). 

This seemed like the cleanest / most efficient way to handle this (essentially remove the assumption on the little endian byte order)

src/common.cc
```diff
@@ -111,4 +121,337 @@ Join(const std::vector<std::string>& vec, const std::string& delim)
...
+void
+LittleEndianToHost(
+    TRITONSERVER_DataType datatype, char* base, size_t byte_size,
+    std::vector<char*>& partial_result, size_t& next_offset)
+{
+  SwapEndian(datatype, base, byte_size, false, partial_result, next_offset);
+}
+
+TRITONSERVER_DataType
+GetDataTypeForRawInput(
```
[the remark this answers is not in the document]
nnshah1: Was on the fence here - I opted for an error at the call site for when it wasn't able to determine the data type. If there is an issue with the model in terms of being able to send a raw binary input - that is also caught by other checks in normalize, etc. In this case I think having the more general error "couldn't determine data type" was sufficient. The only real check here is that there is an input (otherwise model, model version, etc. should be caught already).

Let me know if you feel otherwise -

src/common.h
nnshah1: @Tabrizian Since we only use these in a few places - perfectly happy to guard them with `#ifdef TRITON_BIG_ENDIAN` instead of using empty macros if it makes things a bit clearer. I think initially I was hoping to avoid needing to define `TRITON_BIG_ENDIAN` - and kind of liked the way the functions or macros still behaved correctly without guards.

src/grpc/infer_handler.cc
nv-kmcgill53: I learned something new today 👍

build.py
nnshah1: I think with @Tabrizian suggestion of using the standard image - we can remove this. The remaining bit would be the miniconda version - but we could also provide the ability to simply exclude miniconda

src/common.cc
```diff
@@ -111,4 +121,337 @@ Join(const std::vector<std::string>& vec, const std::string& delim)
...
+/// Returns uint32_t length with byte order swapped and
+/// returns length in host byte order.
+///
+/// \param[in] length
+/// \param[in] host_byte_order whether original length is in host byte order
+/// \param[out] host_length length in host byte order
+/// \return length with byte order swapped
+///
+uint32_t
+bswap_length(
```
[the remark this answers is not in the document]
nnshah1: Suggestion change all functions to:

```
SwapBytes()
SwapLength()
```
Instead of 
```
SwapEndian()
```

Partial would be via parameter overloading - or alternatively:

```
SwapBytes()
SwapLength()
SwapBytesPartial()
SwapLengthPartial()
```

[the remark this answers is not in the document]
Tabrizian: Looks good!

src/common.cc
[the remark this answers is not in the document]
Tabrizian: I think it would be nicer if we can return the error. We can still keep the message as "couldn't determine data type". I agree that these calls would most likely not return any error but the logic on why the error checking is not required (i.e., normalize is first called) is not apparent in the first glance.

src/common.cc
```diff
@@ -111,4 +121,337 @@ Join(const std::vector<std::string>& vec, const std::string& delim)
...
+  triton::common::TritonJson::Value document;
+  triton::common::TritonJson::Value inputs;
+  triton::common::TritonJson::Value input;
+  const char* datatype = "";
+  size_t datatype_length = 0;
+  const char* buffer;
+  size_t byte_size;
+
+  RETURN_VAL_IF_ERR(
+      TRITONSERVER_ServerModelMetadata(
```
Tabrizian, line 432: I think this call is very costly. Is there a way to avoid making this call for every inference request in raw input case?

src/common.h
Tabrizian: Is it possible to move the ifdefs inside these functions? The body of the function would change based on whether `ifdef TRITON_BIG_ENDIAN` is enabled or not. If it is not enabled, it wouldn't do anything.

src/grpc/infer_handler.cc
Tabrizian: Thanks for explaining! Resolving.

src/common.cc
nnshah1: I agree - we could cache the data types? but if models change that becomes an issue as well. Unfortunately in this case the request itself doesn't have enough information to know the data type. maybe we start here and think of how to optimize?

src/common.h
nnshah1: @nv-kmcgill53 - What do you think - is the compiler smart enough to remove the calls completely? I want to avoid any calls in little endian case entirely - so if compiler would remove an "empty function" as @Tabrizian is suggesting - am good with that as well. I'll dig also - but you might know the answer already.

src/common.cc
Tabrizian: I agree we can optimize this later since it only affects the performance of raw data on big endian. Resolving.

src/common.h
nv-kmcgill53: These calls get optimized out for anything greater than `Debug`. This is also a standard pattern in industry so I have no problem with the way it's defined. Mostly this pattern is used for defining debug functions for printing/logging info which then get optimized out in the `Release` build. @nnshah1 you might be able to get away with just not defining anything at all after it so that the symbol gets inserted in the preprocessor and then gets translated to no-op. I need to play around with that though.

Tabrizian: The current way it is handled should be fine resolving.

qa/L0_big_endian/test.sh
```diff
@@ -0,0 +1,58 @@
...
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# TODO
+# Add end to end tests
+#
```
rmccorm4, line 30: Still needed?

src/common.cc
```diff
@@ -111,4 +121,343 @@ Join(const std::vector<std::string>& vec, const std::string& delim)
   return ss.str();
 }
 
+#ifdef TRITON_BIG_ENDIAN
+
+/// Returns uint16_t input with byte order swapped
+///
+/// \param[in] input
```
rmccorm4, line 128: I like this `\param[in]` syntax! I think it would be especially useful and clarifying for our functions that use `\param[in,out]` . Not in this PR, but I think we should update our API docs to use this syntax.

src/http_server.cc
```diff
@@ -2538,6 +2550,8 @@ HTTPAPIServer::EVBufferToInput(
           v_idx++;
         }
 
+        LittleEndianToHost(dtype, base, base_size, partial_result, next_offset);
```
rmccorm4, line 2553: Does this not complain that `partial_result` isn't defined when `TRITON_BIG_ENDIAN` is undefined?

src/http_server.cc
```diff
@@ -2759,6 +2786,8 @@ HTTPAPIServer::EVBufferToRawInput(
         v_idx++;
       }
 
+      LittleEndianToHost(dtype, base, base_size, partial_result, next_offset);
```
rmccorm4, line 2789: Same as https://github.com/triton-inference-server/server/pull/6043/files#r1264141301

src/http_server.h
```diff
@@ -310,8 +310,9 @@ class HTTPAPIServer : public HTTPServer {
       evbuffer* input_buffer, InferRequestClass* infer_req,
       size_t header_length);
   TRITONSERVER_Error* EVBufferToRawInput(
-      const std::string& model_name, TRITONSERVER_InferenceRequest* irequest,
-      evbuffer* input_buffer, InferRequestClass* infer_req);
+      const std::string& model_name, const int64_t model_version,
```
rmccorm4, line 313: Where did the addition of `model_version` to this function come from?

rmccorm4: NVM, I see it's for getting ServerModelMetadata.

src/common.cc
rmccorm4: Should we add a `NOTE/FIXME` type comment above? I don't think it would be obvious to future readers that this is costly without really staring at it.

src/test/endian_conversion_test.cc
rmccorm4: Looks like this is covered now.

rmccorm4 (commented):
Overall look and feel LGTM, left a few minor comments.

I think the only outstanding item is around the base_image used for testing and `build.py` changes?

src/http_server.cc
nnshah1: get's eaten up

src/http_server.cc
nnshah1: chomp

src/common.cc
nnshah1: as discussed offline - we'll revisit the performance optimizations in a later PR

src/test/endian_conversion_test.cc
nnshah1: added tests for fp32 / fp64 -

nnshah1:
> Overall look and feel LGTM, left a few minor comments.
> 
> I think the only outstanding item is around the base_image used for testing and `build.py` changes?

@nv-kmcgill53 , @rmccorm4 , @Tabrizian - If agreed -I'll remove the base image / build.py changes for now and separate that for E2E testing. The unit level tests do not require them. We can scope these to unit level tests only.

nv-kmcgill53:
>If agreed -I'll remove the base image / build.py changes for now and separate that for E2E testing.

I'll agree to this. I would like to see a CI build for x86, windows, and jetson pass before approval. I think that is it from me.

src/http_server.cc
```diff
@@ -2744,6 +2759,18 @@ HTTPAPIServer::EVBufferToRawInput(
...
+#ifdef TRITON_BIG_ENDIAN
+
+    size_t next_offset = 0;
+    std::vector<char*> partial_result;
+
+    TRITONSERVER_DataType dtype = TRITONSERVER_TYPE_INVALID;
+
+    RETURN_IF_ERR(GetDataTypeForRawInput(server_.get(), model_name, model_version, &dtype)));
+
+#endif
```
rmccorm4, line 2772: nit: Can the extra new lines be removed here? Don't think clang cares.

qa/L0_big_endian/test.sh
nnshah1: yes I think so - these are just unit tests and I think it will be good to have end to end covered as well

src/http_server.cc
```diff
@@ -405,16 +405,23 @@ ReadDataFromJsonHelper(
           const char* cstr;
           size_t len = 0;
           RETURN_IF_ERR(el.AsString(&cstr, &len));
+          if (len > std::numeric_limits<uint32_t>::max()) {
```
Tabrizian, line 408: nice error checking!

Tabrizian (approved):
LGTM! As discussed, it would be great if it can be merged to a side branch until end-to-end testing is added.

rmccorm4 (approved):

PR #6063, opened 2023-07-14
Adding the support tracing of child models invoked from a BLS model

oandreeva-nv:
Mainly testing for:
https://github.com/triton-inference-server/python_backend/pull/277
https://github.com/triton-inference-server/core/pull/234

`opentelemetry_unittest.py` - mostly formatting changes + `send_bls_request` a client for testing tracing on BLS model.

[FIXME] explained in the doc: https://docs.google.com/document/d/1R9oqGcjdcb8vpDiq--ZdGgkCmJ8Vrpt226AsaniFY3o/edit#heading=h.ykdv684u0rnx 
section: OpenTelemetry related issues related to current implementation

[Edit]

I did some refactoring, including:
1) OpenTelemetry Tracer Provider is initialized when TraceManager is created. It is a singleton object, so it is acquired by calling `opentelemetry::trace::Provider::GetTracerProvider()` . Since, I was working on this area of code, I've also added OpenTelemetry Resource specification, i.e. DLIS-5073.

2) OpenTelemetry has an overhead, so we no longer can delete the Trace's `userp` when we see the root trace (i.e. parent id == 0). In my experiments, when BLS model calls for ensemble model, it is highly likely that `TraceRelease` of BLS's model will be called earlier that `TraceRelease` of ensemble, which causes SegFaults. I've added an `instance_tracker_` set, which tracks the number of child traces created for this request and deletes `userp` only when no more `TraceRelease` calls are expected. 

3) I did some experimentation with PerfAnalyzer and concurrent requests to verify and test this implementation.

4) Regarding FIXME - me and Tanmay discussed this issue, I believe we can keep everything as is for now, and introduce changes upon request.

[Edit 2]

Regarding tests.

- To test that HTTP exporter sends traces to specified url, I am using a simple OTel collector, which outputs everything in the output stream.[ Example](https://dl.gitlab-master-pages.nvidia.com/-/dgx/tritonserver/-/jobs/65791822/artifacts/logs/trace_collector_http_exporter.log) . This is done to avoid any implications from OTel exporter expecting a reply from the collector (which was not happenning with `netcat`) and potentially causing unexpected behavior in case of no reply. 

- To verify structure of collected spans, I use OStream exporter, which prints everything in the output stream. [Example](https://dl.gitlab-master-pages.nvidia.com/-/dgx/tritonserver/-/jobs/65791822/artifacts/logs/inference_server_otel_ostream_exporter.log). As you can see, OStream exporter prints spans in json like format, which can easily be post-processed to comply with `json` standard (this is done in the `setup` for unittests), comparing to the collector's output.


Tabrizian (commented):
How would this work if we have a BLS that is calling into an ensemble do we need unit testing for that or it should work properly?

oandreeva-nv:
@Tabrizian Good point, I'll add some testing for that.

docs/user_guide/trace.md
```diff
@@ -427,11 +427,41 @@ The meaning of the trace timestamps is:
 
   * BACKEND_OUTPUT: The tensor in the response of a backend.
 
+## Tracing for BLS models
+
+Triton does not collect traces for child models, invoked from
+[BLS](https://github.com/triton-inference-server/python_backend/tree/main#business-logic-scripting) models.
+
+To include child models into collected traces, user needs to provide `trace`
+argument (as shown in the example), when constructing an InferenceRequest object.
```
rmccorm4, line 436:
```suggestion
Triton does not collect traces for child models invoked from
[BLS](https://github.com/triton-inference-server/python_backend/tree/main#business-logic-scripting)
models by default.

To include child models into collected traces, user needs to provide the `trace`
argument (as shown in the example below), when constructing an InferenceRequest object.
This helps Triton associate the child model with the parent model's trace (`request.trace()`).
```

docs/user_guide/trace.md
```diff
@@ -477,16 +507,30 @@ The following table shows available OpenTelemetry trace APIs settings for
       trace data.
     </td>
     </tr>
+    <tr>
+    <td><code>resource</code></td>
+    <td><code>Empty</code></td>
```
rmccorm4, line 512: Why not set `service.name=triton` by default?

src/tracer.cc
```diff
@@ -575,14 +594,16 @@ TraceManager::Trace::AddEvent(
...
+
+  auto ts = reinterpret_cast<std::shared_ptr<TraceManager::Trace>*>(userp);
+  (*ts)->instance_tracker_.erase(id);
   // The userp will be shared with the trace children, so only delete it
-  // if the root trace is being released
-  if (parent_id == 0) {
-    delete reinterpret_cast<std::shared_ptr<TraceManager::Trace>*>(userp);
+  // if no more TraceRelease calls are expected
+  if ((*ts)->instance_tracker_.empty()) {
+    delete ts;
```
rmccorm4, line 606: Does instance_tracker need to be protected by `ts->mtx_` here?

src/tracer.h
```diff
@@ -150,13 +160,30 @@ class TraceManager {
...
     Trace() : trace_(nullptr), trace_id_(0) {}
     ~Trace();
     std::shared_ptr<TraceSetting> setting_;
     // Group the spawned traces by trace ID for better formatting
     std::mutex mtx_;
     std::unordered_map<uint64_t, std::unique_ptr<std::stringstream>> streams_;
+    // We use the set to track the number of spawned traces, so that
+    // when TraceManager::TraceRelease() with 'trace_userp_' is called
+    // we can safely release 'trace_userp_'
+    std::set<uint64_t> instance_tracker_;
```
tanmayv25, line 186: instance_tracker -> spawned_traces_ 
or something similar.

src/tracer.h
```diff
@@ -122,7 +123,16 @@ class TraceManager {
       const std::string& filepath, const InferenceTraceMode mode,
       const TraceConfigMap& config_map);
 
+#if !defined(_WIN32) && defined(TRITON_ENABLE_TRACING)
+  ~TraceManager()
+  {
+    if (global_setting_->mode_ == TRACE_MODE_OPENTELEMETRY) {
```
tanmayv25, line 129: Can this conditional move into CleanupTracer as well?

tanmayv25 (commented):
some minor comments, otherwise lgtm.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -41,27 +40,29 @@
...
+            json_string = re.sub("\t", "", json_string)
+            json_string = re.sub(r"\b([\w.-]+)\b", r'"\1"', json_string)
+            json_string = re.sub('"span" "kind"', '"span kind"', json_string)
+            json_string = re.sub("{,", "{", json_string)
+            json_string = re.sub(",}", "}", json_string)
+            json_string = re.sub("}{", "},{", json_string)
+            json_string = re.sub(
+                '"events"        : {', '"events"        : [{', json_string
+            )
+            json_string = re.sub('},  "links"', '}],  "links"', json_string)
```
rmccorm4: Can you explain the goal of this part? I find this hard to follow and figure there may be an easier way just working with dictionaries and then reserializing to JSON or something.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -41,27 +40,29 @@
...
+            self.spans = eval(json_string[:-1])
```
rmccorm4, line 65: Use of `eval` is generally a code smell and opens up for [ACE](https://en.wikipedia.org/wiki/Arbitrary_code_execution). I know this is just a test so it's probably not very important, but maybe there's a better way to achieve this? Happy to chat offline to understand what you're trying to do here.

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: The problem is that OStream exporter does not export spans in a json format (i.e. simple `json.loads(json.dumps(data))` won't work), so extracted spans need to be transformed to json. I haven't finished working on it yet, this was put in here for testing on our CI, so it may be simplified later

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: I'll reconsider

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -40,29 +41,54 @@
 
 class OpenTelemetryTest(tu.TestResultCollector):
     def setUp(self):
+        # Extracted spans are in json-like format, thus data needs to be
+        # post-processed, so that `json` could accept it for further
+        # processing
         with open("trace_collector.log", "rt") as f:
```
rmccorm4, line 47: What code generates/outputs this "trace_collector.log" ? Is it Triton code, or open telemetry built-in code?

oandreeva-nv: This line: https://github.com/triton-inference-server/server/pull/6063/files#diff-27e9ca4cd8f292f0e5a94bbabb243e40fe0d47cd55164a0cb49ae6e093c9fbe7R778

It is extracted from server logs into a separate file

qa/L0_trace/trace-config.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+exporters:
+  logging:
+    verbosity: detailed
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      exporters: [logging]
```
rmccorm4, line 45: Not 100% sure if applicable here, but if you set `encoding: json` explcitly, will that remove the need to parse/fix the pseudo-json log format?
```
service:
  telemetry:
    logs:
      encoding: json
```

oandreeva-nv: I'll check, I probably missed that

oandreeva-nv: I'll check for traces, your code seems to be for logs.

oandreeva-nv: This controls collector, it will not have an impact on OStream Exporter. I can check if it is configurable, though

oandreeva-nv: Here are findings: For OTel collector's logging exporter, there is no option to ask a json output: https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/loggingexporter/README.md

For OStream Exporter, which is used on Triton server side, (the one which outputs initial collected traces), there is no option to specify json output either, the format is hardcoded: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/ostream/src/span_exporter.cc#L73

oandreeva-nv: Attached reference refers to specifying collector's logs, it has no impact on the way it exports received traces

src/tracer.cc
```diff
@@ -57,6 +58,10 @@ TraceManager::Create(
   *manager = new TraceManager(
       level, rate, count, log_frequency, filepath, mode, config_map);
 
+  if (mode == TRACE_MODE_OPENTELEMETRY) {
```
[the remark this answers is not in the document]
oandreeva-nv: done

src/tracer.h
```diff
@@ -150,13 +160,30 @@ class TraceManager {
 
   static const char* InferenceTraceModeString(InferenceTraceMode mode);
 
+#if !defined(_WIN32) && defined(TRITON_ENABLE_TRACING)
+  /// Initializes Opentelemetry exporter, processor,
```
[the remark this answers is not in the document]
oandreeva-nv: done

src/tracer.h
oandreeva-nv: done

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: The code is generated by OStream exporter (OpenTelemetry side) : https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/ostream/src/span_exporter.cc#L73

src/tracer.cc
```diff
@@ -617,6 +658,10 @@ TraceManager::TraceActivity(
       reinterpret_cast<std::shared_ptr<TraceManager::Trace>*>(userp)->get();
 
   std::lock_guard<std::mutex> lk(ts->mtx_);
+  if (ts->spawned_traces_tracker_.find(id) ==
+      ts->spawned_traces_tracker_.end()) {
+    ts->spawned_traces_tracker_.emplace(id);
+  }
```
GuanLuo, line 664: How about adding new `TraceSpawn` callback for the original trace object that the user may add to the trace, which will be invoked during `TRITONSERVER_InferenceTraceSpawnChildTrace` if set. The user may use it to trace spawned traces rather than implementing the logic in `TraceActivity` callback.

However, if we were providing the new callback, I would expect it to be where the child trace to be created. In other words, instead of Triton internally creates the child trace, the user will need to do it explicitly in the callback:
```
TRITONSERVER_InferenceTraceSpawnFn_t(parent_trace, *child_trace) {
    TRITONSERVER_InferenceTraceTensorNew(child_trace, ...);
}
```
This is extra effort for the developer but it gives flexibility on how child trace is reported, i.e. the trace level / activity callback can be completely different from parent trace.

docs/user_guide/trace.md
```diff
@@ -477,16 +509,30 @@ The following table shows available OpenTelemetry trace APIs settings for
...
+      <code>--trace-config opentelemetry,resource=<<text>key</text>>=<<text>value</text>></code><br/>
+      For example:<br/>
+      <code>--trace-config opentelemetry,resource=service.name=triton</code><br/>
+      <code>--trace-config opentelemetry,resource=service.version=1</code><br/>
+      Alternatively, key-vaue attributes can be specified through <br/>
+      <a href="https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/#otel_resource_attributes">
+      OTEL_RESOURCE_ATTRIBUTES</a>
+      environment variable.
+    </td>
+    </tr>
```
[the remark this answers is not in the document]
oandreeva-nv: With the suggested syntax, line length goes above 80 chars. The above syntax still generates nice tables

src/tracer.cc
```diff
@@ -362,23 +373,45 @@ TraceManager::Trace::InitTracer(
...
+void
+TraceManager::CleanupTracer()
+{
+  if (global_setting_->mode_ == TRACE_MODE_TRITON) {
+    return;
+  }
+#if !defined(_WIN32) && defined(TRITON_ENABLE_TRACING)
+  std::shared_ptr<otel_trace_api::TracerProvider> none;
+  otel_trace_api::Provider::SetTracerProvider(none);
+#endif
```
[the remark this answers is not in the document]
oandreeva-nv: Could you elaborate, please?

src/tracer.cc
```diff
@@ -404,6 +437,10 @@ TraceManager::Trace::StartSpan(
   if (parent_id == 0 && activity == TRITONSERVER_TRACE_REQUEST_START) {
     parent_span_key = kRootSpan;
   } else if (activity == TRITONSERVER_TRACE_REQUEST_START) {
+    // [FIXME] For BLS requests parent span for children's request spans
+    // should be parent model's compute span. Currently,
+    // this won't work, since parent's compute span will be created
+    // only after children's spans are created.
```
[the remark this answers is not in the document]
oandreeva-nv: Current issue does not have a major impact on OpenTelemetry traces. All child models are still displayed under the parrent's model.  So unless the correct order is requested, we can keep it as is.

Regarding your suggestion, I think it would require a major refactor on trace files to do that. I don't think we need to implement it now.

src/tracer.cc
oandreeva-nv: Should we consider it as a follow up? It seems like a nice addition, but I feel like it needs to be discussed in more details.

src/tracer.cc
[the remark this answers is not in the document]
GuanLuo: Kind of being mentioned in a previous PR, assuming the mode will not be changed at runtime, you can have
```
class OpenTelemetryTraceManager : public TraceManager;
```
which will be created if mode is OpenTelemetry, then there is no runtime check and different branching based on mode selected.

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: In their python instrumentation, they do export spans in `json`: [Example](https://opentelemetry.io/docs/instrumentation/python/automatic/example/). To do this they use [`ConsoleExporter`](https://github.com/open-telemetry/opentelemetry-python/blob/main/docs/examples/auto-instrumentation/server_programmatic.py#L27), defined [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/export/__init__.py#L499). As you can see, they invoke `to_json` on every processed span, defined [here](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L455). Thus, they don't have an explicit parser, that can help us here.

oandreeva-nv: @rmccorm4  and @GuanLuo how about I create a request on opentelemetry-cpp github to add an option for OStream to print spans in json format and will monitor. As soon as they add this, I'll change this part of the test.

rmccorm4: Sharing the json file contrib mentioned offline for future reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/fileexporter, but as concluded it's not shipped with the standard set of exporters out of the box, probably not worth the effort figuring out how to integrate it.

> how about I create a request on opentelemetry-cpp github to add an option for OStream to print spans in json format and will monitor. As soon as they add this, I'll change this part of the test.

No harm in raising the request to add flag for ostream exporter, go for it. After poking around, it doesn't seem like there are quick of easy options for improving the parsing/ingestion of these logs.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -121,32 +153,38 @@ def _check_parent(self, child_span, parent_span):
         # Check that child and parent span have the same trace_id
         # and child's `parent_span_id` is the same as parent's `span_id`
         self.assertEqual(child_span["trace_id"], parent_span["trace_id"])
-        self.assertIn(
-            "parent_span_id",
-            child_span,
+        self.assertNotEqual(
+            child_span["parent_span_id"],
+            "0000000000000000",
```
rmccorm4, line 158: Can this be a descriptive variable instead of magic number? Not sure what all zeroes or number of zeroes importance is.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -214,6 +255,16 @@ def prepare_traces():
...
     triton_client_http.infer("ensemble_add_sub_int32_int32_int32", inputs)
 
+    send_bls_request(model_name="ensemble_add_sub_int32_int32_int32")
+
+
+def send_bls_request(model_name="simple"):
+    with httpclient.InferenceServerClient("localhost:8000") as client:
+        inputs = prepare_data(httpclient)
+        inputs.append(httpclient.InferInput("MODEL_NAME", [1], "BYTES"))
+        inputs[2].set_data_from_numpy(np.array([model_name], dtype=np.object_))
```
rmccorm4, line 265: Use index `-1` rather than `2` if last input to generalize better if prepare_data adds other inputs in future
```suggestion
        inputs[-1].set_data_from_numpy(np.array([model_name], dtype=np.object_))
```

qa/L0_trace/test.sh
```diff
@@ -649,37 +647,109 @@ fi
...
+if [ `grep -c "\"trace_file\":\"bls_trace.log\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 
+set +e
+# Send bls requests to make sure simple model is traced
+for p in {1..4}; do
+    python -c 'import opentelemetry_unittest; \
+        opentelemetry_unittest.send_bls_request(model_name="ensemble_add_sub_int32_int32_int32")'  >> client_update.log 2>&1
+done
```
rmccorm4, line 659: I don't really understand this part - why isn't it somewhere in the python code, rather than being called from bash?

qa/L0_trace/test.sh
```diff
@@ -649,37 +647,109 @@ fi
...
+fi
+
+if [ `grep -c ^simple summary_bls.log` != "1" ]; then
+    cat summary_bls.log
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+if [ `grep -c 'parent id' bls_trace.log` == "1" ]; then
+    cat summary_bls.log
```
rmccorm4, line 683: Should this be `bls_trace.log`?
```suggestion
    cat bls_trace.log
```

rmccorm4, line 683: Should this be `bls_trace.log`?

```suggestion
if [ `grep -c 'parent id' bls_trace.log` == "1" ]; then
    cat bls_trace.log
```

qa/L0_trace/test.sh
```diff
@@ -649,37 +647,109 @@ fi
...
 kill $SERVER_PID
 wait $SERVER_PID
 
 set +e
 
+$TRACE_SUMMARY -t bls_trace.log > summary_bls.log
+
+if [ `grep -c "COMPUTE_INPUT_END" summary_bls.log` != "2" ]; then
+    cat summary_bls.log
+    echo -e "\n***\n*** Test Failed\n***"
```
rmccorm4, line 672: Can you make the failure messages more descriptive? "Test failed: unexpected number of COMPUTE_INPUT_END found" - etc.

src/tracer.cc
```diff
@@ -57,6 +61,8 @@ TraceManager::Create(
   *manager = new TraceManager(
       level, rate, count, log_frequency, filepath, mode, config_map);
 
+  (*manager)->InitTracer(config_map);
```
rmccorm4, line 64: Why not call this in `TraceManager` constructor?

qa/L0_trace/test.sh
oandreeva-nv: That is to test that Triton Trace API logs bls request. I didn't want to write a separate client for this. `opentelemetry_unittest.py` tests OpenTelemetry Trace API.

src/tracer.cc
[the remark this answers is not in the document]
oandreeva-nv: Sure, I can introduce it in a refactoring PR.

qa/L0_trace/test.sh
oandreeva-nv: summary will have a clearer picture in my mind

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: When parent for span is not specified, `parent_span_id` field equal to "0000000000000000".

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -37,6 +37,7 @@
 import tritonclient.http as httpclient
 
 EXPECTED_NUM_SPANS = 16
+NO_PARENT_SPAN = "0000000000000000"
```
rmccorm4, line 40: Who sets this? Is it an open telemetry default value?

On quick search I'm guessing otel based on [this](https://opentelemetry.io/docs/instrumentation/cpp/getting-started/), but not finding a spec or mention of the default. Can you link one in a comment above this?

oandreeva-nv: It is set here: https://github.com/open-telemetry/opentelemetry-cpp/blob/main/exporters/ostream/src/span_exporter.cc#L78C54-L78C68. I'll add comment

rmccorm4: Is there a doc that mentions this instead? If not, use permalink to code as lines will change over time on `main`

src/tracer.cc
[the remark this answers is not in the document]
oandreeva-nv: For reference: all refactoring suggestions are tracked under this ticket: https://jirasw.nvidia.com/browse/DLIS-4803

rmccorm4 (approved):
LGTM with DLIS-4803 for follow-ups

PR #6068, opened 2023-07-17
Fix broken links pointing to the `grpc_server.cc` file

matemijolovic:
`grpc_server.cc` was recently moved to `grpc` folder, so I'm updating the docs links to point to the correct location


Tabrizian:
@matemijolovic Thanks for the fixes.

Tabrizian (approved):

PR #6097, opened 2023-07-24
Fix L0_backend_python lifecycle default instance name

kthui:
After merging https://github.com/triton-inference-server/core/pull/231 PR, the default instance name no longer varies based on the instance group count.

For example, in the past, if `count = 1`, the default instance name is `model_name_0`. If `count > 1`, the default instance names are `model_name_0_0, model_name_0_1, ...` After the change, the default instance names will always be `model_name_0_0, model_name_0_1, ...` regardless of count.


kthui:
cc @jbkyang-nvi for passing L0_backend_python

PR #5935, opened 2023-06-12
add the -L parameter to fix #5934

is:
as title.

PR #6099, opened 2023-07-25
fix dcgm installation in build.py

w169q169:
Building server in container will cause the following error:

```bash
./build.py -v --enable-all
```

bug

In the Americas region, `developer.download.nvidia.com` simply downloads the file.
`developer.download.nvidia.com` just download file
image

In China and other regions, `developer.download.nvidia.com` will return a 302 redirect response. 
The curl command is unable to download the file with the -L option enabled.
image


rmccorm4:
Hi @w169q169, thanks for the contribution? Have you filled out a [CLA](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla)?

@nv-kmcgill53 @mc-nv could you sanity check it's OK to follow redirects here when you get a chance?

rmccorm4:
Looks like this is also a duplicate of #5935

nv-kmcgill53 (commented):
I am hesitant to allow "follow redirects" being checked for security reasons. I need to follow up with our team before we can merge.

mc-nv:
> I am hesitant to allow "follow redirects" being checked for security reasons. I need to follow up with our team before we can merge.

@nv-kmcgill53 I don't see any issue with follow redirect on build process. We are not responsible for security issues users may face during their own build, in our case we should be safe unless there is internal attack.

nv-kmcgill53 (approved):

mc-nv:
@nv-kmcgill53  are we still willing to have this change?

nv-kmcgill53:
@mc-nv You have an outstanding comment [here](https://github.com/triton-inference-server/server/pull/5935#pullrequestreview-1873192732). Do you need to look into the other repos before we merge?

mc-nv:
> @mc-nv You have an outstanding comment [here](https://github.com/triton-inference-server/server/pull/5935#pullrequestreview-1873192732). Do you need to look into the other repos before we merge?

I came with assumption to close this PR unless we are going to prioritize and work on it. 
It's relatively small change, users always can workaround it if needed by adding that flag. 
I haven't seen any complains from our SA's about it.

nv-kmcgill53:
If there is pending work on this it's not getting prioritized. We can close this PR in this case.

PR #6109, opened 2023-07-26
Test python environments 3.8-3.11

jbkyang-nvi:


Tabrizian (commented):
This looks good but I think it only tests a small subset of Python BE functionalities. Do you think we can create variants of `L0_backend_python` that run with a stub other than the default compiled stub? I think if we can just copy the stub created here in `/opt/tritonserver/backends/python/triton_python_backend_stub` and install the Python 3.X in the container the rest of the tests can run as-is.

jbkyang-nvi:
@Tabrizian we can create variants of L0_backend_python, which is already pretty large. Do you think its worthwhile to make the test run for all variants 4 times? It will take a lot of time. The python_backend test takes 139 minutes 51 seconds to run on the latest run

qa/L0_backend_python/test.sh
```diff
@@ -132,296 +133,314 @@ mkdir -p models/dlpack_identity/1/
...
 fi
 
-prev_num_pages=`get_shm_pages`
-run_server
-if [ "$SERVER_PID" == "0" ]; then
-    cat $SERVER_LOG
-    echo -e "\n***\n*** Failed to start $SERVER\n***"
-    exit 1
-fi
+for PYTHON_ENV in ${PYTHON_ENV_VERSION}; do
```
Tabrizian, line 149: I was wondering why do we need to iterate on all the Python versions? I think we want separate jobs that create a stub and setup the environment based on the PYTHON_ENV_VERSION listed. Then, we only need to change the environment version for each job to use a different Python version.

We would have 3 different variants of L0_backend_python for testing different Python versions (`L0_backend_python--py3.x`) + the default L0_backend python. Each job should only test a single Python version with the exception of the `env` subset.

jbkyang-nvi: we are not. We put them in one at a time on the CI.

qa/L0_backend_python/setup_python_enviroment.sh
```diff
@@ -0,0 +1,172 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+RET=0
+set -e
+if [ ${PYTHON_ENV_VERSION} = "10" ]; then
```
tanmayv25, line 29: Can we obtain the default python version programmatically? Specifying `10` explicitly in different places in the workflow would make it hard to update this version when we upgrade default python version.

qa/L0_backend_python/test.sh
```diff
@@ -397,12 +400,14 @@ if [ "$TEST_JETSON" == "0" ]; then
         rm -fr venv
     done
 
-    # In 'env' test we use miniconda for dependency management. No need to run
-    # the test in a virtual environment.
-    (cd env && bash -ex test.sh)
-    if [ $? -ne 0 ]; then
-        echo "Subtest env FAILED"
-        RET=1
+    if [ ${PYTHON_ENV_VERSION} = "10" ]; then
```
tanmayv25, line 403: same as above ^

qa/L0_backend_python/setup_python_enviroment.sh
jbkyang-nvi: I think we can test if the current installed version is 10, and if so fail the test so we can update everything.
We currently test python3.10 as the default but if it is not, we will need to add another test and update the default test. How does that sound?

jbkyang-nvi: @tanmayv25 ^

tanmayv25 (approved):

PR #6126, opened 2023-07-31
Add sanity tests for parallel instance loading

rmccorm4:
1. Test speedup: Same way as parallel model loading - create 2 python model instances each with a sleep(10) in initialize function. Then confirm that total load time took less than 15 seconds (rather than >= 20 for the serial case).
2. Test sanity: For each enabled backend, load N instances in parallel, setup scheduler in such a way that we can guarantee sending 1 request per instance, make sure no failures occur, make sure server gracefully loads/unloads/etc.
    - Looking for feedback on this test. If anyone has any better or additional ideas, let me know.
    - Ideally this test would fail either consistently or frequently when a backend is not thread-safe for instance loading.


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2916,6 +2917,189 @@ def test_concurrent_load_unload(self):
...
 
+    def test_concurrent_model_instance_load_speedup(self):
+        # Initialize client
+        try:
+            triton_client = httpclient.InferenceServerClient(
+                "localhost:8000", verbose=True
+            )
+        except Exception as ex:
+            self.assertTrue(False, "unexpected error {}".format(ex))
+        models = ["identity_fp32"]
```
kthui, line 2928: Maybe we can also add a Python model for testing?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2916,6 +2917,189 @@ def test_concurrent_load_unload(self):
...
+        # Inputs
+        inputs = []
+        for name in input_names:
+            inputs.append(
+                client_type.InferInput(name, shape, dtype.replace("TYPE_", ""))
+            )
+            inputs[-1].set_data_from_numpy(np.ones(shape, dtype=np.float32))
+        return input_config, output_config, inputs
+
+    def test_concurrent_model_instance_load_sanity(self):
```
kthui, line 3001: Just for my understanding. The goal of this test is to check if the included backends are capable of loading model instances concurrently?

qa/L0_lifecycle/lifecycle_test.py
rmccorm4: `identity_fp32` is a python model, `../python_models/identity_fp32`, I can rename it to be consistent with the other test.

qa/L0_lifecycle/lifecycle_test.py
rmccorm4: The goal is to check that nothing crashes and burns internally when trying to load instances in parallel for the supported backends. All it does it load N instances of various kinds, and ensure that exactly N instances are loaded (via sequence batch slot logic) and complete a successful inference (no exception is raised on the N inferences).

rmccorm4: Open to ideas if you think there's a better way to test

kthui: I think the test case does not belong here, because the L0_lifecycle is for testing the correctness of the Triton core, while the correctness of the backend implementation belongs to the testing of each backend. Since the test is similar for all backends, maybe a new L0_backend_thread_safety (or something similar) can be added for this test?
@GuanLuo @tanmayv25 should correct me if you think otherwise

rmccorm4: @kthui do you have any feedback on the test itself other than how it is named? Or do you think it is sufficient for now?

Also @Tabrizian @tanmayv25 any feedback?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2916,6 +2917,189 @@ def test_concurrent_load_unload(self):
...
+                    # Unload the model
+                    try:
+                        triton_client.unload_model(model)
+                    except Exception as ex:
+                        self.assertTrue(False, "unexpected error {}".format(ex))
+
+                    # Allow server to fully unload model before next test iteration
+                    num_tries = 10
+                    for i in range(num_tries):
+                        if triton_client.is_server_ready():
```
Tabrizian, line 3091: Does the server become not ready when a model is unloaded? I was wondering whether we should make `unload_model` blocking? I think it is really confusing that it is async. (cc @kthui @GuanLuo). Also, even if it is async we should probably update this example to better demonstrate the usage: https://github.com/triton-inference-server/client/blob/f7c45d382639f31a5162666034fae7852681f082/src/python/examples/simple_grpc_model_control.py#L98-L101

qa/L0_lifecycle/lifecycle_test.py
Tabrizian: I was wondering how is this testing the parallel instance feature? I see that it tests whether all the instances are ready or not but I don't see any testing in this specific case for parallel instance loading. I believe this test case would pass before this feature too. Please correct me if I'm missing anything.

rmccorm4: @Tabrizian you are correct that this subtest doesn't explicitly the parallelism in terms of speedup, I did that in the other subtest as it is easy to do for a python model, similar to the existing parallel model loading tests.

This sanity subtest was written as more of a sanity check that when loading a model with a significant (N=5, though upon writing I can probably increase this) instance count that (1) nothing crashes and burns, with higher instance counts it would be likely to reveal a race condition or segfault, hang, etc. if there was a bug, and (2) verify that exactly N instances (not more, not less, due to any possible race condition) do actually get loaded and can handle inference.

Upon reflection though, I can probably modify this test to do a serial load (env var) and then a parallel load and just check that the parallel load is faster at all. Initially I was afraid of this check introducing flakiness, but ideally it should consistently be faster on the types of CI systems we have with higher core counts.

Let me know if that makes sense, or if you have other ideas.

qa/L0_lifecycle/lifecycle_test.py
rmccorm4: > Does the server become not ready when a model is unloaded?
Server ready is true when all models are ready for inferencing, so while a model is in the unloading state, server is considered not ready.

> I was wondering whether we should make unload_model blocking? I think it is really confusing that it is async.

I have been thinking about this recently when seeing Jacky's model update tests requiring polling, Kris's BLS test unload issues, and this test as well. I second that it's definitely confusing and unintuitive. 

I do believe we document that it returns without waiting for model to finish in most places. However, I still think it is unituitive and not symmetric with `load_model` which is blocking.

rmccorm4: NOTE: I understand there's good reason for an async unload to allow inflight requests to finish without a long blocking/wait time, but it would be nice to have an alternative or optional blocking call to support both.

Tabrizian (dismissed):

rmccorm4:
I will improve these tests and consider moving to a separate L0 test over time while working on enabling further backends, but would like to get this in to help stress test instance counts of the python/ORT backends during code freeze/release process.

qa/L0_lifecycle/lifecycle_test.py
GuanLuo: My reasoning is that when `unload` returns, the model is no longer accessible from the client. Does it matter from the client's perspective that whether the model is fully cleaned up on server side when `unload` returns?

PR #6130, opened 2023-08-01
Reduce instance count to 1 for python bls model loading test

krishung5:
To avoid unsafely using the model loading API, reduce the model instance count to 1.

### DIFF
--- a/qa/python_models/bls_model_loading/config.pbtxt
+++ b/qa/python_models/bls_model_loading/config.pbtxt
@@ -37,7 +37,7 @@ output [
 
 instance_group [
   {
-    count: 3
+    count: 1
     kind: KIND_CPU
   }
 ]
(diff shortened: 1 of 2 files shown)


qa/python_models/bls_model_loading/config.pbtxt
```diff
@@ -37,7 +37,7 @@ output [
 
 instance_group [
   {
-    count: 3
+    count: 1
```
rmccorm4, line 40: Why does an instance count > 1 make it unsafe to use? It's not related to parallel instance loading, right?

krishung5: No, I don't think it's related to parallel instance loading. It's just that all the unit tests are executed during the initialize phase, and loading three model instances simultaneously might cause interference between them. For instance, unloading a model in one instance could impact another unit test that relies on that model in other model instances. This is what I meant by unsafely using the model loading API. In the general usage where the `unload` call is not recommended, using multiple instances should be fine.

rmccorm4 (approved):

PR #6147, opened 2023-08-04
Use targetmodel from header as model name in SageMaker

nskool:
1. So far, model_name is provided by SageMaker as a 'model_hash' (for other naming and abstraction reasons). We have been using this as model_name, but logging the actual model archive name (provided as a header) as a log-line when LOADING, UNLOADING, and INVOKING the model.
2. However, while this works alright, a better experience is to identify the model with the archive name specified by users when calling SageMaker.
3. This also helps identifying the metrics associated with the model archive when metrics publishing is enabled.
4. Model archive name and the model hash have a 1:1 mapping for a given instance in SageMaker, and hence maybe used interchangeably.
5. This PR enables changes the model 'id' to use the archive name from the header when available, otherwise default to the model hash if not available (which can be the case when testing locally).

### DIFF
--- a/src/sagemaker_server.cc
+++ b/src/sagemaker_server.cc
@@ -226,9 +226,21 @@ SagemakerAPIServer::Handle(evhtp_request_t* req)
             evhtp_send_reply(req, EVHTP_RES_NOTFOUND); /* 404*/
             return;
           }
-          LOG_VERBOSE(1) << "SageMaker MME Custom Invoke Model Path"
-                         << std::endl;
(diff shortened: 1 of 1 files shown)


nskool:
@GuanLuo could you help take a look at this PR

GuanLuo (approved):
Sorry for the delay, will run a pipeline for it. Also, you may need to address the pre-commit check if it fails

nskool:
@GuanLuo did the pipeline pass? Can we merge this in?

PR #6150, opened 2023-08-05
Add test when unload/load requests for same model is received at the same time

kthui:
To produce failure if number 2 path on https://github.com/triton-inference-server/core/pull/236 PR is mistakenly taken.


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2885,6 +2885,21 @@ def test_concurrent_load_unload(self):
...
+            unload_thread = pool.submit(
+                triton_client.unload_model, "identity_zero_1_int32"
+            )
+            load_thread = pool.submit(triton_client.load_model, "identity_zero_1_int32")
+            unload_thread.result()
+            load_thread.result()
+        triton_client.unload_model("identity_zero_1_int32")  # to unload if loaded
+        self.assertTrue(triton_client.is_server_live())
+        self.assertTrue(triton_client.is_server_ready())
+        triton_client.is_model_ready("identity_zero_1_int32")
```
GuanLuo, line 2902: validate?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2885,6 +2885,21 @@ def test_concurrent_load_unload(self):
         self.assertTrue(triton_client.is_server_live())
         self.assertTrue(triton_client.is_server_ready())
         self.assertFalse(triton_client.is_model_ready("identity_zero_1_int32"))
+        # Unload identity_zero_1_int32 and immediately load it
+        # The model can either be loaded or unloaded but server must not crash
```
GuanLuo, line 2889: In the case you try to hit, doesn't the order matter? i.e. load immediately after an unload. How often does this test catch the failure without your earlier fix?

qa/L0_lifecycle/lifecycle_test.py
kthui: It could be true or false, but it is refactored and removed.

qa/L0_lifecycle/lifecycle_test.py
kthui: Yes, the order matters. It is refactored into a separate case to have better checking.

If `load -> unload`: The load will block the unload, which the model is unloaded at the end.
If `unload -> load`: The unload will block the load, which the model is loaded at the end.

Since the unload is async, if it is performed at the end, a poll/timeout is needed to ensure the unload is completed such that `is_model_ready` produces the correct result.

Without the earlier fix, the chance of catching it is about 1/50.

GuanLuo: > Without the earlier fix, the chance of catching it is about 1/50.

This may not be a good test case then.. Or at least you need to put emphasize that one must look into the failure when this particular test fails. Otherwise, I can see when something actually breaks, and luckily get caught by this test, one may just re-run the test and say it is just turbulence unless the re-run hits another 1/50 chance.

kthui: Rewrite the test to perform stress test on load and unload API instead. The rewritten test failed 10/10 without the earlier fix, and passed 10/10 with the earlier fix.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2917,6 +2917,62 @@ def test_concurrent_model_load_unload(self):
...
+
+        def _load_unload():
+            for i in range(num_iterations):
+                try:
+                    triton_client.load_model(model_name)
+                except InferenceServerException as ex:
+                    # Acceptable for an unload to happen after a load completes, but
+                    # before the load can verify its load state.
+                    fail_reasons = [
+                        "unexpected miss in global map",
```
rmccorm4, line 2940: I don't think I've ever seen this error, is this a valid user-facing error?

rmccorm4 (commented):
LGTM overall, just one question

qa/L0_lifecycle/lifecycle_test.py
kthui: Those errors are from https://github.com/triton-inference-server/core/blob/main/src/model_repository_manager.cc#L663-L706.

For model load, notice the `mu_` that ensure atomic operation is release briefly between the actual load completion and validation. If the validation fails, the errors could be printed. i.e. an unload of the same model happens while the `mu_` is briefly released. Thus, the check is there to prevent intermittent failure.
(I was never able to produce any load error, maybe the gap is simply too small, but the probability is not zero I think.)

README.md
```diff
@@ -38,15 +38,15 @@ and corresponds to the 23.07 container release on
...
+multiple deep learning and machine learning frameworks, including TensorRT,
+TensorFlow, PyTorch, ONNX, OpenVINO, Python, RAPIDS FIL, and more. Triton
+Inference Server supports inference across cloud, data center, edge and embedded
+devices on NVIDIA GPUs, x86 and ARM CPU, or AWS Inferentia. Triton Inference
+Server delivers optimized performance for many query types, including real time,
 batched, ensembles and audio/video streaming. Triton inference Server is part of
-[NVIDIA AI Enterprise](https://www.nvidia.com/en-us/data-center/products/ai-enterprise/), 
-a software platform that accelerates the data science pipeline and streamlines 
+[NVIDIA AI Enterprise](https://www.nvidia.com/en-us/data-center/products/ai-enterprise/),
+a software platform that accelerates the data science pipeline and streamlines
```
GuanLuo, line 49: unrelated change?

docs/index.md
```diff
@@ -60,15 +60,15 @@ Triton Inference Server is an open source inference serving software that stream
...
 PyTorch, ONNX, OpenVINO, Python, RAPIDS FIL, and more. Triton supports inference
 across cloud, data center, edge and embedded devices on NVIDIA GPUs, x86 and ARM
 CPU, or AWS Inferentia. Triton Inference Server delivers optimized performance
-for many query types, including real time, batched, ensembles and audio/video 
+for many query types, including real time, batched, ensembles and audio/video
 streaming. Triton inference Server is part of
-[NVIDIA AI Enterprise](https://www.nvidia.com/en-us/data-center/products/ai-enterprise/), 
-a software platform that accelerates the data science pipeline and streamlines 
+[NVIDIA AI Enterprise](https://www.nvidia.com/en-us/data-center/products/ai-enterprise/),
+a software platform that accelerates the data science pipeline and streamlines
```
GuanLuo, line 71: same

docs/user_guide/faq.md
```diff
@@ -165,41 +165,41 @@ the backtrace to better help us resolve the problem.
...
 including savings from reduced time and energy consumption.
 
-### Optimized and Certified to Deploy Everywhere: 
-Cloud, Data Center, Edge Optimized and certified to ensure reliable performance 
-whether it’s running your AI in the public cloud, virtualized data centers, or 
-on DGX systems. 
+### Optimized and Certified to Deploy Everywhere:
+Cloud, Data Center, Edge Optimized and certified to ensure reliable performance
+whether it’s running your AI in the public cloud, virtualized data centers, or
+on DGX systems.
```
GuanLuo, line 205: same

GuanLuo (dismissed):
Need to revert doc changes

kthui:
> Need to revert doc changes

It is the last commit
image
to pass pre-commit. 

The unrelated docs change is reverted.

README.md
kthui: Reverted

docs/index.md
kthui: Reverted

docs/user_guide/faq.md
kthui: Reverted

rmccorm4 (approved):
LGTM. Long term I think we should consider closing the gaps for possible race in parallel loading/unloading. My current understanding is that there is a small section with chance for race condition, but these races are somewhat harmless and return errors. In the future, it would be nice to know that conflicting parallel loads/unloads would be serialized and not have expected errors.

PR #6157, opened 2023-08-07
Improve L0_backend_python debugging

krishung5:
- [x] Improve the debugging for the `bls` test so that the correct server log is printed out when a sub-test fails.
- [x] Copy all logs and core dumps from subdirectories to the upper-level directory so that the CI logic can capture all artifacts.

### DIFF
--- a/qa/L0_backend_python/argument_validation/test.sh
+++ b/qa/L0_backend_python/argument_validation/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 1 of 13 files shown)


qa/L0_backend_python/argument_validation/test.sh
```diff
@@ -75,4 +75,8 @@ else
     echo -e "\n***\n*** Argument validation test PASSED. \n***"
 fi
 
+# Collect all logs and core dumps and copy them to an upper-level directory for
+# proper capture on the CI.
+cp *.*log* core* ../ || true
```
rmccorm4, line 80: Nice updates.

Can this be some function in ../common/util.sh so it only needs to be updated in one spot?

krishung5: Great idea! Updated.

rmccorm4 (approved):

Issue #6153, opened 2023-08-07
[Trace_rate] If trace_rate value is more than 10 digits it crashes the server.


Kanupriyagoyal:
**Description**
Upon providing larger “trace_rate” value in the POST body of the “/v2/models/gbm_model/trace/settings”, it was observed that the container stopped or crashed allowing the server to crash. 

If the application goes down, the user is not able to access his/her data.
It can overload your server and consume all of its resources which makes the service unavailable.

**Triton Information**
r23.04

Are you using the Triton container or did you build it yourself?
Using same instructions as build.py

**To Reproduce**
Steps to reproduce the behavior.
1. Invoke "POST /v2/models/gbm_model/trace/setting" API with "trace_rate" parameter value "1000" and observed that it is updated the "trace_rate" to "1000" in the response.
2. Change the "trace_rate" value to "10000000000" and send the request. As the trace_rate digit is more than 10 digits it fails
3. Post which the container breaks and the APIs becomes inaccessible.

Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).

**Expected behavior**
Application should look at the requests. If there is any limitation at the server, it should return a proper error description to user and should not process further requests.
If number of zeros == 9:
<img width="760" alt="Screenshot 2023-08-07 at 3 34 44 PM" src="https://github.com/triton-inference-server/server/assets/39382030/610211f8-1b95-4bdc-a56a-07a2f75ce5f9">
If number of zeros == 10 or number of digits more than 10:
<img width="760" alt="Screenshot 2023-08-07 at 3 34 56 PM" src="https://github.com/triton-inference-server/server/assets/39382030/3aff4b04-2dee-4ae7-a7a0-5c7c34f57cc9">

pradghos:
@dyastremsky: Any suggestion would help. Thank you !

GuanLuo:
Probably due to overflow, 10^10 takes more than 32 bits and the type for stroing rate is `uint32_t`. CC @oandreeva-nv on whether it can be a quick fix

oandreeva-nv:
I'll take a look

oandreeva-nv:
Hi @pradghos and @Kanupriyagoyal , could you please briefly describe your workload and what ranges you would like to see supported by our Trace APIs?: https://github.com/triton-inference-server/server/blob/main/docs/user_guide/trace.md#global-settings

[...]

PR #6162, opened 2023-08-08
Update docs with NVAIE messaging

dzier:

PR #6173, opened 2023-08-10
Adding safe string to number conversions

oandreeva-nv:
Related issue: https://github.com/triton-inference-server/server/issues/6153

Issue fix is in files `http_server` and `grpc_server` - added catch for out of range exception .

I've also added a wrapper to `ParseOptions` in `command_line_parser` to handle `stoi`, `stoull`, `stoll` exceptions + report option, for which conversion didn't work.

### DIFF
--- a/qa/L0_trace/test.sh
+++ b/qa/L0_trace/test.sh
@@ -481,36 +481,20 @@ if [ -f ./global_trace.log.0 ]; then
     RET=1
 fi
 
-# Check out of range errors
-rm -f ./curl.out
-set +e
-code=`curl -s -w %{http_code} -o ./curl.out -d'{"trace_count":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
-set -e
-if [ "$code" != "400" ]; then
-    cat ./curl.out
-    echo -e "\n***\n*** Test Failed\n***"
-    RET=1
-fi
+SETTINGS="trace_count trace_rate log_frequency"
 
-rm -f ./curl.out
-set +e
-code=`curl -s -w %{http_code} -o ./curl.out -d'{"trace_rate":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
-set -e
-if [ "$code" != "400" ]; then
-    cat ./curl.out
-    echo -e "\n***\n*** Test Failed\n***"
-    RET=1
-fi
-
-rm -f ./curl.out
-set +e
-code=`curl -s -w %{http_code} -o ./curl.out -d'{"log_frequency":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
-set -e
-if [ "$code" != "400" ]; then
-    cat ./curl.out
-    echo -e "\n***\n*** Test Failed\n***"
-    RET=1
-fi
+for SETTING in $SETTINGS; do
+    # Check `out of range` errors
+    rm -f ./curl.out
+    set +e
+    code=`curl -s -w %{http_code} -o ./curl.out -d'{"'${SETTING}'":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
+    set -e
+    if [ "$code" != "400" ]; then
+        cat ./curl.out
+        echo -e "\n***\n*** Test Failed\n***"
+        RET=1
+    fi
+done
 
 set -e
 
--- a/qa/L0_trace/test.sh
+++ b/qa/L0_trace/test.sh
@@ -485,15 +485,8 @@ SETTINGS="trace_count trace_rate log_frequency"
 
 for SETTING in $SETTINGS; do
     # Check `out of range` errors
-    rm -f ./curl.out
-    set +e
-    code=`curl -s -w %{http_code} -o ./curl.out -d'{"'${SETTING}'":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
-    set -e
-    if [ "$code" != "400" ]; then
-        cat ./curl.out
-        echo -e "\n***\n*** Test Failed\n***"
-        RET=1
-    fi
+    update_trace_setting "simple" '{"'${SETTING}'":"10000000000"}'
+    assert_curl_failure "Server modified '${SETTING}' with an out of range value."
 done
 
 set -e
@@ -714,7 +707,7 @@ OTEL_COLLECTOR_LOG="./trace_collector_http_exporter.log"
 # Building the latest version of the OpenTelemetry collector.
 # Ref: https://opentelemetry.io/docs/collector/getting-started/#local
 if [ -d "$OTEL_COLLECTOR_DIR" ]; then rm -Rf $OTEL_COLLECTOR_DIR; fi
-git clone https://github.com/open-telemetry/opentelemetry-collector.git
+git clone --depth 1 --branch v0.82.0 https://github.com/open-telemetry/opentelemetry-collector.git
 cd $OTEL_COLLECTOR_DIR
 make install-tools
 make otelcorecol
--- a/qa/L0_trace/test.sh
+++ b/qa/L0_trace/test.sh
@@ -481,6 +481,37 @@ if [ -f ./global_trace.log.0 ]; then
     RET=1
 fi
 
+# Check out of range errors
+rm -f ./curl.out
+set +e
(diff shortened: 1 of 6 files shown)


oandreeva-nv:
@dyastremsky by any chance, do you know why pre-commit failed on docs, I haven't touched?

dyastremsky:
> @dyastremsky by any chance, do you know why pre-commit failed on docs, I haven't touched?

[This PR](https://github.com/triton-inference-server/server/pull/6162) had an Admin override of the failing checks and was merged. Pre-commit only checks the files included in a commit locally, but it looks like the GitHub-side version checks all files. I created a PR to fix the formatting issues, so that it doesn't impact your PR. Would you be able to review/approve it [here](https://github.com/triton-inference-server/server/pull/6174)?

dyastremsky:
> > @dyastremsky by any chance, do you know why pre-commit failed on docs, I haven't touched?
> 
> [This PR](https://github.com/triton-inference-server/server/pull/6162) had an Admin override of the failing checks and was merged. Pre-commit only checks the files included in a commit locally, but it looks like the GitHub-side version checks all files. I created a PR to fix the formatting issues, so that it doesn't impact your PR. Would you be able to review/approve it [here](https://github.com/triton-inference-server/server/pull/6174)?

The PR to fix this issue was merged. Please feel free to pick up those commits from main, and the pre-commit failures should go away.

qa/L0_trace/test.sh
```diff
@@ -543,6 +543,37 @@ if [ -f ./global_trace.log.0 ]; then
...
+
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -d'{"log_frequency":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
```
rmccorm4, line 575: Since it looks like the check/format is the same, can you just loop over the 3 settings instead of copy/paste?

src/command_line_parser.cc
```diff
@@ -96,44 +97,87 @@ namespace triton { namespace server {
...
+// A wrapper around std::stoi, std::stoull, std::stoll, std::stod
+// to catch `invalid argument` and `out of range` exceptions
+template <typename T>
+T
+SafeConvertion(
+    const std::string& arg, std::function<T(const std::string&)> stox,
+    const std::string& opt_name = "")
+{
+  try {
+    return stox(arg);
```
nnshah1, line 109: One thing you might be able to do is use templates for stoxa and then could get rid of the function argument - and just have ParseOption be safeconversion...  Would something like this work?

```
template <typename T>
T stox(const std::string);

template<>
int stox();

template <typename T>
T ParseOption(string, name) {
  try {
      return stox<T>(arg) ;
} catch () {
}
}
```

oandreeva-nv: I am not sure I understand how it will work with `stod`, `stoull` and `stoll` as well

nnshah1: I might be missing something here - but was thinking you could add template instantiations for `stox()` for each -

oandreeva-nv: Ah, so instead of bindings instantiate templates for `stox` ? That wasn't clear for me initially

nnshah1: yep - if it works. I liked the way you used templates here and then just got to thinking if it could be used instead of bindings ...

oandreeva-nv: Sounds good, I'll try this out

src/command_line_parser.cc
```diff
@@ -96,44 +111,71 @@ namespace triton { namespace server {
...
+// There must be specialization for the types to be parsed into so that
+// the argument is properly validated and parsed. Attempted to use input
+// operator (>>) but it will consume improper argument without error
+// (i.e. parse "1.4" to 'int' will return 1 but we want to report error).
+template <typename T>
+T
+ParseOption(const std::string& arg, const std::string& opt_name = "")
+{
+  try {
+    return StoX<T>(arg);
```
GuanLuo, line 156: Call something like `StringTo<T>`, `X` is already covered by the type argument.

src/command_line_parser.cc
```diff
@@ -1251,7 +1315,8 @@ TritonParser::Parse(int argc, char** argv)
         break;
         break;
       case OPTION_HTTP_THREAD_COUNT:
-        lparams.http_thread_cnt_ = ParseOption<int>(optarg);
+        lparams.http_thread_cnt_ = ParseOption<int>(
+            optarg, GetOptionName(&long_options[option_index]));
```
GuanLuo, line 1319: Instead of passing option name into the `Parse` function just for formatting the exception message. How about `try .. catch`ing the `while(getopt_long(...))` loop and repackage the exception with the option name on exception.

src/command_line_parser.cc
```diff
@@ -1115,6 +1161,19 @@ TritonServerParameters::BuildTritonServerOptions()
   return managed_ptr;
 }
 
+std::string
+TritonParser::GetOptionName(const struct option* long_option)
+{
+  if (long_option != nullptr) {
+#ifdef _WIN32
+    return long_option->name_;
```
GuanLuo, line 1169: Since `option` is supposed to be the mirror implementation for Windows, I think the `_` suffix can be dropped variable name to be identical as the linux implementation and `_WIN32` conditional can be droppd

GuanLuo (dismissed):
Approve with minor comment

src/command_line_parser.cc
nnshah1: The function code definitely stand to be refactored further but I think is a good step.

nnshah1 (approved):

src/command_line_parser.cc
oandreeva-nv: Refactored as suggested

qa/L0_trace/test.sh
```diff
@@ -543,6 +543,21 @@ if [ -f ./global_trace.log.0 ]; then
...
+    # Check `out of range` errors
+    rm -f ./curl.out
+    set +e
+    code=`curl -s -w %{http_code} -o ./curl.out -d'{"'${SETTING}'":"10000000000"}' localhost:8000/v2/models/simple/trace/setting`
+    set -e
+    if [ "$code" != "400" ]; then
+        cat ./curl.out
+        echo -e "\n***\n*** Test Failed\n***"
+        RET=1
+    fi
```
rmccorm4, line 558: I think @fpetrini15 recently added some [helpers](https://github.com/triton-inference-server/server/blob/639f758de1bafded665668be805dc0772f56b246/qa/L0_trace/test.sh#L648-L649) for this common code, so you can probably use that here, something like this:
```bash
for SETTING in $SETTINGS; do
  update_trace_setting "simple" -d'{"'${SETTING}'":"10000000000"}'
  assert_curl_failure "Expected error on setting large value for ${SETTING}"
done
```

GuanLuo (dismissed):

qa/L0_trace/test.sh
```diff
@@ -699,7 +707,7 @@ OTEL_COLLECTOR_LOG="./trace_collector_http_exporter.log"
 # Building the latest version of the OpenTelemetry collector.
 # Ref: https://opentelemetry.io/docs/collector/getting-started/#local
 if [ -d "$OTEL_COLLECTOR_DIR" ]; then rm -Rf $OTEL_COLLECTOR_DIR; fi
-git clone https://github.com/open-telemetry/opentelemetry-collector.git
+git clone --depth 1 --branch v0.82.0 https://github.com/open-telemetry/opentelemetry-collector.git
```
oandreeva-nv, line 710: @fpetrini15 This was the issue you were seeing. The current master branch has requirement for go to be at least 1.20 version.

fpetrini15: Added this edit to my setup and can confirm it resolved the error.

qa/L0_trace/test.sh
oandreeva-nv: Refactored

PR #6174, opened 2023-08-10
Remove extra whitespace

dyastremsky:
Remove extra whitespace that was added in a previous documentation update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,15 +38,15 @@ and corresponds to the 23.07 container release on
 
 ----
 Triton Inference Server is an open source inference serving software that
-streamlines AI inferencing. Triton enables teams to deploy any AI model from 
-multiple deep learning and machine learning frameworks, including TensorRT, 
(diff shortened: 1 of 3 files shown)


matthewkotila (approved):

matthewkotila:
Is there an intention to have a future PR for making pre-commit failures block merging? And for which branches? All? protected?

dyastremsky:
> Is there an intention to have a future PR for making Github Action failures block merging? And for which branches? All? protected?

Thanks for the approval! Great question. Based on Misha's comment [here](https://github.com/triton-inference-server/server/pull/6162#issuecomment-1673985901), it sounds like the main and rYY.MM tickets should be protected (i.e. you should not be able to merge if there is a failing test). If you see that's not the case at some point, please let me and Misha know. We do want failed checks to block merging, though any applied protections would only be valid for non-admins.

PR #6199, opened 2023-08-16
Trace Model Name Validation

fpetrini15:
Changes to validate the model name if specified in a trace request. Accompanying changes to the qa test suite.

### DIFF
--- a/qa/L0_trace/test.sh
+++ b/qa/L0_trace/test.sh
@@ -691,6 +691,132 @@ if [ `grep -o 'parent_id' bls_trace.log | wc -l` != "2" ]; then
     RET=1
 fi
 
+# Attempt to trace non-existent model
+SERVER_ARGS="--trace-file=trace_off_to_min.log --trace-level=OFF --trace-rate=1 --model-control-mode=explicit --model-repository=$MODELSDIR"
+SERVER_LOG="./inference_server_off.log"
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+
+# Explicitly load model
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/repository/models/simple/load`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Non-existent model (get)
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out localhost:8000/v2/models/does-not-exist/trace/setting`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Non-existent model (post)
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/models/does-not-exist/trace/setting -d '{"log_frequency": "1"}'`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Local model
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/models/simple/trace/setting -d '{"log_frequency": "1"}'`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out localhost:8000/v2/models/simple/trace/setting`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Local model (unload)
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/repository/models/simple/unload`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/models/simple/trace/setting -d '{"log_frequency": "1"}'`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out localhost:8000/v2/models/simple/trace/setting`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+# Local model (reload)
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/repository/models/simple/load`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/models/simple/trace/setting -d '{"log_frequency": "1"}'`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out localhost:8000/v2/models/simple/trace/setting`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
+
+kill $SERVER_PID
+wait $SERVER_PID
+
+set +e
+
 # Check opentelemetry trace exporter sends proper info.
 # A helper python script starts listening on $OTLP_PORT, where
 # OTLP exporter sends traces.
--- a/src/grpc/grpc_server.cc
+++ b/src/grpc/grpc_server.cc
@@ -1236,6 +1236,21 @@ CommonHandler::RegisterTrace()
     int32_t count;
     uint32_t log_frequency;
     std::string filepath;
+
+    if (!request.model_name().empty()) {
+      TRITONSERVER_Message* config_message = nullptr;
+      err = TRITONSERVER_ServerModelConfig(
+          tritonserver_.get(), request.model_name().c_str(),
+          -1 /* model version */, 1 /* config_version */, &config_message);
+      if (err != nullptr) {
+        err = TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_INVALID_ARG,
+            (std::string("Request for unknown model : ") + request.model_name())
+                .c_str());
+        GOTO_IF_ERR(err, earlyexit);
+      }
+    }
+
     // Update trace setting
     if (!request.settings().empty()) {
       TraceManager::NewSetting new_setting;
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -1662,6 +1662,19 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
   uint32_t log_frequency;
   std::string filepath;
 
+  if (!model_name.empty()) {
+    TRITONSERVER_Message* config_message = nullptr;
+    auto err = TRITONSERVER_ServerModelConfig(
+        server_.get(), model_name.c_str(), -1 /* model version */,
+        1 /* config_version */, &config_message);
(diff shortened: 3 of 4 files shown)


qa/L0_trace/test.sh
```diff
@@ -691,6 +691,132 @@ if [ `grep -o 'parent_id' bls_trace.log | wc -l` != "2" ]; then
     RET=1
 fi
 
+# Attempt to trace non-existent model
+SERVER_ARGS="--trace-file=trace_off_to_min.log --trace-level=OFF --trace-rate=1 --model-control-mode=explicit --model-repository=$MODELSDIR"
```
rmccorm4, line 695: 1. Use `--trace-config` as the other trace args are deprecated in favor of trace-config.
2. Don't set trace-level / trace-rate if they're not relevant to the test, just use defaults to keep server args minimal to what you need

qa/L0_trace/test.sh
```diff
@@ -691,6 +691,132 @@ if [ `grep -o 'parent_id' bls_trace.log | wc -l` != "2" ]; then
...
+fi
+
+# Non-existent model (get)
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out localhost:8000/v2/models/does-not-exist/trace/setting`
+set -e
+if [ "$code" != "400" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
```
rmccorm4, line 722: Can you make the test failure messages more descriptive? ex: "Updating trace setting on non-existent model succeeded when expected to fail" something like that

qa/L0_trace/test.sh
```diff
@@ -691,6 +691,132 @@ if [ `grep -o 'parent_id' bls_trace.log | wc -l` != "2" ]; then
...
+# Explicitly load model
+rm -f ./curl.out
+set +e
+code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/repository/models/simple/load`
+set -e
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Test Failed\n***"
+    RET=1
+fi
```
rmccorm4, line 713: I think a lot of lines in this file can be condensed with some helper functions. Something along these lines, but you can get creative with it:
```bash
# Helpers
function assert_curl_success {
  message="${1}"
  if [ "$code" != "200" ]; then
    cat ./curl.out
    echo -e "\n***\n*** ${message}\n***"
    RET=1
  fi
}

function assert_curl_failure {
  message="${1}"
  if [ "$code" != "400" ]; then
    cat ./curl.out
    echo -e "\n***\n*** ${message}\n***"
    RET=1
  fi
}

function get_trace_setting {
  model_name="${1}"
  rm -f ./curl.out
  set +e
  code=`curl -s -w %{http_code} -o ./curl.out localhost:8000/v2/models/${model_name}/trace/setting`
  set -e
}

function update_trace_setting {
  model_name="${1}"
  settings="${2}"
  rm -f ./curl.out
  set +e
  code=`curl -s -w %{http_code} -o ./curl.out -X POST localhost:8000/v2/models/${model_name}/trace/setting -d ${settings}`
  set -e
}

...

# Assert trace endpoints return errors for non-existent models
GOOD_MODEL="simple"
BAD_MODEL="does-not-exist"
TRACE_SETTINGS='{"log_frequency": "1"}'

get_trace_setting ${BAD_MODEL}
assert_curl_failure "Expected non-existent model to fail"

update_trace_setting "${BAD_MODEL}" "${TRACE_SETTINGS}" 
assert_curl_failure "Expected non-existent model to fail"

# Assert trace endpoints succeed for existing model
load_model ${GOOD_MODEL}
get_trace_setting ${GOOD_MODEL}
assert_curl_success "Expected existing model to succeed"

update_trace_setting "$GOOD_MODEL}" "${TRACE_SETTINGS}" 
assert_curl_success "Expected existing model to succeed"

unload_model ${GOOD_MODEL}
...
...
```

Can be applicable to some existing parts of the test you didn't touch as well

src/grpc/grpc_server.cc
```diff
@@ -1236,6 +1236,21 @@ CommonHandler::RegisterTrace()
     int32_t count;
     uint32_t log_frequency;
     std::string filepath;
+
+    if (!request.model_name().empty()) {
```
oandreeva-nv, line 1240: I wounder if it would make sense to put this check and `TRITONSERVER_ServerModelConfig` into TraceManager::UpdateTraceSetting ([here](https://github.com/triton-inference-server/server/blob/8a823b2381f12f97219404543ad55b1437a62958/src/tracer.cc#L92) ). This way we only need to change `trace.cc` (cc @rmccorm4 )

qa/L0_trace/test.sh
oandreeva-nv: example of trace config is in the test closer to the end: https://github.com/triton-inference-server/server/pull/6199/files#diff-27e9ca4cd8f292f0e5a94bbabb243e40fe0d47cd55164a0cb49ae6e093c9fbe7R840

Since some of these tests were before we switched to `trace-config`, not all of them were adjusted

src/grpc/grpc_server.cc
fpetrini15: I was originally trying to put the changes there, however, I think it would require exposing `TRITONSERVER_ServerModelConfig` to the trace manager which would violate the hierarchy. Also, by the time `TraceManager::UpdateTraceSetting` is called, the server has already spent time parsing the input request. My ultimate decision to place the changes in the {endpoint}.cc files based on the structure of our `HandleModelConfig` functions, which, in the original ticket, was able to correctly identify the changes to the model repository.

oandreeva-nv: > which would violate the hierarchy.

Could you please clarify? I believe we already expose `TRITONSERVER_` API to our tracer: https://github.com/triton-inference-server/server/blob/8a823b2381f12f97219404543ad55b1437a62958/src/tracer.h#L50

fpetrini15: I was referring to the function `TRITONSERVER_ServerModelConfig`, itself which requires a `TRITONSERVER_Server` pointer parameter that the endpoints possess but `TraceManager` does not. The endpoints do not currently expose this pointer as far as I am aware.

qa/L0_trace/test.sh
fpetrini15: Added more descriptive messages and line number tracking.

src/grpc/grpc_server.cc
oandreeva-nv: Ah, gotcha. Thank you for clarification.

src/grpc/grpc_server.cc
```diff
@@ -1236,6 +1236,21 @@ CommonHandler::RegisterTrace()
...
+      TRITONSERVER_Message* config_message = nullptr;
+      err = TRITONSERVER_ServerModelConfig(
+          tritonserver_.get(), request.model_name().c_str(),
+          -1 /* model version */, 1 /* config_version */, &config_message);
```
rmccorm4, line 1244: Since we don't actually need the model config here, I think [TRITONSERVER_ServerModelIsReady](https://github.com/triton-inference-server/core/blob/e456d2ad22778776dfbea631c5d262c42567b6f1/src/tritonserver.cc#L2553) would be a more lightweight proxy for model existing.

qa/L0_trace/test.sh
```diff
@@ -691,6 +623,70 @@ if [ `grep -o 'parent_id' bls_trace.log | wc -l` != "2" ]; then
     RET=1
 fi
 
+# Attempt to trace non-existent model
+SERVER_ARGS="--model-control-mode=explicit --model-repository=$MODELSDIR"
+SERVER_LOG="./inference_server_off.log"
```
rmccorm4, line 628:
```suggestion
SERVER_LOG="./inference_server_nonexistent_model.log"
```

src/grpc/grpc_server.cc
fpetrini15: Interestingly, when I switch to using this function, no error is returned from the core components: 
`bool ready = false;`
`auto err = TRITONSERVER_ServerModelIsReady(server_.get(), model_name.c_str(), -1 /* model version */, &ready);`
The only indication that something went wrong is that the bool parameter `ready` remains false when the model does not exist / flips to true if a model does exist. However, I don't believe this is the desired effect considering the [HandleModelReady](https://github.com/triton-inference-server/server/blob/328614e1bd85997ff8559b934a33eaa636f46b1b/src/http_server.cc#L1486-L1487) function attempts to pass on an error status from `TRITONSERVER_ServerModelIsReady`. I am thinking that this may be an issue from an unhandled exception [here](https://github.com/triton-inference-server/core/blob/e456d2ad22778776dfbea631c5d262c42567b6f1/src/server.cc#L461C2-L470C26) in core. Seems like an error getting the model will have no effect and the function will still return a `Status::Success`?

rmccorm4: I don't know the original context for why this doesn't return an error if server ready but model not ready. However, you can just check the `ready` arg instead after calling.

As for the model ready handler, it checks [both](https://github.com/triton-inference-server/server/blob/328614e1bd85997ff8559b934a33eaa636f46b1b/src/http_server.cc#L1490-L1491): `(ready && (err == nullptr))`

fpetrini15: Yes, I can pivot on the ready arg, but it seems strange that this code trace would encounter an error that essentially becomes overwritten, especially when `HandleModelReady` anticipates the possibility of an error. I guess keeping the ready flag false becomes the manifestation of the error...

src/http_server.cc
```diff
@@ -1661,6 +1661,17 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
...
+  if (!model_name.empty()) {
+    bool ready = false;
+    auto err = TRITONSERVER_ServerModelIsReady(
+        server_.get(), model_name.c_str(), -1 /* model version */, &ready);
+    if (!ready || err != nullptr) {
+      HTTP_RESPOND_IF_ERR(
+          req, TRITONSERVER_ErrorNew(
+                   TRITONSERVER_ERROR_INVALID_ARG,
+                   ("Request for unknown model : " + model_name).c_str()));
+    }
```
rmccorm4, line 1673: I believe this is leaking `err` without deleting it. You can probably do something like:
```cpp
bool ready = false;
HTTP_RESPOND_IF_ERR(req, TRITONSERVER_ServerModelIsReady(..., &ready));
if (!ready) {
      HTTP_RESPOND_IF_ERR(
          req, TRITONSERVER_ErrorNew(
                   ("Request for unknown model : " + model_name).c_str()));
}
```
Similarly for GRPC server, don't want to overwrite a non-null error without deleting it.

fpetrini15: Good catch, thanks! By that same token, would [this](https://github.com/triton-inference-server/server/blob/8a823b2381f12f97219404543ad55b1437a62958/src/http_server.cc#L1283-L1293C43) also be a memory leak since this newly declared `err` variable could potentially leave scope without being deleted?

rmccorm4: It gets cleaned up [here](https://github.com/triton-inference-server/server/blob/8a823b2381f12f97219404543ad55b1437a62958/src/http_server.cc#L1300)

rmccorm4 (dismissed):

rmccorm4:
LGTM assuming CI passes

src/http_server.cc
fpetrini15: I think that cleans up [this](https://github.com/triton-inference-server/server/blob/8a823b2381f12f97219404543ad55b1437a62958/src/http_server.cc#L1240C3-L1240C37) variable `err` , declared earlier. The variable of the same name that's declared on 1283 looks to go out of scope unhandled on line 1295, no?

rmccorm4: Ah you're right, 1283 appears to shadow the outer `err` var, this is probably leaking as well. As discussed offline, you should quickly verify this with valgrind, and probably replace `auto err = ...` on 1283 with just `err = ...` to use the outer scope variable.

fpetrini15: Confirmed with valgrind that it could indeed cause a memory leak. Adding the fix to the PR.

qa/L0_trace/test.sh
```diff
@@ -80,8 +80,77 @@ cp -r $DATADIR/$MODELBASE $MODELSDIR/simple && \
 
 RET=0
 
+# Helpers =======================================
+function assert_curl_success {
+  message="${1}"
+  if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** ${message} : line ${BASH_LINENO}\n***"
```
rmccorm4, line 88: I think these `BASH_LINENO` will statically call the line number of these functions (ex: 88 or 97), rather than the line of the caller. And I believe it's unexpected complicated to correctly pass line numbers around in bash. So I'd probably just take these out.

src/http_server.cc
```diff
@@ -1280,7 +1280,7 @@ HTTPAPIServer::HandleRepositoryIndex(
     }
 
     TRITONSERVER_Message* message = nullptr;
-    auto err = TRITONSERVER_ServerModelIndex(server_.get(), flags, &message);
+    err = TRITONSERVER_ServerModelIndex(server_.get(), flags, &message);
```
rmccorm4, line 1283: Nice catch! :tada:

rmccorm4 (commented):
LGTM other than last BASH_LINENO comment

qa/L0_trace/test.sh
fpetrini15: When I was making adjustments to the script I tested the error messages (accidentally by making mistakes in the logic) and they worked as expected. `BASH_LINENO` should not statically call the line number it's invoked on, I believe `LINENO` would do that. Instead, according to the documentation `BASH_LINENO` should evaluate to the line number its parent function was called on, which is what I observed. If we still prefer to remove `BASH_LINENO` altogether, I can incorporate the above suggestions and pass additional specification parameters.

rmccorm4: > BASH_LINENO should evaluate to the line number its parent function was called on, which is what I observed. 
Oh nice, TIL. LGTM locally too. Thanks for pushing back :slightly_smiling_face:

rmccorm4 (approved):

PR #6203, opened 2023-08-17
Add utility functions for array manipulation

dyastremsky:
Add utility functions for array manipulation. These are used for updated internal tests and can be re-used for other QA tests.


mc-nv:
Is it applied to any of our tests, I see you added two new functions but can't find any usage of them.

dyastremsky:
> Is it applied to any of our tests, I see you added two new functions but can't find any usage of them.

Right now, these are only used for our internal GitLab tests (updated to use these functions).

mc-nv (approved):

PR #6228, opened 2023-08-24
Test Python execute() to return Triton error code

kthui:
Related PR: https://github.com/triton-inference-server/python_backend/pull/292

This new test ensures the optionally supplied Python `execute()` error code can be picked up on the calling gRPC client.

### DIFF
--- a/qa/L0_backend_python/lifecycle/lifecycle_test.py
+++ b/qa/L0_backend_python/lifecycle/lifecycle_test.py
@@ -58,6 +58,27 @@ class LifecycleTest(tu.TestResultCollector):
     def setUp(self):
         self._shm_leak_detector = shm_util.ShmLeakDetector()
 
+    def test_error_code(self):
+        model_name = "error_code"
+        shape = [1, 1]
+        error = "TRITONSERVER_ERROR_INVALID_ARG"
+        expected_grpc_error = "StatusCode.INVALID_ARGUMENT"
+        with self._shm_leak_detector.Probe() as shm_probe:
+            with grpcclient.InferenceServerClient("localhost:8001") as client:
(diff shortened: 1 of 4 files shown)


qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -58,6 +58,27 @@ class LifecycleTest(tu.TestResultCollector):
     def setUp(self):
         self._shm_leak_detector = shm_util.ShmLeakDetector()
 
+    def test_error_code(self):
+        model_name = "error_code"
+        shape = [1, 1]
+        error = "TRITONSERVER_ERROR_INVALID_ARG"
```
dyastremsky, line 64: Great testing addition!

I found some of this to be confusing until I reviewed the PR closer. Is there a way to simplify the model and naming? For example, the model doesn't need all the error codes if we're only going to test one, unless we expect to reuse the model for other errors soon. I think it'd be good if this can be cleaned up to make it clear that the model takes an error string input and returns a message with the associated error code followed by the error string input. That might come down to updating the variable naming and/or message.

kthui: How about we test all the codes, and also default code on the same test? [Add all supported error codes into test](https://github.com/triton-inference-server/server/pull/6228/commits/ab83a3db5f8cbe5d73e72552fe01b271957546a7)

dyastremsky: Fantastic, this looks great!

dyastremsky (approved):

rmccorm4 (approved):

Tabrizian (approved):

rmccorm4 (approved):

PR #6231, opened 2023-08-25
Add testing for retrieving shape and datatype in backend API

fpetrini15:
Updated BLS tests for new backend APIs. Hard-coded reference to the "fpetrini-dim-shape-apis" tag will be removed upon approval and merge.


fpetrini15:
Important changes for bls.cc start on line 292 where we validate that we can retrieve the set values from the backend APIs.

qa/L0_backend_bls/bls.cc
```diff
@@ -0,0 +1,349 @@
+// Copyright 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
Tabrizian, line 1:
```suggestion
// Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

qa/L0_backend_bls/test.sh
```diff
@@ -101,6 +101,44 @@ set -e
...
+    echo -e "\n***\n*** bls_test.py FAILED. \n***"
+    cat $CLIENT_LOG
+    cat $SERVER_LOG
+    RET=1
+fi
+set -e
+
+kill $SERVER_PID
+wait $SERVER_PID
+
```
Tabrizian, line 141: Need to update the copyright.

Tabrizian (commented):
Looks good in general! Few copyright updates.

qa/L0_backend_bls/bls.cc
fpetrini15: I'll keep this in mind for future updates.

Tabrizian (dismissed):

qa/L0_backend_bls/bls.cc
```diff
@@ -0,0 +1,349 @@
...
+                      output, &output_buffer, output_byte_size_server,
+                      &output_memory_type_server, &output_memory_id_server));
+    if (output_memory_type_server == TRITONSERVER_MEMORY_GPU) {
+      RESPOND_AND_SET_NULL_IF_ERROR(
+          response, TRITONSERVER_ErrorNew(
+                        TRITONSERVER_ERROR_INTERNAL,
+                        "failed to create output buffer in CPU memory"));
+    }
+    // Validate new backen APIs, ensuring that the retrieved values match those
+    // retrieved by the server
```
rmccorm4, line 293: Instead of adding a majority copy of this `bls.cc` file into the server, can we just augment the backend example and mention it's just for validation/demonstration?

Something along the lines of this:

```cpp
// NOTE: This function is just to demonstrate TRITONBACKEND equivalents of TRITONSERVER output
// getters, and verify their correctness. It is not necessary for this backend example.
TRITONSERVER_Error*
ValidateBackendGetters() {
  /* your existing additions, but RETURN_IF_ERROR instead of RESPOND_IF_ERROR everywhere */
}


void
BLSExecutor::ConstructFinalResponse() {
  ...
  
  // Optional: demonstrate backend output getter correctness, not necessary for this example.
  RESPOND_AND_SET_NULL_IF_ERROR(ValidateBackendGetters(...));
  
  ...
}
```
This way you're just adding the 1 new helper function to the example, keeping the code in one place in the example, and not duplicating it in server.

Also, we can add a comment that the helpers can be used for optimizations (original intent of the feature I believe?)

rmccorm4: @fpetrini15 @Tabrizian what do you think?

fpetrini15: @Tabrizian since this is an example, it might be helpful to add the comment as rmccorm4 suggests, but it might be better to demonstrate the optimization. How feasible/is this the appropriate model to demonstrate the possibility of the response optimization?

Tabrizian: I figured out an easier way to unit test the backend APIs without having to create a separate backend. Please see the PR here: https://github.com/triton-inference-server/core/pull/249/files#diff-50be12abfe003477dda43701f231fad704847749ad914a52868dbf8afc508fb9

fpetrini15: Added unit testing for the backend APIs: https://github.com/triton-inference-server/core/pull/247

qa/L0_backend_output_details/test.sh
```diff
@@ -0,0 +1,74 @@
...
+    cp  ../python_models/add_sub/model.py $MODELSDIR/add_sub/1 && \
+
+source ../common/util.sh
+
+RET=0
+
+TEST_LOG="./backend_output_detail_test.log"
+TEST_EXEC=./backend_output_detail_test
+
+rm -fr *.log
```
rmccorm4, line 55: I think logs are removed already above

qa/L0_backend_output_details/test.sh
```diff
@@ -0,0 +1,74 @@
...
+    RET=1
+fi
+set -e
+
+if [ $RET -eq 0 ]; then
+    echo -e "\n***\n*** Test Passed\n***"
+else
+    cat $TEST_LOG
+    cat $CLIENT_LOG
+    cat $SERVER_LOG
```
rmccorm4, line 70: I don't think these client/server logs exist, should be encapsulated in test log

Tabrizian (approved):
Let's change the PR title to "Add testing for retrieving shape and datatype in backend API"

rmccorm4 (approved):

PR #6233, opened 2023-08-26
Add tests/docs for pending request metric

rmccorm4:
Tests/docs for new per-model pending request count / queue size metric: https://github.com/triton-inference-server/core/pull/246


qa/L0_metrics/metrics_queue_size_test.py
```diff
@@ -0,0 +1,265 @@
...
+        self.ensemble_inputs = [
+            tritonclient.http.InferInput(
+                "ENSEMBLE_INPUT0", shape, dtype
+            ).set_data_from_numpy(input_np)
+        ]
+
+        # Verify values for filling request queues
+        self.num_requests = 10
+        self.concurrency = 10
+        self.assertGreaterEqual(self.concurrency, self.num_requests)
```
dyastremsky, line 79: Can you clarify why this is necessary? Is this an extra check that the above constants are not changed in a way that would fail this check (like your comment below states)?

dyastremsky (commented):
Very clean code. Appreciate the addition of the "tritonserver already running" check and use of concurrency to speed up this test.

Commented one clarifying question.

qa/L0_metrics/metrics_queue_size_test.py
rmccorm4: This is mostly just to help requests get into server ASAP for filling batches. In hindsight I don't think it's a hard requirement to make this test consistent. 

The assert's intent is to catch changes to the constants, and raise a question to the developer when changing it.

docs/user_guide/metrics.md
```diff
@@ -97,6 +97,71 @@ Count*. The count metrics are illustrated by the following examples:
...
+than a single queue. Some of the most common will be called out below:
+- Default Scheduler backlogs any requests not currently executing.
+  - Assuming 1 available model instance with the default scheduler settings,
+    and 10 requests are sent in rapid succession.
+  - The 1st request should be picked up for
+    execution immediately, and the remaining 9 requests should be considered
+    pending for this model, until the 1st request is finished. Afterwards, the
+    next request should be picked up and the pending count should be decremented
+    to 8, and so on until all requests are finished and the pending count is 0.
+- Dynamic Batcher queue for dynamically  creating batches from requests
```
fpetrini15, line 129: Nit: Some of the bullet headers are punctuated, some are not. Also I think there's an extra space between "dynamically" and "creating".

qa/L0_metrics/ensemble_delay/config.pbtxt
```diff
@@ -0,0 +1,67 @@
...
+  }
+]
+
+ensemble_scheduling
+{
+  step [
+    {
+      model_name: "dynamic_composing"
+      model_version: -1
+      input_map { key: "INPUT0", value: "ENSEMBLE_INPUT0"}
```
fpetrini15, line 57: Nit: Add a space between the end of the input name and the bracket to keep the file consistent.

fpetrini15 (commented):
Strong changes, only extremely minor nits from me.

qa/L0_metrics/metrics_queue_size_test.py
rmccorm4: I modified the comment to be a bit more clear on this.

dyastremsky (approved):

PR #6257, opened 2023-09-01
Refactor StreamInferHandler for better readability

rmccorm4:
This PR does two things:
1. Refactor the handling of every `step` in `StreamInferHandler::Process` from one large ~400 line function into helper functions for each state.
2. Replace the "double negative" logic of `if not not_finished` into just `if finished` in Common/InferHandler/StreamInferHandler `Process()` functions. I found the double negative logic confusing that the function did:
```cpp
bool Process(...) {
  ...
  finished = true; 
  // (not finished)
  return !finished;
}
``` 
and then the caller checks the negation of that:
```cpp
// if (not (not finished) )
if (  !Process(...) )
```
So instead if's more direct to be:
```cpp
bool Process(...) {
  finished = true;
  return finished;
}

if ( Process(...) {
  ...
}
```

### DIFF
--- a/src/grpc/stream_infer_handler.cc
+++ b/src/grpc/stream_infer_handler.cc
@@ -130,6 +130,241 @@ ModelStreamInferHandler::StartNewRequest()
 }
 
 bool
+ModelStreamInferHandler::RequestStartStep(
+    InferHandler::State* state, bool rpc_ok)
+{
+  // A new stream connection... If RPC failed on a new request then
+  // the server is shutting down and so we should do nothing.
+  if (!rpc_ok) {
+    state->step_ = Steps::FINISH;
+    return true;
+  }
+
+  // Start a new request to replace this one...
+  StartNewRequest();
+
+  if (ExecutePrecondition(state)) {
+    // Since this is the start of a connection, 'state' hasn't been
+    // used yet so use it to read a request off the connection.
+    state->context_->step_ = Steps::READ;
+    state->step_ = Steps::READ;
+    state->context_->responder_->Read(&state->request_, state);
+  } else {
+    // Precondition is not satisfied, cancel the stream
+    state->context_->step_ = Steps::COMPLETE;
+    state->step_ = Steps::COMPLETE;
+    ::grpc::Status status = ::grpc::Status(
+        ::grpc::StatusCode::UNAVAILABLE,
+        std::string("This protocol is restricted, expecting header '") +
+            restricted_kv_.first + "'");
+    state->context_->responder_->Finish(status, state);
+    // Request finished due to error.
+    return true;
+  }
+
+  // Not finished with a request on the start step unless an error occurs above.
+  return false;
+}
+
+bool
+ModelStreamInferHandler::RequestReadStep(
+    InferHandler::State* state, bool rpc_ok)
+{
+  bool finished = false;
+  TRITONSERVER_Error* err = nullptr;
+  const inference::ModelInferRequest& request = state->request_;
+#ifdef TRITON_ENABLE_TRACING
+  state->trace_timestamps_.emplace_back(
+      std::make_pair("GRPC_WAITREAD_END", TraceManager::CaptureTimestamp()));
+#endif  // TRITON_ENABLE_TRACING
+
+  // If done reading and no in-flight requests then can finish the
+  // entire stream. Otherwise just finish this state.
+  if (!rpc_ok) {
+    state->context_->step_ = Steps::WRITEREADY;
+    if (state->context_->IsRequestsCompleted()) {
+      state->context_->step_ = Steps::COMPLETE;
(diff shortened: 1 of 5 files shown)


src/grpc/stream_infer_handler.cc
```diff
@@ -130,362 +130,424 @@ ModelStreamInferHandler::StartNewRequest()
...
+    } else {
+      state->context_->PutTaskBackToQueue(state);
     }
   }
 
-  return !finished;
+  return finished;
+}
+
+bool
```
tanmayv25, line 526: Can you move the Process function above before L111? It would be helpful to look at the state machine right at the beginning on the file?

src/grpc/stream_infer_handler.h
```diff
@@ -114,6 +115,20 @@ class ModelStreamInferHandler
       void* userp);
   bool Finish(State* state);
 
+  // Step helpers
+  bool RequestStartStep(State* state, bool rpc_ok);
+  bool RequestReadStep(State* state, bool rpc_ok);
+  bool RequestCompleteStep(State* state);
+  bool RequestWrittenStep(State* state, bool rpc_ok);
```
tanmayv25, line 122: RequestWrittenStep and RequestWrittenStepNonDecoupled looks confusing.
I would recommend pulling the state->is_decoupled_ out into the Process function. It might be easier to look at which function is being called when in the Process function.

src/grpc/stream_infer_handler.cc
```diff
@@ -130,362 +130,424 @@ ModelStreamInferHandler::StartNewRequest()
...
+                 << ", context " << state->context_->unique_id_ << ", "
+                 << state->unique_id_ << " step " << state->step_;
+
+  // We need an explicit finish indicator. Can't use 'state->step_'
+  // because we launch an async thread that could update 'state's
+  // step_ to be FINISH before this thread exits this function.
+  bool finished = false;
+
+  if (state->step_ == Steps::START) {
+    finished = RequestStartStep(state, rpc_ok);
```
tanmayv25, line 539: Can we pull the step transition from within these helpers to this Process function? It would help in giving a complete picture of the state machine being implemented here.
If not, then may be add a comment describing to which state the state can transition to in these helper functions.

src/grpc/stream_infer_handler.h
```diff
@@ -114,6 +115,20 @@ class ModelStreamInferHandler
       void* userp);
   bool Finish(State* state);
 
+  // Step helpers
+  bool RequestStartStep(State* state, bool rpc_ok);
```
tanmayv25, line 119: StreamStartStep?

src/grpc/stream_infer_handler.cc
rmccorm4: This would be nice, but there's a lot of branching in each function where `if (this) state = state1; else state = state2;` It would be non-trivial effort to extract this out I think.

src/grpc/stream_infer_handler.cc
```diff
@@ -130,362 +130,424 @@ ModelStreamInferHandler::StartNewRequest()
...
+  // and writes to complete.
+  if (!state->context_->PopCompletedResponse(state)) {
+    LOG_ERROR << "Unexpected response for " << Name() << ", rpc_ok=" << rpc_ok
+              << ", context " << state->context_->unique_id_ << ", "
+              << state->unique_id_ << " step " << state->step_;
+    state->context_->finish_ok_ = false;
+  }
+
+  // Write the next response if it is ready...
+  state->context_->WriteResponseIfReady(nullptr);
```
rmccorm4, line 468: @tanmayv25 I don't understand this part. Do you know why we try to Write the next response (not the current' state's response) here in the WRITTEN state for a non-decoupled request? 

For the non-decoupled case, isn't every response's write handled when calling the [response callback](https://github.com/triton-inference-server/server/blob/6930d009b05a64a16eac3c751003c2329c429c6b/src/grpc/stream_infer_handler.cc#L633) or error handling for a given request?

src/grpc/stream_infer_handler.cc
```diff
@@ -130,362 +130,424 @@ ModelStreamInferHandler::StartNewRequest()
...
+  // Not finished with a request on the read step unless an error occurs above.
+  return false;
+}
+
+bool
+ModelStreamInferHandler::RequestCompleteStep(InferHandler::State* state)
+{
+  state->step_ = Steps::FINISH;
+  // Finished with the request.
+  return true;
```
rmccorm4, line 364: Also @tanmayv25 , I don't really understand the point of the COMPLETE state/step. Why don't we directly transition to FINISH state anywhere we would transition to COMPLETE instead? Seems like an unnecessary intermediate step.

rmccorm4:
Closing due to lots of changes around request cancellation. May come back to this later.

PR #6275, opened 2023-09-07
Fix L0_java_memory_growth

nv-kmcgill53:
Compilation inside the test is failing due to `cmake` and `rapidjson-dev` not being installed. 

Adding these pre-requisites doesn't solve the runtime issue. Once a solution is found, it will need to be implemented before this can be merged.


Tabrizian (approved):
I wonder how the test worked before without these packages.

nv-kmcgill53:
> I wonder how the test worked before without these packages.

Last pass was on `2023-06-29`

jbkyang-nvi:
`cmake` and `rapidjson` shouldn't be needed for this test. I'll take a look into this

PR #6282, opened 2023-09-08
Update the request re-use example

Tabrizian:


src/simple.cc
```diff
@@ -902,6 +907,20 @@ main(int argc, char** argv)
             InferResponseComplete, reinterpret_cast<void*>(p)),
         "setting response callback");
 
+    // When re-using the requests, we need to make sure that the request
+    // complete callback has been called for the previous request before
```
nnshah1, line 911:
```suggestion
    // release callback has been called for the previous request before
```

src/simple.cc
```diff
@@ -724,10 +726,14 @@ main(int argc, char** argv)
       TRITONSERVER_InferenceRequestSetId(irequest, "my_request_id"),
       "setting ID for the request");
 
+  std::unique_ptr<std::promise<void>> barrier =
+      std::make_unique<std::promise<void>>();
   FAIL_IF_ERR(
       TRITONSERVER_InferenceRequestSetReleaseCallback(
-          irequest, InferRequestComplete, nullptr /* request_release_userp */),
+          irequest, InferRequestComplete,
```
nnshah1, line 733: maybe rename to InferRequestReleased ?

Tabrizian:
thanks for the quick review. I applied the changes in https://github.com/triton-inference-server/server/pull/6283

PR #6283, opened 2023-09-08
Update the request re-use example

Tabrizian:
https://github.com/triton-inference-server/core/pull/258

### DIFF
--- a/src/simple.cc
+++ b/src/simple.cc
@@ -1,4 +1,4 @@
-// Copyright 2020-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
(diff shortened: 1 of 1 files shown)


src/simple.cc
```diff
@@ -902,6 +906,20 @@ main(int argc, char** argv)
             InferResponseComplete, reinterpret_cast<void*>(p)),
         "setting response callback");
 
+    // When re-using the requests, we need to make sure that the request
```
nnshah1, line 909:
```suggestion
    // We need to make sure that the previous request was released before reusing it.
```

nnshah1 (approved):
slight tweak to comment for readability

rmccorm4 (dismissed):

src/simple.cc
Tabrizian: fixed.

rmccorm4 (approved):
:rocket:

PR #6294, opened 2023-09-11
Exclude extra TRT version-compatible models from tests

dyastremsky:
Some QA tests do not need the version-compatible models. Including them will result in the tests failing if the server is started with version compatibility disabled. Remove these models from those tests.

### DIFF
--- a/qa/L0_infer_zero/test.sh
+++ b/qa/L0_infer_zero/test.sh
@@ -57,7 +57,7 @@ cp -r /data/inferenceserver/${REPO_VERSION}/qa_identity_model_repository/* model
 
 # Remove version-compatible TensorRT models, as they require version-compatibility
 # mode to be turned on when starting the server.
-rm -r models/plan_compatible*
+rm -rf models/plan_compatible*
 
 create_nop_version_dir `pwd`/models
 
--- a/qa/L0_infer_zero/test.sh
+++ b/qa/L0_infer_zero/test.sh
@@ -55,6 +55,10 @@ rm -fr models && mkdir models
 cp -r /data/inferenceserver/${REPO_VERSION}/qa_identity_model_repository/* models/. && \
     cp -r /data/inferenceserver/${REPO_VERSION}/qa_ensemble_model_repository/qa_identity_model_repository/* models/.
 
+# Remove version-compatible TensorRT models, as they require version-compatibility
+# mode to be turned on when starting the server.
+rm -r models/plan_compatible*
+
 create_nop_version_dir `pwd`/models
 
 RET=0
--- a/qa/L0_nullchar_string/test.sh
+++ b/qa/L0_nullchar_string/test.sh
@@ -40,16 +40,22 @@ fi
 
 export CUDA_VISIBLE_DEVICES=0
 
-DATADIR=/data/inferenceserver/${REPO_VERSION}/qa_identity_model_repository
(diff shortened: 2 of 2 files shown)


qa/L0_infer_zero/test.sh
```diff
@@ -55,6 +55,10 @@ rm -fr models && mkdir models
 cp -r /data/inferenceserver/${REPO_VERSION}/qa_identity_model_repository/* models/. && \
     cp -r /data/inferenceserver/${REPO_VERSION}/qa_ensemble_model_repository/qa_identity_model_repository/* models/.
 
+# Remove version-compatible TensorRT models, as they require version-compatibility
+# mode to be turned on when starting the server.
+rm -r models/plan_compatible*
```
rmccorm4, line 60: `-f` to avoid failure in case it's not there for some reason.
```suggestion
rm -rf models/plan_compatible*
```

rmccorm4, line 60: How difficult would it be for compatible plans to be in a separate folder instead, and explicitly pulled in for the compatibility test? ex:
```
qa_trt_compatible_model_repository
```

rmccorm4: I think it's more prone to unexpected errors when a model requiring server-specific flags `--backend-config ... compatible=true` is included with the rest of the common QA models.

qa/L0_nullchar_string/test.sh
```diff
@@ -40,16 +40,22 @@ fi
...
+SERVER_ARGS="--model-repository=models"
 SERVER_LOG="./inference_server.log"
 source ../common/util.sh
 
-rm -f $CLIENT_LOG $SERVER_LOG
+rm -f $CLIENT_LOG $SERVER_LOG models
+
+mkdir -p models
+for MODEL in $MODELS; do
+    cp -r $DATADIR/$MODEL models/.
```
rmccorm4, line 57: nice cleanup :rocket:

qa/L0_infer_zero/test.sh
dyastremsky: We could. I didn't create one because there is only one version-compatible model, so I didn't love the idea of creating another directory just for this one model. And I couldn't find another directory that seemed to make more sense for this model.

Most of the tests seem to just copy the identity models they need rather than the whole repository. If you think there's a clean, worthwhile way to separate out the compatible model, let me know. If you think it'd be the best option, I could just create that separate directory, since we have ones for other things like the trt_plugin one, but I'd like to ultimately find a way to ensure that high-level model directory doesn't grow in an unorderly way.

rmccorm4: I think this is fine for now, but if it comes up again I'd probably move it to a separate folder.

rmccorm4: Regarding high level directory growth, we can always start grouping things like `trt_model_repo` containing `trt_plugin`, `trt_version_compatible`, etc.

rmccorm4 (approved):

qa/L0_infer_zero/test.sh
dyastremsky: That's a good idea! I'll try to keep an eye out for this, we might want to start shifting towards that approach as we create more backend-specific repos. Thanks, Ryan.

PR #6297, opened 2023-09-12
Refactor HTTP server

rmccorm4:
1. Clean up common parsing components of request JSON into helper functions. 
    - This will separate some logic between parsing inputs provided in current format `ParseJsonTritonIO --> {inputs: {name, datatype, shape}, ...}` and accepting json inputs in other formats in future such as `{prompt, parameters}`.
    - Breaks up a ~400 line `EVBufferToInput` function into more manageable chunks
2. Return error messages in response body more consistently for most routes. Improved CLI/`curl` UX, similar to FastAPI and vLLM API server experience.
    - Routes without any handler (not found) will now return **404 instead of 400** at top-level Handler

### Before

Query triton routes:
```bash
rmccormick@ced35d0-lcedt:~$ curl -w "\n" -X POST localhost:8000/v2

rmccormick@ced35d0-lcedt:~$ curl -w "\n" -X POST localhost:8000/v2/does-not-exist

```

Responses are blank, so you have to remember to grab the error code:
```bash
rmccormick@ced35d0-lcedt:~$ curl -w "%{http_code}\n" -X POST localhost:8000/v2
405
rmccormick@ced35d0-lcedt:~$ curl -w "%{http_code}\n" -X POST localhost:8000/v2/does-not-exist
400
```

But then you have to go look up what 405 means, and so on

### After

Query triton routes:

```bash
rmccormick@ced35d0-lcedt:~$ curl -w "\n" -X POST localhost:8000/v2
{"error":"Method Not Allowed"}
rmccormick@ced35d0-lcedt:~$ curl -w "\n" -X POST localhost:8000/v2/does-not-exist
{"error":"Not Found"}
```

Now it's more immediately obvious what's wrong with our queries.

### FastAPI/vLLM for comparison
```bash
rmccormick@ced35d0-lcedt:~$ curl -w "\n" -X POST localhost:8080/
{"detail":"Method Not Allowed"}
rmccormick@ced35d0-lcedt:~$ curl -w "\n" -X POST localhost:8080/does-not-exist
{"detail":"Not Found"}
```

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -2155,11 +2155,8 @@ HTTPAPIServer::HandleSystemSharedMemory(
     }
   }
 
-  if (err == nullptr) {
-    evhtp_send_reply(req, EVHTP_RES_OK);
-  } else {
-    HTTP_RESPOND_IF_ERR(req, err);
-  }
+  HTTP_RESPOND_IF_ERR(req, err);
+  evhtp_send_reply(req, EVHTP_RES_OK);
 }
 
 void
@@ -2277,10 +2274,7 @@ HTTPAPIServer::HandleCudaSharedMemory(
     }
   }
 
-  if (err != nullptr) {
-    HTTP_RESPOND_IF_ERR(req, err);
-  }
-
+  HTTP_RESPOND_IF_ERR(req, err);
   evhtp_send_reply(req, EVHTP_RES_OK);
 }
 
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -75,6 +75,160 @@ namespace triton { namespace server {
   } while (false)
 
 namespace {
+void
+EVBufferAddErrorJsonInternal(evbuffer* buffer, const char* message)
+{
+  triton::common::TritonJson::Value response(
+      triton::common::TritonJson::ValueType::OBJECT);
+  response.AddStringRef("error", message, strlen(message));
+
+  triton::common::TritonJson::WriteBuffer buffer_json;
+  response.Write(&buffer_json);
+
+  evbuffer_add(buffer, buffer_json.Base(), buffer_json.Size());
+}
+
+void
+EVBufferAddErrorJson(evbuffer* buffer, const std::string& error_str)
+{
+  const char* message = error_str.c_str();
+  EVBufferAddErrorJsonInternal(buffer, message);
+}
+
+void
+EVBufferAddErrorJson(evbuffer* buffer, TRITONSERVER_Error* err)
+{
+  const char* message = TRITONSERVER_ErrorMessage(err);
+  EVBufferAddErrorJsonInternal(buffer, message);
+}
+
+
+}  // namespace
+
+TRITONSERVER_Error*
+HTTPServer::Start()
+{
+  if (!worker_.joinable()) {
+    evbase_ = event_base_new();
+    htp_ = evhtp_new(evbase_, NULL);
+    evhtp_enable_flag(htp_, EVHTP_FLAG_ENABLE_NODELAY);
+    if (reuse_port_) {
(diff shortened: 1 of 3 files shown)


src/http_server.cc
```diff
@@ -67,6 +67,44 @@ namespace triton { namespace server {
...
   } while (false)
 
+#define HTTP_RESPOND_WITH_CODE(REQ, CODE, MSG)    \
+  do {                                            \
+    EVBufferAddErrorJson((REQ)->buffer_out, MSG); \
+    evhtp_send_reply((REQ), CODE);                \
+    return;                                       \
+  } while (false)
+
+namespace {
```
rmccorm4, line 77: I will swap this block to be with the rest of the helpers and move the HTTPServer Start/Stop and HTTPMetrics server to be down after the helpers with the rest of the server code in a quick follow-up PR. 

Doing so in this PR introduces a lot of noise to the amount of actual changes.

src/http_server.cc
```diff
@@ -142,8 +180,7 @@ HTTPMetricsServer::Handle(evhtp_request_t* req)
                  << req->uri->path->full;
 
   if (req->method != htp_method_GET) {
-    evhtp_send_reply(req, EVHTP_RES_METHNALLOWED);
-    return;
+    HTTP_RESPOND_WITH_CODE(req, EVHTP_RES_METHNALLOWED, "Method Not Allowed");
```
rmccorm4, line 183: Eventually I think we can provide some mapping between code -> message, but done inline for simplicity for now.

src/http_server.cc
```diff
@@ -1453,13 +1467,8 @@ HTTPAPIServer::HandleRepositoryControl(
...
   }
 
-  if (err == nullptr) {
-    evhtp_send_reply(req, EVHTP_RES_OK);
-  } else {
-    EVBufferAddErrorJson(req->buffer_out, err);
-    evhtp_send_reply(req, EVHTP_RES_BADREQ);
-    TRITONSERVER_ErrorDelete(err);
-  }
+  HTTP_RESPOND_IF_ERR(req, err);
```
nnshah1, line 1470: is there a missing if / else here - to avoid sending OK if we already responded with error?

rmccorm4: The `_IF_ERR` variant checks `err != nullptr`, and returns early if there was an error, so the OK won't get hit.

rmccorm4: I'll fix a couple of the spots where its inside a redundant if block

src/http_server.cc
```diff
@@ -67,6 +67,44 @@ namespace triton { namespace server {
...
+  response.AddStringRef("error", message, strlen(message));
+
+  triton::common::TritonJson::WriteBuffer buffer_json;
+  response.Write(&buffer_json);
+
+  evbuffer_add(buffer, buffer_json.Base(), buffer_json.Size());
+}
+
+void
+EVBufferAddErrorJson(evbuffer* buffer, const std::string& error_str)
```
nnshah1, line 92: currently you don't need the overload for string here vs the const char *message - would this be for the future?

currently you could get away with EvBufferAddError(const char *message), and EvBufferAddError(TritonServer_Error*err)

src/http_server.cc
nnshah1: Maybe we should rename to 'RETURN_AND_RESPOND_IF_ERR' and 'RETURN_AND_RESPONE_WITH_ERR' 

to make consistent with other RETURN_IF_ERR macros?

rmccorm4: I don't mind either way. Can rename to RETURN prefix if you think that's more clear.

rmccorm4: Renamed

nnshah1 (approved):
Nice work!

PR #6298, opened 2023-09-12
Handle cancellation notification in gRPC server

tanmayv25:


src/grpc/infer_handler.h
```diff
@@ -859,7 +990,9 @@ class InferHandlerState {
 
   std::shared_ptr<Context> context_;
   Steps step_;
-  std::mutex step_mtx_;
+  std::recursive_mutex step_mtx_;
```
rmccorm4, line 993: Does this have to be recursive? Is it just to be able to cancel earlier while lock is already held?

tanmayv25: We can not release the state object if the InferResponseComplete is not yet called. The cancelled requests still invokes the Callback with cancelled response. And as state objects are reuse it is possible that InferResponseComplete of cancelled request receives response of a next request.
An alternative is to pass the current state object whose mutex is already held and skip holding the lock for that particular state when iterating through inflight states. But that seems little ugly and recursive locks fit the use-case.

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,137 @@ class InferHandlerState {
...
+      TRITONSERVER_Error* err = nullptr;
+      {
+        std::lock_guard<std::mutex> lock(mu_);
+
+        // Issues the request cancellation to the core.
+        for (auto state : inflight_states_) {
+          LOG_VERBOSE(1) << "Issuing cancellation for " << state->unique_id_;
+          bool is_cancelled = false;
+          err = TRITONSERVER_InferenceRequestIsCancelled(
+              state->irequest_ptr_, &is_cancelled);
```
rmccorm4, line 697: Do we need to bother checking? Or can we just cancel anyway? I don't think there's an error with "double" cancelling, just keeping the same bool true, but one less check to do.

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,137 @@ class InferHandlerState {
...
+          if ((err == nullptr) && (!is_cancelled)) {
+            err = TRITONSERVER_InferenceRequestCancel(state->irequest_ptr_);
+          }
+          if (err != nullptr) {
+            const char* request_id = "";
+            TRITONSERVER_Error* tmp_err = TRITONSERVER_InferenceRequestId(
```
rmccorm4, line 703: Can we use `RETURN_IF_ERROR` pattern instead?

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,137 @@ class InferHandlerState {
...
+        // Issues the request cancellation to the core.
+        for (auto state : inflight_states_) {
+          LOG_VERBOSE(1) << "Issuing cancellation for " << state->unique_id_;
+          bool is_cancelled = false;
+          err = TRITONSERVER_InferenceRequestIsCancelled(
+              state->irequest_ptr_, &is_cancelled);
+          if ((err == nullptr) && (!is_cancelled)) {
+            err = TRITONSERVER_InferenceRequestCancel(state->irequest_ptr_);
+          }
+          if (err != nullptr) {
```
rmccorm4, line 701: Can we use `RETURN_IF_ERROR` pattern instead?

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,137 @@ class InferHandlerState {
...
+
+    // Handles the gRPC context cancellation. This function can be called
+    // multiple times and is supposed to be re-entrant.
+    // Returns whether or not to continue cycling through the gRPC
+    // completion queue or not.
+    bool HandleCancellation(
+        InferHandlerStateType* state, bool rpc_ok, const std::string& name)
+    {
+      if (!IsCancelled()) {
+        LOG_ERROR << "[INTERNAL] HandleCancellation even when the context was "
```
rmccorm4, line 739:
```suggestion
        LOG_ERROR << "[INTERNAL] HandleCancellation called even when the context was "
```

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,137 @@ class InferHandlerState {
...
+                  << state->context_->unique_id_ << ", " << state->unique_id_
+                  << " step " << state->step_;
+        return true;
+      }
+      if ((state->step_ != Steps::CANCELLATION_ISSUED) &&
+          (state->step_ != Steps::CANCELLED)) {
+        LOG_VERBOSE(1) << "Cancellation notification received for " << name
+                       << ", rpc_ok=" << rpc_ok << ", context "
+                       << state->context_->unique_id_ << ", "
+                       << state->unique_id_ << " step " << state->step_;
```
rmccorm4, line 751: Side note, not for this PR, but it might be nice to have some lightly wrapped object or helper around some of these values used for logging purposes:
```cpp
struct RpcCallInfo {
  State state;
  bool rpc_ok;
};

ostream operator ...
{
  out <<  ", rpc_ok=" << rpc_ok << ", context "
        << state->context_->unique_id_ << ", "
        << state->unique_id_ << " step " << state->step_;
  return out;
}
```
So when we call this logging in most of our methods, it can be simplified to something like:
```cpp
LOG_VERBOSE(1) << "Cancellation notification received for: " << RpcCallInfo(state, rpc_ok);
```
Or something along those lines.

src/grpc/stream_infer_handler.cc
```diff
@@ -331,6 +348,8 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
         state->complete_ = true;
         state->context_->PutTaskBackToQueue(state);
       }
+    } else {
+      state->context_->InsertInflightState(state);
```
rmccorm4, line 352: Can you swap this so the first condition `if err == nullptr`, and the error handling is the `else` ? Also, split the comment across each condition:
```cpp
if (err == nullptr) {
    // If there was not an error in issuing the 'state' request then
    // state->step_ == ISSUED and inference request has
    // initiated, so mark this state as inflight.
    // The completion callback will transition to
    // WRITEREADY or WRITTEN.
    state->context_->InsertInflightState(state);
} else {
    // If there was an error then enqueue the
    // error response and show it to be ready for writing.
    ...
}
```

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,137 @@ class InferHandlerState {
...
+                    "]: " + TRITONSERVER_ErrorMessage(err))
+                    .c_str());
+            TRITONSERVER_ErrorDelete(err);
+            return wrapped_err;
+          }
+
+          {
+            std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
+            if (state->step_ != Steps::CANCELLED) {
+              state->step_ = Steps::CANCELLATION_ISSUED;
```
rmccorm4, line 722: Is `CANCELLATION_ISSUED` step required? Can we just cancel on the spot and set to CANCELLED?

src/grpc/stream_infer_handler.cc
```diff
@@ -508,6 +527,23 @@ ModelStreamInferHandler::Finish(InferHandler::State* state)
   return false;
 }
 
+void
+ModelStreamInferHandler::StreamInferRequestComplete(
+    TRITONSERVER_InferenceRequest* request, const uint32_t flags, void* userp)
+{
+  LOG_VERBOSE(1) << "ModelStreamInferHandler::StreamInferRequestComplete";
```
rmccorm4, line 534: Including request_id here would be nice for debugging

src/grpc/stream_infer_handler.cc
```diff
@@ -619,18 +678,40 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
+          << "ModelStreamInferHandler::StreamInferResponseComplete, "
+          << state->unique_id_
+          << ", skipping writing response because of transaction was cancelled";
+
+      // If this was the final callback for the state
+      // then cycle through the completion queue so
+      // that state object can be released.
+      if (state->complete_) {
+        state->step_ = Steps::CANCELLED;
+        state->context_->PutTaskBackToQueue(state);
```
rmccorm4, line 697: Why do we only cancel for the final callback? Say a decoupled request has 10 responses that each take a long time, would we not support cancelling after 1 or 2 responses?

src/grpc/infer_handler.h
tanmayv25: I am not sure. I was trying to conform with the API contract. For example, if `TRITONSERVER_InferenceRequestCancel` starts returning error for a request that was already cancelled, then we would break out of the loop early.

src/grpc/infer_handler.h
tanmayv25: I am trying to include the request id in the message for specific request that failed cancellation.

src/grpc/infer_handler.h
tanmayv25: Its the InferResponseComplete's responsibility to let the handler know when the state object can be cancelled and released.

src/grpc/stream_infer_handler.cc
tanmayv25: The other callback invocations would run into segmentation fault or incorrectly point to the state which is now tracking the next request.
It is the callback's responsibility to retrieve all the cancelled invocations are correctly processed.
Additionally, I presume that a response with cancelled state should have FINAL flag set and not issue anymore response callbacks. This would entail that gRPC frontend will cancel the state transaction as soon as it sees the cancelled response with FINAL flag - which can be 1 or 2 response. I didn't want to make this assumption in the gRPC frontend and make it flexible to handle mulitple cases.

tanmayv25:
I will address the clean-ups mentioned in another PR.

src/grpc/infer_handler.cc
```diff
@@ -982,6 +1022,28 @@ ModelInferHandler::InferResponseComplete(
...
+                   << state->unique_id_
+                   << ", skipping response generation as grpc transaction was "
+                      "cancelled... ";
+
+    // Send state back to the queue so that state can be released
+    // in the next cycle.
+    state->context_->PutTaskBackToQueue(state);
+
+    return;
+  }
```
rmccorm4, line 1045: This logic can probably be moved to a helper and re-used in the 2-3 places it is done

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,143 @@ class InferHandlerState {
...
     void DecrementRequestCounter() { ongoing_requests_--; }
 
+
+    // Inserts the state to a set tracking active requests
+    // within the server core.
+    void InsertInflightState(
+        InferHandlerStateType* state, TRITONSERVER_InferenceRequest* irequest)
+    {
+      std::lock_guard<std::mutex> lock(mu_);
+      state->irequest_ptr_ = irequest;
```
rmccorm4, line 674: Would be good to mention the logic behind saving/checking the request object for sanity checking state cleanup/reuse in CQ, as we discussed offline

src/grpc/infer_handler.h
```diff
@@ -781,6 +911,11 @@ class InferHandlerState {
     std::queue<InferHandlerStateType*> states_;
     std::atomic<uint32_t> ongoing_requests_;
 
+    // Tracks the inflight requests sent to Triton core via this
+    // context. We will use this structure to issue cancellations
+    // on these requests.
+    std::set<InferHandlerStateType*> inflight_states_;
```
rmccorm4, line 917: By using a set, are we possibly hiding any bugs/errors where we insert a duplicate state, with the parallel nature of Handler/ResponseCallback/AsyncNotifyWhenDone logic for adding states onto queue?

src/grpc/infer_handler.h
nnshah1: I would tend to agree there is no need to check here. In terms of the public API there is not a way for the request to become cancelled except from the front end. (backend api doesn't have such an api to cancel the request).  I think we would always allow double cancellation. There is also no way to gurantee that between checking and cancelling that the request doesn't become cancelled (i.e. if in the future we allow for the backend to cancel). I think logging but ignoring the error to cancel is probably best.

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,143 @@ class InferHandlerState {
...
+              // Ignore if failed to get request id.
+              TRITONSERVER_ErrorDelete(tmp_err);
+            }
+            TRITONSERVER_Error* wrapped_err = TRITONSERVER_ErrorNew(
+                TRITONSERVER_ERROR_INTERNAL,
+                std::string(
+                    std::string("[request id:") + request_id +
+                    "]: " + TRITONSERVER_ErrorMessage(err))
+                    .c_str());
+            TRITONSERVER_ErrorDelete(err);
```
Tabrizian: nit: can we make this a helper function (e.g., `AppendRequestIdToError`)?

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,143 @@ class InferHandlerState {
...
+        InferHandlerStateType* state, bool rpc_ok, const std::string& name)
+    {
+      if (!IsCancelled()) {
+        LOG_ERROR
+            << "[INTERNAL] HandleCancellation called even when the context was "
+               "not cancelled for "
+            << name << ", rpc_ok=" << rpc_ok << ", context "
+            << state->context_->unique_id_ << ", " << state->unique_id_
+            << " step " << state->step_;
+        return true;
```
Tabrizian, line 735: Ideally this should never happen is that correct?

src/grpc/infer_handler.h
```diff
@@ -652,12 +651,143 @@ class InferHandlerState {
...
+        if (state->context_->step_ != Steps::CANCELLED) {
+          TRITONSERVER_Error* err = nullptr;
+          err = IssueRequestCancellation();
+          if (err != nullptr) {
+            LOG_VERBOSE(1) << "Failed to cancel request: "
+                           << TRITONSERVER_ErrorMessage(err);
+            TRITONSERVER_ErrorDelete(err);
+          }
+          // Mark the context as cancelled
+          state->context_->step_ = Steps::CANCELLED;
```
Tabrizian, line 750: Is it possible for the `step` to be modified by another thread? Do we need to project it using a mutex?

src/grpc/infer_handler.h
tanmayv25: The items are added in this data structure iff the the InferAsync function was successfully called.

src/grpc/infer_handler.h
tanmayv25: We only have a single ModelStreamInferHandler thread. Hence, we don't need a lock around the context step_.

src/grpc/infer_handler.h
tanmayv25: Updated the flow.

src/grpc/infer_handler.h
tanmayv25: I have some more ideas to improve the code. I propose we improve it in the next iteration.

src/grpc/infer_handler.cc
tanmayv25: I will work on refactoring and reducing code duplication in next iteration.

kthui (approved):

PR #6308, opened 2023-09-13
Tests and docs for ability to specify subdirectory to download for LocalizePath

oandreeva-nv:
Tests and docs for this PR: https://github.com/triton-inference-server/core/pull/254

### DIFF
--- a/docs/user_guide/model_repository.md
+++ b/docs/user_guide/model_repository.md
@@ -142,6 +142,19 @@ $ gsutil iam ch allUsers:objectViewer "${BUCKET_URL}"
 $ gsutil iam ch allUsers:legacyBucketReader "${BUCKET_URL}"
 ```
 
+By default, Triton makes a local copy of a remote model repository in a temporary folder, which is deleted after Triton server is shut down.
+If you would like to control where remote model repository is copied to, you may set the
+`TRITON_GCS_MOUNT_DIRECTORY` environment variable to a path pointing to the existing folder
+on your local machine.
+
+```bash
+export TRITON_GCS_MOUNT_DIRECTORY=/path/to/your/local/directory
+```
+
+Make sure, that `TRITON_GCS_MOUNT_DIRECTORY` exists on your local machine.
+We recommend to provide an empty folder, otherwise make sure that the content of this folder
+is different from the content of your GCS bucket.
+
 #### S3
 
 For a model repository residing in Amazon S3, the path must be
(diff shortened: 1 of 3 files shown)


qa/L0_storage_azure/test.sh
```diff
@@ -169,6 +169,45 @@ for ENV_VAR in "shared_key"; do
...
+
+kill $SERVER_PID
+wait $SERVER_PID
+
+if [ -d "$TRITON_AZURE_MOUNT_DIRECTORY" ] && [ ! -z "$(ls -A $TRITON_AZURE_MOUNT_DIRECTORY)" ]; then
+    echo -e "\n***\n*** Test localization to a specified location failed. \n***"
+    echo -e "\n***\n*** Specified mount folder $TRITON_AZURE_MOUNT_DIRECTORY was not cleared properly. \n***"
+    ls -A $TRITON_AZURE_MOUNT_DIRECTORY
+    exit 1
+fi
```
rmccorm4, line 206: I may be misunderstanding, but isn't the point of this feature to deterministically cache cloud storage models/files for re-use across server runs? If we clean it up after server exit, what's the point of picking a specific directory if it can't be re-used?

oandreeva-nv: That's a good point. The emphasis of the ticket was mostly on specifying the non-random name for the local cache directory. By default, it was mounting to `/tmp/folderXXX`, so in my understanding, if it is more desirable to mount it to some external space through e.g. `/mnt/path/to/external/space`. We are cleaning up this directory by default all the time. It is likely that the full feature was vaguely described.

rmccorm4 (dismissed):
I'm happy with the test if we got agreement that it's fine to clean up the directory after server exit for now.

docs/user_guide/model_repository.md
```diff
@@ -177,6 +190,19 @@ If the environment variables are set they will take a higher priority
 and will be used by Triton instead of the credentials set using the
 aws config command.
 
+By default, Triton makes a local copy of a remote model repository
+in a temporary folder, which is deleted after Triton server is shut down.
+If you would like to control where remote model repository is copied to,
+you may set the `TRITON_AWS_MOUNT_DIRECTORY` environment variable to
```
kthui, line 196: I think we usually refer to Amazon storage as "S3", so it might be more appropriate to name the environment variable as `TRITON_S3_MOUNT_DIRECTORY` rather than `TRITON_AWS_MOUNT_DIRECTORY`.

oandreeva-nv: I was mostly inspired by what we use now, and environmental variables were using `AWS`

kthui: Double checking: "AWS" is used on environment variables and "S3" on everywhere else, but the environment variable names (i.e. "AWS_SECRET_ACCESS_KEY") are specified/used by the S3 client initially, so I think it is up to you if you want to use "S3" or "AWS" on the environment variable name.

kthui (approved):

PR #6332, opened 2023-09-20
Refactor HandleInfer into more readable chunks

rmccorm4:
1. Compartmentalize `HandleInfer` into logical helpers for better readability and re-usabilty for different inference handlers
2. Replace nested `if (err == nullptr)` logic with early return macro style for better readability. Added a callback helper to cleanup on error and resume the request if `connection_paused`. Currently it will always be paused when inference request is created, but left the variable in for clarity.
3. Set JSON content type before early returns so it's properly set on early errors, and add a helper to condense the code a bit.
4. Remove some unnecessary ifdefs around `trace` usage to remove some noise, as we already have [TraceManager](https://github.com/triton-inference-server/server/blob/5dd9398dd76a90a117ce6b3052e15561337fe88b/src/http_server.h#L340) in the code without any guards. The trace will just be `nullptr` when tracing isn't enabled, and the main trace logic is still protected by ifdefs.

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -3036,14 +3036,6 @@ HTTPAPIServer::HandleInfer(
   std::shared_ptr<TraceManager::Trace> trace =
       StartTrace(req, model_name, &triton_trace);
 
-  // Create the inference request object which provides all information needed
-  // for an inference.
-  TRITONSERVER_InferenceRequest* irequest = nullptr;
-  RETURN_AND_RESPOND_IF_ERR(
-      req, TRITONSERVER_InferenceRequestNew(
-               &irequest, server_.get(), model_name.c_str(),
-               requested_model_version));
-
   // Decompress request body if it is compressed in supported type
   evbuffer* decompressed_buffer = nullptr;
   RETURN_AND_RESPOND_IF_ERR(req, DecompressBuffer(req, &decompressed_buffer));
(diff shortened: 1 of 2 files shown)


src/http_server.cc
```diff
@@ -2884,6 +2905,117 @@ ForEachHeader(evhtp_header_t* header, void* arg)
...
+        TRITONSERVER_ERROR_UNSUPPORTED,
+        "HTTP end point doesn't support models with decoupled "
+        "transaction policy");
+  }
+
+  return nullptr;  // success
+}
+
+std::shared_ptr<TraceManager::Trace>
+HTTPAPIServer::StartTrace(
```
oandreeva-nv, line 2928: I may need to re-base on that for context propagation. Nice work!

dyastremsky (approved):
Beautiful code. This was a pleasure to review. Great working making the HTTP server code so readable!

src/http_server.h
```diff
@@ -209,9 +209,8 @@ class HTTPAPIServer : public HTTPServer {
 
     uint32_t IncrementResponseCount();
 
-#ifdef TRITON_ENABLE_TRACING
+    // Only used if tracing enabled
```
oandreeva-nv, line 212: Just checking if this builds ok without enabled tracing. I don't think that tracing libraries are available, when `TRITON_ENABLE_TRACING` is OFF, so that is why we used wrapper `#ifdef TRITON_ENABLE_TRACING` here

src/http_server.h
```diff
@@ -252,10 +251,25 @@ class HTTPAPIServer : public HTTPServer {
   // Get the inference header length. Return 0 if the whole request body is
   // the inference header.
   virtual TRITONSERVER_Error* GetInferenceHeaderLength(
-      evhtp_request_t* req, int32_t content_length, size_t* header_length);
+      evhtp_request_t* req, evbuffer* decompressed_buffer,
+      size_t* header_length);
```
rmccorm4, line 255: Need to fix this so it doesn't break override by vertex/sagemaker servers.

src/http_server.h
rmccorm4: Built fine for me, but will triple check and should be caught by L0_build_variants as well. Will let you know after pipeline.

dyastremsky (approved):

PR #6336, opened 2023-09-20
Refactor model generation scripts

Tabrizian:


qa/common/gen_common.py
```diff
@@ -0,0 +1,144 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Common utilities for model generation scripts
+
+
+def np_to_onnx_dtype(np_dtype):
```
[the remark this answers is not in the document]
dyastremsky: Nit: for consistency and clarity, would you be able to add a "return None" like in the other functions?

qa/common/gen_ensemble_model_utils.py
```diff
@@ -31,37 +31,13 @@
 import numpy as np
 import test_util as tu
 
+from .gen_common import np_to_model_dtype
```
dyastremsky, line 34: We can keep this, but I wanted to make a note that I'm seeing [this type of formatting](https://github.com/triton-inference-server/client/blob/5a9bd15899d64fa199cf7801df948a5f01b76ee6/src/python/library/tritonclient/grpc/_infer_stream.py#L35) in our client repo, where the file starts with an underscore. I could not find a recommended style for Python. I do see we have other repos (Model Navigator, Model Analyzer, PyTriton) that use this naming convention. Either approach is likely fine, though consistency would be helpful.

dyastremsky (commented):
Great work, Iman! Code looks much cleaner.

Added one nit and one optional observation.

qa/common/gen_ensemble_model_utils.py
Tabrizian: The `_` is mostly to distinguish the public APIs from private API. I think for these utility scripts what we have is fine since it is just an internal utility script.

dyastremsky: Got it. Thanks for explaining!

dyastremsky (approved):

qa/common/gen_common.py
```diff
@@ -0,0 +1,160 @@
...
+        return onnx.TensorProto.FLOAT
+    elif np_dtype == np.float64:
+        return onnx.TensorProto.DOUBLE
+    elif np_dtype == np_dtype_string:
+        return onnx.TensorProto.STRING
+    return None
+
+
+def np_to_model_dtype(np_dtype):
+    import numpy as np
```
dyastremsky, line 62: This looks to be a duplicate import. You import Numpy at the start of this script.

dyastremsky: Otherwise, the PR looks good and ready to approve!

PR #6345, opened 2023-09-23
Fixes on the gRPC frontend to handle AsyncNotifyWhenDone() API

tanmayv25:
There are few items that have been identified and fixed:
1. It is unsafe to call ctx_->IsCancelled() if the tag from AsyncNotifyWhenDone(). I am using a dummy State object that points to the first state object of ongoing context tracking the bidi grpc streaming RPC.
2. Introduced a new state step called PARTIAL_COMPLETE. We would need to handle the lifetimes of the state objects ourselves so that State is released only when all the RPC is complete. This means that callbacks have been processed, all the standard loops from the state->steps_ are complete and Done notification has been processed. The sad thing is there is no order guarantee on the standard CQ delivery and Done notification delivery.
3. There is a possibility that server when shutting down might not receive the Done() event. Exploring more on this,

I am adding some tests on the gRPC state cleanup logic for the following permutations and combinations.
[non-streaming, streaming non-decoupled, streaming decoupled][normal infer, cancelled, timeout, kill server, restricted API error].
I think if these tests pass. We should be confident on our end. 

There are some open gaps that got identified in the process of this work:

The server shutdown and handler are asynchronously invoked which might lead to certain race condition. Specifically, when calling Finish() on a listening RPC within handler. There are several points where it can occur. See [here](https://github.com/triton-inference-server/server/pull/6345#discussion_r1341666326). I will fix them in a separate PR as it is going out-of-context with this PR and the issue is already seem to be present in the current implementation. Opened a ticket here to track this: [DLIS-5563](https://jirasw.nvidia.com/browse/DLIS-5563)

### DIFF
--- a/src/grpc/grpc_utils.h
+++ b/src/grpc/grpc_utils.h
@@ -38,6 +38,25 @@
 
 namespace triton { namespace server { namespace grpc {
 
+// The step of processing that the state is in. Every state must
+// recognize START, COMPLETE and FINISH and the others are optional.
+typedef enum {
+  START,
+  COMPLETE,
+  FINISH,
+  ISSUED,
+  READ,
+  WRITEREADY,
+  WRITTEN,
+  WAITING_NOTIFICATION,
+  CANCELLATION_ISSUED,
+  CANCELLED,
+  PARTIAL_COMPLETION
+} Steps;
+
+// Debugging helper
+std::ostream& operator<<(std::ostream& out, const Steps& step);
+
 //
 // GrpcStatusUtil
 //
--- a/src/grpc/grpc_utils.h
+++ b/src/grpc/grpc_utils.h
@@ -41,16 +41,38 @@ namespace triton { namespace server { namespace grpc {
 // The step of processing that the state is in. Every state must
 // recognize START, COMPLETE and FINISH and the others are optional.
 typedef enum {
+  // This marks the starting stage of the RPC
   START,
+  // This marks that RPC is complete.
   COMPLETE,
+  // This marks the stage where all the notifications from the gRPC
+  // completion queue is received and state can be safely released.
   FINISH,
+  // This stage means that RPC has been issued to Triton for inference
+  // and is waiting for the server callbacks or cancellation to be
+  // invokded.
   ISSUED,
+  // This stage means the request has been read from the network and
+  // can be sent to Triton for execution.
   READ,
+  // This stage means that the response is ready to be written back to
+  // the network.
   WRITEREADY,
+  // This stage means that response has been written completely to the
+  // network.
   WRITTEN,
+  // This marks the special stage for the state object to differentiate
+  // the tag delivered from AsyncNotifyWhenDone() method.
   WAITING_NOTIFICATION,
+  // This stage means that the cancellation for the RPC has been issued
+  // to the server.
   CANCELLATION_ISSUED,
+  // This stage marks that the state has been successfully cancelled.
   CANCELLED,
+  // This is intermediary stage where the state has been been partially
+  // completed by grpc responder Finish call or AsyncNotifyWhenDone()
+  // notification. The other next call will move the stage to fully
+  // complete.
   PARTIAL_COMPLETION
 } Steps;
 
--- a/src/grpc/infer_handler.cc
+++ b/src/grpc/infer_handler.cc
@@ -37,45 +37,6 @@ NextUniqueId()
 
 namespace triton { namespace server { namespace grpc {
 
-std::ostream&
-operator<<(std::ostream& out, const Steps& step)
-{
-  switch (step) {
-    case START:
-      out << "START";
-      break;
-    case COMPLETE:
-      out << "COMPLETE";
-      break;
-    case FINISH:
-      out << "FINISH";
-      break;
-    case ISSUED:
-      out << "ISSUED";
-      break;
-    case READ:
-      out << "READ";
-      break;
-    case WRITEREADY:
-      out << "WRITEREADY";
-      break;
-    case WRITTEN:
-      out << "WRITTEN";
-      break;
-    case WAITING_NOTIFICATION:
-      out << "WAITING_NOTIFICATION";
-      break;
-    case CANCELLATION_ISSUED:
-      out << "CANCELLATION_ISSUED";
-      break;
-    case CANCELLED:
-      out << "CANCELLED";
-      break;
-  }
-
-  return out;
-}
-
 TRITONSERVER_Error*
 OutputBufferAttributesHelper(
     TRITONSERVER_ResponseAllocator* allocator, const char* tensor_name,
--- a/src/grpc/infer_handler.cc
+++ b/src/grpc/infer_handler.cc
@@ -62,6 +62,9 @@ operator<<(std::ostream& out, const Steps& step)
     case WRITTEN:
       out << "WRITTEN";
       break;
+    case WAITING_NOTIFICATION:
+      out << "WAITING_NOTIFICATION";
+      break;
     case CANCELLATION_ISSUED:
       out << "CANCELLATION_ISSUED";
       break;
@@ -785,6 +788,7 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
 #endif  // TRITON_ENABLE_TRACING
 
     state->step_ = Steps::FINISH;
+  } else if (state->step_ == Steps::FINISH) {
     finished = true;
   }
 
--- a/src/grpc/infer_handler.h
+++ b/src/grpc/infer_handler.h
@@ -713,7 +713,7 @@ class InferHandlerState {
     void InsertInflightState(
         InferHandlerStateType* state, TRITONSERVER_InferenceRequest* irequest)
     {
-      std::lock_guard<std::mutex> lock(mu_);
+      std::lock_guard<std::recursive_mutex> lock(mu_);
       // The irequest_ptr_ will get populated when it is
       // marked as active which means the request has been
       // successfully enqueued to Triton core using
@@ -726,7 +726,7 @@ class InferHandlerState {
     // within the server core.
     void EraseInflightState(InferHandlerStateType* state)
     {
-      std::lock_guard<std::mutex> lock(mu_);
+      std::lock_guard<std::recursive_mutex> lock(mu_);
       inflight_states_.erase(state);
     }
 
@@ -735,12 +735,13 @@ class InferHandlerState {
     void IssueRequestCancellation()
     {
       {
-        std::lock_guard<std::mutex> lock(mu_);
+        std::lock_guard<std::recursive_mutex> lock(mu_);
 
         // Issues the request cancellation to the core.
         for (auto state : inflight_states_) {
           std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
-          if (state->step_ != Steps::CANCELLED) {
+          if (state->step_ != Steps::CANCELLED &&
+              state->step_ != Steps::COMPLETE) {
             LOG_VERBOSE(1) << "Issuing cancellation for " << state->unique_id_;
             if (state->irequest_ptr_ == nullptr) {
               // The context might be holding some states that have
@@ -759,6 +760,11 @@ class InferHandlerState {
                        << TRITONSERVER_ErrorMessage(err);
             }
             state->step_ = Steps::CANCELLATION_ISSUED;
+          } else if (state->step_ == Steps::COMPLETE) {
+            // The RPC is complete and no callback will be invoked to retrieve
+            // the object. Hence, need to explicitly place the state on the
+            // completion queue.
+            PutTaskBackToQueue(state);
           }
         }
       }
@@ -831,7 +837,7 @@ class InferHandlerState {
     // correct order.
     void EnqueueForResponse(InferHandlerStateType* state)
     {
-      std::lock_guard<std::mutex> lock(mu_);
+      std::lock_guard<std::recursive_mutex> lock(mu_);
       states_.push(state);
     }
 
@@ -857,7 +863,7 @@ class InferHandlerState {
     // that it can be processed later
     void PutTaskBackToQueue(InferHandlerStateType* state)
     {
-      std::lock_guard<std::mutex> lock(mu_);
+      std::lock_guard<std::recursive_mutex> lock(mu_);
       // FIXME: Is there a better way to put task on the
       // completion queue rather than using alarm object?
       // The alarm object will add a new task to the back of the
@@ -875,7 +881,7 @@ class InferHandlerState {
     InferHandlerStateType* WriteResponseIfReady(
         InferHandlerStateType* required_state)
     {
-      std::lock_guard<std::mutex> lock(mu_);
+      std::lock_guard<std::recursive_mutex> lock(mu_);
       if (states_.empty()) {
         return nullptr;
       }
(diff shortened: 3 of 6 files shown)


src/grpc/infer_handler.h
```diff
@@ -767,12 +806,25 @@ class InferHandlerState {
...
 
-      return false;
+        return false;
+      } else {
+        // Should wait for the ResponseComplete callbacks to be invoked.
+        LOG_VERBOSE(1)
+            << "Waiting for the callback to retrieve cancellation for " << name
+            << ", rpc_ok=" << rpc_ok << ", context "
+            << state->context_->unique_id_ << ", " << state->unique_id_
+            << " step " << state->step_;
```
rmccorm4, line 824: Use new `DebugString()` ?

src/grpc/infer_handler.cc
```diff
@@ -785,6 +749,7 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
 #endif  // TRITON_ENABLE_TRACING
 
     state->step_ = Steps::FINISH;
+  } else if (state->step_ == Steps::FINISH) {
     finished = true;
```
rmccorm4, line 753: I believe this introduces an extra state transition/iteration for every COMPLETE state. I think it makes more logical sense to have states transition to the FINISH state first if that's what it's intended for like you're doing here, but just double checking this is expected from your changes.

rmccorm4: Actually, please help clarify my understanding on something. Assume a state is sitting on completion queue in the COMPLETE step.

1. [correct if wrong] The tag loop `cq_->Next(&tag, &ok)` will **pop** the tag (state) off of the completion queue, and pass it to `Process(state)`. So at this point, the state is **no longer on the completion queue**?
2. Process(state) will hit the `else if state->step_ == Steps::COMPLETE)` condition, and simply update the `state->step_ = Steps::FINISH`, and then the `Process` function with exit with `finished == false`. 
3. [correct if wrong] There are no GRPC specific calls (read/write/finish) made, nor any manual `PutTaskBackToQueue` call to put this state back onto the queue. 
4. Since the state has been popped off the queue, not put back onto the queue, and `finished == false`, this state won't get cleaned up by the loop logic right? Is this a leak?
```cpp
  // This condition won't be true for a COMPLETE state with these changes if I 
  // understand correctly, described above, so state won't get released?
 if (!Process(state, ok)) {
        LOG_VERBOSE(1) << "Done for " << Name() << ", " << state->unique_id_;
        state->context_->EraseState(state);
        StateRelease(state);
  }
```

src/grpc/infer_handler.h
```diff
@@ -1168,9 +1246,23 @@ InferHandler<
 
     while (cq_->Next(&tag, &ok)) {
       State* state = static_cast<State*>(tag);
+      if (state->step_ == Steps::WAITING_NOTIFICATION) {
+        State* state_wrapper = state;
+        state = state_wrapper->state_ptr_;
+        state->context_->SetReceivedNotification(true);
```
rmccorm4, line 1252: Is there any way for `state_wrapper->state_ptr_` to be prematurely cleaned up / nullptr / invalid / etc.?

src/grpc/stream_infer_handler.cc
```diff
@@ -142,9 +142,16 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
   // threads.
   {
     std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
-    if (state->IsGrpcContextCancelled()) {
-      bool resume = state->context_->HandleCancellation(state, rpc_ok, Name());
-      return resume;
+    if (state->context_->ReceivedNotification()) {
```
rmccorm4, line 145: Checking `ReceivedNotification()` is protected by `state->step_mtx_`, but setting `state->context_->SetReceivedNotification(true);` appears unprotected. Is this a race?

src/grpc/stream_infer_handler.cc
```diff
@@ -370,9 +379,13 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
 
     next_read_state->context_->responder_->Read(
         &next_read_state->request_, next_read_state);
-
+  } else if (state->step_ == Steps::PARTIAL_COMPLETION) {
+    state->step_ = Steps::COMPLETE;
   } else if (state->step_ == Steps::COMPLETE) {
     state->step_ = Steps::FINISH;
+  } else if (state->step_ == Steps::FINISH) {
```
rmccorm4, line 386: Similar to [other thread](https://github.com/triton-inference-server/server/pull/6345/files#r1337660823), how do these states get put back on queue to process future transitions, assuming `cq_->Next()` is popping them off the queue?

src/grpc/stream_infer_handler.cc
```diff
@@ -370,9 +379,13 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
 
     next_read_state->context_->responder_->Read(
         &next_read_state->request_, next_read_state);
-
+  } else if (state->step_ == Steps::PARTIAL_COMPLETION) {
+    state->step_ = Steps::COMPLETE;
```
rmccorm4, line 383: Similar to [other thread](https://github.com/triton-inference-server/server/pull/6345/files#r1337660823), how does this state get put back on queue to process future transitions, assuming cq_->Next() is popping them off the queue?

src/grpc/infer_handler.h
tanmayv25: DebugString is too much information and is being printed in super-verbose mode. 
Note that it prints all the states associated with the context and their each steps.

src/grpc/infer_handler.cc
tanmayv25: Let's take a step further back. When the step of the state is moved to COMPLETE, the infer handler thread invokes ::grpc::ServerAsyncResponseWriter<inference::ModelInferResponse> 's [Finish()](https://github.com/triton-inference-server/server/blob/tanmayv-fix/src/grpc/infer_handler.cc#L947) call with the state. This marks the beginning of the sending gRPC response. 
Upon the completion of the sending response, there are 2 events that get posted on the completion queue.
1. Finish is complete the completion_queue places the state object passed to the Finish() call.
2. The AsyncNotifyWhenDone() places the StateWrapper object - pointing to the same state - to the completion queue. 

Now these two events can be popped from the completion queue in any order. Hence, we don't differentiate between the two. Whenever the first state is read from the Next call, we silently transition from COMPLETE->FINISH with `finished==false` and rely on the second event to get FINISH step state and return with `finish==true` so the state can be released.

tanmayv25: >Is this a leak? Will we run out of states in state bucket?
Most likely not. I have tested with all the popular cases and both the events are being properly delivered to us in CQ.

src/grpc/infer_handler.h
tanmayv25: We are preserving the state object's lifetime till all the events and callbacks have been processed. We are not going to release the state until all these conditions are met.

src/grpc/stream_infer_handler.cc
tanmayv25: AsyncNotifyWhenDone()/Finish() call will put the state back to the compelition queue.

src/grpc/stream_infer_handler.cc
tanmayv25: There is no hazard in this case as there is a single streaming infer handler thread that receives the notification state and then later processes it.
The mutex protection is mainly to protect the state transitions happening within HandleCancellation()/HandleCompletion() against the ResponseComplete callback.

rmccorm4: Can you decrease the scope of the mutex to only the necessary section for future clarity then?

src/grpc/grpc_utils.h
```diff
@@ -38,6 +38,25 @@
...
+  FINISH,
+  ISSUED,
+  READ,
+  WRITEREADY,
+  WRITTEN,
+  WAITING_NOTIFICATION,
+  CANCELLATION_ISSUED,
+  CANCELLED,
+  PARTIAL_COMPLETION
+} Steps;
```
rmccorm4, line 55: Can you add short but descriptive comments to the states here? Especially the new ones. For example, explaining the purpose of PARTIAL_COMPLETION, and when/why it's used. This will make it more maintainable for others in the future.

src/grpc/infer_handler.h
```diff
@@ -949,6 +1013,7 @@ class InferHandlerState {
     // Clear trace_timestamps_ here so they do not grow indefinitely since
     // states are re-used for performance.
     ClearTraceTimestamps();
+    state_ptr_ = nullptr;
```
rmccorm4, line 1016: Comment variable purpose for wrapper handling cancellation etc.

rmccorm4 (commented):
General flow LGTM. Just looking for some minor comments and descriptions so this code is easier to understand when coming back to it in the future.

src/grpc/grpc_utils.h
```diff
@@ -38,6 +38,47 @@
...
+  // This marks the starting stage of the RPC
+  START,
+  // This marks that RPC is complete.
+  COMPLETE,
+  // This marks the stage where all the notifications from the gRPC
+  // completion queue is received and state can be safely released.
+  FINISH,
+  // This stage means that RPC has been issued to Triton for inference
+  // and is waiting for the server callbacks or cancellation to be
+  // invokded.
```
rmccorm4, line 53:
```suggestion
  // invoked.
```

src/grpc/grpc_utils.h
```diff
@@ -38,6 +38,47 @@
...
+  ISSUED,
+  // This stage means the request has been read from the network and
+  // can be sent to Triton for execution.
+  READ,
+  // This stage means that the response is ready to be written back to
+  // the network.
+  WRITEREADY,
```
rmccorm4, line 60: Can you clarify how this differs for decoupled/non-decoupled, if at all? From memory, I thought this step was more or less skipped for non-decoupled (1 response) case and it basically goes straight to WRITTEN.

src/grpc/grpc_utils.h
```diff
@@ -38,6 +38,47 @@
...
+  WAITING_NOTIFICATION,
+  // This stage means that the cancellation for the RPC has been issued
+  // to the server.
+  CANCELLATION_ISSUED,
+  // This stage marks that the state has been successfully cancelled.
+  CANCELLED,
+  // This is intermediary stage where the state has been been partially
+  // completed by grpc responder Finish call or AsyncNotifyWhenDone()
+  // notification. The other next call will move the stage to fully
+  // complete.
```
rmccorm4, line 75: From talking offline, is there an edge case here where the state has already been processed and gets stuck in PARTIAL_COMPLETION?

src/grpc/grpc_utils.h
tanmayv25: It doesn't get skipped. For non-decoupled models, the states can be in WRITEREADY and be blocked if they are not in front of the queue. For non-decoupled, we have a queue_ structure to ensure the writes are done in the order in which corresponding requests were received. 
For decoupled, we don't want requests generating large numbers of responses to block next requests that give small number of responses. Hence, we end up sending the responses as and when the response in generated by backend.

src/grpc/grpc_utils.h
tanmayv25: I have to think about this more. As per my reading and experimentation, a race condition exists in our current implementation as well.
Specifically, we don't know at this point here: https://github.com/triton-inference-server/server/blob/main/src/grpc/stream_infer_handler.cc#L183-L195
Whether the stream was closed from the client side or the server is in shutdown. By introducing custom sleep, I was able to cause a gRPC assert error because it tries to call Finish() on a compeletion queue that is in shutdown state.

So, it is not getting stuck, but it might cause grpc assert when closing the server while running heavy streaming inference workload.

rmccorm4: >  By introducing custom sleep, I was able to cause a gRPC assert error because it tries to call Finish() on a compeletion queue that is in shutdown state.
>
> So, it is not getting stuck, but it might cause grpc assert when closing the server while running heavy streaming inference workload.

Is this a graceful failure? or do we need to catch this somehow?

tanmayv25: assert_error is not a graceful error. I think a better way would be to prime the handlers for cancellation.

src/grpc/stream_infer_handler.cc
```diff
@@ -515,14 +521,17 @@ ModelStreamInferHandler::Finish(InferHandler::State* state)
   // entire stream. Otherwise just finish this state.
   if (state->context_->IsRequestsCompleted()) {
     state->context_->step_ = Steps::COMPLETE;
-    state->step_ = Steps::COMPLETE;
+    state->step_ = Steps::PARTIAL_COMPLETION;
+    LOG_VERBOSE(2) << "Finishing responder from state " << state->unique_id_;
```
rmccorm4, line 525: From PR description:
> There is a possibility that server when shutting down might not receive the Done() event. Exploring more on this,

@tanmayv25 is this resolved? Or still exploring? Can you update PR description to reflect latest coverage/gaps?

tanmayv25: This is a similar server shutdown problem. The race condition is very specific. However, the good thing is we call `TRITONSERVER_ServerStop()` before shutting down the completion queue. So, only the handler thread will be interacting with the completion queue at this point. Opened DLIS-5563 to address this.

tanmayv25: Updated the PR description as well.

src/grpc/infer_handler.h
```diff
@@ -912,8 +979,24 @@ class InferHandlerState {
 
     // True if there is an ongoing write to the grpc stream
     std::atomic<bool> ongoing_write_;
+
+    // Tracks whether the async notification has been delivered by
+    // completion queue.
+    bool received_notification_;
```
rmccorm4, line 985: Should this be atomic? I think this is concurrently accessed by response callback via `state->IsGrpcContextCancelled()`

src/grpc/infer_handler.h
```diff
@@ -656,25 +640,86 @@ class InferHandlerState {
 
     void GrpcContextAsyncNotifyWhenDone(InferHandlerStateType* state)
     {
-      ctx_->AsyncNotifyWhenDone(state);
+      InferHandlerStateType* wrapped_state =
+          new InferHandlerStateType(Steps::WAITING_NOTIFICATION, state);
+      notify_state_ = state;
+      ctx_->AsyncNotifyWhenDone(wrapped_state);
```
rmccorm4, line 646: Since we already have state, wrapped_state, and wrapped_state->state_ptr, I think we should avoid having yet another state with notify_state. Let's instead toggle some boolean flag on the state like `state.is_notify_state_` or something and check that rather than comparing state pointers.

src/grpc/stream_infer_handler.cc
rmccorm4: Follow-up is fine, thanks

src/grpc/grpc_utils.h
rmccorm4: Ok, separate PR is fine. Add a ticket if needed for tracking.

src/grpc/infer_handler.h
tanmayv25: The response callback only reads the variable. And it is not unsafe to try to read it prematurely before the cancellation notification is registered. The state transition will be ignored.

src/grpc/infer_handler.h
tanmayv25: Done!

rmccorm4 (approved):

PR #6353, opened 2023-09-26
Add L0_grpc_state_cleanup test

tanmayv25:
Locally the server seems to be quite stable for these tests. Other tests in the CI are also looking stable.

To-Do:
~~1. Add tests for non-decoupled streaming.~~
~~2. Add tests for non-streaming inferences.~~

After these two are added. We can be quite confident on the new state transitions in the gRPC frontend. 

Additional follow-up is to move decoupled_infer to a common infer utilities where different tests can utilize these calls.

### DIFF
--- a/qa/L0_grpc_state_cleanup/cleanup_test.py
+++ b/qa/L0_grpc_state_cleanup/cleanup_test.py
@@ -54,34 +54,99 @@ def callback(user_data, result, error):
     else:
         user_data._response_queue.put(result)
 
-
+# These state cleanup tests relies on the test.sh
+# to check whether all the created request objects
+# were properly deleted by the sever. 
+# The purpose on these unittest is to exercise 
+# different portions of the gRPC frontend and
+# and track the state objects.
 class CleanUpTest(tu.TestResultCollector):
     SERVER_PID = None
 
     def setUp(self):
         self.decoupled_model_name_ = "repeat_int32"
-        self.sequence_model_name_ = "simple_sequence"
-        self.identity_model_name_ = "simple_identity"
+        self.identity_model_name_ = "custom_zero_1_float32"
 
-        self.inputs_ = []
-        self.inputs_.append(grpcclient.InferInput("IN", [1], "INT32"))
-        self.inputs_.append(grpcclient.InferInput("DELAY", [1], "UINT32"))
-        self.inputs_.append(grpcclient.InferInput("WAIT", [1], "UINT32"))
+    def _prepare_inputs_and_outputs(self, kind):
+        if kind == "decoupled_streaming":
+            self.inputs_ = []
+            self.inputs_.append(grpcclient.InferInput("IN", [1], "INT32"))
+            self.inputs_.append(grpcclient.InferInput("DELAY", [1], "UINT32"))
+            self.inputs_.append(grpcclient.InferInput("WAIT", [1], "UINT32"))
 
-        self.outputs_ = []
-        self.outputs_.append(grpcclient.InferRequestedOutput("OUT"))
-        self.outputs_.append(grpcclient.InferRequestedOutput("IDX"))
-        self.requested_outputs_ = self.outputs_
+            self.outputs_ = []
+            self.outputs_.append(grpcclient.InferRequestedOutput("OUT"))
+            self.outputs_.append(grpcclient.InferRequestedOutput("IDX"))
+            self.requested_outputs_ = self.outputs_
+        elif kind == "simple" or kind == "streaming":
+            self.inputs_ = []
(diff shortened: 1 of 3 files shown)


qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -0,0 +1,382 @@
...
+                    # Request IDs should generally be provided with each request
+                    # to associate decoupled responses with their requests.
+                    if not response.id:
+                        raise ValueError(
+                            "No response id found. Was a request_id provided?"
+                        )
+
+                    # Detect final response. Parameters are oneof and we expect bool_param
+                    if response.parameters.get("triton_final_response").bool_param:
+                        completed_requests += 1
```
rmccorm4, line 137: If a request gets cancelled before it makes it to the response callback, I don't think this parameter can get set. Will this test/loop just hang forever then? Is this a missed test case?

tanmayv25: We are raising the exception as soon as it gets delivered in the callback. So, I think we should be good.

qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -0,0 +1,560 @@
...
+        # released when the server is interrupted to shutdown in middle of
+        # inference run with final parameters being returned.
+        with self.assertRaises(InferenceServerException) as cm:
+            self._decoupled_infer(
+                request_count=10,
+                repeat_count=10,
+                request_delay=1,
+                kill_server=5,
+                should_error=True,
+                infer_helper_map=[False, True],
```
rmccorm4, line 554: Can just pass infer helpers directly, don't need to map them to these bools

qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -0,0 +1,560 @@
...
+                    raise data_item
+                else:
+                    this_id = data_item.get_response().id
+                    if this_id not in result_dict:
+                        result_dict[this_id] = []
+                    result_dict[this_id].append((recv_count, data_item))
+
+                recv_count += 1
+
+    def _streaming_infer(
```
rmccorm4, line 270: If you feel like refactoring, you can use the same code for decoupled and non-decoupled streaming inference. It's the same process. I don't think you need all these different functions for it.

rmccorm4: Similarly, async_infer and async_stream_infer are ~90% the same code too. Main difference is whether you pass callback in start_stream() or in infer() call, which you can toggle with `if streaming` or something. Should reduce a lot of the code in this test for easier maintenance if you wanted to.

qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -0,0 +1,560 @@
...
+                        self.assertEqual(len(this_data), 1)
+                        self.assertEqual(this_data[0], expected_data)
+                        this_idx = result_list[j][1].as_numpy("IDX")
+                        self.assertEqual(len(this_idx), 1)
+                        self.assertEqual(this_idx[0], j)
+                        expected_data += 1
+
+    ###
+    ### Streaming Tests
+    ###
```
rmccorm4, line 406: Non-streaming tests?

qa/L0_grpc_state_cleanup/test.sh
```diff
@@ -0,0 +1,194 @@
...
+  fi
+  set -e
+done
+
+
+for i in test_simple_infer_error_status \
+                test_streaming_error_status \
+                test_decoupled_error_status; do
+  SERVER_LOG="./inference_server.$i.log"
+  SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=2 --grpc-restricted-protocol=inference,health:infer-key=infer-value"
```
rmccorm4, line 121: Is health endpoint intended to be tested here?

rmccorm4 (commented):
Overall test logic LGTM. Main feedback is mostly to reduce some duplicate code. If it's too much effort to refactor right now, please make a ticket to follow-up.

kthui (approved):
LGTM overall. I wonder if some other part of Triton prints "StateRelease" or "StateNew" to the server log in the future, would it interfere with this test? Agree with @rmccorm4 that some duplicate code could be consolidated.

qa/L0_grpc_state_cleanup/cleanup_test.py
tanmayv25: These helpers are class function and self is not defined in the arguments here.

PR #6365, opened 2023-09-29
Add TRT-LLM backend build to Triton

krishung5:
Currently the build is using internal GitLab repo for both TRT-LLM backend and TRT-LLM repositories. Will switch the link to the GitHub repo once they're ready.

TRT-LLM backend: https://github.com/triton-inference-server/tensorrtllm_backend/pull/3

### DIFF
--- a/build.py
+++ b/build.py
@@ -1714,13 +1714,14 @@ def core_build(
             os.path.join(repo_install_dir, "lib", "libtritonserver.so"),
             os.path.join(install_dir, "lib"),
         )
+    # FIXME: Uncomment this part when using the latest branch
     # [FIXME] Placing the Triton server wheel file in 'python' for now, should
     # have been upload to pip registry and be able to install directly
-    cmake_script.mkdir(os.path.join(install_dir, "python"))
-    cmake_script.cp(
-        os.path.join(repo_install_dir, "python", "tritonserver*.whl"),
-        os.path.join(install_dir, "python"),
-    )
+    # cmake_script.mkdir(os.path.join(install_dir, "python"))
+    # cmake_script.cp(
+    #     os.path.join(repo_install_dir, "python", "tritonserver*.whl"),
+    #     os.path.join(install_dir, "python"),
+    # )
 
     cmake_script.mkdir(os.path.join(install_dir, "include", "triton"))
     cmake_script.cpdir(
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -1726,6 +1768,12 @@ def core_build(
     cmake_script.blankln()
 
 
+def tensorrtllm_prebuild(cmake_script):
+    # Export the TRT_ROOT environment variable
+    cmake_script.cmd("export TRT_ROOT=/usr/local/tensorrt")
```
tanmayv25, line 1773: This can be simplified if we are replacing TRT8 with TRT9 in-place.

build.py
```diff
@@ -1746,7 +1794,16 @@ def backend_build(
     cmake_script.comment()
     cmake_script.mkdir(build_dir)
     cmake_script.cwd(build_dir)
-    cmake_script.gitclone(backend_repo(be), tag, be, github_organization)
+    # FIXME: Use GitHub repo
+    if be == "tensorrtllm":
+        cmake_script.gitclone(
```
tanmayv25, line 1799: Can you remind me what was the reason for not using github instead? I think we are not exposing anything different here.

build.py
```diff
@@ -1319,7 +1319,10 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
         )
         trtllm_buildscript = importlib.util.module_from_spec(spec)
         exec(response.content, trtllm_buildscript.__dict__)
-        df += trtllm_buildscript.create_postbuild()
+        # FIXME: Specify these versions via build flag?
```
tanmayv25, line 1322: I think better alternative is to put it in Triton version map at the top.

build.py
krishung5: Synced offline that we will improve the TRT9 installation in the future as right now it requires to modify the build script in the TRT-LLM repo.

build.py
krishung5: It was for the WAR for private repo.

build.py
krishung5: Great idea. Updated the file.

krishung5:
@tanmayv25 Accidentally amended the commit when solving some pre-commit issues so the addressed changes are not showing as a separate commit. The PR is ready for review, thanks!

tanmayv25 (approved):

PR #6366, opened 2023-09-29
Add basic generate endpoints for LLM tasks

GuanLuo:
Test case is the example to interact with a vLLM-mock model using the endpoints.
Quick curl command:
```
curl -X POST localhost:8000/v2/models/vllm_proxy/generate -d '{"PROMPT": "text", "STREAM": false}'
```
yields
```
{"TEXT" : "response"}
```


src/http_server.cc
```diff
@@ -3078,154 +3078,217 @@ HTTPAPIServer::HandleGenerate(
...
+      &input_metadata));
 
-  // Parse simple request format
-  triton::common::TritonJson::Value request;
-  RETURN_AND_RESPOND_IF_ERR(req, EVRequestToJson(req, &request));
 
-  // Fallbacks
-  std::string prompt = "";
-  triton::common::TritonJson::Value params_json;
+  // [FIXME] decompression should have been done here. before parsing request body
```
rmccorm4, line 3086: Do we need to support compression out of the box? I think it's OK as an enhancement. The simple `curl` use case is pretty minimal as far as I can tell.

GuanLuo: You are correct that it is just for completeness purposes.

src/http_server.cc
```diff
@@ -3078,154 +3077,262 @@ HTTPAPIServer::HandleGenerate(
...
+          TRITONSERVER_InferenceRequestDelete(irequest),
+          "deleting HTTP/REST inference request");
 
-  // TODO: Parse Model Config for candidate input names/keys to compare and
-  // populate
-  triton::common::TritonJson::Value config_json;
-  RETURN_AND_RESPOND_IF_ERR(
-      req, config_json.Parse(config_json_str.c_str(), config_json_str.size()));
+      evhtp_request_resume(req);
+      RETURN_AND_RESPOND_IF_ERR(req, error);
```
rmccorm4, line 3126: This will double delete the error. Can return msg directly instead and let `RETURN_CALLBACK_IF_ERR` delete the error:
```cpp
      evhtp_request_resume(req);
      EVBufferAddErrorJson(req->buffer_out, error);
      evhtp_send_reply(req, EVHTP_RES_BADREQ);
```

src/http_server.cc
```diff
@@ -336,29 +336,27 @@ JsonBytesArrayByteSize(
...
-      *byte_size += len + sizeof(uint32_t);
     }
-    TRITONSERVER_ErrorDelete(assert_err);
+  } else {
+    // Serialized data size is the length of the string itself plus
+    // 4 bytes to record the string length.
+    const char* str;
+    size_t len = 0;
+    RETURN_MSG_IF_ERR(
+        tensor_data.AsString(&str, &len), "Unable to parse JSON bytes array");
```
rmccorm4, line 356: FYI I think this macro leaks memory on errors: 
https://github.com/triton-inference-server/server/blob/f4d427841767601134ad7d13f1952fe7cbb34f72/src/common.h#L65-L69

It looks like it creates a new error and returns early, not deleting the original error.

src/http_server.cc
```diff
@@ -331,140 +369,140 @@ ReadDataFromJsonHelper(
 {
   // FIXME should invert loop and switch so don't have to do a switch
```
jbkyang-nvi, line 370: should we delete this comment?

qa/L0_http/llm_test.py
```diff
@@ -0,0 +1,95 @@
...
+        r.raise_for_status()
+
+        # Validate SSE format
+        self.assertIn("Content-Type", r.headers)
+        self.assertIn("text/event-stream", r.headers["Content-Type"])
+
+        # SSE format (data: []) is hard to parse, use helper library for simplicity
+        client = sseclient.SSEClient(r)
+        res_count = 0
+        for i, event in enumerate(client.events()):
```
Tabrizian, line 85: Do we have some follow up tickets to enhance the Python/C++ clients to use this endpoint?

src/http_server.cc
GuanLuo: will update comment, here doesn't really change the handling to be address (still switch on each element even though all elements are the same data type)

qa/L0_http/llm_models/vllm_proxy/1/model.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
```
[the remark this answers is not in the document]
GuanLuo: Renamed to `mock_llm`

qa/L0_http/generate_endpoint_test.py
```diff
@@ -0,0 +1,354 @@
...
+        return requests.post(
+            url,
+            data=inputs if isinstance(inputs, str) else json.dumps(inputs),
+            headers=headers,
+            stream=stream,
+        )
+
+    def generate(self, model_name, inputs):
+        url = self._get_infer_url(model_name, "generate")
+        return requests.post(url, data=json.dumps(inputs))
```
tanmayv25, line 65: Why don't we support direct string based input here?

qa/L0_http/generate_models/mock_llm/config.pbtxt
```diff
@@ -0,0 +1,62 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+backend: "python"
+
+# Disabling batching in Triton, let vLLM handle the batching on its own.
```
tanmayv25, line 28: remove

qa/L0_http/generate_models/mock_llm/config.pbtxt
```diff
@@ -0,0 +1,62 @@
...
+
+output [
+  {
+    name: "TEXT"
+    data_type: TYPE_STRING
+    dims: [ 1, -1 ]
+  }
+]
+
+# The usage of device is deferred to the vLLM engine
```
tanmayv25, line 56: remove

qa/L0_http/generate_endpoint_test.py
```diff
@@ -0,0 +1,354 @@
...
+            url, data=inputs if isinstance(inputs, str) else json.dumps(inputs)
+        )
+        try:
+            r.raise_for_status()
+            self.assertTrue(False, f"Expected failure, success for {inputs}")
+        except requests.exceptions.HTTPError as e:
+            self.assertIn(msg, r.json()["error"])
+
+    def generate_stream_expect_failure(self, model_name, inputs, msg):
+        r = self.generate_stream(model_name, inputs)
```
tanmayv25, line 79: nit: As this will generate multiple responses, can one of the latter response be an error? Should we add this test case? We may use SSEClient trick here as well to make this function more generic?

tanmayv25: I see the case being included below. Can somehow we use this function to detect error in next responses? 
This can even take the response idx that is supposed to generate an error.

qa/L0_http/generate_endpoint_test.py
```diff
@@ -0,0 +1,354 @@
...
+        # With "REPETITION" > 0, the first response is valid response and set
+        # HTTP code to success, so user must validate each response
+        inputs["REPETITION"] = 1
+        r = self.generate_stream(self._model_name, inputs)
+        r.raise_for_status()
+
+        self.check_sse_responses(r, [{"TEXT": text}, {"error": "An Error Occurred"}])
+
+    def test_race_condition(self):
+        # HTTP response send logic and Triton response complete logic are
```
tanmayv25, line 248: I don't fully understand what is this testing... Do you want to test the thread-safety on the server?

qa/L0_http/generate_endpoint_test.py
```diff
@@ -0,0 +1,354 @@
...
+        input2 = {
+            "PROMPT": "hello",
+            "STREAM": True,
+            "REPETITION": 3,
+            "param": "segfault",
+        }
+        threads = []
+
+        def thread_func(model_name, input):
+            self.generate_stream(model_name, input1).raise_for_status()
```
tanmayv25, line 262: input1->input?

qa/L0_http/generate_endpoint_test.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+        for thread in threads:
+            thread.start()
+        for thread in threads:
+            thread.join()
+
+    def test_one_response(self):
+        # "STREAM" controls response behavior,
+        # True sends two responses, one with infer response and one with flag
+        # only, where generate endpoint will be able to handle as in
```
tanmayv25, line 279: > True sends two responses, one with infer response and one with flag
   only, where generate endpoint will be able to handle as in non-decoupled mode.

Can you improve the comment here?  May be callout generate_stream behavior as well? 
How are flag only responses represented for generate_stream?

qa/L0_http/generate_endpoint_test.py
```diff
@@ -0,0 +1,354 @@
...
+            r.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            self.assertIn(
+                "generate expects model to produce exactly 1 response",
+                r.json()["error"],
+            )
+
+    def test_many_response(self):
+        inputs = {"PROMPT": "hello world", "STREAM": True, "REPETITION": 2}
+        r = self.generate_stream(self._model_name, inputs)
```
tanmayv25, line 311: no need to iterate through responses?

src/http_server.h
```diff
@@ -230,6 +253,91 @@ class HTTPAPIServer : public HTTPServer {
     std::atomic<uint32_t> response_count_;
   };
 
+  class GenerateRequestClass : public InferRequestClass {
```
tanmayv25, line 256: nit: Can we move this to  a separate file? Something similar to how we broke down composing handlers in gRPC?
Makes sense to just start with Generate handler.

src/http_server.h
```diff
@@ -230,6 +253,91 @@ class HTTPAPIServer : public HTTPServer {
...
+    const MappingSchema* response_schema_{nullptr};
+    const bool streaming_{false};
+    // Pointer to associated Triton request, this class does not own the
+    // request and must not reference it after a successful
+    // TRITONSERVER_ServerInferAsync.
+    TRITONSERVER_InferenceRequest* triton_request_{nullptr};
+    // Placeholder to completing response, this class does not own
+    // the response.
+    TRITONSERVER_InferenceResponse* triton_response_{nullptr};
+    // As InferResponseComplete and ChunkResponseCallback are called in
```
tanmayv25, line 333: I see... So, this was the race condition the test was trying to exercise?

src/http_server.cc
```diff
@@ -3015,40 +3063,36 @@ HTTPAPIServer::ForwardHeaders(
...
+      req, ModelInputMetadata(
+               model_name, requested_model_version, &input_metadata,
+               &meta_data_root));
 
-  // Get the header length
-  size_t header_length = 0;
-  RETURN_AND_RESPOND_IF_ERR(
-      req, GetInferenceHeaderLength(req, content_length, &header_length));
+
+  // [FIXME] decompression should have been done here. before parsing request
```
tanmayv25, line 3094: We should keep in mind to document these limitations. However, I don't think compression is any useful for text data..

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
-  infer_request->trace_ = trace;
+  std::unique_ptr<GenerateRequestClass> generate_request;
+  if (streaming) {
+    generate_request.reset(new GenerateRequestClass(
+        server_.get(), req, GetResponseCompressionType(req),
+        generate_stream_request_schema_.get(),
+        generate_stream_response_schema_.get(), streaming, irequest));
+  } else {
+    generate_request.reset(new GenerateRequestClass(
+        server_.get(), req, GetResponseCompressionType(req),
```
tanmayv25, line 3115: So, we are supporting the response compression but not request compression? We should include it in our testing if we are planning to leave it here.

qa/L0_http/generate_endpoint_test.py
GuanLuo: Yes

src/http_server.h
GuanLuo: That is planned, but not in this PR as I would like to move out InferRequestClass as well

src/http_server.h
GuanLuo: Yes

src/http_server.cc
GuanLuo: No, both directions are not supported, this is passed due to the parent class expects it in constructor. The compression logic is not properly encapsulated so it doesn't automatically kick in, and frankly I think some investigation is needed to figure out how chunk + compression works. Therefore, compression is not under consideration for now.

qa/L0_http/generate_endpoint_test.py
GuanLuo: Tested in `test_generate_stream` and other variants.

qa/L0_http/generate_endpoint_test.py
GuanLuo: This is different from `test_generate_stream_response_error` that the error is not part of the inference response (error before async infer). This kind of error are returned in the same way as other Triton endpoint error and can't be checked with the SSE client.

qa/L0_http/generate_endpoint_test.py
GuanLuo: flag only end response chunk, not actual message are sent back. Same as current grpc streaming

src/common.h
```diff
@@ -63,10 +63,12 @@ const std::vector<std::string> TRITON_RESERVED_REQUEST_PARAMS{
   do {                                                                 \
     TRITONSERVER_Error* err__ = (X);                                   \
     if (err__ != nullptr) {                                            \
-      return TRITONSERVER_ErrorNew(                                    \
+      auto new_err = TRITONSERVER_ErrorNew(                            \
```
nnshah1, line 66: Future scope / question:

Can we reset the message of the the error instead of creating a new one and deleting the old one?  Maybe requires a new / unwanted api TRITONSERVER_ErrorSetMessage() ....

src/http_server.cc
```diff
@@ -296,29 +336,27 @@ JsonBytesArrayByteSize(
...
-  for (size_t i = 0; i < tensor_data.ArraySize(); i++) {
-    triton::common::TritonJson::Value el;
-    RETURN_IF_ERR(tensor_data.At(i, &el));
-
-    // Recurse if not last dimension...
-    TRITONSERVER_Error* assert_err =
-        el.AssertType(triton::common::TritonJson::ValueType::ARRAY);
-    if (assert_err == nullptr) {
+  // Recurse if not last dimension...
+  TRITONSERVER_Error* assert_err =
```
nnshah1, line 340: Question / Future Scope:

Not sure why we use the assertion here instead of `IsArray()` with the error we have to clean up , etc. - do we need it?

https://github.com/triton-inference-server/common/blob/cf617c93c75d8e03da962115a2471e4b89062aae/include/triton/common/triton_json.h#L941

src/http_server.cc
```diff
@@ -329,142 +367,144 @@ ReadDataFromJsonHelper(
...
-    if (assert_err == nullptr) {
+  // FIXME should move 'switch' statement outside the recursive function and
+  // pass in a read data callback once data type is confirmed.
+  // Currently 'switch' is performed on each element even through all elements
+  // have the same data type.
+
+  // Recurse on array element if not last dimension...
+  TRITONSERVER_Error* assert_err =
+      tensor_data.AssertType(triton::common::TritonJson::ValueType::ARRAY);
+  if (assert_err == nullptr) {
```
nnshah1, line 378: same as above could we use IsArray() ?

src/http_server.cc
```diff
@@ -2775,6 +2801,28 @@ HTTPAPIServer::ParseJsonTritonRequestID(
...
+  struct evbuffer_iovec* v = nullptr;
+  int v_idx = 0;
+  int n = evbuffer_peek(req->buffer_in, -1, NULL, NULL, 0);
+  if (n > 0) {
+    v = static_cast<struct evbuffer_iovec*>(
+        alloca(sizeof(struct evbuffer_iovec) * n));
+    if (evbuffer_peek(req->buffer_in, -1, NULL, v, n) != n) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INTERNAL,
+          "[POC] unexpected error getting request buffers");
```
nnshah1, line 2818: Question what does `[POC]` mean in this context?

src/http_server.cc
nnshah1: Instead of getting it from the req - can we error if it's not IDENTITY and/or pass IDENTITY as the type (and leave a note / TODO)

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
+  for (const auto& m : members) {
+    auto it = schema->children_.find(m);
+    if (it != schema->children_.end()) {
+      switch (it->second->kind_) {
+        case MappingSchema::Kind::EXACT_MAPPING: {
+          // Read meta data
+          RETURN_IF_ERR(ExactMappingInput(m, generate_request, input_metadata));
+          break;
+        }
+        case MappingSchema::Kind::MAPPING_SCHEMA: {
```
nnshah1, line 3268: Question: are providing support for mapping in 23.10 or exact mapping only?

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
+  } else {
+    // Parse data type and shape
+    std::string value;
+    it->second.MemberAsString("datatype", &value);
+    auto dtype = TRITONSERVER_StringToDataType(value.c_str());
+
+    // Perform shape validation, assume the value must be either
+    // primitive type or 1-D array.
+    triton::common::TritonJson::Value tensor_data;
+    if (!generate_request.Find(name.c_str(), &tensor_data)) {
```
nnshah1, line 3312: Question / Future:

Could `Members()` return a map of name to JSON Value - and `ExactMappingInput()` take name and value? Just to avoid having to check again - seems we already know its a member (if I understand)

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INVALID_ARG,
+          (std::string("unexpected key not found in generate request, "
+                       "expecting key '") +
+           name + "'")
+              .c_str());
+    }
+
+    auto assert_err =
+        tensor_data.AssertType(triton::common::TritonJson::ValueType::ARRAY);
```
nnshah1, line 3322: could we use IsArray() to avoid having to delete the error?

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
+      // a two-pass process to pad the request value to match input shape.
+      // 1. iterate shape for fixed dimension to distribute 'element_cnt'.
+      // 2. Set most inner dynamic shape to the remaining element count,
+      //    other dynamic shape to be 1.
+      for (auto rit = shape_vec.rbegin(); rit != shape_vec.rend(); ++rit) {
+        if (*rit != -1) {
+          if (element_cnt % *rit) {
+            return TRITONSERVER_ErrorNew(
+                TRITONSERVER_ERROR_INVALID_ARG,
+                "The schema can not convert input to tensor with proper shape");
```
nnshah1, line 3356: Update with tensor nam

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
+      for (auto rit = shape_vec.rbegin(); rit != shape_vec.rend(); ++rit) {
+        if (*rit == -1) {
+          *rit = element_cnt;
+          element_cnt = 1;
+        }
+      }
+      if (element_cnt != 1) {
+        return TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_INVALID_ARG,
+            "The schema can not convert input to tensor with proper shape");
```
nnshah1, line 3370: Add tensor name for error where it makes sense

src/http_server.cc
```diff
@@ -3060,120 +3104,449 @@ HTTPAPIServer::HandleInfer(
...
+      byte_size = element_cnt * TRITONSERVER_DataTypeByteSize(dtype);
+    }
+
+    std::vector<int64_t> shape_vec;
+    {
+      triton::common::TritonJson::Value value;
+      if (!it->second.Find("shape", &value)) {
+        return TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_INTERNAL,
+            "unexpected 'shape' not found in model metadata");
```
nnshah1, line 3339: add model name where makes sense for error messages

src/http_server.cc
```diff
@@ -3568,6 +3941,416 @@ HTTPAPIServer::InferRequestClass::IncrementResponseCount()
...
+  if (strncmp(request_id, "", 1)) {
+    triton_outputs.emplace(
+        "response_id", TritonOutput(TritonOutput::Type::RESERVED, request_id));
+  }
+  const char* model_name;
+  int64_t model_version;
+  RETURN_IF_ERR(TRITONSERVER_InferenceResponseModel(
+      response, &model_name, &model_version));
+  triton_outputs.emplace(
+      "model_name", TritonOutput(TritonOutput::Type::RESERVED, model_name));
```
nnshah1, line 4109: Question: model name, response_id, and model version - are always sent? (responseid if request id is present).

Can we use 'id' to match infer ? Or 'request_id` for both - to avoid any confusion with partial results, etc.

src/http_server.cc
```diff
@@ -3568,6 +3941,416 @@ HTTPAPIServer::InferRequestClass::IncrementResponseCount()
...
+    const std::map<
+        std::string, HTTPAPIServer::GenerateRequestClass::TritonOutput>&
+        output_metadata,
+    const MappingSchema* schema,
+    triton::common::TritonJson::Value* generate_response,
+    std::set<std::string>* mapped_outputs)
+{
+  for (auto& nested : schema->children_) {
+    switch (nested.second->kind_) {
+      case MappingSchema::Kind::MAPPING_SCHEMA: {
```
nnshah1, line 4232: same as above - double checking if we support mapping schema or just exact match in first version

src/http_server.cc
```diff
@@ -3568,6 +3941,416 @@ HTTPAPIServer::InferRequestClass::IncrementResponseCount()
...
+{
+  mapped_outputs->emplace(name);
+
+  switch (triton_output.type) {
+    case TritonOutput::Type::RESERVED: {
+      generate_response->AddStringRef(
+          name.c_str(), triton_output.value.c_str());
+      break;
+    }
+    case TritonOutput::Type::PARAMETER: {
```
nnshah1, line 4281: do parameters automatically become top level properties?

do we support keeping them in a "parameters" object (for either request or response parameters?)

src/http_server.h
```diff
@@ -44,6 +45,25 @@
...
+  enum class Kind {
+    EXACT_MAPPING,
+    // An object of this kind means it is a nested mapping schema.
+    MAPPING_SCHEMA
+  };
+  std::map<std::string, std::unique_ptr<MappingSchema>> children_;
+  // Whether an unspecified key is allowed. If true,
+  // * for requests, the unspecified key will be converted to Triton input
+  //   following the EXACT_MAPPING rule.
+  // * for responses, the Triton output will be converted to JSON key-value
```
nnshah1, line 59: for responses - it is also true it follows EXACT_MAPPING rule if not specfiied- is that right?

src/http_server.h
```diff
@@ -230,6 +253,91 @@ class HTTPAPIServer : public HTTPServer {
...
+            input_metadata,
+        const MappingSchema* schema,
+        triton::common::TritonJson::Value& generate_request);
+
+    const MappingSchema* RequestSchema() { return request_schema_; }
+    const MappingSchema* ResponseSchema() { return response_schema_; }
+
+   private:
+    struct TritonOutput {
+     public:
```
nnshah1, line 298: nit: struct are public by default - could remove -

src/http_server.h
```diff
@@ -367,6 +500,14 @@ class HTTPAPIServer : public HTTPServer {
   re2::RE2 systemsharedmemory_regex_;
   re2::RE2 cudasharedmemory_regex_;
   re2::RE2 trace_regex_;
+
+  // [DLIS-5551] currently always perform basic conversion
+  std::unique_ptr<MappingSchema> generate_request_schema_{new MappingSchema()};
```
nnshah1, line 505: Question: does basic conversion include schema mapping?

If not could we comment out / error that code for clarity in this version?

nnshah1 (commented):
Looks good to me - some questions and minor suggestions.

Need to update some error messages to include tensor / model names.

for unsupported compression type - would be good to error / fix to identity

if exact mapping is what is supported - would comment / error on any other type for now

src/http_server.cc
GuanLuo: No mapping, mapping is placeholder for nested object

src/common.h
GuanLuo: Right, that needs new API.. You can create ticket for it if you want, but I don't see how useful it can be.

src/http_server.cc
GuanLuo: Yes. Can you create a ticket for that? I don't want to change the json library in this PR.

src/http_server.cc
GuanLuo: Still under the nested object category and need to specify the schema to group model outputs into an object.

src/http_server.h
GuanLuo: Is this going to block PR approval? Commenting out this needs to touch many places and soon to be un-commented. It will error out already if user pass a complex message, see `test_complex_schema`.

Tabrizian (approved):

src/http_server.cc
nnshah1: nvm - thought members was a method of a new class not part of JSON libraryt - agree - not in scope

rmccorm4: @nnshah1 if you file a ticket, an enhancement I would also like to see is to include the JSON key/name on conversions somewhere. Maybe `Value` can have a `name_` field or something.

Currently if you pass a bad datatype like where a string is expected for the value:
```
{"prompt": 123}
```
the error message would be something like 
```
"attempt to access JSON non-string as string"
```

But it would be a lot easier to debug if the error was:
```
"attempt to access JSON non-string as string for 'prompt' field"
```
or
```
"attempt to access JSON non-string (123) as string"
```
so at least you can associate which field was wrong.

ref: https://github.com/triton-inference-server/common/blob/cf617c93c75d8e03da962115a2471e4b89062aae/include/triton/common/triton_json.h#L713

src/http_server.h
nnshah1: It's more about code readability / reasoning - difficult to understand if its supported or not. Also may decide not to support it / etc.

Maybe just a comment before the convert functions?

src/http_server.cc
nnshah1: @GuanLuo - recommend we put the same error handling as in line 3090 - can we add check for responsecompressiontype there as well - to ensure it is IDENTITY?

src/http_server.cc
```diff
@@ -3568,6 +3946,416 @@ HTTPAPIServer::InferRequestClass::IncrementResponseCount()
...
+      RETURN_IF_ERR(TRITONSERVER_InferenceResponseOutput(
+          triton_response_, triton_output.index, &cname, &datatype, &shape,
+          &dim_count, &base, &byte_size, &memory_type, &memory_type_id,
+          &userp));
+
+      auto info = reinterpret_cast<AllocPayload::OutputInfo*>(userp);
+      // sanity check
+      if (info->kind_ != AllocPayload::OutputInfo::JSON) {
+        return TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_INTERNAL, "Unexpected output response type");
```
nnshah1, line 4334: need output name in error

src/http_server.cc
```diff
@@ -3568,6 +3946,416 @@ HTTPAPIServer::InferRequestClass::IncrementResponseCount()
...
+        case TRITONSERVER_PARAMETER_BOOL:
+        case TRITONSERVER_PARAMETER_INT:
+        case TRITONSERVER_PARAMETER_STRING:
+          triton_outputs.emplace(
+              name, TritonOutput(TritonOutput::Type::PARAMETER, pidx));
+          break;
+        case TRITONSERVER_PARAMETER_BYTES:
+          return TRITONSERVER_ErrorNew(
+              TRITONSERVER_ERROR_UNSUPPORTED,
+              "Response parameter of type 'TRITONSERVER_PARAMETER_BYTES' is "
```
nnshah1, line 4141: would need name of parameter causing error

nnshah1 (changes requested):
Please review error messages where name of input / output etc. are needed to know where conversion / parsing went wrong.

Nice to have: Check for response compression type as identity (maybe already covered) 

Nice to have: comment above conversion function indicating other mapping types are WIP

src/http_server.cc
nnshah1: @GuanLuo - is this feasible? I think we do need to have the field name in error message.

GuanLuo: Yes for `MemberAsXXX` functions, note that it is also change in json library.

nnshah1 (approved):
Nice work! LGTM

src/http_server.cc
```diff
@@ -108,6 +108,46 @@ AddContentTypeHeader(evhtp_request_t* req, const char* type)
...
+  if (!params_json.Find(parameter.c_str(), &value)) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        ("parameter key '" + parameter + "' was not found in the JSON")
+            .c_str());
+  }
+
+  if (value.IsString()) {
+    std::string string_value;
+    RETURN_IF_ERR(value.AsString(&string_value));
```
nnshah1, line 127: One option would be to use RETURN_MSG_IF_ERR on the json conversions and form a message with the name of the variable.

We could create a single string with the name and use it.

rmccorm4 (approved):
🔥

PR #6387, opened 2023-10-04
Fix L0_sdk

tanmayv25:


qa/L0_sdk/test.sh
```diff
@@ -155,7 +155,7 @@ fi
 # Check wheels, note that even TRITON_VERSION is passed as version field for
 # wheel generation. The version number will be normalized by setuptools, so
 # we need to replace the text here as well to match the normalized version.
-WHLVERSION=`cat /workspace/TRITON_VERSION | sed 's/dev\./\.dev/'`
+WHLVERSION=`cat /workspace/TRITON_VERSION | sed 's/dev/\.dev0/'`
```
mc-nv, line 158: Have a suggestion, may be it worth to use following logic instead and do not stick to the version after all? 
```
if [[ $(ls  triton_client/python/tritonclient-*-py3-none-any.whl  triton_client/python/tritonclient-*-py3-none-*linux*.whl | wc -l ) != 2 ]] ; then
    echo -e "*** unexpected number of wheels in triton_client/python\n"
    echo -e "*** available wheels in triton_client/python\n"
    ls -ltr triton_client/python/
    RET=1
fi
```

jbkyang-nvi (approved):

qa/L0_sdk/test.sh
tanmayv25: I would prefer keeping it as it is now. If we end up changing the wheel nomenclature then we can rivisit to relax the naming convention on this test.

PR #6392, opened 2023-10-05
Add TRT-LLM backend build to Triton (#6365)

krishung5:
* Add trtllm backend to build

* Temporarily adding version map for 23.07

* Fix build issue

* Update comment

* Comment out python binding changes

* Add post build

* Update trtllm backend naming

* Update TRTLLM base image

* Fix cmake arch

* Revert temp changes for python binding PR

* Address comment

### DIFF
--- a/build.py
+++ b/build.py
@@ -26,6 +26,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import argparse
+import importlib.util
 import multiprocessing
 import os
 import os.path
@@ -35,6 +36,8 @@ import stat
 import subprocess
 import sys
 from inspect import getsourcefile
+
+import requests
 
 #
 # Build Triton Inference Server.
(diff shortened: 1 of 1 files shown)


krishung5:
There will be another merge request on the GitLab to bring all the TRT-LLM backend changes to the main. Both PRs will need to be merged before code freeze.

build.py
```diff
@@ -1278,6 +1315,27 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
     pip3 install --upgrade numpy && \
     rm -rf /var/lib/apt/lists/*
 """
+    # Add dependencies needed for tensorrtllm backend
+    if "tensorrtllm" in backends:
+        be = "tensorrtllm"
+        import importlib.util
+
+        import requests
```
mc-nv, line 1323: Wondering why `import` statement is in function body? 

Outcome will be the same, if package installation missed function won't work. <br>

Having the `import` statement in the header of the body gives us an advantage to catch issue with configuration of the runtime environment without prerequisites or condition match.

krishung5: That's good point. I think the intention behind this was that we only want to import those packages when we need them. e.g. the `importlib.util` and `requests` are only needed when building FT and TRT-LLM backends.

krishung5:
Rebase

build.py
mc-nv: I don't see any specific PIP package installation for given `import` statement assume it safe to move to the top.

krishung5: Addressed.

mc-nv (approved):

PR #6394, opened 2023-10-06
Update build.py to build vLLM backend

dyastremsky:


build.py
```diff
@@ -1279,6 +1279,23 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
...
+        df += install_miniconda(argmap["CONDA_VERSION"], target_machine)
+        vllm_conda_env_bash = "https://raw.githubusercontent.com/triton-inference-server/vllm_backend/main/tools/environment.yml"
+
+        # The last copy is needed so that user could run vllm models with unpacked
+        # conda env
+        df += """
+# conda environment is required for the vllm backend
+RUN conda install -y mamba -c conda-forge  && \
+    wget "{vllm_conda_env_bash}" && \
+    mamba env create -f environment.yml && \
```
tanmayv25, line 1293: I am looking into another ways than providing `environment.yml` with hard coded package versions.

nnshah1 (dismissed):
LGTM - overall script could use refactor - but out of scope here.

build.py
nnshah1: Should we start with environment.yml and then change if we find something better?

tanmayv25: Ok. Let's do this.

build.py
```diff
@@ -1768,6 +1785,29 @@ def backend_build(
     cmake_script.blankln()
 
 
+def backend_clone(
+    be,
+    cmake_script,
```
tanmayv25, line 1790: cmake_script -> clone_script? 
Just for this scope?

build.py
```diff
@@ -1768,6 +1785,29 @@ def backend_build(
...
+):
+    repo_clone_dir = os.path.join(install_dir, "backends")
+
+    cmake_script.commentln(8)
+    cmake_script.comment(f"'{be}' backend")
+    cmake_script.comment("Delete this section to remove backend from build")
+    cmake_script.comment()
+    cmake_script.mkdir(repo_clone_dir)
+    cmake_script.cwd(repo_clone_dir)
+    cmake_script.gitclone(backend_repo(be), tag, be, github_organization)
```
tanmayv25, line 1803: Additionally, it will be helpful if we can provide vllm version from this script to the backend to use from the VERSION_MAP above.

tanmayv25: This can be done later.

build.py
dyastremsky: Sure, done.

build.py
```diff
@@ -1768,6 +1785,29 @@ def backend_build(
...
+):
+    repo_clone_dir = os.path.join(install_dir, "backends")
+
+    clone_script.commentln(8)
+    clone_script.comment(f"'{be}' backend")
+    clone_script.comment("Delete this section to remove backend from build")
+    clone_script.comment()
+    clone_script.mkdir(repo_clone_dir)
+    clone_script.cwd(repo_clone_dir)
+    clone_script.gitclone(backend_repo(be), tag, be, github_organization)
```
tanmayv25, line 1803: Need to address this: 
Olga:
Regarding build.py, current (locally working approach) is these lines: https://github.com/triton-inference-server/server/blob/oandreeva_vllm_ci/build.py#L1283-L1295
+ build.py currently clones the whole github repo to `install/backends`, this would need to be adjusted slightly to not copy `ci` and `tools` folders

dyastremsky: Good find. Added. Looking online, there is no way to exclude subdirectories from a clone (or, inversely, only include some subdirectories), so I will simply force remove these extra directories for now. Long-term, we could look at whether we're copying these directories somewhere and could make sure we're only copying the directory with the Python-based backend itself.

tanmayv25: Why not clone into tmp_dir(repo-dir) as all other backends and only copy the model.py in the backend directory(install dir)?

dyastremsky: Good idea, thanks for talking about this. Done!

tanmayv25:
Impressive work @dyastremsky!

tanmayv25 (approved):

rmccorm4 (approved):
🚀

PR #6396, opened 2023-10-06
support parameters object in generate route

nnshah1:
Adding a specific key word for generate endpoint to support passing parameters such as:

curl -s localhost:8000/v2/models/mock_llm/generate_stream -d '{"PROMPT":"hello","STREAM":true, "parameters"\
        :{"REPETITION":5}}'

### DIFF
--- a/qa/L0_http/generate_endpoint_test.py
+++ b/qa/L0_http/generate_endpoint_test.py
@@ -363,6 +363,18 @@ class GenerateEndpointTest(tu.TestResultCollector):
         inputs = {"PROMPT": [text], "STREAM": True, "parameters":{"REPETITION": rep_count}}
         self.generate_stream_expect_success(self._model_name, inputs, text, rep_count)
 
+        # non object parameters
+        inputs = {"PROMPT": [text], "STREAM": True, "parameters": 1}
+
+        r = self.generate(self._model_name, inputs)
+        try:
+            r.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            self.assertIn(
+                "Expected JSON object for keyword: 'parameters'", r.json()["error"]
+            )
+
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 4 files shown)


qa/L0_http/generate_endpoint_test.py
```diff
@@ -356,6 +356,13 @@ def test_complex_schema(self):
                 "attempt to access JSON non-string as string", r.json()["error"]
             )
 
+    def test_parameters(self):
+        # Test reserved nested object for parameters
+        text = "hello world"
+        rep_count = 3
+        inputs = {"PROMPT": [text], "STREAM": True, "parameters":{"REPETITION": rep_count}}
```
rmccorm4, line 363: Any negative tests? Would this fail if parameters value wasn't an object? ex:
```
        inputs = {"PROMPT": [text], "STREAM": True, "parameters": "some_string"}

```

rmccorm4 (commented):
LGTM, just need to lint for pre-commit

rmccorm4 (approved):

src/http_server.cc
```diff
@@ -1082,6 +1082,19 @@ HTTPAPIServer::HTTPAPIServer(
...
+  // or parameters based on model config
+
+  // Note: may change in future versions
+  const std::string parameters_field = "parameters";
+  generate_stream_request_schema_->children_.emplace(
+      parameters_field,
+      new MappingSchema(MappingSchema::Kind::MAPPING_SCHEMA, true));
+  generate_request_schema_->children_.emplace(
+      parameters_field,
+      new MappingSchema(MappingSchema::Kind::MAPPING_SCHEMA, true));
```
GuanLuo, line 1097: Given that there will be a specific LLM schema, can start grouping the schema building into a function call.

rmccorm4 (approved):

GuanLuo (approved):

PR #6406, opened 2023-10-10
TRT-LLM backend build changes

krishung5:
For 23.10, we won't bring any changes to the structure of TRT-LLM backend. This PR fixes some path issues so that Triton can build using CMake. Also adding some temporary workaround to solve the repo permission issues. We will update those once the GitHub repo is ready.


tanmayv25 (approved):

krishung5:
@tanmayv25 It seems like their script to install new TRT also installs some cudnn packages, which increase the image size. I changed that part back for the post build for now, and I think we could improve that in the future.

tanmayv25 (dismissed):

tanmayv25 (approved):

krishung5:
Rebased

PR #6412, opened 2023-10-11
Document generate HTTP endpoint

GuanLuo:


docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Generate Extension
+
+This document describes Triton's generate extension. The generate
+extension provides a simple text oriented endpoint schema for interacting
```
dyastremsky, line 32: Nit: text-oriented

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+with LLM models. This generate endpoint is specific to HTTP/REST frontend.
+
+## HTTP/REST
+
+In all JSON schemas shown in this document `$number`, `$string`, `$boolean`,
```
dyastremsky, line 37: Nit: "All JSON schemas..." (additional word: in)

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+
+### generate v.s. generate_stream
+
+Both URLs expect the same request JSON object, and generate the same response
+JSON object. However, `generate` return exactly 1 response JSON object, while
+`generate_stream` may return various number of responses based on the inference
+results. `generate_stream` returns the responses as
+[Server-Sent Events](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events)
+(SSE), where each response will be a "data" chunk in the HTTP response body.
+Also note that an error may be returned during inference, whereas the HTTP
```
dyastremsky, line 60: Nit: "Also, note that an error may be returned during inference, where the HTTP response code has been set in the first response of the SSE, which can result in receiving an error object while the status code shows success (200) Therefore, the"...

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+receiving error object while status code is success (200). Therefore the user
+must always check whether an error object is received when generating responses
+through `generate_stream`.
+
+### Generate Request JSON Object
+
+The generate request object, identified as *$generate_request*, is
+required in the HTTP body of the POST request. The model name and
+(optionally) version must be available in the URL. If a version is not
+provided the server may choose a version based on its own policies or
```
dyastremsky, line 71: Nit: "provided, the server..."

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+      "text_input" : $string,
+      "parameters" : $parameters #optional
+    }
+
+* "text_input" : The text input that the model should generate output from.
+* "parameters" : An optional object containing zero or more parameters for this
+  generate request expressed as key/value pairs. See
+  [Parameters](#parameters) for more information.
+
+#### Parameters
```
dyastremsky, line 85: Isn't there also the easier route that can be shown where if you have a parameter that has a distinct name from any inputs, you can simply put it as an input and it will work?

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+      $parameter, ...
+    }
+
+    $parameter = $string : $string | $number | $boolean
+
+Parameters are model-specific. In other words, the accepting names and values
+varies according to the model. The user should check with the model
+specification to set the parameters.
+
+### Generate Response JSON Object
```
dyastremsky, line 102: I don't know if this deviates too much from existing documentation, but would you be able to provide an easy example to save users time and help with understanding? For example:
`curl -X POST localhost:8000/v2/models/vllm_model/generate -d '{"prompt": "text", "stream": false, "temperature": 0}'`

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+$string, $number, or $boolean.
+
+    $parameters =
+    {
+      $parameter, ...
+    }
+
+    $parameter = $string : $string | $number | $boolean
+
+Parameters are model-specific. In other words, the accepting names and values
```
dyastremsky, line 98: Nit: remove accepting? Is this supposed to mean something along the lines of "the keys and values are specific to the model"?

dyastremsky (commented):
Fantastic work, this is such a wonderful feature! The content looks great. Added some nits on wording and a couple of requests for potential additions for clarity.

docs/protocol/extension_generate.md
dyastremsky: An example like this one with a description could also be a good way to show how you can send both inputs and parameters quite easily.

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Generate Extension
+
+This document describes Triton's generate extension. The generate
+extension provides a simple text oriented endpoint schema for interacting
+with LLM models. This generate endpoint is specific to HTTP/REST frontend.
```
nnshah1, line 33:
```suggestion
with large language models (LLMs). This generate endpoint is specific to HTTP/REST frontend.
```

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+```
+POST v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]/generate
+
+POST v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]/generate_stream
+```
+
+### generate v.s. generate_stream
+
+Both URLs expect the same request JSON object, and generate the same response
+JSON object. However, `generate` return exactly 1 response JSON object, while
```
nnshah1, line 55:
```suggestion
JSON object. However, `generate` returns exactly 1 response JSON object, while
```

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+`generate_stream` may return various number of responses based on the inference
+results. `generate_stream` returns the responses as
+[Server-Sent Events](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events)
+(SSE), where each response will be a "data" chunk in the HTTP response body.
+Also note that an error may be returned during inference, whereas the HTTP
+response code has been set at first response of the SSE, which can result in
+receiving error object while status code is success (200). Therefore the user
+must always check whether an error object is received when generating responses
```
nnshah1, line 63: Can we hyper link here to the error object?

docs/protocol/extension_generate.md
nnshah1, line 98:
```suggestion
Parameters are model-specific. The user should check with the model specification to set the parameters.
```

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,126 @@
...
+required in the HTTP body of the POST request. The model name and
+(optionally) version must be available in the URL. If a version is not
+provided the server may choose a version based on its own policies or
+return an error.
+
+    $generate_request =
+    {
+      "text_input" : $string,
+      "parameters" : $parameters #optional
+    }
```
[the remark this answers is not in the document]
GuanLuo: Do you want this to be part of the generate schema? Seems to be de-valuing the purpose of adding reserved "parameters" field.

docs/protocol/extension_generate.md
GuanLuo: Maybe below? The subject of the sentence is the types.
```suggestion
In all JSON schemas shown in this document, `$number`, `$string`, `$boolean`,
```

docs/protocol/extension_generate.md
GuanLuo: See https://github.com/triton-inference-server/server/pull/6412#discussion_r1355419563, documenting this then introduces ambiguity on how to specify additional inputs. If there are two (documented) ways to achieve the same goal, when should the user pick one over the other.

docs/protocol/extension_generate.md
rmccorm4: Do we want to include "model_name" and "model_version" in output description here? I think these are returned currently.

docs/protocol/extension_generate.md
```diff
@@ -0,0 +1,171 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+-->
+
+# Generate Extension
+
+**Note:** The Generate Extension is *provisional* and likely to change in future versions.
```
rmccorm4, line 31: Optional, but can get a nice render for notes using this syntax:
```
> [!NOTE]
> The Generate Extension is *provisional* and likely to change in future versions.
```

ex:
image

rmccorm4 (commented):
LGTM overall. Only remaining comments from me:

1. (optional) note syntax for more visible render
2. I would also like to see an full example `curl` for both generate and generate_stream mentioned in other comments, just so users have something easy to copy/paste 
3. I agree with Guan that it may be better to hide the implicit top-level parameter behavior for now if we want to encourage `parameters` keyword usage.

docs/protocol/extension_generate.md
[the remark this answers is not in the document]
dyastremsky: I think the parameters route might be easier for users, or at least I would find it easier as a user.

I think there could be a way to document both approaches without confusing users, if done right. This approach is good because it allows people to be more explicit with which are parameters, so the feature still has value (e.g. if an input and parameter have identical names). I think the question though should be less about what de-values a feature and more about what provides the most ease and value to users.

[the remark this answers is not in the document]
nnshah1: I'm good either way - we can add a note:

```suggestion
   > [!NOTE] > Any additional properties in the request object are passed either as parameters or tensors based on model specification. 
```

but in all examples let's use 'parameters' for consistency (i.e. not show both forms). 

I believe we'll refine this further and may either lock the schema further or make it more flexible. Since this is provisional we won't guarantee backwards compat (yet)

nnshah1 (dismissed):
LGTM - I think the note by @rmccorm4 makes sense I had not found the best syntax.

nnshah1 (approved):

rmccorm4 (approved):

PR #6425, opened 2023-10-12
Improve Python Backend Testing

rmccorm4:
shm_util.py
- :warning: Change shm probe check from inequality `!=` to strictly growth `>`. Some subtests like `io/io_test.py` were failing with slight **decreases** in shm usage after adding probe to the test, and I didn't spend too much time to figure out why.
- [debug improvement] Add debug string (model name/type) to SHM leak probe to more immediately see failing cases in CI logs

python_unittest.py
- [coverage] add back shm leak testing for bls and decoupled tests
- [coverage] run more iterations of unittest while monitoring SHM for NIGHTLY runs

python_test.py
- [coverage] add shm probe to some uncovered test cases

bls
- [reduce intermittence] improve bls_tensor_lifecycle gpu memory intermittence by using torch cuda allocator memory stats rather than global gpu memory stats to avoid noise from other gpu processes

ensemble
- [reduce intermittence] increase `atol` tolerance for FP32 comparisons to reduce intermittence on outputs within `1e-06` of each other previously causing failures
- [cleanup] re-use common code and break up single test into 2 cpu/gpu tests

io
- [coverage] add shm probe to test
- [cleanup] flatten some of the nested loops with `itertools.product`
- [FIXME] test sees small **decrease** in SHM usage - would be good to understand why - see comment.

logging
- [coverage] add shm probe to test
- [FIXME/out-of-scope] this test can probably be refactored to use `python_unittest.py` structure like some of the other tests

restart
- [coverage] add shm probe to test
- [FIXME/out-of-scope] this test can probably be refactored to use `python_unittest.py` structure like some of the other tests

---

decoupled
- [WIP] add back SHM leak testing


qa/L0_backend_python/python_unittest.py
```diff
@@ -55,32 +55,14 @@ def _run_unittest(self, model_name):
...
-                # this we need to use the shared memory probe only on the later
-                # call so that the probe can detect the leak correctly.
-                self._run_unittest(model_name)
+        # Warmup model, let it do any upfront initializations required
+        self._run_unittest(model_name)
 
-                # [FIXME] See DLIS-3684
+        with self._shm_leak_detector.Probe() as shm_probe:
+            for i in range(5):
+                print(f"[{model_name}] Test iteration: {i}")
```
rmccorm4, line 64: @Tabrizian does this change make sense? 

I'm trying to express:
1. Do any upfront "warmup" that would take some initial amount of memory for every use case
2. Then probe several follow-up runs to make sure the memory doesn't increase over the initial required upper bound

rmccorm4: Just want to sanity check that logic makes sense, and that I'm not potentially missing a leak by doing it this way.

qa/L0_backend_python/ensemble/ensemble_test.py
```diff
@@ -70,36 +69,18 @@ def test_ensemble(self):
                 self.assertIsNotNone(output0)
                 self.assertIsNotNone(output1)
 
-                self.assertTrue(np.allclose(output0, 2 * input_data_0))
-                self.assertTrue(np.allclose(output1, 2 * input_data_1))
+                # Set a big enough tolerance to reduce intermittence. May be
+                # better to test integer outputs in the future for consistency.
+                self.assertTrue(np.allclose(output0, 2 * input_data_0, atol=1e-06))
+                self.assertTrue(np.allclose(output1, 2 * input_data_1, atol=1e-06))
```
rmccorm4, line 75: Improved intermittence on ensemble test:
```
# ensemble test atol

FAIL: [-1.4901161e-08 -2.9802322e-08  0.0000000e+00  5.9604645e-08
  0.0000000e+00  0.0000000e+00  0.0000000e+00  1.4901161e-08
 -5.9604645e-08  0.0000000e+00  0.0000000e+00 -5.9604645e-08
 -3.2130629e-08  0.0000000e+00  0.0000000e+00  0.0000000e+00]
```

Tabrizian: Nice improvement!!

qa/L0_backend_python/python_unittest.py
Tabrizian: I think it is probably fine from leak detection but would significantly affect the test time since now we would run each test 5 times.

rmccorm4: Yeah I'll play with the numbers, just wanted to confirm the methodology

qa/L0_backend_python/decoupled/decoupled_test.py
```diff
@@ -64,11 +68,50 @@ def test_decoupled_execute_error(self):
         shape = [2, 2]
         number_of_requests = 2
         user_data = UserData()
-        with grpcclient.InferenceServerClient("localhost:8001") as triton_client:
-            triton_client.start_stream(callback=partial(callback, user_data))
+        with self._shm_leak_detector.Probe(debug_str=model_name) as shm_probe:
```
rmccorm4, line 71: This file looks like a lot of changes, but it's just adding probe context around the large block of each test and then indenting it.

qa/L0_backend_python/io/io_test.py
```diff
@@ -101,57 +100,56 @@ def _run_ensemble_test(self):
                             self.assertTrue(np.all(output0 == input0))
 
     def test_ensemble_io(self):
-        # Only run the shared memory leak detection with the default trial
```
rmccorm4, line 104: This was previously skipping decoupled trial - trying to add decoupled test back.

qa/L0_backend_python/decoupled/decoupled_test.py
```diff
@@ -79,215 +122,191 @@ def test_decoupled_execute_error(self):
...
-            result = user_data._completed_requests.get()
-            check_result(result)
+                result = user_data._completed_requests.get()
+                check_result(result)
 
     def test_decoupled_bls_stream(self):
         # Test combinations of BLS and decoupled API in Python backend.
         model_name = "decoupled_bls_stream"
         in_values = [4, 2, 0, 1]
-        shape = [1]
```
rmccorm4, line 138: this was unused

qa/L0_backend_python/io/io_test.py
```diff
@@ -101,57 +100,56 @@ def _run_ensemble_test(self):
...
-        if TRIAL == "default":
-            with self._shm_leak_detector.Probe():
-                self._run_ensemble_test()
-        else:
-            self._run_ensemble_test()
```
rmccorm4, line 109: This was previously skipping decoupled trial - trying to add decoupled test back.

qa/L0_backend_python/python_unittest.py
```diff
@@ -56,11 +56,15 @@ def _run_unittest(self, model_name):
     def test_python_unittest(self):
         model_name = os.environ["MODEL_NAME"]
 
-        # Warmup model, let it do any upfront initializations required
+        # Stress test memory growth test more on nightly runs
+        nightly = int(os.environ.get("TRITON_NIGHTLY", "0"))
```
rmccorm4, line 60: TODO: Pass TRITON_NIGHTLY env var to container in test template

qa/common/shm_util.py
```diff
@@ -448,7 +440,7 @@ def __exit__(self, type, value, traceback):
                 if current_shm_size != prev_shm_size:
                     shm_leak_detected = True
                     print(
-                        f"Shared memory leak detected: {current_shm_size} (current) != {prev_shm_size} (prev)."
+                        f"[{self._debug_str}] Shared memory leak detected: {current_shm_size} (current) != {prev_shm_size} (prev)."
```
rmccorm4, line 443: This is just to make it more immediately clear what subtest is leaking in job logs.

qa/L0_backend_python/python_unittest.py
kthui: @rmccorm4 I'm trying to understand you implementation here on warmup. The [existing implementation](https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/python_unittest.py#L60-L82) will do a warmup only if the model name matches `bls`, `bls_memory`, `bls_memory_async` or `bls_request_rescheduling`. A warmup is not performed for all other model names. On your implementation, it will perform a warmup for all model names. Am I understanding it correctly? Why do we want to perform a warmup for all models?

rmccorm4: I think my goal was just to standardize the test for every use case, rather than having special cases as much as possible. If it significantly impacts total test time, it's probably not a good idea. This is just something I was experimenting with at the time.

rmccorm4: If you have any similar thoughts but want to do it differently, feel free - I don't think my latest changes were passing CI when I was last looking into this.

kthui: Interestingly, I got it passed locally, after merging in main changes. I will push them here and run the CI, and see what happens.

qa/common/shm_util.py
```diff
@@ -434,12 +435,14 @@ def __exit__(self, type, value, traceback):
             for current_shm_size, prev_shm_size in zip(
                 current_shm_sizes, self._shm_region_free_sizes
             ):
-                if current_shm_size != prev_shm_size:
+                if current_shm_size > prev_shm_size:
```
kthui, line 438: @rmccorm4 Do you think the comparison should be `<` instead of `>`? I think the `*_shm_size` here is not referring to the occupied size, but the free size on the shm segment. If there is a leak, then `current_free_size < previous_free_size`.

qa/L0_backend_python/io/io_test.py
kthui: @Tabrizian Do you know why we were skipping shm leak detector probe on decoupled trial? It seems like the decoupled trial will always leak the exact same amount as @rmccorm4 found.

Tabrizian: Based on our slack conversation I think you found the root cause. And just to set the record straight it is not really a leak since the memory will be deallocated on next inference.

PR #6435, opened 2023-10-16
Extending restricted api to http

nnshah1:
Adding support for restricting access to http apis based on header key and value (similar to grpc restricted protocols)

- Refactored command line parsing to create a map of features to header key and values
- Added check to all Handle APIs

### DIFF
--- a/Dockerfile.win10.min
+++ b/Dockerfile.win10.min
@@ -97,7 +97,7 @@ RUN git clone --single-branch --depth=1 -b %VCPGK_VERSION% https://github.com/mi
 WORKDIR /vcpkg
 RUN bootstrap-vcpkg.bat
 RUN vcpkg.exe update
-RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
+RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows boost-core:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
 RUN vcpkg.exe integrate install
 
 LABEL VCPGK_VERSION=${VCPGK_VERSION}
--- a/Dockerfile.win10.min
+++ b/Dockerfile.win10.min
@@ -92,7 +92,7 @@ WORKDIR /
 #
 # Installing Vcpkg
 #
-ARG VCPGK_VERSION=2023.07.21
+ARG VCPGK_VERSION=2022.11.14
 RUN git clone --single-branch --depth=1 -b %VCPGK_VERSION% https://github.com/microsoft/vcpkg.git
 WORKDIR /vcpkg
 RUN bootstrap-vcpkg.bat
--- a/Dockerfile.win10.min
+++ b/Dockerfile.win10.min
@@ -97,7 +97,7 @@ RUN git clone --single-branch --depth=1 -b %VCPGK_VERSION% https://github.com/mi
 WORKDIR /vcpkg
 RUN bootstrap-vcpkg.bat
 RUN vcpkg.exe update
-RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows boost-core:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
+RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
 RUN vcpkg.exe integrate install
 
 LABEL VCPGK_VERSION=${VCPGK_VERSION}
--- a/qa/L0_http/test.sh
+++ b/qa/L0_http/test.sh
@@ -667,6 +667,49 @@ set -e
 kill $SERVER_PID
 wait $SERVER_PID
 
+### Test Restricted  APIs ###
+
+SERVER_ARGS="--model-repository=${MODELDIR} \
+             --http-restricted-api=model-repository,health:k1=v1 \
+             --http-restricted-api=metadata,health:k2=v2"
+run_server
+EXPECTED_MSG="api 'health' can not be specified in multiple config groups"
+if [ "$SERVER_PID" != "0" ]; then
+    echo -e "\n***\n*** Expect fail to start $SERVER\n***"
+    kill $SERVER_PID
+    wait $SERVER_PID
+    RET=1
+elif [ `grep -c "${EXPECTED_MSG}" ${SERVER_LOG}` != "1" ]; then
+    echo -e "\n***\n*** Failed. Expected ${EXPECTED_MSG} to be found in log\n***"
+    cat $SERVER_LOG
+    RET=1
+fi
+
+kill $SERVER_PID
+wait $SERVER_PID
+
+### Test Restricted  APIs ###
+
+SERVER_ARGS="--model-repository=${MODELDIR} \
+             --http-restricted-api=model-repository:admin-key=admin-value \
+             --http-restricted-api=inference,health:infer-key=infer-value"
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+set +e
+python $PYTHON_UNIT_TEST RestrictedProtocolTest > $CLIENT_LOG 2>&1
+if [ $? -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Python GRPC Restricted Protocol Test Failed\n***"
+    RET=1
+fi
+set -e
+kill $SERVER_PID
+wait $SERVER_PID
+
 ###
 
 if [ $RET -eq 0 ]; then
--- a/qa/L0_http/test.sh
+++ b/qa/L0_http/test.sh
@@ -44,6 +44,7 @@ RET=0
 
 CLIENT_PLUGIN_TEST="./http_client_plugin_test.py"
 BASIC_AUTH_TEST="./http_basic_auth_test.py"
+RESTRICTED_API_TEST="./http_restricted_api_test.py"
 NGINX_CONF="./nginx.conf"
 # On windows the paths invoked by the script (running in WSL) must use
 # /mnt/c when needed but the paths on the tritonserver command-line
@@ -588,13 +589,15 @@ kill $SERVER_PID
 wait $SERVER_PID
 
 # Run python unit test
-rm -r ${MODELDIR}/*
+MODELDIR=python_unit_test_models
+mkdir -p $MODELDIR
+rm -rf ${MODELDIR}/*
 cp -r $DATADIR/qa_identity_model_repository/onnx_zero_1_float32 ${MODELDIR}/.
 cp -r $DATADIR/qa_identity_model_repository/onnx_zero_1_object ${MODELDIR}/.
 cp -r $DATADIR/qa_identity_model_repository/onnx_zero_1_float16 ${MODELDIR}/.
 cp -r $DATADIR/qa_identity_model_repository/onnx_zero_3_float32 ${MODELDIR}/.
 cp -r ${MODELDIR}/onnx_zero_1_object ${MODELDIR}/onnx_zero_1_object_1_element && \
-    (cd models/onnx_zero_1_object_1_element && \
+    (cd $MODELDIR/onnx_zero_1_object_1_element && \
         sed -i "s/onnx_zero_1_object/onnx_zero_1_object_1_element/" config.pbtxt && \
         sed -i "0,/-1/{s/-1/1/}" config.pbtxt)
 
@@ -668,8 +671,8 @@ kill $SERVER_PID
 wait $SERVER_PID
 
 ### Test Restricted  APIs ###
-
-SERVER_ARGS="--model-repository=${MODELDIR} \
+MODELDIR="`pwd`/models"
+SERVER_ARGS="--model-repository=${MODELDIR}
              --http-restricted-api=model-repository,health:k1=v1 \
              --http-restricted-api=metadata,health:k2=v2"
 run_server
@@ -685,14 +688,11 @@ elif [ `grep -c "${EXPECTED_MSG}" ${SERVER_LOG}` != "1" ]; then
     RET=1
 fi
 
-kill $SERVER_PID
-wait $SERVER_PID
-
 ### Test Restricted  APIs ###
 
 SERVER_ARGS="--model-repository=${MODELDIR} \
              --http-restricted-api=model-repository:admin-key=admin-value \
-             --http-restricted-api=inference,health:infer-key=infer-value"
+             --http-restricted-api=inference:infer-key=infer-value"
 run_server
 if [ "$SERVER_PID" == "0" ]; then
     echo -e "\n***\n*** Failed to start $SERVER\n***"
@@ -700,10 +700,11 @@ if [ "$SERVER_PID" == "0" ]; then
     exit 1
 fi
 set +e
-python $PYTHON_UNIT_TEST RestrictedProtocolTest > $CLIENT_LOG 2>&1
+
+python $RESTRICTED_API_TEST RestrictedAPITest > $CLIENT_LOG 2>&1
 if [ $? -ne 0 ]; then
     cat $CLIENT_LOG
-    echo -e "\n***\n*** Python GRPC Restricted Protocol Test Failed\n***"
+    echo -e "\n***\n*** Python HTTP Restricted Protocol Test Failed\n***"
     RET=1
 fi
 set -e
--- a/qa/L0_http/test.sh
+++ b/qa/L0_http/test.sh
@@ -671,6 +671,8 @@ kill $SERVER_PID
 wait $SERVER_PID
 
 ### Test Restricted  APIs ###
+### Repeated API not allowed
+
 MODELDIR="`pwd`/models"
 SERVER_ARGS="--model-repository=${MODELDIR}
              --http-restricted-api=model-repository,health:k1=v1 \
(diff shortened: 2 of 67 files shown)


src/http_server.cc
```diff
@@ -1515,6 +1528,10 @@ HTTPAPIServer::HandleModelReady(
     evhtp_request_t* req, const std::string& model_name,
     const std::string& model_version_str)
 {
+  if (RestrictedAPI(req, "health")) {
```
nnshah1, line 1531: could move to a macro if it fits better

nv-kmcgill53: If this trend of adding checks to headers continues then we may want to think about a full pre-/post-request process implementation where users can opt-in to the checks they want done.

nv-kmcgill53 (commented):
Looks good for the most part. I think we do want to take 10 min and think about if we can sanely eliminate the check for the hot path.

src/common.h
```diff
@@ -170,4 +171,6 @@ bool Contains(const std::vector<std::string>& vec, const std::string& str);
 /// \return The joint string.
 std::string Join(const std::vector<std::string>& vec, const std::string& delim);
 
+typedef std::map<std::string, std::pair<std::string, std::string>>
```
GuanLuo, line 174: `using`?

GuanLuo (approved):

src/grpc/grpc_server.cc
```diff
@@ -252,9 +252,7 @@ class CommonHandler : public HandlerBase {
       TraceManager* trace_manager,
       inference::GRPCInferenceService::AsyncService* service,
       ::grpc::health::v1::Health::AsyncService* health_service,
-      ::grpc::ServerCompletionQueue* cq,
-      std::map<std::string, std::pair<std::string, std::string>>
-          restricted_keys);
+      ::grpc::ServerCompletionQueue* cq, RestrictedFeatureMap restricted_keys);
```
Tabrizian, line 255:
```suggestion
      ::grpc::ServerCompletionQueue* cq, RestrictedFeatureMap& restricted_keys);
```

qa/L0_http/test.sh
```diff
@@ -44,6 +44,7 @@ RET=0
 
 CLIENT_PLUGIN_TEST="./http_client_plugin_test.py"
 BASIC_AUTH_TEST="./http_basic_auth_test.py"
+RESTRICTED_API_TEST="./http_restricted_api_test.py"
```
Tabrizian, line 47: I think this file is not committed.

src/http_server.cc
```diff
@@ -1295,6 +1300,10 @@ void
 HTTPAPIServer::HandleRepositoryIndex(
     evhtp_request_t* req, const std::string& repository_name)
 {
+  if (RestrictedAPI(req, "model-repository")) {
+    return;
```
Tabrizian, line 1304: Does only returning from this function is sufficient or we also need to respond the request with error?

qa/L0_http/test.sh
nnshah1: thanks for catching!

src/http_server.cc
nnshah1: The function itself sends the response - agree the naming isn't the best.

Could move to a macro: RETURN_AND_RESPOND_IF_RESTRICTED

but that also seemed pretty verbose.

CHECK_RESTRICTED_AND_RESPOND

...

src/http_server.cc
nnshah1: For now I'd like to keep this to bringing parity with grpc restrictions - we could generalize later but I would want to make sure we have a use case in mind. reasonable?

The option itself is pretty generic - so can check any header name - but only allows one check per api type ...

We do other checks like method / model name / etc. -

src/http_server.cc
```diff
@@ -3411,6 +3465,10 @@ HTTPAPIServer::HandleInfer(
     evhtp_request_t* req, const std::string& model_name,
     const std::string& model_version_str)
 {
+  if (RestrictedAPI(req, "inference")) {
```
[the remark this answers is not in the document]
nv-kmcgill53: Follow up: I wonder if we can use perf analyzer to show what, if any, overhead might be here. i.e. Run a test with/without the inference api restricted.

src/http_server.cc
nv-kmcgill53: That's reasonable. I meant for this to be more of a comment on what we should consider for the future more than an actual change request.

src/http_server.cc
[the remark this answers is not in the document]
nv-kmcgill53: I think so. I don't think this has to be very verbose

src/http_server.cc
```diff
@@ -4487,16 +4546,38 @@ HTTPAPIServer::Create(
...
+HTTPAPIServer::RestrictedAPI(evhtp_request_t* req, const std::string api)
+{
+  auto restricted_api = restricted_apis_.find(api);
+  if (restricted_api == restricted_apis_.end()) {
+    return false;
+  }
+  auto header = restricted_api->second.first;
+  auto expected_value = restricted_api->second.second;
+  const char* actual_value = evhtp_kv_find(req->headers_in, header.c_str());
+  if ((actual_value == nullptr) || (actual_value != expected_value)) {
```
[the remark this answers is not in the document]
nnshah1: Currently the message mirrors what we provide for grpc.

I actually would provide less information (i.e. ideally the error message shouldn't leak any details about the header or the value) -

That is - the message should rather be "This API is restricted please see http-restricted-api documentation".

The fact that it is a header or a specific header name ideally we do not divulge ...

src/http_server.cc
```diff
@@ -4487,16 +4546,38 @@ HTTPAPIServer::Create(
...
   return nullptr;
 }
 
+bool
+HTTPAPIServer::RestrictedAPI(evhtp_request_t* req, const std::string api)
+{
+  auto restricted_api = restricted_apis_.find(api);
+  if (restricted_api == restricted_apis_.end()) {
+    return false;
+  }
```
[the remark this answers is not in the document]
nnshah1: There are a couple things I think we could do:

1) Have the table be a fixed array keyed on an enum  (the number of restricted apis is fixed and we don't need a map)
2) introduce a thread safe function level static to hold the value:

```
static const auto restricted_api_key_value = ..
  
if (restricted_api_key_value && Restricted()) {
  return;
}
```

Given the number of checks already in the hotpath - let me know what you think - I'm not opposed to it - but not sure if it would be meaningful.

src/http_server.cc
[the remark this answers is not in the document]
nv-kmcgill53: I was thinking in the literal sense that the message is, "header not found," where we don't divulge the header or the expected value but the user knows that the header isn't on the request and can troubleshoot that. Again, "wrong header value," would be just that message telling the user that the header was found but the value is wrong.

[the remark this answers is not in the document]
nnshah1: I think even that would leak info (i.e. that the got the header right but the value wrong) ...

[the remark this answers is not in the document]
nv-kmcgill53: Ya this is a balance between usability and obfuscation. I'll defer to you on this.

src/http_server.cc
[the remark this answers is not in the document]
nnshah1: will handle in seperate doc update

src/http_server.cc
```diff
@@ -1271,6 +1283,8 @@ HTTPAPIServer::InferResponseFree(
 void
 HTTPAPIServer::HandleServerHealth(evhtp_request_t* req, const std::string& kind)
 {
+  RETURN_AND_RESPOND_IF_RESTRICTED(req, "health");
```
rmccorm4, line 1286: Can the majority of these be checked via `kind` in top level handler? ex: https://github.com/triton-inference-server/server/blob/a2762710aa3f133ae8c1abcb998b4c09c8b02194/src/http_server.cc#L4439-L4442

We already determine the "key" via regexes in top level handler, so wonder if it can just be directly passed for each matched regex or something.

```cpp
  std::string model_name, version, kind;
  if (RE2::FullMatch(
          std::string(req->uri->path->full), model_regex_, &model_name,
          &version, &kind)) {
    // NEW
    RETURN_AND_RESPOND_IF_RESTRICTED(req, kind);  // kind in [infer, stats, metadata, tracing, ...]
    // EXISTING
    if (kind == "ready") {
    ...
```

May not be a 1:1 mapping for every single one, but might cover some subsets.

nnshah1: I looked at that briefly, but they don't seem to match that well - I think we'd have to create another map from kind to field. I can see moving to an enum - but I think matching in each function maybe the easiest to reason about. 

However, open to change. One thing I added in this go was static variables to only evaluate if an api is restricted once - might be harder to optimized in the top level handler ...

qa/L0_http/test.sh
```diff
@@ -667,6 +670,47 @@ set -e
 kill $SERVER_PID
 wait $SERVER_PID
 
+### Test Restricted  APIs ###
+MODELDIR="`pwd`/models"
+SERVER_ARGS="--model-repository=${MODELDIR}
+             --http-restricted-api=model-repository,health:k1=v1 \
+             --http-restricted-api=metadata,health:k2=v2"
```
rmccorm4, line 677: IIUC, we have a test with:
1. two groups, two endpoints per group, but health endpoint appears twice, so failure
2. two groups, one endpoint per group, success

Can you add a test like (1) but with success to test multiple groups sharing same key works OK?
3. two groups, two endpoints per group, all unique endpoints, success

src/http_server.cc
nnshah1: There seems to be limited overlap from the regex to the names .....

https://github.com/triton-inference-server/server/blob/a2762710aa3f133ae8c1abcb998b4c09c8b02194/docs/customization_guide/inference_protocols.md#limit-endpoint-access-beta

rmccorm4: I'm fine with current addition per function. An enum would be nice, but just a nice-to-have. Just wanted to see if this brought up any "lightbulbs" on current organization. Will resolve this thread.

src/http_server.cc
```diff
@@ -80,6 +80,17 @@ namespace triton { namespace server {
     return;                                         \
   } while (false)
 
+#define RETURN_AND_RESPOND_IF_RESTRICTED(REQ, API_KEY)                     \
+  do {                                                                     \
+    static auto const is_restricted_api = restricted_apis_.count(API_KEY); \
+    static auto const restricted_api = restricted_apis_.find(API_KEY);     \
```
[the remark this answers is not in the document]
nnshah1: static initialization is now thread safe and happens only once:

https://stackoverflow.com/a/8102145 

the restricted_apis_ map is initialized at start up and then is constant - so no race condition with count and find being out of sync.

Since it's static auto const - it will be initialized once and if is_restricted_api() is false it will be a quick exit (no lookup). 

My hope would be the execution would be smart enough to optimize out the check afterwards via branch prediction ...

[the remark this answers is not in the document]
nv-kmcgill53: What I am thinking about is once the program is running for a bit and we have two requests on two different APIs say, inference and model config, and both are checking to see if they are restricted. Is there a race condition in the storage of the `is_restricted_api` variable? Or is this per thread storage?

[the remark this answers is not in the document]
nnshah1: the static const variable is local to the function it is defined - so each function has its own copy that tracks jus that api...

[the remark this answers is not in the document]
nv-kmcgill53: 💡 Got it! Don't do PRs late at night is what I am learning. The macro defined the `static auto const` in each of the functions thus making the variable easily predicted since it's checking a const string. Sorry for throwing you through the ringer on this. This is better than I first thought.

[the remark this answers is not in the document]
nnshah1: NP, I am wondering now though if the ultimate would be a const local variable initialized from a fixed offset into a table .... I didn't realize the penalty of accessing a static ... may move to a fixed table because we don't need the flexibility of the map (we have a fixed set of categories to restrict)

https://stackoverflow.com/a/21033042

src/http_server.cc
```diff
@@ -80,6 +80,17 @@ namespace triton { namespace server {
     return;                                         \
   } while (false)
 
+#define RETURN_AND_RESPOND_IF_RESTRICTED(REQ, API_KEY)                     \
+  do {                                                                     \
+    static auto const is_restricted_api = restricted_apis_.count(API_KEY); \
```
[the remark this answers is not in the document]
nnshah1: we could pass in the map as well to the macro - the main benefit of the macro is to create the static const variables that only get initialized once - with the goal of removing any unnecessary checks in the flow if the user hasn't enabled any restrictions. 

let me think if there is a simpler way to optimize.

What would you think about removing the map and using a fixed table based on an enumeration? (the idea is that the number of categories is fixed at compile time and we don't need to have an arbitrary map of categories)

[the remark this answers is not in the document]
Tabrizian: > What would you think about removing the map and using a fixed table based on an enumeration? (the idea is that the number of categories is fixed at compile time and we don't need to have an arbitrary map of categories)

can you elaborate more on this?

nv-kmcgill53 (dismissed):

src/http_server.cc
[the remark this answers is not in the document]
GuanLuo: template with non-type template parameter?

qa/L0_http/test.sh
nnshah1: added support for tests to match grpc tests - covers two categories with a single header value.

added test for unknown category

all restrictions - deem that out of scope for this PR - but good for future coverage

src/command_line_parser.cc
```diff
@@ -463,6 +462,14 @@ TritonParser::SetupOptions()
...
       {OPTION_HTTP_THREAD_COUNT, "http-thread-count", Option::ArgInt,
        "Number of threads handling HTTP requests."});
+  http_options_.push_back(
+      {OPTION_HTTP_RESTRICTED_API, "http-restricted-api",
+       "<string>:<string>=<string>",
+       "Specify restricted HTTP api setting. The format of this "
+       "flag is --http-restricted-api=<apis>,<key>=<value>. Where "
+       "<api> is a comma-separated list of apis to be restricted. "
+       "<key> will be additional header key to be checked when a HTTP request "
+       "is received, and <value> is the value expected to be matched."});
```
rmccorm4, line 472: What do you think of including options here to avoid docs lookup?
```cpp
...
"Possible options are: " + Join(RESTRICTED_CATEGORY_NAMES, " ")
```

Same for grpc help text.

(I think I recently added a Join() helper to common.h)

src/restricted_features.h
```diff
@@ -0,0 +1,113 @@
...
+  INVALID,
+  CATEGORY_COUNT = INVALID
+};
+
+/// Restricted Category Names
+const std::array<const std::string, RestrictedCategory::CATEGORY_COUNT>
+    RESTRICTED_CATEGORY_NAMES{
+        "health",        "metadata",     "inference",
+        "shared-memory", "model-config", "model-repository",
+        "statistics",    "trace",        "logging"};
```
rmccorm4, line 56: Is an "invalid" needed here at the end for completeness or returning error strings anywhere?

nnshah1: used for convenience to indicate when a string is not found in the category names -

Tabrizian (approved):

src/restricted_features.h
```diff
@@ -0,0 +1,113 @@
...
+  METADATA,
+  INFERENCE,
+  SHARED_MEMORY,
+  MODEL_CONFIG,
+  MODEL_REPOSITORY,
+  STATISTICS,
+  TRACE,
+  LOGGING,
+  INVALID,
+  CATEGORY_COUNT = INVALID
```
nv-kmcgill53, line 48: Idle comment, no action required: I normally see the error condition at the beginning of the enum list so that the value is forward ABI compatible as new enums are added. I assume you have a reason for putting the error at the end of the list.

nnshah1: end in this case is to make ToCategory() work easily for items that aren't found, end() - begin() == size

nv-kmcgill53 (approved):

nv-kmcgill53 (approved):

src/common.h
```diff
@@ -168,6 +169,19 @@ bool Contains(const std::vector<std::string>& vec, const std::string& str);
 /// \param vec The vector of strings to join.
 /// \param delim The delimiter to join with.
 /// \return The joint string.
-std::string Join(const std::vector<std::string>& vec, const std::string& delim);
+
+// std::string Join(const std::vector<std::string>& vec, const std::string&
```
rmccorm4: Update/remove extra doc strings and commented out functions here

rmccorm4: Oh nvm my view was out of sync

src/command_line_parser.cc
```diff
@@ -1904,57 +1919,64 @@ TritonParser::ParseBackendConfigOption(const std::string& arg)
   return {name_string, setting_string, value_string};
 }
 
-std::tuple<std::string, std::string, std::string>
-TritonParser::ParseGrpcRestrictedProtocolOption(const std::string& arg)
+void
+TritonParser::ParseRestrictedFeatureOption(
+    const std::string& arg, const std::string& option_name,
+    const std::string& key_prefix, const std::string& feature_type,
```
rmccorm4, line 1925: nit: it's called `feature_type` in .cc and `feature_name` in .h -- let's make them match for parsing code

src/command_line_parser.cc
```diff
@@ -1398,16 +1416,11 @@ TritonParser::Parse(int argc, char** argv)
...
-          for (const auto& p : protocols) {
-            pg.protocols_.emplace(p);
-          }
-          pg.restricted_key_ = std::make_pair(key, value);
-          lgrpc_options.protocol_groups_.emplace_back(pg);
+          ParseRestrictedFeatureOption(
+              optarg, long_options[option_index].name,
+              std::string(
+                  triton::server::grpc::kRestrictedProtocolHeaderTemplate),
+              "protocol", lgrpc_options.restricted_protocols_);
```
rmccorm4, line 1423: 1. Why is feature_name "protocol" for grpc and "api" for http? 
2. Can http_options and grpc_options have the same member variable name, either `.restricted_protocols_` or `.restricted_apis_` rather than being different?

nnshah1: Mainly just to align with the command line setting / documentation. gRPC specifically calls out protocols throughout and that didn't make sense for HTTP - where the closest seemed to be API. 

could make them both restricted_features_ if clearer -

nnshah1: Oh sorry, I wouldn't to change the `feature_type` from "protocol" as its already part of the existing docs / help messages. So for printing / documentation I think keeping it "protocol" and "api" makes sense. 

For the internal variable I think it matters less.

rmccorm4: both restricted_features_ LGTM

nnshah1: deferring name change to documentation PR / larger discussion on if it's ok to change from grpc protocol

rmccorm4 (approved):

rmccorm4 (approved):

Dockerfile.win10.min
```diff
@@ -97,7 +97,7 @@ RUN git clone --single-branch --depth=1 -b %VCPGK_VERSION% https://github.com/mi
 WORKDIR /vcpkg
 RUN bootstrap-vcpkg.bat
 RUN vcpkg.exe update
-RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
+RUN vcpkg.exe install openssl:x64-windows openssl-windows:x64-windows rapidjson:x64-windows re2:x64-windows boost-interprocess:x64-windows boost-stacktrace:x64-windows boost-core:x64-windows zlib:x64-windows pthread:x64-windows b64:x64-windows
```
nnshah1, line 100: @nv-kmcgill53 @mc-nv 

I believe I caught an intermediate windows build state that caused this - but that it shouldn't technically be required - as it should be a dependency already 

let me know if we should remove it or keep it -

nv-kmcgill53: I don't see a problem listing it here.

rmccorm4 (approved):

PR #6436, opened 2023-10-16
Adjust test_concurrent_same_model_load_unload_stress

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/276

Increase concurrency due to reduced windows size of replicating a load while async unloading.

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2919,7 +2919,7 @@ class LifeCycleTest(tu.TestResultCollector):
 
     def test_concurrent_same_model_load_unload_stress(self):
         model_name = "identity_zero_1_int32"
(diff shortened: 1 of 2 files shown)


rmccorm4:
Is this change necessary? Shouldn't 16 threads be sufficient to repro? or was this more reliable in your testing?

kthui:
> Is this change necessary? Shouldn't 16 threads be sufficient to repro? or was this more reliable in your testing?

With 16 thread, I was only able to reproduce around 1/3 runs locally, after the change in core. With 32 threads, I was able to reproduce 100% locally.

I think we might get away with 24 threads, probably less reliable, do you want to try it instead?

rmccorm4:
32 is fine - just wanted to understand the motivation. LGTM.

rmccorm4 (approved):

PR #6454, opened 2023-10-19
Adding emergency vllm latest release

oandreeva-nv:
vLLM's latest `emergency` release

### DIFF
--- a/build.py
+++ b/build.py
@@ -78,7 +78,7 @@ TRITON_VERSION_MAP = {
         "2023.0.0",  # Standalone OpenVINO
         "2.4.7",  # DCGM version
         "py310_23.1.0-1",  # Conda version
-        "0.2.1",  # vLLM version
+        "0.2.1.post1",  # vLLM version
     )
 }
 


dyastremsky (approved):

build.py
```diff
@@ -78,7 +78,7 @@
         "2023.0.0",  # Standalone OpenVINO
         "2.4.7",  # DCGM version
         "py310_23.1.0-1",  # Conda version
-        "0.2.1",  # vLLM version
+        "0.2.1.post1",  # vLLM version
```
mc-nv, line 81: I thought that the pip resolves `post` as an newer version of the package and this value is not need to be updated unless we wan't to stick with a specific one. 
Further more: 
```
docker run --rm -it python:3 bash
root@2f294ce14ad7:/# pip3 install vllm
Collecting vllm
  Downloading vllm-0.2.1.post1.tar.gz (132 kB)
```
or 
```
$ docker run --rm -it python:3 bash
root@36cb7753add5:/# pip install vllm==0.2.1
Collecting vllm==0.2.1
  Downloading vllm-0.2.1.tar.gz (132 kB)
  ```

oandreeva-nv: Unfortunately, we do specify version for `pip install vllm` [here](https://github.com/triton-inference-server/server/blob/328d51940c3402f3594bcab9c2c6155740cf043f/build.py#L1371), so it will not install `post1` release. 

And we specify the release version (I believe) for consistency (with tests especially)

PR #6466, opened 2023-10-23
Add Python-based backends CI

pskiran1:
Changes:

Load all the models listed below and test their functionality:
- 2 vLLM backend models
- 2 add_sub backend (Python-based) models
- 1 Python backend model
- 1 Pytorch backend model.


qa/python_models/python_based_backends/vllm_opt/model.json
```diff
@@ -0,0 +1,5 @@
+{
+    "model": "facebook/opt-125m",
+    "disable_log_requests": "true",
+    "gpu_memory_utilization": 0.25
+}
```
oandreeva-nv, line 5: can we copy this file and `config` from vllm_backend repo `samples` folder? This way we can eliminate the necessity to maintain the `vllm_opt` repo here

pskiran1: Updated test.sh to use vllm_backend repo samples.

dyastremsky (dismissed):

tanmayv25:
Do we need to test against vLLM backend model here? Isn't `2 add_sub backend (Python-based) models` good proxy for vllm backend models?
Otherwise you would need to use vLLM backend installed to run these tests.

oandreeva-nv:
I am currently confused why we have `json` model files for pytorch Python-based backend. I believe the whole point of PyTorch Python-based backend. is to help serve `model.py` which define `torch.nn.Module` base class. Or what is '2 add_sub backend (Python-based) models' mean ? Could you please clarify my concern?

pskiran1:
> Do we need to test against vLLM backend model here? Isn't `2 add_sub backend (Python-based) models` good proxy for vllm backend models? Otherwise you would need to use vLLM backend installed to run these tests.

@tanmayv25, Yes, I think `2 add_sub backend (Python-based) models` covers python-based backend testing. Would you recommend removing vLLM model testing here?

qa/common/gen_qa_pytorch_model.py
```diff
@@ -0,0 +1,124 @@
...
+        (torch.rand(1, 4, dtype=torch.float), torch.rand(1, 4, dtype=torch.float)),
+    )
+
+    os.makedirs(model_dir, exist_ok=True)
+    model_path = os.path.join(model_dir, "model.pt")
+
+    traced_model.save(model_path)
+
+
+def generate_config(config_path):
```
oandreeva-nv, line 57: nit: I believe add_sub's config is similar to what this function is generating. Maybe we can re-use this function to generate add_sub config as well? Maybe as an input it can take backend's name and dynamically adjust produced `config` .

pskiran1: @oandreeva-nv, just for confirmation, would recommend to generate [qa/python_models/python_based_backends/add_sub_backend/config.pbtxt](https://github.com/triton-inference-server/server/pull/6466/files#diff-85437ebf20c6bf63c7225f32423b65dfa4056538b204cf6d4a7e0fc10a6ba6fa) using this function?

qa/python_models/python_based_backends/add_sub_backend/model.py
```diff
@@ -0,0 +1,121 @@
...
+          be the same as `requests`
+        """
+
+        responses = []
+
+        for request in requests:
+            in_0 = pb_utils.get_input_tensor_by_name(request, "INPUT0")
+            in_1 = pb_utils.get_input_tensor_by_name(request, "INPUT1")
+
+            if self.add_sub_config["operation"] == "add":
```
oandreeva-nv, line 105: I believe we can eliminate dependency on single output [here](https://github.com/triton-inference-server/server/pull/6466/files#diff-c02333b955665a5e56699cccc5cce927d0c8ec9ac1fe6a783dc2fd3cc8a3e802R57), if we make this model to output both add tensor and sub tenor. In this case we can also shorten our `initialize` function to get rid of some checks and `add_model.json` and `sub_model.json` . CC @tanmayv25  what do you think?

qa/common/gen_qa_pytorch_model.py
oandreeva-nv: @pskiran1 How about we add inputs and outputs through [auto_complete_config](https://github.com/triton-inference-server/python_backend#auto_complete_config) for add_sub backend model then? In this case, we can make this function conditional, e.g.  `generate_config(config_path, backend_name, add_io = True)`. If `add_io` is false, it will only generate " backend: "{}"".format(backend_name)" something like this. Could you please try it out?

qa/python_models/python_based_backends/add_sub_backend/model.py
oandreeva-nv: I would also suggest adding second version to at least one model. It can be the same e.g. pytorch model, just in a separate version folder

tanmayv25: The reason for "operation" field in the add_sub json model is to test whether two models belonging to the same python based backend can be loaded and served. In order to clean up the dependency of single output, we can do the following:

1. variant1.json model: returns sum in OUTPUT0 and difference in OUTPUT1.
2. variant2.json model: returns difference in OUTPUT0 and sum in OUTPUT1.

What are your thoughts?

qa/common/gen_qa_pytorch_model.py
tanmayv25: Using auto_complete_config is a good idea here!

qa/python_models/python_based_backends/add_sub_backend/model.py
oandreeva-nv: I am fine with keeping this as is, if this is too much work. I thought we can also add some versions to some models

pskiran1: @oandreeva-nv @tanmayv25,
Thank you. The above changes may still require some conditional checks in the model.py. Therefore, proceeding with the existing approach. Updated the add_sub backend model.py to prepare configuration using the auto-config feature and removed a separate config.pbtxt file.
In addition, removed add_model.json, and sub_model.json files and creating them using a simple echo command in the test.sh file.
```
echo '{ "operation": "add" }' > ${MODEL_REPOSITORY}/add/1/model.json
```

Also, added multiple versions for the add_sub backend (add) and python backend models, along with appropriate tests to ensure expected behavior.

qa/common/gen_qa_pytorch_model.py
pskiran1: Updated the add_sub backend model.py to prepare inputs/outputs configuration using the `auto-complete-config` feature and removed a separate config.pbtxt file. For simplicity, generating minimal config.pbtxt in test.sh using command,
`echo "backend: \"add_sub\"" > ${MODEL_REPOSITORY}/add/config.pbtxt`

qa/L0_backend_python/python_based_backends/python_based_backends_test.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def test_python_model(self):
+        self.assertTrue(
+            self.triton_client.is_model_ready(self.python_model, model_version="2")
+        )
+        self._test_add_sub_model(
+            model_name=self.python_model, shape=[16], model_version="2"
+        )
+
+    def test_pytorh_model(self):
```
tanmayv25, line 81:
```suggestion
    def test_pytorch_model(self):
```

tanmayv25 (dismissed):

dyastremsky (approved):

PR #6472, opened 2023-10-23
L0_model_config Fix

fpetrini15:
Minor fix to the L0_model_config test suite.
Core changes: https://github.com/triton-inference-server/core/pull/278

### DIFF
--- a/qa/L0_model_config/autofill_noplatform/custom/unknown_backend.unknown/expected
+++ b/qa/L0_model_config/autofill_noplatform/custom/unknown_backend.unknown/expected
@@ -1,1 +1,1 @@
-Invalid argument: unable to find 'libtriton_unknown.so' for model 'unknown_backend.unknown'
+Invalid argument: unable to find 'unknown/model.py' for model 'unknown_backend.unknown'
--- a/qa/L0_model_config/autofill_noplatform/custom/unknown_backend.unknown/expected
+++ b/qa/L0_model_config/autofill_noplatform/custom/unknown_backend.unknown/expected
@@ -1,1 +1,1 @@
-Invalid argument: unable to find 'unknown/model.py' for model 'unknown_backend.unknown'
+Invalid argument: unable to find 'libtriton_unknown.so' or 'unknown/model.py' for model 'unknown_backend.unknown'


tanmayv25:
We should fix server core to output error as `unable to find 'libtriton_unknown.so' or 'unknown/model.py' for model 'unknown_backend.unknown'`

krishung5 (approved):

qa/L0_model_config/autofill_noplatform/custom/unknown_backend.unknown/expected
```diff
@@ -1 +1 @@
-Invalid argument: unable to find 'libtriton_unknown.so' for model 'unknown_backend.unknown'
+Invalid argument: unable to find 'libtriton_unknown.so' or 'unknown/model.py' for model 'unknown_backend.unknown'
```
jbkyang-nvi, line 1: nit: space

fpetrini15: Is this in reference to the lack of a newline or is there a space issue I am not seeing here?

jbkyang-nvi: The lack of newline. Sorry for being unclear

fpetrini15: No worries! I didn't use a newline because the previous version did not have one and figured the comparison in bash would be sensitive to this change. Taking a closer look in the repo it seems like we are inconsistent in our use of a newline. I know our bash files should end with one. Do we try to apply this principle to all files?

jbkyang-nvi: The `newline` check should have fixed this. I think it just doesn't check for this sort of file. I saw the previous file does have an extra line in it but your change deleted that. I think it would be good to change that though (although not a urgent or strong preference)

fpetrini15: I'll make a note to include this in my next server-based PR that way I don't have to create and new PR just for this.

PR #6484, opened 2023-10-26
Update `main` post-23.10 release  

mc-nv:


kthui (approved):

build.py
```diff
@@ -72,12 +72,14 @@
...
         "23.11dev",  # triton container
-        "23.09",  # upstream container
+        "23.10",  # upstream container
         "1.16.0",  # ORT
         "2023.0.0",  # ORT OpenVINO
         "2023.0.0",  # Standalone OpenVINO
         "2.4.7",  # DCGM version
         "py310_23.1.0-1",  # Conda version
+        "9.1.0.1",  # TRT version for building TRT-LLM backend
+        "12.2",  # CUDA version for building TRT-LLM backend
```
krishung5, line 82: I think these were added by mistake when resolving merge conflicts. Will need to remove these two line.

krishung5 (changes requested):

build.py
mc-nv: Addressed

build.py
```diff
@@ -72,13 +72,13 @@
...
         "23.11dev",  # triton container
-        "23.09",  # upstream container
+        "23.10",  # upstream container
         "1.16.0",  # ORT
         "2023.0.0",  # ORT OpenVINO
         "2023.0.0",  # Standalone OpenVINO
         "2.4.7",  # DCGM version
         "py310_23.1.0-1",  # Conda version
-        "0.2.1.post1",  # vLLM version
+        "9.1.0.1",  # TRT version for building TRT-LLM backend
```
krishung5, line 81:
```suggestion
        "0.2.1.post1", # vLLM version
```
It's the vLLm version that we need to keep, and remove the TRT-LLM related ones.

build.py
```diff
@@ -1357,9 +1357,11 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
     if "vllm" in backends:
         # [DLIS-5606] Build Conda environment for vLLM backend
         # Remove Pip install once vLLM backend moves to Conda environment.
+        # [DLIS-5650] Pre-installing torch 2.0.1, since vllm 0.2
+        # requires torch >= 2.0.0, but it doesn't work with torch 2.1.0.
         df += """
 # vLLM needed for vLLM backend
-RUN pip3 install vllm=={}
+RUN pip3 install torch==2.0.1 vllm=={}
 """.format(
```
jbkyang-nvi, line 1365: Should we put pinned torch version in the Triton version map?

qa/common/gen_jetson_trt_models
```diff
@@ -34,9 +34,7 @@
 # Make all generated files accessible outside of container
 umask 0000
 
-# Set the version of the models
-TRITON_VERSION=${TRITON_VERSION:=23.09}
-# Set the CUDA device to use
+TRITON_VERSION=${TRITON_VERSION:=23.10}
```
jbkyang-nvi, line 37: did we want to remove the comments as well?

jbkyang-nvi (commented):
Should we also change client repo as well?

build.py
mc-nv: I believe it's by a reason. 
You can confirm it with @oandreeva-nv

krishung5 (approved):

build.py
mc-nv: @krishung5 - please review again

build.py
oandreeva-nv: We actually don't need this `torch==2.0.1` anymore since L81 already has newer version of vLLM, than the one used in 23.10 release. To simplify current process, we can merge it as is and I can follow up and remove this particular commit

build.py
```diff
@@ -1357,9 +1357,11 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
     if "vllm" in backends:
         # [DLIS-5606] Build Conda environment for vLLM backend
         # Remove Pip install once vLLM backend moves to Conda environment.
+        # [DLIS-5650] Pre-installing torch 2.0.1, since vllm 0.2
+        # requires torch >= 2.0.0, but it doesn't work with torch 2.1.0.
```
oandreeva-nv, line 1361:
```suggestion

```

build.py
```diff
@@ -1357,9 +1357,11 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
...
         df += """
 # vLLM needed for vLLM backend
-RUN pip3 install vllm=={}
+RUN pip3 install torch==2.0.1 vllm=={}
```
oandreeva-nv, line 1364:
```suggestion
RUN pip3 install vllm=={}
```

build.py
oandreeva-nv: or, if possible, just commit suggestions

mc-nv: I don't see a suggestion...

krishung5 (approved):

PR #6490, opened 2023-10-27
Improve L0_backend_python test stability (ensemble / gpu_tensor_lifecycle)

rmccorm4:
- Increase float comparison tolerance for ensemble test, reduce failures due to float noise
- Increase stability on gpu_tensor_lifecycle test by comparing directly with torch allocated memory. Global gpu memory can be noisy with other processes using it, and that testing can be deferred to more holistic gpu memory growth tests. I left in some logging of total gpu memory usage to aid future debugging if needed.


dyastremsky (approved):
Nice work!

Tabrizian (approved):
nice improvements!

krishung5 (approved):
Nice work, Ryan!

rmccorm4:
Ran L0_backend_python 5 times and none failed due to either of these subtests, so seems like an improvement. Will monitor future tests for a bit.

PR #6503, opened 2023-11-01
Downgrade vcpkg version to WAR RE2 library inconsistency

GuanLuo:
Windows build with newer vcpkg results in HTTP server not properly parse URL with regex (provided by RE2), combining with the segfault from GRPC server suggests that a different version of the library is dynamically linked. GRPC itself brings an older version of RE2 and builds against that, while HTTP server links a newer RE2 shared library installed by vcpkg.

This PR is WAR to use a different but compatible library at runtime, the long-term fix is to isolate the dependency between HTTP and GRPC so they may use different version of the library.

### DIFF
--- a/Dockerfile.win10.min
+++ b/Dockerfile.win10.min
@@ -92,7 +92,7 @@ WORKDIR /
 #
 # Installing Vcpkg
 #
-ARG VCPGK_VERSION=2023.07.21
+ARG VCPGK_VERSION=2022.11.14
 RUN git clone --single-branch --depth=1 -b %VCPGK_VERSION% https://github.com/microsoft/vcpkg.git
 WORKDIR /vcpkg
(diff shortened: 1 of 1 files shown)


mc-nv (approved):

nv-kmcgill53:
> the long-term fix is to isolate the dependency between HTTP and GRPC so they may use different version of the library.

Can you make a ticket and put it in my backlog. Thanks!

nv-kmcgill53 (approved):

GuanLuo:
@nv-kmcgill53 DLIS-5748

nnshah1:
we faced this previously as well and had a solution to change grpc to use the system version adding here for reference:

https://github.com/triton-inference-server/third_party/pull/41

PR #6509, opened 2023-11-02
Add testing for Python backend request rescheduling

krishung5:
Python backend: https://github.com/triton-inference-server/python_backend/pull/319


krishung5:
Added grpc endpoint test with non-decoupled model based on offline discussion with @Tabrizian.

qa/L0_backend_python/request_rescheduling/grpc_endpoint_test.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def callback(user_data, result, error):
+    if error:
+        user_data._completed_requests.put(error)
+    else:
+        user_data._completed_requests.put(result)
+
+
+class GRPCENDPOINTTest(tu.TestResultCollector):
```
Tabrizian, line 55:
```suggestion
class GrpcEndpointTest(tu.TestResultCollector):
```

qa/python_models/bls_request_rescheduling/model.py
```diff
@@ -0,0 +1,133 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import time
+import unittest
+
+import numpy as np
+import triton_python_backend_utils as pb_utils
+
+
+class REQUESTRESCHEDULINGTest(unittest.TestCase):
```
Tabrizian, line 34:
```suggestion
class RequestReschedulingTest(unittest.TestCase):
```

Tabrizian (commented):
Nice testing! Some minor comments

qa/python_models/bls/model.py
```diff
@@ -275,7 +275,7 @@ def _send_bls_sequence_requests(self, correlation_id, is_decoupled):
             else:
                 infer_response = infer_request.exec()
 
-            self.assertFalse(infer_response.has_error(), infer_response.error())
+            self.assertFalse(infer_response.has_error())
```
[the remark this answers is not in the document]
krishung5: The correct function should be `infer_response.error().message()` to print out the error string. However, it will throw an error when there is no error object with infer_response if we call `infer_response.error().message()` here, so removing these.

Tabrizian (approved):

Issue #6481, opened 2023-10-25
reuse-http-port flag is integer and not boolean


VictorM-PS:
**Description**
The command feature `reuse-http-port` to merge several triton servers in a single port is indicated boolean in the help flag, however it is actually implemented as integer.

**Triton Information**
What version of Triton are you using?
23.04

Are you using the Triton container or did you build it yourself?
Triton container with fastertransformer_backend.

**To Reproduce**

```
root@ddf7471aa162:/workspace# /opt/tritonserver/bin/tritonserver --allow-http true --model-repository=./models/ner-model/  --load-model=*  --allow-http true --http-port 5023 --reuse-http-port true
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
Aborted (core dumped)
```

When running it with 1:

```
root@ddf7471aa162:/workspace# /opt/tritonserver/bin/tritonserver --allow-http true --model-repository=./models/ner-model/  --load-model=*  --allow-http true --http-port 5023 --reuse-http-port=1
I1025 14:02:19.135604 307 pinned_memory_manager.cc:240] Pinned memory pool is created at '0x7fa262000000' with size 268435456
I1025 14:02:19.139103 307 cuda_memory_manager.cc:105] CUDA memory pool is created on device 0 with size 67108864
...
```

Your test files cover it with the 1 flag but `-h` documentation is wrong/not clear as it indicates boolean true/false.

**Expected behavior**
The command line should execute with `--reuse-http-port true`.

dyastremsky:
Thank you for catching and reporting this bug, as well as detailing how our test files cover it. We have filed a ticket (DLIS-5697) and will fix this.

dyastremsky:
Thanks for letting us know about this! Closed with #6511.

PR #6511, opened 2023-11-02
Parse reuse-grpc-port and reuse-http-port as booleans

dyastremsky:
The reuse-grpc-port and reuse-http-port flags are documented in the options as booleans but parsed as integers. This PR makes it so they are parsed as booleans.

Addresses [6481](https://github.com/triton-inference-server/server/issues/6481).

### DIFF
--- a/qa/L0_socket/test.sh
+++ b/qa/L0_socket/test.sh
@@ -310,12 +310,12 @@ SERVER2_LOG="./inference_server2.log"
 for p in http grpc; do
     # error if servers bind to the same http/grpc port without setting the reuse flag
     if [ "$p" == "http" ]; then
-        SERVER_ARGS="--model-repository=$DATADIR --metrics-port 8002 --reuse-grpc-port=1"
-        SERVER0_ARGS="--model-repository=$DATADIR --metrics-port 8003 --reuse-grpc-port=1"
+        SERVER_ARGS="--model-repository=$DATADIR --metrics-port 8002 --reuse-grpc-port=true"
+        SERVER0_ARGS="--model-repository=$DATADIR --metrics-port 8003 --reuse-grpc-port=true"
         SERVER1_ARGS="--model-repository=$DATADIR --metrics-port 8004 --reuse-grpc-port=true"
     else
-        SERVER_ARGS="--model-repository=$DATADIR --metrics-port 8002 --reuse-http-port=1"
-        SERVER0_ARGS="--model-repository=$DATADIR --metrics-port 8003 --reuse-http-port=1"
+        SERVER_ARGS="--model-repository=$DATADIR --metrics-port 8002 --reuse-http-port=true"
+        SERVER0_ARGS="--model-repository=$DATADIR --metrics-port 8003 --reuse-http-port=true"
         SERVER1_ARGS="--model-repository=$DATADIR --metrics-port 8004 --reuse-http-port=true"
     fi
     # make sure the first server is launched successfully, then run the other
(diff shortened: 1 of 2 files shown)


qa/L0_socket/test.sh
```diff
@@ -357,9 +357,9 @@ for p in http grpc; do
     #   (a) Test default metrics-address being same as http-address
     #   (b) Test setting metrics-address explicitly to 0.0.0.0
     #   (c) Test setting metrics-address explicitly to 127.0.0.1
-    SERVER0_ARGS="--model-repository=$DATADIR --metrics-port 8002 --reuse-http-port=1 --reuse-grpc-port=1"
-    SERVER1_ARGS="--model-repository=$DATADIR --metrics-address 0.0.0.0 --metrics-port 8003 --reuse-http-port=1 --reuse-grpc-port=1"
-    SERVER2_ARGS="--model-repository=$DATADIR --metrics-address 127.0.0.2 --metrics-port 8004 --reuse-http-port=1 --reuse-grpc-port=1"
+    SERVER0_ARGS="--model-repository=$DATADIR --metrics-port 8002 --reuse-http-port=true --reuse-grpc-port=true"
+    SERVER1_ARGS="--model-repository=$DATADIR --metrics-address 0.0.0.0 --metrics-port 8003 --reuse-http-port=1 --reuse-grpc-port=true"
```
fpetrini15, line 361:
```suggestion
    SERVER1_ARGS="--model-repository=$DATADIR --metrics-address 0.0.0.0 --metrics-port 8003 --reuse-http-port=true --reuse-grpc-port=true"
```

qa/L0_socket/test.sh
```diff
@@ -357,9 +357,9 @@ for p in http grpc; do
...
+    SERVER2_ARGS="--model-repository=$DATADIR --metrics-address 127.0.0.2 --metrics-port 8004 --reuse-http-port=1 --reuse-grpc-port=true"
```
fpetrini15, line 362:
```suggestion
    SERVER2_ARGS="--model-repository=$DATADIR --metrics-address 127.0.0.2 --metrics-port 8004 --reuse-http-port=true --reuse-grpc-port=true"
```

kthui:
I wonder what you think about changing the docs instead? In case people are supplying an `int` instead of `bool` in the real world?

dyastremsky:
> I wonder what you think about changing the docs instead? In case other people are silently supplying an `int` instead of `bool` in the real world?

For context, the docs are the ones [here](https://github.com/triton-inference-server/server/blob/e865b47d9e9e1beaca12c51fd6b6ab925bc8cd96/src/command_line_parser.cc#L453) (in the parser itself).

As far as real world usage, as you can see from the fact that the test was passing before the two commits Francesco just added that this should work with values 0 and 1 because they are accepted by [ParseOption](https://github.com/triton-inference-server/server/blob/e865b47d9e9e1beaca12c51fd6b6ab925bc8cd96/src/command_line_parser.cc#L168-L184). Beyond that, I think previously any non-0 was accepted as "true" and now any non-zero integer other than 1 would throw an error.

This is a good point, so CC'ing @nnshah1 and @GuanLuo for their thoughts.

kthui (approved):
>  I think previously any non-0 was accepted as "true" and now any non-zero integer other than 1 would throw an error.

Thanks for clarifying, I don't know how many people could fall into this "trap", other than that, LGTM

nnshah1:
Good with the change - we can release note it to give people a heads up.

dyastremsky:
Awesome, thank you Jacky and Neelay! I'll add it to the release notes.

PR #6516, opened 2023-11-03
Test file override outside of model directory

rmccorm4:


build.py
```diff
@@ -1257,6 +1258,14 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
             {backend_dependencies} && \
     rm -rf /var/lib/apt/lists/*
 
+# Install boost version >= 1.78 for boost::span
```
rmccorm4, line 1261: Note this duplication is to have the built boost shared libraries (libboost_filesystem*.so) inside the final container as well.

nnshah1 (approved):

Tabrizian (approved):

PR #6518, opened 2023-11-04
Bring back Python backend based PyTorch backend

kthui:
Related PRs:
* https://github.com/triton-inference-server/core/pull/285
* https://github.com/triton-inference-server/common/pull/103
* https://github.com/triton-inference-server/backend/pull/94
* https://github.com/triton-inference-server/pytorch_backend/pull/117

Minor changes to L0_model_config due to adding "runtime" field into model configuration.
Add L0_pytorch_python_runtime test for the Python backend based PyTorch backend.


qa/L0_model_config/autofill_noplatform_success/custom/no_backend.identity/expected
```diff
@@ -31,3 +31,4 @@ output_pinned_memory {
 }
 }
 backend: "identity"
+runtime: ""
```
rmccorm4, line 34: Why does `onnxruntime` expect `libtriton_onnxruntime.so`, but `identity` doesn't expect `libtriton_identity.so`?

kthui: It is because autocomplete can only fill in "known" backends, while onnx is a "known" backend and identity is a custom backend.
See: https://github.com/triton-inference-server/core/pull/285/files#diff-f947d03afcf01249667bedc5b955404da083564c22809b6666ba923356f7c829R977
Therefore, the runtime field for identity backend is left empty on autocomplete, if not provided, until filled in later when the model is loaded.

rmccorm4: I see, thanks

qa/L0_pytorch_python_runtime/unit_test.py
```diff
@@ -0,0 +1,152 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+import unittest
+
+import torch
+
+sys.modules["triton_python_backend_utils"] = unittest.mock.MagicMock()
```
rmccorm4, line 34: Slick idea :eyes:

qa/L0_pytorch_python_runtime/unit_test.py
```diff
@@ -0,0 +1,152 @@
...
+from py_runtime import _gather_torch_tensors, _scatter_torch_tensors
```
rmccorm4, line 35: Can you add a comment that this is importing from python based implementation of pytorch backend? It wasn't clear to me at first and had to look around to figure that out. Can also mention the mock is just to satisfy import for python model implementation, but is only using the helpers and not the model itself.

qa/L0_pytorch_python_runtime/test.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/bash
```
rmccorm4, line 1: What do you think of moving this test into pytorch backend repo with `BACKEND:PYTORCH` child pipeline to keep things modular?

qa/L0_pytorch_python_runtime/unit_test.py
kthui: Sure.
[Add comments on unit test imports](https://github.com/triton-inference-server/server/pull/6518/commits/b59f7d79706e4ff614886fdccf6b134aa98adffe)

qa/L0_pytorch_python_runtime/test.sh
kthui: Can you elaborate a bit more on your thought? We already have `PYTORCH:TEST` child pipeline and this test is one of the tests on the child pipeline. If we are going to move all the tests on `PYTORCH:TEST` child pipeline from server repo to pytorch_backend repo, maybe we can file another ticket for it?

qa/L0_model_config/autofill_noplatform_success/onnx/cpu_instance/expected
```diff
@@ -32,4 +32,5 @@ optimization {
     enable: true
   }
 }
-backend: "onnxruntime"
\ No newline at end of file
+backend: "onnxruntime"
+runtime: ""
```
Tabrizian, line 36: should auto-complete fill in this field with correct value as well (i.e., this would be `/opt/tritonserver/backends/onnxruntime/libtriton_onnxruntime.so`)?

qa/L0_pytorch_python_runtime/test.sh
rmccorm4: I meant `PYTORCH:TEST` and yes if it is being included in that child pipeline on gitlab side, then we're good for now.

rmccorm4, line 1: Can you add a generic test that tries to set `runtime` field to some invalid values, and test that Triton gracefully handles it?
```
# config.pbtxt

# unexpected/bad value
runtime: "invalid_value"
```
and
```
# config.pbtxt

# no python implementation found
backend: "onnxruntime"
runtime: "model.py"
```
Maybe we can have something like this:
```
L0_runtime_invalid
L0_runtime_pytorch_python
```
and future tests can be `L0_runtime_*` for organization.

qa/L0_model_config/autofill_noplatform_success/onnx/cpu_instance/expected
kthui: That was my original thought, but the issue is how does auto-complete determine if the model uses C++ or Python runtime? A simple solution is to have backend specific auto-complete implementation. For example, if it the model uses PyTorch backend and a `*.pt` model file is provided without any `*.py` file, then use C++. If a `*.py` file is provided, then use Python. See [this commit](https://github.com/triton-inference-server/core/pull/285/commits/2e5b50963f4cd41b440a4359a1d038256f25179f). From previous discussions with @nnshah1 and @rmccorm4, we want to stop adding more backend specific logic into auto-complete and opt for a generic implementation, which Python runtime falls into the category of [custom backend](https://github.com/triton-inference-server/core/pull/285/files#diff-f947d03afcf01249667bedc5b955404da083564c22809b6666ba923356f7c829R1179) on auto-complete. Backends can implement either C++ or Python runtime (or both), we do not want to auto-complete into Python runtime on a C++ only backend, which requires checking the backend installation directory. The current structure limits auto-complete to only see model directory, but not backend directory, because the latter is resolved only when the model is loaded, which happens after auto-complete. A easier approach is to resolve runtime, if not already filled, when the model is loaded, which has both the model and backend directory information.

Going back to the original question, auto-complete does not touch the runtime field, it is determined/filled when the model is loaded, if it is not already filled.

qa/L0_pytorch_python_runtime/test.sh
kthui: [Add invalid runtime test](https://github.com/triton-inference-server/server/pull/6518/commits/cc6b256f5945a375da2ef272f98fcd5084e717ff)

Tabrizian (approved):

rmccorm4 (approved):
🚀

PR #6524, opened 2023-11-06
Add testing for implicit state enhancements

Tabrizian:


qa/L0_implicit_state/implicit_state.py
```diff
@@ -83,6 +83,80 @@ def test_wrong_implicit_state_name(self):
...
+            model_name="growable_memory",
+            inputs=inputs,
+            sequence_id=2,
+            sequence_start=True,
+            sequence_end=False,
+        )
+        output_state = output.as_numpy("OUTPUT_STATE")
+        expected_output_state = np.zeros(output_state.shape, dtype=np.int8)
+        np.testing.assert_equal(output_state, expected_output_state)
+
```
rmccorm4, line 128: Should this kind of checking be added (re-used) to `test_implicit_state_single_buffer` as well?

qa/L0_sequence_batcher/test.sh
```diff
@@ -332,6 +336,23 @@ for MODEL in $MODELS; do
       (cd queue_delay_models/$(basename $MODEL)_full && \
         sed -i "s/$(basename $MODEL)/$(basename $MODEL)_full/" config.pbtxt && \
         sed -i "s/minimum_slot_utilization: 0/minimum_slot_utilization: 1/" config.pbtxt)
+
+    # TODO: Enable single state buffer testing for sequence batcher
```
rmccorm4, line 340: still needed?

qa/L0_implicit_state/implicit_state.py
```diff
@@ -83,6 +83,80 @@ def test_wrong_implicit_state_name(self):
         err_str = str(e.exception).lower()
         self.assertIn("state 'undefined_state' is not a valid state name", err_str)
 
+    def test_implicit_state_single_buffer(self):
```
rmccorm4, line 86: Is is easy enough to test that growable memory isn't used (or returns error) when set to false?  ex: set growable_memory to false, send request that tries to grow state tensor, assert failure

qa/L0_implicit_state/implicit_state.py
Tabrizian: For single buffer we throw away the old buffer so what we are testing here is unique to growable buffer. For growable buffer the backend only appends data to the end of the buffer and here we make sure that the output is written correctly.

qa/L0_implicit_state/implicit_state.py
Tabrizian: You can grow the buffer even when you don't use `growable_memory`. The only catch is that the old memory is thrown away and a new memory is returned to the user.

qa/L0_sequence_batcher/test.sh
Tabrizian: Yes, I'm planning to add more testing for single state buffer later.

qa/L0_implicit_state/implicit_state.py
rmccorm4: I see, thanks

qa/L0_implicit_state/implicit_state.py
rmccorm4: I guess I'm more curious if that test is missing some testing, because it looks like all it does is send 2 inferences and then not check or validate anything afterwards. If inference simply succeeding is sufficient, that's fine - just wanted to double check if it was. It looks like there may be some verification in the model / `implicit_state.cc`

src/command_line_parser.cc
```diff
@@ -740,6 +741,18 @@ TritonParser::SetupOptions()
        "<GPU device ID>:<pool byte size>. This option can be used multiple "
        "times, but only once per GPU device. Subsequent uses will overwrite "
        "previous uses for the same GPU device. Default is 64 MB."});
+  memory_device_options_.push_back(
+      {OPTION_CUDA_VIRTUAL_ADDRESS_SIZE, "cuda-virtual-address-size",
+       "<integer>:<integer>",
```
rmccorm4, line 746: Is there any testing for this flag? I don't see if at a glance in the tests. I think some quick edge case coverage tests would be, assert server startup fails for:
1. size too big
2. size too small
3. size not valid with alignment

and then for the existing successful tests :
1. no size provided (test default works)
2. valid size provided (test cli override works)

qa/L0_implicit_state/implicit_state.py
Tabrizian: Yes, the verification is mainly done in implicit_state.cc.

src/command_line_parser.cc
Tabrizian: I did some testing offline and it looks like. Will add more coverage for this in follow up PRs if that's ok.

rmccorm4: Follow up is fine 👍

src/test/implicit_state/src/implicit_state.cc
```diff
@@ -783,9 +809,32 @@ TRITONBACKEND_ModelInstanceExecute(
         continue;
       }
 
-      const int32_t ipbuffer_state =
-          *reinterpret_cast<const int32_t*>(input_state_buffer);
-      ipbuffer_state_int = ipbuffer_state;
+      // When using single state buffer, input/output tensors should point to
+      // the buffer.
+      if ((test_case_buffer_int == 2 || test_case_buffer_int == 3) &&
```
rmccorm4, line 814: It would help readability here to use an enum or something for these values , but can be follow up as well.

rmccorm4 (approved):
One comment about followup, but lgtm

nnshah1 (approved):

PR #6526, opened 2023-11-07
Enhance GCS credentials documentations

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/280

Mention support for JSON authorized user credentials, and reformat the order of how the credentials are fetched and attempted.


Tabrizian (approved):
nice doc improvement!

PR #6542, opened 2023-11-08
install wget in Docker building

JoanFM:
When running `build.py` it seems that `wget` is not installed and fails to download `boost`


Tabrizian:
fixed in a59f3cf.

JoanFM:
> fixed in [a59f3cf](https://github.com/triton-inference-server/server/commit/a59f3cf09cc9cc6b266acf41752bb9e28a9f0481).

Thanks!!

PR #6546, opened 2023-11-09
Add timeout to client apis and tests

jbkyang-nvi:
Client PR: https://github.com/triton-inference-server/client/pull/429


Dockerfile.sdk
```diff
@@ -34,7 +34,6 @@ ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:23.10-py3-min
 ARG TRITON_CLIENT_REPO_SUBDIR=clientrepo
 ARG TRITON_COMMON_REPO_TAG=main
 ARG TRITON_CORE_REPO_TAG=main
-ARG TRITON_BACKEND_REPO_TAG=main
```
jbkyang-nvi, line 37: don't need backend repo in client build

src/grpc/grpc_utils.cc
```diff
@@ -73,6 +77,18 @@ operator<<(std::ostream& out, const Steps& step)
...
+  if (dstr != nullptr) {
+    delay_response = atoi(dstr);
+    // Will delay the write of the response by the specified time.
+    // This can be used to test the flow where there are other
+    // responses available to be written.
+    LOG_VERBOSE(1) << "Delaying the write of the response by " << delay_response
+                   << " seconds";
+    std::this_thread::sleep_for(std::chrono::seconds(delay_response));
+  }
+
```
jbkyang-nvi, line 91: Added this to create artificial delay in the grpc server's delays.

qa/L0_client_timeout/test.sh
```diff
@@ -50,27 +51,63 @@ rm -f *.log
...
+    echo -e "\n***\n*** Test Failed: GRPC non-infer APIs\n***"
+    RET=1
+fi
+
+set -e
+kill $SERVER_PID
+wait $SERVER_PID
+
+# Test infer APIs
+export TRITONSERVER_SERVER_DELAY_GRPC_RESPONSE_SEC=
```
tanmayv25, line 95: unset TRITONSERVER_SERVER_DELAY_GRPC_RESPONSE_SEC

src/grpc/grpc_utils.cc
```diff
@@ -73,6 +77,18 @@ operator<<(std::ostream& out, const Steps& step)
 void
 GrpcStatusUtil::Create(::grpc::Status* status, TRITONSERVER_Error* err)
 {
+  const char* dstr = getenv("TRITONSERVER_SERVER_DELAY_GRPC_RESPONSE_SEC");
```
tanmayv25, line 80: Instead of capturing this artificial delay here, I would recommend placing the sleep here: https://github.com/triton-inference-server/server/blob/main/src/grpc/grpc_server.cc#L237
This will cause all non-infer response writes to be delayed by the specified time interval.
Additionally, reading an environment variable is a bit heavy and can impact performance. I would recommend to call getenv once in each handler and then reuse it every time when writing the response.

tanmayv25 (changes requested):

src/grpc/grpc_server.cc
```diff
@@ -2388,9 +2407,15 @@ Server::Server(
   model_stream_infer_cq_ = builder_.AddCompletionQueue();
 
   // A common Handler for other non-inference requests
```
tanmayv25, line 2409: Move this comment to just above L2415.
and add a new comment saying here that response_delay is only used for testing and debugging purposes.

src/grpc/grpc_utils.cc
```diff
@@ -26,6 +26,10 @@
 
 #include "grpc_utils.h"
 
+#include <chrono>
```
tanmayv25, line 29: Do we need these includes here? 
I would suggest to remove them.

qa/L0_client_timeout/test.sh
```diff
@@ -204,11 +238,55 @@ set -e
...
+SERVER_ARGS="${SERVER_ARGS} --model-control-mode=explicit --load-model=custom_identity_int32 --log-verbose 2"
+run_server
+if [ "$SERVER_PID" == "0" ]; then
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    cat $SERVER_LOG
+    exit 1
+fi
+set +e
+
+for i in test_grpc_server_live \
```
tanmayv25, line 252: nit: These looks like too many test cases to iterate through. Can we create a new python test for non-infer API that can be directly invoked from here?
This would allow all the sub-tests to be executed without listing each of them here.

qa/L0_client_timeout/client_timeout_test.py
```diff
@@ -58,6 +58,299 @@ class ClientTimeoutTest(tu.TestResultCollector):
     def setUp(self):
         self.model_name_ = "custom_identity_int32"
         self.input0_data_ = np.array([[10]], dtype=np.int32)
+        self.input0_data_byte_size_ = 32
+        self.SMALL_INTERVAL = 0.1  # seconds for a timeout
+        self.INFER_SMALL_INTERVAL = 2.0  # seconds for a timeout
+        self.NORMAL_INTERVAL = 5.0  # seconds for server to load then receive request
+
+    def test_grpc_server_live(self):
```
tanmayv25, line 66: Move them to a new non_infer_client_timeout_test.py?

src/grpc/grpc_utils.cc
```diff
@@ -26,6 +26,9 @@
 
 #include "grpc_utils.h"
 
+#include <cstdlib>
```
tanmayv25, line 29: Do we need these?

tanmayv25 (approved):

PR #6563, opened 2023-11-10
Fix Dockerfile syntax error in `build.py`

nlile:
Corrected string concatenation in build.py to resolve a malformed RUN command in Dockerfile.buildbase, which improperly combined 'wget' and 'rm -rf /var/lib/apt/lists/*'.


Tabrizian:
Fixed in a59f3cf

PR #6577, opened 2023-11-14
changing references to concrete instances

nnshah1:
Underlying issue was the storage of a reference to the RestrictedFeatures object. In general this can be done as long as the lifetime of the object is known to be longer than the holder (similar to a pointer). However in this case for vertex and sagemaker a reference to a temporary was being stored. In http / grpc cases the reference was to an object with longer lifetime and so no issue was present.

Changed to copy the value and store it as this initialization only happens once.

### DIFF
--- a/src/http_server.h
+++ b/src/http_server.h
@@ -362,6 +362,15 @@ class HTTPAPIServer : public HTTPServer {
       const int32_t port, const bool reuse_port, const std::string& address,
       const std::string& header_forward_pattern, const int thread_cnt,
       const RestrictedFeatures& restricted_apis);
+  explicit HTTPAPIServer(
+      const std::shared_ptr<TRITONSERVER_Server>& server,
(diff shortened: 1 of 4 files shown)


src/http_server.h
```diff
@@ -362,6 +362,15 @@ class HTTPAPIServer : public HTTPServer {
       const int32_t port, const bool reuse_port, const std::string& address,
       const std::string& header_forward_pattern, const int thread_cnt,
       const RestrictedFeatures& restricted_apis);
```
Tabrizian, line 364: one other way to remove the need for a separate constructor would be to remove the const& and add a default value for restricted_apis.

Tabrizian (approved):

rmccorm4 (approved):

GuanLuo (approved):

PR #6585, opened 2023-11-15
Mention TRT LLM backend supports request cancellation

kthui:
Add TRT LLM backend into the list of backends that support request cancellation.


Tabrizian (approved):
Please cherry-pick it in the r23.11 branch too.

PR #6590, opened 2023-11-16
Fix generation endpoint segfault. Fix #6576

### LINKED ISSUE #6576: segmentation fault on /generate_stream when the client closes the connection
If the client tries to close the connection when the server is still generating, the server will crash on segmentation fault. 100% reproducible.

GuanLuo:
See inline comment for detail


rmccorm4:
Closing as I believe it was fixed here: https://github.com/triton-inference-server/server/pull/6591

Feel free to re-open if needed

PR #6591, opened 2023-11-16
Support for http generate request cancellation and segfault fix

nnshah1:
Hooking the request free callback from evhtp to support request cancellation.

From inspection, all request free operations have to go through:

https://github.com/triton-inference-server/third_party/blob/8af7da2ae3f4938a8b497728fcca6460797c402f/libevhtp/libevhtp/evhtp.c#L1233 

And connection free frees the request first:

https://github.com/triton-inference-server/third_party/blob/8af7da2ae3f4938a8b497728fcca6460797c402f/libevhtp/libevhtp/evhtp.c#L5151

So when a client disconnects we can trap the request free callback and cancel the request. 

There are three objects with lifetimes to maintain:

1) evhtp_req. Request object for libevhtp. This is created by libevhtp and freed by libevhtp when a connection is freed. We add the request_fini hook to get notified when this object is freed and use it as indication that the client has exited and cancelled the request. The req object can not be used after the request free hook. We remove the hook if the response is complete. 

2) InferRequestClass. InferRequestClass is created by the httpserver to wrap the evhtp_req and the TRITONSERVER_InferenceRequest objects. It is used in the response callbacks to send http responses. The lifetime of the object is between the request creation and the final response for a request.

3) TRITONSERVER_InferenceRequest. InferenceRequest is created by the httpserver and used to send inputs to the core. It is also used to cancel requests. Its lifetime is between request creation and the final response AND the request release callback (sent from the core).

We use a shared_ptr to manage the lifetime of the InferenceRequest. The shared_ptr is held by the request_release_payload (freed in the request_release_callback) and by the InferRequestClass (freed in the final response callbacks). This enables us to use the request object to cancel requests after it has been released by the core but before the final response is received.

Some minor refactoring to store the inference request in the InferRequestClass. 
Moved the Callbacks to the InferRequestClass to make Generate and InferRequestClass similar and to avoid having to make protected members public.


src/http_server.cc
```diff
@@ -3628,15 +3644,32 @@ HTTPAPIServer::BADReplyCallback(evthr_t* thr, void* arg, void* shared)
   delete infer_request;
 }
 
+std::unordered_set<evhtp_request*>
+    HTTPAPIServer::InferRequestClass::active_requests_ = {};
+
+evhtp_res
+HTTPAPIServer::InferRequestClass::RequestFiniHook(
```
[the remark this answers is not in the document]
nnshah1: That's correct. I think different requests would call this concurrently. Each request* should be on its own thread synchronized with the thread used for processing that request. I believe as long as each request is being processed on a specific thread then there's no concurrency issue as they won't be reading / writing to the same element.

That being said we can add a mutex - 

Another thought I had was to add the state to the associated irequest. My only concern there is how to ensure the irequest is still valid when the ev request is freed. I don't think there is a gurantee there.

qa/L0_http/generate_models/mock_llm/1/model.py
```diff
@@ -87,7 +87,10 @@ def exec_decoupled(self, requests):
                 for _ in range(rep_count):
                     if delay is not None:
                         time.sleep(delay)
-                    sender.send(response)
+                    if not sender.is_cancelled():
+                        sender.send(response)
```
Tabrizian, line 91: Does the server segfault if we send responses when the request is cancelled and connection is closed?

src/http_server.cc
```diff
@@ -3607,14 +3615,22 @@ HTTPAPIServer::OKReplyCallback(evthr_t* thr, void* arg, void* shared)
...
   HTTPAPIServer::InferRequestClass* infer_request =
       reinterpret_cast<HTTPAPIServer::InferRequestClass*>(arg);
 
   evhtp_request_t* request = infer_request->EvHtpRequest();
-  evhtp_send_reply(request, EVHTP_RES_BADREQ);
-  evhtp_request_resume(request);
+
+  if (InferRequestClass::active_requests_.count(request) == 0) {
+    if (infer_request->triton_request_ != nullptr) {
+      TRITONSERVER_InferenceRequestCancel(infer_request->triton_request_);
```
Tabrizian, line 3628: error checking

src/http_server.cc
```diff
@@ -4105,8 +4163,15 @@ HTTPAPIServer::GenerateRequestClass::EndResponseCallback(
   auto infer_request =
       reinterpret_cast<HTTPAPIServer::GenerateRequestClass*>(arg);
 
-  infer_request->SendChunkResponse(true /* end */);
-  evhtp_send_reply_chunk_end(infer_request->EvHtpRequest());
+  if (InferRequestClass::active_requests_.count(infer_request->req_) == 0) {
+    if (infer_request->triton_request_ != nullptr) {
+      TRITONSERVER_InferenceRequestCancel(infer_request->triton_request_);
```
Tabrizian, line 4168: error checking.

src/http_server.cc
```diff
@@ -4105,8 +4163,15 @@ HTTPAPIServer::GenerateRequestClass::EndResponseCallback(
   auto infer_request =
       reinterpret_cast<HTTPAPIServer::GenerateRequestClass*>(arg);
 
-  infer_request->SendChunkResponse(true /* end */);
-  evhtp_send_reply_chunk_end(infer_request->EvHtpRequest());
+  if (InferRequestClass::active_requests_.count(infer_request->req_) == 0) {
```
Tabrizian, line 4166: I think this is introducing a check in the hot path of inferencing. Is there any way to avoid requiring this check?

src/http_server.cc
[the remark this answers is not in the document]
Tabrizian: > Another thought I had was to add the state to the associated irequest. My only concern there is how to ensure the irequest is still valid when the ev request is freed. I don't think there is a gurantee there.

We handle the lifetime of the irequest so I think that's probably a better approach if I'm undressing it correctly.

Tabrizian (commented):
Also, please make sure that L0_infer_valgrind tests pass after this change.

qa/L0_http/generate_models/mock_llm/1/model.py
nnshah1: no - server can handle if responses are sent. this was changed to showcase the feature

src/http_server.cc
nnshah1: check is on response sending - and I don't think there is anyway to avoid it - as the basic issue is whether or not the evhtp request has been freed. So we have to check before using it.

Note: this could have happened even in non generative case I believe because if a request took a long time to process and the client connection was freed we would have also run into this.

The other option is to keep the request paused - but I'm not sure if there are other scenarios which can lead to the request being freed.

nnshah1: Updated to check if the req is null - this should be limited overhead.

src/http_server.cc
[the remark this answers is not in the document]
nnshah1: updated to not require a global state. We set the request in the InferClass now explicitly to null when the request is freed. If an InferClass is freed first - we unset the hook. 

All operations (set_hook, check req, unset hook) are synchronized on the same event thread (per connection / request).

src/http_server.cc
```diff
@@ -4105,8 +4173,17 @@ HTTPAPIServer::GenerateRequestClass::EndResponseCallback(
...
-  evhtp_send_reply_chunk_end(infer_request->EvHtpRequest());
+  if (infer_request->req_ == nullptr) {
+    if (infer_request->triton_request_ != nullptr) {
+      LOG_TRITONSERVER_ERROR(
+          TRITONSERVER_InferenceRequestCancel(infer_request->triton_request_),
+          "cancelling request");
+    }
+  } else {
+    infer_request->SendChunkResponse(true /* end */);
+    evhtp_send_reply_chunk_end(infer_request->EvHtpRequest());
```
rmccorm4, line 4184: Not a blocking comment for this PR - but could be a non-cherry-picked follow up. 

Are `infer_reqest->req_` and `infer_request->EvHtpRequest()` the same object? Seems like they're used interchangeably or sometimes both used in the same function like this one.

src/http_server.cc
```diff
@@ -3628,15 +3648,34 @@ HTTPAPIServer::BADReplyCallback(evthr_t* thr, void* arg, void* shared)
...
 }
 
+evhtp_res
+HTTPAPIServer::InferRequestClass::RequestFiniHook(
+    evhtp_request* request, void* arg)
+{
+  HTTPAPIServer::InferRequestClass* infer_request =
+      reinterpret_cast<HTTPAPIServer::InferRequestClass*>(arg);
+  if (infer_request->req_ != request) {
+    LOG_ERROR << "[INTERNAL] mismatched request in fini hook";
```
rmccorm4, line 3658: I assume this is not expected to happen, so should this return a EVHTP_RES_BADREQ to promote visibility of failure if it does happen?

rmccorm4 (approved):
Looks generally good to me. Left some comments for follow-ups, doesn't need to block cherry pick.

src/http_server.cc
nnshah1: The error code isn't actually checked or acted upon for the hook. Agree for completeness / readability.

src/http_server.cc
nnshah1: they are the same. previously the callbacks were accessing the object from a different class and so `infer_request->EvHtpRequest()` was necessary but with the refactor to move the callbacks to the infer class - I think we can remove it. will handle in follow up PR.

tanmayv25 (dismissed):
This looks great! I will introduce RequestReleasePayload class in gRPC frontend too which would fix some lifetime issues that we are still observing.

rmccorm4 (dismissed):
LGTM, not too sure if sagemaker server has any unique behavior, so will want to make sure L0_sagemaker passes.

GuanLuo (approved):

nnshah1:
> Also, please make sure that L0_infer_valgrind tests pass after this change.

Tested and expected tests passing.

rmccorm4 (approved):

PR #6595, opened 2023-11-17
Add documentation on generative sequence

GuanLuo:


docs/user_guide/model_configuration.md
```diff
@@ -980,6 +980,44 @@ indicating sequence start, end, ready and correlation ID. See
...
+
+Because only one request is used to represent the whole sequence, the user
+doesn't need to and should not set
+[control inputs](architecture.md#control-inputs) mentioned in the previous
+section. They will be filled internally by the scheduler.
+
+There are similarities between generative sequence and [decoupled](#decoupled)
+setting that undetermined number of responses may be generated for a request.
+One major difference is that generative sequence may reschedule and execute the
+same request repeatedly to fully utilize the batching feature in Triton: an
```
Tabrizian, line 1010:
```suggestion
same request repeatedly to fully utilize the batching feature in Triton: a
```

docs/user_guide/model_configuration.md
```diff
@@ -980,6 +980,44 @@ indicating sequence start, end, ready and correlation ID. See
...
+decoupled model generate all responses associated to a request within the same
```
Tabrizian, line 1011:
```suggestion
decoupled model generates all responses associated to a request within the same
```

Tabrizian (commented):
Thanks Guan! Could you also add a comment regarding that this is how you can achieve inflight batching in Triton? Additionally, it would be good to mention the supported backends and expected workflow (i.e., BLS on a step model).

nnshah1:
> Thanks Guan! Could you also add a comment regarding that this is how you can achieve inflight batching in Triton? Additionally, it would be good to mention the supported backends and expected workflow (i.e., BLS on a step model).

I agree we should add a section on inflight batching. For use case examples - I would instead point to an example and not add here to the documentation - what do you think? I think adding to the docs may get hard to read and would only be one possible scenario? Maybe in the section on inflight batching we could describe an example flow.

Tabrizian:
>what do you think? 
Sounds good. We can point to an example later.

rmccorm4 (approved):

PR #6602, opened 2023-11-17
Ading 'ldconfig' to configure dynamic linking in container

mc-nv:


qa/L0_infer/test.sh
```diff
@@ -38,6 +38,8 @@ if [ ! -z "$TEST_REPO_ARCH" ]; then
     REPO_VERSION=${REPO_VERSION}_${TEST_REPO_ARCH}
 fi
 
+ldconfig || true
```
nv-kmcgill53, line 41: Is this for iGPU?

mc-nv: Yes, this if for iGPU but shouldn't impact regular containers, as it fixing the dynamic linking.

nv-kmcgill53 (approved):

PR #6610, opened 2023-11-20
Restricted HTTP API documentation

nnshah1:


GuanLuo (dismissed):

docs/customization_guide/inference_protocols.md
```diff
@@ -145,24 +146,31 @@ protocol types mentioned above:
...
   * `logging` : [logging endpoint](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_logging.md).
 
-* `restricted-key` : Key to determine the GRPC request header to be checked when a
-request to the protocol is received. The completed header will be in the form of
-`triton-grpc-protocol-<restricted-key>`
+* `restricted-key` : The GRPC / HTTP request header
+to be checked when a request is received. The
+completed header for GRPC will be in the form of
+`triton-grpc-protocol-<restricted-key>`. The completed header for HTTP
+will be in the form of `<restricted-key>`.
```
rmccorm4, line 153: Just a question, not expecting any changes. Any reason we didn't follow similar format for HTTP like `triton-http-api-<restricted-key>` ?

docs/customization_guide/inference_protocols.md
```diff
@@ -145,24 +146,31 @@ protocol types mentioned above:
...
 ```
-tritonserver --grpc-restricted-protocol=shared-memory,model-config,model-repository,statistics,trace:<admin-key>=<admin-value> ...
+tritonserver --grpc-restricted-protocol=shared-memory,model-config,model-repository,statistics,trace:<admin-key>=<admin-value> \
+             --http-restricted-api=shared-memory,model-config,model-repository,statistics,trace:<admin-key>=<admin-value> ...
 ```
 
+GRPC requests to `admin` protocols require that an additional header
+`triton-grpc-protocol-<admin-key>` is provided with value
+`<admin-value>`. HTTP requests to `admin` APIs required that an
+additional header `<admin-key>` is provided with value `<admin-value`.
```
rmccorm4, line 172:
```suggestion
additional header `<admin-key>` is provided with value `<admin-value>`.
```

docs/customization_guide/inference_protocols.md
```diff
@@ -114,26 +114,27 @@ These options can be used to configure the KeepAlive settings:
...
 
 The option can be specified multiple times to specifies multiple groups of
-protocols with different restriction settings.
+protocols or APIs with different restriction settings.
 
-* `protocols` : A comma-separated list of protocols to be included in this
-group. Note that currently a given protocol is not allowed to be included in
-multiple groups. The following protocols are currently recognized by all network
-protocol types mentioned above:
+* `protocols / APIs` : A comma-separated list of protocols / APIs to be included in this
```
rmccorm4, line 135: (another nit not for this release)

I think we touched on this briefly when implementing, but if this is still considered beta or subject to change, my 2 cents would be to try to unify naming in the next release or something. ex:

To restrict access to certain Triton APIs over the network...:
```
--grpc-restricted-protocol
--http-restricted-protocol
```

Rather than having to say `protocols / API` in every spot, and having different arg names for each. Upon reading it, my first question is, "what's the difference between a protocol and an API here?"

I think the cli args can say `protocol` just to keep the `grpc` version unchanged since it came first, but we can describe them as either protocol **or** API in the text around it and be consistent with it.

rmccorm4:
LGTM other than this one comment: https://github.com/triton-inference-server/server/pull/6610/files#r1399722121

The rest of the comments are just nits/discussion - not looking for changes

docs/customization_guide/inference_protocols.md
nnshah1: Not really. In hindsight it might have been better to follow the same pattern. I think I was confused at why we prefixed a fixed string instead of just using the passed in header value. We can maybe revisit the behavior of grpc.

docs/customization_guide/inference_protocols.md
nnshah1: sounds good - we can add a ticket to unify behavior / naming. Since its beta we can decide whether to keep the grpc handling as is or modify it. having `--http-restricted-protocol` seems odd as it is a protocol itself but maybe its the least bad unified name. could consider, `--http-restricted-feature`, `--grpc-restricted-feature`. perhaps ...

rmccorm4 (approved):

PR #6620, opened 2023-11-21
Extend request objects lifetime and fixes possible segmentation fault

tanmayv25:
The  lifetime management is similar to what is being proposed in the PR for HTTP endpoint:
https://github.com/triton-inference-server/server/pull/6591#pullrequestreview-1740568846

### DIFF
--- a/src/grpc/stream_infer_handler.cc
+++ b/src/grpc/stream_infer_handler.cc
@@ -262,6 +262,12 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
       err = TRITONSERVER_InferenceRequestNew(
           &irequest, tritonserver_.get(), request.model_name().c_str(),
           requested_model_version);
+      state->inference_request_ = {
+          irequest, [](TRITONSERVER_InferenceRequest* request) {
+            LOG_TRITONSERVER_ERROR(
+                TRITONSERVER_InferenceRequestDelete(request),
+                "deleting gRPC inference request");
+          }};
     }
 
     if (err == nullptr) {
@@ -285,9 +291,13 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
           tritonserver_, shm_manager_, request, std::move(serialized_data),
           response_queue_, &state->alloc_payload_);
     }
+
+    auto request_release_payload =
+        std::make_unique<RequestReleasePayload>(state->inference_request_);
     if (err == nullptr) {
       err = TRITONSERVER_InferenceRequestSetReleaseCallback(
-          irequest, InferRequestComplete, nullptr /* request_release_userp */);
+          irequest, InferRequestComplete,
+          request_release_payload.get() /* request_release_userp */);
     }
     if (err == nullptr) {
       err = TRITONSERVER_InferenceRequestSetResponseCallback(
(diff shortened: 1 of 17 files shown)


src/grpc/stream_infer_handler.cc
```diff
@@ -285,9 +291,13 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
           tritonserver_, shm_manager_, request, std::move(serialized_data),
           response_queue_, &state->alloc_payload_);
     }
+
+    auto request_release_payload =
```
nnshah1, line 295: One thing to check is line src/grpcs/stream_infer_handler.cc#352

I found that when moving to the shared pointer / custom deleter I needed to remove any explicit TRITONSERVER_InferenceRequestDelete() calls as they would be called automatically via shared pointer cleanup. In the case that ownership is shared with request_release_payload and state - then it gets cleaned up when those are released.

If they are not shared (if some error happens before then) - then the shared pointer locally would get released. I'm not sure in this case if that is the case or not -but I would quickly check).

Its a bit of a tricky condition - do we have tests that cover memory leaks in error handling?

tanmayv25: Nice catch! The custom deleter would be invoked if the ownership of `request_release_payload` is held in case of InferAsync fails.

tanmayv25:
I am not sure if we have added the valgrind tests for errors but I quickly verified the error cases and it seems to be cleaning everything properly. 

```

root@334e8ee83eee:/opt/tritonserver/qa/L0_model_queue# cat verbose_1.log | grep "deleting gRPC"
I1122 01:30:12.793627 8306 infer_handler.cc:865] deleting gRPC inference request
I1122 01:30:12.794137 8306 infer_handler.cc:865] deleting gRPC inference request
I1122 01:30:12.794228 8306 infer_handler.cc:865] deleting gRPC inference request
I1122 01:30:22.793527 8306 infer_handler.cc:865] deleting gRPC inference request
root@334e8ee83eee:/opt/tritonserver/qa/L0_model_queue# cat verbose_1.log | grep "creating gRPC"
I1122 01:30:12.791597 8306 infer_handler.cc:862] creating gRPC inference request
I1122 01:30:12.791775 8306 infer_handler.cc:862] creating gRPC inference request
I1122 01:30:12.792236 8306 infer_handler.cc:862] creating gRPC inference request
I1122 01:30:12.792872 8306 infer_handler.cc:862] creating gRPC inference request
```

In the past we have seen some leak tests being triggered but I don't think we have comprehensive coverage.

src/grpc/stream_infer_handler.cc
```diff
@@ -262,6 +262,12 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
       err = TRITONSERVER_InferenceRequestNew(
           &irequest, tritonserver_.get(), request.model_name().c_str(),
           requested_model_version);
+      state->inference_request_ = {
```
nnshah1, line 265: do we need to check err here? Or is it guranteed irequest is set to null or not changed? (meaning is it safe to delete it with InferenceRequestDelete() if InferenceRequestNew() doesn't succeed?)

nnshah1:
LGTM - one question on calling delete if new fails -

src/grpc/stream_infer_handler.cc
tanmayv25: The current TRITONSERVER_InferenceRequestNew implementation does not change the request pointer on error.
However, let's play it safe and place it within the check.

rmccorm4 (approved):
Nice changes!

nnshah1 (approved):

PR #6639, opened 2023-11-28
Allow configuring GRPC max connection age and max connection age grace

siweili11:
Add ability to configure GRPC max connection age and max connection age grace. The definitions of the two parameters can be found in:

https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification

https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html#gaf4574abe94c339c6f21163bca6e7b6b7

Same as the initial GRPC proposal: https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md#proposal, the reason why we need to configure the two parameters is to spread load more quickly, by forcing the client to periodically reconnect to a L4 LB.


jbkyang-nvi:
Hi! Thanks for your contribution. Can you fill in the [CLA](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) for this?

siweili11:
Hi, thanks for the guidance! I have reformatted my changes and rebased. Please take a look again when you have a chance.

I'll fill in the CLA before this code change is merged

src/grpc/grpc_server.h
```diff
@@ -66,13 +66,16 @@ struct SslOptions {
...
+// https://grpc.io/docs/guides/keepalive/
 struct KeepAliveOptions {
   int keepalive_time_ms_{7200000};
   int keepalive_timeout_ms_{20000};
   bool keepalive_permit_without_calls_{false};
   int http2_max_pings_without_data_{2};
   int http2_min_recv_ping_interval_without_data_ms_{300000};
   int http2_max_ping_strikes_{2};
+  int max_connection_age_ms_{INT_MAX};
+  int max_connection_age_grace_ms_{INT_MAX};
```
jbkyang-nvi, line 78: actually per convention can you just set this to 0?

src/grpc/grpc_server.cc
```diff
@@ -2366,6 +2366,12 @@ Server::Server(
     builder_.AddChannelArgument(
         GRPC_ARG_HTTP2_MAX_PING_STRIKES,
         keepalive_options.http2_max_ping_strikes_);
+    builder_.AddChannelArgument(
```
jbkyang-nvi, line 2369: can you just check if the parameter is populated and only `AddChannelArgument` if it is nonzero?

jbkyang-nvi (changes requested):

src/grpc/grpc_server.h
siweili11: If I understand the code correctly, the value inside {} is the default value of the option if the flag is not set. So I want the default values here to be consistent with GRPC default values, see https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification. 

Basically GRPC by default tend to re-use existing connections as possible as it can to reduce network overhead, I think we should preserve this behavior in Triton unless the flags are set.

Please let me know if my understanding is incorrect.

src/grpc/grpc_server.cc
siweili11: Sorry I'm still new to this. Can you please provide more details on how to check whether the parameter is populated?

siweili11: Or do you mean I should make code changes such that AddChannelArgument will only be invoked when the parameter is set by the user with a non-default value?

src/grpc/grpc_server.h
jbkyang-nvi: Sorry for the confusion here. I was trying to say instead of specifying the max connection age, just don't specify `max_connection_age_ms_` or `max_connection_age_grace_ms_` upon initialization of the `KeepAliveOptions` struct and never pass these parameters to the server if the values are 0.

siweili11: Thanks for the explanation, I have updated my code change according to comments

src/grpc/grpc_server.h
```diff
@@ -66,13 +66,16 @@ struct SslOptions {
...
 // GRPC KeepAlive: https://grpc.github.io/grpc/cpp/md_doc_keepalive.html
+// https://grpc.io/docs/guides/keepalive/
 struct KeepAliveOptions {
   int keepalive_time_ms_{7200000};
   int keepalive_timeout_ms_{20000};
   bool keepalive_permit_without_calls_{false};
   int http2_max_pings_without_data_{2};
   int http2_min_recv_ping_interval_without_data_ms_{300000};
   int http2_max_ping_strikes_{2};
+  int max_connection_age_ms_{};
```
jbkyang-nvi, line 77: explicitly initialize to 0 here

jbkyang-nvi (commented):
Looks good! If you make the changes and file the CLA, I will kick off a pipeline run in our internal CI, we should be good to go.

siweili11:
Hi, my company has sent an email to triton-cla@nvidia.com requesting to add me to my company CLA. Can you please approve when you have a chance?

jbkyang-nvi:
@siweili11  can you *explicitly* initialize to 0?

siweili11:
Thanks. I have updated my code change to initialize as 0

jbkyang-nvi (approved):
Thanks for your contribution!

siweili11:
Thanks for merging! May I know when this change can be included in a released image?

jbkyang-nvi:
This will be available in the 24.01 image (our freeze passed a week ago)

jbkyang-nvi:
> Hi，how can i set these params when start a trion server?

you should be able to set `--grpc-max-connection-age` and `--grpc-max-connection-age-grace` in the command line when you run tritonserver. Details can be found if you do `tritonserver --help`

PR #6663, opened 2023-12-05
Add test for optional internal tensor within an ensemble

GuanLuo:


qa/L0_optional_input/optional_input_test.py
```diff
@@ -142,13 +142,8 @@ def check_status(self, model_name, batch_exec, request_cnt, infer_cnt):
             stats = self.triton_client_.get_inference_statistics(model_name, "1")
             self.assertEqual(len(stats.model_stats), 1, "expect 1 model stats")
             actual_exec_cnt = stats.model_stats[0].execution_count
-            if actual_exec_cnt == exec_cnt:
+            if stats.model_stats[0].execution_count > 0:
```
tanmayv25, line 145: Is this intended? Comparing with `exec_cnt` seemed like we were waiting for a number of executions to have completed before comparing the stats.

GuanLuo: The check for `exec_cnt` is incorrectly added for this `check_status` function, as you can see the function doesn't even expect `exec_cnt` to be passed. It seems to be some global change was mistakenly applied to this test because the function name is the same, here I just apply a WAR to bypass the incorrect checking.

tanmayv25 (approved):

PR #6668, opened 2023-12-06
Allow users to specify github organizations for triton core, backend, common, third-party, etc.

ClifHouck:
This change adds several CMake variables to allow the use of other github organizations for triton core, backend, common, third-party, etc. That way a developer can easily pull in changes from their own forks of triton related libraries and have it included in their build process.


ClifHouck:
Created an issue for this https://github.com/triton-inference-server/server/issues/6718

ClifHouck:
It seems like there's some work going on to support this PR. Any feedback or review of the work done in this PR would be appreciated.

PR #6669, opened 2023-12-07
Omit registered_model_meta file when uploading to registry.

albertoperdomo2:
It seems that MLFLow broke backwards compatibility with this plugin by introducing a new meta file to keep traceability of the models [#9402](https://github.com/mlflow/mlflow/commit/85037652d6caf4a78d8ac73ca734e433e86b1fd9). The file intent is good, but it's still less informative for our use case (it only contains name and version, although it might replace our `mlflow-meta.json` in the future).


albertoperdomo2:
cc: @GuanLuo

albertoperdomo2:
@nnshah1 it is already fixed! 🙌🏽

PR #6678, opened 2023-12-09
Fix extra content-type headers in HTTP server

rmccorm4:
**Overview**

evhtp_add_header was appending extra headers when already set, rather than overwriting existing value, ending with some cases of extra headers.

Rather than adding more logic and branching throughout the code, just updating the `AddContentType` header to first remove the header if found before adding new one.

---

Before:
```
# print(r.headers) # on /generate_stream success
{'Content-Type': 'application/json, text/event-stream; charset=utf-8', 'Transfer-Encoding': 'chunked'}

# print(r.headers) # on /generate_stream error
{'Content-Type': 'application/json, application/json'}
```

After:
```
# print(r.headers) # on /generate_stream success
{'Content-Type': 'text/event-stream; charset=utf-8', 'Transfer-Encoding': 'chunked'}

# print(r.headers) # on /generate_stream error
{'Content-Type': 'application/json'}
```

---

**Testing**

Added more strict equality checking to header values in generate endpoint tests

### DIFF
--- a/docs/protocol/extension_generate.md
+++ b/docs/protocol/extension_generate.md
@@ -55,17 +55,22 @@ POST v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]/generate_stream
 
 ### generate v.s. generate_stream
 
-Both URLs expect the same request JSON object, and generate the same response
-JSON object. However, `generate` returns exactly 1 response JSON object, while
-`generate_stream` may return multiple responses based on the inference
-results. `generate_stream` returns the responses as
+Both URLs expect the same request JSON object, and generate the same response.
+However, there are some differences in the format used to return each:
+* `generate` returns exactly 1 response JSON object with a
+`Content-Type` of `application/json`
+* `generate_stream` may return multiple responses based on the inference
+results. Successful responses will be sent as
 [Server-Sent Events](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events)
-(SSE), where each response will be a "data" chunk in the HTTP response body.
-Also, note that an error may be returned during inference, whereas the HTTP
-response code has been set in the first response of the SSE, which can result in
-receiving an [error object](#generate-response-json-error-object) while status
-code shows success (200). Therefore the user must always check whether an error
-object is received when generating responses through `generate_stream`.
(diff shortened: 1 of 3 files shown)


qa/L0_http/generate_endpoint_test.py
```diff
@@ -71,6 +71,9 @@ def generate_expect_failure(self, model_name, inputs, msg):
         r = requests.post(
             url, data=inputs if isinstance(inputs, str) else json.dumps(inputs)
         )
+        # Content-Type header should always be JSON for errors
+        self.assertEqual(r.headers["Content-Type"], "'application/json'")
```
nnshah1: Can we add a note to the generate endpoint documentation with that clarification (for initial errors content type will be application/json). Another option could be to use text/event-stream for all errors as well. I could not find a reference that spelled it out one way or the other.

docs/protocol/extension_generate.md
```diff
@@ -55,17 +55,27 @@ POST v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]/generate_stream
...
+an [error JSON object](#generate-response-json-error-object).
+    * Note that the HTTP response code is set in the first response of the SSE,
+    so if the first response succeeds but an error occurs in a subsequent
+    response for the request, it can result in receiving an error object
+    while the status code shows success (200). Therefore, the user must
+    always check whether an error object is received when generating
+    responses through `/generate_stream`.
+    * If the request fails before inference begins, then a JSON error will
+    be returned with `Content-Type` of `application/json`, similar to errors
+    from other endpoints and the status code should correctly reflect an error.
```
nnshah1, line 78:
```suggestion
    from other endpoints.
```

docs/protocol/extension_generate.md
```diff
@@ -55,17 +55,27 @@ POST v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]/generate_stream
 
 ### generate v.s. generate_stream
```
nnshah1, line 56:
```suggestion
### generate vs. generate_stream
```

GuanLuo (approved):

nnshah1 (approved):

PR #6686, opened 2023-12-12
Add test case for decoupled model raising exception

krishung5:
PYBE: https://github.com/triton-inference-server/python_backend/pull/327


qa/L0_backend_python/decoupled/decoupled_test.py
```diff
@@ -289,6 +289,35 @@ def test_decoupled_execute_cancel(self):
         self.assertIn("[execute_cancel] Request not cancelled at 1.0 s", log_text)
         self.assertIn("[execute_cancel] Request cancelled at ", log_text)
 
+    def test_decoupled_raise_exception(self):
+        # The decoupled_raise_exception model raises an exception for the request.
+        # This test case is making sure that repeated exceptions are properly handled.
+
+        model_name = "decoupled_raise_exception"
+        shape = [2, 2]
+        number_of_requests = 2
```
rmccorm4, line 298: In some cases of my local testing when I observed this, it took a handful of requests. Could you crank this up to 10 or something just for peace of mind?

rmccorm4: Also maybe an additional call to `is_model_ready` or something at the end just to test that some utilities still function after a model returns errors. I think there was a recent issue where even these types of calls would hang when the model failed for some other reasons.

krishung5: Sounds good! Increased the number of requests and added a `is_model_ready` check.

rmccorm4 (approved):

Tabrizian (approved):

PR #6695, opened 2023-12-13
Fix iGPU CMakeFile tags

kthui:


Dockerfile.QA
nv-kmcgill53, line 1: @kthui Can you confirm my syntax is correct in this file

kthui: Pre-commit should be fixed by [Pre-commit fix](https://github.com/triton-inference-server/server/pull/6695/commits/83183e71a0346af37681f24ac0ea6b7fb4e8e221)

src/test/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ find_package(GTest REQUIRED)
 #
 # caffe2plan
 #
-if(${TRITON_ENABLE_TENSORRT})
+if(${TRITON_ENABLE_TENSORRT} AND NOT ${TRITON_IGPU_BUILD})
```
mc-nv, line 37: Are we going to fee our `build.py` with extra CMake argument. 
I wasn't able from glance find it.

kthui: I think @nv-kmcgill53 is right that we cannot build this part as it involves some peculiar linking. It used to be there I think.

nv-kmcgill53: @mc-nv We are currently adding this to the build. [See here](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/blob/3fe14c296226d262d5299c516f589ba5bd564a20/.gitlab-ci.yml#L1126)

PR #6713, opened 2023-12-16
Modify HTTP frontend to return error code reflecting Triton error. Add test

GuanLuo:


src/sagemaker_server.cc
```diff
@@ -407,49 +407,27 @@ SagemakerAPIServer::SagemakeInferRequestClass::InferResponseComplete(
...
 #endif  // TRITON_ENABLE_TRACING
 
-  if (err == nullptr) {
-    evthr_defer(infer_request->thread_, OKReplyCallback, infer_request);
-  } else {
+  if (err != nullptr) {
     EVBufferAddErrorJson(infer_request->req_->buffer_out, err);
+    // [FIXME] In http_server.cc, error handling is enhanced to reporting
+    // different error code according to the Triton error code, holding
+    // the change from SageMaker endpoint as it may not fit their SLA.
```
Tabrizian, line 414: do we need similar fix for vertex endpoint as well?

Tabrizian (dismissed):

qa/L0_http/http_test.py
```diff
@@ -205,6 +207,31 @@ def test_content_encoding_unsupported_client(self):
...
+        # aligns with error detail (server unavailable).
+        t = threading.Thread(
+            target=self._raw_binary_helper, args=(model, input_bytes, input_bytes)
+        )
+        t.start()
+        time.sleep(0.5)
+        try:
+            self._raw_binary_helper(model, input_bytes, input_bytes)
+        except requests.exceptions.HTTPError as ex:
+            self.assertEqual(
```
rmccorm4, line 225: 1. Should the test be changed to expect exception to occur everytime? Looks like current test would pass if no exception occurs.
2. If so, will 0.5secs be flaky?

src/sagemaker_server.cc
GuanLuo: I will leave the same comment in vertex endpoint. Will also need to check with the right group of people for the same before applying the change

qa/L0_http/http_test.py
GuanLuo: > Should the test be changed to expect exception to occur everytime? Looks like current test would pass if no exception occurs.

Good catch, will fix to fail if there is no exception

src/sagemaker_server.cc
GuanLuo: Actually it is harder for vertex AI.. The endpoint just call into the http_server base class's functions and it gets the same behavior already.. I would just let it be what it is and fix later if specific error code must be returned in vertex AI deployment.

rmccorm4 (approved):
LGTM assuming vertex/sagemaker OK

Issue #302, opened 2019-05-22
Time to get_async_run_results is very slow


ThiagoMateo:
Hi every one, I get result by using async_run method. Time for async_run is 0.007930517196655273, but time for get_async_run_results is 0.11087822914123535. so it is very slow. 
```
   while not last_request:
        input_batch = []
        for idx in range(batch_size):
            input_batch.append(image_data[image_idx])
            image_idx = (image_idx + 1) % len(image_data)
            if image_idx == 0:
                last_request = True
                batch_size = len(input_batch)
                break
        request_ids.append(ctx.async_run(
            {input_name: input_batch},
            {boxes: (InferContext.ResultFormat.RAW),
             scores: (InferContext.ResultFormat.RAW)},
            batch_size))
    time_infer_stop = time.time()
    print("v_detection inference time: ", time_infer_stop - time_infer_start)

    time_post_start = time.time()
    # For async, retrieve results according to the send order
    for request_id in request_ids:
        results.append(ctx.get_async_run_results(request_id, True))
   
    time_post_stop = time.time()
    print("v_detection post processing time: ", time_post_stop - time_post_start)
    print("total_time_process: ", time_post_stop - time_infer_start)
```
how can i reduce get_async_run_results time?

deadeyegoodwin:
The async_run method itself just sends the request to the server, it doesn't wait for the response.
get_async_run_results will return quickly if the response has already been received, but if the response has not yet been received then it must wait. You can see this is you put a sleep(1 sec) before the call to get_async_run_results. If you do that you should find that it returns very quickly.

ThiagoMateo:
hi @deadeyegoodwin , thank for your help. but how to put a sleep?

deadeyegoodwin:
It is just a standard python function.

import time
time.sleep(5)   # Delays for 5 seconds. You can also use a float value.

Of course this is only to demonstrate what I described above, you don't want to leave the sleep in your "production" code.

ThiagoMateo:
so, are there no way to reduce inference time?

deadeyegoodwin:
The inference time is determined by the model and the framework. Using the async API doesn't speed up the actual inference time on the server... it just allows that client thread to do something else instead of just waiting for the response (which is what the thread does in the non-async API).

Have you looked at perf_client: https://docs.nvidia.com/deeplearning/sdk/tensorrt-inference-server-master-branch-guide/docs/client.html#performance-example-application

[...]

PR #6714, opened 2023-12-16
Remove Python backend double unit test

kthui:
The issue appears fixed by #302, so the double unit test before checking for memory leak can be reduced down to a single unit test.

### DIFF
--- a/qa/L0_backend_python/python_unittest.py
+++ b/qa/L0_backend_python/python_unittest.py
@@ -75,8 +75,6 @@ class PythonUnittest(tu.TestResultCollector):
                 # call so that the probe can detect the leak correctly.
                 self._run_unittest(model_name)
 
-                # [FIXME] See DLIS-3684
-                self._run_unittest(model_name)
                 with self._shm_leak_detector.Probe() as shm_probe:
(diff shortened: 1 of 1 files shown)


Tabrizian (approved):
Can we remove the if condition as well?

krishung5 (approved):

rmccorm4 (approved):

rmccorm4:
> Can we remove the if condition as well?

Technically I think this is removing the "triple" unit test - and IIRC, the BLS/decoupled cases in the `if` condition would still fail without a "warmup" or "double" run when I last looked into it.

I'll defer to Jacky on that in case he finds something different from his testing/experiments.

kthui:
> > Can we remove the if condition as well?
> 
> Technically I think this is removing the "triple" unit test - and IIRC, the BLS/decoupled cases in the `if` condition would still fail without a "warmup" or "double" run when I last looked into it.
> 
> I'll defer to Jacky on that in case he finds something different from his testing/experiments.

I did a quick check, I can confirm 'the BLS/decoupled cases in the `if` condition would still fail without a "warmup" or "double" run'.

PR #6725, opened 2023-12-20
Simplify cmake install command

Tabrizian:


mc-nv (dismissed):
Have no objection on this if it compiles all our builds. 

Hope there is no custom targets that will be excluded.

build.py
```diff
@@ -290,15 +290,8 @@ def cmake(self, args):
...
-            verbose_flag = "" if self._verbose else "-clp:ErrorsOnly"
-            self.cmd(
-                f"msbuild.exe -m:{FLAGS.build_parallel} {verbose_flag} -p:Configuration={FLAGS.build_type} {target}.vcxproj",
-                check_exitcode=True,
-            )
-        else:
-            verbose_flag = "VERBOSE=1" if self._verbose else "VERBOSE=0"
-            self.cmd(f"make -j{FLAGS.build_parallel} {verbose_flag} {target}")
+        verbose_flag = "-v" if self._verbose else ""
+        self.cmd(f"cmake --build . -j{FLAGS.build_parallel} {verbose_flag} -t {target}")
```
nv-kmcgill53, line 294: This omits the `build_type` flag, no? With this change we/users won't be able to set `--build-type=Debug`.

nv-kmcgill53 (changes requested):

build.py
mc-nv: I believe we switch code generator in favor of `cmake` if it works then we may have to drop the msbuild installation as well inside the container.

Tabrizian: Added `--config` to pass the build type.

build.py
```diff
@@ -290,15 +290,10 @@ def cmake(self, args):
...
-            self.cmd(
-                f"msbuild.exe -m:{FLAGS.build_parallel} {verbose_flag} -p:Configuration={FLAGS.build_type} {target}.vcxproj",
-                check_exitcode=True,
-            )
-        else:
-            verbose_flag = "VERBOSE=1" if self._verbose else "VERBOSE=0"
-            self.cmd(f"make -j{FLAGS.build_parallel} {verbose_flag} {target}")
+        verbose_flag = "-v" if self._verbose else ""
+        self.cmd(
+            f"cmake --build . --config {FLAGS.build_type} -j{FLAGS.build_parallel} {verbose_flag} -t {target}"
```
mc-nv, line 295: Aren't build directory needs to be declared similar to other flags?

build.py
Tabrizian: > I believe we switch compiler in favor of cmake if it works then we may have to drop the msbuild installation as well inside the container.

This is mostly a wrapper around `make`, `msbuild`, `gcc`, etc. We still need the underlying dependencies.

build.py
Tabrizian: could you elaborate? I don't see a build directory variable before these changes.

mc-nv: In our scenarios we use `.` reference as CMake-generated project binary tree to run`cmake --build <dir>` .
I'm aware that  outcome of `build.py` will be a generated script which will create the directory for CMake-generated project binary tree and switch the location to it during the execution. 

However from the glance on a given line I see that the all argument values are given as a variables except CMake-generated project binary tree. 
Defined function by itself doesn't explains why `.` location is used and it's creates a confusion for those who not familiar yet with results. 
```python
f"cmake --build . --config {FLAGS.build_type} -j{FLAGS.build_parallel} {verbose_flag} -t {target}"
```

nv-kmcgill53 (approved):
LGTM

build.py
Tabrizian: what is the action here? Does adding a comment help clarify why "." is used?

nv-kmcgill53 (approved):

PR #6758, opened 2024-01-02
Update trace summary script

pskiran1:


Tabrizian (commented):
I feel like still we should be looking for ways to correctly record `GRPC WAITREAD` or remove this value.

One scenario that I can imagine being confusing for the user is when they compare the value of `GRPC infer request` in gRPC and HTTP and notice much longer values for gRPC since it takes into account the gaps between the requests as well.

tanmayv25:
@pskiran1  Were you able to investigate whether gRPC server allows us to record **actual** request read time? Similar to libevhtp which allow us to record a timestamp on the first set of bytes, does gRPC have any hooks which can be used to detect the network activity.

If not, then I would advocate for removing the value as well to avoid any confusion.

pskiran1:
As we discussed internally, gRPC does not provide a direct hook for recording request read time, and using alternative methods such as interceptors may add latency/overhead to the current server. Therefore, we have decided to remove the `GRPC WAITREAD` related values in the summary to avoid confusion.

docs/user_guide/trace.md
```diff
@@ -274,29 +274,27 @@ GRPC inference requests are reported separately.
...
+GRPC infer request (avg): 383.601us
+	Send (avg): 62.816us
+	Handler (avg): 392.924us
+  		Overhead (avg): 51.968us
+  		Queue (avg): 21.45us
+  		Compute (avg): 319.506us
+  			Input (avg): 27.76us
+  			Infer (avg): 227.844us
+  			Output (avg): 63.902us
 ```
```
tanmayv25, line 299: Can you add a note here about why Receive (avg) is missing in the gRPC summary here?

tanmayv25:
a small change otherwise looks good to me.

Tabrizian (approved):

Tabrizian:
@pskiran1 Actually, I realized that you need to remove it from grpc trace recording too.

Tabrizian:
https://github.com/triton-inference-server/server/blob/8fc7b10e19e53fb130c6171bb15161ebaaadbf9c/src/grpc/infer_handler.cc#L667

PR #6762, opened 2024-01-03
Fix L0_lifecycle on insufficient hardware concurrency

kthui:
There is a check on "L0_lifecycle" "test_concurrent_same_model_load_unload_stress" case which fails the entire test if the case was not able to reproduce a load while async unloading. This check can be problematic on machines with less concurrency, because they may not be able to reproduce a load while async unloading at all. This PR changes the check from failing the entire test to log a warning on the test printout.

For example, on a machine that reproduces a load while async unloading:
```
+ python lifecycle_test.py LifeCycleTest.test_concurrent_same_model_load_unload_stress
+ '[' 0 -ne 0 ']'
+ cat ./test_concurrent_same_model_load_unload_stress.statistics.log
{"failed to unload 'identity_zero_1_int32', versions that are still available: 1": 115}     <---- acceptable failure logged
```

On a machine that does not reproduces a load while async unloading:
```
+ python lifecycle_test.py LifeCycleTest.test_concurrent_same_model_load_unload_stress
+ '[' 0 -ne 0 ']'
+ cat ./test_concurrent_same_model_load_unload_stress.statistics.log
{'Cannot replicate a load while async unloading. CPU count: 32. num_threads: 32.': 1}     <---- logged warning
```

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -2996,6 +2996,16 @@ class LifeCycleTest(tu.TestResultCollector):
             for model_name in model_names:
                 self.assertEqual(is_load, triton_client.is_model_ready(model_name))
 
+    # TODO: Consider revisiting this test
+    # The goal of this test is only to ensure the server does not crash when
+    # bombarded with concurrent load/unload requests for the same model.
+    # Some clean-up:
+    # 1. Improve core logic so all load/unload requests will always success, so
+    #    'load_fail_reasons' and 'unload_fail_reasons' can be removed.
+    # 2. Is it still necessary to track the ability to replicate a load while
+    #    async unloading?
+    # 3. What is the ideal number of threads and iterations, across different
+    #    machines, that the server is sufficiently stressed?
     def test_concurrent_same_model_load_unload_stress(self):
         model_name = "identity_zero_1_int32"
         num_threads = 32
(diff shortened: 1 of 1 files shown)


kthui:
What do you think about adding some margins between the `minimum expected number of threads` and `detected number of threads`?

Currently, the warning is silent when `detected number of threads <= minimum expected number of threads`. What do you think about `detected number of threads <= ( minimum expected number of threads * 2 )`?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -3062,10 +3063,16 @@ def _load_unload():
...
         self.assertTrue(triton_client.is_server_live())
         self.assertTrue(triton_client.is_server_ready())
-        self.assertTrue(
-            load_before_unload_finish[0],
-            "The test case did not replicate a load while async unloading. Consider increase concurrency.",
-        )
+
+        # This test can replicate a load while async unloading on machines with
+        # sufficient concurrency. Regardless on whether it is replicated or not,
+        # the server must not crash.
```
rmccorm4, line 3069: WAR LGTM, but for this test as a whole, I think it would be good to leave in a `# TODO: Consider revisiting this test` comment or similar, as I think it's a bit of a strange test due to hand-wavy conditions.

kthui: Sure. [Add some TODOs for the sub-test](https://github.com/triton-inference-server/server/pull/6762/commits/5cb234e12b4e6a03b87468d3efea5f92c1e4a272)

rmccorm4 (approved):

PR #6768, opened 2024-01-07
Fixes for Clang warnings

HennerM:
I was trying to compile triton server with default Clang 15 and got a few warnings that triggered as errors, among those:

* deprecation errors from open telemetry showing up, which shouldn't be an issue here. They come up because the headers aren't included as "SYSTEM" headers.
* unused parameter in TritonParser::SetGlobalTraceArgs(). There is a `explicit_disable_trace` passed as bool, that was updated but never read. I believe this was intended to be a `bool&` @oandreeva-nv ? (error: parameter 'explicit_disable_trace' set but not used [-Werror,-Wunused-but-set-parameter])
* An unused `[this]` capture (-Wunused-lambda-capture)
* Assignment with `std::move` of temporary variable resulting in: ` error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]` I am not quite sure what's the problem, but I think the std::move is redundant here anyway?
* a few conditionally uninitialised variables in tests  variable 'ba_memory_type' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
* unused private field in ` HTTPAPIServer`:  error: private field 'end_' is not used [-Werror,-Wunused-private-field]
    bool end_{false};

There is more in the core and backend repos that I will raise separately


src/command_line_parser.cc
```diff
@@ -2134,7 +2134,7 @@ void
 TritonParser::SetGlobalTraceArgs(
     TritonServerParameters& lparams, bool trace_level_present,
     bool trace_rate_present, bool trace_count_present,
-    bool explicit_disable_trace)
+    bool& explicit_disable_trace)
```
oandreeva-nv, line 2137: Nice catch! Thanks for the fix

rmccorm4:
Hi @HennerM, thanks for the contributions, these look great! Can you update the branch and resolve the conflicts? Then we can run some quick tests to make sure nothing breaks and approve.

rmccorm4:
Also just to double check, have you sent in a CLA as described here: https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla?

PR #6769, opened 2024-01-07
Use OTEL_SERVICE_NAME environment variable if set

### LINKED ISSUE #5987: OpenTelemetry Tracing allow to configure service name
**Is your feature request related to a problem? Please describe.**
We are using the Open Telemetry tracing feature of triton, however we are not able to configure the [`service.name`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/README.md#service) of the exported traces.

**Describe the solution you'd like**
With most Open Telemetry SDKs the service name can be configured via the environment variable `OTEL_SERVICE_NAME`, see [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#general-sdk-configuration). However when I tested this with triton the service name still was the default value `unknown_service` after I set the environment variable.

**Describe alternatives you've considered**
As an alternative the service name can be configured via a triton tracing config flag.

**Additional context**
The service name is a very important attribute of span, which allows to build a service graph and visualize the different services which are part of the trace.

HennerM:
`triton-inference-server` is the default OpenTelemetry service name if none other is set. Most libraries and applications read the `OTEL_SERVICE_NAME` environment variable though to get the name. This is override by the explicit setting of `triton-inference-server`.

Fixes #5987

### DIFF
--- a/src/tracer.cc
+++ b/src/tracer.cc
@@ -363,7 +363,7 @@ TraceManager::InitTracer(const triton::server::TraceConfigMap& config_map)
       otlp::OtlpHttpExporterOptions opts;
       otel_resource::ResourceAttributes attributes = {};
       attributes[otel_resource::SemanticConventions::kServiceName] =
-          "triton-inference-server";
+          triton::server::GetEnvironmentVariableOrDefault("OTEL_SERVICE_NAME", "triton-inference-server");
       auto mode_key = std::to_string(TRACE_MODE_OPENTELEMETRY);
       auto otel_options_it = config_map.find(mode_key);
       if (otel_options_it != config_map.end()) {


oandreeva-nv:
Thanks for this PR! 
I think it would need to be re-based later, when this PR: https://github.com/triton-inference-server/server/pull/6785 is merged (end of this week). 

I would like to also suggest to write a test, e.g. export `OTEL_SERVICE_NAME` before server start for opentelemetry unittests, and add a subroutine for `_test_trace`

oandreeva-nv:
@HennerM, please rebase this PR. 
As for tests, I think it would be enough for now to simply export environment variable ` OTEL_SERVICE_NAME`  [here](https://github.com/triton-inference-server/server/blob/7d9f6cdc1a2a31bf6650df85627b019023c725fa/qa/L0_trace/test.sh#L717),
start server, send request and grep `service name`, from collected traces. You can refer to how to do it [here](https://github.com/triton-inference-server/server/blob/7d9f6cdc1a2a31bf6650df85627b019023c725fa/qa/L0_trace/test.sh#L763-L826) , and sample grep command [here](https://github.com/triton-inference-server/server/blob/7d9f6cdc1a2a31bf6650df85627b019023c725fa/qa/L0_trace/test.sh#L888)

PR #6778, opened 2024-01-10
Add Python API Test

nnshah1:


qa/L0_python_api/test.sh
```diff
@@ -25,25 +25,35 @@
...
 
 RET=0
 
-rm -f $TEST_LOG
+rm -f $BINDING_TEST_LOG
 
 set +e
 
-python test_binding.py > $TEST_LOG 2>&1
+python test_binding.py > $BINDING_TEST_LOG 2>&1
```
rmccorm4, line 36:
```
python: can't open file '/opt/tritonserver/qa/L0_python_api/test_binding.py': [Errno 2] No such file or directory
python: can't open file '/opt/tritonserver/qa/L0_python_api/test_api.py': [Errno 2] No such file or directory
```
I think you need to either `cd test/` here before running, or modify where the tests get copied to.

nnshah1: thanks! resolved by changing the copy command

PR #6784, opened 2024-01-11
Enable shared memory leak detector on decoupled ensemble IO test

kthui:
Related PR: https://github.com/triton-inference-server/python_backend/pull/336

Enable shm leak detection on decoupled ensemble IO test.


kthui:
Will be part of another PR.

Issue #5853, opened 2023-05-25
How to pass traceparent in the OpenTelemetry tracing feature


bestend:
**Description**
I specified the trace mode as follows

```
"--trace-config=mode=opentelemetry"
```

The system I'm running is going through several phases, one of which includes triton.

When I pass the traceparent value from the previous server to triton, I expect triton to pass it to the opentelemetry protocol.

The method I tried was to insert the traceparent value in the header when calling triton over http.

```
-H 'Traceparent: 00-80e1afed08e019fc1110464cfa66635c-7a085853722dc6d2-01'
```
However, the value is not being passed correctly.

Here is the trace information I logged from triton to otel collector.

```
Resource SchemaURL:
Resource attributes:
     -> service.name: Str(karlo-triton<c6dfdb1a238c4aa4b4d4398a71d08290>)
     -> telemetry.sdk.version: Str(1.8.3)
     -> telemetry.sdk.name: Str(opentelemetry)
     -> telemetry.sdk.language: Str(cpp)
ScopeSpans #0
ScopeSpans SchemaURL:
InstrumentationScope triton-server
Span #0
    Trace ID : b1619b9ffb5cb1c741a46254817c3183
    Parent ID :
    ID : 8dc68e0febc217c0
    Name : InferRequest
    Kind : Server
    Start time : 2023-05-25 02:49:51.069104369 +0000 UTC
    End time : 2023-05-25 02:49:56.428076964 +0000 UTC
    Status code : Unset
    Status message :
```

Can anyone help me?

dyastremsky:
CC: @oandreeva-nv

oandreeva-nv:
@bestend Currently Triton provides limited support for OpenTelemetry and Context Propagation for OpenTelemetry is on our road map but is not supported yet.

bestend:
@oandreeva-nv Thank you for your confirmation.  Looking forward to seeing the feature added.

Legion2:
Hi, we also looking forward to triton supporting trace context propagation from the incoming GRPC requests. Our GRPC clients use otel context propagators to inject the relevant information into the GRPC metadata. This feature would allow us to see the triton traces as part (child span) of the incoming request. This is a critical part of a distributed tracing system.

oandreeva-nv:
@Legion2 Noted, thank you for the request

MatthieuToulemont:
Hello, I am delighted to see this added to the roadmap !! 

[...]

PR #6785, opened 2024-01-11
Support for Context Propagation for OTel trace mode

oandreeva-nv:
Resolves issue: https://github.com/triton-inference-server/server/issues/5853
I've picked up @HennerM work, he has CLA: https://github.com/triton-inference-server/server/issues/5853#issuecomment-1719863258

This PR adds 2 classes: `HttpTextMapCarrier` and `GrpcServerCarrier`, that take care of extracting OTel relevant header/metadata from an inference request. Next, when request comes to the server, OTel relevant context is extracted and stored in `TraceStartOptions start_options`.

`TraceStartOptions` was added for the following reason. Currently we can set up trace options through `rate` and `count`. However, users may want to trace every request, for which OTel trace was initiated on client side. In this case, it is hard to properly tune `rate` (I would say, even impossible). Thus, I've added logic to initiate trace every time a valid OTel context is passed and ignoring `rate` and `count`. 

Interestingly, we are allowing to start `tritonserver` with `rate=0` (checked locally). I've added a check for `rate_ !=0` in `TraceManager::TraceSetting::SampleTrace(bool force_sample)` and would like to keep this for now for the following reason. If users would like to trace only requests, for which trace was initiated on client side, they can start tritonserver with `rate=0`. I will follow up with a separate PR, which will introduce a dedicated trace-config option, something like : `--trace-config opentelemetry, context-based-tracing=ON`, which will eliminate current `rate=0` WAR, since this is slightly out of scope of this PR and current PR is already big.

I've re-factored tests. Extracted small sub-routines and made open telemetry tests easy to work with and to expand.

Additionally, there were issues with ABSEIL, so I removed it from OTel's installation: https://github.com/triton-inference-server/third_party/pull/52

### DIFF
--- a/qa/L0_trace/opentelemetry_unittest.py
+++ b/qa/L0_trace/opentelemetry_unittest.py
@@ -28,78 +28,105 @@ import sys
 
 sys.path.append("../common")
 import json
+import queue
 import re
+import shutil
+import subprocess
+import time
 import unittest
+from functools import partial
 
 import numpy as np
 import test_util as tu
 import tritonclient.grpc as grpcclient
 import tritonclient.http as httpclient
+from tritonclient.utils import InferenceServerException
 
-EXPECTED_NUM_SPANS = 16
-# OpenTelemetry OStream exporter sets `parent_span_id` to "0000000000000000",
-# if current span is a root span, i.e. there is no parent span.
-# https://github.com/open-telemetry/opentelemetry-cpp/blob/b7fd057185c4ed2dff507b859cbe058b7609fb4a/exporters/ostream/src/span_exporter.cc#L78C54-L78C68
-NO_PARENT_SPAN = "0000000000000000"
+NO_PARENT_SPAN_ID = ""
+COLLECTOR_TIMEOUT = 10
+
+
+def callback(user_data, result, error):
+    if error:
+        user_data.put(error)
+    else:
+        user_data.put(result)
+
+
+def prepare_data(client):
+    inputs = []
+    input0_data = np.full(shape=(1, 16), fill_value=-1, dtype=np.int32)
+    input1_data = np.full(shape=(1, 16), fill_value=-1, dtype=np.int32)
+
+    inputs.append(client.InferInput("INPUT0", [1, 16], "INT32"))
+    inputs.append(client.InferInput("INPUT1", [1, 16], "INT32"))
+
+    # Initialize the data
+    inputs[0].set_data_from_numpy(input0_data)
+    inputs[1].set_data_from_numpy(input1_data)
+
+    return inputs
+
+
+def send_bls_request(model_name="simple", headers=None):
+    with httpclient.InferenceServerClient("localhost:8000") as client:
+        inputs = prepare_data(httpclient)
+        inputs.append(httpclient.InferInput("MODEL_NAME", [1], "BYTES"))
+        inputs[-1].set_data_from_numpy(np.array([model_name], dtype=np.object_))
+        client.infer("bls_simple", inputs, headers=headers)
 
 
 class OpenTelemetryTest(tu.TestResultCollector):
     def setUp(self):
-        # Extracted spans are in json-like format, thus data needs to be
-        # post-processed, so that `json` could accept it for further
-        # processing
-        with open("trace_collector.log", "rt") as f:
-            data = f.read()
-            # Removing new lines and tabs around `{`
-            json_string = re.sub("\n\t{\n\t", "{", data)
-            # `resources` field is a dictionary, so adding `{` and`}`
-            # in the next 2 transformations, `instr-lib` is a next field,
-            # so whatever goes before it, belongs to `resources`.
-            json_string = re.sub(
-                "resources     : \n\t", "resources     : {\n\t", json_string
-            )
-            json_string = re.sub(
-                "\n  instr-lib     :", "}\n  instr-lib     :", json_string
-            )
-            # `json`` expects "key":"value" format, some fields in the
-            # data have empty string as value, so need to add `"",`
-            json_string = re.sub(": \n\t", ':"",', json_string)
-            json_string = re.sub(": \n", ':"",', json_string)
-            # Extracted data missing `,' after each key-value pair,
-            # which `json` exppects
-            json_string = re.sub("\n|\n\t", ",", json_string)
-            # Removing tabs
-            json_string = re.sub("\t", "", json_string)
-            # `json` expects each key and value have `"`'s, so adding them to
-            # every word/number/alpha-numeric entry
-            json_string = re.sub(r"\b([\w.-]+)\b", r'"\1"', json_string)
-            # `span kind`` represents one key
-            json_string = re.sub('"span" "kind"', '"span kind"', json_string)
-            # Removing extra `,`
-            json_string = re.sub("{,", "{", json_string)
-            json_string = re.sub(",}", "}", json_string)
-            # Adding `,` between dictionary entries
-            json_string = re.sub("}{", "},{", json_string)
-            # `events` is a list of dictionaries, `json` will accept it in the
-            # form of "events" : [{....}, {.....}, ...]
-            json_string = re.sub(
-                '"events"        : {', '"events"        : [{', json_string
-            )
-            # Closing `events`' list of dictionaries
-            json_string = re.sub('},  "links"', '}],  "links"', json_string)
-            # Last 2 symbols are not needed
-            json_string = json_string[:-2]
-            # Since now `json_string` is a string, which represents dictionaries,
-            # we  put it into one dictionary, so that `json` could read it as one.
-            json_string = '{ "spans" :[' + json_string + "] }"
-            self.spans = json.loads(json_string)["spans"]
+        self.collector_subprocess = subprocess.Popen(
+            ["./otelcol", "--config", "./trace-config.yaml"]
+        )
+
+        time.sleep(5)
+
+        self.filename = "collected_traces.json"
+        self.client_headers = dict(
+            {"traceparent": "00-0af7651916cd43dd8448eb211c12666c-b7ad6b7169242424-01"}
+        )
 
         self.simple_model_name = "simple"
         self.ensemble_model_name = "ensemble_add_sub_int32_int32_int32"
         self.bls_model_name = "bls_simple"
         self.root_span = "InferRequest"
 
+    def tearDown(self):
+        self.collector_subprocess.kill()
+        self.collector_subprocess.wait()
+        time.sleep(5)
+        test_name = unittest.TestCase.id(self).split(".")[-1]
+        shutil.copyfile(self.filename, self.filename + "_" + test_name + ".log")
+
+    def _parse_trace_log(self, trace_log):
+        """
+        Helper function that parses file, containing collected traces.
+
+        Args:
+            trace_log (str): Name of a file, containing all traces.
+
+        Returns:
+            traces (List[dict]): List of json objects, representing each span.
+        """
+        traces = []
+        with open(trace_log) as f:
+            for json_obj in f:
+                entry = json.loads(json_obj)
+                traces.append(entry)
+
+        return traces
+
     def _check_events(self, span_name, events):
+        """
+        Helper function that verifies passed events contain expected entries.
+
+        Args:
+            span_name (str): name of a span.
+            events (List[str]): list of event names, collected for the span with the name `span_name`.
+        """
         root_events_http = [
             "HTTP_RECV_START",
             "HTTP_RECV_END",
@@ -144,7 +171,11 @@ class OpenTelemetryTest(tu.TestResultCollector):
             self.assertFalse(all(entry in events for entry in request_events))
             self.assertFalse(all(entry in events for entry in compute_events))
 
-        elif span_name == self.simple_model_name:
+        elif span_name in [
+            self.simple_model_name,
+            self.ensemble_model_name,
+            self.bls_model_name,
+        ]:
             # Check that all request related events (and only them)
             # are recorded in request span
             self.assertTrue(all(entry in events for entry in request_events))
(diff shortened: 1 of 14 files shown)


qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -28,78 +28,112 @@
...
+
+    # Initialize the data
+    input_data = np.expand_dims(input_data, axis=0)
+
+    if is_binary:
+        inputs[0].set_data_from_numpy(input_data)
+        inputs[1].set_data_from_numpy(input_data)
+    else:
+        inputs[0].set_data_from_numpy(input_data, binary_data=is_binary)
+        inputs[1].set_data_from_numpy(input_data, binary_data=is_binary)
```
rmccorm4, line 73: this is the same logically right?
```suggestion
    inputs[0].set_data_from_numpy(input_data, binary_data=is_binary)
    inputs[1].set_data_from_numpy(input_data, binary_data=is_binary)
```

oandreeva-nv: suggested code will not work for grpc (tried myself), it mentions that for grpc client `binary_data` is not an option

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -28,78 +28,112 @@
...
+        user_data.put(result)
+
+
+def prepare_data(client, is_binary=True):
+    inputs = []
+    input_data = np.array(
+        [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15], dtype=np.int32
+    )
+    inputs.append(client.InferInput("INPUT0", [1, 16], "INT32"))
+    inputs.append(client.InferInput("INPUT1", [1, 16], "INT32"))
```
rmccorm4, line 63: Can use `np.arange` over hard-coded list for the same
```suggestion
    dim = 16
    input_data = np.arange(dim, dtype=np.int32)
    inputs.append(client.InferInput("INPUT0", [1, dim], "INT32"))
    inputs.append(client.InferInput("INPUT1", [1, dim], "INT32"))
```

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -28,78 +28,112 @@
...
-            self.spans = json.loads(json_string)["spans"]
+        self.collector_subprocess = subprocess.Popen(
+            ["./otelcol", "--config", "./trace-config.yaml"]
+        )
+
+        time.sleep(5)
+
+        self.filename = "collected_traces.json"
+        self.client_headers = dict(
+            {"traceparent": "00-0af7651916cd43dd8448eb211c12666c-b7ad6b7169242424-01"}
```
rmccorm4, line 96: What's the meaning of this traceparent value? Can you assign to a constant or add a comment?

oandreeva-nv: This simulates OTel context being injected on client side: https://www.w3.org/TR/trace-context/#design-overview

Essentially, this is what will be injected to headers/metadata on the client side, and what will be extracted on the server side. The second long value is `trace id`, the third - active `span id`, which should become a parent span.

Since there server side changes are related to extraction of this info through `Propagator`, by simply adding this header during tests, should suffice. I'll later add an example client with OpenTelemetry trace initialization for reference

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -144,7 +178,11 @@ def _check_events(self, span_name, events):
             self.assertFalse(all(entry in events for entry in request_events))
             self.assertFalse(all(entry in events for entry in compute_events))
 
-        elif span_name == self.simple_model_name:
+        elif span_name in [
+            self.simple_model_name,
+            self.ensemble_model_name,
+            self.bls_model_name,
+        ]:
```
rmccorm4, line 185: Maybe move these into another attribute so there's only 1 place to update if new models are added later:
```suggestion
        elif span_name in self.test_models:
```

```
self.test_models = [
            self.simple_model_name,
            self.ensemble_model_name,
            self.bls_model_name,
        ]

File: qa/L0_trace/opentelemetry_unittest.py [hunk 5]
```diff
@@ -28,78 +28,112 @@
...
+        Helper function that parses file, containing collected traces.
+
+        Args:
+            trace_log (str): Name of a file, containing all traces.
+
+        Returns:
+            traces (List[dict]): List of json objects, representing each span.
+        """
+        traces = []
+        with open(trace_log) as f:
```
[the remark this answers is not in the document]
oandreeva-nv: I've upgraded OTel collector and they've added functionality to extract spans in a json format to file, so I've adapted tests

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: I'll add a comment to tests

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -153,121 +191,560 @@ def _check_events(self, span_name, events):
...
+        """
+        rex_name_field = re.compile("(?<='name': ')[a-zA-Z0-9_\- ]+(?=')")
+
+        span_names = []
+        for span in spans:
+            # Check that collected spans have proper events recorded
+            span_name = span[0]["name"]
+            span_names.append(span_name)
+            span_events = span[0]["events"]
+            event_names_only = rex_name_field.findall(str(span_events))
```
rmccorm4, line 251: Can you share sample event format that's getting parsed? Just curious what's being parsed by the regex.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -153,121 +191,560 @@ def _check_events(self, span_name, events):
...
+                   parents and children in the dictionary form:
+                        <parent_span_name> (str) : <children_names> (List[str])
+        """
+        seen_spans = dict({})
+        for span in spans:
+            cur_span = span[0]["spanId"]
+            seen_spans[cur_span] = span[0]["name"]
+
+        parent_child_dict = dict({})
+        for span in spans:
```
rmccorm4, line 285: Any reason for this `dict({})` syntax over `{}` or `dict()`?
```suggestion
        seen_spans = {}
        for span in spans:
            cur_span = span[0]["spanId"]
            seen_spans[cur_span] = span[0]["name"]

        parent_child_dict = {}
        for span in spans:
```

oandreeva-nv: Nicely spotted. I can incorporate this
No particular reason in this case

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -153,121 +191,560 @@ def _check_events(self, span_name, events):
...
+        for a `simple` model, HTTP client and context propagation,
+        i.e. client specifies OTel headers, defined in `self.client_headers`.
+        """
+        triton_client_http = httpclient.InferenceServerClient(
+            "localhost:8000", verbose=True
+        )
+        inputs = prepare_data(httpclient)
+        triton_client_http.infer(
+            self.simple_model_name, inputs, headers=self.client_headers
+        )
```
rmccorm4, line 501: **This is totally optional "follow-up PR" material** if you like the idea, but I think a lot of these functions for:
1. http/grpc
2. with/without context propogation
3. simple/ensemble/bls

Could be condensed a lot with something like `unittest.subtest` or `parameterized`:
```python
def test_trace(self):
  clients = [httpclient, grpcclient]
  header_list = [None, self._client_headers]
  models = self.test_models # [self.simple_model_name, self.ensemble, self.bls ...]

  permutations = itertools.product(clients, header_list, models)
  for p in permutations:
    client_type, headers, model = p[0], p[1], p[2]
    with self.subtest(permutation=p):
        client = client_type.InferenceServerClient(...) # can probably be smart and include url somewhere
        inputs = prepare_data(client_type) 
        client.infer(model, inputs, headers)
        self._test_trace(model, headers=headers)
```

Doesn't have to be exactly like this, but seems like there's room for cleanup or re-using logic

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: image

qa/L0_trace/trace-config.yaml
```diff
@@ -34,12 +34,18 @@ receivers:
...
+processors:
+  batch:
+    send_batch_size: 10
+    timeout: 10s
+
 exporters:
-  logging:
-    verbosity: detailed
+  file:
+    path: ./collected_traces.json
```
rmccorm4, line 44: Ah I think this answered my other question about json/parsing.

qa/L0_trace/test.sh
```diff
@@ -698,68 +698,40 @@ set +e
...
+MODEL_PATH="/opt/ml/models/123456789abcdefghi/model"
+rm -r ${MODEL_PATH}
+mkdir -p "${MODEL_PATH}"
+cp -r $DATADIR/$MODELBASE/* ${MODEL_PATH} && \
+    rm -r ${MODEL_PATH}/2 && rm -r ${MODEL_PATH}/3 && \
+        sed -i "s/onnx_int32_int32_int32/simple/" ${MODEL_PATH}/config.pbtxt
+
+
+SERVER_ARGS="--allow-sagemaker=true --model-control-mode=explicit \
+                --load-model=simple --load-model="ensemble_add_sub_int32_int32_int32" \
```
rmccorm4, line 727: Is this nested quote for ensemble model name OK?

oandreeva-nv: Yes, from recent local run:
```
I0111 22:47:57.913510 1734 server.cc:676]
+------------------------------------+---------+--------+
| Model                              | Version | Status |
+------------------------------------+---------+--------+
| bls_simple                         | 1       | READY  |
| ensemble_add_sub_int32_int32_int32 | 1       | READY  |
| simple                             | 1       | READY  |
+------------------------------------+---------+--------+
```
And from CI's logs:
```

I0111 19:03:11.859289 3243 server.cc:676] 
+------------------------------------+---------+--------+
| Model                              | Version | Status |
+------------------------------------+---------+--------+
| bls_simple                         | 1       | READY  |
| ensemble_add_sub_int32_int32_int32 | 1       | READY  |
| simple                             | 1       | READY  |
+------------------------------------+---------+--------+
```
It is getting loaded, but I'll remove quotes for consistency

src/grpc/infer_handler.cc
```diff
@@ -916,8 +916,29 @@ ModelInferHandler::Execute(InferHandler::State* state)
...
+          otel_trace_api::GetSpan(start_options.propagated_context)
+              ->GetContext();
+      if (span_context.IsValid()) {
+        start_options.force_sample = true;
+      }
+#else
+      LOG_ERROR << "Unsupported trace mode: "
+                << TraceManager::InferenceTraceModeString(mode);
+#endif
+    }
```
rmccorm4, line 939: (can be follow-up) Could it be moved to a helper?
```cpp
auto start_options = GetTraceStartOptions(state, request);
state->trace_ = std::move(
        trace_manager_->SampleTrace(request.model_name(), start_options));
```

src/grpc/stream_infer_handler.cc
```diff
@@ -309,8 +309,29 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
+            otel_trace_api::GetSpan(start_options.propagated_context)
+                ->GetContext();
+        if (span_context.IsValid()) {
+          start_options.force_sample = true;
+        }
+#else
+        LOG_ERROR << "Unsupported trace mode: "
+                  << TraceManager::InferenceTraceModeString(mode);
+#endif
+      }
```
rmccorm4, line 332: can it re-use code/helper from infer_handler?

src/http_server.cc
```diff
@@ -3055,11 +3055,30 @@ HTTPAPIServer::StartTrace(
...
+    otel_trace_api::SpanContext span_context =
+        otel_trace_api::GetSpan(start_options.propagated_context)->GetContext();
+    if (span_context.IsValid()) {
+      start_options.force_sample = true;
+    }
+#else
+    LOG_ERROR << "Unsupported trace mode: "
+              << TraceManager::InferenceTraceModeString(mode);
+#endif
+  }
```
rmccorm4, line 3077: Same as https://github.com/triton-inference-server/server/pull/6785/files#r1449553152

src/tracer.h
```diff
@@ -397,7 +414,7 @@ class TraceManager {
         const std::unordered_map<uint64_t, std::unique_ptr<std::stringstream>>&
             streams);
 
-    std::shared_ptr<Trace> SampleTrace();
+    std::shared_ptr<Trace> SampleTrace(bool force_sample = false);
```
rmccorm4, line 417: Add comment describing reason for force_sample

src/tracer.cc
```diff
@@ -1022,10 +1056,10 @@ TraceManager::TraceFile::SaveTraces(
 }
 
 std::shared_ptr<TraceManager::Trace>
-TraceManager::TraceSetting::SampleTrace()
+TraceManager::TraceSetting::SampleTrace(bool force_sample)
 {
   bool create_trace = false;
-  {
+  if (rate_ != 0) {
```
rmccorm4, line 1062: Add comment on rate==0 WAR

src/grpc/infer_handler.cc
```diff
@@ -916,8 +916,29 @@ ModelInferHandler::Execute(InferHandler::State* state)
   if (err == nullptr) {
     TRITONSERVER_InferenceTrace* triton_trace = nullptr;
 #ifdef TRITON_ENABLE_TRACING
-    state->trace_ =
-        std::move(trace_manager_->SampleTrace(request.model_name()));
+    TraceStartOptions start_options;
+    InferenceTraceMode mode = TRACE_MODE_TRITON;
+    trace_manager_->GetTraceMode(request.model_name(), &mode);
```
rmccorm4, line 921: Looks like this adds a lock to the hot path. Will DM you to discuss further

rmccorm4 (commented):
Logic and tests LGTM - just want to discuss this one comment: https://github.com/triton-inference-server/server/pull/6785/files#r1449559949.

Most nits/cleanups/helper function comments can be a follow-up PR if it will impact code freeze and isn't changing any logic.

src/grpc/infer_handler.cc
oandreeva-nv: Yes, it most likely should

src/grpc/infer_handler.cc
oandreeva-nv: Technically, I can do the following re-factoring:
`std::shared_ptr<TraceSetting> trace_setting` can become a part of `TraceStartOptions`. 

And then when we initialize `start_options` on line 919, I can get `trace_setting` for current model through [the steps](https://github.com/triton-inference-server/server/blob/main/src/tracer.cc#L296-L301). This way, lock is happenning only once (as it is now) and then we just pass start options to `TraceManager::SampleTrace` , where we move on to [this line](https://github.com/triton-inference-server/server/pull/6785/files#diff-c05667609085c993fd282ff18d0429945c44492b95c0d01fcbeec6946fda2db2R316) since `trace_setting` is already obtained

oandreeva-nv: @GuanLuo I would like to have your opinion on this as well if possible

src/http_server.cc
```diff
@@ -3055,11 +3055,30 @@ HTTPAPIServer::StartTrace(
...
+  TraceStartOptions start_options;
+  InferenceTraceMode mode = TRACE_MODE_TRITON;
+  trace_manager_->GetTraceMode(model_name, &mode);
+  if (mode == TRACE_MODE_OPENTELEMETRY) {
+#ifndef _WIN32
+    const HttpTextMapCarrier carrier(req->headers_in);
+    auto prop =
+        otel_cntxt::propagation::GlobalTextMapPropagator::GetGlobalPropagator();
+    auto ctxt = opentelemetry::context::Context();
+    start_options.propagated_context = prop->Extract(carrier, ctxt);
```
GuanLuo, line 3067: `propagated_context` will be set to an valid default if there is no context from client?

src/http_server.cc
```diff
@@ -3055,11 +3055,30 @@ HTTPAPIServer::StartTrace(
...
+    otel_trace_api::SpanContext span_context =
+        otel_trace_api::GetSpan(start_options.propagated_context)->GetContext();
+    if (span_context.IsValid()) {
```
GuanLuo, line 3070: Will this always be evaluated to `true` in the case of no client context? And causes force sampling as long as the trace is in OpenTelemetry mode?

GuanLuo (commented):
Overall looks good, leave comment for clarification

src/http_server.cc
oandreeva-nv: Good question. Here we use an instance of class [`HttpTraceContext`](https://github.com/open-telemetry/opentelemetry-cpp/blob/c4f39f2be8109fd1a3e047677c09cf47954b92db/api/include/opentelemetry/trace/propagation/http_trace_context.h#L38C1-L38C23) as a global propagator. We set it [here](https://github.com/triton-inference-server/server/pull/6785/files#diff-c05667609085c993fd282ff18d0429945c44492b95c0d01fcbeec6946fda2db2R427)

[`Extract`](https://github.com/open-telemetry/opentelemetry-cpp/blob/c4f39f2be8109fd1a3e047677c09cf47954b92db/api/include/opentelemetry/trace/propagation/http_trace_context.h#L52) method will use [`ExtractImpl`](https://github.com/open-telemetry/opentelemetry-cpp/blob/c4f39f2be8109fd1a3e047677c09cf47954b92db/api/include/opentelemetry/trace/propagation/http_trace_context.h#L163), meaning our `Get`, e.g. HTTP [Get](https://github.com/triton-inference-server/server/pull/6785/files#diff-54db4fdff0f413084068211acf3ab675ae22c4fec2b02a680d9809dbeb756172R151), to extract `traceparent` and `tracestate` info from headers/metadata. In none is found, it will return invalid `SpanContext` and [this](https://github.com/open-telemetry/opentelemetry-cpp/blob/c4f39f2be8109fd1a3e047677c09cf47954b92db/api/include/opentelemetry/trace/propagation/http_trace_context.h#L56) line will create invalid span. So in the returned context there will be `invalid span` stored.

I believe the reason, why everything is working when no context is passed, is because by default `StartSpanOptions.parent` is initialized to invalid context [here](https://github.com/open-telemetry/opentelemetry-cpp/blob/c4f39f2be8109fd1a3e047677c09cf47954b92db/api/include/opentelemetry/trace/span_startoptions.h#L62), 
So essentially, OpenTelemetry handles invalid spans as parents on their end to start a parent-less span. 

However, I will add changes to handle this behavior better.

src/http_server.cc
oandreeva-nv: I've touched this in the above comment. When no context is propagated, and invalid span will be stored in `start_options.propagated_context`, so `span_context.IsValid()` will return false.

This behavior is tested in `test_http_trace_triggered`, where I set `rate=5` and send 5 requests without `otel` headers -> only one is traced, then send 5 requests with `otel` headers and check that all 5 are traced

src/grpc/infer_handler.cc
oandreeva-nv: discussed offline

src/http_server.cc
```diff
@@ -3055,11 +3055,13 @@ HTTPAPIServer::StartTrace(
     TRITONSERVER_InferenceTrace** triton_trace)
 {
 #ifdef TRITON_ENABLE_TRACING
+  HttpTextMapCarrier carrier(req->headers_in);
```
Tabrizian, line 3058: do these carriers just record all the headers associated with gRPC/HTTP requests?

src/tracer.cc
```diff
@@ -289,19 +288,70 @@ TraceManager::GetTraceSetting(
...
-    trace_setting =
-        (m_it == model_settings_.end()) ? global_setting_ : m_it->second;
+  std::lock_guard<std::mutex> r_lk(r_mu_);
+  auto m_it = model_settings_.find(model_name);
+  trace_setting =
+      (m_it == model_settings_.end()) ? global_setting_ : m_it->second;
+}
+
+TraceManager::TraceStartOptions
+TraceManager::GetTraceStartOptions(
```
Tabrizian, line 302: As a future refactor it might be worth having separate classes for `TritonTrace` and `OTelTrace`. This could help with not having to propagate `#ifdefs` for WiNDOWS and checking for open telemetry in every tracing function.

qa/L0_trace/test.sh
```diff
@@ -698,68 +698,40 @@ set +e
...
+        sed -i "s/onnx_int32_int32_int32/simple/" ${MODEL_PATH}/config.pbtxt
+
+
+SERVER_ARGS="--allow-sagemaker=true --model-control-mode=explicit \
+                --load-model=simple --load-model=ensemble_add_sub_int32_int32_int32 \
+                --load-model=bls_simple --trace-config=level=TIMESTAMPS \
+                --trace-config=rate=1 --trace-config=count=-1 --trace-config=mode=opentelemetry \
                 --trace-config=opentelemetry,resource=test.key=test.value \
                 --trace-config=opentelemetry,resource=service.name=test_triton \
+                --trace-config=opentelemetry,url=localhost:$OTLP_PORT/v1/traces \
```
Tabrizian, line 732: I wonder how the test used to work without url configuration?

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -153,121 +198,560 @@ def _check_events(self, span_name, events):
...
+                   containing root span info. For this test `traceID`
+                   and `parentSpanId` are required.
+            expected_parent_span_dict (dict): dictionary, containing expected
+                   parents and children in the dictionary form:
+                        <parent_span_name> (str) : <children_names> (List[str])
+        """
+        parent_span_id = NO_PARENT_SPAN_ID
+
+        if headers != None:
+            parent_span_id = headers["traceparent"].split("-")[2]
```
Tabrizian, line 324: So context propagation is making sure that the headers associated with the request are included in the trace?

src/tracer.cc
```diff
@@ -289,19 +288,70 @@ TraceManager::GetTraceSetting(
...
+TraceManager::GetTraceSetting(
+    const std::string& model_name, std::shared_ptr<TraceSetting>& trace_setting)
 {
-  std::shared_ptr<TraceSetting> trace_setting;
-  {
-    std::lock_guard<std::mutex> r_lk(r_mu_);
-    auto m_it = model_settings_.find(model_name);
-    trace_setting =
-        (m_it == model_settings_.end()) ? global_setting_ : m_it->second;
+  std::lock_guard<std::mutex> r_lk(r_mu_);
```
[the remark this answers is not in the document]
oandreeva-nv: Currently, I am re-using already existing locks. I believe next step (since we've moved to C++17) not only to rename them, but also change to dedicated read locks (`std:shared_lock` I believe)

src/http_server.cc
oandreeva-nv: This one is specific to `HTTP` requests. `Carries` are initialized with headers/metadata, and then OTel propagator will use them to extract relevant headers. 
In this comment I provide a deep dive of what is going on:
https://github.com/triton-inference-server/server/pull/6785#discussion_r1449664084

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: I am not sure if `making sure` is the right word. 

Context propagation describes a set of operations that needs to be performed on client and server side that results in a unified trace. The steps are:
On client side, headers/metadata needs to be populated with `traceId` and `spanId` of parent's trace (in a special format) , there is more info, that gets injected, but I'l omit it for now.

On server side, these headers should be extracted, and started trace should have the same `traceId` as passed, and newly started spans should have parent span with `id`, passed in headers.

If nothing found in headers/metadata, we start a fresh new trace with new `traceId`

src/tracer.cc
oandreeva-nv: Yes, I agree and similar idea was added to ticket: https://jirasw.nvidia.com/browse/DLIS-4803

qa/L0_trace/test.sh
oandreeva-nv: Previously, setting exporter's port was tested as a stadalone bash test here: https://github.com/triton-inference-server/server/pull/6785/files#diff-27e9ca4cd8f292f0e5a94bbabb243e40fe0d47cd55164a0cb49ae6e093c9fbe7L710-L741

and `opentelemetry_unittests.py` was testing spans with OStreamExporter, that was used if environment variable was set

qa/L0_trace/opentelemetry_unittest.py
rmccorm4: Ah right, I forgot grpcclient was asymmetric like that... No worries then.

qa/L0_trace/test.sh
Tabrizian: got it, thanks Olga!

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: In this case it is easier to pass `span[0]["events"]` as a string and do regex.

`span[0]["events"]` is actually a list of dictionaries (on image it is seen better), that is why for this test I didn't want to parse `span[0]["events"]` list to collect all names.

Tabrizian (dismissed):

qa/L0_trace/opentelemetry_unittest.py
Tabrizian: Got it, thanks for correcting my understanding.

src/CMakeLists.txt
```diff
@@ -507,9 +507,9 @@ if(${TRITON_ENABLE_TRACING})
     tracer.cc tracer.h
   )
 
-  if (NOT WIN32)
-    target_compile_features(tracing-library PRIVATE cxx_std_${TRITON_MIN_CXX_STANDARD})
+  target_compile_features(tracing-library PRIVATE cxx_std_${TRITON_MIN_CXX_STANDARD})
```
oandreeva-nv, line 510: I've noticed that this line was previously ignored by Windows, so in this PR as well

qa/L0_trace/opentelemetry_unittest.py
rmccorm4: Isn't it same? 
image

oandreeva-nv: nice, bad on my part. Will remove `regex`

src/tracer.cc
```diff
@@ -1022,10 +1069,13 @@ TraceManager::TraceFile::SaveTraces(
...
 std::shared_ptr<TraceManager::Trace>
-TraceManager::TraceSetting::SampleTrace()
+TraceManager::TraceSetting::SampleTrace(bool force_sample)
 {
   bool create_trace = false;
-  {
+  // [FIXME: DLIS-6033]
+  // A current WAR for initiating trace based on propagated context only
+  // Currently this is implemented through setting trace rate = 0
+  if (!force_sample && rate_ != 0) {
```
rmccorm4, line 1078: So if I'm understanding correctly, for `rate_ > 0` - we will always trace when a context is propagated (client header provided -> force_sample=True) - but these propogation-initiated traces will not affect the counts/logic usedwhen deciding to do normal traces (ex: every 1000th request) ? 

So for example with default `rate == 1000`, before:
- request 0 and request 1000 will be traced. 

But now. if I added client headers to some intermediate request say request 7, then instead:
-  request 0, request 7 (context propagation), and request 1001 would get traced?

If so, will this have any impact or confusing behavior on related tools like trace_summary tool or others?

rmccorm4: Also out of curiosity, can `rate_ == 0` here? I didn't look through all the code, but I noticed it would probably hit a division by zero / floating point exception here without this check, so I wanted to try it out:
```
tritonserver --model-store /mnt/triton/models --trace-rate 0 --trace-level TIMESTAMPS
```
and I sent some inferences, but didn't get any errors.

oandreeva-nv: Yes, that is also what I found. I think I talk about it in the description of this PR. `% by 0` was definitely happening before this PR, and I have no idea why it wasn't segfaulting. With this PR, if `rate_==0`, we are skipping this block.

oandreeva-nv: Regarding earlier comment: Good catch. We've discussed with Guan to not update counters, when tracing is happening "by force". Let me correct this.

rmccorm4 (approved):
Nice work. cleanup of traceStartOptions helpers into trace manager looks great!

src/tracer.cc
oandreeva-nv: @rmccorm4 and @GuanLuo I've decided to keep updating counters as is (if `rate_` > 0) for now. I've started checking all parameters and it seems like e.g `created_` is not used, so this area would benefit from a clean up.
I made it a scope of this ticket: DLIS-6033

GuanLuo (approved):

PR #6801, opened 2024-01-17
Fix documentaation genarion

mc-nv:


docs/Dockerfile.docs
```diff
@@ -27,27 +27,50 @@
...
+      attrs  \
+      breathe \
+      docutils \
+      exhale \
+      ipython \
+      myst-nb \
+      nbclient \
+      nbsphinx \
+      rst-to-myst \
+      sphinx==5.0.0 \
```
mc-nv, line 66: @nnshah1  do we actually need to keep specific version on the packages for this task?

docs/Dockerfile.docs
```diff
@@ -27,27 +27,50 @@
...
+    && rm /tmp/protoc-gen-doc.tar.gz
 
 # install sphinx et al
-RUN pip3 install sphinx==4.5.0 nbclient==0.5.13 \
-    docutils==0.16 ablog==0.10.33.post1 myst-nb==0.17.2 rst-to-myst==0.3.4 nbsphinx==0.8.8 \
-    sphinx-book-theme==0.3.2 sphinx-copybutton==0.5.2 sphinx-design==0.4.1 sphinx-prompt==1.5.0  \
-    sphinxcontrib-bibtex==2.5.0 sphinx-tabs==3.2.0 \
-    exhale==0.2.3 breathe==4.14.1 sphinx-sitemap==2.5.0 ipython==8.12.1 attrs==21.4.0
+RUN pip3 install \
+      ablog \
```
nnshah1, line 56: generally - fixing the versions is better / safer - I would prefer to selectively upgrade 
previously we ran into an isssue where we had a breaking change come in because the dependencies were not pinned.

GuanLuo (approved):

nnshah1 (changes requested):

docs/Dockerfile.docs
mc-nv: I pinned only sphinx version for now, supposing that python pip module is smart enough to install only packages with a compatible version.
But there is no right path to do it, we may not pin version and start getting in trouble or pin it and get the issue with dropped support for it…

IMO this image is kind of side car for us, we shouldn’t worry much about it.

nnshah1 (approved):

Issue #6745, opened 2023-12-27
[feature request] Static (or mostly static) build instructions (currently build with only  `libboost_filesystem.a` fails)


vadimkantorov:
Uncommenting `sudo rm /lib/*/libboost_filesystem.so*` makes the build fail, but without any specific error message (`libboost_filesystem1.81.a` is indeed existing).

Starting with `r23.12`, `tritonserver` started to require `libboost_filesystem1.81.so`, not just in build time, but also in runtime

It would be nice to have instructions on how to get a static (or mostly static) build to avoid the need for a very recent libboost. I've been using the following GitHub Actions script to obtain a built `tritonserver` binary for reuse on end cpu-only system for testing some python-backend models.

```yaml
name: buildtritoninferenceserver
on: workflow_dispatch

jobs:

  buildtritoninferenceserver:
    runs-on: ubuntu-22.04
    steps:
      - name: Install Prerequisites
        run:  sudo add-apt-repository -y ppa:mhier/libboost-latest && sudo apt-get update && sudo apt-get install -y git build-essential rapidjson-dev libssl-dev libre2-dev libb64-dev libarchive-dev libnuma-dev libboost1.81-dev libboost-filesystem1.81-dev

#      - name: Keep only static libboost_filesystem.a
#        run:  sudo rm /lib/*/libboost_filesystem.so* && ls /lib/*/libboost_filesystem*

      - name: Clone Triton
        run: git clone https://github.com/triton-inference-server/server --branch r23.12 --single-branch --depth 1

      - name: Build Triton
        run: cd server && python ./build.py -v --no-container-build --enable-logging --enable-stats --enable-tracing --build-dir="$PWD/build" --backend python --backend=ensemble --extra-core-cmake-arg=TRITON_ENABLE_GRPC=OFF --extra-core-cmake-arg=TRITON_ENABLE_HTTP=ON  --extra-core-cmake-arg=TRITON_ENABLE_ENSEMBLE=ON

      - name: Archive artifacts
        uses: actions/upload-artifact@v2
        with:
          name: opt
          path: server/build/opt/
```

Tabrizian:
@rmccorm4 any ideas how to fix this issue? Also, we recently upgraded the Triton server CMake files to work with C++17 so we may remove this dependency in the near future.

rmccorm4:
[...]

PR #6810, opened 2024-01-19
Remove boost::filesystem from builds

rmccorm4:
Server: https://github.com/triton-inference-server/server/pull/6810
Core: https://github.com/triton-inference-server/core/pull/319

Fixes: https://github.com/triton-inference-server/server/issues/6745

### DIFF
--- a/build.py
+++ b/build.py
@@ -974,7 +974,6 @@ RUN pip3 install --upgrade pip && \
 RUN wget -O /tmp/boost.tar.gz \
         https://archives.boost.io/release/1.80.0/source/boost_1_80_0.tar.gz && \
     (cd /tmp && tar xzf boost.tar.gz) && \
-    cd /tmp/boost_1_80_0 && ./bootstrap.sh --prefix=/usr && ./b2 install && \
     mv /tmp/boost_1_80_0/boost /usr/include/boost
 
 # Server build requires recent version of CMake (FetchContent required)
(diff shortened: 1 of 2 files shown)


build.py
```diff
@@ -974,7 +974,6 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
 RUN wget -O /tmp/boost.tar.gz \
         https://archives.boost.io/release/1.80.0/source/boost_1_80_0.tar.gz && \
     (cd /tmp && tar xzf boost.tar.gz) && \
-    cd /tmp/boost_1_80_0 && ./bootstrap.sh --prefix=/usr && ./b2 install && \
```
oandreeva-nv, line 977: May I ask you to briefly clarify where else `boost` is used and why we are keeping it here?

rmccorm4: We use the header-only boost libs (`span` in cache, `interprocess_lock` in python backend, `stacktrace` in server, and some others) in several places. `boost::filesystem` was the only one that needed a runtime shared library requiring this extra `bootstrap+install` step.

oandreeva-nv: Got it, thanks

oandreeva-nv (approved):

rmccorm4:
Passed L0_model_config and the relevant `test_file_override_security` subtest of L0_lifecycle.

build.py
vadimkantorov: Is there still an external dependency on libboost? Or could these header-only libs be included in-tree?

vadimkantorov:
boost is now not required, right? starting from what triton release?

thanks!

PR #6823, opened 2024-01-23
Add unit test reports to L0_json, L0_metrics, L0_response_cache, L0_backend_python

rmccorm4:
- Generates junit test reports from various test and subtests. This is not an exhaustive list of possible CI tests, just an initial set.
    - Had to remove use of `tu.TestResultCollector` as it interfered with `pytest` usage. Wasn't able to find a quick reason/solution as to why - it involved the use of `super()` and certain expected object types. Reverting the classes to inheriting from `unittest.TestCase` directly fixed the issue.
    - Similarly, removed the use of `check_test_results` since the `tu.TestResultCollector` was no longer generating the text file with pass/fail counts. For the short term, I think the numbers shown in gitlab from the reports can replace this for now. Longer term, I think we can modify `check_test_results` to understand the `.xml` reports as well, rather than our custom `.json`.
  

Example report generated on gitlab side:
image


krishung5 (approved):
Lgtm. Great improvements!

PR #6830, opened 2024-01-25
Python Backend Windows Support

fpetrini15:
**Goal:** Modify the test scripts in L0_backend_python to be compatible with the windows port.

_Note:_ Tests and subtests that make use of unsupported features such as GPU tensors, custom execution environments, and the model load/unload API have been stripped out and tagged with the corresponding Jira ticket intended to re-introduce them.

The bulk of these changes surround installing dependencies within the Windows virtualenv, using the non-default `localhost` client, killing servers with our platform-agnostic function `kill_server`, and stripping out tests based on whether the testing platform is Windows.

PYBE changes: https://github.com/triton-inference-server/python_backend/pull/294


qa/L0_backend_python/test.sh
```diff
@@ -35,23 +35,41 @@ if [ -z "$REPO_VERSION" ]; then
     exit 1
 fi
 
-DATADIR=${DATADIR:="/data/inferenceserver/${REPO_VERSION}"}
+# On windows the paths invoked by the script (running in WSL) must use
+# /mnt/c when needed but the paths on the tritonserver command-line
+# must be C:/ style.
+export TEST_WINDOWS=0
+if [[ "$(< /proc/sys/kernel/osrelease)" == *microsoft* ]]; then
```
kthui, line 42: I wonder why don't we just let the CI provide the `TEST_WINDOWS` flag, similar to how the `TEST_JETSON` is provided?

qa/python_models/init_args/model.py
```diff
@@ -36,10 +37,18 @@ def check_init_args(args):
...
+    if sys.platform == "win32":
+        TRITON_DIR = os.getenv("TRITON_DIR", "c:\\tritonserver").replace("/", "\\")
+        expected_args["model_repository"] = (
+            TRITON_DIR + "\\qa\\L0_backend_python\\models\\init_args"
+        )
+    else:
+        expected_args["model_repository"] = (
+            os.getenv("TRITON_DIR", "/opt/tritonserver")
+            + "/qa/L0_backend_python/models/init_args"
+        )
```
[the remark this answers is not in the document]
fpetrini15: Very slick suggestion 🧠

qa/L0_backend_python/test.sh
fpetrini15: I suppose we could do this, I was following the [previous convention](https://github.com/triton-inference-server/server/blob/main/qa/L0_sequence_batcher/test.sh#L96) from our existing Windows tests.

qa/L0_backend_python/python_test.py
```diff
@@ -39,15 +39,20 @@
 import tritonclient.http as httpclient
 from tritonclient.utils import *
 
+# By default, find tritonserver on "localhost", but for windows tests
+# we overwrite the IP address with the TRITONSERVER_IPADDR envvar
+_tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR", "localhost")
+
 TEST_JETSON = bool(int(os.environ.get("TEST_JETSON", 0)))
```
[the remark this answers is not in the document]
fpetrini15: Is the rationale for this so that we don't confused these variables with constants?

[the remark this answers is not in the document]
kthui: It is mainly for consistency in style, for example
```
_tritonserver_ipaddr =     os.environ.get("TRITONSERVER_IPADDR", "localhost")
TEST_JETSON          = ... os.environ.get("TEST_JETSON",         0          ) ...
```
these two are doing almost the exact same thing, so the style in the naming of the variables should be consistent.

You definitely brought up a good point - constants. I wouldn't call those variables constants because they differ every time the script is loaded based on the set environment variables, where constants should be constant. Ideally, I would put those variables inside the class below, for example:
```python
class PythonTest(unittest.TestCase):

  _tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR", "localhost")
  ...

  def setUp(self):
    ...
```
so they become protected variables of the class. If possible, this is even better:
```python
class PythonTest(unittest.TestCase):

  def setUp(self):
    self._tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR", "localhost")
    self._test_jetson = bool(int(os.environ.get("TEST_JETSON", 0)))
    self._test_windows = bool(int(os.environ.get("TEST_WINDOWS", 0)))
    self._shm_leak_detector = shm_util.ShmLeakDetector()
```
https://www.geeksforgeeks.org/protected-variable-in-python/

qa/L0_backend_python/test.sh
```diff
@@ -387,23 +407,31 @@ fi
...
     for TEST in ${SUBTESTS}; do
         # Run each subtest in a separate virtual environment to avoid conflicts
         # between dependencies.
         setup_virtualenv
 
         (cd ${TEST} && bash -ex test.sh)
-        if [ $? -ne 0 ]; then
-        echo "Subtest ${TEST} FAILED"
-        RET=1
+            if [ $? -ne 0 ]; then
```
krishung5, line 425: nit: I think there's an extra indent before the if statement?

qa/L0_backend_python/restart/test.sh
```diff
@@ -26,16 +26,17 @@
...
-SERVER_ARGS="--model-repository=`pwd`/models --backend-directory=${BACKEND_DIR} --log-verbose=1"
+SERVER_ARGS="--model-repository=${MODELDIR}/restart/models --backend-directory=${BACKEND_DIR} --log-verbose=1"
 SERVER_LOG="./restart_server.log"
 source ../../common/util.sh
 source ../common.sh
 
 rm -fr *.log free_memory.txt
 
+if [[ ${TEST_WINDOWS} == 1 ]]; then
+    pip install pytest tritonclient[all]
```
krishung5, line 37: Maybe we can consider moving this part to utils.sh: https://github.com/triton-inference-server/server/blob/main/qa/common/util.sh#L513-L518 to remove the duplicated code in each test.

qa/L0_backend_python/decoupled/models/decoupled_bls/1/model.py
```diff
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
krishung5, line 25: nit: update copyright for this and some other files.

qa/L0_backend_python/test.sh
fpetrini15: I'll create a follow-up ticket for this since I believe we have plans to completely overhaul the windows testing process on the gitlab side. This overhaul is the only reason I'm hesitant to change it now, given the CI stability and the proximity to the deadline. Introducing it would also require changes to other test scripts that leave the scope of this ticket.

qa/L0_backend_python/python_test.py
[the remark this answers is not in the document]
fpetrini15: This makes a lot of sense Jacky, thank you for clarifying. Doing a quick search on our repo, it seems like our use of these variables isn't too extensive, so I'll add this task to the folllow-up ticket mentioned below.

qa/L0_backend_python/restart/test.sh
fpetrini15: This is a great suggestion! Originally, I figured the dependencies might change from test to test. Turns out not really..

qa/L0_backend_python/python_test.py
```diff
@@ -122,50 +127,56 @@ def _optional_input_infer(self, model_name, has_input0, has_input1):
...
+            # Send a small paylaod to make sure it is still working properly
+            total_byte_size = 2 * 1024 * 1024
+            shape = [total_byte_size]
+            with self._shm_leak_detector.Probe() as shm_probe:
+                self._infer_help(model_name, shape, dtype)
 
     # GPU tensors are not supported on jetson
     # CUDA Shared memory is not supported on jetson
-    if not TEST_JETSON:
+    if not _test_jetson and not _test_windows:
```
kthui, line 170: This could be in a follow up ticket. We should use a [decorator](https://docs.python.org/3.10/library/unittest.html#unittest.skipIf) for skipping tests on different platforms.

kthui (approved):
LGTM. Make sure to address @krishung5 comments before merging.

qa/L0_backend_python/python_test.py
fpetrini15: Created DLIS-6129 to track the effort to address this comment and your previous ones.

krishung5 (approved):
Great work! LGTM.

PR #6833, opened 2024-01-25
Handle 0 dimension output for generate endpoint

krishung5:
For the TRT-LLM backend, it is possible that the output tensor has the 0 dim shape if the end token is predicted at the first step. This PR fixed the handle for 0 dim output, and added test case for it.

Before the fix, when using generate/generate_stream endpoint for zero dim tensors:
**generate**
```
root@a2826b5-lcedt:/opt/tritonserver/tensorrtllm_backend# curl -X POST localhost:8000/v2/models/ensemble/generate --data-binary @sample.txt
{"error":"attempt to access non-existing array index '0'"}
```
**generate_stream**
```
root@a2826b5-lcedt:/opt/tritonserver/tensorrtllm_backend# curl -X POST localhost:8000/v2/models/ensemble/generate_stream --data-binary @sample
.txt
data: {"error":"attempt to access non-existing array index '0'"}
```

After the fix:
**generate**
```
root@a2826b5-lcedt:/opt/tritonserver/tensorrtllm_backend# curl -X POST localhost:8000/v2/models/ensemble/generate --data-binary @sample.txt
{"context_logits":0.0,"cum_log_probs":0.0,"generation_logits":0.0,"model_name":"ensemble","model_version":"1","output_log_probs":[],"sequence_end":false,"sequence_id":0,"sequence_start":false,"text_output":[]}
```
**generate_stream**
```
root@a2826b5-lcedt:/opt/tritonserver/tensorrtllm_backend# curl -X POST localhost:8000/v2/models/ensemble/generate_stream --data-binary @sample
.txt
data: {"context_logits":0.0,"cum_log_probs":0.0,"generation_logits":0.0,"model_name":"ensemble","model_version":"1","output_log_probs":[],"sequence_end":false,"sequence_id":0,"sequence_start":false,"text_output":[]}
```

sample.txt looks like this:
```
{
  "text_input": "I have to work on an AI project. So now, I will explain the project I have to do first :The data is divided into three parts. \n\n1. Training data: train.csv \n2. Test data: test.csv \n3. Submission file: sample\\_submission.csv \n\nThe train.csv file consists of 3 columns of id, text and label, and the test.csv file consists of 2 columns of id and text. The sample\\_submission.csv file consists of id and label columns second: There are 8 labels in total. The data is news article. I want to make a model to classify this. First of all, I want to know the number of classes in the train.csv file. I mean, I want to make sure the class is in balance. I'm talking about EDA. Can you make this code first? and at this code please add about Visualization and also printing each class count.",
  "max_tokens": 1024,
  "bad_words": "",
  "stop_words": "",
  "end_id": 40
}
```

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -4489,13 +4489,13 @@ HTTPAPIServer::GenerateRequestClass::ExactMappingOutput(
           *generate_response, triton::common::TritonJson::ValueType::ARRAY);
       RETURN_IF_ERR(WriteDataToJson(
           &data_json, cname, datatype, base, byte_size, element_count));
-      if (element_count > 1) {
-        RETURN_IF_ERR(generate_response->Add(cname, std::move(data_json)));
-      } else {
+      if (element_count == 1) {
         // if only 1 element, strip out the array
         triton::common::TritonJson::Value el;
         RETURN_IF_ERR(data_json.At(0, &el));
         RETURN_IF_ERR(generate_response->Add(cname, std::move(el)));
+      } else {
+        RETURN_IF_ERR(generate_response->Add(cname, std::move(data_json)));
       }
       break;
     }
--- a/qa/L0_http/generate_endpoint_test.py
+++ b/qa/L0_http/generate_endpoint_test.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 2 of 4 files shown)


src/http_server.cc
```diff
@@ -4489,13 +4489,13 @@ HTTPAPIServer::GenerateRequestClass::ExactMappingOutput(
...
           &data_json, cname, datatype, base, byte_size, element_count));
-      if (element_count > 1) {
-        RETURN_IF_ERR(generate_response->Add(cname, std::move(data_json)));
-      } else {
+      if (element_count == 1) {
         // if only 1 element, strip out the array
         triton::common::TritonJson::Value el;
         RETURN_IF_ERR(data_json.At(0, &el));
         RETURN_IF_ERR(generate_response->Add(cname, std::move(el)));
+      } else {
```
rmccorm4, line 4497: Was the wrong case here `element_count == 0` ?

krishung5: Yes. The error `attempt to access non-existing array index '0'` occurred before this fix was because `data_json` was an empty array and this line  `RETURN_IF_ERR(data_json.At(0, &el));` was called for `element_count == 0`.

rmccorm4: Can you share the full output from generate endpoint for one of these zero dim tensors after the fix?

rmccorm4: Both `/generate` and `/generate_stream` *

krishung5: Sure! Added the output in the PR description.

rmccorm4 (approved):

src/http_server.cc
rmccorm4: Thanks!

PR #6834, opened 2024-01-25
Use libmamba solver for L0_backend_python env test. Fix pytest not found

krishung5:
Update the miniconda version to use libmamba solver to speed up the test time. From the output we can see that `conda-libmamba-solver` is used as default solver.
```
(base) root@a2826b5-lcedt:/opt/tritonserver/qa/L0_backend_python/env# conda info

     active environment : base
    active env location : /opt/tritonserver/qa/L0_backend_python/env/miniconda
            shell level : 1
       user config file : /root/.condarc
 populated config files : 
          conda version : 23.11.0
    conda-build version : not installed
         python version : 3.10.13.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=skylake_avx512
                          __conda=23.11.0=0
                          __cuda=12.2=0
                          __glibc=2.35=0
                          __linux=6.5.0=0
                          __unix=0=0
       base environment : /opt/tritonserver/qa/L0_backend_python/env/miniconda  (writable)
      conda av data dir : /opt/tritonserver/qa/L0_backend_python/env/miniconda/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /opt/tritonserver/qa/L0_backend_python/env/miniconda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/tritonserver/qa/L0_backend_python/env/miniconda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/23.11.0 requests/2.31.0 CPython/3.10.13 Linux/6.5.0-14-generic ubuntu/22.04.3 glibc/2.35 solver/libmamba conda-libmamba-solver/23.12.0 libmambapy/1.5.3
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

Also fixed the `pytest not found` error in the tests that are using different versions of Python. It's likely due to the Python version used to install `virtualenv`(within Dockerfile.QA) mismatches with the Python version used for the test.

### DIFF
--- a/qa/L0_backend_python/test.sh
+++ b/qa/L0_backend_python/test.sh
@@ -390,8 +390,7 @@ if [ "$TEST_JETSON" == "0" ]; then
     for TEST in ${SUBTESTS}; do
         # Run each subtest in a separate virtual environment to avoid conflicts
         # between dependencies.
-        virtualenv --system-site-packages venv
-        source venv/bin/activate
+        setup_virtualenv
 
         (cd ${TEST} && bash -ex test.sh)
         if [ $? -ne 0 ]; then
@@ -399,8 +398,7 @@ if [ "$TEST_JETSON" == "0" ]; then
         RET=1
         fi
 
-        deactivate
-        rm -fr venv
+        deactivate_virtualenv
     done
 
     if [ ${PYTHON_ENV_VERSION} = "10" ]; then
@@ -418,8 +416,7 @@ SUBTESTS="lifecycle restart model_control examples argument_validation logging c
 for TEST in ${SUBTESTS}; do
     # Run each subtest in a separate virtual environment to avoid conflicts
     # between dependencies.
-    virtualenv --system-site-packages venv
-    source venv/bin/activate
+    setup_virtualenv
 
     (cd ${TEST} && bash -ex test.sh)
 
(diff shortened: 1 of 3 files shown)


rmccorm4 (approved):
Nice test time speedups! Might want a second opinion from @Tabrizian just in case

qa/L0_backend_python/test.sh
```diff
@@ -153,6 +153,8 @@ else
   # GPU tensor tests are disabled on jetson
 fi
 
+pip3 install pytest
```
Tabrizian, line 156: Now that we are using pytest in all the tests it might be worth installing it in the QA container.

Tabrizian (approved):
Thanks Kris!!

qa/L0_backend_python/test.sh
krishung5: I think Pytest is already there with our default Python version. Installing Pytest in the test is just for the different python environments.

PR #6837, opened 2024-01-26
Add test for shutdown while loading model

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/324

Add a new test that starts loading a model, and before the model can finish loading, the server shutdowns. Given a model is still loading while shutting down, the unload must happen after the load is completed. If the test cannot assert there is exactly one load complete and one unload in order, then the test fails.

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -3347,6 +3347,18 @@ class LifeCycleTest(tu.TestResultCollector):
         # The server will shutdown after this sub-test exits. The server must shutdown
         # without any hang or runtime error.
 
+    def test_shutdown_while_loading(self):
+        triton_client = self._get_client()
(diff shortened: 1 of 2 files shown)


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -3317,6 +3317,7 @@ def test_model_config_overwite(self):
 
         # Touch the local config.pbtxt and reload the file to ensure the local config
         # is preferred because it has a more recent mtime.
+        time.sleep(0.1)  # make sure timestamps are different
```
rmccorm4, line 3320: NOTE: This was an idea by Jacky to try to improve the stability in CI in the event the system clock/internals doesn't actually update quickly enough to reflect a different timestamp when checked in core. Currently the test is failing intermittently around timestamps in this subtest.

rmccorm4 (approved):

Issue #6693, opened 2023-12-12
Interval-Based OpenTelemetry Tracing in Triton Server


teith:
**Is your feature request related to a problem? Please describe.**
I'm exploring Triton Inference Server's integration with OpenTelemetry for tracing. I'm interested in whether Triton can emit OpenTelemetry traces at regular intervals or under specific conditions, instead of for every request. This would be beneficial for efficient performance monitoring without the overhead of continuous detailed tracing.

**Describe the solution you'd like**
I'd like an option in Triton to configure OpenTelemetry trace emission at set intervals or based on certain triggers, rather than per-request.

**Describe alternatives you've considered**
I considered manually retrieving trace files through Nginx at specific times, but this approach seems inefficient and less practical.

**Additional context**
Interval-based or conditional tracing would offer a balanced approach between gathering detailed performance data and managing system overhead, optimizing the monitoring and debugging process.

denti:
+1

oandreeva-nv:
Hi @teith , you can specify rate at which you want to collect traces: https://github.com/triton-inference-server/server/blob/main/docs/user_guide/trace.md#global-settings

For example,  `tritonserver --trace-config mode=opentelemetry --trace-config rate=100 ` will only trace every 100th inference request.

teith:
Hi, @oandreeva-nv!
Thanks for your response. I was actually looking for a way to send groups of 100 traces in one file to a webhook, not just every 100th trace. This would help manage the tracing data better and reduce system load.

oandreeva-nv:
I see, thanks for clarification. [I've created a ticket for this feature: 5943]

oandreeva-nv:
I'll close this issue, since this feature is currently targeted to 24.02 release and currently on master. Relevant PR: https://github.com/triton-inference-server/server/pull/6842

oandreeva-nv:
Relevant docs: https://github.com/triton-inference-server/server/blob/main/docs/user_guide/trace.md#opentelemetry-trace-support

PR #6733, opened 2023-12-22
export opentelemetry spans with batch processor

theoclark:
We have observed slowdowns of up to 50% when running tritonserver with opentelemetry. We have traced this to the SimpleSpan Exporter and have observed these slowdowns go away when switching to the BatchSpanProcessor.

PR #6842, opened 2024-01-26
Adding OpenTelemetry Batch Span Processor

oandreeva-nv:
Fixes issue: https://github.com/triton-inference-server/server/issues/6693
Picked up work from [this PR](https://github.com/triton-inference-server/server/pull/6733) by @theoclark , he has CLA : https://github.com/triton-inference-server/server/pull/6733#issuecomment-1874349122

Currently, OpenTelemetry Trace API uses SimpleSpanProcessor, which send spans one by one, when they are ready. This impacts Triton's performance, when tracing is on and mode is `opentelemetry`. To mitigate this issue, SimpleSpanProcessor was replaced by a BatchSpanProcessor(https://opentelemetry.io/docs/specs/otel/trace/sdk/#batching-processor), which collects spans in batches of a specified size and sends them out to the endpoint periodically. 

There are 3 customization parameters:
`bsp_max_queue_size` - the maximum queue size. After the size is reached spans are dropped. The default value is 2048.
`bsp_schedule_delay` - the maximum delay interval in milliseconds between two consecutive exports. The default value is 5000.
`bsp_max_export_batch_size` - the maximum batch size of every export. It must be smaller or equal to `bsp_max_queue_size`. If the queue reaches `bsp_max_export_batch_size`  a batch will be exported even if `bsp_schedule_delay` milliseconds have not elapsed. The default value is 512.

Note, that specification has 4 parameters, but [opentelemetry-cpp](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/sdk/include/opentelemetry/sdk/trace/batch_span_processor_options.h) client only supports 3 at the moment.

In this PR:
- I add an ability to specify `bsp_max_queue_size`, `bsp_schedule_delay` and `bsp_max_export_batch_size` through cmd line arguments. Alternatively, these parameters could also be specified from the environment variables.
- I verify these arguments in `command_line_parser` and throw exception if argument is specified incorrectly.
- I've also adjusted `TraceConfig` to store pairs, where second element can be one of `std::string`,`int`, `uint32_t`. Note, that cmdline parser will always collect strings of arguments' values. During the validation process I convert numerical values to integers and store integers instead of strings. This is done for multiple reasons: 1) We need to validate args 2) Since validation makes sure that `StringToInt` runs successfully, we can store converted value in `trace_config_map` and pass to TraceManager with valid parameters and proper types.
- I've added docs
- I've added tests

### DIFF
--- a/docs/user_guide/trace.md
+++ b/docs/user_guide/trace.md
@@ -487,7 +487,15 @@ Default parameters for the Batch Span Processor are provided in
 As a general recommendation, make sure that `bsp_max_queue_size` is large enough
 to hold all collected spans, and `bsp_schedule_delay` does not cause frequent
 exports, which will affect Triton Server's latency. Also, users should keep
-in mind that minimal number of produced spans on Trton Server side is 3:
+in mind that at minimum Triton collects 3 levels of spans:
+
+* Top level. The top-level span collects timestamps for when request was received
+by Triton, and when the response was sent.
+* Mid level
+* Low level
+
+
+minimal number of produced spans on Trton Server side is 3:
 The `InferRequest` span, which collects timestamps for when request was received
 by Triton, and when the response was sent; the request span, which is named
 after the model the request was sent to and records when request was started and
--- a/docs/user_guide/trace.md
+++ b/docs/user_guide/trace.md
@@ -486,20 +486,26 @@ Default parameters for the Batch Span Processor are provided in
 [`OpenTelemetry trace APIs settings`](#opentelemetry-trace-apis-settings).
 As a general recommendation, make sure that `bsp_max_queue_size` is large enough
 to hold all collected spans, and `bsp_schedule_delay` does not cause frequent
-exports, which will affect Triton Server's latency. Also, users should keep
-in mind that at minimum Triton collects 3 levels of spans:
+exports, which will affect Triton Server's latency. A minimal Triton trace
+consists of 3 spans: top level span, model span, and compute span.
 
-* Top level. The top-level span collects timestamps for when request was received
-by Triton, and when the response was sent.
-* Mid level
-* Low level
+* __Top level span__. The top-level span collects timestamps for when
+request was received by Triton, and when the response was sent. Any Triton
+trace contains only 1 top level span.
+* __Model span__. Model spans collect information, when request for
+this model was started, when it was placed in a queue, and when it was ended.
+A minimal Triton trace contains 1 model span.
+* __Compute span__. Compute spans record compute timestamps. A minimal
+Triton trace contains 1 compute span.
 
+The total amount of spans depends on the complexity of your model.
+A general rule is any base model - a single model that performs computations -
+produces 1 model span and one compute span. For ensembles, every composing
+model produces model and compute spans in addition to one model span for the
+ensemble. [BLS](#tracing-for-bls-models) models produce the same number of
+model an compute spans as the total amount of involved in BLS request models,
+including the main BLS model.
 
-minimal number of produced spans on Trton Server side is 3:
-The `InferRequest` span, which collects timestamps for when request was received
-by Triton, and when the response was sent; the request span, which is named
-after the model the request was sent to and records when request was started and
-ended; the `compute` span, which records compute timestamps.
 
 ### Differences in trace contents from Triton's trace [output](#json-trace-output)
 
--- a/docs/user_guide/trace.md
+++ b/docs/user_guide/trace.md
@@ -489,7 +489,7 @@ to hold all collected spans, and `bsp_schedule_delay` does not cause frequent
 exports, which will affect Triton Server's latency. Also, users should keep
 in mind that minimal number of produced spans on Trton Server side is 3:
 The `InferRequest` span, which collects timestamps for when request was received
-by Triton, and when the request was sent; the request span, which is named
+by Triton, and when the response was sent; the request span, which is named
 after the model the request was sent to and records when request was started and
 ended; the `compute` span, which records compute timestamps.
 
--- a/docs/user_guide/trace.md
+++ b/docs/user_guide/trace.md
@@ -1,5 +1,5 @@
 <!--
-# Copyright 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2019-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -467,6 +467,32 @@ flag as follows:
 $ tritonserver --trace-config mode=opentelemetry \
     --trace-config opentelemetry,url=<endpoint> ...
 ```
+
+Triton's OpenTelemetry trace mode uses
+[Batch Span Processor](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#batch-span-processor),
+which batches ended spans and sends them in bulk. Batching helps
+with data compression and reduces the number of outgoing connections
+required to transmit the data. This processor supports both size and
+time based batching. Size-based batching is controlled by 2 parameters:
+`bsp_max_export_batch_size` and `bsp_max_queue_size`, while time-based batching
+is controlled by `bsp_schedule_delay`. Collected spans will be exported when
+the batch size reaches `bsp_max_export_batch_size`, or delay since last export
+reaches `bsp_schedule_delay`, whatever comes first. Additionally, user should
+make sure that `bsp_max_export_batch_size` is always less than
+`bsp_max_queue_size`, otherwise the excessive spans will be dropped
+and trace data will be lost.
+
+Default parameters for the Batch Span Processor are provided in
+[`OpenTelemetry trace APIs settings`](#opentelemetry-trace-apis-settings).
+As a general recommendation, make sure that `bsp_max_queue_size` is large enough
+to hold all collected spans, and `bsp_schedule_delay` does not cause frequent
+exports, which will affect Triton Server's latency. Also, users should keep
+in mind that minimal number of produced spans on Trton Server side is 3:
+The `InferRequest` span, which collects timestamps for when request was received
+by Triton, and when the request was sent; the request span, which is named
+after the model the request was sent to and records when request was started and
+ended; the `compute` span, which records compute timestamps.
+
 ### Differences in trace contents from Triton's trace [output](#json-trace-output)
 
 OpenTelemetry APIs produce [spans](https://opentelemetry.io/docs/concepts/observability-primer/#spans)
@@ -552,7 +578,8 @@ The following table shows available OpenTelemetry trace APIs settings for
     <td><code>bsp_max_export_batch_size</code></td>
     <td align="center">512</td>
     <td>
-      Maximum batch size. Must be less than or equal to `bsp_max_queue_size`.<br/>
+      Maximum batch size. Must be less than or equal to
+      <code>bsp_max_queue_size</code>.<br/>
       This setting can also be specified through <br/>
       <a href="https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#batch-span-processor">
       OTEL_BSP_MAX_EXPORT_BATCH_SIZE</a>
--- a/docs/user_guide/trace.md
+++ b/docs/user_guide/trace.md
@@ -520,6 +520,45 @@ The following table shows available OpenTelemetry trace APIs settings for
       environment variable.
     </td>
     </tr>
+    <tr>
+    <td><a href="https://opentelemetry.io/docs/specs/otel/trace/sdk/#batching-processor">
+      Batch Span Processor</a>
+    </td>
+    <td></td><td></td>
+    </tr>
+    <tr>
+    <td><code>bsp_max_queue_size</code></td>
+    <td align="center">2048</td>
+    <td>
+      Maximum queue size. <br/>
+      This setting can also be specified through <br/>
+      <a href="https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#batch-span-processor">
+      OTEL_BSP_MAX_QUEUE_SIZE</a>
+      environment variable.
+    </td>
+    </tr>
+    <tr>
+    <td><code>bsp_schedule_delay</code></td>
+    <td align="center">5000</td>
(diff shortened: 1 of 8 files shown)


rmccorm4:
> Currently, OpenTelemetry Trace API uses SimpleSpanProcessor, which send spans one by one, when they are ready. This impacts Triton's performance, when tracing is on and mode is opentelemetry.

If it's not too much work to check, is there a visible impact in PA throughput/latency for an example model with tracing ON vs OFF ?

oandreeva-nv:
@rmccorm4 I remember a user reported around x2 slowdown. If I don't find it, I'll run some experiments myself

oandreeva-nv:
From https://github.com/triton-inference-server/server/pull/6733#issue-2053941958:
> We have observed slowdowns of up to 50% when running tritonserver with opentelemetry. We have traced this to the SimpleSpan Exporter and have observed these slowdowns go away when switching to the BatchSpanProcessor.

docs/user_guide/trace.md
```diff
@@ -520,6 +520,45 @@ The following table shows available OpenTelemetry trace APIs settings for
...
+      <a href="https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#batch-span-processor">
+      OTEL_BSP_SCHEDULE_DELAY</a>
+      environment variable.
+    </td>
+    </tr>
+    <tr>
+    <td><code>bsp_max_export_batch_size</code></td>
+    <td align="center">512</td>
+    <td>
+      Maximum batch size. Must be less than or equal to `bsp_max_queue_size`.<br/>
```
rmccorm4, line 555: I think this needs to be changed to render correctly since not using markdown directly
```suggestion
      Maximum batch size. Must be less than or equal to <code>bsp_max_queue_size</code>.<br/>
```

qa/L0_cmdline_trace/test.sh
```diff
@@ -618,11 +618,140 @@ set -e
...
+run_server
+if [ "$SERVER_PID" != "0" ]; then
+    echo -e "\n***\n***Fail: Server start should have failed $SERVER\n***"
+    cat $SERVER_LOG
+    set -e
+    kill $SERVER_PID
+    wait $SERVER_PID
+    set +e
+    exit 1
+fi
```
rmccorm4, line 641: Can this be grouped into some helper to condense this?
```bash
function assert_server_startup_failed() {
  if [ "$SERVER_PID" != "0" ]; then
      echo -e "\n***\n***Fail: Server start should have failed $SERVER\n***"
      cat $SERVER_LOG
      set -e
      kill $SERVER_PID
      wait $SERVER_PID
      set +e
      exit 1
  fi
}

...
run_server
assert_server_startup_failed

...
run_server
assert_server_startup_failed
```

qa/L0_cmdline_trace/test.sh
```diff
@@ -618,11 +618,140 @@ set -e
 kill $SERVER_PID
 wait $SERVER_PID
 
+set +e
```
rmccorm4, line 621: Random nit: but I also think it would help organization if `L0_cmdline_trace` was renamed to `L0_trace_cmdline` - if you feel like tweaking that at some point, doesn't have to be in this PR.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -377,16 +377,21 @@ def _test_trace(
         """
         time.sleep(COLLECTOR_TIMEOUT)
         traces = self._parse_trace_log(self.filename)
-        self.assertEqual(len(traces), 1, "Unexpected number of traces collected")
+        expected_traces_number = 1
```
rmccorm4, line 380: Is a "batch" of say 512 spans considered 1 trace or 512 traces?

rmccorm4: and are we testing the batch size anywhere in here?

qa/L0_trace/test.sh
```diff
@@ -897,4 +897,132 @@ kill $SERVER_PID
...
+    cat $SERVER_LOG
+    exit 1
+fi
+
+rm collected_traces.json
+$OTEL_COLLECTOR --config ./trace-config.yaml >> $OTEL_COLLECTOR_LOG 2>&1 & COLLECTOR_PID=$!
+
+set +e
+python -c 'import opentelemetry_unittest; \
+    opentelemetry_unittest.send_bls_request(model_name="ensemble_add_sub_int32_int32_int32")'  >> client_update.log 2>&1
```
rmccorm4, line 929: should the model be "bls_simple" for bls requests?

rmccorm4: same in 2 other places below

qa/L0_trace/test.sh
```diff
@@ -897,4 +897,132 @@ kill $SERVER_PID
 wait $SERVER_PID
 set +e
 
+################################################################################
```
rmccorm4, line 900: I think you need to update copyrights to 2024 throughout some of these files as well

src/command_line_parser.cc
```diff
@@ -2137,37 +2144,38 @@ TritonParser::SetGlobalTraceArgs(
     bool explicit_disable_trace)
 {
   for (const auto& global_setting : lparams.trace_config_map_[""]) {
+    auto user_setting = std::get<std::string>(global_setting.second);
```
rmccorm4, line 2147: optional: With c++17, you could use structured bindings if you wanted. Something like:
```cpp
for (const auto& [setting, value] : lparams.trace_config_map_[""])
  if (setting == "rate") {
    lparams.trace_rate_ = ParseOption<int>(value);
  }
  ...
```

src/tracer.cc
```diff
@@ -409,29 +405,16 @@ TraceManager::InitTracer(const triton::server::TraceConfigMap& config_map)
...
-            attributes[key] = value;
-          }
-        }
-      }
-      auto exporter = otlp::OtlpHttpExporterFactory::Create(opts);
-      auto processor = otel_trace_sdk::SimpleSpanProcessorFactory::Create(
-          std::move(exporter));
+      otel_trace_sdk::BatchSpanProcessorOptions processor_options;
+
+      ProcessOpenTelemetryParameters(
```
rmccorm4, line 412: nice refactor :tada:

src/tracer.cc
```diff
@@ -477,6 +460,45 @@ TraceManager::CleanupTracer()
...
+    const triton::server::TraceConfigMap& config_map,
+    otlp::OtlpHttpExporterOptions& exporter_options,
+    otel_resource::ResourceAttributes& attributes,
+    otel_trace_sdk::BatchSpanProcessorOptions& processor_options)
+{
+  attributes[otel_resource::SemanticConventions::kServiceName] =
+      std::string("triton-inference-server");
+  auto mode_key = std::to_string(TRACE_MODE_OPENTELEMETRY);
+  auto otel_options_it = config_map.find(mode_key);
+  if (otel_options_it != config_map.end()) {
```
rmccorm4, line 474: nit: It can help reduce nesting to check the inverse and exit early:
```cpp
if (otel_options_it == config_map.end()) {
  return;
}

for (...) {
}
```

src/tracer.cc
```diff
@@ -477,6 +460,45 @@ TraceManager::CleanupTracer()
...
+      }
+      if (setting.first == "resource") {
+        auto user_setting = std::get<std::string>(setting.second);
+        auto pos = user_setting.find('=');
+        auto key = user_setting.substr(0, pos);
+        auto value = user_setting.substr(pos + 1);
+        attributes[key] = value;
+      }
+      if (setting.first == "bsp_max_queue_size") {
+        processor_options.max_queue_size = std::get<uint32_t>(setting.second);
```
rmccorm4, line 488: nice use of `std::variant` :brain:

rmccorm4 (commented):
LGTM overall! Most comments are just optional nits aside from a few questions in the tests.

qa/L0_trace/test.sh
oandreeva-nv: the `model_name` here is the name of a child model, i.e. to which `bls_simple` sends requests. `bls_simple` is hard coded here: https://github.com/triton-inference-server/server/blob/dfe9dde17c28535c357ceb304cf4286f018dcbf5/qa/L0_trace/opentelemetry_unittest.py#L81

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: > Is a "batch" of say 512 spans considered 1 trace or 512 traces?
"batch" can collect spans from multiple traces. Let's assume that we trace one model that can produce >= 512 spans in one request. In this case if max_batch_size is 512 and delay is long enough so that we can collect 512 traces, then when exporter exports them, they all belong to 1 trace.

In unittests, I set up the collector so that it will write trace in the file, when all spans are received, that is why I only expect 1 trace. But this trace contains multiple spans.

> and are we testing the batch size > 1 anywhere in here?
All unittests are run with default parameters for batch span processor: https://github.com/triton-inference-server/server/blob/acb5f445350b6927a7bac7b0d99b0241ccfeede7/qa/L0_trace/test.sh#L726-L733 
i.e. I don't specify them. Thus the batch size is 512 and delay 5000, so spans are exported every 5 seconds

src/command_line_parser.cc
oandreeva-nv: right, I forgot about this luxury in C++17

qa/L0_trace/test.sh
rmccorm4: Oops I see, didn't look close enough. Thanks!

qa/L0_trace/opentelemetry_unittest.py
rmccorm4: Makes sense, thanks 🙇‍♂️

docs/user_guide/trace.md
```diff
@@ -467,6 +467,32 @@ flag as follows:
...
+Default parameters for the Batch Span Processor are provided in
+[`OpenTelemetry trace APIs settings`](#opentelemetry-trace-apis-settings).
+As a general recommendation, make sure that `bsp_max_queue_size` is large enough
+to hold all collected spans, and `bsp_schedule_delay` does not cause frequent
+exports, which will affect Triton Server's latency. Also, users should keep
+in mind that minimal number of produced spans on Trton Server side is 3:
+The `InferRequest` span, which collects timestamps for when request was received
+by Triton, and when the request was sent; the request span, which is named
+after the model the request was sent to and records when request was started and
+ended; the `compute` span, which records compute timestamps.
```
rmccorm4, line 494: This section/description sounds a little unclear. Currently it says the 3 spans are:
1. `InferRequest`
2. request
3. `compute`

"InferRequest" and "request" sound a bit ambiguous/confusing to me.

(short term)

Without changing any current logic/names in the code, maybe calling the middle span the "Scheduler" or "Model" span would be more clear?:
1. `InferRequest`
2. Scheduler/Model
3. `Compute`

(longer term)

Or if we were actually changing some names in the codes/tests, then maybe something like this would eventually be clearest:
1. Network (HTTP/GRPC RECV/SEND)
2. InferRequest or Scheduler (REQUEST START/END)
3. Compute (COMPUTE START/END)

What do you think?

docs/user_guide/trace.md
```diff
@@ -467,6 +467,32 @@ flag as follows:
...
+and trace data will be lost.
+
+Default parameters for the Batch Span Processor are provided in
+[`OpenTelemetry trace APIs settings`](#opentelemetry-trace-apis-settings).
+As a general recommendation, make sure that `bsp_max_queue_size` is large enough
+to hold all collected spans, and `bsp_schedule_delay` does not cause frequent
+exports, which will affect Triton Server's latency. Also, users should keep
+in mind that minimal number of produced spans on Trton Server side is 3:
+The `InferRequest` span, which collects timestamps for when request was received
+by Triton, and when the request was sent; the request span, which is named
```
rmccorm4, line 492:
```suggestion
by Triton, and when the response was sent; the request span, which is named
```

rmccorm4 (commented):
Code and tests LGTM - last comments are just on docs: https://github.com/triton-inference-server/server/pull/6842/files#r1472390626

docs/user_guide/trace.md
oandreeva-nv: I would like to avoid changing span names at least in this PR. I agree with the idea of calling the middle span as something related to Scheduler/Model. Let me think for a name

oandreeva-nv: I did my best in trying to clarify each type of spans

docs/user_guide/trace.md
```diff
@@ -467,6 +467,46 @@ flag as follows:
...
+A minimal Triton trace contains 1 model span.
+* __Compute span__. Compute spans record compute timestamps. A minimal
+Triton trace contains 1 compute span.
+
+The total amount of spans depends on the complexity of your model.
+A general rule is any base model - a single model that performs computations -
+produces 1 model span and one compute span. For ensembles, every composing
+model produces model and compute spans in addition to one model span for the
+ensemble. [BLS](#tracing-for-bls-models) models produce the same number of
+model an compute spans as the total amount of involved in BLS request models,
```
rmccorm4, line 506:
```suggestion
model and compute spans as the total amount of models involved in the BLS request,
```

docs/user_guide/trace.md
```diff
@@ -467,6 +467,46 @@ flag as follows:
...
+consists of 3 spans: top level span, model span, and compute span.
+
+* __Top level span__. The top-level span collects timestamps for when
+request was received by Triton, and when the response was sent. Any Triton
+trace contains only 1 top level span.
+* __Model span__. Model spans collect information, when request for
+this model was started, when it was placed in a queue, and when it was ended.
+A minimal Triton trace contains 1 model span.
+* __Compute span__. Compute spans record compute timestamps. A minimal
+Triton trace contains 1 compute span.
```
rmccorm4, line 499:
```suggestion
* __Top level span__: The top-level span collects timestamps for when
request was received by Triton, and when the response was sent. Any Triton
trace contains only 1 top level span.
* __Model span__: Model spans collect information, when request for
this model was started, when it was placed in a queue, and when it was ended.
A minimal Triton trace contains 1 model span.
* __Compute span__: Compute spans record compute timestamps. A minimal
Triton trace contains 1 compute span.
```

rmccorm4 (approved):
Nice changes!

PR #6873, opened 2024-02-12
Add unit test reports to L0_dlpack_multi_gpu and L0_warmup

krishung5:
Adding unit test reports to L0_dlpack_multi_gpu and L0_warmup to also fix the test failures as we removed the use of `check_test_results` for `python_unittest.py` and it was used in these tests as well.


rmccorm4 (dismissed):
Nice catch :rocket:

krishung5:
@rmccorm4 Just found out L0_warmup is using the python_unittest.py as well, need your review on that file. I think these tests should be all the ones that are using the unit test script. Thanks!

rmccorm4 (approved):

PR #6890, opened 2024-02-16
Nuke onnx models from test 

oandreeva-nv:
I disabled some onnx tests, but mostly replaced `onnx` models with `libtorch` models.
The only question left with L0_mlflow. I don't know as of now if `onnx` models are required for this test or not. 
If yes, we should probably disable this test.


mc-nv (approved):

qa/L0_infer_zero/test.sh
```diff
@@ -54,6 +54,7 @@ rm -f $SERVER_LOG $CLIENT_LOG
 rm -fr models && mkdir models
 cp -r /data/inferenceserver/${REPO_VERSION}/qa_identity_model_repository/* models/. && \
     cp -r /data/inferenceserver/${REPO_VERSION}/qa_ensemble_model_repository/qa_identity_model_repository/* models/.
+rm `find ./models/ -name '*onnx*'` -rf
```
mc-nv, line 57: In terms of deletion I would recommend use embedded syntax for `find` utility: 
```
find ./models/ -name '*onnx*' -delete
```

oandreeva-nv: The above gave me:
```
+ find ./models0/ -name '*onnx*' -delete
find: cannot delete './models0/onnx_nobatch_sequence_object': Directory not empty
find: cannot delete './models0/onnx_nobatch_sequence_int32': Directory not empty
find: cannot delete './models0/onnx_nobatch_sequence_bool': Directory not empty
```
so I'll keep `rm ... ` for now

oandreeva-nv: I believe not all the time, but sometime. Since this PR will be reverted at some point, I'll stick to `rm ...`, but I'll keep in mind `find ... -delete` for future

mc-nv (approved):

PR #6893, opened 2024-02-16
Remove ignore files that are not in use by repository

mc-nv:
By defininition these files are designed for a specific purpose: 
* [.gitignore](https://git-scm.com/docs/gitignore) - Specifies intentionally untracked files to ignore to prevent undesired files to be pushed to repository with `git push`. 
* [.helmignore](https://helm.sh/docs/chart_template_guide/helm_ignore_file/) - The `.helmignore` file is used to specify files you don't want to include in your helm chart to be pushed to repository with `helm push`.

Those files can be useful in development process but it is not applicable to the Triton server repository.


oandreeva-nv:
@mc-nv Could you please briefly clarify why we ca remove them and they will not be needed for deployment?

nv-kmcgill53 (approved):

oandreeva-nv (approved):

PR #6914, opened 2024-02-23
Enhance bound check for shm offset

kthui:
Enhance shared memory bound checking during inference to make sure the offset supplied by the user cannot go beyond the shared memory region when the region was registered.


qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -323,6 +324,29 @@ def test_unregisterall(self):
...
 
+    def test_infer_offset_out_of_bound(self):
+        # Shared memory offset outside output region - Throws error
+        error_msg = []
+        shm_handles = self._configure_sever()
+        if _protocol == "http":
+            # -0x20 in int64_t
+            offset = 18446744073709551584
+        else:
+            # grpc will throw if > int64, instead test for exceed shm region
```
rmccorm4, line 335: int64 or 64?

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -323,6 +324,29 @@ def test_unregisterall(self):
             self.assertTrue(len(status_after.regions) == 0)
         self._cleanup_server(shm_handles)
 
+    def test_infer_offset_out_of_bound(self):
+        # Shared memory offset outside output region - Throws error
+        error_msg = []
+        shm_handles = self._configure_sever()
+        if _protocol == "http":
+            # -0x20 in int64_t
+            offset = 18446744073709551584
```
rmccorm4, line 333: what's the significance of this value? Do we have a known bounds in this test, and can instead just call "max_bounds+1" or "min_bounds-1" or something?

src/shared_memory_manager.cc
```diff
@@ -346,12 +346,26 @@ SharedMemoryManager::GetMemoryInfo(
         std::string("Unable to find shared memory region: '" + name + "'")
             .c_str());
   }
+
+  // populate 'shm_mapped_addr'
+  size_t max_offset = 0;
```
rmccorm4, line 351: Is the logic any easier or more comprehensive if we consider `min_region_addr <= target_addr <= max_region_addr` bounds rather than what looks like the current `target_addr <= max_region_addr`

src/shared_memory_manager.cc
```diff
@@ -346,12 +346,26 @@ SharedMemoryManager::GetMemoryInfo(
...
   if (it->second->kind_ == TRITONSERVER_MEMORY_CPU) {
     *shm_mapped_addr = (void*)((uint8_t*)it->second->mapped_addr_ +
                                it->second->offset_ + offset);
+    max_offset = it->second->offset_;
   } else {
     *shm_mapped_addr = (void*)((uint8_t*)it->second->mapped_addr_ + offset);
```
rmccorm4, line 357: Will `it->second->offset_` be zero for the `it->second->kind_ != TRITONSERVER_MEMORY_CPU` case? If so, can these cases be condensed into a single one to both do:
```cpp
    *shm_mapped_addr = (void*)((uint8_t*)it->second->mapped_addr_ +
                               it->second->offset_ + offset);
```

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -323,6 +324,29 @@ def test_unregisterall(self):
             self.assertTrue(len(status_after.regions) == 0)
         self._cleanup_server(shm_handles)
 
+    def test_infer_offset_out_of_bound(self):
+        # Shared memory offset outside output region - Throws error
```
rmccorm4, line 328: Is it possible to either loop around or go backwards in terms of offset? i.e. can/should we test a target address before the region rather than after it?

qa/L0_shared_memory/shared_memory_test.py
kthui: The gRPC type is int64, so if that is overflowed (if > 2^63 - 1), then gRPC will throw an error, so I try to test for a off-by-1 corner instead on the gRPC connection.

I can change the wording if it is confusing.

qa/L0_shared_memory/shared_memory_test.py
rmccorm4: I think putting this number in some format like `2**64` or `2**64 - 31` would be more clear in what you're trying to express here

src/shared_memory_manager.cc
kthui: I suppose `size_t` is unsigned, so any offset cannot be `< 0`. Then the `min_region_addr = target_addr`? Given `original_target_addr + (min_offset = 0) = original_target_addr = target_addr`, I think there is nothing to check for `min_region_addr` unless the `original_target_addr` is invalid.

src/shared_memory_manager.cc
kthui: The `it->second->offset_` is ignored when summing the `shm_mapped_addr` for `it->second->kind_ != TRITONSERVER_MEMORY_CPU` case. I wouldn't assume `it->second->offset_` to be zero for that case, because if it is always zero, then the both cases should have been condensed from the beginning.

@GuanLuo do you recall why the CPU shared memory address is calculated as
```
original address + offset when memory is registered + additional offset when inferencing
```
while the GPU shared memory address is 
```
original address                 +                    additional offset when inferencing
```
?

qa/L0_shared_memory/shared_memory_test.py
kthui: > Is it possible to either loop around or go backwards in terms of offset?

Yes. We are dealing with unsigned values at the core, so `18446744073709551584` offset is actually looping around, which ends up 32 bytes before the start of the original shared memory address.

qa/L0_shared_memory/shared_memory_test.py
kthui: Sure. [Improve comments](https://github.com/triton-inference-server/server/pull/6914/commits/2bd4cd8ec333c064b13533f54fb8f666396ae626)

src/shared_memory_manager.cc
GuanLuo: I don't know, perhaps the address pointer has taken offset into account for CPU?

GuanLuo (commented):
Shouldn't you also check the byte size of the read / write as well? Couldn't the same issue happen if one specifies offset = (max_offset - 1) and instructs Triton to write > 1 bytes?

kthui:
> Shouldn't you also check the byte size of the read / write as well? Couldn't the same issue happen if one specifies offset = (max_offset - 1) and instructs Triton to write > 1 bytes?

This scope is exactly the next ticket after this one, do you want to do both of them together?

src/shared_memory_manager.cc
kthui: @rmccorm4 I think lets not change it in this PR? In case something unexpected breaks?

rmccorm4: Sure no problem, was just curious. No need to follow up on it unless you find it useful later.

src/shared_memory_manager.cc
kthui: I think the above is not a clear explanation, let me put it differently:
Since all the variables here are unsigned, the sum of those variables can only be larger except when it overflows, which is why checking for `min_region_addr` is necessary. Assuming the `offset_` and `byte_size_` supplied during shm registration are valid (which should be checked during registration), then the maximum theoretical offset obtained from summing those three values cannot overflow, i.e. `max_offset = offset_ + byte_size_`. Thus, we have two good values to compare `offset > max_offset` and that will be sufficient for the check. I think this is a better check because it does not only ensure the `target_addr` is within range, but also ensures the computation of the `target_addr` did not overflow, i.e. `target_addr = mapped_addr_ + offset_ + offset`.

> what looks like the current `target_addr <= max_region_addr`

The solution is not comparing address, but validating offsets that sum to the address.

nv-kmcgill53:
This is increasing the scope by a little bit but could you provide an improvement to the shared memory manager where we check that the offset provided by the user is a multiple of the host's page size? You could probably test for this if `mmap` [returns an error](https://github.com/triton-inference-server/server/blob/main/src/shared_memory_manager.cc#L156-L162).
```
const size_t page_size = sysconf(_SC_PAGE_SIZE);
if ( (offset % page_size != 0) {
      return TRITONSERVER_ErrorNew(
        TRITONSERVER_ERROR_INTERNAL, std::string(
                                         "Bad argument: offset, " + std::to_string(offest) + 
                                         ", is not a multiple of the host's page size (" + 
                                         std::to_string(page_size) + "): " +
                                         std::string(std::strerror(errno)))
                                         .c_str());
}
```

qa/L0_shared_memory/shared_memory_test.py
rmccorm4: Thanks for improving the comments, for the actual offset itself, I meant changing the number directly to be clear:
```suggestion
            offset = 2**64 - 32
```

kthui: [Improve comment and offset](https://github.com/triton-inference-server/server/pull/6914/commits/bf5a8dc4bc84acb135c49a6bf9a6e73f89075aee)

kthui:
@nv-kmcgill53 if I understand correctly, you are talking about an improved bound check when ***registering*** a shm region, not when ***inferencing*** with a shm region? The next next ticket after this one is about improving checks when registering a shm region. Maybe we can add this check on a separate PR, unless you think the shm registration bound check improvement and this one must be committed together?

src/shared_memory_manager.cc
nnshah1, line 351: Small suggestion - what if we refactored to set the offsets, check them, and then only set the shm variable? I think one thing was a little odd was seeing the address set and then the offsets checked. 

By setting the variables you can also keep a single place to set the address - just a thought.

```
size_t max_offset = 0;
size_t offser = 0;
 
if (it->second->byte_size_ > 0) {
    max_offset = it->second->byte_size_ - 1;
  }
  if (it->second->kind_ == TRITONSERVER_MEMORY_CPU) {
     offset = it->second->offset + offset;
     max_offset += it -> second-> offset;
  } 
 
  if (offset > max_offset) {
    *shm_mapped_addr = nullptr;
    return TRITONSERVER_ErrorNew(
        std::string("Invalid offset for shared memory region: '" + name + "'")
            .c_str());
  }
   *shm_mapped_addr = (void*)((uint8_t*)it->second->mapped_addr_ + offset);
```

nv-kmcgill53:
> @nv-kmcgill53 if I understand correctly, you are talking about an improved bound check when _**registering**_ a shm region, not when _**inferencing**_ with a shm region? The next next ticket after this one is about improving checks when registering a shm region. Maybe we can add this check on a separate PR, unless you think the shm registration bound check improvement and this one must be committed together?

Yes, if there is other work slated for improvement on the *registering* of shared memory then, yes, that change should go in there.

kthui:
> > @nv-kmcgill53 if I understand correctly, you are talking about an improved bound check when _**registering**_ a shm region, not when _**inferencing**_ with a shm region? The next next ticket after this one is about improving checks when registering a shm region. Maybe we can add this check on a separate PR, unless you think the shm registration bound check improvement and this one must be committed together?
> 
> Yes, if there is other work slated for improvement on the _registering_ of shared memory then, yes, that change should go in there.

Sure. Linked this conversation in that ticket.

src/shared_memory_manager.cc
kthui: Sure. [Separate logic between computation and validation](https://github.com/triton-inference-server/server/pull/6914/commits/c750fdf9f4b2bda12b359d219d76f02900528ce1)

rmccorm4 (approved):
LGTM assuming the byte size check will be in a follow up

PR #6929, opened 2024-02-29
Add note on spacing

### LINKED ISSUE #6924: Issue on page /user_guide/response_cache.html
Incorrect documentation:
This page mentions that cache config is set by `--cache-config <cache_name>,<key>=<value>`, but actually it should be `--cache-config=<cache_name>,<key>=<value>`.
Setting the former results in an error and the docstring documentation says the following:
```
--cache-config <<string>,<string>=<string>>
	Specify a cache-specific configuration setting. The format
	of this flag is --cache-config=<cache_name>,<setting>=<value>. Where
	<cache_name> is the name of the cache, such as 'local' or 'redis'.
	Example: --cache-config=local,size=1048576 will configure a 'local'
	cache implementation with a fixed buffer pool of size 1048576 bytes.
```

rmccorm4:
Fixes #6924

### DIFF
--- a/docs/user_guide/model_configuration.md
+++ b/docs/user_guide/model_configuration.md
@@ -723,7 +723,7 @@ The above configuration creates 3 model instances, one on each device
 themselves as "R1" is local for their own device, however, they will
 contend for "R2" because it is specified as a global resource which
 means "R2" is shared across the system. Though these instances don't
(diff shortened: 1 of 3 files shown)


rmccorm4:
CC @anantzoid

fpetrini15 (approved):
Thanks for fielding this!

PR #6938, opened 2024-03-04
Add test for max queue delay timeout prompt response

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/333

Add a test for the prompt rejection of request upon timeout, without waiting for the availability of a payload slot.


oandreeva-nv (approved):
TMLG

PR #6941, opened 2024-03-04
Update Dockerfile.sdk with OpenAI support

tgerdesnv:
Add flag to enable compile of OpenAI support in PA

### DIFF
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -127,6 +127,7 @@ RUN cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
           -DTRITON_ENABLE_PERF_ANALYZER_C_API=ON \
           -DTRITON_ENABLE_PERF_ANALYZER_TFS=ON \
           -DTRITON_ENABLE_PERF_ANALYZER_TS=ON \
+          -DTRITON_ENABLE_PERF_ANALYZER_OPENAI=ON \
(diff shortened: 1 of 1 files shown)


Dockerfile.sdk
```diff
@@ -127,6 +127,7 @@ RUN cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
           -DTRITON_ENABLE_PERF_ANALYZER_C_API=ON \
           -DTRITON_ENABLE_PERF_ANALYZER_TFS=ON \
           -DTRITON_ENABLE_PERF_ANALYZER_TS=ON \
+          -DTRITON_ENABLE_PERF_ANALYZER_OPENAI=ON \
```
tgerdesnv, line 130: @mc-nv is this used in the CI? If so, what do I need to run to test it? If not, is there an additional place I should be applying this change?

mc-nv: > is this used in the CI? If so, what do I need to run to test it?

No, It's not used by GitLab itself. <br>
It updates cached variables `-DTRITON_ENABLE_PERF_ANALYZER_OPENAI=ON`. For CMake it means that the list of the  variables during the runtime will be extended with provided one which may change the result outputs.

>  If not, is there an additional place I should be applying this change?

We generally building `client` repository code using `Dockerfile.sdk` instructions, exception is Windows SDK build, which have to be updated to ship binaries compiled with given variable `-DTRITON_ENABLE_PERF_ANALYZER_OPENAI=ON`.

mc-nv (approved):

PR #6955, opened 2024-03-06
Update Dockerfile.win10.min

mc-nv:


Dockerfile.win10.min
```diff
@@ -87,7 +75,7 @@ RUN choco install git docker unzip -y
 #
 # Installing python
 #
-ARG PYTHON_VERSION=3.10.11
+ARG PYTHON_VERSION=3.8.10
```
nv-kmcgill53, line 78: I forgot the reasoning for the downgrade of python, could you please explain again?

nv-kmcgill53 (approved):
LGTM, just want a clarification on the python version change.

Dockerfile.win10.min
mc-nv: It's throw compilation errors, by downgrading it we were able to bypass it.

Issue #6887, opened 2024-02-15
Async HTTP Python Client not working properly


mutkach:
**Description**

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa6 in position 5: invalid start byte` when requesting a model config using an async http client. 

When observing the http response directly using `tritonclient.http.aio` without wrapper, I've noticed that async client does not decompress the http response by itself, so it seems to be fixable with setting `auto_decompress=True` on `aiohttp.ClientSession`. I believe in my case the compression is imposed elsewhere (by nginx?). 

If that was intended (as the `auto_decompress=True` is the default setting) then some additional logic is required to process the compressed response (calling to `brotli.decompress()` fixed it in my case). 
In any case I'll be happy to provide the necessary fixes.

**Triton Information**
- Triton server version 2.38.0
- Triton container `23.09` 
- `tritonclient==2.33.0, 2.42.0`
- `nvidia-pytriton==0.2.5, 0.3.0` and `0.5.1` when built from source

**To Reproduce**
 
```
import asyncio
import numpy
from pytriton.client import AsyncioModelClient, ModelClient
from pytriton.client.utils import get_model_config

HOST = "HOST:80"
model_name = "intent_classifier" #does not work
message = "Simple and correct query for testing"

async def run_classification(inferer_server_endpoint: str, clf_model_name: str, message: str, **_):
    async with AsyncioModelClient(inferer_server_endpoint, clf_model_name) as client:
        inference_client = client.create_client_from_url(inferer_server_endpoint)
        config = await inference_client.get_model_config(model_name)
        print(config)

def sync_classification(inferer_server_endpoint: str, clf_model_name: str, message: str, **_):
    with ModelClient(inferer_server_endpoint, clf_model_name) as client:
        inference_client = client.create_client_from_url(inferer_server_endpoint)
        config = inference_client.get_model_config(model_name)
        print(config)

if __name__ == "__main__":
    sync_classification(HOST, model_name, message) # Works
    #asyncio.run(run_classification(HOST, model_name, message)) # Does not
```

config is attached
[classifier_config.txt](https://github.com/triton-inference-server/server/files/14293384/classifier_config.txt)

**Expected behavior**
Model configuration should be returned in a valid json format to be parsed into python dict

kthui:
Thanks for reporting the issue, I have filed a ticket for us to investigate further.

> In any case I'll be happy to provide the necessary fixes.

Any contribution is welcomed!

[...]

PR #6975, opened 2024-03-12
Add AsyncIO HTTP compression test

kthui:
Verifies the Python AsyncIO HTTP client can handle data compression, plus some refactoring.

Original request and response compression algorithms:
| request | response | request != response | is original test case |
| ----- | ----- | ----- | ----- |
| - | - | ❌ | ❌ |
| - | deflate | ✅ | ✅ |
| deflate | - | ✅ | ✅ |
| deflate  | deflate | ❌ | ❌ |
| - | gzip | ✅ | ✅ |
| gzip | - | ✅ | ✅ |
| gzip | gzip | ❌ | ❌ |
| deflate | gzip | ✅ | ✅ |
| gzip | deflate | ✅ | ✅ |

closes: https://github.com/triton-inference-server/server/issues/6887

### DIFF
--- a/qa/L0_data_compression/test.sh
+++ b/qa/L0_data_compression/test.sh
@@ -70,7 +70,37 @@ fi
 set -e
 
 # End-to-end testing with simple model
+function run_data_compression_infer_client() {
+    local client_path=$1
+    local request_algorithm=$2
+    local response_algorithm=$3
+    local log_path=$4
+
+    local python_or_cpp=`echo -n "$client_path" | tail -c 3`
+    if [ "$python_or_cpp" == ".py" ]; then
+        local infer_client="python $client_path"
+        local request_algorithm_parameter="--request-compression-algorithm $request_algorithm"
+        local response_algorithm_parameter="--response-compression-algorithm $response_algorithm"
+    else  # C++
+        local infer_client=$client_path
+        local request_algorithm_parameter="-i $request_algorithm"
+        local response_algorithm_parameter="-o $response_algorithm"
+    fi
+
+    if [ "$request_algorithm" != "" ] && [ "$response_algorithm" == "" ]; then
+        $infer_client -v $request_algorithm_parameter >> $log_path 2>&1
+        return $?
+    fi
+    if [ "$request_algorithm" == "" ] && [ "$response_algorithm" != "" ]; then
+        $infer_client -v $response_algorithm_parameter >> $log_path 2>&1
+        return $?
+    fi
+    $infer_client -v $request_algorithm_parameter $response_algorithm_parameter >> $log_path 2>&1
+    return $?
+}
+
 SIMPLE_INFER_CLIENT_PY=../clients/simple_http_infer_client.py
+SIMPLE_AIO_INFER_CLIENT_PY=../clients/simple_http_aio_infer_client.py
 SIMPLE_INFER_CLIENT=../clients/simple_http_infer_client
 
 CLIENT_LOG=`pwd`/client.log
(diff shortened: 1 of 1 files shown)


qa/L0_data_compression/test.sh
```diff
@@ -70,7 +70,37 @@ fi
 set -e
 
 # End-to-end testing with simple model
+function run_data_compression_infer_client() {
+    local client_path=$1
+    local request_algorithm=$2
+    local response_algorithm=$3
+    local log_path=$4
+
+    local python_or_cpp=`echo -n "$client_path" | tail -c 3`
```
fpetrini15, line 79: [Optional] I'm only a little weary of using a hard count from the end to determine the file extension. Technically, this would work for all extensions I can imagine us using in the future e.g., `.py`, `.cc`, `.rs`, etc. However, if the extension was `.cpp` as the variable name suggests, this line would not store the intended string.

Maybe it would be more future-proof if we did something like:
```bash
filename=$(basename ${client_path})
extension=".${filename##*.}"
```

fpetrini15:
Really great work condensing the tests Jacky; it's much cleaner now! Also the table in the description was very slick 🚀.

Just one minor ~optional~ comment.

qa/L0_data_compression/test.sh
```diff
@@ -70,7 +70,37 @@ fi
...
+
+    if [ "$request_algorithm" != "" ] && [ "$response_algorithm" == "" ]; then
+        $infer_client -v $request_algorithm_parameter >> $log_path 2>&1
+        return $?
+    fi
+    if [ "$request_algorithm" == "" ] && [ "$response_algorithm" != "" ]; then
+        $infer_client -v $response_algorithm_parameter >> $log_path 2>&1
+        return $?
+    fi
+    $infer_client -v $request_algorithm_parameter $response_algorithm_parameter >> $log_path 2>&1
```
rmccorm4, line 98: Also optional, but an alternative I find slightly quicker to parse is to build up the args and call the command in only one place, something like this:
```suggestion
    compression_args=""
    if [ "$request_algorithm" != "" ]; then
        compression_args+="${request_algorithm_parameter} "
    fi
    
    if [ "$response_algorithm" != "" ]; then
        compression_args+="${response_algorithm_parameter} "
    fi
    
    $infer_client ${compression_args} >> $log_path 2>&1
```

rmccorm4 (dismissed):
Nice cleanup Jacky! :rocket: 

Also big fan of the table in the description to summarize the changes!

qa/L0_data_compression/test.sh
kthui: This is a good suggestion if we are sure there is an extension (e.g. `file.py`, `file.cpp`), but there is no extension for the c++ executable, i.e.
```
$ path="../clients/simple_http_infer_client" && filename=$(basename ${path}) && extension=".${filename##*.}" && echo $extension
.simple_http_infer_client
```
which I think `.simple_http_infer_client` is no better than the last three characters `ent`.

I don't know if one day there might be a Java executable, e.g. `../clients/simple_http_java_infer_client`, and uses a different set of command line options than the c++ one, which we will have to come back and rewrite the test anyway.

What do you think about leaving it simple for now (just the last three characters)?

fpetrini15: Oof 🤦‍♂️ that's a really good point, I was stuck in source file land. In that case I am okay leaving it checking the last three characters!

rmccorm4: Personally I'm fine with current implementation, or just making it more straightforward and just defining what type each client is:
(may not be syntactically correct)
```bash
function run_data_compression_infer_client {
    ...
    if [ "$request_algorithm" != "" ] && [ "$response_algorithm" == "" ]; then
        $infer_client -v $request_algorithm_parameter >> $log_path 2>&1
        return $?
    fi
    if [ "$request_algorithm" == "" ] && [ "$response_algorithm" != "" ]; then
        $infer_client -v $response_algorithm_parameter >> $log_path 2>&1
        return $?
    fi

    $infer_client -v $request_algorithm_parameter $response_algorithm_parameter >> $log_path 2>&1
    if [ $? -ne 0 ]; then
       RET=1
       echo "<this specific test failed>"
    fi
}


function run_python_data_compression_test {
        local infer_client="python $client_path"
        local request_algorithm_parameter="--request-compression-algorithm $request_algorithm"
        local response_algorithm_parameter="--response-compression-algorithm $response_algorithm"
        run_data_compression_infer_client "${infer_client}" "$REQUEST_ALGORITHM" "$RESPONSE_ALGORITHM" "$CLIENT_LOG"
}

function run_cpp_data_compression_test {
  ...
}

python_clients="${SIMPLE_INFER_CLIENT_PY} ${SIMPLE_AIO_INFER_CLIENT_PY}"
cpp_clients = "${SIMPLE_INFER_CLIENT}"

    for REQUEST_ALGORITHM in "deflate" "gzip" ""; do
        for RESPONSE_ALGORITHM in "deflate" "gzip" ""; do
            if [ "$REQUEST_ALGORITHM" == "$RESPONSE_ALGORITHM" ]; then
                continue
            fi

            set +e
            for client in ${PYTHON_CLIENTS}; do
              run_python_data_compression_test
            }

            for client in ${CPP_CLIENTS}; do
              run_cpp_data_compression_test
            }
            set -e
        done
    done
```

but either way I don't think this part needs to be over-engineered

qa/L0_data_compression/test.sh
kthui: Sure, build up the arguments in one place: [Improve command line option handling](https://github.com/triton-inference-server/server/pull/6975/commits/1d869967bcb7e540ef8c60eb845ef57a07335523)

fpetrini15 (approved):
LGTM, great work!

PR #6985, opened 2024-03-14
Exposing trace context to python backend

oandreeva-nv:
The purpose of this PR:
This PR is a Stage 1 towards adding an ability for users to create custom traces in python backend. 
This set of PRs is heavily focused on OpenTelemetry trace mode. 

Functionality:
- Prepares OpenTelemetry trace context via `PrepareTraceContext` and attaches it to TRITONSERVER_InferenceTrace instance through `TRITONSERVER_InferenceTraceSetContex` at the moment in time, when `REQUEST_START` callback happens. That one is happening from the core side, so when PBE model starts execution, it should have context attached to its trace.
- Context is passed as a string and for OpenTelemetry is essentially 2 headers `traceparent` : "..." and `tracestate`: "...". Some of you may know about them from OpenTelemetry context propagation discussions. It's the same. 
- We do nothing with trace Context in Triton mode. On the core side `TRITONSERVER_InferenceTrace` is initiated with an empty string as a context, since it is never updated, pbe will get an empty string and return `None` with `request.trace().get_context()` 

Below is a POC illustrated.

`mode=triton` tests are on the way.

Related PRs:
python_be: https://github.com/triton-inference-server/python_backend/pull/346
core: https://github.com/triton-inference-server/core/pull/334
  

POC:
image

created with model.py:
```python
import time
import json

import numpy as np
import triton_python_backend_utils as pb_utils

from opentelemetry import trace
from opentelemetry.propagate import inject
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import (
    BatchSpanProcessor
)
from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter



class TritonPythonModel:

    def initialize(self, args):
    	trace.set_tracer_provider(TracerProvider())
    	self.tracer = trace.get_tracer_provider().get_tracer("pbe")
    	trace.get_tracer_provider().add_span_processor(
	    BatchSpanProcessor(OTLPSpanExporter(endpoint="http://localhost:4318/v1/traces"))
	)

    def execute(self, requests):
        logger = pb_utils.Logger
        logger.log_info(f"got {len(requests)} requests")

        responses = []
        for request in requests:
            inp = pb_utils.get_input_tensor_by_name(request, "INPUT0").as_numpy()
            context = request.trace().get_context()
            if context == None:
            	context = "{}"
            ctx = TraceContextTextMapPropagator().extract(carrier=json.loads(context))

            # emulate some work
            with self.tracer.start_as_current_span('child', context=ctx) as span:
            	time.sleep(3)
            
            time.sleep(3)

            output_tensors = [pb_utils.Tensor("OUTPUT0", inp.astype(np.float32))]
            inference_response = pb_utils.InferenceResponse(output_tensors=output_tensors)
            responses.append(inference_response)

        return responses
```

### DIFF
--- a/qa/L0_trace/opentelemetry_unittest.py
+++ b/qa/L0_trace/opentelemetry_unittest.py
@@ -41,7 +41,7 @@ import requests
 import test_util as tu
 import tritonclient.grpc as grpcclient
 import tritonclient.http as httpclient
-from tritonclient.utils import InferenceServerException, np_to_triton_dtype
+from tritonclient.utils import InferenceServerException
 
 NO_PARENT_SPAN_ID = ""
 COLLECTOR_TIMEOUT = 10
--- a/qa/L0_trace/opentelemetry_unittest.py
+++ b/qa/L0_trace/opentelemetry_unittest.py
@@ -29,6 +29,7 @@ import sys
 sys.path.append("../common")
 import json
 import queue
+import re
 import shutil
 import subprocess
 import time
@@ -40,7 +41,7 @@ import requests
 import test_util as tu
 import tritonclient.grpc as grpcclient
 import tritonclient.http as httpclient
-from tritonclient.utils import InferenceServerException
+from tritonclient.utils import InferenceServerException, np_to_triton_dtype
 
 NO_PARENT_SPAN_ID = ""
 COLLECTOR_TIMEOUT = 10
@@ -755,6 +756,22 @@ class OpenTelemetryTest(tu.TestResultCollector):
         )
         time.sleep(5)
         self._test_simple_trace(headers=self.client_headers)
+
+    def test_trace_context_exposed_to_pbe(self):
+        """
+        Tests trace context, propagated to python backend.
+        """
+        triton_client_http = httpclient.InferenceServerClient(
+            "localhost:8000", verbose=True
+        )
+        xs = np.zeros((1, 1)).astype(np.float32)
+        inputs = httpclient.InferInput("INPUT0", xs.shape, np_to_triton_dtype(xs.dtype))
+        result = triton_client_http.infer(
+            self.ensemble_model_name, inputs, headers=self.client_headers
+        )
+        context = result.as_numpy("OUTPUT0")
+        context_pattern = re.compile(r"\d{2}-[0-9a-f]{32}-[0-9a-f]{16}-\d{2}")
+        self.assertIsNotNone(re.match(context_pattern, context))
 
 
 if __name__ == "__main__":
--- a/qa/L0_trace/opentelemetry_unittest.py
+++ b/qa/L0_trace/opentelemetry_unittest.py
@@ -105,6 +105,7 @@ class OpenTelemetryTest(tu.TestResultCollector):
         self.simple_model_name = "simple"
         self.ensemble_model_name = "ensemble_add_sub_int32_int32_int32"
         self.bls_model_name = "bls_simple"
+        self.trace_context_model = "trace_context"
         self.test_models = [
             self.simple_model_name,
             self.ensemble_model_name,
@@ -764,14 +765,19 @@ class OpenTelemetryTest(tu.TestResultCollector):
         triton_client_http = httpclient.InferenceServerClient(
             "localhost:8000", verbose=True
         )
-        xs = np.zeros((1, 1)).astype(np.float32)
-        inputs = httpclient.InferInput("INPUT0", xs.shape, np_to_triton_dtype(xs.dtype))
-        result = triton_client_http.infer(
-            self.ensemble_model_name, inputs, headers=self.client_headers
-        )
-        context = result.as_numpy("OUTPUT0")
+        expect_none = np.array([False], dtype=bool)
+        inputs = httpclient.InferInput("expect_none", [1], "BOOL")
+        inputs.set_data_from_numpy(expect_none)
+        try:
+            result = triton_client_http.infer(self.trace_context_model, inputs=[inputs])
+        except InferenceServerException as e:
+            self.fail(e.message())
+
+        context = result.as_numpy("OUTPUT0")[()].decode("utf-8")
+        context = json.loads(context)
+        self.assertIn("traceparent", context.keys())
         context_pattern = re.compile(r"\d{2}-[0-9a-f]{32}-[0-9a-f]{16}-\d{2}")
-        self.assertIsNotNone(re.match(context_pattern, context))
+        self.assertIsNotNone(re.match(context_pattern, context["traceparent"]))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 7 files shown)


src/tracer.cc
```diff
@@ -300,7 +300,9 @@ TraceManager::GetTraceStartOptions(
 {
   TraceManager::TraceStartOptions start_options;
   GetTraceSetting(model_name, start_options.trace_setting);
-  if (start_options.trace_setting->mode_ == TRACE_MODE_OPENTELEMETRY) {
+  if (!start_options.trace_setting->level_ ==
+          TRITONSERVER_TRACE_LEVEL_DISABLED &&
+      start_options.trace_setting->mode_ == TRACE_MODE_OPENTELEMETRY) {
```
oandreeva-nv, line 305: This fixes memory leak when `mode` is opentelemetry, but trace level is `OFF`, i.e. tracing disabled. Not happening in the default scenario, only when user specifies `mode=opentelemetry`, but doesn't specify `level=TIMESTAMPS`

src/tracer.cc
```diff
@@ -560,6 +562,9 @@ TraceManager::Trace::StartSpan(
     if (std::string(request_id) != "") {
       span->SetAttribute("triton.request_id", request_id);
     }
+    triton::common::TritonJson::WriteBuffer buffer;
+    PrepareTraceContext(span, &buffer);
```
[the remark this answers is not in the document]
oandreeva-nv: I would like users to be more clear in what they want. Since by default trace mode is `triton` on front side, I think it is reasonable to have this requirement on the python_be side as well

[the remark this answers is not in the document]
oandreeva-nv: Since in this stage mode is not exposed to python_be, I think it is reasonable to get rid of `mode` argument in `get_context`. It will still return `None` if passed context is an empty string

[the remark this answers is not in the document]
rmccorm4: I prefer not having the `mode` for now if not needed, since it seems like the user can choose the mode, but the mode is already set on the server side.

qa/L0_trace/trace_context.py
```diff
@@ -0,0 +1,48 @@
...
+                auto_complete_model_config.add_output(output)
+
+        auto_complete_model_config.set_max_batch_size(1)
+
+        return auto_complete_model_config
+
+    def initialize(self, args):
+        self.model_config = json.loads(args["model_config"])
+        output_config = pb_utils.get_output_config_by_name(self.model_config, "OUTPUT0")
+        self.output_dtype = pb_utils.triton_string_to_numpy(output_config["data_type"])
```
rmccorm4, line 36: I think these are unused
```suggestion
        pass
```

qa/L0_trace/trace_context.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def execute(self, requests):
+        responses = []
+        for request in requests:
+            context = request.trace().get_context(mode="opentelemetry")
```
rmccorm4, line 41: I prefer removing `mode` if not needed, as discussed in this thread: https://github.com/triton-inference-server/server/pull/6985/files#r1525367279

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -756,6 +757,22 @@ def test_sagemaker_invoke_trace_simple_model_context_propagation(self):
         time.sleep(5)
         self._test_simple_trace(headers=self.client_headers)
 
+    def test_trace_context_exposed_to_pbe(self):
```
rmccorm4, line 760: I think having a test for a case where `request.trace().get_context()` is `None` (such as server in `triton` mode) would be good.

src/tracer.cc
rmccorm4: Can be follow-up if complicated, but shouldn't the code not get this far if `--trace-level OFF` (tracing disabled)?

I would expect something earlier in the call chain looking like:
```cpp
// ex: StartTrace() - don't know the entrypoint
Status SomeTraceEntrypointFunction() {
  if (level == TRITONSERVER_TRACE_LEVEL_DISABLED) {
    return Status(INVALID_ARG, "Tracing is disabled");
    // or depending on how code works
    return Status::Success; // do nothing when tracing disabled
  }
}
```

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: yeap, on the way

src/tracer.cc
oandreeva-nv: The problem is `GetTraceStartOptions` is called from http/grpc code and if `mode` was specifically chosen as `opentelemetry` , we check for OTel context in headers, if those exist, we do a force sample. 

Thus, added check will force sample (based on the context) trace, only when `level` != `OFF`.

src/tracer.cc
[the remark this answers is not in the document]
oandreeva-nv: refactored

qa/L0_trace/opentelemetry_unittest.py
oandreeva-nv: added more tests: https://github.com/triton-inference-server/server/pull/6985/files#diff-27e9ca4cd8f292f0e5a94bbabb243e40fe0d47cd55164a0cb49ae6e093c9fbe7R1041-R1083

src/tracer.cc
rmccorm4: I see, thanks for explaining 🙏

rmccorm4 (dismissed):

qa/L0_trace/trace_context.py
oandreeva-nv: updated

qa/L0_trace/trace_context.py
oandreeva-nv: cleaned up

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -756,6 +758,27 @@ def test_sagemaker_invoke_trace_simple_model_context_propagation(self):
...
+        """
+        Tests trace context, propagated to python backend.
```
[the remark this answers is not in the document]
oandreeva-nv: I'll prepare a follow up PR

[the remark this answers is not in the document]
oandreeva-nv: For the reference: DLIS-6341

rmccorm4 (approved):
Nice work :rocket:

qa/L0_trace/trace_context.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
kthui, line 1: @oandreeva-nv do we need a copyright statement for this file?

oandreeva-nv: Good catch, will follow up shortly

PR #6986, opened 2024-03-15
HTTP live connections on server shutdown

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/335

When graceful shutdown begin, 
1. The HTTP endpoints stop accepting new connections (existing connections may continue).
2. The server will poll until all connections on the HTTP endpoints end, or `exit_timeout_secs_` expires, whichever is earlier.
3. The server passes the remaining timeout to the core and initiates core shutdown.
4. The core begins its shutdown routine as it is currently.

Example shutdown:
```
I0315 19:34:17.024361 101974 main.cc:329] Timeout 30: Found 1 HTTP connections
I0315 19:34:18.024569 101974 main.cc:329] Timeout 29: Found 1 HTTP connections
I0315 19:34:19.024747 101974 main.cc:329] Timeout 28: Found 1 HTTP connections
I0315 19:34:20.025022 101974 server.cc:307] Waiting for in-flight requests to complete.
I0315 19:34:20.025078 101974 server.cc:323] Timeout 27: Found 0 model versions that have in-flight inferences
I0315 19:34:20.025321 101974 server.cc:338] All models are stopped, unloading models
I0315 19:34:20.025461 101974 server.cc:347] Timeout 27: Found 1 live models and 0 in-flight non-inference requests
I0315 19:34:21.025611 101974 server.cc:347] Timeout 26: Found 1 live models and 0 in-flight non-inference requests
I0315 19:34:21.440670 101974 model_lifecycle.cc:620] successfully unloaded 'add_sub' version 1
I0315 19:34:22.025764 101974 server.cc:347] Timeout 25: Found 0 live models and 0 in-flight non-inference requests
```

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -238,14 +238,14 @@ HTTPServer::Start()
 TRITONSERVER_Error*
 HTTPServer::Stop(uint32_t* exit_timeout_secs, const std::string& service_name)
 {
+  {
+    std::lock_guard<std::mutex> lock(conn_mu_);
+    accepting_new_conn_ = false;
+  }
   if (exit_timeout_secs != nullptr) {
     while (*exit_timeout_secs > 0) {
-      {
-        std::lock_guard<std::mutex> lock(conn_mu_);
-        if (conn_cnt_ == 0) {
-          stop_accepting_new_conn_ = true;
-          break;
-        }
+      if (conn_cnt_ == 0) {  // conn_cnt_ can only decrease
+        break;
       }
       LOG_INFO << "Timeout " << *exit_timeout_secs << ": Found " << conn_cnt_
                << " " << service_name << " service connections";
@@ -286,13 +286,13 @@ HTTPServer::Dispatch(evhtp_request_t* req, void* arg)
 evhtp_res
 HTTPServer::NewConnection(evhtp_connection_t* conn, void* arg)
 {
-  HTTPServer* t = static_cast<HTTPServer*>(arg);
+  HTTPServer* server = static_cast<HTTPServer*>(arg);
   {
-    std::lock_guard<std::mutex> lock(t->conn_mu_);
-    if (t->stop_accepting_new_conn_) {
+    std::lock_guard<std::mutex> lock(server->conn_mu_);
+    if (!server->accepting_new_conn_) {
       return EVHTP_RES_SERVUNAVAIL;  // reset connection
     }
-    t->conn_cnt_++;
+    server->conn_cnt_++;
   }
   evhtp_connection_set_hook(
       conn, evhtp_hook_on_connection_fini,
@@ -303,10 +303,10 @@ HTTPServer::NewConnection(evhtp_connection_t* conn, void* arg)
 evhtp_res
 HTTPServer::EndConnection(evhtp_connection_t* conn, void* arg)
 {
-  HTTPServer* t = static_cast<HTTPServer*>(arg);
+  HTTPServer* server = static_cast<HTTPServer*>(arg);
   {
-    std::lock_guard<std::mutex> lock(t->conn_mu_);
-    t->conn_cnt_--;
+    std::lock_guard<std::mutex> lock(server->conn_mu_);
+    server->conn_cnt_--;
   }
   return EVHTP_RES_OK;
 }
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -238,8 +238,6 @@ HTTPServer::Start()
 TRITONSERVER_Error*
 HTTPServer::Stop(uint32_t* exit_timeout_secs, const std::string& service_name)
 {
-  accept_new_conn_ = false;
-
   if (exit_timeout_secs != nullptr) {
     while (*exit_timeout_secs > 0) {
       uint32_t conn_cnt = conn_cnt_;
@@ -285,9 +283,6 @@ HTTPServer::Dispatch(evhtp_request_t* req, void* arg)
 evhtp_res
 HTTPServer::NewConnection(evhtp_connection_t* conn, void* arg)
 {
-  if ((static_cast<HTTPServer*>(arg))->accept_new_conn_ == false) {
-    return EVHTP_RES_SERVUNAVAIL;
-  }
   evhtp_connection_set_hook(
       conn, evhtp_hook_on_connection_fini,
       (evhtp_hook)(void*)HTTPServer::EndConnection, arg);
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -240,11 +240,14 @@ HTTPServer::Stop(uint32_t* exit_timeout_secs, const std::string& service_name)
 {
   if (exit_timeout_secs != nullptr) {
     while (*exit_timeout_secs > 0) {
-      uint32_t conn_cnt = conn_cnt_;
-      if (conn_cnt == 0) {
-        break;
+      {
+        std::lock_guard<std::mutex> lock(conn_mu_);
+        if (conn_cnt_ == 0) {
+          stop_accepting_new_conn_ = true;
+          break;
+        }
       }
-      LOG_INFO << "Timeout " << *exit_timeout_secs << ": Found " << conn_cnt
+      LOG_INFO << "Timeout " << *exit_timeout_secs << ": Found " << conn_cnt_
                << " " << service_name << " service connections";
       std::this_thread::sleep_for(std::chrono::seconds(1));
       (*exit_timeout_secs)--;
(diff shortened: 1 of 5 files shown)


src/http_server.cc
```diff
@@ -267,6 +268,26 @@ HTTPServer::Dispatch(evhtp_request_t* req, void* arg)
   (static_cast<HTTPServer*>(arg))->Handle(req);
 }
 
+evhtp_res
+HTTPServer::NewConnection(evhtp_connection_t* conn, void* arg)
```
whoisj, line 272: does any of this need a fence or mutex?

kthui: I am not aware of any need from the evhtp library side, the example I found online does not use any fence/lock/mutex: https://chromium.googlesource.com/external/github.com/ellzey/libevhtp/+/refs/heads/master/examples/test.c#349

On our HTTP server side, the only thing shared is the `void* arg` which is simply `this`, the only two variables it interacts on are `accept_new_conn_` and `conn_cnt_` which are already safeguarded by `std::atomic<...>`. Here is their definition:
```
std::atomic<bool> accept_new_conn_;
std::atomic<uint32_t> conn_cnt_;
```

It is ok for them to use separate "mutexes" because the flow on the `HTTPServer::NewConnection()` is to take a peek into `accept_new_conn_`. If it is `false, the new connection is rejected (connection reset). If it is `true`, the function moves on to `conn_cnt_` and does not interact with `accept_new_conn_` anymore.

src/main.cc
```diff
@@ -299,6 +300,42 @@ StartEndpoints(
   return true;
 }
 
+uint32_t
+PrepareStopEndpoints(uint32_t timeout_secs)
```
nnshah1, line 304:
```suggestion
WaitForPendingConnections(uint32_t timeout_secs)
```

src/http_server.h
```diff
@@ -82,13 +83,17 @@ class HTTPServer {
   TRITONSERVER_Error* Start();
   TRITONSERVER_Error* Stop();
 
+  void StopAcceptingNewConnection() { accept_new_conn_ = false; }
```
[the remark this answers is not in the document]
nnshah1: @kthui - do these need to be public ?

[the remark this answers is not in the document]
kthui: Yes, they are called from the `main.cc` file, not internal to the `HTTPServer` class. It does not compile if they are placed under protected
```
/opt/server/src/main.cc: In function 'uint32_t {anonymous}::WaitForExistingConnections(uint32_t)':
/opt/server/src/main.cc:309:48: error: 'void triton::server::HTTPServer::StopAcceptingNewConnections()' is protected within this context
  309 |     g_http_service->StopAcceptingNewConnections();
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
In file included from /opt/server/src/command_line_parser.h:48,
                 from /opt/server/src/main.cc:57:
/opt/server/src/http_server.h:97:8: note: declared protected here
   97 |   void StopAcceptingNewConnections() { accept_new_conn_ = false; }
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/server/src/main.cc:320:55: error: 'uint32_t triton::server::HTTPServer::ConnectionCount()' is protected within this context
  320 |       uint32_t count = g_http_service->ConnectionCount();
      |                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
In file included from /opt/server/src/command_line_parser.h:48,
                 from /opt/server/src/main.cc:57:
/opt/server/src/http_server.h:98:12: note: declared protected here
   98 |   uint32_t ConnectionCount() { return conn_cnt_; }
      |            ^~~~~~~~~~~~~~~
```

src/main.cc
kthui: Picked the latter, since they are existing connections.

src/http_server.h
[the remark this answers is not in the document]
nnshah1: My initial thinking was that we would update ```http_server -> stop()``` to wait for existing connections before stopping the web server and then stop the server. 

So effectively - change the order in main such that:

1) Signal
2) StopEndpoints()
2(a) Wait for existing connections, close socket.
3) StopServer()

In that case the stop accepting new connections / connection count would be internal to the http_server object - and used within the stop() call.

Is that feasible / possible here - if it's complex to refactor into the http_server->stop() and reorder stopendpoints before stop server - we can add a ticket to follow up and a note to explore.

To me though the stop / wait for existing connections / connection count - should be part of the internal working of the http_server class and not directly exposed to the application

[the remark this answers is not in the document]
kthui: Old flow:
1. SIGINT
2. Instruct HTTP endpoints to stop accepting new connections
    2.1. Wait until all existing connections are closed
3. Stop the core
4. Stop HTTP endpoints

New flow:
1. SIGINT
2. Stop HTTP endpoints
    2.1. Stop accepting new connections
    2.2. Wait until all existing connections are closed
    2.3. Stop the endpoints
3. Stop the core

[Stop HTTP service immediately after all connections close](https://github.com/triton-inference-server/server/pull/6986/commits/5cd928125759b7618f354d8aa30bcf310a5eac79)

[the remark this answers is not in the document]
kthui: > One small thing: you may be able to remove the 'accept new connections' flag if you unbind the socket first

If checking `std::atomic<bool> accept_new_conn_` upon every new connection is a performance concern, we can remove it as well, making the flow:
1. SIGINT
2. Stop HTTP endpoints
    ~~2.1. Stop accepting new connections~~
    2.2. Wait until all ~~existing~~ connections are closed
    2.3. Stop the endpoints
3. Stop the core

src/main.cc
```diff
@@ -509,6 +520,11 @@ main(int argc, char** argv)
     triton::server::signal_exit_cv_.wait_for(lock, wait_timeout);
   }
 
+  // Stop the HTTP[, gRPC, and metrics] endpoints, and update exit timeout.
+  uint32_t exit_timeout_secs = g_triton_params.exit_timeout_secs_;
+  StopEndpoints(&exit_timeout_secs);
+  TRITONSERVER_ServerSetExitTimeout(server_ptr, exit_timeout_secs);
```
Tabrizian, line 526: Is error checking required?

kthui: The error is a part of the API, in case it is needed for the future, but it does not currently produce an error, since it is only assigning a parameter value to a class object.

If there is an error in setting core exit timeout, nothing can be done here but either push forward with the shutdown (with a possibly incorrect timeout value for the core) or force exit immediately. If the timeout value is incorrect on the core, it is visible on the server log during exit and an additional error message I think is redundant.

src/http_server.cc
```diff
@@ -267,6 +283,34 @@ HTTPServer::Dispatch(evhtp_request_t* req, void* arg)
...
+{
+  HTTPServer* t = static_cast<HTTPServer*>(arg);
```
nnshah1, line 289:
```suggestion
  HTTPServer* server = static_cast<HTTPServer*>(arg);
```

src/http_server.cc
```diff
@@ -235,8 +236,24 @@ HTTPServer::Start()
 }
 
 TRITONSERVER_Error*
-HTTPServer::Stop()
+HTTPServer::Stop(uint32_t* exit_timeout_secs, const std::string& service_name)
 {
+  if (exit_timeout_secs != nullptr) {
```
nnshah1, line 241: I think the general pattern would be something like:

```
stop_accepting_new_conn_ = true
if (exit_timeout_secs != nullptr) {
while (*exit_timeout_secs >0 && conn_cnt_>0) {
  sleep(1);
  *exit_timeout_secs --;
}
}
```
Note: not sure why we need to pass a pointer here for exit_timeout_secs - could we just set it to 0 if it wasn't set?
Feel like I'm missing something though that came up in testing. will try to sync offline

src/http_server.cc
kthui: updated

src/http_server.cc
kthui: updated to our conclusion synced offline

> not sure why we need to pass a pointer here for exit_timeout_secs - could we just set it to 0 if it wasn't set?

The purpose for the pointer is to pass the remaining exit_timeout_secs back to the caller, so it can be set in the core.

Both arguments are optional to remain compatible with the previous definition of `HTTPServer::Stop()`, and the exit_timeout_secs is default initialized to nullptr.

nnshah1 (dismissed):
LGTM

src/http_server.cc
```diff
@@ -235,8 +236,24 @@ HTTPServer::Start()
...
 TRITONSERVER_Error*
-HTTPServer::Stop()
+HTTPServer::Stop(uint32_t* exit_timeout_secs, const std::string& service_name)
 {
+  {
+    std::lock_guard<std::mutex> lock(conn_mu_);
+    accepting_new_conn_ = false;
+  }
+  if (exit_timeout_secs != nullptr) {
+    while (*exit_timeout_secs > 0) {
```
nnshah1, line 246:
```suggestion
    while ( (*exit_timeout_secs > 0) && (conn_cnt_>0)) {
```

nnshah1: small tweak - I think you can make this a little simpler by just checking for both conditions in the while loop:
```
while ( (*exit_timeout_secs>0) &&  (conn_cnt_>0)) {
  std::this_thread::sleep_for(std::chrono::seconds(1)));
 (*exit_timeout_secs)--;
}
```

The logic is a little easier to follow then a break in the loop - it becomes a natural condition for exit

kthui: sure, updated: [Group checking logic](https://github.com/triton-inference-server/server/pull/6986/commits/6dbd249c65215add12c900362c7a16b2c683fa34)

nnshah1 (approved):

PR #7025, opened 2024-03-22
Specify path for PyTorch model extension library

mc-nv:


nv-kmcgill53:
Is the pyTorch team adding this back to their containers? I thought they were not going to be shipping this because it was a mistake.

mc-nv:
> Is the pyTorch team adding this back to their containers? I thought they were not going to be shipping this because it was a mistake.

I made a mistake by removing it, by mistyping path for `custom_modulo`.
By default path is autogenerated unless variable is provided.

I went into code and found  that we can gain control over it.
https://github.com/pytorch/pytorch/blob/main/torch/utils/cpp_extension.py#L2032-L2045

nv-kmcgill53 (approved):

PR #7032, opened 2024-03-26
Update README.md 2.44.0 / 24.03

mc-nv:


README.md
```diff
@@ -30,5 +30,230 @@
...
+  protocol](https://github.com/kserve/kserve/tree/master/docs/predict-api/v2)
+- A [C API](docs/customization_guide/inference_protocols.md#in-process-triton-server-api) and
+  [Java API](docs/customization_guide/inference_protocols.md#java-bindings-for-in-process-triton-server-api)
+  allow Triton to link directly into your application for edge and other in-process use cases
+- [Metrics](docs/user_guide/metrics.md) indicating GPU utilization, server
+  throughput, server latency, and more
+
+**New to Triton Inference Server?** Make use of
+[these tutorials](https://github.com/triton-inference-server/tutorials)
+to begin your Triton journey!
```
nv-kmcgill53, line 77: @rmccorm4 Should we start pointing to the Triton CLI here, or do you need more time to sure up that repo?

nv-kmcgill53 (approved):

README.md
rmccorm4: I think we can hold off until we figure out the positioning and where to advertise it, but this is a good point we need to start thinking about soon. It's very LLM focused right now and may not apply as a good starting point generically here.

PR #7046, opened 2024-03-27
Add Documentation from Additional Repositories to nvidia.docs.com

yinggeh:
Closes issue DLIS-6234, Closes DLIS-3429

Usage of the new script (Part of gitlab PR) : 
python3 docs/generate_docs.py --repo-tag=client:main --repo-tag=python_backend:main --github-organization=triton-inference-server

We expect new repos to be added here.


yinggeh:
@nnshah1 Feel free to add more reviewers.

lkomali:
seems good to me. Waiting for Neelay's comments.

docs/contents.md
```diff
@@ -102,3 +102,46 @@ customization_guide/test
 examples/jetson/README
 examples/jetson/concurrency_and_dynamic_batching/README
 ```
+
+```{toctree}
+:maxdepth: 1
+:caption: Client
+
+client/README
```
tanmayv25, line 110: If we add/remove a section in the client, perf_analyzer or python backend repositories, then we would have to come back here in the server repo to add/remove it from here. 
This design doesn't seem scalable and exposes contents of one repository to another. 
Is it possible to have a file within client and python backend repo which encapsulates these sections within the specific repo only? We can then import these directly here.

yinggeh: This is absolutely possible. Note that
1. Since `contents.md` is the master_doc and we do not know which repos are included prior to running, contents.md must be generated dynamically.
2. Where do we want to put `contents.md` inside each repo? For example, `client/contents.md` or `client/docs/contents.md`? Right now `contents.md` will be the only file inside of `client/docs`.
We need to discuss on whether it brings us more benefits than overheads.

tanmayv25: As per our discussion we will address the scalability as a follow-up item.

tanmayv25:
@yinggeh  you'd have to rebase the branch with main.

yinggeh:
@tanmayv25 Merged latest main.

tanmayv25 (approved):

PR #7048, opened 2024-03-27
[Windows] Support CPU shared memory (Client/Frontend)

fpetrini15:
**Goal:** Support CPU shared memory between the server and client for Windows

**Sub-goals:** Modify L0_shared_memory to run on bare-metal Windows using only Python.

Client changes: https://github.com/triton-inference-server/client/pull/551

Some things to note: 
- When I can verify that the Linux tests pass using only the Python script, I will remove `test.sh`
- L0_shared_memory uses a graphdef model by default. I swapped it with Python so that it would be supported on both Windows and Linux. I still need to go back and investigate how the model ends up in L0_shared_memory (not generated by script) and remove it.
- Some of the default paths need to be modified to reflect the testing environment and will be modified pre-merge.

### DIFF
--- a/qa/L0_shared_memory/shared_memory_test.py
+++ b/qa/L0_shared_memory/shared_memory_test.py
@@ -56,7 +56,7 @@ class SharedMemoryTest(tu.TestResultCollector):
     def _setUp(self, protocol, log_file_path):
         self._tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR", "localhost")
         self._test_windows = bool(int(os.environ.get("TEST_WINDOWS", 0)))
-        self._shm_key_prefix = "/" if not self._test_windows else "Global\\"
+        self._shm_key_prefix = "/" if not self._test_windows else ""
         self._timeout = os.environ.get("SERVER_TIMEOUT", 120)
         self._protocol = protocol
         self._test_passed = False
@@ -95,9 +95,11 @@ class SharedMemoryTest(tu.TestResultCollector):
 
     def _build_server_args(self):
         if self._test_windows:
-            backend_dir = "C:\\opt\\tritonserver\\backends"
-            model_dir = "C:\\opt\\tritonserver\\qa\\L0_shared_memory\\models"
-            self._server_executable = "C:\\opt\\tritonserver\\bin\\tritonserver.exe"
+            backend_dir = os.environ.get("BACKEND_DIR", "C:\\opt\\tritonserver\\backends")
+            model_dir = os.environ.get("MODELDIR", (os.getcwd() + "\\models"))
+            self._server_executable = os.environ.get(
+                "SERVER", "C:\\tritonserver\\bin\\tritonserver.exe"
+            )
         else:
             triton_dir = os.environ.get("TRITON_DIR", "/opt/tritonserver")
             backend_dir = os.environ.get("BACKEND_DIR", f"{triton_dir}/backends")
@@ -178,21 +180,44 @@ class SharedMemoryTest(tu.TestResultCollector):
         input1_data = np.ones(shape=16, dtype=np.int32)
         shm.set_shared_memory_region(shm_ip0_handle, [input0_data])
         shm.set_shared_memory_region(shm_ip1_handle, [input1_data])
-        self.triton_client.register_system_shared_memory(
-            "input0_data", (self._shm_key_prefix + "input0_data"), register_byte_size, offset=register_offset
-        )
-        self.triton_client.register_system_shared_memory(
-            "input1_data", (self._shm_key_prefix + "input1_data"), register_byte_size, offset=register_offset
-        )
-        self.triton_client.register_system_shared_memory(
-            "output0_data", (self._shm_key_prefix + "output0_data"), register_byte_size, offset=register_offset
-        )
-        self.triton_client.register_system_shared_memory(
-            "output1_data", (self._shm_key_prefix + "output1_data"), register_byte_size, offset=register_offset
-        )
+        try:
+            self._triton_client.register_system_shared_memory(
+                "input0_data",
+                (self._shm_key_prefix + "input0_data"),
+                register_byte_size,
+                offset=register_offset,
+            )
+            self._triton_client.register_system_shared_memory(
+                "input1_data",
+                (self._shm_key_prefix + "input1_data"),
+                register_byte_size,
+                offset=register_offset,
+            )
+            self._triton_client.register_system_shared_memory(
+                "output0_data",
+                (self._shm_key_prefix + "output0_data"),
+                register_byte_size,
+                offset=register_offset,
+            )
+            self._triton_client.register_system_shared_memory(
+                "output1_data",
+                (self._shm_key_prefix + "output1_data"),
+                register_byte_size,
+                offset=register_offset,
+            )
+        except utils.InferenceServerException as e:
+            shm_handles = [
+                shm_ip0_handle,
+                shm_ip1_handle,
+                shm_op0_handle,
+                shm_op1_handle,
+            ]
+            self._cleanup_server(shm_handles)
+            raise (e)
         return [shm_ip0_handle, shm_ip1_handle, shm_op0_handle, shm_op1_handle]
 
     def _cleanup_server(self, shm_handles):
+        self._triton_client.unregister_system_shared_memory()
         for shm_handle in shm_handles:
             shm.destroy_shared_memory_region(shm_handle)
 
@@ -326,6 +351,7 @@ class SharedMemoryTest(tu.TestResultCollector):
             self.assertEqual(len(shm_status), 1)
         else:
             self.assertEqual(len(shm_status.regions), 1)
+        self._triton_client.unregister_system_shared_memory()
         shm.destroy_shared_memory_region(shm_op0_handle)
         self._test_passed = True
 
@@ -355,6 +381,7 @@ class SharedMemoryTest(tu.TestResultCollector):
             )
         except Exception as ex:
             self.assertIn("registering an active shared memory key", str(ex))
+        self._triton_client.unregister_system_shared_memory()
         shm.destroy_shared_memory_region(shm_op0_handle)
         self._test_passed = True
 
@@ -379,6 +406,7 @@ class SharedMemoryTest(tu.TestResultCollector):
             self.assertEqual(len(shm_status), 0)
         else:
             self.assertEqual(len(shm_status.regions), 0)
+        self._triton_client.unregister_system_shared_memory()
         shm.destroy_shared_memory_region(shm_op0_handle)
         self._test_passed = True
 
@@ -405,6 +433,7 @@ class SharedMemoryTest(tu.TestResultCollector):
             self.assertEqual(len(shm_status), 0)
         else:
             self.assertEqual(len(shm_status.regions), 0)
+        self._triton_client.unregister_system_shared_memory()
         shm.destroy_shared_memory_region(shm_op0_handle)
         self._test_passed = True
 
@@ -438,6 +467,7 @@ class SharedMemoryTest(tu.TestResultCollector):
             self.assertEqual(len(shm_status), 1)
         else:
             self.assertEqual(len(shm_status.regions), 1)
+        self._triton_client.unregister_system_shared_memory()
         shm.destroy_shared_memory_region(shm_op0_handle)
         self._test_passed = True
 
@@ -664,6 +694,7 @@ class SharedMemoryTest(tu.TestResultCollector):
                 register_byte_size=0,
                 register_offset=create_byte_size + 1,
             )
+        self._test_passed = True
 
 
 if __name__ == "__main__":
--- a/src/shared_memory_manager.cc
+++ b/src/shared_memory_manager.cc
@@ -36,6 +36,8 @@
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#else
+#define TRITON_SHM_FILE_ROOT "C:\\triton_shm\\"
 #endif
 
 namespace triton { namespace server {
@@ -122,7 +124,7 @@ OpenCudaIPCRegion(
 
 TRITONSERVER_Error*
 SharedMemoryManager::OpenSharedMemoryRegion(
-    const std::string& shm_key, ShmFile** shm_file)
+    const std::string& shm_key, std::shared_ptr<ShmFile>& shm_file)
 {
 #ifdef _WIN32
   HANDLE shm_handle = OpenFileMapping(
@@ -138,8 +140,7 @@ SharedMemoryManager::OpenSharedMemoryRegion(
         std::string("Unable to open shared memory region: '" + shm_key + "'")
             .c_str());
   }
-  // Dynamic memory will eventually be owned by uniqe_ptr
-  *shm_file = new ShmFile(shm_handle);
+  shm_file = std::make_shared<ShmFile>(shm_handle);
 #else
   // get shared memory region descriptor
   int shm_fd = shm_open(shm_key.c_str(), O_RDWR, S_IRUSR | S_IWUSR);
@@ -150,8 +151,7 @@ SharedMemoryManager::OpenSharedMemoryRegion(
         std::string("Unable to open shared memory region: '" + shm_key + "'")
             .c_str());
   }
-  // Dynamic memory will eventually be owned by uniqe_ptr
-  *shm_file = new ShmFile(shm_fd);
+  shm_file = std::make_shared<ShmFile>(shm_fd);
 #endif
   return nullptr;
 }
@@ -161,14 +161,42 @@ SharedMemoryManager::GetSharedMemoryRegionSize(
     const std::string& shm_key, ShmFile* shm_file, uint64_t* shm_region_size)
 {
 #ifdef WIN32
-  BY_HANDLE_FILE_INFORMATION info;
-  if(!GetFileInformationByHandle(shm_file->shm_handle_, &info)) {
-    LOG_VERBOSE(1) << "GetFileInformationByHandle failed with error code: " << GetWindowsError();
+  // Open file for reading
+  LPCSTR backing_file_path =
+      std::string(TRITON_SHM_FILE_ROOT + shm_key).c_str();
+  HANDLE backing_file_handle = CreateFile(
+      backing_file_path, 0, FILE_SHARE_READ, NULL, OPEN_EXISTING,
+      FILE_ATTRIBUTE_NORMAL, NULL);
+  if (backing_file_handle == INVALID_HANDLE_VALUE) {
+    LOG_VERBOSE(1) << "Failed to open backing file with error code: "
+                   << GetWindowsError();
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INTERNAL,
         std::string("Invalid shared memory region: '" + shm_key + "'").c_str());
   }
-  uint64_t file_size = ((uint64_t)info.nFileSizeHigh << 32) | info.nFileSizeLow;
+  // Construct its size
+  uint64_t file_size;
+  DWORD high_order_size;
+  DWORD low_order_size = GetFileSize(backing_file_handle, &high_order_size);
+  if (low_order_size == INVALID_FILE_SIZE) {
+    CloseHandle(backing_file_handle);
+    LOG_VERBOSE(1) << "GetFileSize failed with error code: "
+                   << GetWindowsError();
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        std::string("Invalid shared memory region: '" + shm_key + "'").c_str());
+  } else if (high_order_size != NULL) {
+    file_size = ((uint64_t)high_order_size << 32) | low_order_size;
+  } else {
+    file_size = low_order_size;
+  }
+  if (!CloseHandle(backing_file_handle)) {
+    LOG_VERBOSE(1) << "failed to close backing file with error: "
+                   << GetWindowsError();
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        std::string("Invalid shared memory region: '" + shm_key + "'").c_str());
+  }
   *shm_region_size = file_size;
 #else
   struct stat file_status;
@@ -201,6 +229,7 @@ SharedMemoryManager::CheckSharedMemoryRegionSize(
   RETURN_IF_ERR(GetSharedMemoryRegionSize(shm_key, shm_file, &shm_region_size));
   // User-provided offset and byte_size should not go out-of-bounds.
   if ((offset + byte_size) > shm_region_size) {
+    CloseSharedMemoryRegion(shm_file);
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INVALID_ARG,
         std::string(
@@ -263,7 +292,7 @@ SharedMemoryManager::MapSharedMemory(
       byte_size);
 
   if (*mapped_addr == NULL) {
-    CloseSharedMemoryRegion(shm_handle);
+    CloseSharedMemoryRegion(shm_file);
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INTERNAL,
         std::string(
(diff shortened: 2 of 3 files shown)


qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -29,98 +29,110 @@
...
-        shm_op0_handle = shm.create_shared_memory_region("dummy_data", "/dummy_data", 8)
-        shm.set_shared_memory_region(
-            shm_op0_handle, [np.array([1, 2], dtype=np.float32)]
+            self._url = self._tritonserver_ipaddr + ":8001"
+            self._triton_client = grpcclient.InferenceServerClient(
+                self._url, verbose=True
+            )
+        self._build_model_repo()
+        self._build_server_args()
+        self._shared_memory_test_server_log = open(log_file_path, "w")
```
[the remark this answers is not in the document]
fpetrini15: I believe we can safely ignore these warnings. The file is opened during `_setUp` and closed in `tearDown`. According to the docs, `tearDown` will [always be called](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown:~:text=Method%20called%20immediately%20after%20the%20test%20method%20has%20been%20called%20and%20the%20result%20recorded.%20This%20is%20called%20even%20if%20the%20test%20method%20raised%20an%20exception) when `setUp` is successful. In this case, since we use a custom version of "setUp", the default unittest `setUp` method will always be successful, so the file should always be closed.

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -29,98 +29,110 @@
...
+        self._server_process = util.run_server(
+            server_executable=self._server_executable,
+            launch_command=self._launch_command,
+            log_file=self._shared_memory_test_server_log,
+        )
+        util.wait_for_server_ready(
+            self._server_process, self._triton_client, self._timeout
+        )
+        # Redirect console output to client log and begin test
+        self._client_log = open(SharedMemoryTest.shared_memory_test_client_log, "a")
```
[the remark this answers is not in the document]
fpetrini15: I believe we can safely ignore these warnings. The file is opened during _setUp and closed in tearDown. According to the docs, tearDown will [always be called](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown:~:text=Method%20called%20immediately%20after%20the%20test%20method%20has%20been%20called%20and%20the%20result%20recorded.%20This%20is%20called%20even%20if%20the%20test%20method%20raised%20an%20exception) when setUp is successful. In this case, since we use a custom version of "setUp", the default unittest setUp method will always be successful, so the file should always be closed.

src/shared_memory_manager.cc
```diff
@@ -26,22 +26,199 @@
...
+  DWORD low_order_size = upperbound_size & 0xFFFFFFFF;
+
+  *shm_handle = OpenFileMapping(
+      FILE_MAP_ALL_ACCESS,  // read/write access
+      FALSE,                // cannot inherit handle
+      shm_key.c_str());     // name of mapping object
+
+  if (*shm_handle == NULL) {
+    LOG_VERBOSE(1) << "OpenFileMapping failed with error code: "
+                   << std::to_string(GetLastError());
```
nv-kmcgill53, line 65: This goes for all instances: I'd like to not just get the numerical error code back from windows, I'd like to see the message which it translates to as well. Please see [shared_library.cc](https://github.com/triton-inference-server/core/blob/main/src/shared_library.cc#L68-L79) in the core repo for ideas.  We should probably wrap this type of error handling in a function to make this easier to use.

src/shared_memory_manager.cc
```diff
@@ -60,72 +238,202 @@ SharedMemoryManager::GetCUDAHandle(
...
+  }
+
+  *memory_type = it->second->kind_;
+  *device_id = it->second->device_id_;
+
+  return nullptr;
 }
 
 TRITONSERVER_Error*
 SharedMemoryManager::GetStatus(
```
nv-kmcgill53, line 294: Does this need to be implemented twice? I understand the iteration cycle you went though, that's fine to make sure it works, but there is no Windows specific code in this function.

src/shared_memory_manager.cc
```diff
@@ -60,72 +238,202 @@ SharedMemoryManager::GetCUDAHandle(
...
-      std::string("Shared memory feature is currently not supported on Windows")
+      std::string(
+          "GPU shared memory features are currently not supported on Windows")
           .c_str());
 }
+
 #endif  // TRITON_ENABLE_GPU
 
 TRITONSERVER_Error*
 SharedMemoryManager::GetMemoryInfo(
```
nv-kmcgill53, line 249: Does this need to be implemented twice? I understand the iteration cycle you went though, that's fine to make sure it works, but there is no Windows specific code in this function.

src/shared_memory_manager.cc
```diff
@@ -60,72 +238,202 @@ SharedMemoryManager::GetCUDAHandle(
...
+         it != shared_memory_map_.cend(); it = next_it) {
+      ++next_it;
+      if (it->second->kind_ == TRITONSERVER_MEMORY_CPU) {
+        TRITONSERVER_Error* err = UnregisterHelper(it->first, memory_type);
+        if (err != nullptr) {
+          unregister_fails.push_back(it->first);
+        }
+      }
+    }
+  }
```
nv-kmcgill53, line 423: Add a comment about checking GPU shared memory here and to check if we still need a windows implementation of `UnregisterAll` once we have GPU shared memory support

src/shared_memory_manager.h
```diff
@@ -130,14 +106,80 @@ class SharedMemoryManager {
...
   /// specified type
   TRITONSERVER_Error* UnregisterHelper(
       const std::string& name, TRITONSERVER_MemoryType memory_type);
 
-  /// A struct that records the shared memory regions registered by the shared
-  /// memory manager.
+/// A struct that records the shared memory regions registered by the shared
+/// memory manager.
+#ifdef _WIN32
+  struct SharedMemoryInfo {
```
nv-kmcgill53, line 142: My intuition says that there is a better way to do this, but I don't have a solution atm. The only way these structs differ is the file `fd`/`HANDLE` which we hold onto. They are both "handles" in that they are both integers which we ask the operating system about and act on.  Should we alias/typedef the struct, make a union, something else? As a starting point maybe something like:

```
#ifdef _WIN32
  typedef HANDLE file_handle;
#else
  typedef int file_handle;
#endif
```

This would be defined at the top of the file so that we can replace the type of the `shm_handle_` as `file_handle` and only define the struct once. I'd ask around on this to make sure I'm not crazy.

nv-kmcgill53 (changes requested):

src/shared_memory_manager.cc
fpetrini15: This is a great idea! And given that Windows error handling will now be present in several repos, I'm wondering if it makes sense to extend Triton's general error handling to accept Windows error codes and convert them for this purpose.

src/shared_memory_manager.cc
fpetrini15: The current file structure has duplicated functions for Windows and Unix, with the Windows function simply returning a "not supported error".

Originally, I kept this structure for simplicity. However, as I continued developing, I found it more legible to have separate file segments for Windows/Unix, rather than inlining the differences with ifdefs. However, I understand the software bloat issue here and received similar feedback in the client PR, so I'll try to merge as much together as I can here as well.

src/shared_memory_manager.h
fpetrini15: Yes, in fact, that is exactly what I currently have implemented on the client side.

Tanmay's feedback was to make a substruct containing platform-dependent members. Currently it would only manage int vs HANDLE, but could be extended in the future. It would look something like this:
```c++
struct ShmFile {
#ifdef _WIN32
  HANDLE shm_file_;
  ShmFile(HANDLE shm_file): shm_file_(shm_file) {};
#else
  int platform_handle_;
  ShmFile(int shm_file): shm_file_(shm_file) {};
#endif // _WIN32
};
```
where `SharedMemoryInfo`'s `shm_handle` member would be a pointer to a ShmFile object.

src/shared_memory_manager.cc
fpetrini15: New error will look something like this:
```
OpenFileMapping failed with error code: 2 - The system cannot find the file specified.
```

src/shared_memory_manager.cc
fpetrini15: Apologies, I was concerned with the other changes and am looking at this comment post-push. Do you recall where this comment was in reference to?

nv-kmcgill53: This is in reference to the if-statement checking if the memory type is CPU. At the end of the if-block, where the `else` statement would go, I would like a comment stating that we will check for the GPU memory type once it's available. Also, I would like a comment to revisit this methods being a duplicate of the linux implementation once we check for the GPU type.

src/shared_memory_manager.h
```diff
@@ -130,31 +106,71 @@ class SharedMemoryManager {
...
+          mapped_addr_(mapped_addr), kind_(kind), device_id_(device_id)
     {
     }
 
     std::string name_;
     std::string shm_key_;
     size_t offset_;
     size_t byte_size_;
-    int shm_fd_;
+    std::unique_ptr<ShmFile> platform_handle_;
```
nv-kmcgill53, line 173: I didn't get this far down when I wrote the above comment. We are wrapping this struct in a `unique_ptr` anyway so why does the underlying data structure need to be a smart pointer? I am making a lot of rhetorical questions, but it's always asking what value are we getting out of using this more complicated thing?

Again, I would say this could be converted into a raw pointer with no impact as long as we free it in the destructor or whenever it needs to be freed. After looking at the code, `SharedMemoryInfo` isn't the one owning the pointer, the `shared_memory_map_` really is since this lives as long as the entry in the map does. Directly below, `mapped_addr_` would cause a memory leak if we didn't clean up the shared memory regions properly, but AFAIK it doesn't so we should be fine.

nv-kmcgill53 (commented):
This review is a bit ramble-y but it's very tricky as well. You've done a great job so far, I'm teasing out the nuances so you can provide a good template of how to program for multiple OSes in a sustainable way.

docs/protocol/extension_shared_memory.md
```diff
@@ -45,10 +45,10 @@ memory. These parameters and their type are:
   registered shared memory region. Region names share a namespace for
   system-shared-memory regions and CUDA-shared-memory regions.
 
-- "shared_memory_offset" : int64 value is the offset, in bytes, into
+- "shared_memory_offset" : unsigned integer value is the offset, in bytes, into
```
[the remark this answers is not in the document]
fpetrini15: We did not change to 32 bits. Technically, our python interface will always [pass](https://github.com/triton-inference-server/client/blob/main/src/python/library/tritonclient/utils/shared_memory/__init__.py#L67) a pointer to `uint64`. However, given that the C++ source takes in a `size_t`, which could either be a 32 or 64-bit unsigned integer depending on the OS, it seemed most accurate to me to simply say an "unsigned integer". 

Would you prefer the docs say "size_t"? "A 32 or 64-bit unsigned"? Or simply a "uint64"?

src/shared_memory_manager.cc
```diff
@@ -26,113 +26,59 @@
...
-#include "triton/common/logging.h"
+#endif
 
 namespace triton { namespace server {
 
 namespace {
 
+#ifdef _WIN32
+const std::string
+GetWindowsError()
```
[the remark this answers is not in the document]
fpetrini15: Glad you like it. It was inspired from a very wise review comment 😉

src/shared_memory_manager.h
```diff
@@ -130,31 +106,71 @@ class SharedMemoryManager {
...
   /// A helper function to remove the named shared memory blocks of
   /// specified type
   TRITONSERVER_Error* UnregisterHelper(
       const std::string& name, TRITONSERVER_MemoryType memory_type);
 
+  struct ShmFile {
+#ifdef _WIN32
+    HANDLE shm_file_;
+    ShmFile(void* shm_file) { shm_file_ = static_cast<HANDLE>(shm_file); };
+    HANDLE GetShmFile() { return shm_file_; };
```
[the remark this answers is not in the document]
fpetrini15: You are on the money when you suggest I designed it this way to provide a unified interface for both the HANDLE and unique_ptr.

I can change this to use a raw pointer, however, as I am sure you are aware, there has been a significant push to remove references to "new" and "delete" in C++ developer land. Additionally, use of raw pointers has been discouraged by the team in other PRs, granted the raw pointers were not pointing to primitives.

src/shared_memory_manager.cc
```diff
@@ -243,32 +253,37 @@ SharedMemoryManager::RegisterSystemSharedMemory(
 
   // register
   void* mapped_addr;
-  int shm_fd = -1;
+  // Safe initializer for Unix case where shm_file must be dereferenced to
+  // base in order to store file descriptor.
+  int shm_safe_initializer = -1;
```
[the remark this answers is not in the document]
fpetrini15: This is a great catch! I was preoccupied solving the Linux segfault, I didn't go back and rethink the HANDLE.

src/shared_memory_manager.cc
```diff
@@ -26,113 +26,59 @@
...
+  return retString;
+}
+#endif
+
 TRITONSERVER_Error*
-OpenSharedMemoryRegion(const std::string& shm_key, int* shm_fd)
+OpenSharedMemoryRegion(const std::string& shm_key, void** shm_file)
 {
+#ifdef _WIN32
+  HANDLE* shm_handle = static_cast<HANDLE*>(shm_file);
```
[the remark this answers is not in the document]
fpetrini15: 🧠

src/shared_memory_manager.h
fpetrini15: I think this is mostly to do with the safest possible design. The current main branch code in client uses raw pointers and leaks memory because it's easy to forget to release the memory. Using the unique_ptrs, what we lose in additional verbosity, we gain in not needing to update a destructor and being confident memory will be safely freed.

I think this will be particularly helpful when/if ShmFile becomes a more complex data structure.

qa/common/util.py
```diff
@@ -0,0 +1,103 @@
...
+
+def kill_server(server_process):
+    # Only kill process if it's stil running
+    if server_process and not server_process.poll():
+        print("*\n*\n*\nTerminating server\n*\n*\n*\n")
+        # Terminate gracefully for Linux
+        if sys.platform == "win32":
+            server_process.kill()
+        else:
+            server_process.send_signal(signal.SIGINT)
```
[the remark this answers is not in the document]
fpetrini15: To be clear, you're suggesting something like:

```python
...
else:
    server_process.send_signal(signal.SIGINT)

try:
    server_process.wait(timeout=30)
except subprocess.TimeoutExpired:
    print("=== Server did not shutdown properly")
    server_process.kill()

```

docs/protocol/extension_shared_memory.md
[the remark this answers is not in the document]
nv-kmcgill53: I think `size_t` is the best solution. It packs all the issues we are describing into one word which devs can understand.

qa/common/util.py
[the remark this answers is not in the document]
nv-kmcgill53: Ya... But I'm not sure of the impact this has on just testing. I guess if we are starting and stopping triton multiple times in a script we could potentially hang. But idk if we just wait for this to become a real problem.

src/shared_memory_manager.cc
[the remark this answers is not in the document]
nv-kmcgill53: To ponder the topic just a step further: I looked through the documentation of `static_cast` and expected there to be some sort of error if the cast is in someway "bad." There isn't, but probably because the underlying pointer could have been reassigned at some undetermined point in the code and then cast to a completely different type.

qa/common/util.py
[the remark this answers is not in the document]
fpetrini15: This is a simple enough add and I think it adds good value. Thanks for the suggestion! It'll be in the next push.

src/shared_memory_manager.h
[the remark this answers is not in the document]
fpetrini15: Changed structure according to Guan's comments. Resolving as our discussion no longer applies.

src/shared_memory_manager.cc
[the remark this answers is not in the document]
fpetrini15: Changed structure according to Guan's comments. We now pass around a pointer to ShmFile so our discussion no longer applies.

src/shared_memory_manager.cc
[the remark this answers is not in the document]
fpetrini15: Changed structure according to Guan's comments. We no longer make use of `static_cast` so our discussion no longer applies.

src/shared_memory_manager.h
fpetrini15: Discussed offline. Decision was to change the structure of ShmFile, but to continue using smart pointers.

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -29,124 +29,143 @@
...
+        Path(os.getcwd()).parents[0] / "python_models/add_sub/config.pbtxt"
+    )
 
-    def test_valid_create_set_register(self):
-        # Create a valid system shared memory region, fill data in it and register
-        if _protocol == "http":
-            triton_client = httpclient.InferenceServerClient(_url, verbose=True)
+    # Custom setup method to allow passing of parameters
+    def _setUp(self, protocol, log_file_path):
+        self._tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR", "localhost")
```
rmccorm4, line 57: Does this need to be configurable in practice? Do we expect to use shared memory for anything other than a co-located server on localhost?

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -29,124 +29,143 @@
...
-        shm.set_shared_memory_region(
-            shm_op0_handle, [np.array([1, 2], dtype=np.float32)]
+            self._url = self._tritonserver_ipaddr + ":8001"
+            self._triton_client = grpcclient.InferenceServerClient(
+                self._url, verbose=True
+            )
+        self._build_model_repo()
+        self._build_server_args()
+        self._shared_memory_test_server_log = open(log_file_path, "w")
+        self._server_process = util.run_server(
```
rmccorm4, line 79: How does `util.run_server` interact with `test.sh` also starting server? Is there conflict or issue there?

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -29,124 +29,143 @@
...
-        shm_op0_handle = shm.create_shared_memory_region("dummy_data", "/dummy_data", 8)
-        triton_client.unregister_system_shared_memory("dummy_data")
-        shm_status = triton_client.get_system_shared_memory_status()
-        if _protocol == "http":
-            self.assertTrue(len(shm_status) == 0)
+    def _build_server_args(self):
+        if self._test_windows:
+            backend_dir = "C:\\opt\\tritonserver\\backends"
+            model_dir = "C:\\opt\\tritonserver\\qa\\L0_shared_memory\\models"
+            self._server_executable = "C:\\opt\\tritonserver\\bin\\tritonserver.exe"
```
rmccorm4, line 97: Probably more of a random note or follow-up, but I was under the impression something like `Pathlib.Path("/opt/tritonserver/backends")` would translate to `"C:\\opt\\tritonserver\\backends"` for free when run on Windows. If so you could probably condense the cases to work for both.

Did you see otherwise?

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -29,124 +29,143 @@
...
-            shm_op0_handle = shm.create_shared_memory_region(
-                "dummy_data", "/dummy_data", -1
-            )
-            shm.destroy_shared_memory_region(shm_op0_handle)
-        except Exception as ex:
-            self.assertTrue(str(ex) == "unable to initialize the size")
+    # Constant members
+    shared_memory_test_client_log = Path(os.getcwd()) / "client.log"
+    model_dir_path = Path(os.getcwd()) / "models"
+    model_source_path = Path(os.getcwd()).parents[0] / "python_models/add_sub/model.py"
```
rmccorm4, line 50: Future follow-up, but might be nice to have some kind of `utils.relative_path([path, to, thing])`.

ex maybe something like this:
```python
model_dir_path = utils.relative_path("models")
model_source_path = utils.relative_path("..", "python_models", "add_sub", "model.py")
```

qa/L0_shared_memory/shared_memory_test.py
fpetrini15: TBD: Currently on the Windows testing side of things, it's passed in as a variable and is different from "localhost". Still trying to get a CI pipeline up to see the new behavior for this test in particular. Will remove if no issue.

qa/L0_shared_memory/shared_memory_test.py
fpetrini15: I don't believe they should overlap. For this test my ultimate goal is to remove `test.sh` entirely.

qa/L0_shared_memory/shared_memory_test.py
fpetrini15: No, I believe you are right. ATM they are set to my local path and were hardcoded for convenience. They need to be modified and will once I determine the CI environment.

qa/L0_shared_memory/shared_memory_test.py
rmccorm4: Isn't this also getting run in the linux case that runs `test.sh`? or is there changes on the gitlab-side to not run `test.sh` at all?

fpetrini15: There are changes such that `test.sh` will not run at all.

src/shared_memory_manager.h
```diff
@@ -130,7 +106,72 @@ class SharedMemoryManager {
...
+  /// block.
+  /// \return a TRITONSERVER_Error indicating success or failure.
+  TRITONSERVER_Error* GetCUDAHandle(
+      const std::string& name, cudaIpcMemHandle_t** cuda_mem_handle);
+#endif  // TRITON_ENABLE_GPU
+
  private:
+  /// A struct that records the platform-dependent handle to a shared memory
+  /// file.
+  struct ShmFile {
```
nv-kmcgill53, line 136: Much better!

src/shared_memory_manager.cc
```diff
@@ -243,32 +253,42 @@ SharedMemoryManager::RegisterSystemSharedMemory(
 
   // register
   void* mapped_addr;
-  int shm_fd = -1;
+  ShmFile* shm_file = nullptr;
```
nv-kmcgill53, line 256: Guan was right :)

nv-kmcgill53 (dismissed):
LGTM. Great work on this!

src/shared_memory_manager.cc
```diff
@@ -176,55 +178,63 @@ CloseSharedMemoryRegion(int shm_fd)
...
   return nullptr;
 }
-#endif  // TRITON_ENABLE_GPU
-
-}  // namespace
 
 SharedMemoryManager::~SharedMemoryManager()
 {
   UnregisterAll(TRITONSERVER_MEMORY_CPU);
+#ifndef _WIN32
```
GuanLuo, line 235: nit: I don't think you need this ifdef since no CUDA shared memory will be registered on Windows in the first place.

src/shared_memory_manager.cc
GuanLuo: nit: wrap in smart pointer and move assign to SharedMemoryInfo after successfully initialized. In this way we can move clean-up logic into its destructor to handle early return due to unexpected error.

GuanLuo (commented):
Left some comments, can be addressed in the future PR that adds clean up logic

src/shared_memory_manager.cc
fpetrini15: I ended up wrapping ShmFile in a shared pointer because a unique pointer seemed incompatible with what we were trying to achieve with the lookup. This may change depending on how we move forward and fix the current issue regarding key reuse.

src/shared_memory_manager.cc
fpetrini15: Placed due to an abundance of caution. Removed without testing error.

Octoslav:
@fpetrini15 
Hello! Any updates?

PR #7050, opened 2024-03-27
Add trace mode and trace config entries in trace settings API

indrajit96:
PR for
https://jirasw.nvidia.com/browse/DLIS-4960

/trace/setting API updated to return
1. trace_mode
2. trace_config

Old output:
{"trace_level":["TIMESTAMPS"],"trace_rate":"1","trace_count":"-1","log_frequency":"0","trace_file":""}

New Output:
{"trace_level":["TIMESTAMPS"],"trace_rate":"1","trace_count":"-1","log_frequency":"0","trace_file":"","trace_mode":"OPENTELEMETRY","url":"localhost:4318/v1/traces","bsp_max_queue_size":"1","bsp_max_export_batch_size":"512","bsp_schedule_delay":"5000"}

Tested for both modes TRITON and OPENTELEMETRY.

### DIFF
--- a/qa/L0_trace/test.sh
+++ b/qa/L0_trace/test.sh
@@ -188,6 +188,9 @@ fi
 if [ `grep -c "\"trace_file\":\"trace_off_to_min.log\"" ./curl.out` != "1" ]; then
     RET=1
 fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
 
 send_inference_requests "client_min.log" 10
 
@@ -244,6 +247,9 @@ fi
 if [ `grep -c "\"trace_file\":\"global_trace.log\"" ./curl.out` != "1" ]; then
     RET=1
 fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
 
 # Use a different name
 update_trace_setting "simple" '{"trace_file":"simple_trace.log","log_frequency":"2"}'
@@ -263,6 +269,9 @@ if [ `grep -c "\"log_frequency\":\"2\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 if [ `grep -c "\"trace_file\":\"simple_trace.log\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 
@@ -345,6 +354,9 @@ fi
 if [ `grep -c "\"trace_file\":\"update_trace.log\"" ./curl.out` != "1" ]; then
     RET=1
 fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
 
 # Send requests to simple where trace is explicitly disabled
 send_inference_requests "client_update.log" 10
@@ -370,6 +382,9 @@ if [ `grep -c "\"log_frequency\":\"0\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 if [ `grep -c "\"trace_file\":\"global_trace.log\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 
@@ -440,6 +455,9 @@ fi
 if [ `grep -c "\"trace_file\":\"global_count.log\"" ./curl.out` != "1" ]; then
     RET=1
 fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
 
 # Set trace count
 update_global_trace_setting '{"trace_count":"5"}'
@@ -459,6 +477,9 @@ if [ `grep -c "\"log_frequency\":\"0\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 if [ `grep -c "\"trace_file\":\"global_count.log\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 
@@ -482,6 +503,9 @@ if [ `grep -c "\"log_frequency\":\"0\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 if [ `grep -c "\"trace_file\":\"global_count.log\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 
@@ -598,6 +622,9 @@ if [ `grep -c "\"log_frequency\":\"0\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 if [ `grep -c "\"trace_file\":\"bls_trace.log\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
     RET=1
 fi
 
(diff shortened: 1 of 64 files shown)


rmccorm4:
Please update PR title to be more descriptive for the changes :pray:

indrajit96:
> Please update PR title to be more descriptive for the changes 🙏
Fixed thanks :)
github automatically took my branch name as PR title

src/grpc/grpc_server.cc
```diff
@@ -1414,7 +1417,31 @@ CommonHandler::RegisterTrace()
...
         std::to_string(log_frequency));
     (*response->mutable_settings())["trace_file"].add_value(filepath);
-
+    (*response->mutable_settings())["trace_mode"].add_value(
+        trace_manager_->InferenceTraceModeString(trace_mode));
+    {
+      auto mode_key = std::to_string(trace_mode);
+      auto trace_options_it = config_map.find(mode_key);
+      if (trace_options_it != config_map.end()) {
+        for (const auto& element : trace_options_it->second) {
```
rmccorm4, line 1426: You can use probably "structured bindings" here with something like
```cpp
for (const auto& [key, value] : trace_options_it->second_) {
  ...
}
```

src/grpc/grpc_server.cc
```diff
@@ -1414,7 +1417,31 @@ CommonHandler::RegisterTrace()
...
+            continue;
+          }
+          std::string valueAsString;
+          if (std::holds_alternative<std::string>(element.second)) {
+            valueAsString = std::get<std::string>(element.second);
+          } else if (std::holds_alternative<int>(element.second)) {
+            valueAsString = std::to_string(std::get<int>(element.second));
+          } else if (std::holds_alternative<uint32_t>(element.second)) {
+            valueAsString = std::to_string(std::get<uint32_t>(element.second));
+          }
```
rmccorm4, line 1437: Can these `if`/`else` statements be condensed to a single `std::to_string(element.second)` for all cases? or do we need the intermediate `std::get` conversions for each case?
```cpp
          const auto valueAsString = std::to_string(element.second);
          (*response->mutable_settings())[element.first].add_value(
              valueAsString);
```

src/grpc/grpc_server.cc
```diff
@@ -1414,7 +1417,31 @@ CommonHandler::RegisterTrace()
...
+          (*response->mutable_settings())[element.first].add_value(
+              valueAsString);
+        }
+      } else {
+        LOG_VERBOSE(1) << "Trace Config Empty";
+      }
```
rmccorm4, line 1443: This log can probably be removed unless you think it adds significant value

src/http_server.cc
```diff
@@ -1996,7 +2000,31 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
...
+        continue;
+      }
+      std::string valueAsString;
+      if (std::holds_alternative<std::string>(element.second)) {
+        valueAsString = std::get<std::string>(element.second);
+      } else if (std::holds_alternative<int>(element.second)) {
+        valueAsString = std::to_string(std::get<int>(element.second));
+      } else if (std::holds_alternative<uint32_t>(element.second)) {
+        valueAsString = std::to_string(std::get<uint32_t>(element.second));
+      }
```
rmccorm4, line 2021: Can this logic be shared as a single helper used by both http/grpc frontends?

src/grpc/grpc_server.cc
rmccorm4: Thanks for looking into this, agreed on readabiity, I'm good with how it currently is.

src/http_server.cc
```diff
@@ -1996,7 +2000,29 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
       trace_response.AddString("log_frequency", std::to_string(log_frequency)));
   RETURN_AND_RESPOND_IF_ERR(
       req, trace_response.AddString("trace_file", filepath));
-
+  RETURN_AND_RESPOND_IF_ERR(
+      req,
+      trace_response.AddString(
+          "trace_mode", trace_manager_->InferenceTraceModeString(trace_mode)));
+  auto mode_key = std::to_string(trace_mode);
+  auto trace_options_it = config_map.find(mode_key);
```
rmccorm4, line 2008: Food for thought in the future with Olga, but I found this code misleading to read.

Upon reading it, I naively thought `std::to_tsring(trace_mode)` would be something like `"TRITON"` or `"OPENTELEMETRY"`, but it's actually "0" or "1" based on order in the enum. `InferenceTraceModeString` is the one that actually converts to a string. TLDR; I found it confusing that the keys in this map are "0" and "1" rather than the actual string versions of the modes.

No action here, just mentioning something I found misleading to think about down the line.

rmccorm4 (dismissed):
Nice changes! :rocket:

src/grpc/grpc_server.cc
```diff
@@ -1391,7 +1393,8 @@ CommonHandler::RegisterTrace()
     // Get current trace setting, this is needed even if the setting
     // has been updated above as some values may not be provided in the request.
     trace_manager_->GetTraceSetting(
-        request.model_name(), &level, &rate, &count, &log_frequency, &filepath);
+        request.model_name(), &level, &rate, &count, &log_frequency, &filepath,
+        &trace_mode, &config_map);
```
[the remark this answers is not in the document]
indrajit96: Yes I will take this up in the Update PR for
https://jirasw.nvidia.com/browse/DLIS-4795

rmccorm4:
Random note @indrajit96 but I also see the message in [this block](https://github.com/triton-inference-server/server/blob/2eee9ed18e10ba9c3a925123667b2ecf93d9bfa0/src/command_line_parser.cc#L2217-L2218) (not from your changes) is wrong. The warning should mention deprecrated `--trace-log-frequency` and not `--trace-file`. 

Feel free to add to this PR, or submit a separate small message-fix PR

src/tracer.h
```diff
@@ -64,6 +64,7 @@ namespace triton { namespace server {
 
 using TraceConfig = std::vector<
     std::pair<std::string, std::variant<std::string, int, uint32_t>>>;
+// Key is trace mode,
```
oandreeva-nv, line 67: Not only trace mode, IIRC, for global settings the key is an empty string

src/http_server.cc
```diff
@@ -4570,6 +4596,7 @@ HTTPAPIServer::Handle(evhtp_request_t* req)
     } else if (kind == "") {
       // model metadata
       HandleModelMetadata(req, model_name, version);
+
```
oandreeva-nv, line 4599: Do we need this line?

oandreeva-nv:
As a follow-up, I think it is nice to send to the user only relevant settings, i.e. if mode is `opentelemetry`, then `trace file` and `log frequency` are irrelevant.

Next suggestion, can we lowercase mode in the response, please? When users set up mode, it is in lower case, so it would be nice to send it back in lower case as well

indrajit96:
> As a follow-up, I think it is nice to send to the user only relevant settings, i.e. if mode is `opentelemetry`, then `trace file` and `log frequency` are irrelevant.
> 
> Next suggestion, can we lowercase mode in the response, please? When users set up mode, it is in lower case, so it would be nice to send it back in lower case as well

Resolved

src/command_line_parser.cc
```diff
@@ -2214,7 +2214,7 @@ TritonParser::SetTritonTraceArgs(
         lparams.trace_filepath_ = value;
       } else if (setting == "log-frequency") {
         if (trace_log_frequency_present) {
-          std::cerr << "Warning: Overriding deprecated '--trace-file' "
+          std::cerr << "Warning: Overriding deprecated '--trace-log-frequency' "
                        "in favor of provided file in --trace-config!"
```
rmccorm4, line 2218:
```suggestion
                       "in favor of provided log-frequency in --trace-config!"
```

src/http_server.cc
```diff
@@ -1991,12 +1995,40 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
...
       req, trace_response.AddString("trace_count", std::to_string(count)));
+  if (trace_mode == TRACE_MODE_TRITON) {
+    RETURN_AND_RESPOND_IF_ERR(
+        req, trace_response.AddString(
+                 "log_frequency", std::to_string(log_frequency)));
+    RETURN_AND_RESPOND_IF_ERR(
+        req, trace_response.AddString("trace_file", filepath));
+  }
+  std::string trace_mode_lower =
+      trace_manager_->InferenceTraceModeString(trace_mode);
```
rmccorm4, line 2006: @oandreeva-nv @indrajit96 can we just change `InferenceTraceModeString` to return the lowercase variants instead of transforming it?

qa/L0_trace/test.sh
```diff
@@ -188,6 +188,9 @@ fi
 if [ `grep -c "\"trace_file\":\"trace_off_to_min.log\"" ./curl.out` != "1" ]; then
     RET=1
 fi
+if [ `grep -c "\"trace_mode\":\"TRITON\"" ./curl.out` != "1" ]; then
```
rmccorm4, line 191: note these tests will be expecting "triton" instead "opentelemetry" after your lowercase changes

qa/L0_trace/test.sh
```diff
@@ -806,6 +833,18 @@ fi
...
+fi
+if [ `grep -c "\"bsp_max_export_batch_size\":\"512\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"bsp_schedule_delay\":\"5000\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
+if [ `grep -c "\"bsp_max_queue_size\":\"2048\"" ./curl.out` != "1" ]; then
+    RET=1
+fi
```
oandreeva-nv, line 847: We might also check that settings do not contain `trace file` and `log frequency`. Same for trace mode triton : no OpenTelemetry - related settings

oandreeva-nv (approved):

PR #7074, opened 2024-04-05
Tests for Top Level Request Caching for Ensemble Models

lkomali:
Related PR: https://github.com/triton-inference-server/core/pull/338

Added 4 new tests in L0_response_cache to test top level request caching for ensemble models
Test 1: When cache and decoupled enabled in ensemble model config: Error
Test 2: When cache enabled in ensemble model config and decoupled enabled in composing model: Error
Test 3: When cache enabled only in ensemble model: If cache hit = 0 or cache hit count doesn't exist - Fail 
Expected: Non-zero cache hit count
Test 4: When cache enabled in all models: If cache hit = 0 or cache hit > inference count - Fail
Expected: Non zero cache hit count and cache hit count < successful inference count.

No modifications in L0_perf_analyzer_report test. Removed wait $SERVER_PID as is not returning/exiting with 0 because Triton server is taking a long time to shut down.


qa/L0_perf_analyzer_report/test.sh
```diff
@@ -73,13 +73,7 @@ function check_cache_output {
...
     num_cache_hit_lines=$(cat ${CLIENT_LOG} |  grep -i "${CACHE_STRING}" | wc -l)
     num_cache_hit_zero_lines=$(cat ${CLIENT_LOG} |  grep -i "${ERROR_STRING}" | wc -l)
-    # Top-level ensemble model requests do not currently support caching and
-    # will always report a cache hit count of zero if any composing model
-    # has caching enabled. So we check that at least one model reports
-    # non-zero cache hits for now.
-    # TODO: When ensemble models support cache hits, this should just fail
-    #       for any occurrence of ERROR_STRING
-    if [ ${num_cache_hit_lines} -eq ${num_cache_hit_zero_lines} ]; then
+    if [ ${num_cache_hit_zero_lines} -eq ${num_cache_hit_lines} ]; then
```
rmccorm4, line 76: Based on the previous comment in the code, I think this test case was supposed to fail if **any** models had a cache hit of zero count, rather than all, after adding this feature. However, upon second thought, I suppose if ensemble model has caching enabled, then it will intercept the cache hits before the composing models get any. Just commenting to double check your opinion on the logic we should test here.

qa/L0_perf_analyzer_report/test.sh
```diff
@@ -166,7 +160,7 @@ set -e
 
 # Cleanup
 kill $SERVER_PID
-wait $SERVER_PID
```
rmccorm4, line 169: Why is this removed?

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+import re
+import sys
+
+sys.path.append("../common")
+sys.path.append("../clients")
+
+import logging
+import unittest
+
+import infer_util as iu
```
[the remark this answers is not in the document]
rmccorm4: address all the codeQL warnings

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+        results = triton_client.infer(
+            model_name=self.ensemble_model_name, inputs=self.input_tensors
+        )
+        stats = triton_client.get_inference_statistics(
+            model_name=self.ensemble_model_name, as_json=True
+        )
+        return stats
+
+    def _run_ensemble(self):
+        stats = self._get_infer_result()
```
rmccorm4, line 71: rather than doing two inferences in this function , discarding the first stats, and checking the second stats in each test - we should probably do something closer to this for each test:
1. check stats are all 0
2. do inference
3. check cache miss is 1, and cache hit is 0
4. do inference
5. check cache miss is 1, check cache hit is 1

What do you think?

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+        stats = self._get_infer_result()
+        return stats["model_stats"][1]["inference_stats"]
+
+    def _update_config(self, config_file_path, config_pattern, config_to_add):
+        with open(config_file_path, "r") as f:
+            config_data = f.read()
+            if config_pattern not in config_data:
+                with open(config_file_path, "a") as f:
```
rmccorm4, line 79: We should probably overwrite and not append, as it may make the test not [idempotent](https://en.wikipedia.org/wiki/Idempotence) (runnable multiple times in a row expecting same results) if appending.

lkomali: Makes sense, I will change it.

qa/L0_perf_analyzer_report/test.sh
rmccorm4: I see you mentioned this:
> Removed wait $SERVER_PID as is not returning/exiting with 0 because Triton server is taking a long time to shut down.

Can you elaborate on "long time" here? If it's slow, maybe it's expected. If it's hanging, there may be a bug.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ensemble_model_name = "simple_graphdef_float32_float32_float32"
+        self.composing_model_name = "graphdef_float32_float32_float32"
+        self.model_directory = os.path.join(os.getcwd(), "models", "ensemble_models")
+        input0_data = np.ones((1, 16), dtype=np.float32)
+        input1_data = np.ones((1, 16), dtype=np.float32)
+        self.input_tensors = [
+            grpcclient.InferInput("INPUT0", [1, 16], "FP32"),
+            grpcclient.InferInput("INPUT1", [1, 16], "FP32"),
```
rmccorm4, line 55: Use things like `input0_data`.shape and `np_to_triton_dtype(input0_data.dtype)` where applicable so there are less code changes required if we ever need to update shapes/types.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Ideally, we should do this. An issue that I encountered was with trying to kill Triton server which. When I add the lines
kill $SERVER_ID
wait $SERVER_ID
The wait function is not returning or maybe returning with an error code, because after wait command, the other lines are not executing.
Without killing the server and re-launching it, the stats won't be empty.
Let me know if my understanding is wrong.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Will fix this

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+            folder
+            for folder in contents
+            if os.path.isdir(os.path.join(self.model_directory, folder))
+        ]
+        for model_dir in folders:
+            config_file_path = os.path.join(
+                self.model_directory, str(model_dir), "config.pbtxt"
+            )
+            config_to_remove = r"response_cache\s*\{[^}]*\}\n*"
+            self._remove_extra_config(config_file_path, config_to_remove)
```
rmccorm4, line 177: Add leading `_` to helper functions that aren't starting with `test_` for consistency.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+            self._update_config(config_file_path, config_pattern, config_to_add)
+
+    def _remove_extra_config(self, config_file_path, config_to_remove):
+        with open(config_file_path, "r") as f:
+            config_data = f.read()
+        updated_config_data = re.sub(config_to_remove, "", config_data)
+        with open(config_file_path, "w") as f:
+            f.write(updated_config_data)
+
+    def test_ensemble_top_level_cache(self):
```
rmccorm4, line 104: Do you have any tests that validate the output results are actually correct? We should check somewhere that the cached outputs are correct.

qa/L0_perf_analyzer_report/test.sh
lkomali: I made changes in Perf analyzer, which will return 0 when the composing model is not executed in case of a cache hit. Keeping that in mind, I left the condition as is, as there will be cache hit: 0 in the stats but from the composing model.
Reason for the change: There's a check in Perf analyzer expecting stats for all models.
Now that I think of it, I shouldn't return 0 for composing models, when they are not executed.
I will think about this again and the best way to fix.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,182 @@
...
+        input0_data = np.ones((1, 16), dtype=np.float32)
+        input1_data = np.ones((1, 16), dtype=np.float32)
+        self.input_tensors = [
+            grpcclient.InferInput("INPUT0", [1, 16], "FP32"),
+            grpcclient.InferInput("INPUT1", [1, 16], "FP32"),
+        ]
+        self.input_tensors[0].set_data_from_numpy(input0_data)
+        self.input_tensors[1].set_data_from_numpy(input1_data)
+
+    def _get_infer_result(self):
```
rmccorm4, line 60: This function name is misleading if we are returning stats instead of the actual inference outputs/response.

qa/L0_response_cache/ensemble_cache_test.py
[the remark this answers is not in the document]
lkomali: I will fix the codeQL warnings.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: I will change the name

qa/L0_response_cache/ensemble_cache_test.py
lkomali: I will add a test.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Will fix this.

qa/L0_perf_analyzer_report/test.sh
rmccorm4: It could be reasonable to return 0 for composing models if they have caching enabled but don't get executed. 

But another way to think about it from testing perspective is to have at least two test cases for ensemble caching:
1. Ensemble caching enabled and composing models caching enabled -- expect **exactly 1** non-zero cache hit line (for the ensemble), and N-1 zero cache hit lines (composing models that didn't get executed but had caching enabled)
2. Ensemble caching disabled, but composing models caching enabled -- expect exactly N or N-1 non-zero cache hit lines in the output (depending on whether we output a line for the top-level ensemble or not - would need to see some example outputs for this)

qa/L0_response_cache/ensemble_cache_test.py
rmccorm4: Are these comments meant for the other thread about removing the `wait $SERVER_PID`? I don't see the total relevance of the `wait` function for this thread. If you wanted to avoid starting/stopping server between tests there are other things you can do too. Rather than assert they are all zero, you can assert they are all some expected baseline value (based on test knowledge or something - but this may be fragile if it depends on test order). Or you can load the models under new names, or just unload/load the same model, etc. - you can do things to clear the stats if it helps improve the test sanity.

qa/L0_perf_analyzer_report/test.sh
lkomali: Makes sense. Will make the changes.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Ok, will fix it.

qa/L0_response_cache/test.sh
```diff
@@ -197,26 +197,36 @@ function check_server_failure_decoupled_model {
   fi
 }
 
-function test_ensemble_model_with_cache_util {
+function run_server_ensemble_model {
   SERVER_ARGS="--model-repository="${3}" --cache-config local,size=1048576000"
-  if [ "${SERVER_PID}" == "0" ]; then
-    run_server
+  run_server
```
rmccorm4, line 202: nit: Add named args to start of function for readability, something like this:
```bash
function run_server_ensemble_model {
  unit_test="${1}"
  error_msg="${2}"
  model_dir="${3}"
  ...
}

File: qa/L0_response_cache/test.sh [hunk 11]
```diff
@@ -140,52 +170,90 @@ function check_server_expected_failure {
...
-if [ "$SERVER_PID" != "0" ]; then
-    echo -e "\n***\n*** Failed: $SERVER started successfully when it was expected to fail\n***"
-    cat $SERVER_LOG
+function check_server_failure_decoupled_model {
+  MODEL_REPOSITORY="${1}"
+  MODEL_CONTROL_MODE="${2}"
+  MODEL_LOAD_TYPE="${3}"
+  EXTRA_ARGS="--model-control-mode=${MODEL_CONTROL_MODE} --load-model=${MODEL_LOAD_TYPE}"
+  SERVER_ARGS="--model-repository=${MODEL_REPOSITORY} --cache-config local,size=1048576000 ${EXTRA_ARGS}"
+  source ../common/util.sh
```
rmccorm4, line 179: Why is this sourced as part of the function? This should be moved higher up and outside of the function.

rmccorm4: Actually looks like it's already included higher up, so this can just be removed.

qa/L0_perf_analyzer_report/test.sh
lkomali: After some thinking, I think this test should be added to L0_response_cache, since this check is related to the behavior of cache after adding top-level response caching feature, but not how Perf_analyzer handles the change.
Already, the first case is handled in the L0_response_cache with the test **"test_all_models_with_cache_enabled"**
where I enable cache in all models and check if cache hit count is not greater than inference count.
I can quickly add the second scenario too.
Let me know your thoughts on this.

rmccorm4: Sure, feel free to augment the L0_response_cache for more of this checking. Let's leave this check as-is in L0_perf_analyzer_report.

qa/L0_response_cache/test.sh
lkomali: I removed the SERVER_ARGS defined outside the function as it is not used. This should be inside the function as ${MODEL_REPOSITORY} changes for decoupled_cache test case and ensemble cache testcase. Other testcases use different SERVER_ARGS.

qa/L0_response_cache/test.sh
```diff
@@ -31,11 +31,42 @@ TEST_LOG="./response_cache_test.log"
...
+    REPO_VERSION=${REPO_VERSION}_${TEST_REPO_ARCH}
+fi
 # Only localhost supported in this test for now, but in future could make
 # use of a persistent remote redis server, or similarly use --replicaof arg.
 export TRITON_REDIS_HOST="localhost"
 export TRITON_REDIS_PORT="6379"
 REDIS_LOG="./redis-server.unit_tests.log"
+ENSEMBLE_CACHE_TEST_PY="./ensemble_cache_test.py"
+SERVER=/opt/tritonserver/bin/tritonserver
+# SERVER_ARGS="--model-repository=`pwd`/models --response-cache-byte-size=8192"
```
oandreeva-nv, line 53: Do we need this line?

qa/L0_response_cache/test.sh
```diff
@@ -140,52 +171,77 @@ function check_server_expected_failure {
...
+  if [ "${SERVER_PID}" != "0" ]; then
+    echo -e "\n***\n*** Failed: ${SERVER} started successfully when it was expected to fail\n***"
+    cat ${SERVER_LOG}
     RET=1
 
-    kill $SERVER_PID
-    wait $SERVER_PID
-else
+    kill ${SERVER_PID}
+    wait ${SERVER_PID}
```
oandreeva-nv, line 189: Why do we need curly brackets?

qa/L0_response_cache/test.sh
```diff
@@ -140,52 +171,77 @@ function check_server_expected_failure {
...
+# Check that server fails to start for a "decoupled" model with cache enabled
+check_server_failure_decoupled_model ${MODEL_DIR}  "decoupled_cache"
+
 
 # Test with model expected to load successfully
 EXTRA_ARGS="--model-control-mode=explicit --load-model=identity_cache"
 
-# Test old cache config method
-# --response-cache-byte-size must be non-zero to test models with cache enabled
+#  Test old cache config method
```
oandreeva-nv, line 233: Why do we need an extra space here and in the following comments?

qa/L0_response_cache/test.sh
lkomali: we don't need. probably missed in removing it.

qa/L0_response_cache/test.sh
```diff
@@ -272,15 +328,71 @@ run_server
...
+CACHE_FUNCTION="EnsembleCacheTest.setup_cache_ensemble_model"
+python ${ENSEMBLE_CACHE_TEST_PY} ${CACHE_FUNCTION} >> ${CLIENT_LOG} 2>&1
+DECOUPLED_FUNCTION="EnsembleCacheTest.setup_decoupled_ensemble_model"
+python ${ENSEMBLE_CACHE_TEST_PY} ${DECOUPLED_FUNCTION} >> ${CLIENT_LOG} 2>&1
+check_server_failure_decoupled_model ${ENSEMBLE_MODEL_DIR} ${ENSEMBLE_MODEL}
+RESET_CONFIG_FUNCTION="EnsembleCacheTest.reset_config"
+python ${ENSEMBLE_CACHE_TEST_PY} ${RESET_CONFIG_FUNCTION} >> ${CLIENT_LOG} 2>&1
+
+
+# Test Ensemble Model With Cache Enabled and Decoupled Mode in Composing Model
```
oandreeva-nv, line 346: ditto here and similar following comments to the tests: I'd prefer capitalization of only the first letter

oandreeva-nv: and a space after `#`

qa/L0_response_cache/test.sh
```diff
@@ -272,15 +328,71 @@ run_server
...
+python ${ENSEMBLE_CACHE_TEST_PY} ${RESET_CONFIG_FUNCTION} >> ${CLIENT_LOG} 2>&1
+
+#Test Ensemble Model with Top Level Caching Enabled
+CACHE_FUNCTION="EnsembleCacheTest.setup_cache_ensemble_model"
+python ${ENSEMBLE_CACHE_TEST_PY} ${CACHE_FUNCTION} >> ${CLIENT_LOG} 2>&1
+TEST_NAME="EnsembleCacheTest.test_ensemble_top_level_cache"
+ERROR_MESSAGE="\n***\n*** Failed: Expected Top Level Request Caching\n***"
+CACHE_SIZE=10480
+run_server_ensemble_model "${TEST_NAME}" "${ERROR_MESSAGE}" "${CACHE_SIZE}"
+
```
oandreeva-nv, line 362:
```suggestion
```

qa/L0_response_cache/test.sh
```diff
@@ -272,15 +328,71 @@ run_server
...
+#Test Ensemble Model with cache enabled in all models
+CACHE_FUNCTION="EnsembleCacheTest.setup_cache_all_models"
+python ${ENSEMBLE_CACHE_TEST_PY} ${CACHE_FUNCTION} >> ${CLIENT_LOG} 2>&1
+TEST_NAME="EnsembleCacheTest.test_all_models_with_cache_enabled"
+ERROR_MESSAGE="\n***\n*** Failed: Expected cache to return Top-Level request's response\n***"
+CACHE_SIZE=10480
+run_server_ensemble_model "${TEST_NAME}" "${ERROR_MESSAGE}" "${CACHE_SIZE}"
+
+
```
oandreeva-nv, line 372:
```suggestion
```

qa/L0_response_cache/test.sh
```diff
@@ -272,15 +328,71 @@ run_server
...
+#Test Composing model cache enabled
+CACHE_FUNCTION="EnsembleCacheTest.setup_cache_composing_model"
+python ${ENSEMBLE_CACHE_TEST_PY} ${CACHE_FUNCTION} >> ${CLIENT_LOG} 2>&1
+TEST_NAME="EnsembleCacheTest.test_composing_model_cache_enabled"
+ERROR_MESSAGE="\n***\n*** Failed: Expected only composing model's input/output to be inserted in cache\n***"
+CACHE_SIZE=10480
+run_server_ensemble_model "${TEST_NAME}" "${ERROR_MESSAGE}" "${CACHE_SIZE}"
+
+
```
oandreeva-nv, line 381:
```suggestion
```

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+    def _check_valid_stats(self, model_inference_stats):
+        if "count" not in model_inference_stats["success"]:
+            self.assertTrue(False, "unexpected error while retrieving statistics")
+
+    def _check_single_cache_miss_success_inference(self, model_inference_stats):
+        if (
+            "count" not in model_inference_stats["cache_miss"]
+            and "count" not in model_inference_stats["cache_hit"]
+        ):
+            self.assertTrue(False, "unexpected error with response cache")
```
oandreeva-nv, line 200: I'm not sure I fully understand what is happening here. 

If we want to make sure that both `"count" in model_inference_stats["cache_miss"]` and `"count"  in model_inference_stats["cache_hit"]` are True, then I would suggest:

```
expr = "count" in model_inference_stats["cache_miss"] and "count"  in model_inference_stats["cache_hit"]
assertTrue(expr, msg="unexpected error with response cache")
```
reference: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue 

`self.assertTrue(False, ...)` is a bit confusing, i.e. I know that False is not True, what we're asserting here?

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+        if (
+            "count" in model_inference_stats["cache_hit"]
+            and int(model_inference_stats["cache_hit"]["count"]) > 0
+        ):
+            self.assertTrue(False, "unexpected cache hit")
```
oandreeva-nv, line 205: ditto

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+        composing_model_stats = self.inference_handler.get_inference_statistics(
+            self.composing_model
+        )
+        if (
+            "count" not in ensemble_model_stats["cache_hit"]
+            or int(ensemble_model_stats["cache_hit"]["count"]) == 0
+        ):
+            self.assertTrue(
+                False, "unexpected error in top-level ensemble request caching"
+            )
```
oandreeva-nv, line 280: ditto here and everywhere else

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def test_ensemble_top_level_cache(self):
+        model_inference_stats = self.inference_validator.run_ensemble(
+            self.ensemble_model
+        )
+        if (
+            "count" not in model_inference_stats["cache_hit"]
+            or model_inference_stats["cache_hit"]["count"] != "0"
+        ):
+            self.assertFalse(
```
oandreeva-nv, line 260: `assertFalse(False, ....)` <- not sure what was intended here, but this assert will always pass

qa/L0_response_cache/test.sh
lkomali: It's not required. Followed the style from lines 158,159 which was already present in the file.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+        return outputs
+
+
+class InferenceValidator:
+    def __init__(self):
+        self.inference_handler = InferenceHandler()
+
+    def _check_valid_output(self, output):
+        if output is None:
+            self.assertTrue(False, "unexpected error in inference")
```
oandreeva-nv, line 181: use `assertIsNotNone` : https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNotNone

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def _check_valid_output_inference(self, inference_outputs, cached_outputs):
+        if not np.array_equal(inference_outputs, cached_outputs):
+            self.assertTrue(False, "mismtached outputs")
```
oandreeva-nv, line 185: use `assertIsEqual`

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def _check_zero_stats_baseline(self, model_inference_stats):
+        if "count" in model_inference_stats["success"]:
+            self.assertTrue(False, "unexpected non-zero inference statistics")
```
oandreeva-nv, line 189: `assertIsIn`

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def _check_valid_stats(self, model_inference_stats):
+        if "count" not in model_inference_stats["success"]:
+            self.assertTrue(False, "unexpected error while retrieving statistics")
```
oandreeva-nv, line 193: `assertIsIn`

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def test_composing_model_cache_enabled(self):
+        composing_model_stats = self.inference_validator.run_ensemble(
+            self.composing_model
+        )
+        ensemble_model_stats = self.inference_handler.get_inference_statistics(
+            self.ensemble_model
+        )
+        if "count" not in composing_model_stats["cache_miss"]:
+            self.assertTrue(
```
oandreeva-nv, line 293: `assertIsIn`

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+                False, "unexpected error in caching composing model response"
+            )
+        if "count" in ensemble_model_stats["cache_hit"]:
+            self.assertTrue(False, "unexpected top-level response caching")
```
oandreeva-nv, line 297: `assertIsIn`

qa/L0_response_cache/ensemble_cache_test.py
lkomali: I followed this https://github.com/triton-inference-server/server/blob/main/qa/L0_simple_ensemble/ensemble_test.py
when I first started writing testcases. 
I will make changes as suggested.

oandreeva-nv: No problem, that test was added in 2019, not sure what state python unittests library was in, but now it has a variety of methods, which worth exploring and using, since they provide better messages and are tailored to what you are trying to check

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,319 @@
...
+    def _check_valid_output(self, output):
+        assert output is not None, "Unexpected error in inference"
+
+    def _check_valid_output_inference(self, inference_outputs, cached_outputs):
+        assert np.array_equal(inference_outputs, cached_outputs), "Mismatched outputs"
+
+    def _check_zero_stats_baseline(self, model_inference_stats):
+        assert (
+            "count" not in model_inference_stats["success"]
+        ), "Unexpected error while retrieving statistics"
```
oandreeva-nv, line 187: I'd recommend picking a style and stick to it. Currently some asserts are followed by condition in brackets and some are not.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+        self.assertNotEqual(
+            model_inference_stats["cache_hit"]["count"],
+            "0",
+            "Unexpected error in top-level ensemble response caching",
+        )
+        self.assertLessEqual(
+            int(model_inference_stats["cache_hit"]["count"]),
+            1,
+            "Unexpected multiple cache hits",
```
oandreeva-nv, line 258: could you please clarify why we're validating with `LessEqual` ? I would assume based on `run_ensemble`, it should be `assertEquals`

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,319 @@
...
+            "count",
+            composing_model_stats["cache_hit"],
+            "Unexpected composing model cache hits",
+        )
+        self.reset_config()
+
+    def test_composing_model_cache_enabled(self):
+        composing_model_stats = self.inference_validator.run_ensemble(
+            self.composing_model
+        )
```
oandreeva-nv, line 284: is `composing_model` also an ensemble?

qa/L0_response_cache/ensemble_cache_test.py
oandreeva-nv: I would also recommend adding expected and received values to the error msg, to ease debugging

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Actually, the run_ensemble runs the ensemble model but the self.composing_model parameter is to define which model stats to verify. I'm doing a bunch of stats verification in run_ensemble. In the case of response_cache enable only in composing model, the top_level ensemble model stats will not have any cache statistics. 
I can see that it's cofusing. I will modify the function name to avoid confusion.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: On a second thought, this assertion is not required. I want to test if composing model doesn't have cache hits. I will modify to test this case on composing model stats.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,296 @@
...
+    def _get_model_statistics(self, model):
+        # Get the stats for the requested model
+        model_stats = self.triton_client.get_inference_statistics(
+            model_name=model, as_json=True
+        )
+        return model_stats["model_stats"][1]["inference_stats"]
+
+    def _run_inference_and_validate(self, model):
+        # Utility function to call run_ensemble for inference and validate expected baseline output and stats
+        self.triton_client.load_model(self.ensemble_model)
```
oandreeva-nv, line 118: Should the past argument be `model` and not `self.ensemble_model` ?

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def _get_model_statistics(self, model):
+        # Get the stats for the requested model
+        model_stats = self.triton_client.get_inference_statistics(
+            model_name=model, as_json=True
+        )
+        return model_stats["model_stats"][1]["inference_stats"]
+
+    def _run_inference_and_validate(self, model):
+        # Utility function to call run_ensemble for inference and validate expected baseline output and stats
```
oandreeva-nv, line 117: May I ask to specify the test plan for this sub-routine? Something like this: https://github.com/triton-inference-server/server/blob/20627268b9f7e951e8be7f49f0d28bf76f178961/qa/L0_trace/opentelemetry_unittest.py#L236-L250

This way it would be clearer why we load models in question and what this routine is doing

qa/L0_response_cache/test.sh
oandreeva-nv: Could you please remove curly brackets here and everywhere where they are not needed?

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Sure, will add it.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: The parameter for _run_inference_and_validate should be model because the model can be ensemble model or composing model. The passed model's stats will be validated according to testcase. 
In case of 3rd testcase which has response cache enabled only in composing model, the ensemble model stats will have empty fields for cache related metrics. That's the reason why I'm separately passing model parameter to define which mode's stats to be verified.

oandreeva-nv: My question was mostly for `load_model` argument. It was a slightly confusing and not clear from the start why we do that, so that is why I asked to clarify test plan.

oandreeva-nv (commented):
I think this PR is in a very good state. Nice work on re-factoring since the original commit! Just a couple of small additions and we'll be able to merge it

qa/L0_perf_analyzer_report/test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 2: Remove (c) for multi-year
```suggestion
# Copyright 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

qa/L0_perf_analyzer_report/test.sh
```diff
@@ -166,7 +160,7 @@ set -e
 
 # Cleanup
 kill $SERVER_PID
-wait $SERVER_PID
+
```
rmccorm4, line 163: @lkomali doesn't this `wait $SERVER_PID` need to be added back after you fixed the root cause with inflight request counter?

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,296 @@
...
+    def _remove_config(self, config_file, config_to_remove):
+        # Utility function to remove extra added config from the config files
+        with open(config_file, "r") as f:
+            config_data = f.read()
+        updated_config_data = re.sub(config_to_remove, "", config_data)
+        with open(config_file, "w") as f:
+            f.write(updated_config_data)
+
+    def _reset_config_files(self):
+        # Utillity function to reset all config files to original
```
rmccorm4, line 91:
```suggestion
        # Utility function to reset all config files to original
```

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,296 @@
...
+        output1 = output.as_numpy("OUTPUT1")
+        outputs = [output0, output1]
+        return outputs
+
+    def _get_model_statistics(self, model):
+        # Get the stats for the requested model
+        model_stats = self.triton_client.get_inference_statistics(
+            model_name=model, as_json=True
+        )
+        return model_stats["model_stats"][1]["inference_stats"]
```
rmccorm4, line 114: Add comment on why `[1]` here -> which model? If it's for a specific model, I would add some small assert that `model_stats["model_stats"][1]["name"]` equals the expected model you're checking for

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,296 @@
...
+        """
+        Test caching behavior when response cache enabled only in
+        composing model's config file.
+        Expected result: One cache hit in composing model stats. No cache related metric counts in
+        ensemble model stats.
+        """
+        self._update_config(
+            self.composing_config_file, RESPONSE_CACHE_PATTERN, RESPONSE_CACHE_CONFIG
+        )
+        self._run_inference_and_validate(self.composing_model)
```
rmccorm4, line 236: Should we be running inference on the ensemble model here, and validating that ensemble did inference but has no cache stats, and that composing model does have correct cache stats? Looks like we're doing inference on composing model directly here so not actually testing the ensemble flow.

qa/L0_response_cache/test.sh
rmccorm4: I'm pro curly brackets 😁 maybe we should leave it as is and add a bash linter to avoid ambiguity in style preference?

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,296 @@
...
+            composing_model_stats["cache_hit"],
+            f"No cache hit recorded for ensemble model '{self.composing_model}'. Expected exactly one cache hit.",
+        )
+        self.assertNotIn(
+            "count",
+            ensemble_model_stats["cache_hit"],
+            f"Unexpected number of cache hits recorded for ensemble model '{self.ensemble_model}'. Expected empty cache metrics",
+        )
+
+    def test_ensemble_cache_insertion_failure(self):
```
rmccorm4, line 255: Can you add a comment to this test explaining why the insertinos are expected to fail? Assuming it's due to tensors bigger than small cache size set in bash script.

Can you also `grep` the server log in bash script for the expected insertion failure error message if any?

qa/L0_response_cache/test.sh
```diff
@@ -31,11 +31,56 @@ TEST_LOG="./response_cache_test.log"
...
 export TRITON_REDIS_HOST="localhost"
 export TRITON_REDIS_PORT="6379"
 REDIS_LOG="./redis-server.unit_tests.log"
+ENSEMBLE_CACHE_TEST_PY="./ensemble_cache_test.py"
+SERVER=/opt/tritonserver/bin/tritonserver
+CLIENT_LOG="./client.log"
+TEST_RESULT_FILE='test_results.txt'
+SERVER_LOG=./inference_server.log
+RESET_CONFIG_FUNCTION="_reset_config_files"
+CACHE_SIZE=10840
```
rmccorm4, line 57: Add comment on size choice - why this number? what's the requirements/goal?

rmccorm4 (commented):
Looking really good! Only minor comments

qa/L0_response_cache/test.sh
oandreeva-nv: We need a style to stick to

rmccorm4: My preference stems from past use of [shellcheck](https://github.com/koalaman/shellcheck): https://www.shellcheck.net/wiki/SC2250 - and I extend it to just preferring the brackets everywhere for consistency rather than only in problematic places. But it would be a separate effort to standardize all files to one or the other, they're currently mixed throughout I believe

qa/L0_response_cache/ensemble_cache_test.py
lkomali: The model's have two versions version 1 and version 3. Version 1 stats are at index 0 and version 3 stats are index 1.
Version 3 is loaded. so to access it's stats, index is set to 1.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: We are running inference on ensemble model only. The model parameter is only to verify baseline stats in run_inference_and_validate. For this testcase, ensemble model stats are going to be empty. So I passed model as a parameter to correctly verify corresponding model's stats, for this testcase it's composing model.

qa/L0_response_cache/test.sh
lkomali: I used a random number. Something that worked for me. 
For test_cache_insertion testcase, I used CACHE_SIZE=200 because the data to be inserted is more than 200. Found the value through calculations and testing.

qa/L0_response_cache/ensemble_cache_test.py
lkomali: Added it.

qa/L0_response_cache/test.sh
```diff
@@ -140,42 +185,65 @@ function check_server_expected_failure {
...
+
+# Check that server fails to start for a "decoupled" model with cache enabled
+check_server_failure_decoupled_model ${MODEL_DIR}  "decoupled_cache"
 
 # Test with model expected to load successfully
 EXTRA_ARGS="--model-control-mode=explicit --load-model=identity_cache"
 
 # Test old cache config method
-# --response-cache-byte-size must be non-zero to test models with cache enabled
+#  --response-cache-byte-size must be non-zero to test models with cache enabled
```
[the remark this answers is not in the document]
lkomali: Fixed.

qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -0,0 +1,316 @@
...
+        model_stats has inference stats for version 1 at index 0 and inference stats for version 3 at index 1.
+        """
+        return model_stats["model_stats"][1]["inference_stats"]
+
+    def _run_inference_and_validate(self, model):
+        """
+        Helper function that takes model as a parameter to verify the corresponding model's stats
+        The passed model is composing model for test case `test_ensemble_composing_model_cache_enabled`
+        For other testcases, the top-level ensemble model stats are verified.
+            * loads the simple_graphdef_float32_float32_float32 and graphdef_float32_float32_float32
```
oandreeva-nv, line 128: nit: clarify which of these models is an ensemble and which is not

qa/L0_response_cache/test.sh
rmccorm4: deferring this to later if that's OK with you @oandreeva-nv , looking to get this PR in

oandreeva-nv: I'm ok

rmccorm4 (approved):
Great work Harshini! :rocket: 

Can you add a link on slack to a CI pipeline running with all your latest changes to make sure everything looks good?

PR #7076, opened 2024-04-05
build: Have ability to split the build in CI 

mc-nv:
MOTIVATION: 
Triton Server CI is monolithic, I would like to add ability splitting the CI and build it per backend if needed

CHANGES: 
1. Added flags `--ci-split` and `--cache-from-map`
2. Update logic of file creation in case of flag `--ci-split`

VERIFICATION: 
```
$ ./build.py --enable-all --split-build --dryrun
$ ls build
cmake_build                 cmake_build_backend_ensemble  cmake_build_backend_onnxruntime  cmake_build_backend_pytorch  cmake_build_backend_tensorflow  cmake_build_cache_redis  Dockerfile
cmake_build_agent_checksum  cmake_build_backend_fil       cmake_build_backend_openvino     cmake_build_backend_repeat   cmake_build_backend_tensorrt    cmake_build_collect      Dockerfile.buildbase
cmake_build_backend_dali    cmake_build_backend_identity  cmake_build_backend_python       cmake_build_backend_square   cmake_build_cache_local         docker_build             Dockerfile.cibase
 ```
 
 ```
$ ./build.py --enable-all --dryrun
$ ls build
cmake_build  docker_build  Dockerfile  Dockerfile.buildbase  Dockerfile.cibase
 ```

### CODE REVIEW

File: build.py [hunk 1]
```diff
@@ -2338,6 +2347,19 @@ def enable_all():
         required=False,
         help="Override specified backend CMake argument in the build as <backend>:<name>=<value>. The argument is passed to CMake as -D<name>=<value>. This flag only impacts CMake arguments that are used by build.py. To unconditionally add a CMake argument to the backend build use --extra-backend-cmake-arg.",
     )
+    parser.add_argument(
+        "--ci-split",
```
rmccorm4, line 2351: If it would have any value to end users as well, how about naming it something like `--split-build` or something instead, not specific to CI?

build.py
```diff
@@ -2338,6 +2347,19 @@ def enable_all():
...
+        action="store_true",
+        required=False,
+        help="Requires to split CI build into multiple builds. Will generate cmake build script separatelly for each backend",
```
rmccorm4, line 2354:
```suggestion
        help="Split the intermediate build artifacts into independently build-able targets",
```

build.py
```diff
@@ -2338,6 +2347,19 @@ def enable_all():
...
+    )
+    parser.add_argument(
+        "--cache-from-map",
+        action="append",
+        required=False,
+        help="Requires to split CI build into multiple builds. Will generate cmake build script separatelly for each backend",
```
rmccorm4, line 2360:
```suggestion
        help="A set of docker images to `--cache-from` where applicable to help speedup builds",
```

build.py
mc-nv: it should be work only in CI we shouldn't expose it to the customers.
You have a valid point `ci` vs `build` where second sound more reasonable. 
I have no expectation that users will use it.

build.py
```diff
@@ -2639,9 +2661,12 @@ def enable_all():
                 components,
                 backends,
             )
-
         # Commands to build each backend...
         for be in backends:
+            # Define fucntion to create cmake_script with backend name as suffix
```
rmccorm4, line 2666: Make sure to do `pre-commit install` locally in the repo and possibly `pre-commit run --all-files` so it can locally fix the issues being reported by the :x:

build.py
rmccorm4: PR LGTM assuming CI build is passing using these new flags. Would still prefer to rename this flag to mention `build` instead of `ci`.

build.py
```diff
@@ -2338,6 +2348,23 @@ def enable_all():
         required=False,
         help="Override specified backend CMake argument in the build as <backend>:<name>=<value>. The argument is passed to CMake as -D<name>=<value>. This flag only impacts CMake arguments that are used by build.py. To unconditionally add a CMake argument to the backend build use --extra-backend-cmake-arg.",
     )
+    parser.add_argument(
+        "--split-build",
+        action="store_true",
+        required=False,
+        help="Split the intermediate build artifacts into independently build-able targets",
```
nv-kmcgill53, line 2355:
```suggestion
        help="Split the intermediate build artifacts into independent build-able targets",
```
nit: grammar

build.py
```diff
@@ -2338,6 +2348,23 @@ def enable_all():
...
+        "--cache-from-map",
+        action="append",
+        required=False,
+        help="A set of docker images to `--cache-from` where applicable to help speedup builds",
+        default=[
+            "tritonserver_buildbase",
+            "tritonserver_buildbase_cache0",
+            "tritonserver_buildbase_cache1",
+        ],
+    )
```
nv-kmcgill53, line 2367: I understand that it was called a `map` before, but it's really just a list. What do you think about changing this to something like `--cache-image` or `--cache-images`? We can make use of the multiple arguments in a couple of different ways. We can use the `action="append"` or the `nargs='+'` fields ([SO comment for example](https://stackoverflow.com/questions/15753701/how-can-i-pass-a-list-as-a-command-line-argument-with-argparse)). I might lean towards the `--cache-image` with `action="append"` API just because we have precedent.

nv-kmcgill53 (commented):
Just an API comment, since API's are sticky you try to get them right the first time because the overhead of changing them later is high.

build.py
mc-nv: I think name was used to  keep consistency with original command which is relates to: `docker build --cache-from` 
```
$ docker build --help | grep cache-from
      --cache-from stringArray        External cache sources (e.g., "user/app:cache", "type=local,src=path/to/dir")
```

nv-kmcgill53: How about `--cache-from-list`?

PR #7083, opened 2024-04-08
Validate the memory requested for the infer request is not out of bounds

jbkyang-nvi:
Previously in https://github.com/triton-inference-server/server/pull/6914 we checked out of bounds `offset` for shared memory requests. This PR updates to a more comprehensive test of verifying that the block of memory is in fact in bounds.
Client change: https://github.com/triton-inference-server/client/pull/565


qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -40,6 +40,8 @@
 import tritonclient.utils.shared_memory as shm
 from tritonclient import utils
 
+DEFAULT_SHM_BYTE_SIZE = 64
```
Tabrizian, line 43: nit: can we make `DEFAULT_SHM_BYTE_SIZE` a member of `SharedMemoryTest` class?

```
class SharedMemoryTest:
  def setUp(self):
    self.shared_memory_byte_size = 64
```

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -328,6 +346,7 @@ def test_infer_offset_out_of_bound(self):
         # Shared memory offset outside output region - Throws error
         error_msg = []
         shm_handles = self._configure_sever()
+        print("testing")
```
Tabrizian, line 349: remove print

kthui (commented):
LGTM! apart from things @Tabrizian pointed out

qa/L0_shared_memory/shared_memory_test.py
jbkyang-nvi: Since it's a const, would it be ok to just do:
```
class SharedMemoryTest:
    DEFAULT_SHM_BYTE_SIZE=64
```
?

kthui: I think it is fine as long as the variable does not change across instances of the class. I don't know about the all upper-case variable name, because the example from the Python docs tutorial use all lower-case.
https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables
I will defer that to @Tabrizian .

jbkyang-nvi: I was following: https://peps.python.org/pep-0008/#constants

qa/L0_shared_memory/test.sh
```diff
@@ -62,32 +63,34 @@ for i in \
...
                 RET=1
             fi
         fi
-        set -e
-
+        cat $TMP_CLIENT_LOG >>$CLIENT_LOG
+        rm $TMP_CLIENT_LOG
         kill $SERVER_PID
         wait $SERVER_PID
+        set -e
```
Tabrizian, line 87: question: why do we need to move `set -e`? If it is before kill/wait it can help us detect non-graceful exits.

qa/L0_shared_memory/test.sh
```diff
@@ -62,32 +63,34 @@ for i in \
         fi
 
         export CLIENT_TYPE=$client_type
-        echo "Test: $i, client type: $client_type" >>$CLIENT_LOG
+        TMP_CLIENT_LOG="./tmp_client.log"
+        echo "Test: $i, client type: $client_type" >>$TMP_CLIENT_LOG
 
         set +e
-        python $SHM_TEST SharedMemoryTest.$i >>$CLIENT_LOG 2>&1
+        python $SHM_TEST SharedMemoryTest.$i >>$TMP_CLIENT_LOG 2>&1
```
Tabrizian, line 70: question: why do we need to use a different log file?

jbkyang-nvi: This way we can echo the failing part of the client test that is failing and not _all_ tests that have been run so far.

qa/L0_shared_memory/test.sh
jbkyang-nvi: Yes but at the same time on local runs it will exit before the server is properly killed. I am also a little hesitant on this. Since the tests are orthogonal to each other, I though it might make more sense to run everything and see which tests are failing instead of only failing on the first one

qa/L0_shared_memory/test.sh
Tabrizian: but the `TMP_CLIENT_LOG` contains the results from all of the clients as far as I can tell from your code. We're appending to the same file log.

qa/L0_shared_memory/test.sh
Tabrizian: As long as `set -e` is after `python ...` it would run all the tests. We need to either check for the return code of `wait` or move `set -e` before the `kill & wait` calls.

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -348,6 +370,28 @@ def test_infer_offset_out_of_bound(self):
         self.assertIn("Invalid offset for shared memory region", error_msg[0])
         self._cleanup_server(shm_handles)
 
+    def test_infer_byte_size_out_of_bound(self):
```
Tabrizian, line 373: Could you please test it for CUDA shared memory as well?

qa/L0_shared_memory/test.sh
jbkyang-nvi: we remove the `$TMP_CLIENT_LOG` at L84 after we copy to the `$CLIENT_LOG`

qa/L0_shared_memory/shared_memory_test.py
jbkyang-nvi: We have a ticket for GPU shared memory. Can we add that test in that PR?

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -295,7 +317,7 @@ def test_too_big_shm(self):
     def test_mixed_raw_shm(self):
         # Mix of shared memory and RAW inputs
         error_msg = []
-        shm_handles = self._configure_sever()
+        shm_handles = self._configure_sever(shm_byte_size=256)
```
rmccorm4, line 320: My PR will have some conflicts with this PR, but you should be able to use the `create_byte_size` and `register_byte_size` args in the same way. Let me know if you're good with it: https://github.com/triton-inference-server/server/pull/7093

qa/L0_shared_memory/test.sh
Tabrizian: @jbkyang-nvi friendly ping on this one.

qa/L0_shared_memory/shared_memory_test.py
Tabrizian: Sounds good.

Tabrizian (dismissed):

kthui (dismissed):

qa/L0_shared_memory/shared_memory_test.py
rmccorm4: Francesco's PR will also likely have some merge conflicts: https://github.com/triton-inference-server/server/pull/7048

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -292,7 +328,7 @@ def test_too_big_shm(self):
     def test_mixed_raw_shm(self):
         # Mix of shared memory and RAW inputs
         error_msg = []
-        shm_handles = self._configure_sever(shm_byte_size=256)
+        shm_handles = self._configure_server()
```
rmccorm4: Was this supposed to have a value of 256 passed to it?

jbkyang-nvi: I think it was only needed because I had a bug so I removed it..

PR #7092, opened 2024-04-09
Disable Dynamic Log File

yinggeh:
Disable dynamic log file location, aka `--log_file` API.

### DIFF
--- a/src/grpc/grpc_server.cc
+++ b/src/grpc/grpc_server.cc
@@ -1463,32 +1463,10 @@ CommonHandler::RegisterLogging()
         static std::string setting_name = "log_file";
         auto it = request.settings().find(setting_name);
         if (it != request.settings().end()) {
-          const auto& log_param = it->second;
-          if (log_param.parameter_choice_case() !=
-              inference::LogSettingsRequest_SettingValue::ParameterChoiceCase::
-                  kStringParam) {
-            err = TRITONSERVER_ErrorNew(
-                TRITONSERVER_ERROR_INVALID_ARG,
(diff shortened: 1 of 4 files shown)


src/grpc/grpc_server.cc
```diff
@@ -1463,32 +1463,10 @@ CommonHandler::RegisterLogging()
...
-            // to the global object.
-            err = TRITONSERVER_ServerOptionsSetLogFile(
-                nullptr, log_file_path.c_str());
-            if (err != nullptr) {
-              GOTO_IF_ERR(err, earlyexit);
-            }
-          }
+          err = TRITONSERVER_ErrorNew(
+              TRITONSERVER_ERROR_UNSUPPORTED,
+              "updating log file was deprecated and no longer supported");
```
nnshah1, line 1468: would continue to use the variable `setting_name` for completeness

nnshah1:
```suggestion
              "log file location can not be updated through network protocol");
```

nnshah1 (approved):

src/grpc/grpc_server.cc
nnshah1: little tricky - given that it's the only one probably clear enough from context. 

Was thinking:
```
 setting_name + " can not be updated through newtork protocol"  
```

nnshah1: @Tabrizian - any thoughts one way or the other?

Tabrizian: Error message looks good to me.

PR #7093, opened 2024-04-10
Validate system shared memory region size when registering a region

rmccorm4:
This PR:
- [x] Validate System SHM region size during registration
- [x] Add system shm registration tests

Deferred to separate PRs:
- [ ] Validate CUDA SHM region size
    - need to investigate the right way to query device pointer allocated size from cuIPC handle
- [ ] Validate System SHM file/user permissions
    - need a test to validate the existing code would fail with bad permissions when container is setup with a non-root user (ex: `shm_open`)

ref: DLIS-6198

### DIFF
--- a/src/shared_memory_manager.cc
+++ b/src/shared_memory_manager.cc
@@ -147,11 +147,11 @@ OpenSharedMemoryRegion(const std::string& shm_key, int* shm_fd)
 }
 
 TRITONSERVER_Error*
-GetSharedMemoryFileInfo(
+GetSharedMemoryRegionSize(
     const std::string& shm_key, int shm_fd, size_t* shm_region_size)
 {
-  struct stat sb;
-  if (fstat(shm_fd, &sb) == -1) {
+  struct stat file_status;
+  if (fstat(shm_fd, &file_status) == -1) {
     LOG_VERBOSE(1) << "fstat on shm_fd failed, errno: " << errno;
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INTERNAL,
@@ -160,15 +160,34 @@ GetSharedMemoryFileInfo(
 
   // According to POSIX standard, type off_t can be negative, so for sake of
   // catching possible under/overflows, assert that the size is non-negative.
-  if (sb.st_size < 0) {
+  if (file_status.st_size < 0) {
     LOG_VERBOSE(1) << "File size of shared memory region must be non-negative";
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INTERNAL,
         std::string("Invalid shared memory region: '" + shm_key + "'").c_str());
   }
 
-  // FIXME: Consider file permissions
-  *shm_region_size = static_cast<size_t>(sb.st_size);
+  *shm_region_size = static_cast<size_t>(file_status.st_size);
+  return nullptr;  // success
+}
+
+TRITONSERVER_Error*
+CheckSharedMemoryRegionSize(
+    const std::string& name, const std::string& shm_key, int shm_fd,
+    size_t offset, size_t byte_size)
+{
+  size_t shm_region_size = 0;
+  RETURN_IF_ERR(GetSharedMemoryRegionSize(shm_key, shm_fd, &shm_region_size));
+  // User-provided offset and byte_size should not go out-of-bounds.
+  if ((offset + byte_size) > shm_region_size) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        std::string(
+            "failed to register shared memory region '" + name +
+            "': invalid args")
+            .c_str());
+  }
+
   return nullptr;  // success
 }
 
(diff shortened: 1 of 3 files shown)


src/shared_memory_manager.cc
```diff
@@ -145,6 +146,35 @@ OpenSharedMemoryRegion(const std::string& shm_key, int* shm_fd)
...
+TRITONSERVER_Error*
+GetSharedMemoryRegionInfo(
+    const std::string& shm_key, int shm_fd, size_t* shm_region_size)
+{
+  struct stat sb;
+  if (fstat(shm_fd, &sb) == -1) {
+    LOG_VERBOSE(1) << "fstat on shm_fd failed, errno: " << errno;
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        std::string("Unable to stat shared memory region: '" + shm_key + "'")
```
nnshah1, line 158:
```suggestion
        std::string("Invalid shared memory region: '" + shm_key + "'")
```

src/shared_memory_manager.cc
```diff
@@ -145,6 +146,35 @@ OpenSharedMemoryRegion(const std::string& shm_key, int* shm_fd)
...
+            .c_str());
+  }
+
+  // According to POSIX standard, type off_t can be negative, so for sake of
+  // catching possible under/overflows, assert that the size is non-negative.
+  if (sb.st_size < 0) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        std::string(
+            "File size of shared memory region: '" + shm_key +
```
nnshah1, line 168:
```suggestion
            "Invalid shared memory region: '" + shm_key +
```

nnshah1: maybe add the reason to the log? but keep message generic ...

src/shared_memory_manager.cc
```diff
@@ -145,6 +146,35 @@ OpenSharedMemoryRegion(const std::string& shm_key, int* shm_fd)
   return nullptr;
 }
 
+TRITONSERVER_Error*
+GetSharedMemoryRegionInfo(
```
fpetrini15, line 150: Design looks solid! My only concern is naming overloading. This file has:
- CUDASharedMemoryInfo (struct)
- SharedMemoryInfo (struct)
- GetMemoryInfo (method)
- GetSharedMemoryRegionInfo (new method)

Would it make sense to name this method something more specific like `GetSharedMemoryRegionSize` to make it more readily distinguishable from the rest? I know a user can read the code to tell the difference and the current name makes it more readily extensible in the future, but just a thought.

rmccorm4: I didn't choose Size in the name since it may return some other fields like permissions in the near term.

The main distinction of this function is that it's checking info from actual shared memory file object, rather than some internal data structure we setup with user-provided values.

Open to other names though if that helps distinguish the functionality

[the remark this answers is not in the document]
fpetrini15: Yes, all of this will need to be merged with Windows, but I can handle this post-merge.

fpetrini15: That's what I figured. Carry on then, no issue 🚀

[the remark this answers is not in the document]
rmccorm4: Thanks for being willing to handle it as a follow-up Francesco!

rmccorm4: Renamed `RegionInfo` to `FileInfo` in hopes for some extra clarity

rmccorm4:
Added tests for system shm registration to L0_shared_memory.

- Tests fail locally before server-side changes, and pass after server-side changes ✅

nnshah1 (approved):
LGTM

nnshah1:
@rmccorm4 , @fpetrini15 

Small, non blocking items: 

I also thought while reading that `GetSharedMemoryRegionSize` was more intuitive than `Info` as we weren't getting more information. At the moment - I'm hoping permissions are checked when we actual do the map for reading / writing and don't need to be checked independently. If we do have to do more checks we could update it then. Also the comment says `metadata`, maybe `GetSharedMemoryRegionMetadata` ?

Another small tweak could be to wrap the validation into a function: `CheckRegionSize()` 

`struct stat sb` -> `struct stat file_status`?

`sb` wasn't immediately clear to me - it's not used very much 

These are all minor - and non blocking.

src/shared_memory_manager.cc
```diff
@@ -145,6 +146,32 @@ OpenSharedMemoryRegion(const std::string& shm_key, int* shm_fd)
   return nullptr;
 }
 
+TRITONSERVER_Error*
+GetSharedMemoryFileInfo(
+    const std::string& shm_key, int shm_fd, size_t* shm_region_size)
+{
+  struct stat sb;
```
nnshah1, line 153:
```suggestion
  struct stat file_status;
```

Tabrizian (dismissed):

fpetrini15 (approved):
TM,LG! 🚀

nnshah1 (approved):

PR #7062, opened 2024-04-01
Decoupled Async Execute

kthui:
Related PR: https://github.com/triton-inference-server/python_backend/pull/350

Tests for Python backend decoupled async execute function. Covered cases:
* A async execute function may start multiple coroutines and those are ran concurrently.
* Multiple async execute coroutines are ran concurrently.
* An async execute coroutine started late but finished early will respond before another async execute coroutine that started early but finished late.
* A coroutine may raise an exception and it will be logged.

PR #7100, opened 2024-04-11
Decoupled Async Execute (#7062)

kthui:
* Add async execute decoupled test

* Add decoupled bls async exec test

* Enhance test with different durations for concurrent executes


kthui:
Pick https://github.com/triton-inference-server/server/pull/7062 into r24.04

Tabrizian (approved):

PR #7106, opened 2024-04-12
Disable Dynamic Trace File

yinggeh:
Disable dynamic trace file location, aka `--trace_file` API.


nnshah1 (approved):
Thanks for the fast work! LGTM

src/tracer.cc
indrajit96, line 1: If we don't support update of trace_file now 
Should we clean up "filepath_specified" from the tracer..cc altogether? As filepath_specified_ will always be false?

Can take up in another PR as well as the current changes are for a PSIRT

yinggeh: No. It' not. Trace file can be specified when launching tritonserver. In this case filepath_specified is true.

rmccorm4 (approved):
LGTM - is there a separate PR for client/PA tests in the works?

PR #7127, opened 2024-04-17
Use a lower concurrency with more repetition for L0_memory_growth

krishung5:
The test is failing due to onnxruntime throwing CUDA OOM error. Decrease the concurrency and increase the repetition to make sure that CUDA memory isn't exhausted, and we still have enough amount of requests sending from PA to observe if the memory grows or not.


Tabrizian (approved):

krishung5:
Use a larger window to avoid intermittent PA unstable issue.

Tabrizian (approved):

PR #7131, opened 2024-04-18
Re-enable PA trace testing but remove setting trace file

matthewkotila:
Re-adds in PA trace testing but just avoids using the discontinued trace-file setting

### DIFF
--- a/qa/L0_perf_analyzer/test.sh
+++ b/qa/L0_perf_analyzer/test.sh
@@ -66,8 +66,10 @@ SEQ_WRONG_OUTPUT_JSONDATAFILE=`pwd`/../common/perf_analyzer_input_data_json/seq_
 
 REPEAT_INT32_JSONDATAFILE=`pwd`/../common/perf_analyzer_input_data_json/repeat_int32_data.json
 
+TRACE_FILE="trace.json"
+
 SERVER=/opt/tritonserver/bin/tritonserver
-SERVER_ARGS="--model-repository=${DATADIR} --trace-config triton,file=trace.json"
+SERVER_ARGS="--model-repository=${DATADIR} --trace-config triton,file=${TRACE_FILE}"
 SERVER_LOG="./inference_server.log"
 
 ERROR_STRING="error | Request count: 0 | : 0 infer/sec"
@@ -642,7 +644,6 @@ for PROTOCOL in grpc http; do
 
     # Testing that trace logging works
     set +e
-    TRACE_FILE="trace.json"
     rm ${TRACE_FILE}*
     $PERF_ANALYZER -v -i $PROTOCOL -m simple_savedmodel_sequence_object -p 2000 -t5 --sync \
     --trace-level TIMESTAMPS --trace-rate 1000 --trace-count 100 --log-frequency 10 \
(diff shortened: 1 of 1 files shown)


qa/L0_perf_analyzer/test.sh
```diff
@@ -639,6 +639,30 @@ for PROTOCOL in grpc http; do
         fi
         set -e
     done
+
+    # Testing that trace logging works
+    set +e
+    curl localhost:8000/v2/trace/setting -d '{"trace_level":["TIMESTAMPS"]}'
```
rmccorm4, line 645: Shouldn't the `$PERF_ANALYZER ... --trace-level TIMESTAMPS` do this for you?

matthewkotila: Good catch--fixed

rmccorm4 (approved):

qa/L0_perf_analyzer/test.sh
```diff
@@ -639,6 +639,29 @@ for PROTOCOL in grpc http; do
         fi
         set -e
     done
+
+    # Testing that trace logging works
+    set +e
+    TRACE_FILE="trace.json"
+    rm ${TRACE_FILE}*
+    $PERF_ANALYZER -v -i $PROTOCOL -m simple_savedmodel_sequence_object -p 2000 -t5 --sync \
```
rmccorm4, line 647: Having an extra block for something like this after David Y's changes would be a nice sanity check, and automatic reminder to test it if it gets re-enabled in the future:
```bash
${PERF_ANALYZER} ... --trace-file trace.json ...
if [ $? -eq 0]; then
  cat ${CLIENT_LOG}
  echo -e "\n***\n*** Test Failed. Expected to fail for unknown arg --trace-file"
  RET=1
fi
```

rmccorm4: @dyastremsky

matthewkotila: Nice idea--added

qa/L0_perf_analyzer/test.sh
```diff
@@ -662,6 +662,19 @@ for PROTOCOL in grpc http; do
     fi
     curl localhost:8000/v2/trace/setting -d '{"trace_level":["OFF"]}'
     set -e
+
+    # Testing that setting trace file does not work
+    $PERF_ANALYZER -v -i $PROTOCOL -m simple_savedmodel_sequence_object \
+    -p 2000 -t5 --sync --trace-file trace.json --trace-level TIMESTAMPS \
+    --trace-rate 1000 --trace-count 100 --log-frequency 10 \
+    --input-data=$SEQ_JSONDATAFILE -s ${STABILITY_THRESHOLD} >$CLIENT_LOG 2>&1
```
rmccorm4, line 670: nit: I'd remove all unnecessary args other than `--trace-file` and the `-m model` if they're irrelevant to the test, since we're just expecting it to fail at parsing/startup.

matthewkotila: Good note, I wasn't sure, but upon further reflection, it seems like they shouldn't matter. I left `-v` as that's a pattern and I also left setting the network protocol as I want to test both lower level clients send the trace file option correctly and receive the error correctly.

qa/L0_perf_analyzer/test.sh
```diff
@@ -639,6 +639,41 @@ for PROTOCOL in grpc http; do
         fi
         set -e
     done
+
+    # Testing that trace logging works
+    set +e
+    TRACE_FILE="trace.json"
```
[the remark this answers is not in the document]
matthewkotila: Nice catch--changed

rmccorm4 (approved):
:rocket:

PR #7132, opened 2024-04-18
Enhance OTEL testing to capture and verify Cancellation Requests and Non-Decoupled model inference.

indrajit96:
Added tests for 
1. For request cancellations, make sure spans are ended properly in various scenarios 
2. Verify tracing behavior, after this PR: https://github.com/triton-inference-server/server/pull/6017 

Tests added
1. test_non_decoupled() : Verify tracing for a non-decoupled inference
2. test_grpc_trace_all_input_required_model_cancel() : Verify trace after an inference request is cancelled in COMPUTE Phase
3. test_grpc_trace_model_cancel_in_queue() : Verify trace after an inference request is cancelled in QUEUE before COMPUTE Phase


qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -527,6 +609,52 @@
...
+            grpcclient.InferInput("WAIT", [1], "UINT32").set_data_from_numpy(
+                self.input_data["WAIT"]
+            ),
+        ]
+
+        triton_client = grpcclient.InferenceServerClient(
+            url="localhost:8001", verbose=True
+        )
+        # Expect the inference is successful
+        res = triton_client.infer(
```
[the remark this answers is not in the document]
oandreeva-nv: if we expect successful inference here, then I would recommend also checking that `res` matches an expected value

oandreeva-nv:
Could you please add a description, clarifying what tests were added?

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -199,6 +214,36 @@ def _check_events(self, span_name, events):
             )
             self.assertFalse(all(entry in events for entry in compute_events))
 
+    def _check_events_cancel(self, events):
```
oandreeva-nv, line 217: Maybe instead of a new `_check_events_cancel` we can somehow refactor `_check_events` and use it?

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -270,6 +315,11 @@ def _verify_contents(self, spans, expected_counts):
                 "Unexpeced number of " + name + " spans collected",
             )
 
+    def _verify_contents_cancel(self, spans):
```
oandreeva-nv, line 318: Same here, is it possible to refactor `_verify_contents` ?

oandreeva-nv:
Could you please attach a picture of a trace as displayed in jaeger with cancelled request. 

Another question, have you considered cases when request was cancelled, when it was in a queue and when it was already in a compute stage?

indrajit96:
> Could you please attach a picture of a trace as displayed in jaeger with cancelled request.
> 
> Another question, have you considered cases when request was cancelled, when it was in a queue and when it was already in a compute stage?



> Could you please attach a picture of a trace as displayed in jaeger with cancelled request.
> 
> Another question, have you considered cases when request was cancelled, when it was in a queue and when it was already in a compute stage?

Screenshot 2024-05-02 at 10 51 00 AM

Fixed this, added a model which waits for a delay before executing hence cancellation request is recieved before the execution starts

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -339,6 +399,22 @@
...
+    def _test_trace_cancel(
+        self,
+    ):
+        time.sleep(2 * COLLECTOR_TIMEOUT)
+        traces = self._parse_trace_log(self.filename)
+        expected_counts = dict(
+            {"compute": 1, self.input_all_required_model_name : 1, self.root_span: 1}
+        )
+        parsed_spans = traces[0]["resourceSpans"][0]["scopeSpans"][0]["spans"]
+        root_span = [
```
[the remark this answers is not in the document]
oandreeva-nv: `root_span` is not used, probably can remove

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -420,6 +497,24 @@ def _test_simple_trace(self, headers=None):
...
+            {"compute": 1, self.non_decoupled_model_name_: 1, self.root_span: 1}
+        )
+        expected_parent_span_dict = dict(
+            {"InferRequest": ["repeat_int32"], "repeat_int32": ["compute"]}
+        )
+        self._test_trace(
+            headers=headers,
+            expected_number_of_spans=expected_number_of_spans,
+            expected_counts=expected_counts,
+            expected_parent_span_dict=expected_parent_span_dict,
```
oandreeva-nv, line 515:
```suggestion
            expected_parent_span_dict=expected_parent_span_dict
```

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -232,7 +292,7 @@ def _test_resource_attributes(self, attributes):
             ),
         )
 
-    def _verify_contents(self, spans, expected_counts):
+    def _verify_contents(self, spans, expected_counts, is_cancel):
```
oandreeva-nv, line 295: nit: `is_cancel` -> `is_cancelled`

and please update `Args` field bellow on L304 with a new field.

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -397,7 +474,7 @@ def _test_trace(
         ][0]
         self.assertEqual(len(parsed_spans), expected_number_of_spans)
 
-        self._verify_contents(parsed_spans, expected_counts)
+        self._verify_contents(parsed_spans, expected_counts, False)
```
oandreeva-nv, line 477: I would recommend not use just False or True as an argument, it might be not clear what they stand for in this context. Please use `is_cancelled=False`

rmccorm4:
Can you update the PR title to be more descriptive? (cancellation, decoupled responses, etc. rather than JIRA ticket number)

qa/L0_trace/opentelemetry_unittest.py
```diff
@@ -339,6 +389,22 @@ def _verify_headers_propagated_from_client_if_any(self, root_span, headers):
             ),
         )
 
+    def _test_trace_cancel(self, is_queued):
+        # TO accommodate for delay in the model
```
oandreeva-nv, line 393: Could you please clarify what delay in the model? and if it is a model related, why are we using `COLLECTOR_TIMEOUT` variable?

oandreeva-nv:
I think it's in a good state. Let's address Ryan's comment and provide a clear description for 3 tests, added in this PR, to the PR description

qa/L0_trace/opentelemetry_unittest.py
indrajit96: We want to capture a cancellation request traces WHILE the inference is in the COMPUTE stage.
Because the model "input_all_required" has a delay/wait in the compute phase so the cancellation request can be send while the request is waiting in the compute phase.
The idea here is to wait before we try and read the traces from the file.

If we do not wait, the file in empty without traces.
Updated the comments in the test to reflect his better.

oandreeva-nv (approved):
LGTM! Thank you for this hard work!

PR #7134, opened 2024-04-18
Update L0_trace for DLIS-6462

indrajit96:
Remove checking on trace_mode when trace_file update fails.

### DIFF
--- a/qa/L0_trace/test.sh
+++ b/qa/L0_trace/test.sh
@@ -239,9 +239,6 @@ assert_curl_failure "trace_file updated through network protocol expects an erro
 if [ `grep -c "\"error\":\"trace file location can not be updated through network protocol\"" ./curl.out` != "1" ]; then
     RET=1
 fi
-if [ `grep -c "\"trace_mode\":\"triton\"" ./curl.out` != "1" ]; then
-    RET=1
-fi
 
 # Use a different name
 update_trace_setting "simple" '{"log_frequency":"2"}'


qa/L0_trace/test.sh
```diff
@@ -239,9 +239,6 @@ assert_curl_failure "trace_file updated through network protocol expects an erro
 if [ `grep -c "\"error\":\"trace file location can not be updated through network protocol\"" ./curl.out` != "1" ]; then
     RET=1
 fi
-if [ `grep -c "\"trace_mode\":\"triton\"" ./curl.out` != "1" ]; then
```
oandreeva-nv, line 242: can we please add a `TODO` or something, to indicate that when `trace_file` update is back, we need to test `trace_mode` as well.

rmccorm4: Don't we have other places in the test that test the returned mode is correct? Isn't the `trace_file` update / error check more standalone?

yinggeh:
Can you please remove line 567 `RET=0`? This line resets `RET` thus failures before that were not captured. cc @GuanLuo

indrajit96:
> Can you please remove line 567 `RET=0`? This line resets `RET` thus failures before that were not captured. cc @GuanLuo

Resolved

yinggeh (approved):

yinggeh:
Would you merge into 24.04?

PR #7141, opened 2024-04-19
Add metrics model namespacing label test

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/343

Add test to ensure "namespace" label exists on metrics when the server is launched with `--model-namespacing=true`.


oandreeva-nv (commented):
Should we also check that for default behavior (assuming it is `model-namespacing=false`), namespacing label is not present?

kthui:
> Add test for namespace off

Sure. [Add test for namespace off](https://github.com/triton-inference-server/server/pull/7141/commits/98167f5e31c2055ce7e402c4429dde88006556fa)

oandreeva-nv (approved):
LGTM

PR #7142, opened 2024-04-20
Fix TensorRT-LLM

mc-nv:


build.py
```diff
@@ -1708,6 +1708,30 @@ def tensorrtllm_prebuild(cmake_script):
     # Export the TRT_ROOT environment variable
     cmake_script.cmd("export TRT_ROOT=/usr/local/tensorrt")
     cmake_script.cmd("export ARCH=$(uname -m)")
+    cmake_script.cmd(
+        'export LD_LIBRARY_PATH="/usr/local/cuda/compat/lib.real:${LD_LIBRARY_PATH}"'
```
Tabrizian, line 1712: Why is this line required?

Tabrizian (approved):
@krishung5 might be more familiar with these changes.

build.py
mc-nv: Because `libcuda.so.1` can't be resolved till `--runtime=nvidia` is passed. 
It's a workaround that allows us refer to the links in build container, can be deprecated in future. 

You can create task and assign to me.

build.py
```diff
@@ -1102,17 +1102,17 @@ def create_dockerfile_linux(
 RUN python3 -m pip install --upgrade pip \\
       && pip3 install transformers
 
-# Uninstall unused nvidia packages
-RUN if pip freeze | grep -q "nvidia.*"; then \\
-        pip freeze | grep "nvidia.*" | xargs pip uninstall -y; \\
-    fi
-RUN pip cache purge
```
krishung5, line 1109: I was  wondering why do we need to remove this section? I think this is just to reduce some bytes for the container size. Do we see issues if remove those libraries?

build.py
```diff
@@ -1102,17 +1102,17 @@ def create_dockerfile_linux(
...
-
 # Drop the static libs
 RUN ARCH="$(uname -i)" \\
       && rm -f ${TRT_ROOT}/targets/${ARCH}-linux-gnu/lib/libnvinfer*.a \\
           ${TRT_ROOT}/targets/${ARCH}-linux-gnu/lib/libnvonnxparser_*.a
 
+# Install TensorRT-LLM
+RUN python3 -m pip install /opt/tritonserver/backends/tensorrtllm/tensorrt_llm-*.whl -U --pre --extra-index-url https://pypi.nvidia.com \\
```
krishung5, line 1111: I see that we are pip installing TRT-LLM in the `Dockerfile.triton.trt_llm_backend`. Do we still need to pip install TRT-LLM again here?

build.py
mc-nv: Because it eventually crash the CI

krishung5 (approved):

PR #7143, opened 2024-04-20
Change TensorRT-LLM

mc-nv:


krishung5:
@mc-nv I think the changes in this PR are merged within other previous PR. Should we close this one?

build.py
```diff
@@ -1108,14 +1094,7 @@ def create_dockerfile_linux(
 RUN python3 -m pip install --upgrade pip \\
       && pip3 install transformers
 
-# Drop the static libs
-RUN ARCH="$(uname -i)" \\
-      && rm -f ${TRT_ROOT}/targets/${ARCH}-linux-gnu/lib/libnvinfer*.a \\
-          ${TRT_ROOT}/targets/${ARCH}-linux-gnu/lib/libnvonnxparser_*.a
-
```
krishung5, line 1115: Why removing these lines? Is dropping libs causing some issues?

mc-nv: Removing of those libs doesn't make our container image smaller if that's what is it designed for.
Is the removing of those libs has any other logical purpose?

krishung5: Thought it would reduce the container size.. Thanks for the explanation.

krishung5 (approved):
LGTM! Might need to rebase onto main branch.

PR #7163, opened 2024-04-26
Add extra footer to documentation

mc-nv:


docs/_static/NVIDIA-LogoBlack.svg
yinggeh, line 1: Is this image ever used?

Tabrizian (approved):

Tabrizian:
@mc-nv if the image is not used let's remove it.

docs/_static/NVIDIA-LogoBlack.svg
mc-nv: That image is official logo, that came with request to be added/used on a page. However by going over other examples on https://docs.nvidia.com  it doesn't look like logo widely used by company in the footer.

At the same time logo has `svg` vector based format, which is more preferable for HTML configuration. 
I decided to add it to the `_static` folder in case if there will be a push back or a scenario to use it.

yinggeh (approved):

PR #7172, opened 2024-04-29
Fix Python Shm Client Leak

fpetrini15:
**Goal:** The python client was not properly deallocating dynamic memory when destroying shared memory regions. This PR bundle fixes the leak and adds an additional test case.

**Sub-Goal:** Fix the formatting in the .rst document to fix an unrelated CI failure.

Client PR: https://github.com/triton-inference-server/client/pull/614

### DIFF
--- a/qa/L0_shared_memory/shared_memory_test.py
+++ b/qa/L0_shared_memory/shared_memory_test.py
@@ -34,6 +34,7 @@ import os
 import unittest
 
 import numpy as np
+import psutil
 import test_util as tu
 import tritonclient.grpc as grpcclient
 import tritonclient.http as httpclient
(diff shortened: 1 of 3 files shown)


qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -444,6 +445,32 @@ def test_register_out_of_bound(self):
                 register_offset=create_byte_size + 1,
             )
 
+    def test_python_client_leak(self):
```
Tabrizian, line 448: Does it run fast? Also, do you think there is anything that could make this test flaky (e.g. the threshold percentage)?

@krishung5 do you think we can enhance `L0_client_memory_growth` with this test?

fpetrini15: Yes, in my experience it runs very quickly. I added the threshold because in previous experiments I would see a "warm-up" or sorts where memory would very slightly increase at the beginning, then remain stable. Now when I test locally, it remains stable the whole time. I left the threshold in JIC.

I don't envision it being flaky. If it increases pass this threshold value, I imagine we have a true memory leak. But I can monitor it over the coming months.

Tabrizian (approved):

qa/L0_shared_memory/shared_memory_test.py
krishung5: @Tabrizian Yes, I think it will be a great enhancement in [memory_growth_test.py](https://github.com/triton-inference-server/client/blob/main/src/python/examples/memory_growth_test.py). Filed a ticket for this: DLIS-6649.

@fpetrini15 Nice enhancement!

fpetrini15: @rmccorm4 is to thank for this enhancement. I just took his PoC and made a test out of it. 🚀

qa/L0_shared_memory/shared_memory_test.py
```diff
@@ -444,6 +445,32 @@ def test_register_out_of_bound(self):
...
+                )
+
+            shm_handle = shm.create_shared_memory_region(
+                "shmtest", "/shmtest", byte_size
+            )
+            shm.destroy_shared_memory_region(shm_handle)
+            i += 1
+        final_mem_usage = process.memory_info().rss / 1024**2
+        self.assertTrue(
+            (initial_mem_usage <= final_mem_usage <= threshold),
```
rmccorm4, line 470: TIL Python allows you to do this and it works as expected :brain:

PR #7176, opened 2024-05-01
Add testing for L0_iterative_tutorial

Tabrizian:
https://github.com/triton-inference-server/tutorials/pull/90


oandreeva-nv (commented):
Do you think it make sense to make a more general structure for tutorial's tests in this PR or follow up PRs?

e.g. `qa/L0_iterative_tutorial/test.sh` ->  `L0_tutorials/iterative_tutorial/test.sh`

Tabrizian:
@oandreeva-nv I think we did this for Python backend but the feedback was that it makes it harder to determine which part of the test is flaky since they are all bundled together. Additionally, the tests can be executed in parallel if they are separate.

oandreeva-nv (dismissed):
LGTM

qa/L0_iterative_tutorial/test.sh
```diff
@@ -0,0 +1,49 @@
+# Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 1: nit similar to Olga's comment - can we rename this to `L0_tutorial_iterative_scheduling` (and future `L0_tutorial_*`) so we can quickly identify/group the tutorial repo based tests?

qa/L0_iterative_tutorial/test.sh
```diff
@@ -0,0 +1,49 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+rm -rf tutorials
+git clone https://github.com/triton-inference-server/tutorials.git  -b ${TUTORIAL_REPO_TAG} --depth 1
+
+cd tutorials/Conceptual_Guide/Part_7-iterative_scheduling/tests
+RET=0
+
+# Install the dependencies
+pip3 install "transformers[torch]"
```
rmccorm4, line 34: Do we need to install the tritonserver python wheel for the test?

Tabrizian: In the devel container it is already installed: https://github.com/triton-inference-server/server/blob/20627268b9f7e951e8be7f49f0d28bf76f178961/Dockerfile.QA#L388-L389

rmccorm4: Sweet, thanks for confirming :rocket:

qa/L0_tutorial_iterative_scheduling/test.sh
```diff
@@ -0,0 +1,49 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+rm -rf tutorials
+git clone https://github.com/triton-inference-server/tutorials.git  -b ${TUTORIAL_REPO_TAG} --depth 1
+
+cd tutorials/Conceptual_Guide/Part_7-iterative_scheduling/tests
+RET=0
+
+# Install the dependencies
+pip3 install "transformers[torch]"
```
rmccorm4, line 34: Should we lock down the version to whatever worked while testing to avoid unexpected breakage?

qa/L0_tutorial_iterative_scheduling/test.sh
```diff
@@ -0,0 +1,49 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+rm -rf tutorials
+git clone https://github.com/triton-inference-server/tutorials.git  -b ${TUTORIAL_REPO_TAG} --depth 1
```
rmccorm4, line 28: Should we default `TUTORIAL_REPO_TAG` for ergonomics of local testing outside of a pipeline? Something like this

```bash
TUTORIAL_REPO_TAG=${TUTORIAL_REPO_TAG:-"main"}
```

PR #7178, opened 2024-05-02
Validate CUDA SHM region registration size

krishung5:
- [x] Validate CUDA SHM region size during registration
- [x] Add CUDA SHM registration tests
- [x] Refactor tests for CUDA SHM and System SHM

### DIFF
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -331,7 +331,7 @@ class CudaSharedMemoryTest(tu.TestResultCollector):
         # Verify various edge cases of registered region size don't go out of bounds of the actual created shm region's size.
         with self.assertRaisesRegex(
             InferenceServerException,
-            r"failed to register CUDA shared memory region.*register size exceeds CUDA shared memory region size",
+            "failed to register shared memory region.*invalid args",
         ):
             self._configure_server(
                 create_byte_size=create_byte_size,
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -331,7 +331,7 @@ class CudaSharedMemoryTest(tu.TestResultCollector):
         # Verify various edge cases of registered region size don't go out of bounds of the actual created shm region's size.
         with self.assertRaisesRegex(
             InferenceServerException,
-            "failed to register CUDA shared memory region.*invalid args",
+            "failed to register shared memory region.*invalid args",
         ):
             self._configure_server(
                 create_byte_size=create_byte_size,
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -331,7 +331,7 @@ class CudaSharedMemoryTest(tu.TestResultCollector):
         # Verify various edge cases of registered region size don't go out of bounds of the actual created shm region's size.
         with self.assertRaisesRegex(
             InferenceServerException,
-            "failed to register cuda memory region.*register size exceeds cuda shared memory region size",
+            r"failed to register CUDA shared memory region.*register size exceeds CUDA shared memory region size",
         ):
             self._configure_server(
                 create_byte_size=create_byte_size,
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -327,7 +327,7 @@ class CudaSharedMemoryTest(tu.TestResultCollector):
         self._cleanup_server(shm_handles)
 
     def test_register_out_of_bound(self):
-        create_byte_size = 64
+        create_byte_size = self.DEFAULT_SHM_BYTE_SIZE
         # Verify various edge cases of registered region size don't go out of bounds of the actual created shm region's size.
         with self.assertRaisesRegex(
             InferenceServerException,
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2019-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2019-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -33,6 +33,7 @@ sys.path.append("../common")
 import os
 import unittest
 
+import infer_util as iu
 import numpy as np
 import test_util as tu
 import tritonclient.grpc as grpcclient
@@ -42,6 +43,24 @@ from tritonclient.utils import *
 
 
 class CudaSharedMemoryTest(tu.TestResultCollector):
+    DEFAULT_SHM_BYTE_SIZE = 64
+
+    def setUp(self):
+        self._setup_client()
+
+    def _setup_client(self):
+        self.protocol = os.environ.get("CLIENT_TYPE", "http")
+        if self.protocol == "http":
+            self.url = "localhost:8000"
+            self.triton_client = httpclient.InferenceServerClient(
+                self.url, verbose=True
+            )
+        else:
+            self.url = "localhost:8001"
+            self.triton_client = grpcclient.InferenceServerClient(
+                self.url, verbose=True
+            )
+
     def test_invalid_create_shm(self):
         # Raises error since tried to create invalid cuda shared memory region
         try:
@@ -52,19 +71,15 @@ class CudaSharedMemoryTest(tu.TestResultCollector):
 
     def test_valid_create_set_register(self):
         # Create a valid cuda shared memory region, fill data in it and register
-        if _protocol == "http":
-            triton_client = httpclient.InferenceServerClient(_url, verbose=True)
-        else:
-            triton_client = grpcclient.InferenceServerClient(_url, verbose=True)
         shm_op0_handle = cshm.create_shared_memory_region("dummy_data", 8, 0)
         cshm.set_shared_memory_region(
             shm_op0_handle, [np.array([1, 2], dtype=np.float32)]
         )
-        triton_client.register_cuda_shared_memory(
+        self.triton_client.register_cuda_shared_memory(
             "dummy_data", cshm.get_raw_handle(shm_op0_handle), 0, 8
         )
-        shm_status = triton_client.get_cuda_shared_memory_status()
-        if _protocol == "http":
+        shm_status = self.triton_client.get_cuda_shared_memory_status()
+        if self.protocol == "http":
             self.assertEqual(len(shm_status), 1)
         else:
             self.assertEqual(len(shm_status.regions), 1)
@@ -72,14 +87,10 @@ class CudaSharedMemoryTest(tu.TestResultCollector):
 
     def test_unregister_before_register(self):
         # Create a valid cuda shared memory region and unregister before register
-        if _protocol == "http":
-            triton_client = httpclient.InferenceServerClient(_url, verbose=True)
-        else:
-            triton_client = grpcclient.InferenceServerClient(_url, verbose=True)
         shm_op0_handle = cshm.create_shared_memory_region("dummy_data", 8, 0)
-        triton_client.unregister_cuda_shared_memory("dummy_data")
-        shm_status = triton_client.get_cuda_shared_memory_status()
-        if _protocol == "http":
+        self.triton_client.unregister_cuda_shared_memory("dummy_data")
+        shm_status = self.triton_client.get_cuda_shared_memory_status()
+        if self.protocol == "http":
             self.assertEqual(len(shm_status), 0)
         else:
             self.assertEqual(len(shm_status.regions), 0)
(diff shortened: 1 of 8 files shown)


src/shared_memory_manager.cc
```diff
@@ -365,6 +405,18 @@ SharedMemoryManager::RegisterCUDASharedMemory(
   // Get CUDA shared memory base address
   TRITONSERVER_Error* err =
       OpenCudaIPCRegion(cuda_shm_handle, &mapped_addr, device_id);
+  if (err != nullptr) {
```
rmccorm4, line 408: Is this a duplicate check to the err check on line 420 below?

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,46 @@ OpenCudaIPCRegion(
...
+  size_t shm_region_size = 0;
+  RETURN_IF_ERR(GetCudaSharedMemoryRegionSize(data_ptr, shm_region_size));
+
+  // User-provided offset and byte_size should not go out-of-bounds.
+  if (byte_size > shm_region_size) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        std::string(
+            "failed to register CUDA shared memory region '" + name +
+            "': register size exceeds CUDA shared memory region size")
```
rmccorm4, line 301: In system shm equivalent, I was guided to keep the error messages more vague: https://github.com/triton-inference-server/server/blob/96e4c0445d1329ee006fac6c58fdb39cbbf3e8a0/src/shared_memory_manager.cc#L187-L188.

@nnshah1 asuming you'd like the same here.

rmccorm4 (commented):
Refactor and general approach LGTM! Only minor comments, thanks for tackling this so quickly @krishung5 !

src/shared_memory_manager.cc
nnshah1: this is on the border - but I think keeping the reasoning as clear but not leaking details is better. 

"invalid size" could be an alternate - but lets' match / use the same messaging as in system shared memory

src/shared_memory_manager.cc
krishung5: Good catch! Removed the duplicate.

src/shared_memory_manager.cc
krishung5: Sounds good. Updated the error message.

src/shared_memory_manager.cc
```diff
@@ -297,8 +297,8 @@ CheckCudaSharedMemoryRegionSize(
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INVALID_ARG,
         std::string(
-            "failed to register CUDA shared memory region '" + name +
-            "': register size exceeds CUDA shared memory region size")
+            "failed to register shared memory region '" + name +
```
rmccorm4, line 300: failed to register `CUDA` shared memory region

qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
```diff
@@ -331,7 +331,7 @@ def test_register_out_of_bound(self):
         # Verify various edge cases of registered region size don't go out of bounds of the actual created shm region's size.
         with self.assertRaisesRegex(
             InferenceServerException,
-            r"failed to register CUDA shared memory region.*register size exceeds CUDA shared memory region size",
+            "failed to register shared memory region.*invalid args",
```
rmccorm4, line 334: failed to register `CUDA` shared memory region

src/shared_memory_manager.cc
```diff
@@ -406,25 +406,19 @@ SharedMemoryManager::RegisterCUDASharedMemory(
   TRITONSERVER_Error* err =
       OpenCudaIPCRegion(cuda_shm_handle, &mapped_addr, device_id);
   if (err != nullptr) {
+    std::string err_str = TRITONSERVER_ErrorMessage(err);
+    TRITONSERVER_ErrorDelete(err);
```
rmccorm4, line 410: Can this block just be condensed to a `RETURN_IF_ERROR(OpenCudaIPCRegion(...));` instead?

qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
krishung5: Yes, the api for cuda shm is `cuda_shared_memory` and for system shm is `register_system_shared_memory`, so the user should know if this is for cuda shm or system shm. In this case I think we could provide the `CUDA` keyword in the error message if we want.

krishung5: I updated the error message to include `CUDA` keyword for now. @nnshah1 Let me know if there's any concern for this.

rmccorm4 (approved):

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,46 @@ OpenCudaIPCRegion(
...
+  if (result != CUDA_SUCCESS) {
+    const char* errorString;
+    if (cuGetErrorString(result, &errorString) != CUDA_SUCCESS) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INTERNAL, "Failed to get CUDA error string");
+    }
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL,
+        std::string(
+            "Failed to get CUDA address range: " + std::string(errorString))
```
nnshah1, line 282: do these errors go back to the client?

nnshah1: if so - I recommend logging the detailed error but not passing the detailed error back to client / user of api

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,46 @@ OpenCudaIPCRegion(
...
+{
+  size_t shm_region_size = 0;
+  RETURN_IF_ERR(GetCudaSharedMemoryRegionSize(data_ptr, shm_region_size));
+
+  // User-provided offset and byte_size should not go out-of-bounds.
+  if (byte_size > shm_region_size) {
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        std::string(
+            "failed to register CUDA shared memory region '" + name +
```
nnshah1, line 300: im slightly in favor of removing CUDA - but more important is the level of detail in line 282

nnshah1 (changes requested):
want to review level of detail passed back to client / user of api

src/shared_memory_manager.cc
krishung5: Good catch. Yes, the errors will be passed back to the client. Changed it to just pass a less informative error to the client.

src/shared_memory_manager.cc
krishung5: Np! Removed `CUDA` from the error.

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,45 @@ OpenCudaIPCRegion(
...
+TRITONSERVER_Error*
+GetCudaSharedMemoryRegionSize(CUdeviceptr data_ptr, size_t& shm_region_size)
+{
+  CUdeviceptr* base = nullptr;
+  CUresult result = cuMemGetAddressRange(base, &shm_region_size, data_ptr);
+  if (result != CUDA_SUCCESS) {
+    const char* errorString;
+    if (cuGetErrorString(result, &errorString) != CUDA_SUCCESS) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INTERNAL, "Failed to get CUDA error string");
```
nnshah1, line 277:
```suggestion
          TRITONSERVER_ERROR_INTERNAL, "Failed to get shared memory information");
```

nnshah1: same here - let's log details - but error be generic

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,45 @@ OpenCudaIPCRegion(
...
+    }
+    // Should not pass the detailed error message back to the client.
+    LOG_ERROR << "Failed to get CUDA address range: " << errorString;
+    return TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INTERNAL, "Failed to get CUDA address range");
```
nnshah1, line 282:
```suggestion
        TRITONSERVER_ERROR_INTERNAL, "Failed to get shared memory information");
```

nnshah1: let's define one message for this method - and keep it generic

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,45 @@ OpenCudaIPCRegion(
...
+  }
+  return nullptr;
+}
+
+TRITONSERVER_Error*
+CheckCudaSharedMemoryRegionSize(
+    const std::string& name, CUdeviceptr data_ptr, size_t byte_size)
+{
+  size_t shm_region_size = 0;
+  RETURN_IF_ERR(GetCudaSharedMemoryRegionSize(data_ptr, shm_region_size));
```
nnshah1, line 292: instead of return if error - we could treat this as internal - 

I think I would just do
```
success = GetSharedMemoryRegionSice()
if not sucesss or (byte_size > shm_region_size)
{
 return "Failed to register shared memory region .... "
}
```

nnshah1: that way we have one consistent simple error message (details of why it failed are not so relevant to client in this case) - and we have logs for more detailed analysis.

nnshah1 (changes requested):
Want to normalize to a single error message for "failed to register shared memory .. invalid args"

src/shared_memory_manager.cc
krishung5: Makes sense! Do you think we'd still need to keep the generic error message if we are only returning a single error message for "failed to register shared memory .. invalid args"?
```
return TRITONSERVER_ErrorNew(
        TRITONSERVER_ERROR_INTERNAL, "Failed to get shared memory information");
```
I think we could probably do something like
```
{
   // In GetCudaSharedMemoryRegionSize
   // Return the details with Triton error object, which will NOT be exposed to client
   return TRITONSERVER_ErrorNew(
             TRITONSERVER_ERROR_INTERNAL, "Failed to get CUDA error string");
}
...
...
{
   auto err = GetCudaSharedMemoryRegionSize(data_ptr, shm_region_size);
   // User-provided offset and byte_size should not go out-of-bounds.
   if (err != nullptr || byte_size > shm_region_size) {
    // Should not pass the detailed error message back to the client.
    LOG_ERROR << TRITONSERVER_ErrorMessage(err); // log the error in one place
    return TRITONSERVER_ErrorNew(
        std::string(
            "failed to register shared memory region '" + name +
            "': invalid args")
            .c_str());
}

```

src/shared_memory_manager.cc
```diff
@@ -264,6 +264,51 @@ OpenCudaIPCRegion(
...
+  size_t shm_region_size = 0;
+  auto err = GetCudaSharedMemoryRegionSize(data_ptr, shm_region_size);
+
+  // User-provided offset and byte_size should not go out-of-bounds.
+  if (err != nullptr || byte_size > shm_region_size) {
+    if (err != nullptr) {
+      // Should not pass the detailed error message back to the client.
+      LOG_ERROR << TRITONSERVER_ErrorMessage(err);
+      TRITONSERVER_ErrorDelete(err);
+    }
```
krishung5, line 301: @nnshah1 I went ahead and made some changes to this part. Let me know if this makes sense to you. I think we could remove the part where we log out `TRITONSERVER_ERROR_INTERNAL, "Failed to get shared memory information"`. I think that would be duplicated log if we already log out the detailed errors and pass the general one back to the client.

src/shared_memory_manager.cc
jbkyang-nvi: why did we resolve this conversation? It seems like adding CUDA here is still a good idea?

src/shared_memory_manager.cc
jbkyang-nvi: This should still involve CUDA because user could request both system shm and CUDA shm right?

jbkyang-nvi (commented):
LGTM. Minor suggestions. Thanks for refactoring the tests

src/shared_memory_manager.cc
krishung5: I think after the discussion with Neelay, we think that it's better to remove the `CUDA` keyword to expose as little information as possible to the client: https://github.com/triton-inference-server/server/pull/7178#discussion_r1589740725

src/shared_memory_manager.cc
krishung5: Please see here for the context of removing `CUDA`: https://github.com/triton-inference-server/server/pull/7178#discussion_r1591672353

qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
```diff
@@ -295,40 +292,52 @@ def test_too_big_shm(self):
...
+        self.triton_client.unregister_cuda_shared_memory()
+        status_after = self.triton_client.get_cuda_shared_memory_status()
+        if self.protocol == "http":
             self.assertEqual(len(status_after), 0)
         else:
             self.assertEqual(len(status_after.regions), 0)
         self._cleanup_server(shm_handles)
 
+    def test_register_out_of_bound(self):
+        create_byte_size = 64
```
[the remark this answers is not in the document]
krishung5: Good catch! Fixed.

krishung5:
Made the error message less informative for `OpenCudaIPCRegion` and fixed the L0_grpc test.

src/shared_memory_manager.cc
```diff
@@ -255,11 +255,57 @@ OpenCudaIPCRegion(
   cudaError_t err = cudaIpcOpenMemHandle(
       data_ptr, *cuda_shm_handle, cudaIpcMemLazyEnablePeerAccess);
   if (err != cudaSuccess) {
+    // Should not pass the detailed error message back to the client.
```
nnshah1, line 258:
```suggestion
    // Log detailed error message
```

nnshah1 (dismissed):
error logging looks good - one slight suggested tweak to comment

src/shared_memory_manager.cc
krishung5: Updated!

nnshah1 (approved):

PR #7192, opened 2024-05-07
Remove unnecessary wait in case of failed stub creation

indrajit96:
Added a new "Faulty" model to repro DLIS-5819 and cause the server to hang.
The test
1. First loads a "correct" model.
2. Then loads a "faulty" model.

The expectation is the server should NOT hang.

### DIFF
--- a/qa/L0_backend_python/model_control/model_control_test.py
+++ b/qa/L0_backend_python/model_control/model_control_test.py
@@ -82,6 +82,24 @@ class ExplicitModelTest(unittest.TestCase):
                 self.assertFalse(client.is_model_ready(model_name))
                 self.assertFalse(client.is_model_ready(ensemble_model_name))
 
+    def test_faulty_model_load(self):
+        working_model_name = "identity_fp32"
+        faulty_model_name = "faulty_model"
+        with httpclient.InferenceServerClient(f"{_tritonserver_ipaddr}:8000") as client:
+            for _ in range(5):
+                # Load a correct model
+                client.load_model(working_model_name)
+                # Load a faulty model
+                # Load a faulty model
+                try:
+                    client.load_model(faulty_model_name)
+                except:
+                    pass
+                # Check if server is responsive
+                self.assertTrue(client.is_model_ready(working_model_name))
+                # Verify faulty model is not loaded
+                self.assertFalse(client.is_model_ready(faulty_model_name))
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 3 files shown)


qa/L0_backend_python/model_control/model_control_test.py
```diff
@@ -82,6 +82,24 @@ def test_model_reload(self):
...
                 self.assertFalse(client.is_model_ready(ensemble_model_name))
 
+    def test_faulty_model_load(self):
+        working_model_name = "identity_fp32"
+        faulty_model_name = "faulty_model"
+        with httpclient.InferenceServerClient(f"{_tritonserver_ipaddr}:8000") as client:
+            # Load a correct model
+            client.load_model(working_model_name)
+            # Load a faulty model
+            # Load a faulty model
```
krishung5, line 92: Remove the duplicated comment.

qa/L0_backend_python/model_control/model_control_test.py
```diff
@@ -82,6 +82,24 @@ def test_model_reload(self):
...
+            try:
+                client.load_model(faulty_model_name)
+            except:
+                pass
```
krishung5, line 96: I think the try-except can be removed as we expect it to be passing.

qa/L0_backend_python/model_control/test.sh
```diff
@@ -37,9 +37,12 @@ source ../../common/util.sh
 
 mkdir -p models/identity_fp32/1/
 mkdir -p models/simple_identity_fp32/1/
+mkdir -p models/faulty_model/
```
krishung5, line 40: For faulty model, I think we could reuse this one: https://github.com/triton-inference-server/server/blob/main/qa/python_models/auto_complete_error/model.py

krishung5 (changes requested):

qa/L0_backend_python/model_control/model_control_test.py
```diff
@@ -82,6 +82,23 @@ def test_model_reload(self):
                 self.assertFalse(client.is_model_ready(model_name))
                 self.assertFalse(client.is_model_ready(ensemble_model_name))
 
+    def test_faulty_model_load(self):
+        working_model_name = "identity_fp32"
+        faulty_model_name = "auto_complete_error"
+        with httpclient.InferenceServerClient(f"{_tritonserver_ipaddr}:8000") as client:
+            # Load a correct model
+            client.load_model(working_model_name)
```
Tabrizian, line 90: Can you add a comment why a working model is required to reproduce the issue?

krishung5 (dismissed):

qa/L0_backend_python/model_control/model_control_test.py
indrajit96: I am still trying to figure this out.
I think we can merge this Test for 24.05 and meanwhile I will continue my investigation.
We have already merged the python backend change.

PR #7217, opened 2024-05-14
Jetpack 5.1.2 Patch 2

fpetrini15:
**Goal:** Provide a PR to be used as a basis for review comments. This PR adds build functionality/testing to fix arbitrary file uploads via the model load API for the Jetpack 5.1.2 platform. Notably, it will package the version of libboost_filesystem.so.1.80.0 that was used to compile `tritonserver` in a folder called `boost_filesystem` along with our `lib`, `bin`, etc. folders.

_Note:_ This PR will not be merged. Please also note that the change to the lifecycle test simply removes all tests except the file override tests.


build.py
```diff
@@ -1555,6 +1555,11 @@ def core_build(cmake_script, repo_dir, cmake_dir, build_dir, install_dir,
         cmake_script.cp(
             os.path.join(repo_install_dir, 'lib', 'libtritonserver.so'),
             os.path.join(install_dir, 'lib'))
+        # Copy boost filesystem library used to compile tritonserver
+        cmake_script.mkdir(os.path.join(install_dir, 'boost_filesystem'))
+        cmake_script.cp(
+            os.path.join('/usr', 'lib', 'libboost_filesystem.so.1.80.0'),
+            os.path.join(install_dir, 'boost_filesystem'))
```
mc-nv, line 1562: Are we going to update our `x86` and `sbsa` images?

fpetrini15: This is specifically just for our Jetson Xavier platform, which we release as a tar file.

build.py
```diff
@@ -948,8 +948,8 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
 # Current libboost-dev apt packages are < 1.78, so install from tar.gz
 RUN wget -O /tmp/boost.tar.gz \
         https://boostorg.jfrog.io/artifactory/main/release/1.80.0/source/boost_1_80_0.tar.gz && \
-    (cd /tmp && tar xzf boost.tar.gz) && \
-    mv /tmp/boost_1_80_0/boost /usr/include/boost
+    (cd /tmp && tar xzf boost.tar.gz) && cd /tmp/boost_1_80_0 && mv /tmp/boost_1_80_0/boost /usr/include/boost && \
+    ./bootstrap.sh --prefix=/usr && ./b2 --with-filesystem install
```
rmccorm4, line 952: Nice `--with-filesystem` optimization to exclude unnecessary deps :pray:

build.py
mc-nv: What happen if someone will decide to rebuild `x86/sbsa` container from source, will it work as expected?

rmccorm4, line 1562: What is this `install/boost_filesystem` dir used for? So it gets packaged up in jetson tarball?

fpetrini15: As I understand it, Jetson Xavier is exclusively an `aarch64` [platform](https://www.nvidia.com/en-us/autonomous-machines/embedded-systems/jetson-xavier-series/#:~:text=View%20Jetson%20Xavier%20Technical%20Specifications). Any attempt to build an image using this branch outside the scope of this platform would be improper usage.

Are you asking me to put up guardrails in the build script to ensure this branch is not used for unsupported platforms? All this said, I don't believe any of these changes would be breaking for x86 if built from source.

fpetrini15: Exactly. It is just a way to package the version of libboost_filesystem.so.1.80.0 used to compile tritonserver into the final tarball. From there users can add it to `/usr/lib` or a custom location and add it to their linker path.

nv-kmcgill53 (approved):

rmccorm4 (approved):

PR #7219, opened 2024-05-14
Add testing for escaped log messages

nnshah1:
- Test for escaped and unescaped log formats
- Test for log injection
- Minor updates to test search patterns


qa/L0_logging/log_injection_test.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/python
+
+# Copyright 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 3:
```suggestion
# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

qa/L0_logging/log_injection_test.py
```diff
@@ -0,0 +1,55 @@
...
+            print("context creation failed: " + str(e))
+            sys.exit(1)
+
+        input_name = "'nothing_wrong'\nI0205 18:34:18.707423 1 [file.cc:123] THIS ENTRY WAS INJECTED\nI0205 18:34:18.707461 1 [http_server.cc:3570] [request id: <id_unknown>] Infer failed: [request id: <id_unknown>] input 'nothing_wrong"
+        
+        input_data = numpy.random.randn(1, 3).astype(numpy.float32)
+        input_tensor = httpclient.InferInput(input_name, input_data.shape, "FP32")
+        input_tensor.set_data_from_numpy(input_data)
+
+        triton_client.infer(model_name="simple", inputs=[input_tensor])
```
rmccorm4, line 55: Should this be verifying some escaping in server log afterwards?

qa/L0_logging/log_injection_test.py
tanmayv25, line 1: How is this test case invoked? I was expecting some changes in test.sh to call out `log_injection_test.py`.

qa/L0_logging/log_injection_test.py
nnshah1: actual tests coming ...

qa/L0_logging/log_injection_test.py
nnshah1: actual tests coming ...

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,293 @@
...
+# Sample log entry
+# log_entry = "#I0516 17:33:33.476093 2814 cache_manager.cc:480] \"Create CacheManager with cache_dir: '/opt/tritonserver/caches'\""
+
+
+# Regular expression pattern to capture the headers and rows
+table_regex = re.compile(
+    r'\+[-+]+\+\n'  # Match the top border
+    r'\| (?P<header>.*?) \|\n'  # Capture the header
+    r'\+[-+]+\+\n'  # Match the header border
+    r'(?P<rows>(?:\| .*? \|\n)*)'  # Capture the rows
```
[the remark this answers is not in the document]
rmccorm4: :eyes:

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,271 @@
...
+            validate_table(match)
+        else:
+            google.protobuf.text_format.Parse(obj,grpcclient.model_config_pb2.ModelConfig())
+
+class TestLogFormat:
+    @pytest.fixture(autouse=True)
+    def setup(self, request):
+        test_case_name = request.node.name
+        self._server_options = {}
+        self._server_options["log-verbose"] = 256
```
rmccorm4, line 168: Is this `256` supposed to be testing something about the `Log::Level` being a `uint8_t`?

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,271 @@
...
+        self._server_options["log-format"] = log_format.replace("_unescaped", "")
+        self._launch_server(unescaped=True if "_unescaped" in log_format else False)
+        time.sleep(1)
+        self._server_process.kill()
+        return_code = self._server_process.wait()
+        if isinstance(format_regex, str):
+            return
+        self.verify_log_format(self._server_options["log-file"], format_regex)
+
+    def foo_test_injection(self):
```
rmccorm4, line 256:
```suggestion
    # FIXME: PLACEHOLDER
    def foo_test_injection(self):
```

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,271 @@
...
+        ids=IDS,
+    )
+    def test_log_format(self, log_format, format_regex):
+        self._server_options["log-format"] = log_format.replace("_unescaped", "")
+        self._launch_server(unescaped=True if "_unescaped" in log_format else False)
+        time.sleep(1)
+        self._server_process.kill()
+        return_code = self._server_process.wait()
+        if isinstance(format_regex, str):
+            return
```
rmccorm4, line 281: I didn't understand this at a glance, but I'm assuming this is checking whether it's a regex or an empty string `""` from your definition of FORMATS at the top? 

Can it just be something like this then?
```python
if format_regex:
  self.verify_log_format(self._server_options["log-file"], format_regex)
```

And are we testing anything for the cases where `log_format` is not`"default"`?

qa/L0_logging/log_format_test.py
nnshah1: we don't have a defined max for log verbose - it is currently defined as a uint32_t - though I thought about changing it to a uint8_t.

this is distinct from log level, this the verbosity level

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,271 @@
...
+        if match:
+            for field, value in match.groupdict().items():
+                if field in validators:
+                    try:
+                        validators[field](value)
+                    except Exception as e:
+                        raise type(e)(f"{e}\nInvalid {field}: '{match.group(field)}' in log record '{record}'")
+
+        else:
+            raise Exception("Invalid log line")
```
rmccorm4: Can we invert some checks to avoid the deep nesting here? Something like this?
```python
    def validate_log_record(self, record, format_regex):
        match = format_regex.search(record)
        if not match:
            raise Exception("Invalid log line")

        for field, value in match.groupdict().items():
            if field not in validators:
                continue

            try:
                validators[field](value)
            except Exception as e:
                raise type(e)(f"{e}\nInvalid {field}: '{match.group(field)}' in log record '{record}'")
```

qa/L0_logging/log_format_test.py
nnshah1: this is still place holder to test out the parameterization - will remove once have the regex for ISO in place

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,271 @@
+#!/usr/bin/python
```
rmccorm4, line 1: Can you run the pre-commit hooks to avoid the linting check errors?

qa/L0_logging/log_format_test.py
rmccorm4: I see, just looked like magic number to me so was curious

nnshah1: open to other thoughts here - could do max unsigned int -

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,470 @@
...
+            )
+    elif heading:
+        with pytest.raises(json.JSONDecodeError):
+            json.loads(heading)
+    if obj:
+        match = table_border_regex.search(obj[0])
+        if match:
+            validate_table(obj)
+        elif escaped:
+            validate_protobuf(obj)
```
rmccorm4, line 301: Deferred to follow-up PR, but would be nice to have some comments explaining the validations that the only acceptable multi-line logs will be tables/protobuf (perhaps json?), but will otherwise be restricted to single-line logs if escaped correctly.

qa/L0_pinned_memory/test.sh
```diff
@@ -148,7 +155,7 @@ for BACKEND in $BACKENDS; do
         RET=1
     fi
 
-    grep "] pinned" ${ENSEMBLE_NAME}.nonpinned.server.log
+    grep "] \"Pinned memory pool is created" ${ENSEMBLE_NAME}.nonpinned.server.log
```
rmccorm4, line 158: Note: This is checking that the following line is **not** found in the log. This was previously "passing" because "pinned" is never found, this is updating the check to be more correct to look for the correct pinned memory message, and assert it's not found when disabled.

qa/L0_model_update/instance_update_test.py
```diff
@@ -510,11 +510,11 @@ def test_instance_resource_decrease(self):
             if os.environ["RATE_LIMIT_MODE"] == "execution_count":
                 # Make sure the previous max resource limit of 4 is reduced to 3
                 # when no explicit limit is set.
-                self.assertIn("Resource: R1\t Count: 3", f.read())
+                self.assertIn("Resource: R1\\t Count: 3", f.read())
             else:
                 # Make sure the max resource limit is never set to 3 when
                 # explicit limit of 10 is set.
-                self.assertNotIn("Resource: R1\t Count: 3", f.read())
+                self.assertNotIn("Resource: R1\\t Count: 3", f.read())
```
rmccorm4, line 517: Note for possible follow-up, if these end up poorly escaped, maybe we can consider embedding some something like 4 spaces "    " literally in the string rather than using `\t`.

nnshah1: was also thinking these could be a table - or a json instead

qa/L0_logging/test.sh
```diff
@@ -588,12 +588,24 @@ set -e
...
 
+set +e
+
+FORMAT_TEST_LOG="./log_format_test.log"
+
+python3 -m pytest --junitxml=log_format_test.xml log_format_test.py > $FORMAT_TEST_LOG 2>&1
+
+if [ $? -ne 0 ]; then
+    cat $FORMAT_TEST_LOG
+    echo -e "\n***\n*** Test Failed\n***"
```
rmccorm4, line 599: Can be follow-up but should specify which subtest failed for easier debugging and automation of parsing log errors, ex:
```suggestion
    echo -e "\n***\n*** Log Format Test Failed\n***"
```

nnshah1: less of an issue here as they are also picked up in CI as individual xunit style tests

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,470 @@
...
+            test_logs_directory, test_case_name + ".server.log"
+        )
+
+    def _shutdown_server(self):
+        if self._server_process:
+            self._server_process.kill()
+            self._server_process.wait()
+
+    def _launch_server(self, escaped=None):
+        cmd = ["tritonserver"]
```
rmccorm4, line 332: follow-up, why bash process in python? Why not one of the following:
1. Start server in `test.sh` like other tests
2. Use in-process python API within python test
3. Do it this way with subprocess in python, but move it into a common util we can start to re-use in other tests if reducing bash usage.

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,470 @@
...
+import google.protobuf.text_format
+import numpy
+import pytest
+import tritonclient.grpc as grpcclient
+import tritonclient.http as httpclient
+from tritonclient.utils import InferenceServerException
+
+module_directory = os.path.split(os.path.abspath(__file__))[0]
+
+test_model_directory = os.path.abspath(os.path.join(module_directory, "log_models"))
```
rmccorm4, line 45: follow-up: unused?

qa/L0_logging/log_format_test.py
```diff
@@ -0,0 +1,470 @@
...
+    def _setup(self, request):
+        test_case_name = request.node.name
+        self._server_options = {}
+        self._server_options["log-verbose"] = INT32_MAX
+        self._server_options["log-info"] = 1
+        self._server_options["log-error"] = 1
+        self._server_options["log-warning"] = 1
+        self._server_options["log-format"] = "default"
+        self._server_options["model-repository"] = os.path.abspath(
+            os.path.join(module_directory, "log_models")
```
rmccorm4, line 319: follow-up: Use `test_model_directory` defined at top?

qa/L0_logging/log_format_test.py
rmccorm4: I'm surprised the codeql scanner didn't complain about this

qa/L0_logging/log_format_test.py
nnshah1: I'm thinking #3 - reason to do it here was to easily name the logs to match the test cases - and to also control escaping / non escaping via environment variable. Would have been difficult to reuse the start server / seperate out the cases there.

in process api is an option - but log injection was specifically a network issue - so wanted to also stay true to the original intent - I haven't tried to verify if the same holds true if accessing the in process api directly

rmccorm4 (approved):
Functionally LGTM - nits/suggestions for future follow-up

PR #7224, opened 2024-05-15
Remove the dependency on CUDA driver

krishung5:
This PR fixed the CUDA driver dependency issue on CPU-only system introduced in the previous PR https://github.com/triton-inference-server/server/pull/7178.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -180,7 +180,6 @@ if(${TRITON_ENABLE_GPU})
     main
     PRIVATE
       CUDA::cudart
-      -lcuda
   )
 endif() # TRITON_ENABLE_GPU
 
--- a/src/shared_memory_manager.cc
+++ b/src/shared_memory_manager.cc
@@ -118,7 +118,6 @@ SharedMemoryManager::UnregisterHelper(
 }
 }}  // namespace triton::server
 #else
-#include <dlfcn.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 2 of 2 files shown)


src/shared_memory_manager.cc
```diff
@@ -266,14 +266,38 @@ OpenCudaIPCRegion(
...
+  }
+  return nullptr;
+}
+
 TRITONSERVER_Error*
 GetCudaSharedMemoryRegionSize(CUdeviceptr data_ptr, size_t& shm_region_size)
 {
+  void* cu_mem_get_address_range = nullptr;
+  void* cu_get_error_string = nullptr;
+  RETURN_IF_ERR(GetCudaDriverEntryPoint(
```
tanmayv25, line 288: Can you add a comment here that why we are making this call? 
Basically, why linking against the cuda driver directly is not an option for Triton.

src/CMakeLists.txt
```diff
@@ -180,7 +180,6 @@ if(${TRITON_ENABLE_GPU})
     main
     PRIVATE
       CUDA::cudart
-      -lcuda
```
mc-nv, line 183: Thank you for removing this flag.

mc-nv: LGTM

src/shared_memory_manager.cc
krishung5: Added the comment.

tanmayv25 (approved):

PR #7234, opened 2024-05-17
Add docs section on ensemble model caching

rmccorm4:


lkomali:
https://jirasw.nvidia.com/browse/DLIS-6258
Response cache currently doesn't support CPU-GPU input/output mismatch between composing models. 
This should be added to the limitations too.

docs/user_guide/response_cache.md
```diff
@@ -221,6 +221,30 @@ fast/cheap (the model is not compute-bound), the cache can negatively impact
 the overall performance due to the overhead of managing and communicating with
 the cache.
 
+## Ensemble Model Caching
```
rmccorm4, line 224: > https://jirasw.nvidia.com/browse/DLIS-6258
> Response cache currently doesn't support CPU-GPU input/output mismatch between composing models.
> This should be added to the limitations too.

@lkomali this is loosely covered by the first two existing limitations:

> - Only input tensors located in CPU memory will be hashable for accessing the
  cache. If an inference request contains input tensors not in CPU memory, the
  request will not be hashed and therefore the response will not be cached.
> - Only responses with all output tensors located in CPU memory will be eligible
  for caching. If any output tensor in a response is not located in CPU memory,
  the response will not be cached.

I think we can leave it as-is until we see some confusion if any. It's a little difficult to eloquently explain the subtle differences in the docs.

tanmayv25 (approved):

lkomali (approved):
LGTM

PR #7239, opened 2024-05-17
Fix the L0_simple_go_client

tanmayv25:
The client should run at the same location as go.mod.
This is needed because go has moved to module only execution system.

Client PR: https://github.com/triton-inference-server/client/pull/663

### DIFF
--- a/qa/L0_simple_go_client/test.sh
+++ b/qa/L0_simple_go_client/test.sh
@@ -31,7 +31,6 @@ TRITON_REPO_ORGANIZATION=${TRITON_REPO_ORGANIZATION:="http://github.com/triton-i
 TRITON_COMMON_REPO_TAG=${TRITON_COMMON_REPO_TAG:="main"}
 
 GO_CLIENT_DIR=client/src/grpc_generated/go
-SIMPLE_GO_CLIENT=${GO_CLIENT_DIR}/grpc_simple_client.go
 
 SERVER=/opt/tritonserver/bin/tritonserver
 SERVER_ARGS=--model-repository=`pwd`/models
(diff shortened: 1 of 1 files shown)


qa/L0_simple_go_client/test.sh
```diff
@@ -55,26 +55,23 @@ git clone ${TRITON_REPO_ORGANIZATION}/client.git
 go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
```
rmccorm4, line 55: I'm assuming this `@latest` is why this test is breaking or changing behavior every so often

qa/L0_simple_go_client/test.sh
```diff
@@ -55,26 +55,23 @@ git clone ${TRITON_REPO_ORGANIZATION}/client.git
...
-rm -rf ${PACKAGE_PATH}/client
-mkdir -p ${PACKAGE_PATH}
-cp -r client $PACKAGE_PATH
 
 set +e
 
-# Run test for GRPC variant of go client
-GO111MODULE=off go run $SIMPLE_GO_CLIENT >>client.log 2>&1
+# Run test for GRPC variant of go client within go.mod path
+go run grpc_simple_client.go >>client.log 2>&1
```
rmccorm4, line 66: Remove `${SIMPLE_GO_CLIENT}` up top?

qa/L0_simple_go_client/test.sh
tanmayv25: I think the reason was because we upgraded the go version itself. We could have searched for the `@latest` packages and fixed them to a version. But instead have modified the other test scripts like L0_grpc to work with newer behavior. 

go 1.22 has introduced various changes and I think it is ok to upgrade however, we will try fixing the version from future.

rmccorm4 (approved):

PR #7254, opened 2024-05-21
Add response sender test base

kthui:
Set up the base for response sender testing, which enables testing varies ways a model may return response(s), including improper use of response sender.
Currently, this test only tests existing behaviors of decoupled and non-decoupled models, before any work is done on their data pipelines or response sender.

The control of testing is from the test script by setting the input tensors:
* number_of_response_before_return
* send_complete_final_flag_before_return
* return_a_response
* number_of_response_after_return
* send_complete_final_flag_after_return

There are definite some improvements possible on the test base, i.e. reporting the size of requests, allow the final flag to be sent with the last response or separately, add support for dynamic batching requests. but I think those can be added (or already added) in next PRs when needed.

Next PRs:
* https://github.com/triton-inference-server/server/pull/7257
* https://github.com/triton-inference-server/python_backend/pull/360


qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -0,0 +1,293 @@
...
+            if response_id < 1000:
+                self.assertFalse(
+                    response_returned,
+                    "Expect at most one response returned per request.",
+                )
+                response_returned = True
+            elif response_id < 2000:
+                before_return_response_count += 1
+            else:  # response_id < 3000
+                after_return_response_count += 1
```
rmccorm4, line 126: Can you explicitly fail for an unexpected response_id for future test change sanity checks?

```python
if response_id < 1000:
    # ...
elif response_id < 2000:
    # ...
elif response_id < 3000:
    # ...
else: 
    raise ValueError(f"Unexpected response_id: {response_id}")
```

qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def _infer(
+        self,
+        model_name,
+        number_of_response_before_return,
+        send_complete_final_flag_before_return,
+        return_a_response,
+        number_of_response_after_return,
+        send_complete_final_flag_after_return,
+    ):
```
rmccorm4, line 86: Would it make sense to have a sanity check to catch user error here with something like this:

```python
    # Assert that one and only one of these values is True, as they are mutually exclusive
    self.assertEqual([send_complete_final_flag_before_return, return_a_response, send_complete_final_flag_after_return].count(True), 1)
```

qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -0,0 +1,293 @@
...
+            number_of_response_before_return=1,
+            send_complete_final_flag_before_return=False,
+            return_a_response=False,
+            number_of_response_after_return=3,
+            send_complete_final_flag_after_return=True,
+        )
+
+    # Non-decoupled model send 1 response on return.
+    def test_non_decoupled_one_response_on_return(self):
+        self._assert_non_decoupled_infer_success(
```
rmccorm4, line 283: All these tests are looking for successes. Do we have any test cases to assert failures happen?

qa/L0_backend_python/test.sh
```diff
@@ -414,7 +414,7 @@ if [ "$TEST_JETSON" == "0" ]; then
     # [DLIS-5970] Disable io tests for Windows since GPU Tensors are not supported
     # [DLIS-6122] Disable model_control & request_rescheduling tests for Windows since they require load/unload
     if [[ ${TEST_WINDOWS} == 0 ]]; then
-        SUBTESTS+=" variants io python_based_backends async_execute"
+        SUBTESTS+=" variants io python_based_backends async_execute response_sender"
```
rmccorm4, line 417: Is there anything specific about this test that won't be supported on windows?

qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -0,0 +1,293 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 1: Side note: Can we target these test changes into `main` so there is less to review in eventual merge to main if it tests existing features? This way we can also see if you make any changes to these tests in future work.

rmccorm4 (commented):
High level overview of tests look good to me, left some comments.

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: I intentionally left this open, because then we can test improper model behavior using this testing base, for example, sending a complete final flag and then returning a response.

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: Not added yet, they will be added later in the series of PRs.

qa/L0_backend_python/test.sh
kthui: I don't see it now. I think we can move it up and if there is a failure then move it back here?

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: Sure. @Tabrizian are you also ok merging this directly into main?

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: Good idea! Updated.

qa/python_models/response_sender/model_common.py
```diff
@@ -0,0 +1,185 @@
...
+        self._pb_utils = pb_utils
+        self._background_tasks = set()
+
+    def _get_instructions_from_request(self, request):
+        instr = {}
+        return_a_response_np = self._pb_utils.get_input_tensor_by_name(
+            request, "RETURN_A_RESPONSE"
+        ).as_numpy()
+        instr["batch_size"] = return_a_response_np.shape[0]
+        instr["return_a_response"] = bool(return_a_response_np.sum())
```
Tabrizian, line 45: Is the idea behind `bool(...sum())` that even if one of the values in the batch is larger than 0, we would consider it true?

qa/python_models/response_sender/model_common.py
```diff
@@ -0,0 +1,185 @@
...
+
+import numpy as np
+
+
+class ResponseSenderModelCommon:
+    def __init__(self, pb_utils):
+        self._pb_utils = pb_utils
+        self._background_tasks = set()
+
+    def _get_instructions_from_request(self, request):
```
Tabrizian, line 39:
```suggestion
    def _get_instructions_from_request(self, request):
     """Determine the execution instructions from the inputs. This test tries
    to examine all the corner cases with using response sender. It uses the
    following inputs to control the model behaviour:
    
    Assumptions: The request batch size can be larger than one.
    
    There are 5 inputs in the model that control the model behaviour:
    
    * NUMBER_OF_RESPONSE_BEFORE_RETURN(UINT8): Determines the number of responses before returning from execute function.
    * SEND_COMPLETE_FINAL_FLAG_BEFORE_RETURN(BOOL): Determines whether the final flag will be sent before return.
    * RETURN_A_RESPONSE(BOOL): Return the response when the model is returning from `execute` function.
    * NUMBER_OF_RESPONSE_AFTER_RETURN(UINT8): Determines the number of responses after return.
    * SEND_COMPLETE_FINAL_FLAG_AFTER_RETURN(BOOL): Determines whether the final flag will be sent before return.
    
    Note that if the batch size is larger than one the sum of the values in the batch size will be used for determining the value of each input. The response_id is used to determine the difference between responses
    sent before the execute vs the responses sent after the execute.
    """

```

qa/L0_backend_python/response_sender/response_sender_test.py
Tabrizian: Let's keep it in the side branch. I don't think it is testing existing features. If it is merged to `main` it will fail since you can't use `response_sender` in non-decoupled mode.

qa/python_models/response_sender/model_common.py
kthui: Yes, because if there is a batched tensor, there is only one output slot for the tensor on the list that is returned. I think for the purpose of this test, we only need to test if a return has taken place or not on a particular slot, and the content of the returned tensor does not matter, so I basically just picked the simplest way of reducing a list of `bool` into one `bool` - if there is one or more true, we consider it true; if all false, we consider it false.

qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -0,0 +1,295 @@
...
+            send_complete_final_flag_before_return,
+            return_a_response,
+            number_of_response_after_return,
+            send_complete_final_flag_after_return,
+        )
+        callback, responses = self._generate_streaming_callback_and_responses_pair()
+        with grpcclient.InferenceServerClient("localhost:8001") as client:
+            client.start_stream(callback)
+            client.async_stream_infer(model_name, inputs)
+            time.sleep(2)  # to collect all responses
```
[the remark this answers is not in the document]
kthui: Nice suggestion! `stop_stream` indeed wait for all responses to arrive. Removing the wait significantly improved testing speed when the number of sub-test increased.

Removing the wait also revealed a hidden issue, will discuss more on the [future PR](https://github.com/triton-inference-server/server/pull/7292).

qa/python_models/response_sender/model_common.py
kthui: Nice summary! Clarified some statements based on my understanding and added to this PR.

qa/python_models/response_sender/model_common.py
Tabrizian: sounds good, thanks for the explanation.

Tabrizian (approved):

PR #7257, opened 2024-05-22
Add response sender test for new behavior

kthui:
Previous PR: https://github.com/triton-inference-server/server/pull/7254
Related PR: https://github.com/triton-inference-server/python_backend/pull/360

Add the following new tests:
* Non-decoupled model sending a response before / after returning.
* Decoupled / non-decoupled model handling batched requests each using a different style of returning response(s).
* Assert failure on decoupled model returning a response.

Next PR: https://github.com/triton-inference-server/python_backend/pull/361

### DIFF
--- a/qa/L0_backend_python/response_sender/response_sender_test.py
+++ b/qa/L0_backend_python/response_sender/response_sender_test.py
@@ -288,6 +288,26 @@ class ResponseSenderTest(unittest.TestCase):
             send_complete_final_flag_after_return=False,
         )
 
+    # Non-decoupled model send 1 response before return.
+    def test_non_decoupled_one_response_pre_return(self):
+        self._assert_non_decoupled_infer_success(
+            number_of_response_before_return=1,
+            send_complete_final_flag_before_return=True,
+            return_a_response=False,
+            number_of_response_after_return=0,
+            send_complete_final_flag_after_return=False,
+        )
+
+    # Non-decoupled model send 1 response after return.
+    def test_non_decoupled_one_response_post_return(self):
+        self._assert_non_decoupled_infer_success(
+            number_of_response_before_return=0,
+            send_complete_final_flag_before_return=False,
+            return_a_response=False,
+            number_of_response_after_return=1,
+            send_complete_final_flag_after_return=True,
+        )
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/qa/L0_backend_python/response_sender/response_sender_test.py
+++ b/qa/L0_backend_python/response_sender/response_sender_test.py
@@ -32,6 +32,63 @@ from tritonclient.utils import InferenceServerException
 
 
 class ResponseSenderTest(unittest.TestCase):
+    _inputs_parameters_zero_response_pre_return = {
+        "number_of_response_before_return": 0,
+        "send_complete_final_flag_before_return": True,
+        "return_a_response": False,
+        "number_of_response_after_return": 0,
+        "send_complete_final_flag_after_return": False,
+    }
+    _inputs_parameters_zero_response_post_return = {
+        "number_of_response_before_return": 0,
+        "send_complete_final_flag_before_return": False,
+        "return_a_response": False,
+        "number_of_response_after_return": 0,
+        "send_complete_final_flag_after_return": True,
+    }
+    _inputs_parameters_one_response_pre_return = {
+        "number_of_response_before_return": 1,
+        "send_complete_final_flag_before_return": True,
+        "return_a_response": False,
+        "number_of_response_after_return": 0,
+        "send_complete_final_flag_after_return": False,
+    }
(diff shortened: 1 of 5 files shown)


qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -67,6 +68,94 @@ def _get_inputs(
...
+                    send_complete_final_flag_before_return=False,
+                    return_a_response=False,
+                    number_of_response_after_return=1,
+                    send_complete_final_flag_after_return=True,
+                ),
+            ],
+            "number_of_response_before_return": 1,
+            "return_a_response": 1,
+            "number_of_response_after_return": 1,
+        }
```
Tabrizian, line 157: How does the items in higher level dictionary interact with `parallel_inputs`? I couldn't find a place where `number_of_response_before_return` in the high-level dictionary is being used.

kthui: > How does the items in higher level dictionary interact with parallel_inputs?

Each `inputs` on `parallel_inputs` are sent in parallel to each other, so they can be grouped into a `requests` on the backend by the dynamic batcher. They are consumed by `_infer_parallel()` [here](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR171-R172).

> I couldn't find a place where number_of_response_before_return in the high-level dictionary is being used.

Those three variables after `parallel_inputs` are for checking if the responses are valid, for example [here](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR439-R444).

The idea behind those two [get parallel inputs functions](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR71-R157) is to support the two [multiple requests test cases](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR412-R454). The 10 inputs on the two get parallel inputs functions are from the [first 10 tests](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR312-R410) for valid response interactions. I think there is definitely some grouping can be done to avoid repeating those 10 inputs, but I also don't want to add another layer of abstraction which complicates things. What do you think?

Tabrizian (approved):

qa/L0_backend_python/response_sender/response_sender_test.py
Tabrizian: Makes sense. Let's keep it the away it is. Thanks for explaining.

kthui: I grouped the inputs into a class variables, for example
```
  _inputs_parameters_zero_response_pre_return = {
      "number_of_response_before_return": 0,
      "send_complete_final_flag_before_return": True,
      "return_a_response": False,
      "number_of_response_after_return": 0,
      "send_complete_final_flag_after_return": False,
  }
```
When those five values are needed, the test can simply refer to the class variable, for example
```
    # Decoupled model send response final flag before request return.
    def test_decoupled_zero_response_pre_return(self):
        self._assert_decoupled_infer_success(
            **self._inputs_parameters_zero_response_pre_return
        )
```
Base on this change, the purpose of those two [get parallel inputs functions](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR71-R157) are reduced to, for example
```
        parallel_inputs = [
            self._get_inputs(**self._inputs_parameters_one_response_on_return),
            self._get_inputs(**self._inputs_parameters_one_response_pre_return),
            self._get_inputs(**self._inputs_parameters_one_response_post_return),
        ]
        expected_number_of_response_before_return = 1
        expected_return_a_response = True
        expected_number_of_response_after_return = 1
```
and can fit comfortably inside the two `test_*_multiple_requests()` test cases. Thus, the those two [get parallel inputs functions](https://github.com/triton-inference-server/server/pull/7257/files#diff-5105d2786bb8fe3edddf799d95cee554f37b16924d8a3cd3c80ad14e79cfad9cR71-R157) are removed.

Sorry for the last minute change, but I hope this will make the test more manageable in the future.

kthui:
Rebasing onto https://github.com/triton-inference-server/server/tree/jacky-res-sender-main for merging.

Tabrizian (approved):

PR #7261, opened 2024-05-23
Added new flag for GPU peer access API control

indrajit96:
Added a new flag "--enable-peer-access" at Triton startup to control creation of CUDA context at server startup.
Jira : https://jirasw.nvidia.com/browse/DLIS-6705
Core : https://github.com/triton-inference-server/core/pull/361


src/command_line_parser.cc
```diff
@@ -373,7 +373,8 @@ enum TritonOptionId {
   OPTION_BACKEND_CONFIG,
   OPTION_HOST_POLICY,
   OPTION_MODEL_LOAD_GPU_LIMIT,
-  OPTION_MODEL_NAMESPACING
+  OPTION_MODEL_NAMESPACING,
+  OPTION_PEER_ACCESS
```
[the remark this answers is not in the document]
Tabrizian: Looks like we don't do it right now, although we do it for other build options (e.g. tracing, http, etc).

Tabrizian, line 377:
```suggestion
  OPTION_ENABLE_PEER_ACCESS
```

qa/L0_trace/test.sh
```diff
@@ -777,6 +777,7 @@ SERVER_ARGS="--allow-sagemaker=true --model-control-mode=explicit \
                 --load-model=simple --load-model=ensemble_add_sub_int32_int32_int32 \
                 --load-model=repeat_int32 \
                 --load-model=input_all_required \
+                --load-model=dynamic_batch \
```
Tabrizian, line 780: Were we missing this model before? Is it correctly added?

src/command_line_parser.cc
```diff
@@ -461,6 +462,11 @@ TritonParser::SetupOptions()
       {OPTION_MODEL_NAMESPACING, "model-namespacing", Option::ArgBool,
        "Whether model namespacing is enable or not. If true, models with the "
        "same name can be served if they are in different namespace."});
+  model_repo_options_.push_back(
+      {OPTION_PEER_ACCESS, "enable-peer-access", Option::ArgBool,
+       "Whether GPU peer access is enable or not. If true, GPUs can share "
+       "virtual memory address,"
+       "adds cuda context "});
```
Tabrizian, line 469:
```suggestion
       "Whether the server tries to enable peer access or not. Even when this options is set to true,  "
       "peer access could still be not enabled because the underlying system doesn't support it."
       " The server will log a warning in this case. Default is true."});
```

qa/L0_trace/test.sh
indrajit96: Yes this is a breakage cause by one of my changes in L0_trace.
Tested the pipeline for this change fixes currently failing L0_trace

Tabrizian (approved):

Tabrizian (approved):

PR #7265, opened 2024-05-24
Fix gRPC streaming non-decoupled segfault if sending response and final flag separately

kthui:
When using gRPC streaming for non-decoupled models, the frontend does not properly handle the case which the response and final flag are sent separately. The gRPC server will release the state as soon as the response is received, resulting in a possible segfault when the final flag is later received.

The fix is to defer sending the response if a final flag is not coupled with the response, and leave the gRPC step in WRITEREADY. Then, when the final flag is received, it continue sending the response and move to step WRITTEN.

This PR should merge before https://github.com/triton-inference-server/server/pull/7311

### DIFF
--- a/qa/L0_grpc_state_cleanup/cleanup_test.py
+++ b/qa/L0_grpc_state_cleanup/cleanup_test.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
--- a/qa/L0_grpc_state_cleanup/cleanup_test.py
+++ b/qa/L0_grpc_state_cleanup/cleanup_test.py
@@ -70,7 +70,7 @@ class CleanUpTest(tu.TestResultCollector):
         self.repeat_non_decoupled_model_name = "repeat_int32_non_decoupled"
 
     def _prepare_inputs_and_outputs(self, kind):
-        if kind == "decoupled_streaming" or kind == "non_decoupled_streaming":
+        if kind in ("decoupled_streaming", "non_decoupled_streaming"):
             self.inputs_ = []
             self.inputs_.append(grpcclient.InferInput("IN", [1], "INT32"))
             self.inputs_.append(grpcclient.InferInput("DELAY", [1], "UINT32"))
@@ -80,7 +80,7 @@ class CleanUpTest(tu.TestResultCollector):
             self.outputs_.append(grpcclient.InferRequestedOutput("OUT"))
             self.outputs_.append(grpcclient.InferRequestedOutput("IDX"))
             self.requested_outputs_ = self.outputs_
-        elif kind == "simple" or kind == "streaming":
+        elif kind in ("simple", "streaming"):
             self.inputs_ = []
             self.inputs_.append(grpcclient.InferInput("INPUT0", [1, 1], "FP32"))
 
--- a/qa/L0_grpc_state_cleanup/cleanup_test.py
+++ b/qa/L0_grpc_state_cleanup/cleanup_test.py
@@ -67,9 +67,10 @@ class CleanUpTest(tu.TestResultCollector):
     def setUp(self):
         self.decoupled_model_name_ = "repeat_int32"
         self.identity_model_name_ = "custom_zero_1_float32"
+        self.repeat_non_decoupled_model_name = "repeat_int32_non_decoupled"
 
     def _prepare_inputs_and_outputs(self, kind):
-        if kind == "decoupled_streaming":
+        if kind == "decoupled_streaming" or kind == "non_decoupled_streaming":
             self.inputs_ = []
             self.inputs_.append(grpcclient.InferInput("IN", [1], "INT32"))
             self.inputs_.append(grpcclient.InferInput("DELAY", [1], "UINT32"))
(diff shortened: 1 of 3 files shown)


kthui:
***Note: Previous PRs are not yet merged, so there will be additional non-related changes. The real changes are the last 3 commits, and mainly [this commit](https://github.com/triton-inference-server/server/commit/9722ce13b007c0bcef29ba0b8a086bd1eb2bba28) should be concerned.***

qa/L0_backend_python/test.sh
```diff
@@ -409,12 +409,12 @@ fi
 # Disable variants test for Jetson since already built without GPU Tensor support
 # Disable decoupled test because it uses GPU tensors
 if [ "$TEST_JETSON" == "0" ]; then
-    SUBTESTS="ensemble bls decoupled"
+    SUBTESTS="ensemble decoupled"
```
oandreeva-nv, line 412: this is just a temporary removal, right?

kthui: Yes, because there are still some complication in the bls test that will fail the entire L0_backend_python test on the pipeline.

Tabrizian (commented):
Can we target `main` for this change? It looks like it is independent of other Python be changes.

qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -0,0 +1,473 @@
...
+import time
+import unittest
+
+import numpy as np
+import tritonclient.grpc as grpcclient
+from tritonclient.utils import InferenceServerException
+
+
+class ResponseSenderTest(unittest.TestCase):
+    def _get_inputs(
```
oandreeva-nv, line 36: May I kindly ask to provide a brief decryption for helper functions?

oandreeva-nv: Not all of them, e.g. `_assert*` and `_infer` are self-explanatory

src/grpc/stream_infer_handler.cc
```diff
@@ -573,17 +573,19 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
-  // Log appropriate errors
   bool is_complete =
       state->complete_ || (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) != 0;
+
+  // Log appropriate messages
   if (!state->is_decoupled_) {
     if (!is_complete) {
-      LOG_ERROR << "[INTERNAL] ModelStreamInfer received a response without "
-                   "FINAL flag for a model with one-to-one transaction";
+      LOG_VERBOSE(2)
```
oandreeva-nv, line 582: Why did we change `LOG_ERROR` to `LOG_VERBOSE(2)` ?

rmccorm4: It's no longer an error to receive response and final_flag as two separate responses for non-decoupled models, but assuming we're keeping the log for some visibility or test log grepping.

src/grpc/stream_infer_handler.cc
```diff
@@ -745,7 +747,9 @@ ModelStreamInferHandler::StreamInferResponseComplete(
       }
     } else {
       state->step_ = Steps::WRITEREADY;
-      state->context_->WriteResponseIfReady(state);
+      if (is_complete) {
+        state->context_->WriteResponseIfReady(state);
```
rmccorm4, line 751: Can we add a test for a bad user-error case where the model sends two non-null response objects for a non-decoupled model? I'm not sure this is handled here. Would be good to gracefully handle it and raise an error rather than crash or overwrite state.

src/grpc/stream_infer_handler.cc
kthui: Currently we "require" the final flag to be sent with the final/only response for non-decoupled streaming, but this is not an error because the model may send the response first and then the final flag. I change it to `LOG_VERBOSE(2)` because it is not an error, but kept the log for debugging purpose.

tanmayv25: How would frontend handle cases where multiple responses are sent in a non-decoupled case? 
I would suggest changing the message to something more generic if we are keeping it for debugging purpose. For example enhance the verbose message printed out in L564.

kthui:
All non-related commits are removed and the PR is now targeting the main branch.

Tabrizian (dismissed):

src/grpc/stream_infer_handler.cc
kthui: > How would frontend handle cases where multiple responses are sent in a non-decoupled case?

I think the extra responses go into the `response_queue_`, but only the [first response will be written](https://github.com/triton-inference-server/server/blob/main/src/grpc/infer_handler.h#L921). Then, the extra responses are discarded when the [state resets](https://github.com/triton-inference-server/server/blob/main/src/grpc/infer_handler.h#L1048).

kthui: > I would suggest changing the message to something more generic if we are keeping it for debugging purpose. For example enhance the verbose message printed out in L564.

Good idea! Updated: [Enhance logging for non-decoupled on null response and final flag](https://github.com/triton-inference-server/server/pull/7265/commits/42a6a80af9ad6eb9b7c54fb3f7338b060b2f6010)

src/grpc/stream_infer_handler.cc
kthui: Sure. Added the test case for a non-decoupled model returning more than one response during gRPC streaming. The client should get the first response.

src/grpc/stream_infer_handler.cc
Tabrizian: Can we explicitly return an error rather ignoring the second response? It can also be part of the additional core checks that we were targeting.

Tabrizian (dismissed):

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: For `_assert*`, they are mainly checking if the result matches the one expected. For example, if the model should return one response, `_assert*` should check and fail if the model does not return one response.

For `_infer`, because there is only one model backbone behind eight models each contributing to one combination of the three threaded/async, non-decoupled/decoupled and w/wo_dynamic_batching parameters, so the `_infer` is to send an infer request to and return the response from one of the eight models targeted. Specifically, to tell the model on how response(s) should be sent.

src/grpc/stream_infer_handler.cc
kthui: This is a good idea! I don't think it is hard to implement because we should be able to remove the first response from the queue and replace it with an error response. Then, wait for the complete final flag and sent the response.

@Tabrizian @rmccorm4 Can we do it when we are working on additional core checks?
It is because ignoring the second response is the current behavior not introduced by this change. I want to keep the scope of this PR small and only acknowledge the current behavior on the test case, plus the original intent on fixing the issue where the final flag can be sent as a flag only response after the first response.

rmccorm4: If this is existing behavior then I'm good with a separate follow-up. Thanks for the explanation!

qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -67,9 +67,10 @@ class CleanUpTest(tu.TestResultCollector):
     def setUp(self):
         self.decoupled_model_name_ = "repeat_int32"
         self.identity_model_name_ = "custom_zero_1_float32"
+        self.repeat_non_decoupled_model_name = "repeat_int32_non_decoupled"
 
     def _prepare_inputs_and_outputs(self, kind):
-        if kind == "decoupled_streaming":
+        if kind == "decoupled_streaming" or kind == "non_decoupled_streaming":
```
oandreeva-nv, line 73: nit: if kind in ("...", "...")

oandreeva-nv (dismissed):
LGTM, Thanks for elaborations!

qa/L0_grpc_state_cleanup/cleanup_test.py
kthui: I'm ok either way, but use the
```python
if kind in ("kind_1", "kind_2", ...):
...
```
form because the code is shorter.

kthui: [Use in tuple instead of or](https://github.com/triton-inference-server/server/pull/7265/commits/178b65b955dc48222d701e08d6764847be0ebafc)

oandreeva-nv (approved):
LGTM!

qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -79,7 +80,7 @@ def _prepare_inputs_and_outputs(self, kind):
             self.outputs_.append(grpcclient.InferRequestedOutput("OUT"))
             self.outputs_.append(grpcclient.InferRequestedOutput("IDX"))
             self.requested_outputs_ = self.outputs_
-        elif kind == "simple" or kind == "streaming":
+        elif kind in ("simple", "streaming"):
```
rmccorm4, line 83: Just curious, what is `streaming` compared to `decoupled_streaming` and `non_decoupled_streaming`?

rmccorm4 (approved):
LGTM, just one question: https://github.com/triton-inference-server/server/pull/7265/files#r1628592664. Doesn't need to be fixed/addressed here, just curious.

qa/L0_grpc_state_cleanup/cleanup_test.py
kthui: The `decoupled_streaming` and `non_decoupled_streaming` are for the `repeat_int32` and `repeat_int32_non_decoupled` models on repeat backend. The `simple` and `streaming` are for the `custom_zero_1_float32` model on identity backend.

I think we could do some refactoring on how the inputs are prepared, i.e. for the repeat backend, we can have a function that returns a `inputs` that can be used directly on the `client.async_stream_infer()`.

PR #7288, opened 2024-05-28
Revert file copy

mc-nv:


build.py
```diff
@@ -1740,10 +1740,6 @@ def tensorrtllm_postbuild(cmake_script, repo_install_dir, tensorrtllm_be_dir):
         os.path.join(tensorrtllm_be_dir, "build", "libtriton_tensorrtllm*.so"),
         cmake_destination_dir,
     )
-    cmake_script.cp(
```
tanmayv25, line 1743: Shouldn't the post-build look something like this?
https://github.com/triton-inference-server/server/blob/r24.04/build.py#L1722-L1740

mc-nv: You are probably right, I have mess up with changes I"m currently working on for TensoRT-LLM, they have changed the names of the program.

krishung5 (approved):

tanmayv25 (dismissed):

krishung5 (approved):

tanmayv25 (approved):

PR #7258, opened 2024-05-22
Update expected error message

kthui:
Previous PR: https://github.com/triton-inference-server/python_backend/pull/361
Related PR: https://github.com/triton-inference-server/python_backend/pull/362

Update the error message the test expects, given the decoupled data pipeline reports the same error in a different mechanism, which results in the message to be slightly different.

Next PRs:
* https://github.com/triton-inference-server/server/pull/7292
* https://github.com/triton-inference-server/python_backend/pull/363

PR #7292, opened 2024-05-29
Add test for improper response sending from model

kthui:
Previous PRs:
* https://github.com/triton-inference-server/server/pull/7258
* https://github.com/triton-inference-server/python_backend/pull/362

Related PR: https://github.com/triton-inference-server/python_backend/pull/363

Add tests for decoupled models using the `response_sender` and returning a response.
Add tests for non-decoupled models sending multiple responses, using the `response_sender` and/or returning a response.

### DIFF
--- a/qa/L0_backend_python/response_sender/response_sender_test.py
+++ b/qa/L0_backend_python/response_sender/response_sender_test.py
@@ -468,16 +468,12 @@ class ResponseSenderTest(unittest.TestCase):
         #       using `py_future.result()` with error hangs on exit.
 
     # Decoupled model return 1 response and send 1 response.
-    @unittest.skip("Response factory segmentation fault, see the TODO comment.")
     def test_decoupled_one_response_on_and_post_return(self):
         # Note: The returned response will send an error response and complete final
-        #       flag. The response sender is not yet closed due to the model is still
-        #       executing. Then, sending a response and final flag will make their way
-        #       to the frontend, but the request is completed at the frontend, so they
-        #       are not sent to the client.
-        # TODO: The error response will close the response factory, but not the response
-        #       sender, which enables the model to send response and final flag on the
-        #       deleted response factory and causes a segmentation fault.
+        #       flag, and close the response sender and factory. Then, sending a
+        #       response will raise an exception. Since the exception happens after the
+        #       model returns, it cannot be caught by the stub (i.e. in a daemon
+        #       thread), so nothing will happen.
         responses = self._infer(
             model_name="response_sender_decoupled",
             **self._inputs_parameters_one_response_on_and_post_return,
--- a/qa/L0_backend_python/response_sender/response_sender_test.py
+++ b/qa/L0_backend_python/response_sender/response_sender_test.py
@@ -88,6 +88,20 @@ class ResponseSenderTest(unittest.TestCase):
         "number_of_response_after_return": 0,
         "send_complete_final_flag_after_return": False,
     }
+    _inputs_parameters_one_response_pre_and_on_return = {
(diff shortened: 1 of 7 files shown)


qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -517,12 +517,16 @@ def test_decoupled_one_response_pre_and_on_return(self):
...
 
     # Decoupled model return 1 response and send 1 response.
+    @unittest.skip("Response factory segmentation fault, see the TODO comment.")
     def test_decoupled_one_response_on_and_post_return(self):
         # Note: The returned response will send an error response and complete final
         #       flag. The response sender is not yet closed due to the model is still
         #       executing. Then, sending a response and final flag will make their way
         #       to the frontend, but the request is completed at the frontend, so they
         #       are not sent to the client.
+        # TODO: The error response will close the response factory, but not the response
```
Tabrizian, line 527: * If the user choses to send an error response will it close the response factory too or this only happens for unexpected errors (e.g. raising an exception).

     * If it only happens for exceptions, would it be possible to close the response sender for all the requests in the batch to avoid the segfault?

qa/L0_backend_python/response_sender/response_sender_test.py
```diff
@@ -538,10 +542,62 @@ def test_decoupled_one_response_on_and_post_return(self):
...
+            send_complete_final_flag_before_return,
+            return_a_response,
+            number_of_response_after_return,
+            send_complete_final_flag_after_return,
+        )
+        self._assert_responses_exception(responses, expected_message)
+
+    # Non-decoupled model send response final flag after request return.
+    @unittest.skip("Model unload will hang, see the TODO comment.")
+    def test_non_decoupled_zero_response_post_return(self):
```
Tabrizian, line 584: By "hang" do you mean the timeout counter that starts when the request is getting unloaded? If that's the case, I think it is fine since closing the response sender failed so they need to fix their model.

Tabrizian (commented):
Minor questions, fantastic test coverage! Thanks Jacky 🙇

kthui:
Rebase onto https://github.com/triton-inference-server/server/tree/jacky-res-sender-main for merging.

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: > By "hang" do you mean the timeout counter that starts when the ~~request~~ **(model?)** is getting unloaded?

Yes, the hang happens when the server is gracefully shutting down. The model should not raise an exception an any point (or at least catch the exception itself).

qa/L0_backend_python/response_sender/response_sender_test.py
kthui: > If the user choses to send an error response will it close the response factory too or this only happens for unexpected errors (e.g. raising an exception).

I think this happens only for unexpected errors (e.g. raising an exception), because the error of an error response is encapsulated within the response itself. The stub does not know/check if the response contains an error.

> If it only happens for exceptions, would it be possible to close the response sender for all the requests in the batch to avoid the segfault?

Good idea! In this case, the all response factories will be closed anyway, so it make sense to mark all response senders closed (if not already closed) to prevent any more use on the factory pointer.

kthui: Updated: [Enable test_decoupled_one_response_on_and_post_return](https://github.com/triton-inference-server/server/pull/7292/commits/f7a2e56ef1b455d0fe1696950b97a033e87a0bb7)

Tabrizian (approved):

PR #7310, opened 2024-05-31
ci: Support BF16 data type in TensorRT backend

pskiran1:
#### What does the PR do?

- Added new L0_trt_bf16_dtype job to test BF16 datatype for TensorRT backend.
- Updated model generation script to generate BF16 datatype TensorRT models.

`Note: BF16 datatype data type is supported for GPU with CC 8.0 and above only.`

#### Checklist

#### Commit Type:
- [ ] build
- [X] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/tensorrt_backend/pull/90

#### Test plan:

- CI Pipeline ID: 15487143

#### Background
https://docs.nvidia.com/deeplearning/tensorrt/support-matrix/index.html#hardware-precision-matrix

- closes GitHub issue: #xxx


qa/L0_trt_bf16_dtype/trt_bf16_dtype_test.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+    def _infer_helper(self, model_name, shape):
+        inputs = []
+        outputs = []
+        inputs.append(client.InferInput("INPUT0", shape, "BF16"))
+        inputs.append(client.InferInput("INPUT1", shape, "BF16"))
+
+        input0_data = np.ones(shape=shape).astype(np.float32)
+        input1_data = np.ones(shape=shape).astype(np.float32)
```
GuanLuo, line 53: How does this work? Isn't the data size actually 2x of the size computed by shape and data type? Would this still work with the input size validation changes? CC @rmccorm4 how BF16 on the other backend was tested?

qa/common/gen_qa_models.py
```diff
@@ -1076,6 +1076,103 @@ def create_plan_modelfile(
             )
 
 
+def create_plan_bf16_modelfile(
```
GuanLuo, line 1079: Can this just be integrated into existing `create_plan_xxxfile()` functions? Any code path that is differed if dtype is bf16?

docs/user_guide/model_configuration.md
```diff
@@ -452,7 +452,7 @@ library.
 |TYPE_FP32     | kFLOAT       |DT_FLOAT      |FLOAT         |kFloat   |FP32     |float32       |
 |TYPE_FP64     |              |DT_DOUBLE     |DOUBLE        |kDouble  |FP64     |float64       |
 |TYPE_STRING   |              |DT_STRING     |STRING        |         |BYTES    |dtype(object) |
-|TYPE_BF16     |              |              |              |         |BF16     |              |
+|TYPE_BF16     | kBF16        |              |              |         |BF16     |              |
```
tanmayv25, line 455: TensorRT backend also supports INT64 now. Let's add that in L450.

qa/common/gen_qa_models.py
```diff
@@ -2477,7 +2559,7 @@ def create_fixed_models(
         from tensorflow.python.framework import graph_io
 
         tf.compat.v1.disable_eager_execution()
-    if FLAGS.tensorrt:
+    if FLAGS.tensorrt or FLAGS.tensorrt_bf16:
```
tanmayv25, line 2562: Why add additional flag for building BF16 models?

qa/common/gen_qa_models.py
```diff
@@ -2854,6 +2936,112 @@ def create_fixed_models(
             32,
         )
 
+    if FLAGS.tensorrt_bf16:
+        if os.getenv("GPU_MODEL") == "A100":
```
tanmayv25, line 2940: Instead of GPU_MODEL, it will be better if we could put a check on compute capability. 
`if CC >= 8.0`

qa/common/gen_qa_models.py
```diff
@@ -1212,13 +1272,29 @@ def create_plan_modelconfig(
             model_name,
             max_batch,
             version_policy_str,
-            np_to_model_dtype(input_dtype),
+            (
+                trt_dtype_to_model_dtype(input_dtype)
```
tanmayv25, line 1276: Move this special handling within np_to_model_dtype? Otherwise it is a lot of duplicate code.

qa/common/gen_qa_models.py
```diff
@@ -2854,6 +2936,112 @@ def create_fixed_models(
...
+        if os.getenv("GPU_MODEL") == "A100":
+            # Dynamic shape
+            create_plan_modelconfig(
+                FLAGS.models_dir,
+                8,
+                1,
+                [-1, -1],
+                [-1, -1],
+                [-1, -1],
+                trt.bfloat16,
```
tanmayv25, line 2949: Let's create a dummy class called np_dtype_bfloat16 and document there saying as numpy does not support BF16 datatype, we would use this as a representative. 
Internally, we can handle this object to convert to trt.bfloat16 for plan models.

tanmayv25: Let's not expose trt abstraction here and try to keep it generic to add support for bf16 in other backends as well.

qa/common/gen_qa_model_repository
```diff
@@ -49,6 +49,7 @@
 ############################################################################
 
 TRITON_VERSION=${TRITON_VERSION:=24.05}
+MODEL_TYPE=${MODEL_TYPE:="V100"}
```
tanmayv25, line 52: There should be a way to automatically extract compute capability in python script than defining a new env var.

qa/common/gen_qa_models.py
pskiran1: Yes, Tanmay. Earlier, didn't follow this approach because we needed to install the library externally. I hope it is fine. I will make the above change. Thank you.

tanmayv25 (approved):

PR #7311, opened 2024-06-01
Add support for response sender in the default mode

kthui:
Related PR: https://github.com/triton-inference-server/python_backend/pull/364

Add testing for using response sender in non-decoupled model. Add testing for different ways of sending response(s), i.e. sending before, after and/or during `execute()` return, for decoupled/non-decoupled and async/threaded models.

Items to be done on another ticket:
* Enable restart sub-test. [This functionality is never mentioned on documentation, so it is ok to skip it for now]
* Investigate why BLS CUDA shm pool 128 MB is using more memory.

Logistics:
* This PR should merge after https://github.com/triton-inference-server/server/pull/7265
* All commits on this PR came from previously approved PRs (less copyright/documentation updates), the purpose of this PR is to merge the feature into the main branch.

### DIFF
--- a/qa/L0_backend_python/lifecycle/lifecycle_test.py
+++ b/qa/L0_backend_python/lifecycle/lifecycle_test.py
@@ -199,7 +199,7 @@ class LifecycleTest(unittest.TestCase):
                     print(e.message())
                     self.assertTrue(
                         e.message().startswith(
-                            "Failed to process the request(s) for model instance"
+                            "Failed to process the request(s) for model "
                         ),
                         "Exception message is not correct",
                     )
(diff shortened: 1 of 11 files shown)


Tabrizian (dismissed):

qa/python_models/response_sender/model_common.py
```diff
@@ -0,0 +1,210 @@
...
+          * NUMBER_OF_RESPONSE_BEFORE_RETURN (UINT8):
+              Determines the number of responses before returning from execute function.
+          * SEND_COMPLETE_FINAL_FLAG_BEFORE_RETURN (BOOL):
+              Determines whether the final flag will be sent before return.
+          * RETURN_A_RESPONSE (BOOL):
+              Return the response when the model is returning from `execute` function.
+          * NUMBER_OF_RESPONSE_AFTER_RETURN (UINT8):
+              Determines the number of responses after return.
+          * SEND_COMPLETE_FINAL_FLAG_AFTER_RETURN (BOOL):
+              Determines whether the final flag will be sent before return.
```
krishung5, line 56:
```suggestion
              Determines whether the final flag will be sent after return.
```

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -208,45 +208,6 @@ def test_infer_pymodel_error(self):
                         False, "Wrong exception raised or did not raise an exception"
                     )
 
-    def test_incorrect_execute_return(self):
```
krishung5, line 211: qq: why are we removing this test case?

krishung5 (dismissed):
LGTM! Very detailed testing 🚀

qa/L0_backend_python/lifecycle/lifecycle_test.py
kthui: This test case expects the model to return `None` or `[None, ...]`, this is an error if returning a response is the only way for non-decoupled models to send a response. With the introduction of response sender for non-decoupled models, the models may return `None` or `[None, ...]` which indicates the response will be sent later after the `execute()` function returns. Thus, the checks on this test case is outdated and therefore removed. The new behavior will be tested by the new response sender tests.

Currently, it is difficult to determine if a model has "forgotten" to send a response and/or final flag, because there is no restriction on how soon the model must send response(s) after `execute()` return, although the destruction of `response_sender` at non-closed state is a hint of such event. However, for the purpose of this test, it is too late for the exception to be returned to the request, because the request is "done" by the time `response_sender` is destructed.

qa/python_models/response_sender/model_common.py
kthui: Nice catch!

qa/L0_backend_python/lifecycle/lifecycle_test.py
krishung5: Makes sense, thanks for the info!

krishung5 (approved):

Tabrizian (approved):

PR #7325, opened 2024-06-05
fix: Handling grpc cancellation edge-case:: Cancelling at step START

oandreeva-nv:
#### What does the PR do?
This PR fixes an issue with a gRPC ModelInferHandler stopping accepting requests after a number of cancellation received.

Main root-cause and pre-conditions before hand:
At some point  cancellation notification is received at step `START`. When this happens, we never skip this [block](https://github.com/triton-inference-server/server/blob/352b88590424e83564f00baf411e1c017c694f8d/src/grpc/infer_handler.cc#L696-L699)

Since, we never went skip it, we don't create new state for future incoming requests, i.e.  [call `StartNewRequest`](https://github.com/triton-inference-server/server/blob/352b88590424e83564f00baf411e1c017c694f8d/src/grpc/infer_handler.cc#L729-L731)

Thus, completion queue becomes exhausted at some point. gRPC requests come, but there is nothing that accepts it on Triton's side.

Introduced changes make sure we create new request handler in those situations. 
Added test logic:

I start server and send large amount of inference requests and cancel them right away. Pre fix, the clear identification that there are no `ModelInferHandler`'s for any in-coming request is the server stops logging `"New request handler for ModelInfer"`, i.e 
`grep -c "New request handler for ModelInfer"` doesn't change. In all pre-fix scenarious it happens after 
`"Cancellation notification received for ModelInferHandler, rpc_ok=1, context 0, [0-9]* step START"` was logged 4 times, 2 times for 1 request. Since we start 2 `ModelInferHandler`'s threads initially, that make sense, as no new Infer handlers were created to handle incoming requests and Triton just keeps processing what it already has. 

After fix, StartNewRequest is called properly and  `"New request handler for ModelInfer"` keeps growing, as well as `"Cancellation notification received for ModelInferHandler, rpc_ok=1, context 0, [0-9]* step START"`.

#### Checklist`
- [X] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [X] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:
pre-fix: 15596577
post-fix: 15596692

### DIFF
--- a/qa/L0_request_cancellation/grpc_cancellation_test.py
+++ b/qa/L0_request_cancellation/grpc_cancellation_test.py
@@ -64,13 +64,8 @@ def grpc_async_infer_request_with_instant_cancellation(model_name):
     with grpcclient.InferenceServerClient(url="localhost:8001") as client:
         t_end = time.time() + 60
         while time.time() < t_end:
-            future = client.async_infer(
-                model_name=model_name,
-                inputs=inputs_,
-                callback=partial(callback, user_data),
-                outputs=outputs_,
-            )
-            future.cancel()
+            client.start_stream(callback=partial(callback, user_data))
+            client.stop_stream(cancel_requests=True)
 
 
 class GrpcCancellationTest(unittest.IsolatedAsyncioTestCase):
--- a/qa/L0_request_cancellation/grpc_cancellation_test.py
+++ b/qa/L0_request_cancellation/grpc_cancellation_test.py
@@ -62,10 +62,14 @@ def grpc_async_infer_request_with_instant_cancellation(model_name):
     inputs_, outputs_ = prepare_inputs_outputs()
     user_data = UserData()
     with grpcclient.InferenceServerClient(url="localhost:8001") as client:
-        t_end = time.time() + 60
-        while time.time() < t_end:
-            client.start_stream(callback=partial(callback, user_data))
-            client.stop_stream(cancel_requests=True)
+        future = client.async_infer(
+            model_name=model_name,
+            inputs=inputs_,
+            callback=partial(callback, user_data),
+            outputs=outputs_,
+        )
+        time.sleep(2)
+        future.cancel()
 
 
 class GrpcCancellationTest(unittest.IsolatedAsyncioTestCase):
--- a/qa/L0_request_cancellation/grpc_cancellation_test.py
+++ b/qa/L0_request_cancellation/grpc_cancellation_test.py
@@ -50,6 +50,29 @@ def callback(user_data, result, error):
         user_data._completed_requests.put(result)
 
 
+def prepare_inputs_outputs():
+    inputs, outputs = [], []
+    inputs.append(grpcclient.InferInput("INPUT0", [1, 1], "INT32"))
+    outputs.append(grpcclient.InferRequestedOutput("OUTPUT0"))
+    inputs[0].set_data_from_numpy(np.array([[5]], dtype=np.int32))
+    return inputs, outputs
+
+
+def grpc_async_infer_request_with_instant_cancellation(model_name):
+    inputs_, outputs_ = prepare_inputs_outputs()
+    user_data = UserData()
+    with grpcclient.InferenceServerClient(url="localhost:8001") as client:
+        t_end = time.time() + 60
+        while time.time() < t_end:
+            future = client.async_infer(
+                model_name=model_name,
+                inputs=inputs_,
+                callback=partial(callback, user_data),
+                outputs=outputs_,
+            )
+            future.cancel()
+
+
 class GrpcCancellationTest(unittest.IsolatedAsyncioTestCase):
     _model_name = "custom_identity_int32"
     _model_delay = 10.0  # seconds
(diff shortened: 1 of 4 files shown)


src/grpc/infer_handler.cc
```diff
@@ -694,6 +694,16 @@ ModelInferHandler::Process(InferHandler::State* state, bool rpc_ok)
   // Handle notification for cancellation which can be raised
   // asynchronously if detected on the network.
   if (state->IsGrpcContextCancelled()) {
+    if (rpc_ok && (state->step_ == Steps::START) &&
+        (state->context_->step_ != Steps::CANCELLED)) {
```
nnshah1, line 698: question: what does the second clause here imply? 

!= Steps::Cancelled ?

nnshah1 (approved):
LGTM - had question on the condition

qa/L0_request_cancellation/grpc_cancellation_test.py
```diff
@@ -50,6 +50,29 @@ def callback(user_data, result, error):
...
+    inputs[0].set_data_from_numpy(np.array([[5]], dtype=np.int32))
+    return inputs, outputs
+
+
+def grpc_async_infer_request_with_instant_cancellation(model_name):
+    inputs_, outputs_ = prepare_inputs_outputs()
+    user_data = UserData()
+    with grpcclient.InferenceServerClient(url="localhost:8001") as client:
+        t_end = time.time() + 60
+        while time.time() < t_end:
```
tanmayv25, line 66: How about adding artificial delays in ModelInferHandler Process loop? 
That would ensure that we hit cancellation in START step even with a single request instead of running high inference load and expect some of the requests to hit the case.
Something like these: https://github.com/triton-inference-server/server/blob/2ac5f34b3b6dc07ac4c30f38c5df74fa2547fcb9/src/grpc/infer_handler.h#L1016

tanmayv25:
I think we would need similar fix for ModelStreamInferHandler.

src/grpc/infer_handler.cc
oandreeva-nv: To avoid calling `StartNewRequest` twice, at first we fall into HandleCancellation and go through [this](https://github.com/triton-inference-server/server/blob/main/src/grpc/infer_handler.h#L815-L826) block, which returns `true` for `resume`, so we will go into `if (state->IsGrpcContextCancelled()) ` loop for the second time but this time `state->context_->step_ ` is `CANCELLED`

qa/L0_request_cancellation/grpc_cancellation_test.py
oandreeva-nv: Can this be done in a follow up PR? I feel like we still have to analyze all edge case that are left and not processed?

oandreeva-nv: re-factored

src/grpc/infer_handler.cc
GuanLuo: Late to the game, but what is the reasoning of not moving the original "`StartNewRequest()` if at START" to before handling the cancellation? Although I think other code needs to be moved around as well.

qa/L0_request_cancellation/test.sh
```diff
@@ -110,6 +110,59 @@ for TEST_CASE in "test_grpc_async_infer" "test_grpc_stream_infer" "test_aio_grpc
...
+    NEW_REQ_HANDL_COUNT=$(grep -c "New request handler for ModelInferHandler" $SERVER_LOG)
+    if [[ $NEW_REQ_HANDL_COUNT == $INIT_NEW_REQ_HANDL_COUNT ]]; then
+        echo -e "\n***\n*** gRPC Cancellation on step START Test Failed: New request handler for ModelInferHandler was not created \n***"
+        cat $TEST_LOG
+        RET=1
+        break
+    else
+        break
+    fi
+done
```
kthui, line 158: I wonder if the testing procedure can be placed inside a Python unit test?

oandreeva-nv: Yes, good idea. Will think about re-factoring

src/grpc/infer_handler.cc
oandreeva-nv: I am not 100% aware of all underlying processes, meaning `state->step_` and `state->context_->step_` combinations. This change helps to address the bug with known symptoms. Refactoring if the `Process` logic needs proper time and testing IMHO

qa/L0_request_cancellation/test.sh
oandreeva-nv: Converted to unit test

GuanLuo (approved):

src/grpc/infer_handler.cc
GuanLuo: @kthui thoughts? If feasible, this can be done as follow-up and by someone else. Want to make sure if there is room for improvement.

kthui: Yes, I think there is definitely room for improvement/refactoring, i.e. I think the `if (shutdown) { ... }` could also be moved into the `if (state->step_ == Steps::START) { ... } else ...` block, so all procedures for `Steps::START` would be inside the `if (state->step_ == Steps::START) { ... }` block, but it can be done as a follow-up later.

kthui (approved):

tanmayv25 (approved):

src/grpc/infer_handler.cc
oandreeva-nv: Jira ticket: DLIS-6831

PR #7390, opened 2024-06-27
[refactor]: Refactor Frontend Trace OpenTelemetry Implementation

oandreeva-nv:
#### What does the PR do?

This PR introduces an improvement to our frontend OpenTelemetry tracing implementation. The main goal of this refactoring is to enhance the reliability and flexibility of the system.

Key Changes:
-  Introduced a Stack-Based Approach
    - The previous implementation  relied on specific span names to preserve the trace hierarchy. 
    - Current solution uses stacks and doesn't need to know what type of span was started.
    - For ensemble and bls models, any number of sub-traces can be spawned. Since we preserve `parent_id` for every trace, I utilize this and introduced an unordered map to keep spans in stacks. Each trace and sub-trace has its own stack. When stack is empty, I use `parent_id` to check the top of the parent's stack to get parent span's information.
    - This refactor also helps the following `Custom backend tracing` implementation to avoid handling of the complex logic for multiple nested custom spans, i.e. we simply look at the top of the stack.

#### Checklist
- [X] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [] Verified that the PR passes existing CI.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [X] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

N/A 

trace.h 

#### Test plan:

- CI Pipeline ID: 16139557

N/A

### DIFF
--- a/src/tracer.cc
+++ b/src/tracer.cc
@@ -345,13 +345,13 @@ TraceManager::SampleTrace(const TraceStartOptions& start_options)
       auto active_span =
           otel_trace_api::GetSpan(start_options.propagated_context);
       if (active_span->GetContext().IsValid()) {
-        ts->span_stacks_[ts->trace_id_]->push(active_span);
+        ts->span_stacks_[ts->trace_id_]->emplace(active_span);
       }
-      ts->root_span_ =
-          ts->StartSpan("InferRequest", steady_timestamp_ns, ts->trace_id_, 0);
       // Storing "InferRequest" span as a root span
       // to keep it alive for the duration of the request.
-      ts->span_stacks_[ts->trace_id_]->push(ts->root_span_);
+      ts->root_span_ =
+          ts->StartSpan("InferRequest", steady_timestamp_ns, ts->trace_id_);
+      ts->span_stacks_[ts->trace_id_]->emplace(ts->root_span_);
 #else
       LOG_ERROR << "Unsupported trace mode: "
                 << TraceManager::InferenceTraceModeString(ts->setting_->mode_);
@@ -368,7 +368,7 @@ TraceManager::Trace::~Trace()
     setting_->WriteTrace(streams_);
   } else if (setting_->mode_ == TRACE_MODE_OPENTELEMETRY) {
 #ifndef _WIN32
-    EndSpan(kRootSpan, trace_id_);
+    EndSpan(trace_id_);
 #else
     LOG_ERROR << "Unsupported trace mode: "
               << TraceManager::InferenceTraceModeString(setting_->mode_);
@@ -512,7 +512,7 @@ TraceManager::ProcessOpenTelemetryParameters(
 
 void
 TraceManager::Trace::StartSpan(
-    std::string span_key, TRITONSERVER_InferenceTrace* trace,
+    TRITONSERVER_InferenceTrace* trace,
     TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns,
     uint64_t trace_id)
 {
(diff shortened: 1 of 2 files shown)


src/tracer.cc
```diff
@@ -501,15 +512,15 @@ TraceManager::ProcessOpenTelemetryParameters(
...
+    TRITONSERVER_InferenceTrace* trace,
     TRITONSERVER_InferenceTraceActivity activity, uint64_t timestamp_ns,
     uint64_t trace_id)
 {
   uint64_t parent_id;
   LOG_TRITONSERVER_ERROR(
       TRITONSERVER_InferenceTraceParentId(trace, &parent_id),
       "getting trace parent id");
-  std::string parent_span_key = "";
+  auto span_parent_id = parent_id;
```
GuanLuo, line 523: Just use parent_id and zero-initialize it first?

src/tracer.cc
```diff
@@ -580,45 +583,37 @@ TraceManager::Trace::StartSpan(
...
-  // the context and provide it through StartSpanOptions to the child span
-  if (!parent_span_key.empty() && otel_context_.HasKey(parent_span_key)) {
-    auto parent_span = opentelemetry::nostd::get<
-        opentelemetry::nostd::shared_ptr<otel_trace_api::Span>>(
-        otel_context_.GetValue(parent_span_key));
-    options.parent = parent_span->GetContext();
+  // If the new span is a child span, we need to retrieve its parent and
+  // provide it through StartSpanOptions to the child span
+  if (span_stacks_.find(trace_id) != span_stacks_.end() &&
+      !span_stacks_[trace_id]->empty()) {
```
GuanLuo, line 589: Won't this be entered twice if the caller: start span -> end span -> start span? Will this mistakenly treat the trace to "have" a parent?

Tabrizian (commented):
Could you add an example for how this looks like for t

src/tracer.cc
oandreeva-nv: Every `end span` event pops span from the stack, so it doesn't exist anymore. 
The trace frontend currently controls root span creation, request spans and compute spans. The passed `trace_id` in this method obides the logic, explained here: https://github.com/triton-inference-server/server/pull/7390/files#diff-c05667609085c993fd282ff18d0429945c44492b95c0d01fcbeec6946fda2db2R525-R534

src/tracer.h
```diff
@@ -315,58 +323,44 @@ class TraceManager {
...
-
-    /// Adds event to the span, which is retrieved from OpenTelemetry context
-    /// with the provided `span_key`. If activity is
-    /// TRITONSERVER_TRACE_REQUEST_START, or TRITONSERVER_TRACE_COMPUTE_START,
-    /// starts a new span and adds it to `otel_context_`.
+    void EndSpan(const uint64_t& raw_timestamp_ns, uint64_t trace_id);
+
+    /// Adds event to the span on the top of the stack, related to trace
+    /// with `trace_id`. If activity is TRITONSERVER_TRACE_REQUEST_START,
+    /// or TRITONSERVER_TRACE_COMPUTE_START, starts a new span and adds it
```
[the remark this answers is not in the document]
oandreeva-nv: Could you please clarify?

oandreeva-nv:
@Tabrizian this is just a re-factor for the frontend. The actual custom backend implementation PR will be up shortly

src/tracer.cc
oandreeva-nv: I need both, since later on L563, I need true `parent_id`

GuanLuo (approved):

PR #7404, opened 2024-07-02
perf: Improve response throughput of a single gRPC stream

kthui:
#### What does the PR do?
Improve the response throughput of a single gRPC stream on the server frontend.

When there is a large number of responses, responses will compete with each others for the single write slot - a restriction imposed by gRPC that no concurrent write is allowed within a stream. A response that do not get a chance to write will be placed to the back of the completion queue and try again when it comes out. Unless the number of responses reduces immediately, the response that do not get a chance to write will likely not get a chance to write when it comes out. The same happens to other responses at the same time, resulting in a busy loop on the completion queue that impact performance.

This change introduces a new queue for holding responses that do not get a chance to write. When a new response arrives, if the stream is writing another response, then the new response will be placed to the back of the new queue. When the ongoing write is completed, the top response from the new queue will be popped and starts writing.

With this change, the response throughput previously handled by 4 gRPC streams in parallel can be handled by only 1 gRPC stream, with a lot of room to spare.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [x] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

Start reviewing from the bottom of `stream_infer_handler.cc` and make your way to the top of the file.

#### Test plan:
This PR is only a performance improvement, so existing tests should be covering all issues.

- CI Pipeline ID:
#16352903

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/src/grpc/stream_infer_handler.cc
+++ b/src/grpc/stream_infer_handler.cc
@@ -367,14 +367,9 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
         std::lock_guard<std::recursive_mutex> lk1(state->context_->mu_);
         {
           std::lock_guard<std::recursive_mutex> lk2(state->step_mtx_);
-          bool has_prev_ready_response =
-              state->response_queue_->HasReadyResponse();
           state->response_queue_->MarkNextResponseComplete();
-          if (!has_prev_ready_response) {
-            state->context_->ready_to_write_states_.push(state);
-          }
-          if (!state->context_->ongoing_write_ &&
-              !state->context_->ready_to_write_states_.empty()) {
+          state->context_->ready_to_write_states_.push(state);
+          if (!state->context_->ongoing_write_) {
             state->context_->ongoing_write_ = true;
             writing_state = state->context_->ready_to_write_states_.front();
             state->context_->ready_to_write_states_.pop();
--- a/src/grpc/stream_infer_handler.cc
+++ b/src/grpc/stream_infer_handler.cc
@@ -474,6 +474,8 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
         if (!state->context_->ready_to_write_states_.empty()) {
           writing_state = state->context_->ready_to_write_states_.front();
           state->context_->ready_to_write_states_.pop();
+        } else {
+          state->context_->ongoing_write_ = false;
         }
       }
       if (writing_state != nullptr) {
@@ -482,20 +484,13 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
 
       // Finish the state if all the transactions associated with
       // the state have completed.
-      {
+      if (state != writing_state) {
         std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
-        if (state != writing_state) {
-          std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
-          if (state->IsComplete()) {
-            state->context_->DecrementRequestCounter();
-            finished = Finish(state);
-          } else {
-            state->step_ = Steps::ISSUED;
-          }
-        }
-        if (writing_state == nullptr) {
-          std::lock_guard<std::recursive_mutex> lock(state->context_->mu_);
-          state->context_->ongoing_write_ = false;
+        if (state->IsComplete()) {
+          state->context_->DecrementRequestCounter();
+          finished = Finish(state);
+        } else {
+          state->step_ = Steps::ISSUED;
         }
       }
     } else if (state->step_ == Steps::WRITEREADY) {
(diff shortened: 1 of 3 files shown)


src/grpc/stream_infer_handler.cc
```diff
@@ -359,13 +359,31 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
-        state->complete_ = true;
-        state->context_->PutTaskBackToQueue(state);
+        InferHandler::State* writing_state = nullptr;
+        std::lock_guard<std::recursive_mutex> lk1(state->context_->mu_);
+        {
+          std::lock_guard<std::recursive_mutex> lk2(state->step_mtx_);
+          bool has_prev_ready_response =
+              state->response_queue_->HasReadyResponse();
+          state->response_queue_->MarkNextResponseComplete();
+          if (!has_prev_ready_response) {
```
tanmayv25, line 373: Why add this condition? This code is executed when InferAsync call returns an error. The `has_prev_ready_response` should always be false in this case right? As there should be no other responses.

src/grpc/stream_infer_handler.cc
```diff
@@ -359,13 +359,31 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
+            state->context_->ready_to_write_states_.push(state);
+          }
+          if (!state->context_->ongoing_write_ &&
+              !state->context_->ready_to_write_states_.empty()) {
+            state->context_->ongoing_write_ = true;
```
tanmayv25, line 378: Can you add comments here describing why we are circulating through the ready_to_write_states_ when there is no ongoing_write_? 
Describe how the state would be picked up later once the ongoing_write_ is completed.

tanmayv25: Also mention that ongoing_write would be reset once the completion queue returns a written state object.

src/grpc/stream_infer_handler.cc
```diff
@@ -469,61 +486,74 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
-            std::chrono::milliseconds(state->delay_response_ms_));
-      }
-
       // Finish the state if all the transactions associated with
       // the state have completed.
       if (state->IsComplete()) {
         state->context_->DecrementRequestCounter();
         finished = Finish(state);
       } else {
-        // GRPC doesn't allow to issue another write till
```
tanmayv25, line 509: Move the information held in this comment in StateWriteResponse calls? 
Would not want to lose this vital information.

src/grpc/stream_infer_handler.cc
```diff
@@ -701,45 +731,62 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
+      if (!has_prev_ready_response && response) {
+        state->context_->ready_to_write_states_.push(state);
+      }
+      if (!state->context_->ongoing_write_ &&
+          !state->context_->ready_to_write_states_.empty()) {
+        state->context_->ongoing_write_ = true;
+        writing_state = state->context_->ready_to_write_states_.front();
+        state->context_->ready_to_write_states_.pop();
+      }
+      if (is_complete && state->response_queue_->IsEmpty() &&
```
tanmayv25, line 774: Can you add a comment on why this block is needed? The response_queue is empty and we are marking the step_  WRITEREADY.

src/grpc/stream_infer_handler.cc
kthui: I see. The InferAsync only return errors up to scheduler enqueue. Yes, `has_prev_ready_response` should always be false.

Updated: [No need to check for previous response on InferAsync error](https://github.com/triton-inference-server/server/pull/7404/commits/4f10e06b3634660b81c29ccbc062e1e500029797)

src/grpc/stream_infer_handler.cc
kthui: Sure, [Add comments on ready to write lifecycle](https://github.com/triton-inference-server/server/pull/7404/commits/edfaa2ee71eec8bed692feeefa79c56e8d81578a)

src/grpc/stream_infer_handler.cc
kthui: Sure, [Add comments on WRITEREADY cleanup routine](https://github.com/triton-inference-server/server/pull/7404/commits/03efaa6fba0a1f9804893ccd9306a0509165fe15)

src/grpc/stream_infer_handler.cc
kthui: Sure, [Add comments on gRPC does not allow concurrent write](https://github.com/triton-inference-server/server/pull/7404/commits/bc74e68301c7f2f6efc342c436bfeba17749113e)

tanmayv25 (approved):

PR #7408, opened 2024-07-03
build: Reduce intermediate layers

krishung5:
#### What does the PR do?

Reduce intermediate layers in order to reduce TRT-LLM container size.

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

TRT-LLM backend build + tests should pass.

- CI Pipeline ID: 16294063

- closes GitHub issue: #xxx

### DIFF
--- a/build.py
+++ b/build.py
@@ -1082,25 +1082,20 @@ RUN patchelf --add-needed /usr/local/cuda/lib64/stubs/libcublasLt.so.12 backends
 """
     if "tensorrtllm" in backends:
         df += """
-
-RUN ldconfig
-# Remove contents that are not needed in runtime
-RUN ARCH="$(uname -i)" \\
-      && rm -fr ${TRT_ROOT}/bin ${TRT_ROOT}/targets/${ARCH}-linux-gnu/bin ${TRT_ROOT}/data \\
-      && rm -fr  ${TRT_ROOT}/doc ${TRT_ROOT}/onnx_graphsurgeon ${TRT_ROOT}/python \\
-      && rm -fr ${TRT_ROOT}/samples  ${TRT_ROOT}/targets/${ARCH}-linux-gnu/samples
-
 # Install required packages for TRT-LLM models
-RUN python3 -m pip install --upgrade pip \\
-      && pip3 install transformers
-
-# ldconfig for TRT-LLM
-RUN find /usr -name libtensorrt_llm.so -exec dirname {} \; > /etc/ld.so.conf.d/tensorrt-llm.conf
-RUN find /opt/tritonserver -name libtritonserver.so -exec dirname {} \; > /etc/ld.so.conf.d/triton-tensorrtllm-worker.conf
-
+# Remove contents that are not needed in runtime
 # Setuptools has breaking changes in version 70.0.0, so fix it to 69.5.1
 # The generated code in grpc_service_pb2_grpc.py depends on grpcio>=1.64.0, so fix it to 1.64.0
-RUN pip3 install setuptools==69.5.1 grpcio-tools==1.64.0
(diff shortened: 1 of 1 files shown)


nv-kmcgill53:
What is the improvement with this change?

krishung5:
It seems like every Dockerfile command, e.g. FROM, , RUN, etc. creates an image layer of its own. The layers are then used to assemble the final image. With more accumulated layers, the image will be increased in size.

nv-kmcgill53 (approved):

build.py
```diff
@@ -1082,25 +1082,20 @@ def create_dockerfile_linux(
...
+# Remove contents that are not needed in runtime
 # Setuptools has breaking changes in version 70.0.0, so fix it to 69.5.1
 # The generated code in grpc_service_pb2_grpc.py depends on grpcio>=1.64.0, so fix it to 1.64.0
-RUN pip3 install setuptools==69.5.1 grpcio-tools==1.64.0
+RUN ldconfig && \
+    ARCH="$(uname -i)" && \
+    rm -fr ${TRT_ROOT}/bin ${TRT_ROOT}/targets/${ARCH}-linux-gnu/bin ${TRT_ROOT}/data && \
+    rm -fr ${TRT_ROOT}/doc ${TRT_ROOT}/onnx_graphsurgeon ${TRT_ROOT}/python && \
+    rm -fr ${TRT_ROOT}/samples ${TRT_ROOT}/targets/${ARCH}-linux-gnu/samples && \
+    python3 -m pip install --upgrade pip && \
```
mc-nv, line 1094: I would rather keep `pip` to have  clean docker file.

build.py
```diff
@@ -1082,25 +1082,20 @@ def create_dockerfile_linux(
...
-RUN find /usr -name libtensorrt_llm.so -exec dirname {} \; > /etc/ld.so.conf.d/tensorrt-llm.conf
-RUN find /opt/tritonserver -name libtritonserver.so -exec dirname {} \; > /etc/ld.so.conf.d/triton-tensorrtllm-worker.conf
-
+# Remove contents that are not needed in runtime
 # Setuptools has breaking changes in version 70.0.0, so fix it to 69.5.1
 # The generated code in grpc_service_pb2_grpc.py depends on grpcio>=1.64.0, so fix it to 1.64.0
-RUN pip3 install setuptools==69.5.1 grpcio-tools==1.64.0
+RUN ldconfig && \
+    ARCH="$(uname -i)" && \
+    rm -fr ${TRT_ROOT}/bin ${TRT_ROOT}/targets/${ARCH}-linux-gnu/bin ${TRT_ROOT}/data && \
```
mc-nv, line 1091: Double slash `\\` in the and is an escape symbol was to give developers more clean and readable `Dockerfile` output. Not sure if it make sense to remove it.

mc-nv:
> It seems like every Dockerfile command, e.g. FROM, , RUN, etc. creates an image layer of its own. The layers are then used to assemble the final image. With more accumulated layers, the image will be increased in size.

Idea behind image layers is to keep layer history transparent. 
There why it's better to keep them separate unless it vital.

krishung5:
@GuanLuo The container size from the latest pipeline is now only 24.7G, reducing by 8.7G.

krishung5:
Will create a follow up PR to address @mc-nv's comments.

build.py
krishung5: Do you mean just run `pip install` instead of `python3 -m pip install`?

PR #7409, opened 2024-07-03
[fix]: grpc state cleanup fix

oandreeva-nv:
#### What does the PR do?

This PR fixes intermittence of L0_grpc_state_cleanup.

2 sources of intermittence:

1) *_error_status: Was failing periodically because it was seeing `The stream is no longer in valid state, the error detail is reported through provided callback. A new stream should be started after stopping the current stream.`, but expecting `This protocol is restricted, expecting header 'triton-grpc-protocol-infer-key'`
Due to request count > 1, there is a time interval between when the request1 gets picked by gRPC server infer handler and it detect that the protocol was restricted. The requests sent in between this time period from client side will not see  "The stream is no longer in valid state" error.

It was decided to relax the condition and checking for both errors.

2) *_shutdownserver:

Was failing less frequent that 1 due to `Exception was not raised`. I believe something was happening to `os.kill` either being late in killing the process or some delay with killing process, so I've added a sleep after os.kill.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [X] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

From the beginning

#### Test plan:
- CI Pipeline ID: ~16297581~ 16303863

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_grpc_state_cleanup/cleanup_test.py
+++ b/qa/L0_grpc_state_cleanup/cleanup_test.py
@@ -437,7 +437,13 @@ class CleanUpTest(tu.TestResultCollector):
         ]
         with self.assertRaises(InferenceServerException) as cm:
             self._simple_infer(request_count=10)
-        self.assertIn(str(cm.exception), expected_exceptions)
+
+        exception_match = False
+        for expected_exception in expected_exceptions:
+            exception_match |= expected_exception in str(cm.exception)
+        self.assertTrue(
+            exception_match, "Raised unexpected exception {}".format(str(cm.exception))
+        )
 
     def test_simple_infer_shutdownserver(self):
         # This test case is used to check whether all the state objects are
@@ -479,7 +485,13 @@ class CleanUpTest(tu.TestResultCollector):
         ]
         with self.assertRaises(InferenceServerException) as cm:
             self._streaming_infer(request_count=10, should_error=True)
-        self.assertIn(str(cm.exception), expected_exceptions)
+
+        exception_match = False
+        for expected_exception in expected_exceptions:
+            exception_match |= expected_exception in str(cm.exception)
+        self.assertTrue(
+            exception_match, "Raised unexpected exception {}".format(str(cm.exception))
+        )
 
     def test_streaming_infer_shutdownserver(self):
         # This test case is used to check whether all the state objects are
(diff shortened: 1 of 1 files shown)


tanmayv25:
> I believe something was happening to os.kill either being late in killing the process or some delay with killing process, so I've added a sleep after os.kill.

The test case is supposed to be testing that we cleanly handle state objects when the requests are received and cancelled when the server was under shutdown phase.
There might be some issues in server implementation during shutdown.

qa/L0_grpc_state_cleanup/cleanup_test.py
```diff
@@ -428,11 +431,18 @@ def test_simple_infer_timeout(self):
     def test_simple_infer_error_status(self):
         # This test case is used to check whether all the state objects are
         # released when RPC runs into error.
+        expected_exceptions = [
+            "This protocol is restricted, expecting header 'triton-grpc-protocol-infer-key'",
+            "The stream is no longer in valid state, the error detail is reported through provided callback. A new stream should be started after stopping the current stream.",
```
GuanLuo, line 436: Is `_simple_infer` using gRPC streaming?

oandreeva-nv: good catch

GuanLuo:
How can `This protocol is restricted, expecting header 'triton-grpc-protocol-infer-key'` break the stream state? Is that because the check is happening way early in the gRPC communication and the gRPC frontend does recover from it properly?

qa/L0_grpc_state_cleanup/cleanup_test.py
oandreeva-nv: removed

oandreeva-nv:
@tanmayv25 I've removed `time.sleep` and increased request count. In 12/12 tests exception was raised and intermittence seems to be no more. Could you please also help in answering @GuanLuo's question.

tanmayv25:
>How can This protocol is restricted, expecting header 'triton-grpc-protocol-infer-key' break the stream state? Is that because the check is happening way early in the gRPC communication and the gRPC frontend does not recover from it properly?

The client is sending requests into a streaming connection with request delay 0. The requests that hit the protocol restricted error on the server gets returned as a [gRPC exception ](https://github.com/triton-inference-server/client/blob/main/src/python/library/tritonclient/grpc/_infer_stream.py#L157) which is caught in the client library. When error is caught, and the stream is marked [inactive](https://github.com/triton-inference-server/client/blob/44ed1215960e72cd241aef07ea6a33ce3d727265/src/python/library/tritonclient/grpc/_infer_stream.py#L162). 

The exception is caught and sent to the callback function for being handled by the client. The callback function just places the error into a queue which would be picked up by the client in future. However, the client still might send request after the stream is closed as the error object has not been picked from the queue. It is then we see this error appear. The gRPC frontend doesn't even register these requests.

tanmayv25 (approved):

PR #7423, opened 2024-07-08
test: Tests for Metrics API enhancement to include error counters

indrajit96:
#### What does the PR do?
Currently We don't have any mechanism to count any failures that happen in triton core before the request gets to the backend.
Enhances existing metrics API to incorporate failure and failure counts in inference requests.
This PR adds new test

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [x] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:
https://github.com/triton-inference-server/core/pull/377

RespondIfError function call in core

#### Test plan:
Test for **REJECTED**
Modified qa/L0_model_queue/model_queue_test.py to capture rejections in Ensemble Model and Non Ensemble Model in dynamic batcher.
custom_zero_1_float32 model expected to have 4 request rejects

Test for **OTHER**
Modified qa/L0_model_queue/model_queue_test.py to capture OTHER errors in Ensemble Model.
ensemble_zero_1_float32 expects to have 2 request rejected reported as OTHER.

Test for **CANCELED**
Modified qa/L0_request_cancellation/scheduler_test.py to capture various cancellations.
_test_sequence_batch_scheduler_queued_request_cancellation test cancel 2 requests, expected to be reported as 2 in test_direct_sequence_batch_scheduler_request_cancellation() and test_sequence_batch_scheduler_backlog_request_cancellation()

Test for **BACKEND**
Modified qa/L0_backend_python/lifecycle/lifecycle_test.py which generates errors from backend inside test_infer_pymodel_error by using wrong model name.
Expect the error count to be 1.

**Untested** :
- REJECTED request which gets triggered from SequenceBatchScheduler()
- OTHER uncategorized failures for individual (non-ensemble) models
- Follow-up ticket for further testing: `TODO` @indrajit96 

- closes Jira issue: DLIS-5530

### DIFF
--- a/docs/user_guide/metrics.md
+++ b/docs/user_guide/metrics.md
@@ -100,19 +100,25 @@ Count*. The count metrics are illustrated by the following examples:
 |              |Execution Count |`nv_inference_exec_count` |Number of inference batch executions (see [Inference Request Metrics](#inference-request-metrics), does not include cached requests)|Per model|Per request|
 |              |Pending Request Count |`nv_inference_pending_request_count` |Number of inference requests awaiting execution by a backend. This number is incremented when a request is enqueued to the server (`TRITONSERVER_ServerInferAsync`) and is decremented when a backend is about to start executing the request. More details can be found below. |Per model|Per request|
 
-#### Failure Count is further classified into four distinct types:
+#### Failure Count Categories
 
-|Sub-Category      |Sub-Metric | Sub-Metric Name |Description |
-|--------------|----------------|------------|------------|
-|Failure Reason         |Rejected |REJECTED  | Number of inference failures due to request timeout in the schedular. |
-|              |Canceled |CANCELED  |  Number of inference failures due to request cancellation in the core. |
-|              |Backend  |BACKEND |  Number of inference failures due to errors from backend. |
-|              |Other | OTHER  | Number of inference failures due to unspecified reasons in the core. |
-
+| Failed Request Reason |Description |
+|------------|------------|
+| REJECTED  | Number of inference failures due to request timeout in the schedular. |
+| CANCELED  |  Number of inference failures due to request cancellation in the core. |
+| BACKEND |  Number of inference failures during execution of requests in the backend/model. |
+| OTHER  | Number of inference failures due to other uncategorized reasons in the core. |
 
 > **Note**
 >
 > Ensemble failure metrics will reflect the failure counts of their composing models as well as the parent model, but currently do not capture the same granularity for the "reason" label and will default to the "OTHER" reason.
+>
+> For example, if EnsembleA contains ModelA, and ModelA experiences a failed request due to a queue/backlog timeout in the scheduler, ModelA will have a failed request metric reflecting `reason=REJECTED` and `count=1`.
+> Additionally, EnsembleA will have a failed request metric reflecting `reason=OTHER` and `count=2`. 
+> The `count=2` reflects 1 from the internally failed request captured by ModelA, as well as 1 from the failed top-level request sent to EnsembleA by the user/client.
+> The `reason=OTHER` reflects that fact that the ensemble doesn't currently capture the specific reason why 
+> ModelA's request failed at this time.
+
 #### Pending Request Count (Queue Size) Per-Model
 
 The *Pending Request Count* reflects the number of requests that have been
--- a/docs/user_guide/metrics.md
+++ b/docs/user_guide/metrics.md
@@ -100,6 +100,19 @@ Count*. The count metrics are illustrated by the following examples:
 |              |Execution Count |`nv_inference_exec_count` |Number of inference batch executions (see [Inference Request Metrics](#inference-request-metrics), does not include cached requests)|Per model|Per request|
 |              |Pending Request Count |`nv_inference_pending_request_count` |Number of inference requests awaiting execution by a backend. This number is incremented when a request is enqueued to the server (`TRITONSERVER_ServerInferAsync`) and is decremented when a backend is about to start executing the request. More details can be found below. |Per model|Per request|
 
+#### Failure Count is further classified into four distinct types:
+
+|Sub-Category      |Sub-Metric | Sub-Metric Name |Description |
+|--------------|----------------|------------|------------|
+|Failure Reason         |Rejected |REJECTED  | Number of inference failures due to request timeout in the schedular. |
+|              |Canceled |CANCELED  |  Number of inference failures due to request cancellation in the core. |
+|              |Backend  |BACKEND |  Number of inference failures due to errors from backend. |
+|              |Other | OTHER  | Number of inference failures due to unspecified reasons in the core. |
+
+
+> **Note**
+>
+> Ensemble failure metrics will reflect the failure counts of their composing models as well as the parent model, but currently do not capture the same granularity for the "reason" label and will default to the "OTHER" reason.
 #### Pending Request Count (Queue Size) Per-Model
 
 The *Pending Request Count* reflects the number of requests that have been
--- a/docs/user_guide/metrics.md
+++ b/docs/user_guide/metrics.md
@@ -114,9 +114,9 @@ Count*. The count metrics are illustrated by the following examples:
 > Ensemble failure metrics will reflect the failure counts of their composing models as well as the parent model, but currently do not capture the same granularity for the "reason" label and will default to the "OTHER" reason.
 >
 > For example, if EnsembleA contains ModelA, and ModelA experiences a failed request due to a queue/backlog timeout in the scheduler, ModelA will have a failed request metric reflecting `reason=REJECTED` and `count=1`.
-> Additionally, EnsembleA will have a failed request metric reflecting `reason=OTHER` and `count=2`. 
+> Additionally, EnsembleA will have a failed request metric reflecting `reason=OTHER` and `count=2`.
 > The `count=2` reflects 1 from the internally failed request captured by ModelA, as well as 1 from the failed top-level request sent to EnsembleA by the user/client.
-> The `reason=OTHER` reflects that fact that the ensemble doesn't currently capture the specific reason why 
+> The `reason=OTHER` reflects that fact that the ensemble doesn't currently capture the specific reason why
 > ModelA's request failed at this time.
 
 #### Pending Request Count (Queue Size) Per-Model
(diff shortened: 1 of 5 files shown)


qa/L0_request_cancellation/scheduler_test.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 3: Should be appended, not overwritten, ex: `2023-2024`, not `2024`.

qa/L0_request_cancellation/scheduler_test.py
```diff
@@ -228,6 +243,15 @@ def test_scheduler_streaming_request_cancellation(self):
         time.sleep(2)  # ensure reaper thread has responded
         self._assert_streaming_response_is_cancelled(response)
 
+    def test_zerror_metrics(self):
```
rmccorm4, line 246: typo?
```suggestion
    def test_error_metrics(self):
```

rmccorm4, line 246: We should ideally have tests for all of the newly introduced failure `reasons`. So need tests for `rejected`, `backend`, and `other`. Feel free to lay out a rough test plan in text and I can review that too.

qa/L0_request_cancellation/scheduler_test.py
```diff
@@ -84,6 +85,20 @@ def _assert_streaming_response_is_cancelled(self, response):
...
+        r.raise_for_status()
+        return r.text
+
+    def _assert_cancel_metrics_sequence_oldest(self, model_name, count, metrics):
+        expected_metric = f'nv_inference_request_failure{{model="{model_name}",reason="CANCELED",version="1"}} {count}'
+        self.assertIn(expected_metric, metrics)
+
+    def _assert_cancel_metrics_sequence_direct(self, model_name, count, metrics):
+        expected_metric = f'nv_inference_request_failure{{model="{model_name}",reason="CANCELED",version="1"}} {count}'
+        self.assertIn(expected_metric, metrics)
```
rmccorm4, line 100: This can just be one helper function called with the two different model names, right?

qa/L0_request_cancellation/scheduler_test.py
indrajit96: adding the 'z' ensures that the test gets run at the very end.
I want to make sure that metrics test runs at the end to leverage existing tests as much as possible.
Did not find a better way as 
"The order in which the various tests will be run is determined by sorting the test method names with respect to the built-in ordering for strings." As per python unit test docs.

rmccorm4: Instead of relying on test order which may be prone to subtle failures in the future, can you instead make the metrics checks more relative?

Ex: instead of checking the metrics real some absolute value like 4, you would capture them at the start of the test case as a relative baseline X, and compare the new metric value afterwards to say X+4. This should be more stable regardless of the order the tests get run.

Let me know if that makes sense.

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -63,6 +66,29 @@ class LifecycleTest(unittest.TestCase):
...
+        return r.text
+
+    def _metrics_before_test(self, model, reason):
+        pattern = rf'nv_inference_request_failure\{{model="{model}",reason="{reason}",version="1"\}} (\d+)'
+        metrics = self._get_metrics()
+        match = re.search(pattern, metrics)
+        if match:
+            return int(match.group(1))
+        else:
+            return int(0)
```
rmccorm4, line 82: Can you raise some kind of error instead if the metric is not found, to avoid some kind of silent failure?

qa/L0_model_queue/model_queue_test.py
```diff
@@ -283,6 +314,20 @@ def test_policy_reject(self):
                 self.check_deferred_exception()
             except InferenceServerException as ex:
                 self.assertTrue(False, "unexpected error {}".format(ex))
+        expected_count_increase = 2
+        self._assert_metrics(
```
rmccorm4, line 318:
```suggestion
        # NOTE: Ensemble failure metrics will reflect the failure counts of their composing models,
        # but currently do not capture the same granularity for the "reason" label.
        self._assert_metrics(
```

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -207,6 +233,13 @@ def test_infer_pymodel_error(self):
                     self.assertTrue(
                         False, "Wrong exception raised or did not raise an exception"
                     )
+        expected_count_increase = 1
+        self._assert_metrics(
+            model_name,
+            "BACKEND",
```
rmccorm4, line 239: Please (a) update the Failure Count row in https://github.com/triton-inference-server/server/blob/main/docs/user_guide/metrics.md to mention the current set of supported `reasons` and (b) add a small section describing each of the reasons in more detail and when they would be expected to be reflect. For example, "REJECTED" and "BACKEND" may not be immediately clear without explanation.

rmccorm4: (c) please also mention the caveat about ensemble metrics all being `OTHER` for now in that docs section, similar to this comment: https://github.com/triton-inference-server/server/pull/7423/files#r1673381047

rmccorm4 (commented):
Code LGTM overall, will save further nitpicks since this is planned to be cherry-picked. Let's get some docs additions in `metrics.md` with this PR to explain this feature so that the docs update can also be cherry-picked together with the feature. We can then point to the new docs in the release notes.

indrajit96:
Screenshot 2024-07-11 at 10 31 34 AM
Newly added docs

docs/user_guide/metrics.md
```diff
@@ -100,6 +100,19 @@ Count*. The count metrics are illustrated by the following examples:
...
+|--------------|----------------|------------|------------|
+|Failure Reason         |Rejected |REJECTED  | Number of inference failures due to request timeout in the schedular. |
+|              |Canceled |CANCELED  |  Number of inference failures due to request cancellation in the core. |
+|              |Backend  |BACKEND |  Number of inference failures due to errors from backend. |
+|              |Other | OTHER  | Number of inference failures due to unspecified reasons in the core. |
+
+
+> **Note**
+>
+> Ensemble failure metrics will reflect the failure counts of their composing models as well as the parent model, but currently do not capture the same granularity for the "reason" label and will default to the "OTHER" reason.
```
rmccorm4, line 115: Adding a few more details to try to get ahead of future customer questions/confusion.
```suggestion
> Ensemble failure metrics will reflect the failure counts of their composing models as well as the parent model, but currently do not capture the same granularity for the "reason" label and will default to the "OTHER" reason.
>
> For example, if EnsembleA contains ModelA, and ModelA experiences a failed request due to a queue/backlog timeout in the scheduler, ModelA will have a failed request metric reflecting `reason=REJECTED` and `count=1`.
> Additionally, EnsembleA will have a failed request metric reflecting `reason=OTHER` and `count=2`. 
> The `count=2` reflects 1 from the internally failed request captured by ModelA, as well as 1 from the failed top-level request sent to EnsembleA by the user/client.
> The `reason=OTHER` reflects that fact that the ensemble doesn't currently capture the specific reason why 
> ModelA's request failed at this time.
```

docs/user_guide/metrics.md
```diff
@@ -100,6 +100,19 @@ Count*. The count metrics are illustrated by the following examples:
 |              |Execution Count |`nv_inference_exec_count` |Number of inference batch executions (see [Inference Request Metrics](#inference-request-metrics), does not include cached requests)|Per model|Per request|
 |              |Pending Request Count |`nv_inference_pending_request_count` |Number of inference requests awaiting execution by a backend. This number is incremented when a request is enqueued to the server (`TRITONSERVER_ServerInferAsync`) and is decremented when a backend is about to start executing the request. More details can be found below. |Per model|Per request|
 
+#### Failure Count is further classified into four distinct types:
+
+|Sub-Category      |Sub-Metric | Sub-Metric Name |Description |
+|--------------|----------------|------------|------------|
+|Failure Reason         |Rejected |REJECTED  | Number of inference failures due to request timeout in the schedular. |
+|              |Canceled |CANCELED  |  Number of inference failures due to request cancellation in the core. |
+|              |Backend  |BACKEND |  Number of inference failures due to errors from backend. |
```
rmccorm4, line 109:
```suggestion
|              |Backend  |BACKEND |  Number of inference failures during execution of requests in the backend/model. |
```

docs/user_guide/metrics.md
```diff
@@ -100,6 +100,19 @@ Count*. The count metrics are illustrated by the following examples:
...
+|              |Other | OTHER  | Number of inference failures due to unspecified reasons in the core. |
```
rmccorm4, line 110:
```suggestion
|              |Other | OTHER  | Number of inference failures due to other uncategorized reasons in the core. |
```

docs/user_guide/metrics.md
```diff
@@ -100,6 +100,19 @@ Count*. The count metrics are illustrated by the following examples:
 |              |Execution Count |`nv_inference_exec_count` |Number of inference batch executions (see [Inference Request Metrics](#inference-request-metrics), does not include cached requests)|Per model|Per request|
 |              |Pending Request Count |`nv_inference_pending_request_count` |Number of inference requests awaiting execution by a backend. This number is incremented when a request is enqueued to the server (`TRITONSERVER_ServerInferAsync`) and is decremented when a backend is about to start executing the request. More details can be found below. |Per model|Per request|
 
+#### Failure Count is further classified into four distinct types:
```
rmccorm4, line 103:
```suggestion
#### Failure Count Categories

The `nv_inference_request_failure` metric is further categorized to track the types of failures encountered through the use of a `reason` label, described below.

File: docs/user_guide/metrics.md [hunk 11]
```diff
@@ -100,6 +100,19 @@ Count*. The count metrics are illustrated by the following examples:
...
+
+|Sub-Category      |Sub-Metric | Sub-Metric Name |Description |
```
rmccorm4, line 105: I think this can be condensed to 2 columns for now, something like:
---

|Failed Request Reason | Description |
|---|---|
| REJECTED | Number of ... |
| CANCELLED | Number of ... |
| BACKEND | Number of ... |
| OTHER | Number of ... |

rmccorm4 (approved):

Issue #7171, opened 2024-04-29
Input shape validation


HennerM:
**Is your feature request related to a problem? Please describe.**
Triton server doesn't do any validation of the data sent by clients. Specifically validation that the given shape matches the size of the input. For example if one client sends a vector with 15 elements and specifies the shape as [1,10], Triton blindly accepts this and passes it on to the backend, a similar issue arises if the client only sends a vector with 5 elements.

This could potentially lead to data leaking from one request into another one, an example that can trigger this behaviour is given here:

```python
FIXED_LAST_DIM = 5

alice = InferenceServerClient(TRITON_URL)
bob = InferenceServerClient(TRITON_URL)

input_data_1 = np.arange(FIXED_LAST_DIM + 2)[None].astype(np.float32)
print(f"{input_data_1=}")
inputs_1 = [
    InferInput("INPUT0", input_data_1.shape, np_to_triton_dtype(input_data_1.dtype)),
]
inputs_1[0].set_data_from_numpy(input_data_1)
# Usse the wrong dimension on purpose
inputs_1[0].set_shape((1, FIXED_LAST_DIM))

input_data_2 = 100 + np.arange(FIXED_LAST_DIM)[None].astype(np.float32)
print(f"{input_data_2=}")
inputs_2 = [InferInput("INPUT0", shape=input_data_2.shape, datatype=np_to_triton_dtype(input_data_2.dtype))]
inputs_2[0].set_data_from_numpy(input_data_2)

t1 = asyncio.create_task(alice.infer("dummy", inputs_1))
t2 = asyncio.create_task(bob.infer("dummy", inputs_2))

alice_result, bob_result = await asyncio.gather(t1, t2)
print(f"{alice_result.as_numpy('OUTPUT0')=}")
print(f"{bob_result.as_numpy('OUTPUT0')=}")
assert np.allclose(bob_result.as_numpy("OUTPUT0"), input_data_2)
```

With a example model config:
```pbtxt

name: "dummy"
backend: "pytorch"
max_batch_size: 8

input [
  {{
    name: "INPUT0"
    data_type: TYPE_FP32
    dims: [ 5 ]
  }}
]
output [
  {{
    name: "OUTPUT0"
    data_type: TYPE_FP32
    dims: [ 5 ]
  }}
]

dynamic_batching {{
    max_queue_delay_microseconds: 1000000
}}
```

This has been observed with the PyTorch backend, I am not sure if there are provisions in other backends in place.

[...]

PR #7427, opened 2024-07-09
feat: Client Input Byte Size Checks

yinggeh:
#### What does the PR do?
Add client input size check to make sure input shape byte size matches input data byte size.

#### Checklist

#### Commit Type:
- [ ] feat

#### Test plan:
n/a

- CI Pipeline ID:

#### Caveats:
Shared memory byte size checks for string inputs is not implemented.

#### Background
Stop malformed input request at client side before sending to the server.

Relates to https://github.com/triton-inference-server/server/issues/7171

### DIFF
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -149,8 +149,8 @@ RUN mkdir -p qa/common && \
     cp bin/triton_json_test qa/L0_json/. && \
     cp bin/backend_output_detail_test qa/L0_backend_output_detail/. && \
     cp -r deploy/mlflow-triton-plugin qa/L0_mlflow/. && \
-    cp bin/input_byte_size_test qa/L0_input_validation/. && \
-    cp -r docs/examples/model_repository/simple_identity qa/L0_input_validation/models
+    cp -r docs/examples/model_repository/{simple,simple_identity,simple_string} qa/L0_input_validation/models && \
+    cp bin/input_byte_size_test qa/L0_input_validation/.
 
 RUN mkdir -p qa/pkgs && \
     cp python/triton*.whl qa/pkgs/. && \
--- a/qa/L0_input_validation/input_validation_test.py
+++ b/qa/L0_input_validation/input_validation_test.py
@@ -34,7 +34,13 @@ import unittest
 import infer_util as iu
 import numpy as np
 import tritonclient.grpc as tritongrpcclient
-from tritonclient.utils import InferenceServerException, np_to_triton_dtype
+import tritonclient.http as tritonhttpclient
+import tritonclient.utils as utils
+from tritonclient.utils import (
+    InferenceServerException,
+    np_to_triton_dtype,
+    shared_memory,
+)
 
 
 class InputValTest(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


qa/L0_input_validation/test.sh
```diff
@@ -80,50 +83,6 @@ wait $SERVER_PID
...
-  {
-    name: "OUTPUT0"
-    data_type: TYPE_FP32
-    dims: [8]
-  }
-]
-# ensure we batch requests together
-dynamic_batching {
-    max_queue_delay_microseconds: 1000000
-}
```
rmccorm4, line 122: was this dynamic batching config needed for any test cases that are now missing?

Dockerfile.QA
```diff
@@ -149,8 +149,8 @@ RUN mkdir -p qa/common && \
     cp bin/triton_json_test qa/L0_json/. && \
     cp bin/backend_output_detail_test qa/L0_backend_output_detail/. && \
     cp -r deploy/mlflow-triton-plugin qa/L0_mlflow/. && \
-    cp bin/input_byte_size_test qa/L0_input_validation/. && \
-    cp -r docs/examples/model_repository/simple_identity qa/L0_input_validation/models
+    cp -r docs/examples/model_repository/{simple,simple_identity,simple_string} qa/L0_input_validation/models && \
```
rmccorm4, line 152: Do we have any documentation surrounding this byte size validation? It seems like there's a list of exceptions, edge cases, etc. for different input types, different memory types, different backend types (python backend had an issue if I recall correctly?), core vs client vs backend checks, etc. -- it would be good to have it written down somewhere on the expected behaviors and known issues. 

Not sure this needs to be a public-facing doc, but would be valuable for us to have a reference even if only internal.

rmccorm4 (commented):
Left a couple questions

qa/L0_input_validation/test.sh
yinggeh: No. I borrowed the example config file from https://github.com/triton-inference-server/server/issues/7171. I am changing it to existing model after realizing dynamic_batching is irrevalent. It was only to demonstrate the bug pointed out by the user.

Dockerfile.QA
yinggeh: Where is be the best place to document?

rmccorm4: Spoke offline

qa/L0_input_validation/input_validation_test.py
```diff
@@ -116,101 +118,113 @@ def test_input_validation_all_optional(self):
...
+
+            # Compromised input shapes
+            inputs[0].set_shape([2, 8])
+            inputs[1].set_shape([2, 8])
+
+            with self.assertRaises(InferenceServerException) as e:
+                triton_client.infer(model_name=model_name, inputs=inputs)
+            err_str = str(e.exception)
+            self.assertIn(
+                f"unexpected shape for input 'INPUT1' for model 'simple'. Expected [-1,16], got [2,8]",
```
rmccorm4, line 157: Is this a client-side error or core-side error? Client PR only refers to "unexpected element count" errors from what I can see.

If this is a core error message, please rename test to `test_core_input_validation` or something

rmccorm4: I would expect `1, 16` -> `2. 8` to pass an "element count" check where both have a product of 16, but fail a shape check.

yinggeh: This is the error from core (client does not know model_name). This test checks
1. request is not caught by the client check in the case of [1, 16] -> [2, 8].
2. request is caught by the client check in the case of [1, 16] -> [1, 8].

rmccorm4: Right, so if we're not testing client-side validation in this function, then let's rename `test_client_input_shape_validation` to `test_core_input_shape_validation`

qa/L0_input_validation/test.sh
```diff
@@ -80,49 +81,6 @@ wait $SERVER_PID
...
-EOL
-
-cp -r $DATADIR/qa_model_repository/graphdef_object_int32_int32 models/.
```
GuanLuo, line 125: Why removed?

qa/L0_input_validation/input_validation_test.py
```diff
@@ -116,101 +118,113 @@ def test_input_validation_all_optional(self):
...
+            identity_inference(triton_client, null_char_data, False)  # Using JSON data
+
+            # Example using BYTES input tensor with 16 elements, where each
+            # element is a 4-byte binary blob with value 0x00010203. Can use
+            # dtype=np.bytes_ in this case.
+            bytes_data = [b"\x00\x01\x02\x03" for i in range(16)]
+            np_bytes_data = np.array(bytes_data, dtype=np.bytes_)
+            np_bytes_data = np_bytes_data.reshape([1, 16])
+            identity_inference(triton_client, np_bytes_data, True)  # Using binary data
+            identity_inference(triton_client, np_bytes_data, False)  # Using JSON data
```
GuanLuo, line 227: What is this testing?

qa/L0_input_validation/test.sh
yinggeh: Use model "simple_identity" instead to test string inputs.

qa/L0_input_validation/input_validation_test.py
yinggeh: Copied from `client/src/python/examples/simple_http_string_infer_client.py`. It looks like the example demonstrated two ways of preparing string input data. I'll remove one of them.

qa/L0_input_validation/input_validation_test.py
yinggeh: Both checks are in `test_client_input_shape_validation` test case in order to share the same input. Do you want me to split them into two tests cases?

qa/L0_input_validation/input_validation_test.py
```diff
@@ -116,101 +118,113 @@ def test_input_validation_all_optional(self):
...
+
+            # Initialize the data
+            inputs[0].set_data_from_numpy(input0_data)
+            inputs[1].set_data_from_numpy(input1_data)
+
+            # Compromised input shapes
+            inputs[0].set_shape([2, 8])
+            inputs[1].set_shape([2, 8])
+
+            with self.assertRaises(InferenceServerException) as e:
```
rmccorm4, line 153:
```suggestion
            # If number of elements (volume) is correct but shape is wrong, the core will return an error.
            with self.assertRaises(InferenceServerException) as e:
```

qa/L0_input_validation/input_validation_test.py
rmccorm4: Let's just add a comment above the first check: https://github.com/triton-inference-server/server/pull/7427/files#r1705894309

yinggeh: Like it.

PR #7434, opened 2024-07-10
[24.07] Revert vllm to 0.5.0.post1

fpetrini15:
One of CI tests is failing with the latest version of vLLM. We will revert to 0.5.0.post1 which is known to be working until we can investigate further.


pvijayakrish (approved):
minor: Can you please include (in the MR overview) the reason why we did not upgrade to the latest version.

PR #7448, opened 2024-07-15
Update cleanup query for AWS buckets

mc-nv:
#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_backend_python/env/test.sh
+++ b/qa/L0_backend_python/env/test.sh
@@ -229,8 +229,9 @@ aws configure set default.region $AWS_DEFAULT_REGION && \
 BUCKET_URL="s3://triton-bucket-${CI_JOB_ID}"
 
 # Cleanup and delete S3 test bucket if it already exists (due to test failure)
-aws s3 rm $BUCKET_URL --recursive --include "*" && \
-    aws s3 rb $BUCKET_URL || true
+
+AWS_OUTDATED=$(date --date="$(date ) -10 days" +'%Y-%m-%d')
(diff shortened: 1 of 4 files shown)


kthui:
I think we can group the remove logic into a function like the [upload on GCS](https://github.com/triton-inference-server/server/blob/main/qa/common/util.sh#L471-L476), so it can be easily implemented by other AWS tests in the future.

mc-nv:
> I think we can group the remove logic into a function like the [upload on GCS](https://github.com/triton-inference-server/server/blob/main/qa/common/util.sh#L471-L476), so it can be easily implemented by other AWS tests in the future.

This change has nothing to the remove bucket general step. 
It's a cleanup step that is required in order to prevent reaching us limit of 100 buckets, safety trigger for CI. 
IMO make no sense to move into function at this point.

fpetrini15:
I do agree with Jacky that it would probably be better for maintenance purposes to consolidate these into a single function. That way there would only be a single point of change in the future if we need to modify/extend this. I can easily see someone forgetting to modify this command in 1+ locations.

mc-nv:
> I do agree with Jacky that it would probably be better for maintenance purposes to consolidate these into a single, common function. That way there would only be a single point of change in the future if we need to modify/extend this. I can easily see someone forgetting to modify this command in 1+ locations.

What is the sense to replace single line with a function? 
Each scenario is separate one, adding function won't increase readability and understanding of the scenario. 
Further more `aws-cli` is itself multifunctional tool, there why putting more complex tool into wrapper is kind of not the best practice IMO

PR #7449, opened 2024-07-15
chore: PA Migration From Client

fpetrini15:
**Goal:** Build PA separately from client and ensure that there is no impact on CI.

Client changes: https://github.com/triton-inference-server/client/pull/722


Dockerfile.sdk
```diff
@@ -114,26 +118,53 @@ ARG TARGETPLATFORM
...
-          -DTRITON_ENABLE_PERF_ANALYZER=ON \
+          -DTRITON_ENABLE_EXAMPLES=ON -DTRITON_ENABLE_TESTS=ON \
+          -DTRITON_ENABLE_GPU=${TRITON_ENABLE_GPU} /workspace/client
+RUN make -j16 cc-clients java-clients && \
+    rm -fr ~/.m2
+
+# TODO: PA will rebuild the CC clients since it depends on it.
+# This should be optimized so that we do not have to build
+# the CC clients twice. Similarly, because the SDK expectation is
+# that PA is packaged with the python client, we hold off on building
```
GuanLuo, line 142: Why it is packaged with Python client?

Dockerfile.sdk
```diff
@@ -114,26 +118,53 @@ ARG TARGETPLATFORM
...
           -DTRITON_ENABLE_CC_HTTP=ON -DTRITON_ENABLE_CC_GRPC=ON \
-          -DTRITON_ENABLE_PYTHON_HTTP=ON -DTRITON_ENABLE_PYTHON_GRPC=ON \
+          -DTRITON_ENABLE_PYTHON_HTTP=OFF -DTRITON_ENABLE_PYTHON_GRPC=OFF \
           -DTRITON_ENABLE_JAVA_HTTP=ON \
-          -DTRITON_ENABLE_PERF_ANALYZER=ON \
+          -DTRITON_ENABLE_EXAMPLES=ON -DTRITON_ENABLE_TESTS=ON \
+          -DTRITON_ENABLE_GPU=${TRITON_ENABLE_GPU} /workspace/client
+RUN make -j16 cc-clients java-clients && \
+    rm -fr ~/.m2
+
```
GuanLuo, line 138: Looks like a multi-stage build to me: 1. client libs; 2. PA / genai-perf

Dockerfile.sdk
```diff
@@ -114,26 +118,53 @@ ARG TARGETPLATFORM
...
 RUN cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
           -DTRITON_VERSION=`cat /workspace/TRITON_VERSION` \
           -DTRITON_REPO_ORGANIZATION=${TRITON_REPO_ORGANIZATION} \
           -DTRITON_COMMON_REPO_TAG=${TRITON_COMMON_REPO_TAG} \
           -DTRITON_CORE_REPO_TAG=${TRITON_CORE_REPO_TAG} \
           -DTRITON_THIRD_PARTY_REPO_TAG=${TRITON_THIRD_PARTY_REPO_TAG} \
+          -DTRITON_ENABLE_PERF_ANALYZER=OFF \
           -DTRITON_ENABLE_CC_HTTP=ON -DTRITON_ENABLE_CC_GRPC=ON \
-          -DTRITON_ENABLE_PYTHON_HTTP=ON -DTRITON_ENABLE_PYTHON_GRPC=ON \
+          -DTRITON_ENABLE_PYTHON_HTTP=OFF -DTRITON_ENABLE_PYTHON_GRPC=OFF \
```
GuanLuo, line 132: I think you can just turn it on and build it as that is what the client stage will look like at the end, avoiding rebuild can leave to the PA stage

Dockerfile.sdk
```diff
@@ -114,26 +118,53 @@ ARG TARGETPLATFORM
...
 
-WORKDIR /workspace/build
+WORKDIR /workspace/client_build
 RUN cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
           -DTRITON_VERSION=`cat /workspace/TRITON_VERSION` \
           -DTRITON_REPO_ORGANIZATION=${TRITON_REPO_ORGANIZATION} \
           -DTRITON_COMMON_REPO_TAG=${TRITON_COMMON_REPO_TAG} \
           -DTRITON_CORE_REPO_TAG=${TRITON_CORE_REPO_TAG} \
           -DTRITON_THIRD_PARTY_REPO_TAG=${TRITON_THIRD_PARTY_REPO_TAG} \
+          -DTRITON_ENABLE_PERF_ANALYZER=OFF \
```
GuanLuo, line 130: can remove, this is OFF by default.

Dockerfile.sdk
```diff
@@ -114,26 +118,53 @@ ARG TARGETPLATFORM
...
           -DTRITON_ENABLE_PERF_ANALYZER_TS=ON \
           -DTRITON_ENABLE_PERF_ANALYZER_OPENAI=ON \
-          -DTRITON_ENABLE_EXAMPLES=ON -DTRITON_ENABLE_TESTS=ON \
-          -DTRITON_ENABLE_GPU=${TRITON_ENABLE_GPU} /workspace/client
-RUN make -j16 cc-clients python-clients java-clients && \
-    rm -fr ~/.m2
+          -DTRITON_ENABLE_CC_HTTP=ON \
+          -DTRITON_ENABLE_CC_GRPC=ON \
+          -DTRITON_ENABLE_PYTHON_HTTP=ON \
+          -DTRITON_ENABLE_PYTHON_GRPC=ON \
```
GuanLuo, line 159: Should add some TODOs, I don't think these flags will be removed in the final version.

Dockerfile.sdk
fpetrini15: I am not sure what the original motivation was (likely convenience). However, this design has been baked into the python client AFAIK starting back in [2022](https://github.com/triton-inference-server/client/pull/98). If we want to break this expectation, that is a separate discussion.

Dockerfile.sdk
fpetrini15: I general, yes, however, it's complicated by two things:
1. The python client, as discussed above, has functionality to package PA in its pip wheel. Therefore, to guarantee its availability, the python client must be built after PA.
2. PA does (in the current PA repo PR) and should be able to build itself (including its client dependencies) without needing the client libs prebuilt.

Dockerfile.sdk
fpetrini15: I saw it was off by default, but wanted to explicitly mark it as off JIC it changed. Ultimately, this option will be removed entirely from the client CMakeLists.txt

Dockerfile.sdk
fpetrini15: We can build the python client here, however, we need to rebuild it as part of the PA build in order for the pip wheel to contain the PA-associated binaries. For this reason, I figured it was best to skip the python client build at this stage. Am I misunderstanding?

Dockerfile.sdk
fpetrini15: Can you clarify what TODOs are needed here? You're correct in that these flags will **not** be removed in the final version.

GuanLuo: Sorry, I meant "will be".. PA doesn't have dependency on Python client library, right?

GuanLuo (approved):
approving for now, appears that there are follow-up tasks for clean up and please make sure they are listed and tracked somewhere.

Dockerfile.sdk
fpetrini15: No, PA does not depend on the Python client. However, as per current user expectation, the Python client **does** depend on PA.

We may need to discuss as an org if we want to break this expectation now that PA will be its own repo.

PR #7476, opened 2024-07-26
test: Refactor cpu metrics tests to make L0_metrics more stable

rmccorm4:
The previous implementation of tests for the CPU/RAM related metrics were:
1. Unstable on the check for `nv_cpu_utilization` because there was no work going on in the background, making duplicate values appear fairly frequently.
2. Written in bash

This PR:
1. Adds server inference requests in the background to keep the server busy and show some CPU utilization variation during metric observation
2. Moves to python for more flexibility and ease of testing
3. Update the newer styled tests (pinned_memory_metrics, cpu_metrics) to use the new `pytest` style that generates nice reports on the gitlab side for easier regression detection at the unit test level. I also noticed the use of `check_test_results` without using the `tu.ResultCollector` for the pinned memory metrics tests section - so it wasn't actually checking the right test results file anyway, since none was being generated. I moved this one to the pytest style to be consistent with the similar cpu metrics test.
4. Make some windows-friendly changes to the L0_metrics tests based on Francesco's review comments like use of `kill_server` and `TRITONSERVER_IPADDR`.
5. Reduces 4GPU CI runner requirement to 2GPU requirement by only checking 2-gpus instead of 3-gpus in the gpu metrics section. I think the test is equivalent regardless, but there are significantly more 2gpu runners available in CI than 4gpu runners to cut down on job queue times. I'll push a gitlab-side PR to update accordingly, but it will still work as-is with these changes.

Average results of the PR from testing so far typically makes about 20 observations (1 per second for 20 second test), and finds no duplicate values. However, keeping the duplicate tolerance will help stabilize in the event we occasionally find some:
image


qa/L0_metrics/cpu_metrics_test.py
```diff
@@ -0,0 +1,187 @@
...
+import threading
+import time
+import unittest
+from collections import defaultdict
+
+import numpy as np
+import requests
+import tritonclient.http as httpclient
+
+CPU_UTILIZATION = "nv_cpu_utilization"
```
fpetrini15, line 38: I had a good discussion with Jacky a while back on a different PR about how we should name constants and where they should be stored. The rule for python seems to be to name CONSTANTS with all caps. However, we also use this convention when reading environment variables (e.g., `os.environ.get("TRITONSERVER_IPADDR")`). Jacky also had a preference to store variables like this within the class definition. Link to full [discussion.](https://github.com/triton-inference-server/server/pull/6830#discussion_r1473665680). Wanted to know your thoughts on it.

qa/L0_metrics/cpu_metrics_test.py
```diff
@@ -0,0 +1,187 @@
...
+CPU_USED_MEMORY = "nv_cpu_memory_used_bytes"
+CPU_TOTAL_MEMORY = "nv_cpu_memory_total_bytes"
+
+
+def get_metrics():
+    utilization_pattern = re.compile(rf"{CPU_UTILIZATION} (\d+\.?\d*)")
+    used_bytes_pattern = re.compile(rf"{CPU_USED_MEMORY} (\d+)")
+    total_bytes_pattern = re.compile(rf"{CPU_TOTAL_MEMORY} (\d+)")
+
+    r = requests.get("http://localhost:8002/metrics")
```
fpetrini15, line 48: Torn here, but as part of Windows GA we will need to obtain and test CPU metrics--so we will need to run this test. `localhost` is not a friendly term in our Windows CI. Instead we usually follow the convention: 
```python
_tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR", "localhost")
r = requests.get(f"http://{_tritonserver_ipaddr}:8002/metrics")
```
I'm torn because while it would be good to lay this groundwork early, we may also be shifting our Windows testing strategy in the near future.

qa/L0_metrics/test.sh
```diff
@@ -111,6 +111,23 @@ mkdir -p $MODELDIR/${model}/1 && \
...
+# NOTE: CPU utilization is computed based on the metrics interval, so having
+# too small of an interval can skew the results.
+SERVER_ARGS="$BASE_SERVER_ARGS --metrics-interval-ms=1000 --log-verbose=1"
+run_and_check_server
+
+CLIENT_LOG="cpu_metrics_test_client.log"
+python3 ${CLIENT_PY} -v 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+
+kill $SERVER_PID
```
fpetrini15, line 127: Same here regarding Windows GA. 
```bash
kill $SERVER_PID
wait $SERVER_PID
```
is not Windows friendly. To handle shutdown for our joint Windows/Linux tests, we use:
```bash
kill_server
```
Which is a common function that handles shutdown platform agnostically.

qa/L0_metrics/cpu_metrics_test.py
rmccorm4: I have no issue adding a line if it makes it easier for windows later. Note that this is mostly stolen from `pinned_memory_metrics_test.py` which will have the same issues

fpetrini15: Right, that's another reason I'm torn--it'll obviously need to be fixed in other places later on as well. But I guess this could be an opportunity to lessen some of the tech debt. Your call.

qa/L0_metrics/cpu_metrics_test.py
```diff
@@ -0,0 +1,187 @@
...
+            metrics_thread.join()
+
+        self._validate_metric_variance(observed_metrics)
+
+    def test_cpu_metrics_ranges(self):
+        # Test some simple sanity checks on the expected ranges of values
+        # for the CPU related metrics.
+        utilization, used_memory, total_memory = get_metrics()
+        self.assertGreaterEqual(utilization, 0)
+        self.assertLessEqual(utilization, 1.0)
```
fpetrini15, line 178: Also just a reminder that Python can support
```python
self.assertTrue(0 <= utilization <= 1.0)
```
Totally not necessary to change though.

qa/L0_metrics/cpu_metrics_test.py
rmccorm4: Updated python tests to use this convention.

qa/L0_metrics/cpu_metrics_test.py
rmccorm4: I'm not super opinionated on it. Personally I'd prefer:
```python
TRITONSERVER_IPADDR = os.environ.get("TRITONSERVER_IPADDR")
CPU_UTILIZATION = "nv_cpu_utilization
```
over this for consistency purposes:
```
_tritonserver_ipaddr = os.environ.get("TRITONSERVER_IPADDR")
CPU_UTILIZATION = "nv_cpu_utilization
```

I view `tritonserver_ipaddr` as a constant if only set once for the life of the test - but I wouldn't fight for one way over the other.

For now, I followed the example set by other scripts for easier find/replace in the future if we change our mind to reduce the number of special cases.

rmccorm4: Setting things within the class is a good idea too - but in this case I took an example that was using a method defined outside of the class, so I kept it constant. You could argue to move the helper method within the class too - but I don't think it will move the needle much one way or the other

rmccorm4: (note that this test is heavily copied from `pinned_memory_metrics_test.py` so I tried to keep it as similar as possible. It should make further decisions to change things easier to apply in multiple places.

qa/L0_metrics/test.sh
```diff
@@ -111,35 +111,60 @@ mkdir -p $MODELDIR/${model}/1 && \
...
+# NOTE: CPU utilization is computed based on the metrics interval, so having
+# too small of an interval can skew the results.
+SERVER_ARGS="$BASE_SERVER_ARGS --metrics-interval-ms=1000 --log-verbose=1"
+run_and_check_server
+
+CLIENT_PY="./cpu_metrics_test.py"
+CLIENT_LOG="cpu_metrics_test_client.log"
+python3 -m pytest --junitxml="cpu_metrics.report.xml" ${CLIENT_PY} >> ${CLIENT_LOG} 2>&1
+if [ $? -ne 0 ]; then
+    echo -e "\n***\n*** ${CLIENT_PY} FAILED. \n***"
```
fpetrini15, line 126: What's our rule regarding whether we `cat` the client/server logs on failure? I see mixed usage across our tests, but think having failure logs built-in to the pipeline regardless of artifact expiration is valuable.

qa/L0_metrics/test.sh
```diff
@@ -111,35 +111,60 @@ mkdir -p $MODELDIR/${model}/1 && \
...
+    echo -e "\n***\n*** Custom ${CLIENT_PY} FAILED. \n***"
+    RET=1
+fi
 
-kill $SERVER_PID
-wait $SERVER_PID
+kill_server
 set -e
 
+exit 0
```
fpetrini15, line 166:
```suggestion
```

qa/L0_metrics/cpu_metrics_test.py
fpetrini15: Great explanation!

qa/L0_metrics/test.sh
rmccorm4: Nice shout - I'm pro dumping logs to console in case artifacts expire.

qa/L0_metrics/test.sh
rmccorm4: Nice catch :bow:

fpetrini15 (dismissed):
TmLg! Great work Ryan! Look forward to seeing the greener pipelines 🚀 

And thanks for fixing things up to be Windows CI friendly!

krishung5 (approved):
LGTM!

fpetrini15 (approved):
LGTM, great work!

PR #7486, opened 2024-07-30
test: Improve L0_logging stability

rmccorm4:
#### What does the PR do?

Adds more graceful handling to server startup given the way it's launched in the L0_logging test.

Current test is failing intermittently with call to `client.is_server_ready()` raising an `Exception` on a failed connection. This can happen if the frontends aren't ready and accepting requests yet while server is still starting up.

This PR:
1. Gives more time for natural server startup
2. Adds better handling for exceptions raised by client on failing to connect 
3. Exits early if `DATADIR` isn't set correctly to smooth local testing

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Test plan:

- `L0_logging`

### DIFF
--- a/qa/L0_logging/test.sh
+++ b/qa/L0_logging/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2022-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2022-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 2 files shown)


qa/L0_logging/test.sh
```diff
@@ -60,6 +60,11 @@ source ../common/util.sh
 rm -f *.log
 rm -fr $MODELSDIR && mkdir -p $MODELSDIR
 
+if [ ! -d ${DATADIR} ]; then
```
krishung5, line 63: Is this check added just to let us know if the model repository in the dlcluster is not setup correctly? or if there's other cases that we want to catch?

krishung5 (approved):

qa/L0_logging/test.sh
rmccorm4: I ran the test locally with a non-existing datadir and the test ran and failed much later with obscure errors -- so just to catch low hanging fruit later on for future debuggers

PR #7499, opened 2024-08-05
feat: Add GRPC error codes to GRPC streaming if enabled by user.

indrajit96:
#### What does the PR do?
Enhanced GRPC streaming to enable sending GRPC error codes if enabled by user.
User can enabled/disable this feature by including "grpc_strict":true as part of GRPC headers.

#### Checklist
- [X] Related issues are referenced.
- [] Verified that the PR passes existing CI.
- [] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [X] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [Y] test

#### Related PRs:
None

ExtractStateFromHeaders() function and Process() function in stream_infer_handler.cc

#### Test plan:
- Added to qa/L0_backend_python/lifecycle/lifecycle_test.py
- L0_decoupled
- L0_grpc
- L0_grpc_state_cleanup
- L0_request_cancellation
- L0_secure_grpc

#### Background
Amazon request to have GRPC error codes as part of GRPC response in case of errors from core
TEP link : https://docs.google.com/document/d/1TfNAMYLsPuLrduBtAKo64YqV55IWQhcn6zayc_CBY18/edit?pli=1


src/grpc/stream_infer_handler.cc
```diff
@@ -671,14 +682,29 @@ ModelStreamInferHandler::StreamInferResponseComplete(
...
       failed = true;
       ::grpc::Status status;
+      // Converts CORE errors to GRPC error codes
       GrpcStatusUtil::Create(&status, err);
       response->mutable_infer_response()->Clear();
       response->set_error_message(status.error_message());
       LOG_VERBOSE(1) << "Failed for ID: " << log_request_id << std::endl;
+      if (state->context_->grpc_strict_) {
+        // Set to finish
+        // std::lock_guard<std::recursive_mutex> lock(state->grpc_strict_mtx_);
```
rmccorm4, line 695: needed?

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,6 +241,70 @@ def test_infer_pymodel_error(self):
             initial_metrics_value,
         )
 
+    def test_grpc_strict_error_on(self):
```
rmccorm4, line 244: Added some more tests to PR test plan description, just an FYI of some possibly impacted tests

src/grpc/stream_infer_handler.cc
```diff
@@ -596,7 +601,13 @@ ModelStreamInferHandler::StreamInferResponseComplete(
     void* userp)
 {
   State* state = reinterpret_cast<State*>(userp);
-
+  // Ignore Response from CORE in case GRPC Strict as we dont care about
+  if (state->context_->grpc_strict_) {
+    std::lock_guard<std::recursive_mutex> lock(state->context_->mu_);
+    if (state->context_->IsGRPCStrictError()) {
```
rmccorm4, line 607: Do you need both `grpc_stream_errror_state_` to be atomic and using `state->context_->mu_` everywhere you're accessing/modifying?

src/grpc/infer_handler.h
```diff
@@ -669,6 +670,33 @@ class InferHandlerState {
...
+        auto& key = pair.first;
+        std::string param_key = std::string(key.begin(), key.end());
+        std::string grpc_strict_key = "grpc_strict";
+        if (param_key.compare(grpc_strict_key) == 0) {
+          state->context_->grpc_strict_ = true;
+        }
+      }
+    }
+
+    void sendGRPCStrictResponse(InferHandlerStateType* state)
```
rmccorm4, line 687: I think style-wise it should be `SendGRPC` not `sendGRPC` for consistency

src/grpc/infer_handler.h
```diff
@@ -669,6 +670,33 @@ class InferHandlerState {
...
+    {
+      std::lock_guard<std::recursive_mutex> lock(state->context_->mu_);
+      // Check if Error not responded previously
+      // Avoid closing connection twice on multiple errors from core
+      if (!state->context_->IsGRPCStrictError()) {
+        state->context_->responder_->Finish(state->status_, state);
+        // Mark error for this stream
+        state->context_->MarkGRPCStrictError();
+        // Fix Me : Last argument not sure for HandleCancellation
```
rmccorm4, line 696: Still a FIXME?

src/grpc/stream_infer_handler.cc
```diff
@@ -140,6 +140,17 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
   // synchronizing this thread and the ResponseComplete
   // threads.
+  // We need an explicit finish indicator. Can't use 'state->step_'
+  // because we launch an async thread that could update 'state's
+  // step_ to be FINISH before this thread exits this function.
+  bool finished = false;
+  if (state->context_->grpc_strict_) {
+    std::lock_guard<std::recursive_mutex> lock(state->context_->mu_);
+    // Check if stream error detected and already connection ended
+    if (state->context_->IsGRPCStrictError()) {
```
rmccorm4, line 150: Does this require both if statements? Or does `IsGRPCStrictError()` returning `true` require by `grpc_strict_` being true?

qa/L0_backend_python/lifecycle/my_test.py
indrajit96, line 1: Local test pushed by mistake.
Will remove

src/grpc/infer_handler.h
```diff
@@ -779,6 +807,9 @@ class InferHandlerState {
             // The RPC is complete and no callback will be invoked to retrieve
             // the object. Hence, need to explicitly place the state on the
             // completion queue.
+            LOG_VERBOSE(1)
+                << "PutTaskBackToQueue inside IssueRequestCancellation for "
+                << state->unique_id_;
```
rmccorm4, line 812: needed?

qa/python_models/execute_grpc_error/config.pbtxt
```diff
@@ -0,0 +1,52 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "execute_grpc_error"
```
rmccorm4, line 27: It will generally be more convenient down the line to not set the `name` field in case this model is re-used in other test cases, but needs to be renamed. If left unspecified, it will default to the name of the model folder it is in.
```suggestion
```

src/grpc/infer_handler.h
```diff
@@ -669,6 +670,33 @@ class InferHandlerState {
...
+    {
+      std::lock_guard<std::recursive_mutex> lock(state->context_->mu_);
+      // Check if Error not responded previously
+      // Avoid closing connection twice on multiple errors from core
+      if (!state->context_->IsGRPCStrictError()) {
+        state->context_->responder_->Finish(state->status_, state);
+        // Mark error for this stream
+        state->context_->MarkGRPCStrictError();
+        // Fix Me : Last argument not sure for HandleCancellation
+        state->context_->HandleCancellation(state, true, "grpc_strict_name");
```
rmccorm4, line 697:
```suggestion
        state->context_->HandleCancellation(state, true /* rpc_ok */, "grpc_strict_name");
```

src/grpc/stream_infer_handler.cc
```diff
@@ -140,6 +140,17 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
+      return finished;
```
rmccorm4, line 151: For better or worse, I believe we generally return `!finished` (inverse) everywhere throughout the code, or a literal `false/true`. 

Is this line returning `finished` correct?

rmccorm4, line 151: Do you need to set `state->step_ = Steps::FINISH;` here for any dependent cleanup logic?

rmccorm4: It looks like it would affect the flow in the main loop https://github.com/triton-inference-server/server/blob/9ad856c4f67e62226eff40d757e6181ddf97c9a2/src/grpc/stream_infer_handler.cc#L149 and whats returned by this HandleCompletion func: https://github.com/triton-inference-server/server/blob/9ad856c4f67e62226eff40d757e6181ddf97c9a2/src/grpc/infer_handler.h#L693, possibly putting it back on completion queue

src/grpc/stream_infer_handler.cc
indrajit96: I need grpc_stream_errror_state_ to be thread-safe so need the mutex.
Reverting grpc_stream_errror_state_ to bool

src/grpc/infer_handler.h
indrajit96: Yes HandleCancellation needs a string name_, which it uses for logging.
As I have added this inside stream_infer_handler.cc which does not have a Name() function, hence "grpc_strict_name" as a placeholder

src/grpc/stream_infer_handler.cc
indrajit96: grpc_strict_ is a bool per stream set by the user when they begin the stream, indicating that they want to use this feature. Once set it is true throughout the stream.
IsGRPCStrictError() is a function that checks if the particular stream has encountered an error from the core previously. We need to maintain this to avoid double close of the stream.
Hence need locking around it.

qa/python_models/execute_grpc_error/config.pbtxt
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2024, NVIDIA CORPORATION. All rights reserved.
```
Tabrizian, line 1: Is it possible to move the model to the `L0_backend_python/lifecycle` folder? I think it might be easier this way to track which test the models belong to.

src/grpc/infer_handler.h
```diff
@@ -669,6 +670,33 @@ class InferHandlerState {
...
 
+    // Extracts headers from GRPC request and updates state
+    void ExtractStateFromHeaders(InferHandlerStateType* state)
+    {
+      const auto& metadata = state->context_->ctx_->client_metadata();
+      for (const auto& pair : metadata) {
+        auto& key = pair.first;
+        std::string param_key = std::string(key.begin(), key.end());
+        std::string grpc_strict_key = "grpc_strict";
+        if (param_key.compare(grpc_strict_key) == 0) {
```
Tabrizian, line 681: nit:
```suggestion
        if (param_key == grpc_strict_key) {
```

src/grpc/infer_handler.h
```diff
@@ -669,6 +670,33 @@ class InferHandlerState {
...
     }
 
+    // Extracts headers from GRPC request and updates state
+    void ExtractStateFromHeaders(InferHandlerStateType* state)
+    {
+      const auto& metadata = state->context_->ctx_->client_metadata();
+      for (const auto& pair : metadata) {
+        auto& key = pair.first;
+        std::string param_key = std::string(key.begin(), key.end());
+        std::string grpc_strict_key = "grpc_strict";
```
Tabrizian, line 680: Can we name this parameter "triton_grpc_error"?

Tabrizian: Is it possible to move this value to constants.h?

qa/python_models/execute_grpc_error/config.pbtxt
indrajit96: We don't have any models or model folder in
https://github.com/triton-inference-server/server/tree/main/qa/L0_backend_python/lifecycle
In test.sh we copy over from python_models and create the models folder with versions inside.
I have kept this new model in parallel to existing models used in `L0_backend_python/lifecycle`

src/grpc/infer_handler.h
indrajit96: Unable to find a constants.h in server.
Could you help point me to it, maybe I am missing something or vscode is unable to find it.

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,13 +241,13 @@ def test_infer_pymodel_error(self):
...
 
-    def test_grpc_strict_error_on(self):
+    def test_triton_grpc_error_error_on(self):
         model_name = "execute_grpc_error"
         shape = [2, 2]
         number_of_requests = 2
         user_data = UserData()
         triton_client = grpcclient.InferenceServerClient(f"{_tritonserver_ipaddr}:8001")
-        metadata = {"grpc_strict": "true"}
+        metadata = {"triton_grpc_error": "true"}
```
rmccorm4, line 250: 1. Can we add docs on this feature?
2. Was there any discussion on using headers vs request parameters etc for it?

indrajit96: There is a draft TEP (Was not required, I thought would be a good idea) https://docs.google.com/document/d/1TfNAMYLsPuLrduBtAKo64YqV55IWQhcn6zayc_CBY18/edit?pli=1
headers seemed logical after discussions with the SA on all possible options.
Not sure how we would do it via request params.
We need the triton_grpc_error flag set only once while the stream starts so headers seemed logica also has no backward compatibility issues.

src/grpc/infer_handler.h
```diff
@@ -746,7 +792,7 @@ class InferHandlerState {
 
     // Issues the cancellation for all inflight requests
     // being tracked by this context.
-    void IssueRequestCancellation()
+    void IssueRequestCancellation(bool is_triton_grpc_error)
```
rmccorm4, line 795: is this arg unused?

src/grpc/infer_handler.h
```diff
@@ -664,11 +665,56 @@ class InferHandlerState {
 
     bool ReceivedNotification() { return received_notification_; }
 
+    // Returns true ONLY when GRPC_ERROR from CORE is waiting to be processed.
+    bool IsGRPCError()
```
rmccorm4, line 669: This function name feels a little unintuitive that is has a side effect of changing the `grpc_stream_error_state_`.

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,6 +242,145 @@ def test_infer_pymodel_error(self):
...
+                input_data = np.random.randn(*shape).astype(np.float32)
+                input_datas.append(input_data)
+                inputs = [
+                    grpcclient.InferInput(
+                        "IN", input_data.shape, np_to_triton_dtype(input_data.dtype)
+                    )
+                ]
+                inputs[0].set_data_from_numpy(input_data)
+                triton_client.async_stream_infer(model_name=model_name, inputs=inputs)
+                result = user_data._completed_requests.get()
```
oandreeva-nv, line 379: potentially check, that every even `result` has an error as expected and every `odd` does not

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,6 +242,145 @@ def test_infer_pymodel_error(self):
...
+                response_counter += 1
+        # Expect stream to not CLOSE as NOT in triton_grpc_error mode
```
oandreeva-nv, line 381: Let's clarify this comment as something like: we expect `response_counter` == `number_of_requests`, which indicates that after the first reported grpc error stream did NOT close and mode != triton_grpc_error

qa/python_models/execute_grpc_error/config.pbtxt
oandreeva-nv: you can always create a `models` subfolder under `L0_*` tests

src/grpc/stream_infer_handler.cc
oandreeva-nv: should we limit the scope of the lock to only `if (state->context_->IsGRPCStrictError()) { ... } ` part?

src/grpc/stream_infer_handler.cc
```diff
@@ -189,7 +189,7 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
       state->context_->responder_->Finish(status, state);
       return !finished;
     }
-
+    state->context_->ExtractStateFromHeaders(state);
```
oandreeva-nv, line 192: Can this be treated as a part of `ExecutePrecondition` logic, or we would like to keep it separate?

src/grpc/infer_handler.h
```diff
@@ -664,11 +665,63 @@ class InferHandlerState {
...
+    void ExtractStateFromHeaders(InferHandlerStateType* state)
+    {
+      const auto& metadata = state->context_->ctx_->client_metadata();
+      for (const auto& pair : metadata) {
+        auto& key = pair.first;
+        auto& value = pair.second;
+        std::string param_key = std::string(key.begin(), key.end());
+        std::string value_key = std::string(value.begin(), value.end());
+        std::string triton_grpc_error_key = "triton_grpc_error";
+        if (param_key == triton_grpc_error_key) {
```
oandreeva-nv, line 703: I think you can use something like: `metadata.find(...)`:
https://github.com/triton-inference-server/server/blob/53200091b84f08a5e4921f5073137784570283e9/src/grpc/infer_handler.h#L1466

qa/python_models/execute_grpc_error/config.pbtxt
Tabrizian: thanks Olga, that's what I meant.

src/grpc/infer_handler.h
Tabrizian: You are right, it looks like we only have it for core.

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,6 +242,145 @@ def test_infer_pymodel_error(self):
...
+        shape = [2, 2]
+        number_of_requests = 2
+        user_data = UserData()
+        triton_client = grpcclient.InferenceServerClient(f"{_tritonserver_ipaddr}:8001")
+        metadata = {"triton_grpc_error": "true"}
+        triton_client.start_stream(
+            callback=partial(callback, user_data), headers=metadata
+        )
+        stream_end = False
+        with self._shm_leak_detector.Probe() as shm_probe:
```
Tabrizian, line 258: Let's not use shm leak prob as it can make the test intermittent. We only need shm leak prob if we're adding Python backend features.

src/grpc/grpc_utils.h
```diff
@@ -76,6 +76,17 @@ typedef enum {
...
+typedef enum {
+  // No error from CORE seen yet
+  NONE,
+  // Error from CORE encountered, waiting to be picked up by completion queue to
+  // initiate cancellation
+  ERROR_WAITING,
+  // Error from CORE encountered, stream closed
+  // This state is added to avoid double cancellation
+  ERROR_CANCELED
+} Triton_grpc_error_steps;
```
Tabrizian, line 88:
```suggestion
} TritonGRPCErrorSteps;
```

Tabrizian:
I defer the final approval to @tanmayv25 / @kthui who may have more context on this feature request.

src/grpc/stream_infer_handler.cc
indrajit96: Makes sense to include as part of `ExecutePrecondition`

src/grpc/infer_handler.h
tanmayv25, line 1: You can clean-up the code by writing a new class "gRPCErrorTracker" and collect all these functions as part of this class. 
The context can hold the object of this class.

qa/python_models/execute_grpc_error/config.pbtxt
indrajit96: In the existing test.sh we do
`rm -fr *.log ./models` before we start the test. 
https://github.com/triton-inference-server/server/blob/53200091b84f08a5e4921f5073137784570283e9/qa/L0_backend_python/lifecycle/test.sh#L37
For me to add models to `L0_backend_python/lifecycle` I will need to remove this.
We cleanup the models before every test, I will need to remove all the instances where we delete the models folder.
https://github.com/triton-inference-server/server/blob/53200091b84f08a5e4921f5073137784570283e9/qa/L0_backend_python/lifecycle/test.sh#L103
https://github.com/triton-inference-server/server/blob/53200091b84f08a5e4921f5073137784570283e9/qa/L0_backend_python/lifecycle/test.sh#L170
Not sure if I should change existing design might impact other tests too.

qa/L0_backend_python/lifecycle/lifecycle_test.py
indrajit96: I started off with this approach in my tests, but because the model using batch scheduling of size 8 and we can't be sure which test_ would run in which order it's not possible to know is your odd/even request from client is odd/even in the model.py.
Hence just checking status and checking if stream ends as soon as error is reported was a better approach

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,6 +242,141 @@
...
+            input_data = np.random.randn(*shape).astype(np.float32)
+            input_datas.append(input_data)
+            inputs = [
+                grpcclient.InferInput(
+                    "IN", input_data.shape, np_to_triton_dtype(input_data.dtype)
+                )
+            ]
+            inputs[0].set_data_from_numpy(input_data)
+            triton_client.async_stream_infer(model_name=model_name, inputs=inputs)
+            result = user_data._completed_requests.get()
```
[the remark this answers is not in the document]
oandreeva-nv: as a trick, you can do 
`_ = user_data._completed_requests.get()` 
if you don 't plan to use `result`

qa/L0_backend_python/lifecycle/lifecycle_test.py
tanmayv25, line 1: Why are these tests added as part of L0_backend_python? Shouldn't these be better fit for L0_grpc?

qa/L0_backend_python/lifecycle/lifecycle_test.py
```diff
@@ -241,6 +242,141 @@ def test_infer_pymodel_error(self):
...
+        shape = [2, 2]
+        number_of_requests = 2
+        user_data = UserData()
+        triton_client = grpcclient.InferenceServerClient(f"{_tritonserver_ipaddr}:8001")
+        metadata = {"triton_grpc_error": "true"}
+        triton_client.start_stream(
+            callback=partial(callback, user_data), headers=metadata
+        )
+        stream_end = False
+        input_datas = []
```
tanmayv25, line 258: Why do you need `input_datas` variable?

qa/L0_decoupled/test.sh
tanmayv25, line 1: I think adding similar variant for L0_grpc_state_clean_up will also be useful.

qa/L0_backend_python/lifecycle/lifecycle_test.py
tanmayv25: Please take a look at this in a future PR.

tanmayv25 (dismissed):

qa/python_models/execute_grpc_error/config.pbtxt
oandreeva-nv: name the folder differently and copy from it to `models` ?

indrajit96: @tanmayv25 suggesnted I move all the changes from L0_lifecycle to a new L0_ dedicated for this feature going forward.
The reason I made changes here was, L0_lifecycle/models had models where we send errors pragmatically helped me reuse all the models.
Will resolve this comment along with the original comment by @tanmayv25 in a new PR after the cherrypick.
Keeping it unresolved for now will mark this resolved after the new PR

tanmayv25 (approved):

PR #7500, opened 2024-08-05
build: update build.py to pass versions as input parameter and convert version map to dictionary

nvda-mesharma:
#### What does the PR do?
1. Replace TRITON_VERSION_MAP with a dictionary of key and values instead of remembering indexes throughout the script.
2. Add ability to pass backend version (start with vllm version) instead of hard-coding in the build.py
    example - https://github.com/triton-inference-server/vllm_backend/blob/mesharma-ci/ci/build/build_source.sh#L60

#### Checklist
- [ x ] PR title reflects the change and is of format `<commit_type>: <Title>`
- [ x ] Changes are described in the pull request.
- [ x ] Related issues are referenced.
- [ x ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ x ] Added [test plan](#test-plan) and verified test passes.
- [ x ] Verified that the PR passes existing CI.
- [ x] Verified copyright is correct on all changed files.
- [ x] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ x] All template sections are filled out.

#### Commit Type:
- [ x ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
[Not ready for review, not needed to merge this change] https://github.com/triton-inference-server/vllm_backend/pull/49 

all changes are contained within build.py script

#### Test plan:

1. Verified that the changes work with vllm_backedn ci build.
    https://github.com/triton-inference-server/vllm_backend/pull/49
    All actions are successful and pipeline checks have also passed.
2. This is a no-op on release build process

#### CI Pipeline ID:
17157949

#### Caveats:
N/A

#### Background
I was unable to set any specific vllm version while triggering a build and test CI pipeline from vllm_backend repository due to the hard-coding in this build.py script. 

N/A


oandreeva-nv:
Please, follow the PR template :
```
#### What does the PR do?
<!-- Describe your pull request here. Please read the text below the line, and make sure you follow the checklist.-->

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
<!-- Related PRs from other Repositories -->

<!-- call out specific files that should be looked at closely -->

#### Test plan:
<!-- list steps to verify -->
<!-- were e2e tests added?-->

- CI Pipeline ID:
<!-- Only Pipeline ID and no direct link here -->

#### Caveats:
<!-- any limitations or possible things missing from this PR -->

#### Background
<!-- e.g. what led to this change being made. this is optional extra information to help the reviewer -->

- closes GitHub issue: #xxx
```

build.py
```diff
@@ -68,16 +68,15 @@
...
-    "2.49.0dev": (
-        "24.08dev",  # triton container
-        "24.07",  # upstream container
-        "1.18.1",  # ORT
-        "2024.0.0",  # ORT OpenVINO
-        "2024.0.0",  # Standalone OpenVINO
-        "3.2.6",  # DCGM version
-        "0.5.3.post1",  # vLLM version
-    )
+DEFAULT_TRITON_VERSION_MAP = {
```
mc-nv, line 71: Willing to understand motivation behind this utility better also wondering if it pass all other backends?

nvda-mesharma: While updating code I found it very hard to refer to indexes. Prefer using key names to retrieve values instead of indices.
Please let me know how to verify other backend builds / any other verifications I should do here

mc-nv: You need to start the general purpose pipeline with all enabled flags against current branch.

nvda-mesharma: thanks, started 17226628

nv-kmcgill53:
Not asking for change on this just bringing up a limitation of python. There is no idea of `const` in the language and that was the point of using a tuple instead of a map. tuples being immutable expressed that this collection of versions should be used together. But someone could just overwrite the original tuple anyway to get around this anyway.

Again, not rejecting the change, just pointing out the reason for the previous implementation in case this is important for some reason I have not thought of.

build.py
```diff
@@ -2362,6 +2343,12 @@ def enable_all():
         required=False,
         help="Override specified backend CMake argument in the build as <backend>:<name>=<value>. The argument is passed to CMake as -D<name>=<value>. This flag only impacts CMake arguments that are used by build.py. To unconditionally add a CMake argument to the backend build use --extra-backend-cmake-arg.",
     )
+    parser.add_argument(
```
oandreeva-nv, line 2346: @mc-nv Do we have a similar practice of setting library versions through`build.py` args?

nvda-mesharma: yes, we can keep adding. This is also meant to serve as an good example to do that going forward

mc-nv: What do you mean under "library versions"?

build.py
nvda-mesharma: all build steps are passing on this pipeline

build.py
oandreeva-nv: This `add_argument` adds `vllm-version`, i.e. with this PR we would be able to specify vllm library version through `build.py` argument. IIRC, we didn't allow anything like this, but rather had a data structure with specified versions

mc-nv: I have no answer to it, I would like to have a consistency and if we give control under one value trough the argument parser then make sense to extend functionality for any other argument value in the data structure.  

By giving version control to `build.py` with argument, we created precedent: 
- any issue with build, must include check between version in `build.py` and with argument `vllm-version` given to it during the execution.

I have no guidance on it, I don't like wrappers at all and would  rather get rid of it. 
Current scenario brings kind of loop back from `vllm_backend` to `server`. 

From build stand point as long as `vllm_backend` has nothing to include from server, it should be built separately.

oandreeva-nv: I would like to have a consistency as well, thus I recommend to either delete this new flag, or handle all other libraries

nvda-mesharma: I've added arg parser for all libraries

build.py
mc-nv: If the idea is to give users more control against the versions. 
May be it make sense to provide such a values as an variables?

nvda-mesharma: users can now pass values for any of these variables, it still makes sense to keep the default dictionary

build.py
oandreeva-nv: can we additionally mention the usage for them in the docs?

oandreeva-nv: can be in a follow-up pr

build.py
```diff
@@ -2553,6 +2541,54 @@ def enable_all():
...
+        "--release-version",
+        required=False,
+        default=DEFAULT_TRITON_VERSION_MAP["release_version"],
+        help="Provide any release version.",
+    )
+    parser.add_argument(
+        "--triton-container-version",
+        required=False,
+        default=DEFAULT_TRITON_VERSION_MAP["triton_container_version"],
+        help="Provide any released version of project.",
```
oandreeva-nv, line 2554: Is the `help` message make sense, I don't think we release any `dev` versions

nvda-mesharma: This text was previously present. I've only moved it to organize the args better. Happy to remove it.

nvda-mesharma: Can you suggest a better help message

build.py
```diff
@@ -69,16 +69,16 @@
...
+
+DEFAULT_TRITON_VERSION_MAP = {
+    "release_version": "2.50.0dev",
+    "triton_container_version": "24.09dev",
+    "upstream_container_version": "24.08",
+    "ort_version": "1.18.1",
+    "ort_openvino_version": "2024.0.0",
+    "standalone_openvino_version": "2024.0.0",
+    "dcgm_version": "3.2.6",
+    "vllm_version": "0.5.3.post",
```
oandreeva-nv, line 81:
```suggestion
    "vllm_version": "0.5.3.post1",
```

build.py
oandreeva-nv: I think this parameter indicates current development version of the container, not necessary a released version. Based on this, in my vision, the parameter should be something like `triton-development-version`, and help message would be `Provide the current development version of the container`. I'm not sure how relevant it would be to provide the past development versions and how much confusion this would introduce to users. Do we really need to expose this parameter? @nv-kmcgill53 and @mc-nv , what do you think?

oandreeva-nv:
Is it possible to start a new pipeline for this PR to check if it is goes well? + should we test newly added flags?

nvda-mesharma:
I've tested the changes via this pipeline 18898001

build.py
mc-nv: Any kind of logic could be added as an expansion for `build.py`.
My center of thoughts isn't changed, `build.py` requires simplification, any expansion need to be reasonable.

I agree that more moving parts we are adding the more complexity it will bring over time.

build.py
mc-nv: Our end users are developers and most like behavior for them is to clone and change the code, suppose it's overhead.

build.py
nvda-mesharma: afaik, this parameter can be used for dev and prod versions of triton container. A better help text for it would be "Provide any development or released version of project"

build.py
oandreeva-nv: Was an agreement reached on the overhead concern?

nvda-mesharma: @mc-nv Do you still have concern over the changes?

build.py
```diff
@@ -2632,6 +2622,54 @@ def enable_all():
         required=False,
         help="Override specified backend CMake argument in the build as <backend>:<name>=<value>. The argument is passed to CMake as -D<name>=<value>. This flag only impacts CMake arguments that are used by build.py. To unconditionally add a CMake argument to the backend build use --extra-backend-cmake-arg.",
     )
+    parser.add_argument(
+        "--release-version",
+        required=False,
+        default=DEFAULT_TRITON_VERSION_MAP["release_version"],
+        help="Provide any release version.",
```
oandreeva-nv, line 2629:
```suggestion
        help="This flag sets the release version for Triton Inference Server to be build. Default: the latest released version.",
```

oandreeva-nv: `any` word is extremely confusing, since it implies that a user can try any future unreleased versions. Please adjust `help` messages. I've provided my suggestions for `release-version`

oandreeva-nv:
I'm fine with these changes, just clarify help messages, as I mentioned here: https://github.com/triton-inference-server/server/pull/7500/#discussion_r1850904349

If @mc-nv and @nv-kmcgill53 have no objections and there are no disturbances with our current CI, i.e. not the vllm specific, we should be fine.

nv-kmcgill53 (approved):

nvda-mesharma:
Since there were some recent changes, I went ahead and ran a new pipeline to ensure that things were still working as expected 
.../pipelines/20628717

PR #7505, opened 2024-08-07
PA Migration: Update L0_client_build_variants

fpetrini15:
The objective of `L0_client_build_variants` is to build client libraries/binaries by enabling/disabling various optional features and verify that the build succeeds.

One such set of combinations is to enable/disable optional PA features and verify the build succeeds. This test had been passing (false positive) despite running with a client branch that purged PA from the client repo. Examining the logs closer, it was because CMake was simply ignoring PA-related parameters and repeatedly performing the same standard cc-client/python-client build:

```bash
Manually-specified variables were not used by the project:

    TRITON_ENABLE_PERF_ANALYZER
    TRITON_ENABLE_PERF_ANALYZER_C_API
    TRITON_ENABLE_PERF_ANALYZER_TFS
    TRITON_ENABLE_PERF_ANALYZER_TS
```
These changes modify the script to point to PA source code that recognizes these optional parameters, restoring the test's original purpose.

Ideally, these tests will eventually be moved out of Triton CI entirely and into PA CI.

### DIFF
--- a/qa/L0_client_build_variants/test.sh
+++ b/qa/L0_client_build_variants/test.sh
@@ -131,11 +131,12 @@ fi
 
 # TODO: These tests should be PA CI test cases
 # not Triton test cases
+rm -fr /workspace/build
+mkdir -p /workspace/build
 #
 # Build without C API in Perf Analyzer
 #
 (cd /workspace/build && \
-        rm -fr cc_clients && \
         cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
               -DTRITON_ENABLE_CC_HTTP=ON \
               -DTRITON_ENABLE_CC_GRPC=ON \
--- a/qa/L0_client_build_variants/test.sh
+++ b/qa/L0_client_build_variants/test.sh
@@ -129,8 +129,8 @@ else
     exit 1
 fi
 
-# TODO: These tests should be PA CI test cases
-# not Triton test cases
+# TODO: TPRD-342 These tests should be PA CI test
+# cases not Triton test cases
 rm -fr /workspace/build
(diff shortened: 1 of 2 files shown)


rmccorm4:
I was going to ask if we could detect and raise an error to catch these "unused variables" silent issues, but it seems like a common problem with no clear solution (from 2min of research): 
- https://discourse.cmake.org/t/cmake-warnings-as-error/3960/4
- https://stackoverflow.com/questions/54461491/cmake-stop-if-a-manually-specified-variable-is-not-used-by-the-project
- https://discourse.cmake.org/t/fail-cmake-if-there-are-unused-variables-defined/840

So I suppose there's no quick check we can do on our end, other than parsing the output with some custom logic.

qa/L0_client_build_variants/test.sh
```diff
@@ -141,59 +129,27 @@ else
...
-              /workspace/client && \
-        make -j16 cc-clients python-clients)
-if [ $? -eq 0 ]; then
-    echo -e "\n***\n*** No-Perf-Analyzer Passed\n***"
-else
-    echo -e "\n***\n*** No-Perf-Analyzer FAILED\n***"
-    exit 1
-fi
-
+# TODO: These tests should be PA CI test cases
```
rmccorm4, line 132: If you already have a ticket for the follow-up, let's add it here in case it gets swapped for other priorities `TODO (DLIS-XXXX): ...`

rmccorm4 (commented):
LGTM other than adding follow-up ticket comment

rmccorm4 (approved):
Great description and nice catch 🚀

PR #7519, opened 2024-08-12
build: RHEL 8 Compatibility

nv-kmcgill53:
#### What does the PR do?

This PR adds the following:
- Support for RHEL 8 compatible builds. 
  - Public users will need to setup their RHEL-like systems and then use the `--no-container-build` since Triton Inference Server doesn't publish a base container with which to use
  - NVIDIA users may build using a private base container to build
- Updated the `target_platform()` method in `build.py` to distinguish between debian and rhel-like systems. 
  - A follow on PR to change the target platform from "linux" to "debian" will be generated after this. 
- Update to L0_infer to be able to execute on either debian-like or rhel-like systems. 
- Update to `CMakeLists.txt` files to correctly point at the `lib64` directories on rhel-like systems.

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/core/pull/387
https://github.com/triton-inference-server/third_party/pull/59

build.py is the meat of this PR. The CMakeLists.txt files are the same changes which are needed in the Related PRs section.

#### Test plan:

No new e2e tests were added to this repository

- CI Pipeline ID:17365005

#### Background

This is a requested feature from users of Triton, check with Product team for specifics. 

- N/A

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,19 +126,12 @@ FetchContent_Declare(
 # Some libs are installed to ${TRITON_THIRD_PARTY_INSTALL_PREFIX}/{LIB}/lib64 instead
 # of ${TRITON_THIRD_PARTY_INSTALL_PREFIX}/{LIB}/lib on Centos
 set(LIB_DIR "lib")
-# /etc/os-release does not exist on Windows
-if(EXISTS "/etc/os-release")
-  file(STRINGS /etc/os-release DISTRO REGEX "^NAME=")
-  file(STRINGS /etc/os-release DISTRO_LIKE REGEX "^ID_LIKE=")
-  string(REGEX REPLACE "NAME=\"(.*)\"" "\\1" DISTRO "${DISTRO}")
-  string(REGEX REPLACE "ID_LIKE=\"(.*)\"" "\\1" DISTRO_LIKE "${DISTRO_LIKE}")
-  message(STATUS "Distro Name: ${DISTRO}")
-  message(STATUS "Distro Like: ${DISTRO_LIKE}")
-  if(DISTRO_LIKE MATCHES ".*rhel.*" OR DISTRO_LIKE MATCHES ".*centos.*")
+if(LINUX)
+  file(STRINGS "/etc/os-release" DISTRO_ID_LIKE REGEX "ID_LIKE")
+  if(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
     set (LIB_DIR "lib64")
-  endif()
-endif()
-
+  endif(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+endif(LINUX)
 set(TRITON_CORE_HEADERS_ONLY OFF)
 
 FetchContent_MakeAvailable(repo-third-party repo-core)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,13 +125,16 @@ FetchContent_Declare(
 
 # Some libs are installed to ${TRITON_THIRD_PARTY_INSTALL_PREFIX}/{LIB}/lib64 instead
 # of ${TRITON_THIRD_PARTY_INSTALL_PREFIX}/{LIB}/lib on Centos
-set (LIB_DIR "lib")
+set(LIB_DIR "lib")
 # /etc/os-release does not exist on Windows
 if(EXISTS "/etc/os-release")
   file(STRINGS /etc/os-release DISTRO REGEX "^NAME=")
+  file(STRINGS /etc/os-release DISTRO_LIKE REGEX "^ID_LIKE=")
   string(REGEX REPLACE "NAME=\"(.*)\"" "\\1" DISTRO "${DISTRO}")
+  string(REGEX REPLACE "ID_LIKE=\"(.*)\"" "\\1" DISTRO_LIKE "${DISTRO_LIKE}")
   message(STATUS "Distro Name: ${DISTRO}")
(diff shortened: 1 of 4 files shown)


build.py
```diff
@@ -1018,6 +1153,12 @@ def create_dockerfile_cibase(ddir, dockerfile_name, argmap):
         dfile.write(df)
 
 
+def create_dockerfile_rhel(
```
fpetrini15, line 1156: Needed anymore?

CMakeLists.txt
mc-nv, line 1: I don't see a reason in this change, if the idea is to include some directory during the compilation. 
Instead of all this syntax we can extend that path where we are looking for necessary libs, unless there are dependency conflict.

nv-kmcgill53: How would that look in CMake? I don't have a good head about the CMake API to understand what the efficient change would be. Also, I was having issues trying to get the [`function`](https://cmake.org/cmake/help/latest/command/function.html) API to inherit and work properly if you have some suggestions there?

mc-nv: Seems I miss read the configuration.

At glance out of configuration it only play role in installation path for "third_party" dependencies, there why raising a question do we need to introduce any path like `lib64` vs `lib`. 

`LIB_DIR` variable defines the installation path for some dependency libraries. 

Are there any compilation issues if this value remains `LIB_DIR=lib`?

mc-nv: I see it was caused by these PR’s:
https://github.com/triton-inference-server/server/pull/5836
https://github.com/triton-inference-server/core/pull/206
https://github.com/triton-inference-server/client/pull/325
https://github.com/triton-inference-server/third_party/pull/32

It’s weird… - third party contributor and no conversation around it.

@krishung5 , @tanmayv25  - could you please clarify why we have those changes above?

nv-kmcgill53: > Are there any compilation issues if this value remains LIB_DIR=lib?
For rhel-like distros, yes, there are compilation issues. the rhel-like distros define the `lib` directory as `lib64` which is why we need to detect the distribution we are compiling on and then set the correct library directory to make sure we can find the system libraries during linking.

mc-nv: If you mentioning that there is an issue with finding library, we need to be more specific in this case, it could be only impactful in one place. 
May be simple `include_directory` will be more efficient then mocking path across multiple repositories.

nv-kmcgill53: While the `/lib`, `/lib64` problem exists for system libraries it is also needed for the the other git repositories which get auto generated in the build. See line 181 and line 190 for example of usage. I don't think `include_directory` is going to be sufficient here.

mc-nv: Those examples are nothing more than copy-paste from our previous experience. 
This PR is seems to be the same - just expansion of existing solution.

mc-nv (dismissed):

CMakeLists.txt
```diff
@@ -125,13 +125,16 @@ FetchContent_Declare(
...
 if(EXISTS "/etc/os-release")
   file(STRINGS /etc/os-release DISTRO REGEX "^NAME=")
+  file(STRINGS /etc/os-release DISTRO_LIKE REGEX "^ID_LIKE=")
   string(REGEX REPLACE "NAME=\"(.*)\"" "\\1" DISTRO "${DISTRO}")
+  string(REGEX REPLACE "ID_LIKE=\"(.*)\"" "\\1" DISTRO_LIKE "${DISTRO_LIKE}")
   message(STATUS "Distro Name: ${DISTRO}")
-  if(DISTRO MATCHES "CentOS.*")
+  message(STATUS "Distro Like: ${DISTRO_LIKE}")
+  if(DISTRO_LIKE MATCHES ".*rhel.*" OR DISTRO_LIKE MATCHES ".*centos.*")
     set (LIB_DIR "lib64")
```
mc-nv, line 138: May be it make sense to address it as following:
```CMake
if(LINUX)
  file(STRINGS "/etc/os-release" DISTRO_ID_LIKE REGEX "ID_LIKE")
  if(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
    set (LIB_DIR "lib64")
  endif(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
endif(LINUX)
```

build.py
nv-kmcgill53: I guess not. Removed.

CMakeLists.txt
nv-kmcgill53: Tested locally to confirm and fixed across all repositories. Thanks for the suggestion!

mc-nv (approved):

build.py
```diff
@@ -1161,7 +1297,30 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
...
+        re2-devel \\
+        openssl-devel \\
+        libtool \\
+        libcurl-devel \\
+        libb64-devel \\
+        gperftools-devel \\
+        patchelf \\
+        wget \\
+        numactl-devel \\
+        wget
```
fpetrini15, line 1320:
```suggestion
        numactl-devel
```
Duplicate. Won't require new CI.

fpetrini15 (approved):
LGTM, pending passing CI. Great work Kyle!

PR #7524, opened 2024-08-13
build: RHEL8 PyTorch Backend

fpetrini15:
**Goal:** Support the PyTorch Backend on RHEL8 systems. Importantly, these changes will disable optional PyTorch optional features such as `TRITON_PYTORCH_ENABLE_TORCHTRT` and `TRITON_ENABLE_NVTX`, which are not currently supported for the RHEL8 build.

PyTorch Backend: https://github.com/triton-inference-server/pytorch_backend/pull/137


build.py
```diff
@@ -914,6 +917,16 @@ def create_dockerfile_buildbase_rhel(ddir, dockerfile_name, argmap):
...
+"""
+
+    df += """
+# Point to python 3.10 which is what triton expects the default to be
+ENV PYVER 3.10
+ENV PYTHONPATH /opt/python/v
+RUN ln -sf /opt/python/cp${PYVER/./}* ${PYTHONPATH}
+ENV PYBIN ${PYTHONPATH}/bin
+ENV PYTHON_BIN_PATH ${PYBIN}/python${PYVER}
+ENV PATH ${PYBIN}:${PATH}
```
mc-nv, line 929: Why do we need custom location for python installation?

fpetrini15: The manylinux container we are using as a base has several versions of python installed. By default, it will use Python3.7 which is incompatible with some of the packages we need to install. So here we default the version to 3.10.

mc-nv: Why don't we use proper version of Python aligned with all DLFW products instead and expanding existing image?

fpetrini15: Python 3.10 is the properly aligned version for Triton and PyTorch as far as I can tell: https://docs.nvidia.com/deeplearning/frameworks/support-matrix/index.html. Realistically, the best place to fix this would be the origin--the cuda base container. I'm not sure what the rationale is for having it default to 3.7.

fpetrini15: A fix at that level would prevent us and the Pytorch team from having to do this in the first place.

PR #7527, opened 2024-08-14
test: Load new model version should not reload loaded existing model version(s)

kthui:
#### What does the PR do?

Add the following tests on model version reload:
1. If a version is loaded and unmodified, then it should not be reloaded on the next load request.
2. If a version is loaded but modified, then it should be reloaded on the next load request.
3. If a version is not loaded and its model file is in the model directory, then it should be loaded on the next load request.
4. If a version is not loaded nor its model file is in the model directory, then it should be loaded on the next load request.
5. If a generic file is modified on the model directory, i.e. `model_directory/common_dependency.py`, then all loaded version(s) should be reloaded on the next load request.

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [x] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [x] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:

https://github.com/triton-inference-server/core/pull/388

#### Where should the reviewer start?

Start with the core PR.

#### Test plan:

The PR adds the test, see the Test plan on the core PR.

- CI Pipeline ID: 17511909  17650341

#### Caveats:

N/A

#### Background

N/A

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
N/A

PR #7553, opened 2024-08-20
test: Load new model version should not reload loaded existing model …

kthui:


kthui:
Pick https://github.com/triton-inference-server/server/pull/7527 into r24.08

nv-kmcgill53 (approved):

PR #7561, opened 2024-08-22
feat: OpenAI Compatible Frontend

rmccorm4:
### Description

Adds an OpenAI Compatible Frontend for Triton Inference Server as a FastAPI application using the `tritonserver` in-process python bindings for the following endpoints:
- `/v1/models`
- `/v1/completions`
- `/v1/chat/completions`

Additionally there are some other observability routes exposed for convenience:
- `/metrics` - Prometheus-compatible metrics from Triton Core
- `/health/ready` - General health check for inference readiness, similar to NIM schema.

This is a refactor and extension of the original example by @nnshah1 here: https://github.com/triton-inference-server/tutorials/pull/104 to include more thorough testing.

#### Refactors/Changes

1. The file structure was refactored to make each logical component smaller, more discoverable, and more digestible. It was loosely based on the following resources:
    - https://fastapi.tiangolo.com/tutorial/bigger-applications/#an-example-file-structure
    - https://github.com/zhanymkanov/fastapi-best-practices?tab=readme-ov-file#project-structure
 
2. The global variables were converted to use the FastAPI `app` object for storing state instead, such as `app.server` to access the in-process `tritonserver` across different routes handling requests. 

3. Encapsulated info related to model objects in a `ModelMetadata` dataclass.

4. Made the `--tokenizer` an explicit setting provided by user at startup, rather than something we try to infer based on the Triton model name used in the model repository. This is more resilient to BYO/custom models, and models not produced by the Triton CLI. 

5. Added an **optional** explicit `--backend` (request conversion format) override to better deal with edge cases like LLMs defined in custom backends, or ambiguous backends like "python" for ensemble/BLS models where the request format isn't immediately clear from the backend choice. If the explicit override is not used, the frontend will currently pick `vllm` request format if the targeted model's `backend=vllm`, otherwise it will pick `backend=tensorrtllm` for any other backend value.

6. Extracted most of the Triton in-process Python API logic out of the FastAPI route definitions. The pre-initialized in-process tritonserver is attached to the FastAPI app during initialization in `main.py`. This will make it simpler to mock and unit test specific parts in the frontend, as well as teardown/startup the frontend during testing without stopping/starting the tritonserver each time.
 
7. Updated various parts of the generated `schemas/openai.py` that seemed to be generated for pydantic v1 (currently using v2) to account for all of the deprecation warnings that are currently deprecated but will be fully removed in coming versions of pydantic: https://docs.pydantic.dev/latest/migration/

#### Testing Methodology

1. Original set of tests were added using FastAPI's [TestClient](https://fastapi.tiangolo.com/tutorial/testing/) for simplicity in getting started. I later found out that this `TestClient` object, despite interacting with the application and logic as normal, does not actually expose the network host/ports for interaction from other clients. It would probably be good to move away from `TestClient` and use the `OpenAIServer` utility, described below, for all testing later on.
2. OpenAI client tests were then added to show how this implementation can act as a drop-in replacement for current users of the `openai` python client library, which required a `OpenAIServer` utility class to run the application differently then the `TestClient` flow expects, similar to vLLM's utility [here](https://github.com/vllm-project/vllm/blob/7937009a7e82c3c4c9c7f48d11142bee5aac4a30/tests/utils.py#L57).
3. Streaming tests were done through the OpenAI client, as the documentation around `streaming` with `TestClient` were sparse.
4. Overall the tests were meant to add broad support for various types of clients (raw http requests, openai client, genai-perf) and broad testing for the supported parameters (test that changes in the openai-facing values produce different behavior in the backends). Deep dives into each and every parameter and exactly how they would behave for each value (ex: `top_p`, `top_k`) was not done, and I would expect more to be tested by each backend implementation.

### Open Questions

- [x] Q: Should `/v1/models` list all models in the model repository, even if only some of those models are actually compatible with the completions / chat endpoints? It may be difficult to automatically infer which models are "compatible" or not and account for all scenarios, so it may nee to be explicit by the user if we want to limit which models are returned.
    - A: Let's expose a `SERVED_MODEL_NAME` for a cosmetic mapping from something like `tensorrt_llm_bls` -> `meta-llama/Meta-Llama-3.1-8b-Instruct` to let the server/app starter specify how clients should interact with it. They may want to hide the details of ensemble/BLS/etc. to clients behind a more convenient `--served-model-name`.
- [x] Q: What to do with the `docker/Dockerfile*` and general expected user-facing workflow. I added these as examples for myself to use during testing/development and included how they'd be used in the README, but ultimately we will probably be publishing this code within the respective containers and these DIY containers are likely unnecessary.\
    - A: These will be likely removed as we get closer to having something published in a container.

### Open Items for this PR
- [x] Benchmarking and performance improvements, especially under high concurrency and load
- [x] KServe frontend bindings integration
- [x] Fix issues with parameters like `stop` that are passed as a `List` when using vLLM backend.

### Open Items for follow-up PRs
- [ ] README and Documentation improvements (DLIS-7179)
- [ ] Support SERVED_MODEL_NAME mapping for "convenience name" -> "triton model name" (DLIS-7171)
- [ ] Add some tests that use `genai-perf` to maintain compatibility and catch regressions
- [ ] Unsupported OpenAI schema parameters like `logit_bias`, `logprobs`, `n > 1`, `best_of`. These are mostly just unexplored, and haven't yet been scoped out to see the effort involved in supporting them. (DLIS-7183, DLIS-7184)
- [ ] Function Calling / Tools support (DLIS-7168)
- [ ] Unsupported OpenAI schema response fields like `usage` (DLIS-7185)
- [ ] Testing for LoRA/multi-LoRA (DLIS-7186)
- [ ] Migrating `TestClient` tests to use the `OpenAIServer` utility for all testing instead

### Notes

- I left in all the commits over time if you want to watch how the code evolved from the original, and see some of the edge cases that were caught by testing such as `text/event-stream` headers missing for streaming, `temperature` being ignored by TRT-LLM BLS, `content=None` for streaming messages causing genai-perf errors compared to `content=""`, certain sampling parameters being converted to incorrect types internally, etc.


nnshah1:
One general comment on the `/v1/models` open - I think we should limit to the models that support the openai endpoints instead of listing all models. We can make it explicit via command line option like backend and tokenizer. 

Also agree for first pass lets be as explicit as possible - we can simplify after if needed / there is an opportunity.

python/openai/docker/Dockerfile.tensorrtllm
```diff
@@ -0,0 +1,33 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver:24.07-trtllm-python-py3
+FROM ${BASE_IMAGE}
+
+RUN pip install /opt/tritonserver/python/*.whl
+
+COPY requirements.txt /tmp
```
nnshah1, line 32: maybe irrelevant as the dockerfile may be removed - but I found a nicer way to do this that avoids adding the requirements file into the container:
`RUN --mount=type=bind,source=./deps/requirements.txt,target=/tmp/requirements.txt \
    pip install --timeout=2000 --requirement /tmp/requirements.txt`

python/openai/examples/chat.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# or "tensorrt_llm_bls" for TRT-LLM
```
nnshah1, line 2: I think we should in the final version provide an easy way to alias the model name to be the underlying model name and not trt-llm name

python/openai/openai/src/routers/chat_completions.py
```diff
@@ -0,0 +1,200 @@
...
+    CreateChatCompletionResponse,
+    CreateChatCompletionStreamResponse,
+    ObjectType,
+)
+from src.utils.triton import get_output, validate_triton_responses
+
+router = APIRouter()
+
+
+def get_first_response_role(conversation, add_generation_prompt, default_role):
```
nnshah1, line 48: eventually: type hints would be great

python/openai/openai/src/routers/chat_completions.py
```diff
@@ -0,0 +1,200 @@
...
+) -> CreateChatCompletionResponse | StreamingResponse:
+    """
+    Creates a model response for the given chat conversation.
+    """
+
+    model_metadatas = raw_request.app.models
+    if not model_metadatas:
+        raise HTTPException(status_code=400, detail="No known models")
+
+    metadata = model_metadatas.get(request.model)
```
nnshah1, line 116: nit: I think `model_metadata` is plural enough

python/openai/openai/src/routers/chat_completions.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+
+    if not metadata.tokenizer:
+        raise HTTPException(status_code=400, detail="Unknown tokenizer")
+
+    if not metadata.backend:
+        raise HTTPException(status_code=400, detail="Unknown backend")
+
+    triton_model = raw_request.app.server.model(request.model)
+    if request.model != triton_model.name:
```
nnshah1, line 132: curious: I'm not sure how this would be possible - the model api doesn't do any checks and would return exactly what you give it ... was there an issue that you saw?

rmccorm4: This looks to be redundant, probably from some find/replace, will remove :+1:

python/openai/openai/src/utils/tokenizer.py
```diff
@@ -0,0 +1,77 @@
+from typing import Optional, Union
```
nnshah1, line 1: will need to find the right copyright

python/openai/openai/src/utils/triton.py
```diff
@@ -0,0 +1,211 @@
...
+from dataclasses import dataclass
+
+import numpy as np
+import tritonserver
+from fastapi import HTTPException
+from src.schemas.openai import CreateChatCompletionRequest, CreateCompletionRequest
+from src.utils.tokenizer import get_tokenizer
+
+
+# TODO: Stricter pydantic validation would be better in future
```
nnshah1, line 39: pydantic and dataclasses are distinct AFAIK - and dataclasses are lighter weight.

If there isn't a need for serialization / deserializaiton and validation from that perspective - dataclasses may be preferable

python/openai/openai/src/utils/triton.py
```diff
@@ -0,0 +1,211 @@
...
+    # Name of backend used by Triton
+    backend: str
+    # Triton model object handle
+    model: tritonserver.Model
+    # Tokenizers used for chat templates
+    tokenizer: typing.Optional[typing.Any]
+    # Time that model was loaded by Triton
+    create_time: int
+    # Conversion format between OpenAI and Triton requests
+    request_convert_fn: typing.Optional[typing.Any]
```
nnshah1, line 53: nit: maybe `request_converter` - don't see a lot of `fn` suffix -

python/openai/openai/src/utils/triton.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+# TODO: Expose explicit flag to catch edge cases
+def determine_request_format(backend):
```
nnshah1, line 57: nit: if function is internal to module - prefix with `_`

nnshah1 (dismissed):
LGTM - will want to add doc strings for any public interfaces from modules and prefix internal with `_`. I'm not sure about the `src` directory - but need to look through the references cited.

nice work!

python/openai/docker/Dockerfile.tensorrtllm
mc-nv, line 1: Why we need two docker files? Can we have multistage image and manage it?

python/openai/openai/main.py
nnshah1, line 1: my understanding would be that the src folder contains even main.py ...

we have `python/openai/openai/main.py`

maybe instead:
`python/openai_tritonserver/src/main.py`

I think we need to expand `openai` into either `openai_frontend` or `openai_tritonserver` to avoid confusion with `openai`

python/openai/docker/Dockerfile.tensorrtllm
rmccorm4: Naively installing vllm inside of TRT-LLM causes a new and incompatible pytorch to get installed, which breaks TRT-LLM. Installing TRT-LLM inside of vLLM container would probably take some time to figure out.

These dockerfiles are here now for simplicity in dev/testing, but will probably be removed when the source code is ready to be shipped in the release containers, as they will become unnecessary

mc-nv: I was considering following
```Dockerfile
ARG TENSORRT_LLM_IMAGE=nvcr.io/nvidia/tritonserver:24.07-trtllm-python-py3
FROM ${TENSORRT_LLM_IMAGE} AS tensorrt_llm
RUN pip install /opt/tritonserver/python/*.whl
RUN pip install fastapi==0.111.1 openai==1.40.6 pytest==8.1.1 pytest-asyncio==0.23.8

ARG VLLM_IMAGE=nvcr.io/nvidia/tritonserver:24.07-vllm-python-py3
FROM ${VLLM_IMAGE} AS vllm
RUN pip install /opt/tritonserver/python/*.whl
RUN pip install fastapi==0.111.1 openai==1.40.6 pytest==8.1.1 pytest-asyncio==0.23.8
RUN pip install transformers==4.43.1 vllm==0.5.3.post1
```
In this case we can refer customers to commands like: 
```
docker build --target vllm -f Dockerfile . -t vllm-image
docker build --target tensorrt_llm -f Dockerfile . -t tensorrt_llm-image
```

python/openai/openai/src/api_server.py
```diff
@@ -0,0 +1,103 @@
...
+    # Start the tritonserver on FastAPI app startup
+    print("Starting FastAPI app lifespan...")
+    server, model_metadatas = init_tritonserver()
+
+    # NOTE: These are meant for read-only access by routes handling requests
+    # with a single process, and should generally not be modified for the
+    # lifetime of the application. If multiple uvicorn workers are instantiated,
+    # then multiple triton servers would be started, one per worker process.
+    app.server = server
+    app.models = {metadata.name: metadata for metadata in model_metadatas}
```
GuanLuo, line 66: Can this be dynamic in nature? It can be a functor that will trigger Triton API and re-generate the metadata, or reference to something that server owns and the server will update the member when load/unload is performed

python/openai/openai/src/utils/triton.py
```diff
@@ -0,0 +1,211 @@
...
+    # Request conversion from OpenAI format to backend-specific format
+    if backend == "vllm":
+        request_convert_fn = create_vllm_inference_request
+    # Python included to support TRT-LLM BLS model and TRT-LLM python runtime
+    elif backend in ["tensorrtllm", "python"]:
+        request_convert_fn = create_trtllm_inference_request
```
GuanLuo, line 63: Should `create_trtllm_inference_request` be default convert func? Seems simplier if we have a default and let Triton complains if it doesn't align with the model config.

GuanLuo: Also thining about TRT-LLM ensemble

python/openai/openai/src/schemas/openai.py
```diff
@@ -0,0 +1,867 @@
+# generated by fastapi-codegen:
+#   filename:  api-spec/openai_trimmed.yml
```
GuanLuo, line 2: Include the yaml instead and only code-gen at build time?

python/openai/openai/tests/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.fixture
+def prompt():
+    return TEST_PROMPT
+
+
+@pytest.fixture
+def messages():
+    return TEST_MESSAGES
```
GuanLuo, line 103: Can you define the values here instead of define globally and return?

python/openai/openai/tests/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+###
+
+
+# NOTE: OpenAI client requires actual server running, and won't work
+# with the FastAPI TestClient. Run the server at module scope to run
+# only once for all the tests below.
+@pytest.fixture(scope="module")
+def server():
+    model_repository = Path(__file__).parent / f"{TEST_BACKEND}_models"
+    args = ["--model-repository", model_repository, "--tokenizer", TEST_TOKENIZER]
```
GuanLuo, line 68: You can mock Triton core and pass here if the init function allows passing an Triton core object, then you don't need real Triton running to test frontend behavior.

python/openai/openai/src/__init__.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
nnshah1: agree we should come up with a general scheme for layout - can be later - but for reference in other project have been following:

<library/package><language><package-name>

So maybe something like

```openai_frontend/python/openai_frontend``
``openai_frontend/rust/...```
``openai_frontend/python/tests```

Still has the double "openai_frontend" - but one is the logical feature / library and one is the language specific directory structure. 
 for c++ it might be

``openai_frontend/c++/src/```

As the library naming isn't direcctory dependent .

File: python/openai/openai/src/schemas/openai.py (hunk 12 above)
Comment (nnshah1) [reply python/openai/openai/src/schemas/openai.py:2]: I believe there were some manual edits needed - and not sure how easy it would be to seperate those into a seperate file . We should have clear instructions on the steps to generate and modify though - and reference the specific version of fastapi-codegen we used.

File: python/openai/README.md [hunk 16]
```diff
@@ -0,0 +1,148 @@
...
+```
+
+4. Send a `/v1/completions` request:
+  - Note the use of `jq` is optional, but provides a nicely formatted output for JSON responses.
+```bash
+MODEL="llama-3.1-8b-instruct"
+curl -s http://localhost:8000/v1/completions -H 'Content-Type: application/json' -d '{
+  "model": "'${MODEL}'",
+  "prompt": "Machine learning is"
+}' | jq
```
oandreeva-nv, line 53: nit: can be done later, I think expected answer can help users

python/openai/docker/requirements_vllm.txt
```diff
@@ -0,0 +1,3 @@
+transformers==4.43.1
+# Llama3.1 vllm requirements
+vllm==0.5.3.post1
```
oandreeva-nv, line 3: should it be exactly this version or `>=` ?

python/openai/openai/src/utils/triton.py
oandreeva-nv: nit: I find it a bit confusing that the function is named `determine_request_format`, but we actually return a convert function

python/openai/openai/src/utils/triton.py
```diff
@@ -0,0 +1,211 @@
...
+    return request_convert_fn
+
+
+def load_models(server):
+    model_metadatas = []
+    backends = []
+
+    # TODO: Consider support for custom tokenizers
+    tokenizer = None
+    tokenizer_model = os.environ.get("TOKENIZER")
```
oandreeva-nv, line 76: nit: check if the environment has this var

python/openai/docker/requirements_vllm.txt
rmccorm4: `==` for consistency during dev/testing, will ultimately use the deps shipped in the container it goes in

python/openai/openai/src/utils/triton.py
rmccorm4: `get()` returns None if not found, and tokenizer is optional for other routes like `/completions` that don't need chat template

python/openai/openai/src/utils/triton.py
rmccorm4: Yep they are distinct, this comment was mentioning the move to pydantic could be beneficial, and remove the need for some of the in-line code checks for certain values or ranges on the request parameters such as `n > 1` - but more of a future note to self. Agreed pydantic validation can add overhead if unnecessary

GuanLuo:
@KrishnanPrash FYI, use this `python/...` structure as the reference for KServe. This is what I meant when I commented on the file structure.

python/openai/openai/src/__init__.py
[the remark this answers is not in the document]
GuanLuo: I would rather `<language>/<library/package>/<language-specific-package-name>`, which is the common practice for repos that have support for different languages, and much easier to build / package library for corresponding language.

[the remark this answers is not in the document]
nnshah1: @GuanLuo - do you have a reference we can use? - we can add to help guide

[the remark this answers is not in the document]
GuanLuo: You don't need to follow the same file structure within language barrier, but the language barrier should make sure the files within will be self-contained.
https://github.com/NVIDIA/cutlass
https://github.com/rapidsai/ucxx/tree/branch-0.39
https://github.com/tensorflow/tensorflow/tree/master/tensorflow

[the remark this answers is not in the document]
rmccorm4: Updated to the following for now based on the thread:
- `python/openai/tests`
- `python/openai/openai_frontend`
- Removed `src`

ref: https://github.com/triton-inference-server/server/pull/7561/commits/886ee7d212a8f6774d6dfefdeffeb328f32fad24

Happy to rename folders or re-structure later after addressing bigger design feedback

python/openai/openai/main.py
rmccorm4: Resolving in favor of this thread which overlaps: https://github.com/triton-inference-server/server/pull/7561#discussion_r1729588180

python/openai/openai/src/utils/triton.py
rmccorm4: Simplified to use TRT-LLM format for anything other than vLLM for now.

python/openai/openai/src/routers/chat_completions.py
rmccorm4: added some more type hints in a few spots - could still be improved further, but leaving it as-is for now

python/openai/openai/src/utils/triton.py
rmccorm4: Renamed and type hinted as `typing.Callable` -- didn't go as far as the expected callable input/output types yet

python/openai/openai/src/utils/triton.py
rmccorm4: added prefix and renamed

python/openai/openai/src/routers/chat_completions.py
```diff
@@ -0,0 +1,200 @@
...
+    if not metadata.request_convert_fn:
+        raise HTTPException(
+            status_code=400, detail=f"Unknown request format for model: {request.model}"
+        )
+
+    if not metadata.tokenizer:
+        raise HTTPException(status_code=400, detail="Unknown tokenizer")
+
+    if not metadata.backend:
+        raise HTTPException(status_code=400, detail="Unknown backend")
```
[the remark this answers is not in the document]
rmccorm4: added some helpers - will probably still restructure more over time

python/openai/openai/src/routers/chat_completions.py
```diff
@@ -0,0 +1,200 @@
...
+    prompt = metadata.tokenizer.apply_chat_template(
+        conversation=conversation,
+        tokenize=False,
+        add_generation_prompt=add_generation_prompt,
+    )
+
+    request_id = f"cmpl-{uuid.uuid1()}"
+    created = int(time.time())
+
+    responses = triton_model.infer(
```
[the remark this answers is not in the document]
rmccorm4: done

python/openai/openai/src/routers/chat_completions.py
```diff
@@ -0,0 +1,200 @@
...
+        metadata.request_convert_fn(triton_model, prompt, request)
```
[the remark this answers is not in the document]
rmccorm4: unmangled it a bit, could still use more unmangling after the dust settles on a few more things.

python/openai/openai/src/routers/completions.py
```diff
@@ -0,0 +1,151 @@
...
+) -> CreateCompletionResponse | StreamingResponse:
+    """
+    Creates a completion for the provided prompt and parameters.
+    """
+
+    if not request.model:
+        raise Exception("Request must provide a valid 'model'")
+
+    print(f"[DEBUG] Available model metadata: {raw_request.app.models.keys()=}")
+    print(f"[DEBUG] Fetching model metadata for {request.model=}")
```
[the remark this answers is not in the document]
rmccorm4: done

python/openai/openai/src/schemas/openai.py
rmccorm4: The codegen tool produced a lot of older pydantic v1 concepts that are currently deprecated in pydantic v2, and will be removed in pydantic v3 - so I made manual edits there. 

Ideally we'd spend more time to polish and document the codegen path in a way that won't need manual edits - but I'm considering that lower priority for now. I think it would be a good task to take on when adding another endpoint such as embeddings.

nnshah1 (dismissed):
LGTM

python/openai/openai/src/api_server.py
```diff
@@ -0,0 +1,103 @@
...
+        CORSMiddleware,
+        allow_origins=origins,
+        allow_credentials=True,
+        allow_methods=["*"],
+        allow_headers=["*"],
+    )
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI):
```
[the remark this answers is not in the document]
rmccorm4: Refactored with some initial interfaces and passing a pre-prepared engine to the frontend: https://github.com/triton-inference-server/server/pull/7570

python/openai/docker/Dockerfile.tensorrtllm
```diff
@@ -0,0 +1,33 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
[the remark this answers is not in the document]
rmccorm4: Until we have finalized other container delivery paths, I'm going to leave these in for easier collaboration and testing.

python/openai/examples/chat.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rmccorm4: Until we have finalized packaging this up somewhere, I'm going to leave these examples in for easier collaboration and testing.

python/openai/examples/chat.sh
rmccorm4: Will have a separate PR to refactor which models and names get exposed to user through OpenAI frontend

python/openai/docker/Dockerfile.tensorrtllm
rmccorm4: Ultimately these dockerfiles will likely be removed before this is released, so I'm going to defer this for now. But thanks for teaching me something new, didn't know about `--target` :nerd_face:

python/openai/openai/tests/conftest.py
rmccorm4: Will follow-up with mock tests once you're happy with this general refactor: https://github.com/triton-inference-server/server/pull/7570

python/openai/README.md
rmccorm4: Will improve with expected outputs in a follow-up PR focused on docs DLIS-7179

qa/L0_openai/test.sh
```diff
@@ -0,0 +1,75 @@
...
+    rm -rf ${MODEL_REPO}
+
+    # FIXME: This will require an upgrade each release to match the TRT-LLM version
+    # Use Triton CLI to prepare model repository for testing
+    pip install git+https://github.com/triton-inference-server/triton_cli.git@0.0.10
+    # NOTE: Could use ENGINE_DEST_PATH set to NFS mount for pre-built engines in future
+    triton import \
+        --model ${MODEL}  \
+        --backend tensorrtllm \
+        --model-repository "${MODEL_REPO}"
```
GuanLuo, line 33: Is this going to rebuild the model every time? Do we have a model store for TRT-LLM model? Can be passed to this test script as cached location (with TRT-LLM version generated from) and only generate if TRT-LLM version is different from the cached model

rmccorm4: I tried to use NFS mount to cache and re-use engines across test jobs, but in appeared that NFS I/O was slower than just building the model each time for now due to some strange slowness going on with infra currently, see example change I had prior: https://github.com/triton-inference-server/server/commit/81ef479164f6d2f5e2fb1c131312fa991f499b95

Ideally this would not build the model each time in the long term.

faileon:
Would it be possible to include /embeddings endpoint too? It would be the last missing piece for our company to try out Triton for our on-prem RAG/Agentic solutions.

mc-nv:
@rmccorm4 - is this change includes binary extension, if not why we adding it to the this repository?

robertgshaw2-neuralmagic:
This is great!

python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -36,14 +36,34 @@ def _create_vllm_inference_request(
     model, prompt, request: CreateChatCompletionRequest | CreateCompletionRequest
 ):
     inputs = {}
-    excludes = {"model", "stream", "messages", "prompt", "echo"}
+    # Exclude non-sampling parameters so they aren't passed to vLLM
```
rmccorm4, line 39: NOTE: May make more sense to explicitly "include" support sampling parameters, but will be of a similar length or longer. Both approaches likely require periodic updates to either include a new sampling field, or exclude a non-sampling field.

GuanLuo (approved):

chorus-over-flanger:
> Would it be possible to include /embeddings endpoint too? It would be the last missing piece for our company to try out Triton for our on-prem RAG/Agentic solutions.

strong upvote for this feature to be enabled, would be great and, indeed - complete the whole functionality

python/openai/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+## TensorRT-LLM
+
+0. Prepare your model repository for serving a TensorRT-LLM model:
+   https://github.com/triton-inference-server/tensorrtllm_backend?tab=readme-ov-file#quick-start
+
+1. Launch the container:
+  - Mounts the `~/.huggingface/cache` for re-use of downloaded models across runs, containers, etc.
+  - Sets the [`HF_TOKEN`](https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hftoken) environment variable to
+    access gated models, make sure this is set in your local environment if needed.
```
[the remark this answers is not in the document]
rmccorm4: Primarily for the tokenizer used in frontend for chat templates, yea. Also useful if they intend to download weights and build engines in the container too.

.gitignore
```diff
@@ -5,4 +5,13 @@
...
+*.xml
 test_results.txt
+artifacts
+cprofile
+*.prof
+
+# Test exclusions
+qa/L0_openai/openai
+tensorrtllm_models
+custom_tokenizer
```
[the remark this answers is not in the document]
rmccorm4: These are primarily for local dev/debug loop convenience, and to avoid accidentally committing them if keeping them around during dev. 

For example:
- `custom_tokenizer`: not re-downloading a pre-downloaded tokenizer (the test is focused on testing tokenizer from local file path, not on downloading a tokenizer, but downloads if not present) 
- `tensorrtllm_models`: not re-generating trtllm models each time you want to re-run individual test cases locally after the first run

python/openai/openai_frontend/engine/__init__.py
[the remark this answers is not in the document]
rmccorm4: Nice catch - wasn't synced with latest main

python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -0,0 +1,175 @@
...
+def _validate_triton_responses_non_streaming(
+    responses: List[tritonserver._api._response.InferenceResponse],
+):
+    num_responses = len(responses)
+    if num_responses == 1 and responses[0].final != True:
+        raise Exception("Unexpected internal error with incorrect response flags")
+    if num_responses == 2 and responses[-1].final != True:
+        raise Exception("Unexpected internal error with incorrect response flags")
+    if num_responses > 2:
+        raise Exception(f"Unexpected number of responses: {num_responses}, expected 1.")
```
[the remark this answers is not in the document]
rmccorm4: Yeah agreed the separation of the utils/helpers definitely got a little blurry/messy - a restructure like this makes sense.

I would like to further restructure later a few things as well.

I think eventually having imports in a user-facing app looking something like this would look nice:
```python
from tritonserver.core import Server
from tritonserver.frontends.kserve import KServeGrpc
from tritonserver.frontends.openai import FastApiFrontend
from tritonserver.frontends.openai.engine import TritonLLMEngine
```

But will consider restructuring in general as a follow-up PR focused just on structure after some discussion.

kthui (approved):

python/openai/requirements.txt
```diff
@@ -0,0 +1,7 @@
+# FastAPI Application
+fastapi==0.111.1
+openai==1.40.6
+
+# Testing
+pytest==8.1.1
+pytest-asyncio==0.23.8
```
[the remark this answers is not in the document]
rmccorm4: Sure, added separate `requirements-test.txt`

.pre-commit-config.yaml
```diff
@@ -73,12 +73,13 @@ repos:
...
-- repo: local
-  hooks:
-  - id: add-license
-    name: Add License
-    entry: python tools/add_copyright.py
-    language: python
-    stages: [pre-commit]
-    verbose: true
-    require_serial: true
+# FIXME: Only run on changed files when triggered by GitHub Actions
```
rmccorm4, line 76: FYI @GuanLuo - I'll look into fixing this on GitHub Actions separately

kthui (approved):

GuanLuo (approved):

rmccorm4:
Hi @chorus-over-flanger @faileon, thanks for expressing interest in the `/embeddings` route! It's on our radar as another feature to add to the OpenAI frontend support (chat, completions, and models) added in this PR. 

Since the route itself doesn't have too many parameters defined in the [spec](https://platform.openai.com/docs/api-reference/embeddings/create), it may be relatively straightforward to add.

If you have any particular models you'd like to see working as an example for `/embeddings`, do let us know.

If you have any feedback or interest in contributing to the project, let us know as well.

Thanks!

copasseron:
Hi @rmccorm4 ,

thanks a lot on this PR. Huge work very appreciated here. 

Do you know if you're planning to add a support for  Function Calling / Tools support (DLIS-7168) soon ?

DimadonDL:
Hello,

is it already clear when the beta phase will be over? 

Thanks for this very important PR!

chorus-over-flanger:
Hello @rmccorm4 

I've been testing the OpenAI frontend of TIS with vLLM, and I'm impressed with the overall results - excellent work! I'd like to share some findings and observations,

- #7845;

- **Performance:** 
When load-testing, I noticed the OpenAI frontend (TIS/vLLM) shows lower TPS compared to the Generate extension of TIS with vLLM backend. This testing was primarily conducted with quantized 7/8B models. I can provide more detailed test results if needed;

- **Regarding the `embeddings` endpoint:**
In my specific use case, I found that large model embeddings don't perform as well as current SOTA embedding models with million-scale parameters, which led me to switch to a different backend. I noticed that vLLM recently [added](https://docs.vllm.ai/en/latest/models/supported_models.html#text-embedding) support for various embedding models (both billion and million parameter sizes), whereas previously it only supported larger models;

I would also request a `function calling/tools` support in OpenAI frontend, as well as [`structured_output`](https://platform.openai.com/docs/guides/structured-outputs),

huge huge work!

ishandhanani:
You cooked on this one @rmccorm4

DimadonDL:
Hi, is there a timeline for this PR? This would be such a great feature 🤩

oandreeva-nv:
hi @DimadonDL, feel free to try this functionality in our latest `25.02` vllm or trtllm containers. How-tos can be found here: https://github.com/triton-inference-server/server/tree/main/python/openai

DimadonDL:
Hi @oandreeva-nv,

Thanks. I have tested this feature. Did you know when its leaves the beta status?

PR #7567, opened 2024-08-26
fix: Add reference count tracking for shared memory regions

pskiran1:
#### What does the PR do?
This pull request is intended to address the following issue.
Currently, we do not track whether the shared memory (shm) region is being used by any inference request, and we allow unregistering of the shm region at any time. When performing inference, if the user unregisters the shm region, the server attempts to read or write data in the shm region, resulting in a segmentation fault and causing the server to crash.

To address this issue, we have made the following changes:
- Added a new attribute `ref_count_` counter to `ShareMemory`, which represents the count of inference requests currently using it. We increment the counter when parsing the request to the `InferenceRequest` object, and maintain the unique shm region names in the `InferenceRequest` object referencing. Upon response completion or error return, we decrement this counter.
- We now allow unregistering any system/CUDA shm region only if the `ref_count_` is 0 at that moment. This ensures that only unused shm regions can be unregistered. Users can also check the number of inference requests using the shm region `ref_count` by querying the shm region status.
If a user tries to unregister a shm region that is in use, we return the following error:
Single shm: "Cannot unregister shared memory region 'input0_data', it is currently in use by 1 requests."
All shm: "Failed to unregister the following system shared memory regions: input0_data, output0_data, "
- During the server shutdown, we will disregard the `ref_count_` and ensure that all shm regions are unregistered.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 18081701

- closes GitHub issue: #xxx


GuanLuo (commented):
I think that the change should be addressed in a different approach so that the shared memory detail is not leaked into Triton core, some background on the current state of the code is that Triton core (/ request) only need to know data pointer of the input and output, the conversion from shared memory handle to address is encapsulated within the frontend. And this can still stay true for reference counting the shared memory regions.

1. The `SharedMemoryManager` will still be the centralized location for reference counting, but it doesn't expose inc/dec function directly. Instead, it manages it internally, let's use shared_ptr to be the method to maintain ref count:
  * on register, it creates a shared_ptr with deleter which will clean up the shared memory region when shared_ptr's ref count goes to 0
  * on unregister, it release the shared_ptr object that it's holding, then ref count --
  * (SharedMemoryManager API change) on `GetMemoryInfo`, it returns a copy of the holding shared_ptr, which automatically increase the ref_count.
2. Then it is up to the frontend to keep the shared_ptr returned from `GetMemoryInfo` valid until it is done with the shared memory, noticing that both frontends already have done book keeping of the shared memory region used for the request / response, so the shared_ptr can live in the same way as the book keeping information and to be released on corresponding request / output / response release callbacks. In such a way, the reference counting still share similar life cycle as the request but you don't need to inject it into the Triton (core) request object.

There will be corner cases where the use will unregister and re-register the same shared memory region during an inference, but can be addressed through careful design of shared_ptr deleter and SharedMemoryManger modification.

src/shared_memory_manager.h
```diff
@@ -90,11 +132,19 @@ class SharedMemoryManager {
...
+  /// or set to nullptr when no longer needed, to decrease the count and allow
+  /// unregistration.
+  /// \return a TRITONSERVER_Error indicating success or
+  /// failure.
   TRITONSERVER_Error* GetMemoryInfo(
       const std::string& name, size_t offset, size_t byte_size,
       void** shm_mapped_addr, TRITONSERVER_MemoryType* memory_type,
-      int64_t* device_id);
+      int64_t* device_id,
+      std::shared_ptr<const SharedMemoryInfo>* shm_info_ref);
```
GuanLuo, line 147: I think that `shm_info` already contains information that covers other args
```suggestion
      const std::string& name, 
      std::shared_ptr<const SharedMemoryInfo>* shm_info_ref);
```

pskiran1: Yes, Guan. I think if we only return `shm_info_ref` from this method, we might need to move the [validations and shm_mapped_address casting code](https://github.com/triton-inference-server/server/blob/main/src/shared_memory_manager.cc#L472-L502) out of this method. This could result in multiple places where we call `GetMemoryInfo()` and need to rewrite these validations. Or I think we can write another method (as part of this class) that takes `shm_info_ref` as input and performs these validations.
Could you please let me know if you have any thoughts on this.
Thank you.

src/grpc/infer_handler.cc
```diff
@@ -168,7 +168,9 @@ TRITONSERVER_Error* InferGRPCToInput(
     const std::shared_ptr<SharedMemoryManager>& shm_manager,
     const inference::ModelInferRequest& request,
     std::list<std::string>* serialized_data,
-    TRITONSERVER_InferenceRequest* inference_request);
+    TRITONSERVER_InferenceRequest* inference_request,
+    std::vector<std::shared_ptr<const SharedMemoryManager::SharedMemoryInfo>>*
+        ref_shm_regions);
```
GuanLuo, line 173: Can you double check if line 161-173 is needed? The function has been declared in the `infer_handler.h` and I am not sure why it's done again here.

src/grpc/infer_handler.cc
```diff
@@ -1100,6 +1131,8 @@ ModelInferHandler::InferResponseComplete(
   if (response_created) {
     delete response;
   }
+
+  delete response_release_payload;
```
GuanLuo, line 1135: @tanmayv25 can you confirm that the state is going to be reused and thus we should not associate anything that ties to a particular inference to the state lifecycle? This looks right to me if that is the case.

src/grpc/stream_infer_handler.h
```diff
@@ -115,6 +115,23 @@ class ModelStreamInferHandler
...
+
+    StreamResponseReleasePayload(
+        State* state,
+        std::vector<
+            std::shared_ptr<const SharedMemoryManager::SharedMemoryInfo>>&&
+            ref_shm_regions)
+        : state_(state), ref_shm_regions_(ref_shm_regions)
+    {
+    }
+  };
```
GuanLuo, line 133: This can be group with the one define in `infer_handler.h`? Looks the same to me except for the naming.

qa/L0_trt_shape_tensors/test.sh
```diff
@@ -45,7 +45,7 @@ CLIENT_LOG="./client.log"
 SHAPE_TENSOR_TEST=trt_shape_tensor_test.py
 
 SERVER=/opt/tritonserver/bin/tritonserver
-SERVER_ARGS="--model-repository=`pwd`/models"
+SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=1"
```
GuanLuo, line 48: Need this change?

src/shared_memory_manager.h
GuanLuo: I see. Yes then it's better to just keep what you have for now, as we are intended to return a partial buffer of the shared memory based on the request.

src/grpc/infer_handler.cc
pskiran1: Good catch. It seems that these methods are mistakenly declared again here. I have removed them.

qa/L0_trt_shape_tensors/test.sh
pskiran1: I added this to enable more detailed logging, which will be useful for analysis in case of job failure. Since verbose logging was not currently enabled, we had to reproduce the issue again locally to know the behavior of server.

src/shared_memory_manager.cc
```diff
@@ -561,11 +568,19 @@ SharedMemoryManager::GetStatus(
...
+            std::string(
+                "Unable to find cuda shared memory region: '" + name + "'")
+                .c_str());
+      } else {
+        return TRITONSERVER_ErrorNew(
+            TRITONSERVER_ERROR_NOT_FOUND,
+            std::string(
+                "Unable to find system shared memory region: '" + name + "'")
+                .c_str());
+      }
```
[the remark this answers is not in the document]
pskiran1: How about writing a simple function `SharedMemoryTypeString(TRITONSERVER_MemoryType memtype)` which returns `"system"` or `"CUDA"`. So we can write error message only once. We already have [TRITONSERVER_MemoryTypeString()](https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L606-L621), but it is not helpful for shared memory related.

src/grpc/infer_handler.h
```diff
@@ -299,7 +299,9 @@ InferAllocatorPayload(
     const inference::ModelInferRequest& request,
     std::list<std::string>&& serialized_data,
     std::shared_ptr<ResponseQueue<ResponseType>> response_queue,
-    AllocPayload<ResponseType>* alloc_payload)
+    AllocPayload<ResponseType>* alloc_payload,
+    std::vector<std::shared_ptr<const SharedMemoryManager::SharedMemoryInfo>>*
+        ref_shm_regions)
```
nnshah1, line 304: ❓ From a naming point of view we do not typically use the prefix "ref" - do we need it here? Or could these be ```shm_regions``` and elsewhere ```shm_info```

pskiran1: Thanks @nnshah1 for the note. I updated the names accordingly to `shm_regions_info`, and `shm_info`.

src/grpc/infer_handler.cc
```diff
@@ -432,9 +422,17 @@ InferGRPCToInput(
...
       void* tmp;
+      std::shared_ptr<const SharedMemoryManager::SharedMemoryInfo> shm_info =
+          nullptr;
       RETURN_IF_ERR(shm_manager->GetMemoryInfo(
-          region_name, offset, byte_size, &tmp, &memory_type, &memory_type_id));
+          region_name, offset, byte_size, &tmp, &memory_type, &memory_type_id,
+          &shm_info));
       base = tmp;
+
+      if (shm_info != nullptr) {
```
GuanLuo, line 432: Should be able to remove this, a successful call to `GetMemoryInfo` should return valid `shm_info`.

src/grpc/infer_handler.h
```diff
@@ -1263,6 +1273,23 @@ class InferHandler : public HandlerBase {
...
+    State* state_;
+    std::vector<std::shared_ptr<const SharedMemoryManager::SharedMemoryInfo>>
+        shm_regions_info_;
+
+    ResponseReleasePayload(
+        State* state,
+        std::vector<
+            std::shared_ptr<const SharedMemoryManager::SharedMemoryInfo>>&&
+            shm_regions_info)
+        : state_(state), shm_regions_info_(shm_regions_info)
```
GuanLuo, line 1288:
```suggestion
        : state_(state), shm_regions_info_(std::move(shm_regions_info))
```

GuanLuo (approved):

src/grpc/infer_handler.cc
tanmayv25: This is accurate understanding.

tanmayv25 (dismissed):

tanmayv25 (approved):

PR #7570, opened 2024-08-27
refactor: Use thinner API server with an engine interface

rmccorm4:
Refactor for https://github.com/triton-inference-server/server/pull/7561.

Start migration towards thinner API server with some simple Interfaces defined for both an "OpenAIFrontend" and an "LLMEngine". For now, the API server defers almost entirely to the "LLMEngine" to implement each corresponding method in the interface (`engine.chat()`, `engine.completion`, etc.), so the frontend can remain the same but different engines can be used as needed. There may be more of a middle ground in what each component implements later if needed.

The current implementation of the "TritonOpenAIFrontend" happens to use FastAPI under the hood, but it doesn't have to use it. The user-facing interface is currently just a `frontend.start()` and `frontend.stop()`, where the details underneath can use technology other than FastAPI as the implementer desires.

Example application usage:
```python
    # Initialize a Triton Inference Server pointing at some supported LLM models
    server: tritonserver.Server = tritonserver.Server(...).start()

    # Wrap tritonserver in an interface-conforming "LLMEngine"
    engine: TritonLLMEngine = TritonLLMEngine(server)

    # Pass "LLMEngine" to the "OpenAIFrontend" as the backbone for inference and model management
    frontend: FastApiFrontend = FastApiFrontend(
        engine=engine, host=args.host, port=args.port, log_level=args.uvicorn_log_level
    )

    # Blocking call 
    frontend.start()
```

Notes for Reviewers:
- [ ] I don't think these interfaces need to be "final" at this point in time for the short-term. The move towards an interface is  to primarily (a) simplify the responsibilities of each component, (b) facilitate unit testing, and (c) make it easier to manage introducing other features at the application level such as the `kserve` frontend bindings. The "final" interface should go through further review and discussion after this PR.

Follow-ups:
- [ ] Add lighter unit tests with a mock `LLMEngine`, so a running Triton Server isn't necessary for testing frontend behavior


python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -0,0 +1,419 @@
...
+                    object=ObjectType.model,
+                    owned_by="Triton Inference Server",
+                ),
+            )
+
+        return models
+
+    def chat(
+        self, request: CreateChatCompletionRequest
+    ) -> CreateChatCompletionResponse | Iterator[str]:
```
rmccorm4, line 112: Note: this should probably return `Iterator[CreateChatCompletionResponse]` instead of `Iterator[str]` for the streaming case, but would need to somehow wrap the generator to do the conversion from Object to SSE string on frontend side for each yielded item.

Same for completions.

python/openai/openai_frontend/main.py
```diff
@@ -75,19 +89,21 @@ def parse_args():
 
 if __name__ == "__main__":
     args = parse_args()
-    # NOTE: configurations can be passed to FastAPI app through a builder
-    # function like init_app in future, but using env vars for simplicity.
     os.environ["TRITON_MODEL_REPOSITORY"] = args.model_repository
```
rmccorm4, line 92: TODO: Cleanup env var usage

nnshah1:
let's not use the term openai engine - think it is too specific to the api - can we use something more generic like LLMEngine?

python/openai/openai_frontend/engine/engine.py
```diff
@@ -0,0 +1,91 @@
...
+from schemas.openai import (
+    CreateChatCompletionRequest,
+    CreateChatCompletionResponse,
+    CreateCompletionRequest,
+    CreateCompletionResponse,
+    Model,
+)
+
+
+class OpenAIEngine(Protocol):
```
nnshah1, line 41:
```suggestion
class LLMEngine(Protocol):
```

python/openai/openai_frontend/main.py
rmccorm4: Done: https://github.com/triton-inference-server/server/pull/7570/commits/cb4763eec8174da8f42471ece6da6fc8941e23fe

python/openai/openai_frontend/frontend/triton_frontend.py
```diff
@@ -0,0 +1,109 @@
...
+
+import uvicorn
+from engine.triton_engine import TritonLLMEngine
+from fastapi import FastAPI
+from fastapi.middleware.cors import CORSMiddleware
+from frontend.frontend import OpenAIFrontend
+from routers import chat, completions, models, observability
+
+
+class TritonOpenAIFrontend(OpenAIFrontend):
```
GuanLuo, line 39: Not really a `Triton` frontend, it can take any `Engine` types. This is actually a `FastAPI` frontend

python/openai/openai_frontend/main.py
```diff
@@ -73,21 +86,33 @@ def parse_args():
...
+    # Gracefully shutdown when receiving signals for testing and interactive use
+    signal.signal(signal.SIGINT, partial(signal_handler, server, frontend))
+    signal.signal(signal.SIGTERM, partial(signal_handler, server, frontend))
+
+    # Blocking call until killed or interrupted with SIGINT
+    frontend.start()
+
+
+if __name__ == "__main__":
+    main()
```
GuanLuo, line 118: Is this something common that the entry point logic is wrap into another function?

python/openai/openai_frontend/routers/chat.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
GuanLuo, line 1: I wonder if the `routers` should be part of the frontend specialization.

python/openai/openai_frontend/frontend/triton_frontend.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(
+        self,
+        host: str = "localhost",
+        port: int = 8000,
+        log_level: str = "info",
+        engine: Optional[TritonLLMEngine] = None,
```
GuanLuo, line 45: Doesn't seesm to be `Optional` in current implementation.

python/openai/openai_frontend/frontend/triton_frontend.py
rmccorm4: Do you prefer `FastAPIOpenAIFrontend` then?

python/openai/openai_frontend/main.py
rmccorm4: Relatively common:
- https://github.com/vllm-project/vllm/blob/9c71c97ae24ae6f5209a475149808e25554cfe99/vllm/scripts.py#L161-L162
- https://github.com/fastapi/fastapi/blob/9416e89bd7e3491157f8c962012aebd7be7ed5d6/scripts/deploy_docs_status.py#L106

Primarily just makes it so that this function could be imported from another script to be run directly rather than launching the script as a subprocess or something. I'm not taking advantage of this, so it makes no difference one way or the other currently.

python/openai/openai_frontend/frontend/triton_frontend.py
rmccorm4: I was torn if some endpoints like `/v1/models` should work and just return an empty list or something if no engine is attached - but will just mark it required for now based on current implementation.

python/openai/openai_frontend/routers/chat.py
rmccorm4: Can you elaborate? Just moved into the `frontend` folder? Or including the routes as part of the class/implementation?

Let me know if you have a specific preferred change.

python/openai/openai_frontend/frontend/triton_frontend.py
rmccorm4: Removed optional

python/openai/openai_frontend/routers/chat.py
GuanLuo: `frontend/fastapi/routers/...`

python/openai/openai_frontend/frontend/triton_frontend.py
GuanLuo: `FastApiFrontend`

rmccorm4: done

python/openai/openai_frontend/routers/chat.py
rmccorm4: done

rmccorm4:
```
============================================== 76 passed, 15 skipped in 169.19s (0:02:49) ==============================================
```

GuanLuo (approved):

PR #7595, opened 2024-09-04
build/test: RHEL8 EA3

fpetrini15:
#### What does the PR do?
 Goal: Incorporate SBSA builds/tests and extended the smoke tests for x86 and SBSA to include L0_sequence_batcher. 

#### Checklist
- [ X ] PR title reflects the change and is of format `<commit_type>: <Title>`
- [ X ] Changes are described in the pull request.
- [ X ] Verified that the PR passes existing CI.
- [ X ] Verified copyright is correct on all changed files.

#### Commit Type:
- [ X ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ X ] test

PyTorch Backend changes: https://github.com/triton-inference-server/pytorch_backend/pull/139
TensorRT Backend changes: https://github.com/triton-inference-server/tensorrt_backend/pull/99
TensorFlow Backend changes: https://github.com/triton-inference-server/tensorflow_backend/pull/106

### DIFF
--- a/qa/L0_sequence_batcher/test.sh
+++ b/qa/L0_sequence_batcher/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2018-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -182,6 +182,17 @@ export USE_SINGLE_BUFFER
 #   models2 - two instances with batch-size 2
 #   models4 - four instances with batch-size 1
 rm -fr *.log  models{0,1,2,4} queue_delay_models && mkdir models{0,1,2,4} queue_delay_models
+
+# Search BACKENDS to determine if a backend should be tested
+function should_test_backend() {
+  local target_backend=$1
+  for BACKEND in $BACKENDS; do
+    if [[ "${target_backend}" == "${BACKEND}" ]]; then
+      return 1
+    fi
+  done
+  return 0
+}
 
 # Get the datatype to use based on the backend
 function get_datatype () {
@@ -827,8 +838,14 @@ fi
 
 ### Start Preserve Ordering Tests ###
 
-# Test only supported on windows currently due to use of python backend models
-if [ ${WINDOWS} -ne 1 ]; then
+# FIXME: Test only supported on windows currently due to use of python backend models.
+# Now that Windows supports the PYBE, we should check that this tests works once Windows
+# CI is stable.
+
+# These subtests use python models. They should not be executed if 'python' is not one
+# of the backends under test.
+should_test_backend "python"
+if [ $? -eq 1 ] && [ ${WINDOWS} -ne 1 ]; then
     # Test preserve ordering true/false and decoupled/non-decoupled
     TEST_CASE=SequenceBatcherPreserveOrderingTest
     MODEL_PATH=preserve_ordering_models
--- a/qa/L0_sequence_batcher/test.sh
+++ b/qa/L0_sequence_batcher/test.sh
@@ -186,12 +186,11 @@ rm -fr *.log  models{0,1,2,4} queue_delay_models && mkdir models{0,1,2,4} queue_
 # Search BACKENDS to determine if a backend should be tested
 function should_test_backend() {
   local target_backend=$1
-  for BACKEND in $BACKENDS; do
-    if [[ "${target_backend}" == "${BACKEND}" ]]; then
-      return 1
-    fi
-  done
-  return 0
+  if [[ $(echo "${BACKENDS[@]}" | grep -c "${target_backend}") -ne 0 ]]; then
+    echo "true"
+    return
+  fi
+  echo "false"
 }
 
 # Get the datatype to use based on the backend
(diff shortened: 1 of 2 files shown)


qa/L0_sequence_batcher/test.sh
```diff
@@ -827,8 +838,14 @@ fi
...
-# Test only supported on windows currently due to use of python backend models
-if [ ${WINDOWS} -ne 1 ]; then
+# FIXME: Test only supported on windows currently due to use of python backend models.
+# Now that Windows supports the PYBE, we should check that this tests works once Windows
+# CI is stable.
+
+# These subtests use python models. They should not be executed if 'python' is not one
+# of the backends under test.
+should_test_backend "python"
+if [ $? -eq 1 ] && [ ${WINDOWS} -ne 1 ]; then
```
mc-nv, line 848: is this variable something we must introduce manually

fpetrini15: If you mean manually provide the arg list -- yes.

The context here is that the enclosed test below runs tests that use Python models. Traditionally, our tests are written such that we assume that all backends are supported and exclude certain tests based on the target platform (e.g., Windows, Jetson, etc.). This didn't strike me as the right approach. Instead, my thought was to leverage the `BACKENDS` variable, which should generally indicate which backends are supported, to include/exclude certain tests.

This was just an idea I was playing with for this PR, but I'm open to feedback. Alternatively, we can go back to our traditional approach and introduce an external $RHEL_TEST flag or something like we do for Windows here.

qa/L0_sequence_batcher/test.sh
```diff
@@ -183,6 +183,17 @@ export USE_SINGLE_BUFFER
 #   models4 - four instances with batch-size 1
 rm -fr *.log  models{0,1,2,4} queue_delay_models && mkdir models{0,1,2,4} queue_delay_models
 
+# Search BACKENDS to determine if a backend should be tested
+function should_test_backend() {
+  local target_backend=$1
```
nv-kmcgill53, line 188: Sorry for the low quality comment but you could simplify this to the following 
```
if echo "${BACKENDS[@]}" | grep -q "${target_backend}"; then
```

Basically list the backends in one string, and assuming its a distinct set of names, grep for the target backend in the set

qa/L0_sequence_batcher/test.sh
```diff
@@ -183,6 +183,17 @@ export USE_SINGLE_BUFFER
 #   models4 - four instances with batch-size 1
 rm -fr *.log  models{0,1,2,4} queue_delay_models && mkdir models{0,1,2,4} queue_delay_models
 
+# Search BACKENDS to determine if a backend should be tested
+function should_test_backend() {
```
nv-kmcgill53, line 187: This might make sore sense to be put in the `utils.sh` file for other tests

qa/L0_sequence_batcher/test.sh
nv-kmcgill53, line 848: Using the return value on a bash function (i.e. `$?`) is a bit dubious imo. I am more of the opinion of setting variables and then inspecting those. Perhaps @mc-nv has an opinion here?

fpetrini15: I completely agree, but it seems that in bash return values can only be accessed from `$?` (source: https://linuxize.com/post/bash-functions/#return-values).
Since writing this, I realize I can probably just do something like `echo "true"` and store it like `run_test=$(should_test_backend "python")`

qa/L0_sequence_batcher/test.sh
fpetrini15: I was thinking the same thing. Of course, this function will be useless if developers don't know about it and use it properly. If you think it makes sense to add a function like this, we can merge it and I can make an announcement about it.

qa/L0_sequence_batcher/test.sh
mc-nv: Recently I used following to determine either it windows or not, may be it can be suitable here...
```suggestion
if [[ -v WSL_DISTRO_NAME ]] || [[ -v MSYSTEM ]];; then
```

qa/L0_sequence_batcher/test.sh
mc-nv: I'm not a big fan of moving something for functional programming aside, and particularly in our case where we test cross OS and cross-platform. 
I have no idea about this function, but not everything should come out of `utilis.sh`, and for reproducing/testing purpose suppose the more simple and straight forward solution the better.
Note that any problem with those functions from `utils.sh` will impact all others tests that are uses it.

fpetrini15: In a sense @mc-nv I believe you're right. If people don't know about this function, we'll start observing mixed use-cases which is a mess (think about kill/wait vs. calling kill_server). I'll keep the function here if we're all in agreement.

qa/L0_sequence_batcher/test.sh
fpetrini15: This is neat! I modified the idea slightly to use the `-c` option.

qa/L0_sequence_batcher/test.sh
fpetrini15: We want the exact opposite check, but point taken and the code has been updated.

qa/L0_sequence_batcher/test.sh
nv-kmcgill53: Fine by me

nv-kmcgill53 (approved):

PR #7599, opened 2024-09-06
Don't Build `tritonfrontend` for Windows.

fpetrini15:
**Goal:** Don't build `tritonfrontend` for Windows.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -782,8 +782,10 @@ if (NOT WIN32)
   endif() # TRITON_ENABLE_GPU
 endif() # NOT WIN32
 
-# tritonfrontend python package
-add_subdirectory(python)
+if (NOT WIN32)
+  # tritonfrontend python package
+  add_subdirectory(python)
+endif (NOT WIN32)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -782,8 +782,11 @@ if (NOT WIN32)
   endif() # TRITON_ENABLE_GPU
 endif() # NOT WIN32
 
-# tritonfrontend python package
-add_subdirectory(python)
+# DLIS-7292: Extend tritonfrontend to build for Windows
+if (NOT WIN32)
```
rmccorm4, line 786: Were other parts broken other than the CURL/tracing/otel stuff?

rmccorm4 (approved):

src/CMakeLists.txt
rmccorm4: IIRC, the build would fail if individual parts like TRACING were disabled (which we have a follow-up ticket for) - so it would make sense to me if skipping tracing portions due to windows would cause the whole thing to fail.

PR #7605, opened 2024-09-06
build: `tritonfrontend` support for no/partial endpoint builds

KrishnanPrash:
#### What does the PR do?
Currently, for the `tritonfrontend` python package only works for builds where all the supported endpoints are enabled. And for builds with partial endpoints enabled, the building of the `tritonfrontend` wheel is skipped.  

This PR is meant to add support for partial endpoint builds for `tritonfrontend`. With these changes, `tritonfrontend` wheel will build in all cases and match the functionality of the `tritonserver` binary. 

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

`src/python/tritonfrontend/_c/tritonfrontend_pybind.cc` restricts portions of C++ code that is binded.
`src/python/tritonfrontend/_api/__init__.py` restricts which python apis are included in the tritonfrontend wheel.
`build.py` changes allow for wheel to be tested in CI

#### Test plan:
Added testing for `tritonfrontend` to variant builds test.

- CI Pipeline ID: 19165151


src/python/tritonfrontend/_api/__init__.py
```diff
@@ -23,3 +23,13 @@
...
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+try:
+    from ._kservehttp import KServeHttp
+except ImportError:
+    pass
+
+try:
+    from ._kservegrpc import KServeGrpc
+except ImportError:
```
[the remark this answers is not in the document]
rmccorm4: Does adding a comment above `pass` like `# Allow builds with gRPC disabled` make this warning go away?

src/python/tritonfrontend/__init__.py
```diff
@@ -29,5 +29,4 @@
 import builtins
 from importlib.metadata import PackageNotFoundError, version
 
-from tritonfrontend._api._kservegrpc import KServeGrpc
-from tritonfrontend._api._kservehttp import KServeHttp
+from tritonfrontend._api import *
```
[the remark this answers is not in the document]
rmccorm4: Can you be more prescriptive and import specific things here?

rmccorm4 (commented):
Resolve CodeQL warnings

src/python/CMakeLists.txt
```diff
@@ -25,16 +25,9 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 cmake_minimum_required(VERSION 3.18)
-
-message("tritonfrontend python package build skipped when relevant frontends are disabled.")
-message("In order to build tritonfrontend, the following flags are needed: -DTRITON_ENABLE_HTTP=ON -DTRITON_ENABLE_GRPC=ON")
-
 # [DLIS-7232] tritonfrontend package expects all supported packages to be
 # built, without any check/verification for respective frontend enable flags.
 # Support for partial builds(ex: HTTP but not gRPC) will be addressed later.
```
rmccorm4, line 30: > # Support for partial builds(ex: HTTP but not gRPC) will be addressed later.

This is addressed by this PR, right?

KrishnanPrash: Thank you for pointing it out. Just removed the comment.

build.py
```diff
@@ -1857,6 +1857,12 @@ def core_build(
         os.path.join(install_dir, "python"),
     )
 
+    cmake_script.mkdir(os.path.join(install_dir, "python"))
+    cmake_script.cp(
+        os.path.join(repo_install_dir, "python", "tritonfrontend*.whl"),
```
rmccorm4, line 1862: Is this needed? `triton*.whl` right above this

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -96,13 +94,14 @@ endif()
...
-  find_package(CURL CONFIG REQUIRED)
-  find_package(nlohmann_json CONFIG REQUIRED)
-  find_package(opentelemetry-cpp CONFIG REQUIRED)
-  list(APPEND PY_BINDING_DEPENDENCY_LIBS
-    tracing-library
-  )
+  # find_package(absl CONFIG REQUIRED)
+  # find_package(CURL CONFIG REQUIRED)
+  # find_package(nlohmann_json CONFIG REQUIRED)
+  # find_package(opentelemetry-cpp CONFIG REQUIRED)
```
rmccorm4, line 100: remove if not needed

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -122,7 +121,7 @@ pybind11_add_module(
   ${PYTHON_FRONTEND_BINDING_SRCS}
 )
 
-target_include_directories(py-bindings PRIVATE ${CMAKE_SOURCE_DIR}/src)
+# target_include_directories(py-bindings PRIVATE ${CMAKE_SOURCE_DIR}/src)
```
rmccorm4, line 124: remove if not needed

src/CMakeLists.txt
```diff
@@ -521,15 +521,21 @@ if(${TRITON_ENABLE_TRACING})
   if (NOT WIN32)
     target_include_directories(
       tracing-library
-      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+      PUBLIC ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
```
rmccorm4, line 524: @fpetrini15 any general wisdom on necessity of PUBLIC vs PRIVATE and why it fixes it? or @KrishnanPrash

fpetrini15: IIUC, by specifying `tracing-library`'s dependencies as private, any other targets that depend on `tracing-library` will not be able to see its dependencies (not inherited)--they will have to link them again individually. In this case, it appears that to compile the `tracing-library`, CMake needs to have access to the open telemetry include directories. From my meetings with @KrishnanPrash, it seems like anything that depends on `tracing-library` also needs to have access to this directory. In this case, it would make sense to expose this as a public dependency.

Take this with a grain of salt, however, as I am no CMake expert 😅

KrishnanPrash: Changing from PRIVATE to PUBLIC allows the directories to be included with any library that links to the tracing-library, which in this case is py-bindings.

build.py
KrishnanPrash: Thank you for the catch. Just removed it.

src/CMakeLists.txt
```diff
@@ -521,15 +521,21 @@ if(${TRITON_ENABLE_TRACING})
...
     )
 
     target_link_libraries(
       tracing-library
-      PRIVATE
+      PUBLIC
       ${OPENTELEMETRY_CPP_LIBRARIES})
   endif()
 
+  set_target_properties(
```
fpetrini15, line 533: Is this modification still required?

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -153,14 +146,21 @@ if(${TRITON_ENABLE_GPU})
...
+    target_include_directories(
       py-bindings
-      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+      PUBLIC ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+    )
+
+    target_link_libraries(
+      py-bindings
+      PUBLIC
+      ${OPENTELEMETRY_CPP_LIBRARIES}
```
fpetrini15, line 157: Now that these libs are exposed publicly by the `tracing-library` that is already linked to the `py-bindings` target, do we need to do it again here?

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -153,14 +146,21 @@ if(${TRITON_ENABLE_GPU})
 endif()
 
 if(${TRITON_ENABLE_TRACING})
-  target_include_directories(
+    target_include_directories(
       py-bindings
-      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
+      PUBLIC ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
```
fpetrini15, line 151: I think this may also be unnecessary now for the same reason as described below.

src/CMakeLists.txt
KrishnanPrash: Without the `-fPIC` flag set to ON, the CMake build process errors out.

src/python/tritonfrontend/CMakeLists.txt
KrishnanPrash: Addressed in comment above.

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -96,13 +94,10 @@ endif()
 
 if(${TRITON_ENABLE_TRACING})
   message("TRACING/STATS IS CURRENTLY NOT SUPPORTED.")
```
GuanLuo, line 96: I don't understand this part, if not supported, then why introducing the dependency?

src/CMakeLists.txt
KrishnanPrash: Here is a copy of the error message I received when building without :
image

src/python/tritonfrontend/CMakeLists.txt
KrishnanPrash: Currently, when building through `build.py` or `cmake`, tracing can be enabled in the build. From my understanding, this means that `tracer.h` will get built with the `#ifdef TRITON_ENABLE_TRACING` [sections included](https://github.com/triton-inference-server/server/blob/fde6e5887775b1236e25192601b6cd1d7abe3620/src/tracer.h#L40C1-L57C1). Hence, when `http-endpoint-library` or `grpc-endpoint-library` include `tracer.h` ([here](https://github.com/triton-inference-server/server/blob/fde6e5887775b1236e25192601b6cd1d7abe3620/src/http_server.h#L44C1) and [here](https://github.com/triton-inference-server/server/blob/fde6e5887775b1236e25192601b6cd1d7abe3620/src/grpc/grpc_server.h#L35)) it will look for the tracing dependencies that `tracer.h` expects

GuanLuo: > Hence, when http-endpoint-library or grpc-endpoint-library include tracer.h ([here](https://github.com/triton-inference-server/server/blob/fde6e5887775b1236e25192601b6cd1d7abe3620/src/http_server.h#L44C1) and [here](https://github.com/triton-inference-server/server/blob/fde6e5887775b1236e25192601b6cd1d7abe3620/src/grpc/grpc_server.h#L35)) it will look for the tracing dependencies that tracer.h requires.

Then shouldn't `tracing-library` be added as `http-endpoint-library` / `grpc-endpoint-library`'s dependencies?

src/CMakeLists.txt
GuanLuo: Wouldn't hurt to compile with position independent code.. But have you ever looked into compiling the binding as static library? It was mentioned in the design doc right? Was it not feasible?

src/python/tritonfrontend/CMakeLists.txt
GuanLuo: But how can `main`target don't need handling like this? Isn't `main` and `py-bindings` conceptually the same in terms of compilation and linking?

src/CMakeLists.txt
KrishnanPrash: Initially attempted to compiling the binding as a static library as outlined in the design doc ([link](https://github.com/pybind/cmake_example/issues/11#issuecomment-405092832)):
```
add_library(supporting_library STATIC ...)
set_target_properties(supporting_library PROPERTIES POSITION_INDEPENDENT_CODE ON)
...
pybind11_add_module(pymodule MODULE ...)
target_link_libraries(pymodule PRIVATE supporting_library)
```
My initial approach was to link the dependencies(`http-endpoint-library` and `grpc-endpoint-libarry` to the `supporting_library`) and follow a similar procedure, but it resulted in symbol not found errors from the `pymodule` library.

GuanLuo: What symbols are not found..

KrishnanPrash: Apologies for not providing enough details. IIRC the symbol not found errors would be for a mangled function name within `triton::server::HTTPServer` / `triton::server::HTTPAPIServer`.

rmccorm4: @KrishnanPrash please try the relevant build and share the detailed errors with specific symbols - this will help the reviewers better understand the problems and decisions made

src/python/tritonfrontend/CMakeLists.txt
KrishnanPrash: Thank you for the suggestion. The new handling of `tracing-library` requires only linking `py-bindings` to `tracing-library` without including/linking any `opentelemetry` directories or libraries directly to `py-bindings`.

src/CMakeLists.txt
```diff
@@ -249,14 +249,6 @@ if(${TRITON_ENABLE_TRACING})
     main
     PRIVATE TRITON_ENABLE_TRACING=1
   )
-# FIXME: remove, when Windows support is added for Opentelemetry
```
KrishnanPrash, line 252: moved this comment to right above the tracing-library section.

src/python/tritonfrontend/CMakeLists.txt
KrishnanPrash: I think this would be a good change. Currently instead of linking the `tracing-library`, we do something along the lines of:
http_server.h
```cpp
...
#include "tracer.h"
...
```
and in CMakeList.txt to handle the dependencies we do:
```cmake
target_include_directories(
      http-endpoint-library
      PRIVATE ${OPENTELEMETRY_CPP_INCLUDE_DIRS}
    )
```
But now that we are changing the `tracing-library` links/includes to be public, I replaced:
```cmake
target_include_directories(
main/http-endpoint-library/grpc-endpoint-library
PRIVATE 
${OPENTELEMETRY_CPP_INCLUDE_DIRS})
```
with this:
```cmake
target_link_libraries(
main/http-endpoint-library/grpc-endpoint-library
PRIVATE
tracing-library)
```

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -96,17 +94,76 @@ endif()
...
+      tracing-library
   )
 endif()
 
 # ===================== End of Collection ===================================
 
+# ========== Create support-lib (Static Lib for Dependencies) ================
+add_library(
+  support-lib
+    STATIC
```
rmccorm4, line 108: Just curious - What's the file size difference of the wheel when using STATIC vs before?

KrishnanPrash: Wheel size was roughly 10 MB before using STATIC and after as well.

rmccorm4 (commented):
LGTM, but defer to Guan/Francesco

src/python/tritonfrontend/CMakeLists.txt
KrishnanPrash: With assistance from @fpetrini15, I was able to verify that these changes work with the windows build as well.
CI Pipeline ID: 19374177

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -170,12 +220,19 @@ if(${TRITON_ENABLE_STATS})
...
 set_property(TARGET py-bindings PROPERTY OUTPUT_NAME tritonfrontend_bindings)
 
 set_target_properties(
     py-bindings
     PROPERTIES
       BUILD_RPATH "$ORIGIN:/opt/tritonserver/lib"
+      POSITION_INDEPENDENT_CODE ON
 )
-# ===================== End of Python Bindings ==============================
+# ===================== End of Python Bindings ==============================
```
GuanLuo, line 238: new line

GuanLuo (approved):
Left comments, approve to unblock

src/python/tritonfrontend/CMakeLists.txt
```diff
@@ -118,16 +175,13 @@ set(
...
   ${PYTHON_FRONTEND_BINDING_SRCS}
 )
 
-target_include_directories(py-bindings PRIVATE ${CMAKE_SOURCE_DIR}/src)
-
 target_link_libraries(
     py-bindings
     PRIVATE
-      ${PY_BINDING_DEPENDENCY_LIBS}
+    support-lib
```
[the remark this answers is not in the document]
KrishnanPrash: 1. The purpose behind the `support-lib` target was to follow the structure outlined in the design doc ([Link](https://github.com/triton-inference-server/server/pull/7605#discussion_r1797390314)).
2. Yes, the result will be identitical if you directly link against `${PY_BINDING_DEPENDENCY_LIBS}`

GuanLuo (approved):

PR #7617, opened 2024-09-16
Fix: Add mutex lock for state completion check in gRPC streaming to prevent race condition

pskiran1:
#### What does the PR do?
This PR aims to fix an occasional error (caused by a race condition) that occurs when making multiple asynchronous requests to the ASR model in decoupled mode:
```
E0829 13:41:44.609857 147 stream_infer_handler.cc:528] "Should not print this! Decoupled should NOT write via WRITEREADY!"
```

#### Checklist
- [X] PR title reflects the change and is of format `<commit_type>: <Title>`
- [X] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [X] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ ] Added [test plan](#test-plan) and verified test passes.
- [X] Verified that the PR passes existing CI.
- [X] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [X] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 18428769

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #xxx

PR #7618, opened 2024-09-17
Cherry-pick: Fix: Add mutex lock for state completion check in gRPC streaming to prevent race condition

pskiran1:
#### What does the PR do?
Cherry-pick #7617 

- CI Pipeline ID:


pskiran1:
Created for wrong target.

PR #7624, opened 2024-09-18
fix: usage of ReadDataFromJson in array tensors

v-shobhit:
#### What does the PR do?

The `generate` and `generate_stream` endpoints did not seem to work when directly querying TRTLLM backend with input tokens. This is because the `HTTPAPIServer::GenerateRequestClass::ExactMappingInput` does not send the correct size of an array input to `ReadDataFromJson`.

This PR also fixes https://github.com/triton-inference-server/tensorrtllm_backend/issues/369

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

Added a new test case to `L0_http` job.
Internal CI pipeline id: 18800660

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_http/generate_models/mock_llm/config.pbtxt
+++ b/qa/L0_http/generate_models/mock_llm/config.pbtxt
@@ -1,4 +1,4 @@
-# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -41,6 +41,12 @@ input [
     name: "STREAM"
     data_type: TYPE_BOOL
     dims: [ 1, 1 ]
+  },
+  {
+    name: "input_ids"
+    data_type: TYPE_INT32
+    dims: [ 1, -1 ]
+    optional: true
   }
 ]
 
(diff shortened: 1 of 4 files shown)


qa/L0_http/generate_models/mock_llm/config.pbtxt
```diff
@@ -41,6 +41,12 @@ input [
     name: "STREAM"
     data_type: TYPE_BOOL
     dims: [ 1, 1 ]
+  },
+  {
+    name: "input_ids"
```
rmccorm4, line 46: Do we need a new input to test this? Or can we validate the same behavior (support input with list of size > 1 IIUC), by sending multiple prompts? ex: `"PROMPT": ["hello", "world"]` with updated `dims: [1, -1]`

GuanLuo (approved):
LGTM, just to confirm that the added test case **will fail** without this `http_server.cc` change?

pskiran1:
> LGTM, just to confirm that the added test case will fail without this http_server.cc change?

@GuanLuo , thanks for the note. The test case passed with both old and new code (with http_server.cc change).
I had to undo the commit 229e5e85eeef481916d8bf67f24bcf9dfcb68b25, I think this bug is not happening for the String data type.
Could you please review the latest code and approve it?

Now the test case fails on 24.09 with the old code and passes with the new code (including the http_server.cc change).

GuanLuo (approved):

PR #7666, opened 2024-09-27
feat: Add copyright hook

pranavm-nvidia:
#### Checklist
- [ ] I have read the [Contribution guidelines](#../../CONTRIBUTING.md) and signed the [Contributor License
- [ ] I ran pre-commit locally (`pre-commit install, pre-commit run --all`)

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

The main files to be reviewed here are `tools/add_copyright.py`, which provides a new pre-commit hook that will update or add copyright headers to new or modified files, and `.pre-commit-config.yaml`. The logic is more complicated than one might expect because the repository includes many different types of files that require slightly different handling. 

#### Caveats:
The hook expects a certain format for the copyright header. In cases where the file already contains a copyright header in a slightly different format, the header might be inserted twice (the hook will emit a warning in such cases). I have standardized the copyright header format as far as possible, but there could be files which I missed. 

#### Background
Right now, there is no automatic way of updating/inserting copyright headers in the various source files in the repository. Consequently, some files have incorrect copyright years and many don't include headers at all. With this hook, the header no longer needs to be added or changed manually and will automatically be updated when the file is touched.


pranavm-nvidia:
@rmccorm4 I can revert the changes to the copyright headers if it makes it easier to review. Let me know what the rules are for copyrights and I can implement those. My understanding right now is that it should be of the format `Copyright <optional_first_modified_year>-<last_modified_year>`

rmccorm4:
I believe you will find the following formats throughout the code:
1. `Copyright (c) YYYY` - never touched in a new year after creation year
2. `Copyright YYYY` - I don't know if these are valid, or invalid and just snuck in due to human error
3. `Copyright YYYY-YYYY` - Intentionally removed `(c)` to avoid line wrapping with multi-year

I think supporting all of these would provide the least number of files changed and max utility - and then I can separately follow-up to see if we should disallow or remove (2) after confirmation with legal.

rmccorm4:
CC @GuanLuo for viz

tools/add_copyright.py
```diff
@@ -0,0 +1,327 @@
...
+        with open(path, "w") as f:
+            f.write(updated_content)
+
+
+def update_and_get_license() -> str:
+    """
+    Updates the copyright year in the LICENSE file if necessary and then
+    returns its contents.
+    """
+    # TODO: Check if this is right - if the license file needs to have a range,
```
pranavm-nvidia, line 78: Do we want the license file to contain a range? Right now it only has a single year.

pranavm-nvidia:
@rmccorm4 I think this is ready for review now. See the included tests for what the expected behavior is with the various copyright formats.

tools/add_copyright.py
rmccorm4: This is a good question, I'll need to check.

rmccorm4 (commented):
PR looks great to me! Let me run it by the team to make sure everyone is good with it 👍

I think this will save us from a lot of human error and extra commits saying "fix copyright", and remove a step from our PR template checklist!

tools/add_copyright.py
```diff
@@ -0,0 +1,359 @@
...
+
+#
+# Integration Tests
+#
+import tempfile
+
+import pytest
+
+
+# Processes provided text through the copyright hook by writing it to a temporary file.
```
oandreeva-nv, line 283: nit: is there a reason why `imports` are here, and following function definitions are here. Can we move them to a pre- `if __name__ == "__main__"` ?

tools/add_copyright.py
```diff
@@ -0,0 +1,359 @@
...
+@register(has_ext([".cc", ".h"]))
+def cpp(path):
+    update_or_add_header(path, prefix_lines(LICENSE_TEXT, "// "))
+
+
+@register(has_ext([".tpl"]))
+def tpl(path):
+    update_or_add_header(path, "{{/*\n" + prefix_lines(LICENSE_TEXT, "# ") + "\n*/}}")
+
+
```
oandreeva-nv, line 243:
```suggestion
```
Suggestion to remove this for now since we don't use `.tpl` afaik

tools/add_copyright.py
```diff
@@ -0,0 +1,359 @@
...
+    # sys.exit is important here to avoid the test-related imports below during normal execution.
+    sys.exit(main())
+
+
+#
+# Integration Tests
+#
+import tempfile
+
+import pytest
```
oandreeva-nv, line 280: nit: can they be sorted? or there are some dependency issues?

oandreeva-nv:
lgtm in general, left some questions/suggestions

tools/add_copyright.py
pranavm-nvidia: They're grouped according to `stdlib` modules -> external modules -> internal modules, which seems to be the way imports are grouped in other files (e.g. `deploy/mlflow-triton-plugin/mlflow_triton/deployments.py`). I can change it to be purely alphabetical if you prefer that though.

tools/add_copyright.py
pranavm-nvidia: Sure, should I also remove the `.tpl` files in the repo?

oandreeva-nv: please ignore, I was looking in the wrong place, we do have them under `deploy`

tools/add_copyright.py
oandreeva-nv: Thanks for clarification. I'm fine with the current structure. Started codeql a pre-commit checks.

tools/add_copyright.py
```diff
@@ -0,0 +1,359 @@
...
+    )
+    parser.add_argument("files", nargs="*")
+
+    args, _ = parser.parse_known_args()
+    add_copyrights(args.files)
+
+
+if __name__ == "__main__":
+    # sys.exit is important here to avoid the test-related imports below during normal execution.
+    sys.exit(main())
```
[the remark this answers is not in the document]
pranavm-nvidia: Updated to return 0 on success.

rmccorm4 (dismissed):

rmccorm4:
@oandreeva-nv good to ship it?

oandreeva-nv (dismissed):
LGTM

tools/add_copyright.py
```diff
@@ -0,0 +1,360 @@
...
+
+@register(has_ext([".html", ".md"]))
+def html_md(path):
+    update_or_add_header(path, "<!--\n" + prefix_lines(LICENSE_TEXT, "# ") + "\n-->")
+
+
+def add_copyrights(paths):
+    for path in paths:
+        for match, handler in FILE_TYPE_HANDLERS.items():
+            if match(path):
```
GuanLuo, line 252: Can there be warning printed for files that we have not registered handle for?

pranavm-nvidia: Added

GuanLuo (approved):

rmccorm4 (approved):

PR #7501, opened 2024-08-05
feat: Python Deployment of Triton Inference Server

KrishnanPrash:
#### What does the PR do?
This PR is for the `tritonfrontend` python package, containing bindings to the `HTTPAPIServer` and `grpc::Server` classes in the frontend. This allows Triton users to start their respective frontends from within python.

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [x] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Where should the reviewer start?

`server/src/python/tritonfrontend/_c/tritonfrontend_pybind.cc`

#### Test plan:
Working on adding tests, which can be run through pytest, and added as a single command to the existing `L0_python_api` bash script.

- CI Pipeline ID:

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #xxx

PR #7671, opened 2024-09-29
fix: `tritonfrontend` gRPC Streaming Segmentation Fault

KrishnanPrash:
#### What does the PR do?
Currently, the `tritonfrontend` bindings do not support a tracing because a `nullptr` is passed as a `TraceManager` object to the respective frontends. Due to a lack of checks for valid `TraceManager` object, any operations issued assuming the `TraceManager` object is valid, leads to a segmentation fault. 

This PR wraps tracing operations in a `if(trace_manager_obj)` to ensure the `TraceManager` is not `nullptr`.

Additional Changes:
1. Added test case `test_streaming_inference()` in `server/qa/L0_python_api/test_kserve.py` to catch errors in CI pipeline
2. Added beta tag to `tritonfrontend` README.md.
3. Refactored how `testing_utils` is imported and used to prevent shadowing based on this [suggestion](https://github.com/triton-inference-server/server/pull/7501#discussion_r1736907991).

#### Checklist

#### Test plan:
Added `server/qa/L0_python_api/test_kserve.py::test_streaming_inference()`

- CI Pipeline ID: 18855908


qa/L0_python_api/testing_utils.py
```diff
@@ -81,15 +83,63 @@ def send_and_test_inference_identity(frontend_client, url: str) -> bool:
...
+    while completed_requests != num_requests:
+        result = results.get()
+        if isinstance(result, InferenceServerException):
+            if result.status() == "StatusCode.CANCELLED":
+                completed_requests += 1
+            raise result
+
+        # Processing Response
+        text_output = result.as_numpy("text_output")[0].decode()
+        is_final = result.get_response().parameters.get("triton_final_response", False)
```
rmccorm4, line 136: I believe this should be something like 
```python
result.get_response().parameters.get("triton_final_response", False).bool_param
```

Without the `bool_param` part, you might get some truthy value (non-empty dict) that looks like `{"bool_param": False}` or something, and get unexpected behavior.

rmccorm4: See https://github.com/triton-inference-server/triton_cli/blob/36a425e74c91402e7a6c9ac818aa49f3a5c41674/src/triton_cli/client/client.py#L288-L297

qa/L0_python_api/test_model_repository/identity/config.pbtxt
```diff
@@ -30,14 +30,14 @@ max_batch_size: 0
 
 input [
   {
-    name: "INPUT0"
+    name: "text_input"
```
GuanLuo, line 33: Why changing the IO names?

KrishnanPrash: Changed the IO names to make the model compatible with the Triton CLI.

src/grpc/stream_infer_handler.cc
```diff
@@ -324,12 +324,14 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
...
-      GrpcServerCarrier carrier(state->context_->ctx_.get());
-      auto start_options =
-          trace_manager_->GetTraceStartOptions(carrier, request.model_name());
-      state->trace_ = std::move(trace_manager_->SampleTrace(start_options));
-      if (state->trace_ != nullptr) {
-        triton_trace = state->trace_->trace_;
+      if (trace_manager_) {
+        GrpcServerCarrier carrier(state->context_->ctx_.get());
+        auto start_options =
+            trace_manager_->GetTraceStartOptions(carrier, request.model_name());
```
oandreeva-nv, line 330: I see now. I think this is not the only potential place for a SegFault. Do we want to fix it in other places as well? Or we're targeting `streaming` case at the moment?

KrishnanPrash: Would be happy to add checks in other places as well. Could you provide an example where these checks would be needed?

tanmayv25: @KrishnanPrash Can you look for TRITON_ENABLE_TRACING blocks and break the logic into separate utility function in grpc_utils.h?
This function will take trace_manager_ as input. If trace_manager_ is nullptr then the logic is skipped, otherwise the same logic is run.

oandreeva-nv: I think this place (HandleGenerate) seem to be relevant: https://github.com/triton-inference-server/server/blob/dbb064ff930c876b5de46088f675bd60756c2969/src/http_server.cc#L3235-L3237

I can see that HandleInfer on HTTP is guarded: https://github.com/triton-inference-server/server/blob/dbb064ff930c876b5de46088f675bd60756c2969/src/http_server.cc#L3599-L3602

qq to @GuanLuo , I can see that you've added guards for `HandleInfer` (link above), was there a reason not to guard `HandleGenerate` ?

oandreeva-nv: @KrishnanPrash, do you need to support trace update with this PR?

KrishnanPrash: @oandreeva-nv thank you for pointing it out. Added `nullptr` check for `StartTrace()` in `HandleGenerate()`.

KrishnanPrash: @tanmayv25 If it's ok, will move tracing logic to `grpc_utils.cc/h` as helper functions in a follow-up PR [DLIS-7380]

qa/L0_python_api/test_model_repository/identity/config.pbtxt
oandreeva-nv: I would recommend running the general pipeline + L0_backend_python to see if any tests depend on input names

docs/customization_guide/tritonfrontend.md
```diff
@@ -25,7 +25,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 -->
-### Triton Server (tritonfrontend) Bindings
```
oandreeva-nv, line 28: nit: try to keep line length under 80 characters, including READMEs

src/http_server.cc
```diff
@@ -1809,7 +1809,12 @@ HTTPAPIServer::HandleTrace(evhtp_request_t* req, const std::string& model_name)
     return;
   }
 
+
```
oandreeva-nv, line 1812: no need for the extra line?

src/http_server.cc
```diff
@@ -3233,8 +3238,11 @@ HTTPAPIServer::HandleGenerate(
 
   // If tracing is enabled see if this request should be traced.
   TRITONSERVER_InferenceTrace* triton_trace = nullptr;
-  std::shared_ptr<TraceManager::Trace> trace =
-      StartTrace(req, model_name, &triton_trace);
+  std::shared_ptr<TraceManager::Trace> trace;
```
oandreeva-nv, line 3241: Let's also trigger `L0_trace` tests in the new pipeline

src/grpc/stream_infer_handler.cc
GuanLuo: > I can see that you've added guards for HandleInfer (link above), was there a reason not to guard HandleGenerate ?

Probably just a missed spot..

src/grpc/stream_infer_handler.cc
```diff
@@ -324,12 +324,14 @@ ModelStreamInferHandler::Process(InferHandler::State* state, bool rpc_ok)
     if (err == nullptr) {
       TRITONSERVER_InferenceTrace* triton_trace = nullptr;
 #ifdef TRITON_ENABLE_TRACING
-      GrpcServerCarrier carrier(state->context_->ctx_.get());
-      auto start_options =
-          trace_manager_->GetTraceStartOptions(carrier, request.model_name());
-      state->trace_ = std::move(trace_manager_->SampleTrace(start_options));
-      if (state->trace_ != nullptr) {
-        triton_trace = state->trace_->trace_;
+      if (trace_manager_) {
```
nnshah1, line 327: small thing: in some places we use ```<variable> !=nullptr```  - let's choose one for consistency.

src/grpc/stream_infer_handler.cc
nnshah1: I see in the HandleInfer() method - that the trace_manager_ check is not guarded with #IF TRACING compile time macro - should it be?

I guess secondary question - if tracing is compiled in - should trace_manager_ be null? - could we make that a pre-condition instead of runtime check if tracing is enabled? (just a question)

KrishnanPrash: I don't believe `trace_manager_` needs to be guarded with a `#ifdef TRACING_ENABLE_TRACING` compile time macro because `StartTrace()` function has it's own check for if tracing is enabled. 

As for the secondary question, if tracing is compiled in, `trace_manager_` should technically never be `null` after being passed to the services. However, because the bindings do not yet support tracing, a hopefully temporary situation arises of tracing being enabled, but `trace_manager_` being null.

src/grpc/stream_infer_handler.cc
KrishnanPrash: Made check to be `if(trace_manager != nulltpr)`

docs/customization_guide/tritonfrontend.md
```diff
@@ -25,9 +25,15 @@
...
-
-The `tritonfrontend` python package is a set of bindings to Triton's existing frontends implemented in C++. Currently, `tritonfrontend` supports starting up `KServeHttp` and `KServeGrpc` frontends. These bindings used in-combination with Triton's Python In-Process API ([`tritonserver`](https://github.com/triton-inference-server/core/tree/main/python/tritonserver)) and [`tritonclient`](https://github.com/triton-inference-server/client/tree/main/src/python/library) extend the ability to use Triton's full feature set with a couple of lines of Python.
+### Triton Server (tritonfrontend) Bindings (Beta)
+
+The `tritonfrontend` python package is a set of bindings to Triton's existing
```
oandreeva-nv, line 30: happy_ben

qa/L0_python_api/testing_utils.py
```diff
@@ -93,3 +95,58 @@ def send_and_test_inference_identity(frontend_client, url: str) -> bool:
 
     teardown_client(client)
     return input_data[0] == output_data[0].decode()
+
+
+# Sends a streaming inference request to test_model_repository/identity model
```
oandreeva-nv, line 100: nit: personal preference is to follow Google style guide for docstrings&comments: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

src/http_server.cc
```diff
@@ -3233,8 +3237,11 @@ HTTPAPIServer::HandleGenerate(
...
+  if (trace_manager_) {
```
oandreeva-nv, line 3241: do we want to test http frontend to make sure there are no segfaults? or this test was added in previous PRs?

qa/L0_python_api/testing_utils.py
KrishnanPrash: Added TODO to fix in future PR

src/http_server.cc
KrishnanPrash: Just added a test case for http generate as well.

qa/L0_python_api/test_kserve.py
```diff
@@ -29,18 +29,10 @@
 
 import numpy as np
 import pytest
+import testing_utils as utils
```
nnshah1, line 32: small nit: do we need to rename it or just use "testing_utils", or "test_utils"

src/grpc/stream_infer_handler.cc
nnshah1: ah so this is temporary? - once enable tracing is added we can remove the runtime check?

KrishnanPrash: We could remove the runtime check, but personally I am in favor of keeping these checks because it allows to catch unexpected behavior earlier, rather than later. 

After tracing support is provided in `tritonfrontend`, we could probably modify these checks and return an error to fail earlier with a cleaner error message.

qa/L0_python_api/test_kserve.py
KrishnanPrash: Initially was `import testing_utils` but changed to `import testing_utils as utils` per @tanmayv25 suggestion. I don't have a strong preference either way.

oandreeva-nv:
lgtm, I would also let @rmccorm4 take a look

qa/L0_python_api/testing_utils.py
```diff
@@ -93,3 +99,77 @@ def send_and_test_inference_identity(frontend_client, url: str) -> bool:
...
+
+def send_and_test_generate_inference() -> bool:
+    model_name = "identity"
+    url = f"http://localhost:8000/v2/models/{model_name}/generate"
+    input_text = "testing"
+    data = {
+        "INPUT0": input_text,
+    }
+
+    response = requests.post(url, json=data, stream=True)
```
rmccorm4, line 167: comment for future: I'm not sure that `stream=True` here is meaningful. Can revisit later though.

rmccorm4 (approved):

rmccorm4:
LGTM - pending passing pipeline with latest changes

src/http_server.cc
GuanLuo, line 3241: Why not put into `StartTrace()`

KrishnanPrash: Will make these changes as well as a part of this refactoring ticket [DLIS-7380].

PR #7682, opened 2024-10-05
fix: Support sampling parameters of type List for vLLM backend (stop words)

rmccorm4:
Adds support and some smoke testing for parameters of type `List` such as `stop: [".", ","]`. This is done by passing sampling_parameters to vLLM backend as serialized JSON string input rather than the previous `TRITONSERVER_Parameter` approach, which didn't support Lists.

Resolves the error:
```
"Unsupported Value. Can\'t convert <class \'list\'> to Request Parameter"
```

### DIFF
--- a/python/openai/openai_frontend/engine/utils/triton.py
+++ b/python/openai/openai_frontend/engine/utils/triton.py
@@ -40,18 +40,28 @@ def _create_vllm_inference_request(
 
     # NOTE: The exclude_none is important, as internals may not support
     # values of NoneType at this time.
-    sampling_parameters = request.model_dump(
+    sampling_parameters = request.model_dump_json(
         exclude=excludes,
         exclude_none=True,
     )
     inputs["text_input"] = [prompt]
-    inputs["stream"] = [request.stream]
+    inputs["stream"] = np.bool_([request.stream])
+    # Pass sampling_parameters as serialized JSON string input to support List
+    # fields like 'stop' that aren't supported by TRITONSERVER_Parameters yet.
+    inputs["sampling_parameters"] = [sampling_parameters]
     exclude_input_in_output = True
     echo = getattr(request, "echo", None)
     if echo:
         exclude_input_in_output = not echo
+
     inputs["exclude_input_in_output"] = [exclude_input_in_output]
-    return model.create_request(inputs=inputs, parameters=sampling_parameters)
+    # TODO: Remove debug print statements
+    print(f"{request.stop=}")
+    print(f"{request.stream=}")
+    print(f"{sampling_parameters=}")
+    print(f"{type(sampling_parameters)=}")
+    print(f"{inputs=}")
+    return model.create_request(inputs=inputs, parameters={})
 
 
 def _create_trtllm_inference_request(
--- a/python/openai/openai_frontend/engine/utils/triton.py
+++ b/python/openai/openai_frontend/engine/utils/triton.py
@@ -44,23 +44,17 @@ def _create_vllm_inference_request(
         exclude=excludes,
         exclude_none=True,
     )
-    inputs["text_input"] = [prompt]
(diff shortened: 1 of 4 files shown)


python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -40,26 +40,30 @@ def _create_vllm_inference_request(
...
-    inputs["exclude_input_in_output"] = [exclude_input_in_output]
-    return model.create_request(inputs=inputs, parameters=sampling_parameters)
+
+    inputs["text_input"] = [prompt]
+    inputs["stream"] = np.bool_([request.stream])
+    inputs["exclude_input_in_output"] = np.bool_([exclude_input_in_output])
+    # Pass sampling_parameters as serialized JSON string input to support List
+    # fields like 'stop' that aren't supported by TRITONSERVER_Parameters yet.
+    inputs["sampling_parameters"] = [sampling_parameters]
+    return model.create_request(inputs=inputs, parameters={})
```
nnshah1, line 58: nit: can remove parameters, keeping empty can be a little confusing at first look

python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -40,26 +40,30 @@ def _create_vllm_inference_request(
...
-    sampling_parameters = request.model_dump(
+    sampling_parameters = request.model_dump_json(
         exclude=excludes,
         exclude_none=True,
     )
-    inputs["text_input"] = [prompt]
-    inputs["stream"] = [request.stream]
     exclude_input_in_output = True
     echo = getattr(request, "echo", None)
     if echo:
```
nnshah1, line 49: I believe this should be:
```
if echo is not None:
```
instead of ```if echo``` - because the next line assumes that echo could be `False`

python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -77,6 +81,8 @@ def _create_trtllm_inference_request(
         inputs["random_seed"] = np.uint64([[request.seed]])
     if request.temperature is not None:
         inputs["temperature"] = np.float32([[request.temperature]])
+    # NOTE: TRT-LLM doesn't currently support runtime changes of 'echo' and it
+    # is configured at model load time, so we don't handle it here for now.
```
nnshah1, line 85: Can we check at load time and error if a request doesn't match the load time parameter?

python/openai/openai_frontend/engine/utils/triton.py
rmccorm4: done: https://github.com/triton-inference-server/server/pull/7682/commits/9f17b62ecec2ba6b2372d1f0252dcd7355075b86

python/openai/openai_frontend/engine/utils/triton.py
rmccorm4: nice catch, done: https://github.com/triton-inference-server/server/pull/7682/commits/9f17b62ecec2ba6b2372d1f0252dcd7355075b86

python/openai/openai_frontend/engine/utils/triton.py
rmccorm4: This one is a bit tedious because `exclude_input_in_output` is a config field of the `tensorrt_llm_bls` model, but the model we'd be interacting with at this point is the BLS or Ensemble model. Would need some global-er assumption about a single TRT-LLM model that we store somewhere, or a way to associate with TRT-LLM model we're interacting with through the BLS/Ensemble model.

Will think about it a bit more tomorrow - but would rather exclude that change from this PR. I'll also look for an update on the ticket for TRT-LLM adding runtime support for this flag to see when it may be added.

kthui (approved):
LGTM! In summary, we want to pass the sampling parameters as JSON in the inputs, instead of a separate model request parameters which does not support list type.

PR #7691, opened 2024-10-09
docs: Add example outputs to OpenAI Frontend docs

KrishnanPrash:
#### What does the PR do?
Adding outputs and links to related resources in the `server/python/openAI/README.md` docs.


rmccorm4:
Please update target branch to `main` and fix any merge conflicts

python/openai/README.md
```diff
@@ -135,15 +195,14 @@ print(completion.choices[0].message.content)
 8. Run tests (NOTE: The server should not be running, the tests will handle starting/stopping the server as necessary):
 ```bash
 cd server/python/openai/
-pip install -r requirements-test.txt
-
```
rmccorm4, line 139: keep

python/openai/README.md
rmccorm4, line 1: Let's simplify the genai-perf example too by removing the i/o flags:
```bash
genai-perf \
  --model ${MODEL} \
  --tokenizer ${TOKENIZER} \
  --service-kind openai \
  --endpoint-type chat \
  --streaming
```

python/openai/README.md
```diff
@@ -70,6 +70,22 @@ pip install -r requirements.txt
 # NOTE: Adjust the --tokenizer based on the model being used
 python3 openai_frontend/main.py --model-repository tests/vllm_models --tokenizer meta-llama/Meta-Llama-3.1-8B-Instruct
 ```
+The output should contain the following snippets look along the lines of:
```
rmccorm4, line 73:
```suggestion
Once the server has successfully started, you should see something like this:
```

python/openai/README.md
rmccorm4, line 1: In TRT-LLM section, above `MODEL="tensorrt_llm_bls"`, let's add a note that `ensemble` could be used too if setup:
```bash
# "ensemble" can also be used here instead of "tensorrt_llm_bls" if setup
MODEL="tensorrt_llm_bls"
...
```

python/openai/README.md
```diff
@@ -171,20 +251,45 @@ pip install -r requirements.txt
...
 ```bash
 # NOTE: Adjust the --tokenizer based on the model being used
-python3 openai_frontend/main.py --model-repository tests/tensorrtllm_models --tokenizer meta-llama/Meta-Llama-3.1-8B-Instruct
+python3 openai_frontend/main.py --model-repository path/to/models --tokenizer meta-llama/Meta-Llama-3.1-8B-Instruct
 ```
 
 3. Send a `/v1/chat/completions` request:
   - Note the use of `jq` is optional, but provides a nicely formatted output for JSON responses.
 ```bash
+# "ensemble" can also be used here instead of "tensorrt_llm_bls" if setup
```
rmccorm4, line 260: I think the fact that Triton CLI will generate a model whose name is neither `tensorrt_llm_bls` or `ensemble` will be confusing. So I tried to stress the logic behind the variable name choice here. Let me know what you think.
```suggestion
# MODEL should be the client-facing model name in your model repository for a pipeline like TRT-LLM.
# For example, this could also be "ensemble", or something like "gpt2" if generated from Triton CLI
```

rmccorm4: PR LGTM otherwise

KrishnanPrash: I this change is good because it provides more context and reduces the chance of confusion for someone going through the motions.

rmccorm4 (approved):

PR #7694, opened 2024-10-12
test: Tests new histogram metric

yinggeh:
#### What does the PR do?
The PR adds tests to histogram metrics and new `nv_inference_first_response_histogram`.
1. Verify this metric is only created in decoupled models.
2. Tests `--metrics-config histogram_latencies=<bool>`.

#### Checklist

#### Commit Type:
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/core/pull/396

Check the implementation PR first.

#### Test plan:
L0_metrics--base

- CI Pipeline ID:
19270121

#### Background
Standardizing Large Model Server Metrics in Kubernetes

### DIFF
--- a/docs/user_guide/metrics.md
+++ b/docs/user_guide/metrics.md
@@ -204,6 +204,42 @@ metrics are used for latencies:
 
 To disable these metrics specifically, you can set `--metrics-config counter_latencies=false`
 
+#### Histograms
+
+> **Note**
+>
+> The following Histogram feature is experimental for the time being and may be
+> subject to change based on user feedback.
+
+By default, the following
+[Histogram](https://prometheus.io/docs/concepts/metric_types/#histogram)
+metrics are used for latencies:
+
+|Category      |Metric          |Metric Name |Description                |Granularity|Frequency    |Model Type
+|--------------|----------------|------------|---------------------------|-----------|-------------|-------------|
+|Latency       |Response Time    |`nv_inference_first_response_histogram_ms` |Histogram of end-to-end inference request to response time |Per model  |Per request  | Decoupled |
+
+To disable these metrics specifically, you can set `--metrics-config histogram_latencies=false`
+
+Each histogram above may composed of several sub-metrics. For each
+metric, there is a set of `le` metrics tracking the counter for each
+bucket. Additionally, there are `_count` and `_sum` metrics that aggregate
+the count and observed values for each. For example, see the following
+information exposed by the Inference First Response Histogram metrics:
+```
+# HELP nv_first_response_histogram_ms Histogram of end-to-end inference request to response time in milliseconds.
+# TYPE nv_first_response_histogram_ms histogram
+nv_inference_first_response_histogram_ms_count{model="my_model",version="1"} 101
+nv_inference_first_response_histogram_ms_sum{model="my_model",version="1"} 3685
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="10"} 55
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="100"} 97
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="500"} 98
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="1000"} 101
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="+Inf"} 101
+```
+
+Triton has a set of default buckets per metric to track, as shown above. Customization of buckets per metric is  currently not supported.
+
 #### Summaries
 
 > **Note**
--- a/qa/L0_metrics/metrics_config_test.py
+++ b/qa/L0_metrics/metrics_config_test.py
@@ -44,6 +44,7 @@ INF_COUNTER_PATTERNS = [
     "nv_inference_compute_infer_duration",
     "nv_inference_compute_output_duration",
 ]
+INF_HISTOGRAM_DECOUPLED_PATTERNS = ["nv_inference_first_response_histogram"]
 INF_SUMMARY_PATTERNS = [
     "nv_inference_request_summary",
     "nv_inference_queue_summary",
@@ -95,6 +96,17 @@ class MetricsConfigTest(tu.TestResultCollector):
     def test_cache_counters_missing(self):
         metrics = self._get_metrics()
         for metric in CACHE_COUNTER_PATTERNS:
+            self.assertNotIn(metric, metrics)
+
+    # Histograms
+    def test_inf_histograms_decoupled_exist(self):
+        metrics = self._get_metrics()
+        for metric in INF_HISTOGRAM_DECOUPLED_PATTERNS:
+            self.assertIn(metric, metrics)
+
+    def test_inf_histograms_decoupled_missing(self):
+        metrics = self._get_metrics()
+        for metric in INF_HISTOGRAM_DECOUPLED_PATTERNS:
             self.assertNotIn(metric, metrics)
 
     # Summaries
--- a/qa/L0_metrics/test.sh
+++ b/qa/L0_metrics/test.sh
@@ -283,6 +283,41 @@ python3 ${PYTHON_TEST} MetricsConfigTest.test_cache_summaries_missing 2>&1 | tee
 check_unit_test
 kill_server
 
+# Enable histograms (decoupled)
+decoupled_model_name="dlpack_io_identity_decoupled"
+mkdir -p "${MODELDIR}/${decoupled_model_name}/1/"
+cp ../python_models/${decoupled_model_name}/model.py ${MODELDIR}/${decoupled_model_name}/1/
+cp ../python_models/${decoupled_model_name}/config.pbtxt ${MODELDIR}/${decoupled_model_name}/
+
+SERVER_ARGS="${BASE_SERVER_ARGS} --load-model=${decoupled_model_name}"
+run_and_check_server
+python3 ${PYTHON_TEST} MetricsConfigTest.test_inf_counters_exist 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_inf_histograms_decoupled_exist 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_inf_summaries_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_cache_counters_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_cache_summaries_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+kill_server
+
+# Disable histograms (decoupled)
+SERVER_ARGS="${BASE_SERVER_ARGS} --load-model=${decoupled_model_name} --metrics-config histogram_latencies=false"
+run_and_check_server
+python3 ${PYTHON_TEST} MetricsConfigTest.test_inf_counters_exist 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_inf_histograms_decoupled_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_inf_summaries_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_cache_counters_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+python3 ${PYTHON_TEST} MetricsConfigTest.test_cache_summaries_missing 2>&1 | tee ${CLIENT_LOG}
+check_unit_test
+kill_server
+
 # Enable summaries, counters still enabled by default
 SERVER_ARGS="${BASE_SERVER_ARGS} --load-model=identity_cache_off --metrics-config summary_latencies=true"
 run_and_check_server


docs/user_guide/metrics.md
```diff
@@ -204,6 +204,42 @@ metrics are used for latencies:
...
+nv_inference_first_response_histogram_ms_count{model="my_model",version="1"} 101
+nv_inference_first_response_histogram_ms_sum{model="my_model",version="1"} 3685
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="10"} 55
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="100"} 97
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="500"} 98
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="1000"} 101
+nv_inference_first_response_histogram_ms{model="my_model",version="1", le="+Inf"} 101
+```
+
+Triton has a set of default buckets per metric to track, as shown above. Customization of buckets per metric is  currently not supported.
```
kthui, line 241:
```suggestion
Triton has a set of default buckets per metric to track, as shown above. Customization of buckets per metric is currently not supported.
```

kthui (approved):

qa/L0_metrics/metrics_config_test.py
```diff
@@ -97,6 +98,17 @@ def test_cache_counters_missing(self):
         for metric in CACHE_COUNTER_PATTERNS:
             self.assertNotIn(metric, metrics)
 
+    # Histograms
+    def test_inf_histograms_decoupled_exist(self):
+        metrics = self._get_metrics()
+        for metric in INF_HISTOGRAM_DECOUPLED_PATTERNS:
+            self.assertIn(metric, metrics)
+
+    def test_inf_histograms_decoupled_missing(self):
```
rmccorm4, line 107:
```suggestion
    # "Time to First Response" metric currently only exists for decoupled models
    # so verify that it is not reported for a non-decoupled model.
    def test_inf_histograms_nondecoupled_missing(self):
```

qa/L0_metrics/metrics_config_test.py
```diff
@@ -97,6 +98,17 @@ def test_cache_counters_missing(self):
         for metric in CACHE_COUNTER_PATTERNS:
             self.assertNotIn(metric, metrics)
 
+    # Histograms
+    def test_inf_histograms_decoupled_exist(self):
+        metrics = self._get_metrics()
+        for metric in INF_HISTOGRAM_DECOUPLED_PATTERNS:
+            self.assertIn(metric, metrics)
```
rmccorm4, line 105: Since we saw earlier tonight that a recent change might've broken the metric and the buckets weren't actually getting updated - please add a test that verifies the buckets get updated after an inference so we can be confident it is still working when making changes

qa/L0_metrics/metrics_config_test.py
yinggeh: histograms_decoupled means the set of histograms existing in decoupled models. It's missing in either non-decoupled model or in decoupled model but `--metrics-config histogram_latencies=true` is not set.

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+        decoupled_model_sum = get_histogram_metric_key(
+            first_response_family, decoupled_model_name, "1", "sum"
+        )
+        self.assertIn(decoupled_model_count, histogram_dict)
+        self.assertGreaterEqual(histogram_dict[decoupled_model_count], 1)
+        self.assertIn(decoupled_model_sum, histogram_dict)
+        self.assertGreaterEqual(
+            histogram_dict[decoupled_model_sum], wait_secs * MILLIS_PER_SEC
+        )
```
kthui, line 167: Can you run the inference multiple times to show the `count` and `sum` are correctly aggregated, across multiple requests?

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+            inputs=inputs,
+            outputs=outputs,
+        )
+        triton_client.stop_stream()
+
+        # Wait until the results are available in user_data
+        time_out = 10
+        while (len(user_data) == 0) and time_out > 0:
+            time_out = time_out - 1
+            time.sleep(1)
```
kthui, line 112: Are we only expecting one response from the request?

If I understand correctly, the ensemble is to simulate a LLM with a 
postprocessing step. The LLM is expected to return more than one word under most if not all circumstances. Can you update the decoupled model to have it returns multiple responses? And verify the `count` and `sum` are correctly aggregated with respect to the responses per request?

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # Display and validate the available results
+        if len(user_data) == 1:
+            # Check for the errors
+            if type(user_data[0]) == InferenceServerException:
+                print(user_data[0])
+                sys.exit(1)
```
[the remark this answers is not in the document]
yinggeh: This doesn't look correct to me. `user_data[0]` shouldn't be `InferenceServerException`.

qa/L0_metrics/histogram_metrics_test.py
yinggeh: This tests aim to verify not only the ensemble model but also each decoupled sub-model is updated per request. Only the time to first response is observed to histogram so I picked one response per request.

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,168 @@
...
+            key, value = match.rsplit(" ")
+            histogram_dict[key] = int(value)
+
+        return histogram_dict
+
+    def async_stream_infer(self, model_name, inputs, outputs):
+        try:
+            triton_client = grpcclient.InferenceServerClient(url="localhost:8001")
+        except Exception as e:
+            print("context creation failed: " + str(e))
```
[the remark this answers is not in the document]
yinggeh: I checked other tests and didn't see checks for triton_client. I will remove try catch block.

qa/L0_metrics/ensemble_decoupled/ensemble/config.pbtxt
```diff
@@ -0,0 +1,72 @@
...
+    key: "WAIT_SECONDS"
+    value: "temp_output"
+   }
+   output_map {
+    key: "DUMMY_OUT"
+    value: "OUTPUT"
+   }
+  }
+ ]
+}
```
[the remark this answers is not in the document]
yinggeh: Updated

qa/L0_metrics/test.sh
```diff
@@ -406,6 +458,20 @@ kill_server
 expected_tests=6
 check_unit_test "${expected_tests}"
 
+# Test histogram data in ensemble decoupled model
+MODELDIR="${PWD}/ensemble_decoupled"
+SERVER_ARGS="--model-repository=${MODELDIR} --metrics-config histogram_latencies=true --log-verbose=1"
+PYTHON_TEST="histogram_metrics_test.py"
+rm -rf "${MODELDIR}"/ensemble/1
+mkdir "${MODELDIR}"/ensemble/1
```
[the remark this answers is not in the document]
yinggeh: Updated

qa/L0_metrics/histogram_metrics_test.py
yinggeh: Updated

qa/L0_metrics/histogram_metrics_test.py
yinggeh: Also the models in this ensemble model have input and output both with shape [1]. The output is a single value which feeds to the next model input as the wait seconds in `execute` function, which is used to test `sum` value.

qa/L0_metrics/histogram_metrics_test.py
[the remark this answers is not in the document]
kthui: Could use the `assertNotIsInstance()`?

What I meant here is calling `sys.exit(1)` should be avoided on a Python unit test, and use `assert...()` whenever possible.

qa/L0_metrics/histogram_metrics_test.py
kthui: > Only the time to first response is observed to histogram so I picked one response per request.

Can you verify the time to first response is not updated if a second response is returned?

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -133,38 +129,42 @@ def test_ensemble_decoupled(self):
...
+            )
+
+            decoupled_model_count = get_histogram_metric_key(
+                first_response_family, decoupled_model_name, "1", "count"
+            )
+            decoupled_model_sum = get_histogram_metric_key(
+                first_response_family, decoupled_model_name, "1", "sum"
+            )
+            self.assertIn(decoupled_model_count, histogram_dict)
+            self.assertGreaterEqual(histogram_dict[decoupled_model_count], request_num)
```
kthui, line 162: Can we use `assertEqual()`?

yinggeh: Should be assertEqual. Nice catch.

qa/L0_metrics/histogram_metrics_test.py
yinggeh: It will take a while to update the models. Expect update on Monday if there's no blocker.

qa/L0_metrics/histogram_metrics_test.py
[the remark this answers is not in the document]
yinggeh: Agree. Updated.

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
```diff
@@ -0,0 +1,74 @@
...
+        for request in requests:
+            # Get input data
+            wait_secs_tensor = pb_utils.get_input_tensor_by_name(
+                request, "WAIT_SECONDS"
+            ).as_numpy()
+
+            # Validate input data
+            wait_secs = wait_secs_tensor[0]
+            if wait_secs < 0:
+                self.raise_value_error(request, "wait_secs cannot be negative")
```
kthui, line 46: This is for testing Python backend error handling on the original code, you can remove it.

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
```diff
@@ -0,0 +1,74 @@
...
+            processed_requests.append(
+                {
+                    "response_sender": request.get_response_sender(),
+                }
+            )
+
+        # This decoupled execute should be scheduled to run in the background
+        # concurrently with other instances of decoupled execute, as long as the event
+        # loop is not blocked.
+        await asyncio.gather(*async_tasks)
```
kthui, line 57: This is intended to test dynamic batching and async task concurrency, you can remove it. Simply call `time.sleep(duration)`.

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
```diff
@@ -0,0 +1,74 @@
...
+
+        for p_req in processed_requests:
```
kthui, line 59: This is needed for async.gather to work, you can send the response(s) directly.

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
```diff
@@ -0,0 +1,74 @@
...
+                response, flags=pb_utils.TRITONSERVER_RESPONSE_COMPLETE_FINAL
+            )
+
+        return None
+
+    def raise_value_error(self, requests, msg):
+        for request in requests:
+            response_sender = request.get_response_sender()
+            response_sender.send(flags=pb_utils.TRITONSERVER_RESPONSE_COMPLETE_FINAL)
+        raise ValueError(msg)
```
kthui, line 74: Can be deleted, it is for testing Python backend exception handling.

qa/L0_metrics/ensemble_decoupled/async_execute_decouple/1/model.py
```diff
@@ -0,0 +1,91 @@
...
+        processed_requests = []
+        async_tasks = []
+        for request in requests:
+            # Get input data
+            wait_secs_tensor = pb_utils.get_input_tensor_by_name(
+                request, "WAIT_SECONDS"
+            ).as_numpy()
+            response_num_tensor = pb_utils.get_input_tensor_by_name(
+                request, "RESPONSE_NUM"
+            ).as_numpy()
```
kthui, line 44: I think you can get away with keeping this decoupled model. The non-decoupled model is a subset of this one by setting `decoupled = False` on the model config and asserting the `RESPONSE_NUM = 1`.

kthui (commented):
Make sure to rebase on main that has the compute capability fix merged.

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,189 @@
...
+            decoupled_model_count = get_histogram_metric_key(
+                first_response_family, decoupled_model_name, "1", "count"
+            )
+            decoupled_model_sum = get_histogram_metric_key(
+                first_response_family, decoupled_model_name, "1", "sum"
+            )
+            self.assertIn(decoupled_model_count, histogram_dict)
+            self.assertEqual(histogram_dict[decoupled_model_count], request_num)
+            self.assertIn(decoupled_model_sum, histogram_dict)
+            self.assertGreaterEqual(
```
kthui, line 172: Can you also assert the number is within a range?

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
yinggeh: Updated

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
yinggeh: Updated

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
yinggeh: Updated

qa/L0_metrics/ensemble_decoupled/async_execute/1/model.py
yinggeh: Updated

qa/L0_metrics/ensemble_decoupled/async_execute_decouple/1/model.py
yinggeh: Updated

qa/L0_metrics/histogram_metrics_test.py
yinggeh: Updated

kthui (commented):
Nice work adding a more comprehensive test!

Only two things left:
1. Verify the number of response(s) returned on each inference matches the number expected.
2. Verify the histogram bucket key and value pairs after each inference.

qa/L0_metrics/ensemble_decoupled/async_execute_decouple/1/model.py
```diff
@@ -0,0 +1,59 @@
...
+        response_num = pb_utils.get_input_tensor_by_name(
+            request, "RESPONSE_NUM"
+        ).as_numpy()[0]
+        output_tensors = [
+            pb_utils.Tensor("WAIT_SECONDS", np.array([wait_secs], np.float32)),
+            pb_utils.Tensor("RESPONSE_NUM", np.array([1], np.uint8)),
+        ]
+
+        # Wait
+        time.sleep(wait_secs.item())
```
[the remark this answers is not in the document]
yinggeh: I had to call .item() because it couldn't cast into python float type. See error below.
`time.sleep(wait_secs)\nTypeError: 'numpy.float32' object cannot be interpreted as an integer\n`

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,191 @@
...
+MILLIS_PER_SEC = 1000
+
+
+def get_histogram_metric_key(
+    metric_family, model_name, model_version, metric_type, le=""
+):
+    if metric_type in ["count", "sum"]:
+        return f'{metric_family}_{metric_type}{{model="{model_name}",version="{model_version}"}}'
+    elif metric_type == "bucket":
+        return f'{metric_family}_{metric_type}{{model="{model_name}",version="{model_version}",le="{le}"}}'
```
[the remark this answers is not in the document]
yinggeh: Updated

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,191 @@
...
+                first_response_family, decoupled_model_name, "1", "sum"
+            )
+            self.assertIn(decoupled_model_count, histogram_dict)
+            self.assertEqual(histogram_dict[decoupled_model_count], request_num)
+            self.assertIn(decoupled_model_sum, histogram_dict)
+            self.assertTrue(
+                wait_secs * MILLIS_PER_SEC * request_num
+                <= histogram_dict[decoupled_model_sum]
+                < (wait_secs + delta) * MILLIS_PER_SEC * request_num
+            )
```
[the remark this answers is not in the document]
yinggeh: Updated

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,191 @@
...
+                model_name=model_name,
+                inputs=inputs,
+                outputs=outputs,
+            )
+
+        # Wait until the results are available in user_data
+        time_out = 10
+        while (len(user_data) == 0) and time_out > 0:
+            time_out = time_out - 1
+            time.sleep(1)
```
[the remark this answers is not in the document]
yinggeh: Updated

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -0,0 +1,179 @@
...
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import os
+import re
+import sys
+import time
```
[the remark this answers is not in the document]
yinggeh: removed

qa/L0_metrics/metrics_config_test.py
```diff
@@ -97,6 +98,18 @@ def test_cache_counters_missing(self):
         for metric in CACHE_COUNTER_PATTERNS:
             self.assertNotIn(metric, metrics)
 
+    # Histograms
+    def test_inf_histograms_decoupled_exist(self):
+        metrics = self._get_metrics()
+        for metric in INF_HISTOGRAM_DECOUPLED_PATTERNS:
+            for suffix in ["_count", "_sum", ""]:
+                self.assertIn(metric + suffix, metrics)
```
kthui, line 106: Nice work adding checks that the count, sum and bucket metrics exist!

This does not check if the value on each bucket is correct, but you mentioned there is an existing test that will verify the value is correct, so you are not verifying the value here because the existing test will verify the Prometheus histogram metrics is functioning correctly and the tests on histogram_metrics_test.py is verifying the numbers provided to Prometheus is correct via count and sum.

qa/L0_metrics/histogram_metrics_test.py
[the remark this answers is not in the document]
kthui: https://github.com/triton-inference-server/server/pull/7694#discussion_r1811117321

kthui (dismissed):

qa/L0_metrics/metrics_config_test.py
yinggeh: Since this test only checks if specific histograms exist without updating the metrics, all values should be 0. For tests regarding Prometheus metrics APIs and functionalities, please refer to 
https://github.com/triton-inference-server/core/blob/main/src/test/metrics_api_test.cc

rmccorm4:
```
# Prometheus histogram buckets are tested in metrics_api_test.cc::HistogramAPIHelper
```

Just a nit for clarification: The Metrics API tests (ex: `TRITONSERVER_MetricObserve`) aren't going through the same code path that the built-in latency metrics are going through. So although there are custom metrics tests testing histograms, they aren't necessarily testing the built-in histogram latency metrics added here.

The built-in metrics actually just use prometheus APIs directly and some C++ helper functions around them today. Ideally, we would unify these to all use the `TRITONSERVER_Metric` layer in the same way for both built-in metrics and custom metrics for easier test coverage and maintenance in the future.

---

All that being said, I think the current tests included here with the checks around the `_sum` value are fine for now for this PR. When adding other histogram latency metrics, I think it would be a good idea to either (a) add some bucket-related tests to these python unit tests or (b) attempt to unify the internal metrics to use the same custom metrics APIs or internal C++ classes around them for better re-use of the `metrics_api_test.cc` tests.

docs/user_guide/metrics.md
```diff
@@ -204,6 +204,42 @@ metrics are used for latencies:
...
+|--------------|----------------|------------|---------------------------|-----------|-------------|-------------|
+|Latency       |Request to First Response Time    |`nv_inference_first_response_histogram_ms` |Histogram of end-to-end inference request to the first response time |Per model  |Per request  | Decoupled |
+
+To enable these metrics specifically, you can set `--metrics-config histogram_latencies=true`
+
+Each histogram above may composed of several sub-metrics. For each
+metric, there is a set of `le` metrics tracking the counter for each
+bucket. Additionally, there are `_count` and `_sum` metrics that aggregate
+the count and observed values for each. For example, see the following
+information exposed by the Request to First Response Histogram metrics:
```
rmccorm4, line 228:
```suggestion
Each histogram above is composed of several sub-metrics. For each
histogram metric, there is a set of `le` (less than or equal to) thresholds tracking the counter for each
bucket. Additionally, there are `_count` and `_sum` metrics that aggregate
the count and observed values for each. For example, see the following
information exposed by the "Time to First Response" Histogram metrics:
```

qa/L0_metrics/test.sh
```diff
@@ -270,11 +270,63 @@ mkdir -p "${MODELDIR}/identity_cache_on/1"
...
+check_unit_test
+kill_server
+
+# Check default settings: Histograms should be disabled in decoupled model
+decoupled_model_name="async_execute_decouple"
+mkdir -p "${MODELDIR}/${decoupled_model_name}/1/"
+cp ../python_models/${decoupled_model_name}/model.py ${MODELDIR}/${decoupled_model_name}/1/
+cp ../python_models/${decoupled_model_name}/config.pbtxt ${MODELDIR}/${decoupled_model_name}/
+
+SERVER_ARGS="${BASE_SERVER_ARGS} --load-model=${decoupled_model_name}"
```
rmccorm4, line 309: nit for future follow-up (not this PR), a lot of this could probably be condensed with a for loop.

rmccorm4 (commented):
Nice work! Only a minor comment on `metrics.md` for this PR, but LGTM otherwise.

docs/user_guide/metrics.md
yinggeh: This is better. Updated the doc.

rmccorm4 (approved):

PR #7700, opened 2024-10-14
fix: Fix bug when targeting the TRT-LLM backend ensemble

blongnv:
#### What does the PR do?
If a user follows the example code and tries to use the `ensemble` instead of `tensorrt_llm_bls`, they will get this error:

```
curl -s http://localhost:9000/v1/chat/completions -H 'Content-Type: application/json' -d '{"model": "ensemble", "messages": [{"role": "user", "content": "Say this is a test!"}]}' | jq
INFO:     127.0.0.1:40064 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
{
  "detail": "Unknown backend"
}
```

After the PR fix, the response will be correct:
```
curl -s http://localhost:9000/v1/chat/completions -H 'Content-Type: application/json' -d '{"model": "ensemble", "messages": [{"role": "user", "content": "Say this is a test!"}]}' | jq
INFO:     127.0.0.1:47946 - "POST /v1/chat/completions HTTP/1.1" 200 OK
{
  "id": "cmpl-fefc9e2a-8a74-11ef-8e2a-0242c0a80102",
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "content": "This is a test!",
        "tool_calls": null,
        "role": "assistant",
        "function_call": null
      },
      "logprobs": null
    }
  ],
  "created": 1728942057,
  "model": "ensemble",
  "system_fingerprint": null,
  "object": "chat.completion",
  "usage": null
}
```

#### Checklist
- [ ] I ran pre-commit locally (`pre-commit install, pre-commit run --all`)

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

- closes GitHub issue: #xxx

### DIFF
--- a/python/openai/openai_frontend/engine/triton_engine.py
+++ b/python/openai/openai_frontend/engine/triton_engine.py
@@ -250,9 +250,9 @@ class TritonLLMEngine(LLMEngine):
         for name, _ in self.server.models().keys():
             model = self.server.model(name)
             backend = model.config()["backend"]
-            if not backend:
-                # Check platform field as a backup, this will support 'ensemble' models
-                backend = model.config()["platform"]
+            # Explicitly handle ensembles to avoid any runtime validation errors
+            if not backend and model.config()["platform"] == "ensemble":
+                backend = "ensemble"
             print(f"Found model: {name=}, {backend=}")
 
             metadata = TritonModelMetadata(
--- a/python/openai/openai_frontend/engine/triton_engine.py
+++ b/python/openai/openai_frontend/engine/triton_engine.py
@@ -250,6 +250,9 @@ class TritonLLMEngine(LLMEngine):
         for name, _ in self.server.models().keys():
             model = self.server.model(name)
             backend = model.config()["backend"]
+            if backend == "":
+                # look for ensemble setting if backend is undefined
+                backend = model.config()["platform"]
             print(f"Found model: {name=}, {backend=}")
 
             metadata = TritonModelMetadata(
--- a/python/openai/openai_frontend/engine/triton_engine.py
+++ b/python/openai/openai_frontend/engine/triton_engine.py
@@ -250,8 +250,8 @@ class TritonLLMEngine(LLMEngine):
         for name, _ in self.server.models().keys():
             model = self.server.model(name)
             backend = model.config()["backend"]
-            if backend == "":
-                # look for ensemble setting if backend is undefined
+            if not backend:
+                # Check platform field as a backup, this will support 'ensemble' models
                 backend = model.config()["platform"]
             print(f"Found model: {name=}, {backend=}")
 


python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -250,6 +250,9 @@ def _get_model_metadata(self) -> Dict[str, TritonModelMetadata]:
         for name, _ in self.server.models().keys():
             model = self.server.model(name)
             backend = model.config()["backend"]
+            if backend == "":
+                # look for ensemble setting if backend is undefined
```
rmccorm4, line 254:
```suggestion
            if not backend:
                # Check platform field as a backup, this will support 'ensemble' models
```

rmccorm4:
ref: DLIS-7451 to follow-up with a test for `ensemble` path to maintain support and catch regressions

python/openai/openai_frontend/engine/triton_engine.py
blongnv: This actually does not work. It returns an empty string, not None

rmccorm4:
Pipeline: 19345468 ✅

rmccorm4 (dismissed):

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -250,6 +250,9 @@ def _get_model_metadata(self) -> Dict[str, TritonModelMetadata]:
         for name, _ in self.server.models().keys():
             model = self.server.model(name)
             backend = model.config()["backend"]
+            if not backend:
+                # Check platform field as a backup, this will support 'ensemble' models
+                backend = model.config()["platform"]
```
nnshah1, line 255: @rmccorm4  should we check against a list of  supported platform types here to rule out any accidents

rmccorm4: Sure, we can be strict here just in case. 

There's currently no use case I'm aware of where the platform would be used to specify a backend (ex: `platform: tensorrt_plan`) that doesn't have a backend equivalent, and we generally encourage use of backend for this purpose for pretty much everything other than ensembles - so I think it's fine to be strict for now.

rmccorm4: @blongnv could you modify it to be something like this?
```python
# Explicitly handle ensembles to avoid any runtime validation errors
if not backend and model.config()["platform"] == "ensemble":
  backend = "ensemble"
```

blongnv: Implemented, please check

rmccorm4 (approved):

PR #7701, opened 2024-10-15
test: Allow ensemble to create the final response even if some of the outputs are not created

kthui:
#### What does the PR do?
Add test when the composing models of an ensemble do not provide all outputs on the model configuration. The inference should complete normally and return all outputs that are provided.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:
https://github.com/triton-inference-server/core/pull/399

N/A

#### Test plan:
N/A

- CI Pipeline ID: 19348789

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/qa/L0_simple_ensemble/ensemble_test.py
+++ b/qa/L0_simple_ensemble/ensemble_test.py
@@ -26,6 +26,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import os
 import random
 import sys
 import time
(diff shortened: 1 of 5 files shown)


GuanLuo (approved):

rmccorm4 (approved):

qa/L0_simple_ensemble/ensemble_test.py
```diff
@@ -177,6 +178,41 @@ def callback(start_time, result, error):
         if expected_flags != response_flags:
             self.assertTrue(False, "unexpeced sequence flags mismatch error")
 
+    def test_ensemble_partial_add_sub(self):
```
rmccorm4, line 181: I think we should have more testing in general to programatically track expected behavior of certain edge cases, such as:
1. What if a composing model doesn't produce the outputs necessary to feed inputs of the next step?
2. In an ensemble with outputs from 2 parallel steps, what if 1 step has outputs and the other step has no outputs?
3. What if an ensemble doesn't produce any of its outputs?

These aren't immediately clear to me from scanning the code what would happen, so I think it would be good to enforce expected behavior via testing, and be able to catch if the behavior subtlely changes from future ensemble code changes

rmccorm4: @kthui

PR #7715, opened 2024-10-17
docs: Add support matrix for model parallelism in OpenAI Frontend

rmccorm4:
Add support matrix (and known limitations) around multi-gpu models for vLLM/TRTLLM

### DIFF
--- a/python/openai/README.md
+++ b/python/openai/README.md
@@ -315,3 +315,13 @@ available arguments and default values.
 
 For more information on the `tritonfrontend` python bindings, see the docs
 [here](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/tritonfrontend.md).
+
+## Known Limitations
+
+- For multi-GPU models (ex: Tensor Parallelism > 1), here is the current support matrix:
+  - [x] vLLM ([EngineArgs](https://github.com/triton-inference-server/vllm_backend/blob/main/README.md#using-the-vllm-backend))
+  - [x] TensorRT-LLM ([Orchestrator Mode](https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#orchestrator-mode))
+    - `export TRTLLM_ORCHESTRATOR=1`
+    - `NOTE`: Orchestrator mode does not currently support multi-node models
+  - [ ] TensorRT-LLM ([Leader Mode](https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#leader-mode))
+    - Coming soon
--- a/python/openai/README.md
+++ b/python/openai/README.md
@@ -316,7 +316,7 @@ available arguments and default values.
 For more information on the `tritonfrontend` python bindings, see the docs
 [here](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/tritonfrontend.md).
 
-## Known Limitations
+## Model Parallelism Support
 
 - For multi-GPU models (ex: Tensor Parallelism > 1), here is the current support matrix:
   - [x] vLLM ([EngineArgs](https://github.com/triton-inference-server/vllm_backend/blob/main/README.md#using-the-vllm-backend))
--- a/python/openai/README.md
+++ b/python/openai/README.md
@@ -73,6 +73,7 @@ python3 openai_frontend/main.py --model-repository tests/vllm_models --tokenizer
 
 <details>
 <summary>Example output</summary>
+
 ```
 ...
 +-----------------------+---------+--------+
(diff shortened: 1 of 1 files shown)


kthui (approved):

python/openai/README.md
```diff
@@ -315,3 +315,15 @@ available arguments and default values.
 
 For more information on the `tritonfrontend` python bindings, see the docs
 [here](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/tritonfrontend.md).
+
+## Known Limitations
+
+- For multi-GPU models (ex: Tensor Parallelism > 1), here is the current support matrix:
```
nnshah1, line 321: are there similar restrictions for PP?

python/openai/README.md
```diff
@@ -315,3 +315,15 @@ available arguments and default values.
 
 For more information on the `tritonfrontend` python bindings, see the docs
 [here](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/tritonfrontend.md).
+
+## Known Limitations
```
nnshah1, line 319:
```suggestion
## Model Parallelism Support
```

python/openai/README.md
```diff
@@ -315,3 +315,15 @@ available arguments and default values.
...
+
+- For multi-GPU models (ex: Tensor Parallelism > 1), here is the current support matrix:
+  - [x] vLLM ([EngineArgs](https://github.com/triton-inference-server/vllm_backend/blob/main/README.md#using-the-vllm-backend))
+    - ex: Configure `tensor_parallel_size: 2` in the
+      [model.json](https://github.com/triton-inference-server/vllm_backend/blob/main/samples/model_repository/vllm_model/1/model.json)
+  - [x] TensorRT-LLM ([Orchestrator Mode](https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#orchestrator-mode))
+    - `export TRTLLM_ORCHESTRATOR=1`
+    - **NOTE**: Orchestrator mode does not currently support multi-node models
```
nnshah1, line 327: I think we would need to define this - I think people will get confused with multi-node and multi-gpu ....

Do we need to mention multi-gpu ?

python/openai/README.md
```diff
@@ -315,3 +315,15 @@ available arguments and default values.
...
+  - [ ] TensorRT-LLM ([Leader Mode](https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#leader-mode))
```
nnshah1, line 328: this is a mode and not a model parellism thing .... not sure best way - maybe a section on model parallelism and a seperate known limitations or a subsection in

python/openai/README.md
rmccorm4: I wanted to be clear about the limitations of orchestrator mode, but it's mentioned in the linked orchestrator docs, so I think we could remove it if you're concerned about mentioning it:
> Since this mode uses [MPI_Comm_spawn](https://www.open-mpi.org/doc/v4.1/man3/MPI_Comm_spawn.3.php), it might not work properly with [slurm](https://slurm.schedmd.com/) deployments. Additionally, this currently only works for single node deployments.

python/openai/README.md
rmccorm4: These are currently the different ways you can do model parallelism with Triton's LLM backends and what you need to be aware of when thinking about how you're going to achieve it with Triton. 

I think it's relatively clear in this current structure as an eye chart, with links to the details of each (vllm backend documentation on parallelism is currently a bit light), and adds some discover-ability into the methods.

I don't think the OpenAI README needs to go into significantly deeper depth on the concepts other than linking to the relevant sections - but let me know if you want to see something specific added

python/openai/README.md
rmccorm4: The explanations here should hold for PP as well - but it seems that everyone primarily talks about TP these days.

python/openai/README.md
```diff
@@ -315,3 +315,15 @@ available arguments and default values.
 
 For more information on the `tritonfrontend` python bindings, see the docs
 [here](https://github.com/triton-inference-server/server/blob/main/docs/customization_guide/tritonfrontend.md).
+
+## Model Parallelism Support
+
+- For multi-GPU models (ex: Tensor Parallelism > 1), here is the current support matrix:
```
nnshah1, line 321:
```suggestion
```

I think with the heading you can remove the sentence -

python/openai/README.md
nnshah1: As currently written - I'd leave it out of here - I think the relationship between multi-node and multi-gpu would need more clarification to be useful - I would lean on the orchestrator mode docs.

python/openai/README.md
nnshah1: I've got some thoughts but would be bigger than this change - so will take them offline.

python/openai/README.md
rmccorm4: done

python/openai/README.md
rmccorm4: Removed in favor or orchestrator docs

nnshah1 (approved):
LGTM!

PR #7717, opened 2024-10-18
Removing caching on windows.

mc-nv:
related to: https://github.com/triton-inference-server/onnxruntime_backend/pull/276

### DIFF
--- a/build.py
+++ b/build.py
@@ -667,6 +667,16 @@ def onnxruntime_cmake_args(images, library_paths):
             else TRITON_VERSION_MAP[FLAGS.version][2],
         )
     ]
+    # If variable is set, ONNX Runtime package will be downloaded from the specified URL
+    if os.getenv("TRITON_ONNXRUNTIME_PACKAGE_URL"):
+        cargs.append(
+            cmake_backend_arg(
+                "onnxruntime",
+                "TRITON_ONNXRUNTIME_PACKAGE_URL",
+                None,
+                os.getenv("TRITON_ONNXRUNTIME_PACKAGE_URL"),
+            )
+        )
 
     # TRITON_ENABLE_GPU is already set for all backends in backend_cmake_args()
     # TODO: TPRD-334 TensorRT extension is not currently supported by our manylinux build
--- a/build.py
+++ b/build.py
@@ -1689,7 +1689,9 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
             if FLAGS.container_memory:
                 baseargs += ["--memory", FLAGS.container_memory]
 
-        baseargs += ["--cache-from={}".format(k) for k in cachefrommap]
+        if target_platform() != "windows":
+            baseargs += ["--cache-from={}".format(k) for k in cachefrommap]
+
         baseargs += ["."]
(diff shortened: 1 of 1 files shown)


mc-nv:
related to: https://github.com/triton-inference-server/onnxruntime_backend/pull/276

build.py
```diff
@@ -667,6 +667,16 @@ def onnxruntime_cmake_args(images, library_paths):
             else TRITON_VERSION_MAP[FLAGS.version][2],
         )
     ]
+    # If variable is set, ONNX Runtime package will be downloaded from the specified URL
+    if os.getenv("TRITON_ONNXRUNTIME_PACKAGE_URL"):
```
nv-kmcgill53, line 671: I would like for this to be available as a command line option as well. We can do this one or two ways, (1) `--onnxruntime-package-url` or (2) `--extra-backend-cmake-arg=onnxruntime:TRITON_ONNXRUNTIME_PACKAGE_URL=http://example.com`.  

My reasoning here is that it should be clear from the command what behavior we are inducing in the build. If we are relying _only_ on an environment variable then it will make the build that much more opaque to the user and for us when/if we need to debug a user's issue.

nv-kmcgill53 (changes requested):

nv-kmcgill53:
Overall, I like the direction this is going and the solution, just needs some fine tuning.

build.py
```diff
@@ -1679,7 +1679,9 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
             if FLAGS.container_memory:
                 baseargs += ["--memory", FLAGS.container_memory]
 
-        baseargs += ["--cache-from={}".format(k) for k in cachefrommap]
+        if target_platform() != "windows":
```
fpetrini15, line 1682: Won't this change adversely impact Triton users who do not have access to the prebuilt ORT binaries?

mc-nv: Not it shouldn't. 
1. `--cache-from` flag given to top level docker process, not to Docker-out-of-Docker.
2. Preload cache on our runners take about 10-15 minutes.

We can drop this change if your concerns are adamant.

fpetrini15: I'm probably fine with the change, just trying to understand. I'm imagining a scenario where a user is attempting to do a rebuild of Triton, in which case, I'd assume utilizing these caches would speed up their rebuild time.

Seems like the main benefit from our end is that it speeds our builds up by 10-15 minutes, though?

mc-nv: > Seems like the main benefit from our end is that it speeds our builds up by 10-15 minutes, though?
Do you have a confirmation that it speeds up our build?

fpetrini15: I mean, it's a build cache--there wouldn't be much point of having it if it didn't speed up the build. I don't have the cycles to confirm via experimentation.

All that said, I think most Windows users download the release asset and do not build from scratch. Additionally, if this would improve our CI build times I view it as a net win. If build performance issues are reported, it's simple to revert.

fpetrini15 (approved):

nv-kmcgill53 (approved):

PR #7720, opened 2024-10-18
refactor: moving `tritonfrontend` to `@handle_triton_error` decorator

KrishnanPrash:
#### What does the PR do?
Abstracts error handling in the `tritonfrontend/_api` away with the decorator `@handle_triton_error`. This removes the need to wrap every function in `_api/_<service>.py` with a `try: ... catch: ...` manually. 

Side Note for the dunder functions:
For `_api/_<service>.py::__enter__()`: Since this method just calls `start()` and `start()` has `@handle_triton_error`, any error that occurs in `__enter__()` will be handled by the desired `try: ... catch: ...`.
For `_api/_<service>.py::__exit__(self, exc_type, exc_value, traceback)`: If an exception occurs when inside the context, it will gracefully call exit. But to be captured by `@handle_triton_error` it must be thrown again.

Look at `src/python/tritonfrontend/_api/_error_mapping.py`

#### Test plan:
Ran with existing test cases `L0_python_api` to verify that errors are being handled identically. 

- CI Pipeline ID: 19504059


src/python/tritonfrontend/_api/_error_mapping.py
```diff
@@ -47,3 +49,15 @@
     AlreadyExistsError: tritonserver.AlreadyExistsError,
     UnsupportedError: tritonserver.UnsupportedError,
 }
+
+
+def handle_triton_error(func):
+    def error_handling_wrapper(self, *args, **kwargs):
```
GuanLuo, line 55: Do you need `self` here?

GuanLuo (approved):

src/python/tritonfrontend/_api/_error_mapping.py
KrishnanPrash: Removed. Thanks for suggestion.

src/python/tritonfrontend/_api/_kservegrpc.py
```diff
@@ -90,47 +88,38 @@ def __post_init__(self):
...
     def __enter__(self):
         self.triton_frontend.start()
         return self
 
+    @handle_triton_error
     def __exit__(self, exc_type, exc_value, traceback):
         self.triton_frontend.stop()
         if exc_type:
-            raise ERROR_MAPPING[exc_type](exc_value) from None
+            raise exc_type(exc_value) from None
```
rmccorm4, line 117: Why did this one keep the `from None` but the others didn't?

PR #7735, opened 2024-10-23
test: Add small delay to L0_lifecycle test_load_new_model_version after each model reload

kthui:
#### What does the PR do?
Add a small delay to L0_lifecycle test_load_new_model_version after each model reload, to prevent flaky results due to model version load/unload is not completed by the time it is checked by the test script.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:
N/A

Start by looking into the L0_lifecycle failure of pipeline 19628457

#### Test plan:
L0_lifecycle pass after the patch.

- CI Pipeline ID: 19662667

#### Caveats:
N/A

#### Background
See https://github.com/triton-inference-server/server/pull/7730 and job 117883843

N/A

### DIFF
--- a/qa/L0_lifecycle/lifecycle_test.py
+++ b/qa/L0_lifecycle/lifecycle_test.py
@@ -3529,6 +3529,8 @@ class LifeCycleTest(tu.TestResultCollector):
             f.truncate(0)
             f.seek(0)
             f.write(config)
+        # make sure the disk operation is done before reloading
+        time.sleep(0.1)
         # reload the model
         client.load_model(model_name)
 
@@ -3550,6 +3552,8 @@ class LifeCycleTest(tu.TestResultCollector):
 
         # simulate a dependency change to all versions
         Path(os.path.join("models", model_name, "dummy_dependency.py")).touch()
+        # make sure the disk operation is done before reloading
+        time.sleep(0.1)
         # reload the model
         client.load_model(model_name)
 
@@ -3577,6 +3581,8 @@ class LifeCycleTest(tu.TestResultCollector):
             f.truncate(0)
             f.seek(0)
             f.write(config)
+        # make sure the disk operation is done before reloading
+        time.sleep(0.1)
         # reload the model
         client.load_model(model_name)
 
@@ -3604,6 +3610,8 @@ class LifeCycleTest(tu.TestResultCollector):
             f.truncate(0)
             f.seek(0)
             f.write(config)
+        # make sure the disk operation is done before reloading
+        time.sleep(0.1)
         # reload the model
         client.load_model(model_name)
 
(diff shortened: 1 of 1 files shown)


rmccorm4 (commented):
Just to check my understanding, is the flakiness coming from checks for model versions expected to be "not ready"?

I.e. The calls to "load" in some test cases are actually triggering a "reload" that may "unload" some instances (removed versions from from the diff of new config), and the server doesn't wait for each unload to completely finish before returning to the client, so the checks for "not ready" models may fail if it hasn't fully unloaded on server side yet?

I would expect "loads" to block on server side until actually complete and model considered ready before returning to client. Whereas I believe that "unloads" may not block until considered "not ready", so I want to double check if the flakiness is coming from one of these "unload" cases via a version-policy-change-only "reload".

kthui:
I looked deeper into the failed server and client log, and found the [last load operation](https://github.com/triton-inference-server/server/blob/98edccd88ac7b8526c6bc5309640ca55204850c9/qa/L0_lifecycle/lifecycle_test.py#L3596-L3608) was successful according to the client log, but according to the server log the load operation did not commence. Subsequently, the [check for model version 1 is loaded](https://github.com/triton-inference-server/server/blob/98edccd88ac7b8526c6bc5309640ca55204850c9/qa/L0_lifecycle/lifecycle_test.py#L3611) failed.

If I remove the [model config update](https://github.com/triton-inference-server/server/blob/98edccd88ac7b8526c6bc5309640ca55204850c9/qa/L0_lifecycle/lifecycle_test.py#L3598-L3606) from the test, I was able to reproduce the same client and server log locally.

Base on the finding, my hypothesis for the failure is the model config update has been written to disk, according to the Python script, which at that point the server should see the updated file, but for some reason the server was seeing the old version by the time the load operation arrived. I will update the test accordingly.

kthui:
The test script on the container is updated to the latest and re-ran, see job 118303410

qa/L0_lifecycle/lifecycle_test.py
rmccorm4, line 1: Moving to thread.

> Base on the finding, my hypothesis for the failure is the model config update has been written to disk, according to the Python script, **which at that point the server should see the updated file, but for some reason the server was seeing the old version by the time the load operation arrived**. I will update the test accordingly.

Is there a bug in checking file modification time or something?

kthui: I think this is unique to the CI, because I ran the test from the failing container locally 10 times and they all passed. I ran the test ~560 times last night on the CI without any `time.sleep()` and they all passed, see job 118186442.

kthui: I think another possibility is the timestamp is the same before and after the file update, but this breaks our assumption that a ns precision timestamp is sufficient at determining the order of events.

rmccorm4 (approved):
If it starts failing again even with this change, we should re-evaluate the methodology here

Issue #7688, opened 2024-10-08
Possible bug in reference counting with shared memory regions


hcho3:
**Description**
https://github.com/triton-inference-server/server/pull/7567 introduced a reference counter to shared memory region to prevent users from releasing shared memory regions while any inference requests are still being processed.

I encountered an issue where the shared memory region could not be freed even when no inference request was ongoing.

Error I got:
```
tritonclient.utils.InferenceServerException: [StatusCode.INTERNAL] Cannot unregister shared memory region
'input_eec14889775b4c29a99d81458b0feb1a', it is currently in use
```

**Triton Information**
Triton Server 2.50.0 from NGC container 24.09

**To Reproduce**
1. Extract [model_repository.zip](https://github.com/user-attachments/files/17299383/model_repository.zip) into a directory named `model_repository/`. The directory should have the following layout after extraction:

```
model_repository/
model_repository/xgboost_json
model_repository/xgboost_json/config.pbtxt
model_repository/xgboost_json/1
model_repository/xgboost_json/1/xgboost.json
```
The example uses an XGBoost model saved as `xgboost_json` format and loads it using the FIL backend.
Notably, the model is configured to use the CPU for inference (`instance_group [{ kind: KIND_CPU }]`).

2. Launch the Triton server using the latest NGC container:
```
docker run -it --rm --network=host --ipc=host --pid=host --shm-size=1g \
  --gpus '"device=0"' \
  -v $(pwd)/model_repository:/model_repository \
  nvcr.io/nvidia/tritonserver:24.09-py3 \
  tritonserver --model-repository=/model_repository
```

3. Create the script named `test.py` with the following content:

<details>
<summary>
The `test.py` script (Click to expand
</summary>

```python
from argparse import ArgumentParser
from collections import namedtuple
from uuid import uuid4

import numpy as np
import tritonclient.grpc as triton_grpc
import tritonclient.http as triton_http
import tritonclient.utils as triton_utils
import tritonclient.utils.cuda_shared_memory as shm

STANDARD_PORTS = {"http": 8000, "grpc": 8001}
TritonInput = namedtuple("TritonInput", ("name", "handle", "input"))
TritonOutput = namedtuple("TritonOutput", ("name", "handle", "output"))


class TritonMessage:
    """Adapter to read output from both GRPC and HTTP responses"""

    def __init__(self, message):
        self.message = message

    def __getattr__(self, attr):
        try:
            return getattr(self.message, attr)
        except AttributeError:
            try:
[...]

PR #7743, opened 2024-10-25
feat: Enable deferred unregistering of shared memory regions after inference

pskiran1:
#### What does the PR do?
Currently, an error is returned if an attempt is made to unregister a shared memory region that is in use during inference (introduced in #7567). This PR introduces the following enhancement:

- When the unregister API is called for a region still in use, the API will now mark the region for pending unregistration. The server will automatically remove the region once all in-progress inferences utilizing it have completed.
- In the interim, users can query the server to check if the region is still retained, allowing for safe cleanup once the server has fully released the region.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 19695563

- closes GitHub issue: #7688


qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
```diff
@@ -401,74 +401,31 @@ def test_infer_byte_size_out_of_bound(self):
...
         second_client = httpclient.InferenceServerClient("localhost:8000", verbose=True)
 
-        with self.assertRaises(InferenceServerException) as ex:
-            second_client.get_cuda_shared_memory_status("input0_data")
-        self.assertIn(
-            "Unable to find cuda shared memory region: 'input0_data'",
-            str(ex.exception),
-        )
+        status = second_client.get_cuda_shared_memory_status()
+        self.assertEqual(len(status), 0)
```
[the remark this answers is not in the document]
pskiran1: Thank you. I modified `_test_shm_not_found` to check the shared memory status by name, similar to the previous implementation. I also cleaned up the CUDA shared memory test similar to your recent change #7729.

src/http_server.h
```diff
@@ -272,13 +272,34 @@ class HTTPAPIServer : public HTTPServer {
...
+        auto shm_name = shm_regions_info_.back()->name_;
+        auto shm_memory_type = shm_regions_info_.back()->kind_;
+        auto awaiting_unregister =
+            shm_regions_info_.back()->awaiting_unregister_;
+
+        // Delete shared_ptr to decrement reference count
+        shm_regions_info_.pop_back();
+
+        if (awaiting_unregister) {
+          auto err = shm_manager_->Unregister(shm_name, shm_memory_type);
```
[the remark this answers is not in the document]
pskiran1: Done. Added a `nullptr` check on `shm_manager_` before calling `Unregister()` and log verbose if it is `nullptr`.

src/shared_memory_manager.h
```diff
@@ -72,6 +72,7 @@ class SharedMemoryManager {
     void* mapped_addr_;
     TRITONSERVER_MemoryType kind_;
     int64_t device_id_;
+    bool awaiting_unregister_;
```
nnshah1, line 75: question: could this be done with smart pointer referencing counting instead of an explicit flag?

pskiran1: @nnshah1, yes, I believe we can implement this with `weak_ptr`.
@GuanLuo has previously provided input on an alternative approach using `weak_ptr` and also discussed the current method.
Here’s a summary of the `weak_ptr` approach:
Currently, the shared memory (shm) manager uses a map of shared_ptr for shm_info. When the unregister API is called on a region still in use, we can convert it to a `weak_ptr` to decrease the reference count. We would track these in a separate `weak_ptr map` until fully unregistered, preventing new regions with the same name and allowing status check.

When the last reference to the `shm_info` `shared_ptr` is released (after inference completes), the `shared_ptr` destructor will trigger. At that point, we can invoke the unregister logic and remove the associated `weak_ptr` from the second map.

To support this approach, we would likely need to modify most of shm manager methods, such as register, status, unregister, and unregisterHelper, etc. to work with the `weak_ptr map`.

For simplicity, I proceeded with the current approach. Please let me know if we’d like to explore the `weak_ptr` approach, and I can work on it. Happy to connect and discuss if any optimizations are needed.
Thank you.

nnshah1: Thanks for the detailed response - it was just that it seemed like a lot of similarity with reference counting and shared pointers - so was hoping it was something quick. But if this is the more practical short term - no issue, 

@GuanLuo - would we want to pre-emptively put a refactor story in? (only if you think its a near term thing that needs to be addressed) -

qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
```diff
@@ -397,171 +409,222 @@ def test_infer_byte_size_out_of_bound(self):
...
+
+        # Unregister all should not result in an error.
+        # If shared memory regions are in use, they will be marked and unregistered after the inference is completed.
+        with httpclient.InferenceServerClient(
+            "localhost:8000", verbose=True
+        ) as second_client:
+            second_client.unregister_cuda_shared_memory()
+
+        # Number of shared memory regions should be the same as the inference is not completed yet
+        self._test_shm_found(shm_names)
```
GuanLuo, line 427: So the region will still be visible between unregister and completion of the inference, what would happen if another inference request is sent during that time, trying to reuse the unregistered region? Can you add a test for this case to demonstrate the behavior?

src/shared_memory_manager.h
GuanLuo: I have raised the point that using smart pointer is a cleaner way to address this problem early on for this development, so you can have a refactor story as it is just a right thing to do. However, I believe that it is just going to be another form of TODOs at this point and what is in this PR will remain unchanged for a long time.

qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
pskiran1: Yes, the shm will remain accessible between unregister and the completion of the inference. When we run another inference request that reuses the unregistered region, the shm region will be extended and unregistered after the second request, as we check the [reference count](https://github.com/triton-inference-server/server/pull/7743/files#diff-b2df344fcbed5fc3756d51cd3ea194ac9edecbcc7a6f40022642520c92cf8f78L688) before unregister. Added a test case to verify this.

GuanLuo (approved):

PR #7744, opened 2024-10-25
build: RHEL8 Python Backend

fpetrini15:
#### What does the PR do?

**Goal:** Provide RHEL8 support for the Python backend.

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

PYBE changes: https://github.com/triton-inference-server/python_backend/pull/385

- CI Pipeline ID:

### DIFF
--- a/build.py
+++ b/build.py
@@ -78,6 +78,7 @@ TRITON_VERSION_MAP = {
         "2024.0.0",  # Standalone OpenVINO
         "3.2.6",  # DCGM version
         "0.5.3.post1",  # vLLM version
+        "3.10.13",  # Python version
     )
 }
 
@@ -908,7 +909,7 @@ RUN curl -o /tmp/cuda-keyring.deb \\
                 )
 
 
-def create_dockerfile_buildbase_rhel(ddir, dockerfile_name, argmap):
+def create_dockerfile_buildbase_rhel(ddir, dockerfile_name, argmap, backends):
     df = """
 ARG TRITON_VERSION={}
 ARG TRITON_CONTAINER_VERSION={}
@@ -988,6 +989,19 @@ RUN wget -O /tmp/boost.tar.gz \\
 #       && apt-get update -q=2 \\
 #       && apt-get install -y --no-install-recommends cmake=3.27.7* cmake-data=3.27.7*
 """
+    if "python" in backends:
+        df += """
+# python3.10 is not available through yum for RHEL8. It must instead must be installed via pyenv.
+# Only copy out libpython3.1* so as to not overwrite the general libpython3.so lib.
+RUN curl https://pyenv.run | bash \\
+    && echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bashrc \\
+    && echo '[[ -d $PYENV_ROOT/bin ]] && export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bashrc \\
+    && echo 'eval "$(pyenv init -)"' >> ~/.bashrc
+RUN CONFIGURE_OPTS=\"--with-openssl=/usr/lib64\" && pyenv install {} \\
+    && cp /root/.pyenv/versions/{}/lib/libpython3.1* /usr/lib64/""".format(
+            TRITON_VERSION_MAP[FLAGS.version][7], TRITON_VERSION_MAP[FLAGS.version][7]
+        )
+
     if FLAGS.enable_gpu:
         df += install_dcgm_libraries(argmap["DCGM_VERSION"], target_machine())
     df += """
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -78,6 +78,7 @@
         "2024.0.0",  # Standalone OpenVINO
         "3.2.6",  # DCGM version
         "0.5.3.post1",  # vLLM version
+        "3.12.1",  # RHEL Python version
```
mc-nv, line 81: Default install version for ubuntu is 3.12.3 - can we align ?

build.py
```diff
@@ -1514,6 +1540,37 @@ def add_cpu_libs_to_linux_dockerfile(backends, target_machine):
...
+""".format(
+        major_minor_version
+    )
+    df += """
+# The python library version available for install via 'yum install python3.X-devel' does not
+# match the version of python inside the RHEL base container. This means that python packages
+# installed within the container will not be picked up by the python backend stub process pybind
+# bindings. It must instead must be installed via pyenv.
+RUN curl https://pyenv.run | bash
+ENV PATH="/root/.pyenv/bin:$PATH"
```
mc-nv, line 1565: This path won't be available to users if they decide to use `triton-server` user, will it impact them?

build.py
```diff
@@ -1514,6 +1540,37 @@ def add_cpu_libs_to_linux_dockerfile(backends, target_machine):
...
+RUN eval "$(pyenv init -)"
```
mc-nv, line 1566: Why we need this version switch? is it needed for test purpose?

build.py
```diff
@@ -1514,6 +1540,37 @@ def add_cpu_libs_to_linux_dockerfile(backends, target_machine):
...
+RUN CONFIGURE_OPTS=\"--with-openssl=/usr/lib64\" && pyenv install {} \\
+    && cp /root/.pyenv/versions/{}/lib/libpython3* /usr/lib64/""".format(
```
mc-nv, line 1568: Again we have involved root folder.

build.py
fpetrini15: We want to be aligned with DLFW which is using Python v3.12. The base container comes installed with the python 3.12 interpreter, but we also need to python3.12 libraries (e.g., libpython3.12.so) to build the python  backend. We install the library we need using `pyenv`, which was suggested to me via the DLFW team.

build.py
fpetrini15: I can change this, one moment.

build.py
```diff
@@ -1514,6 +1540,36 @@ def add_cpu_libs_to_linux_dockerfile(backends, target_machine):
...
+RUN CONFIGURE_OPTS=\"--with-openssl=/usr/lib64\" && pyenv install {} \\
+    && cp ${{PYENV_ROOT}}/versions/{}/lib/libpython3* /usr/lib64/""".format(
+        version, version
+    )
+    df += """
+# RHEL image has several python versions. It's important
+# to set the correct version, otherwise, packages that are
+# pip installed will not be found during testing.
+ENV PYVER {}
+ENV PYTHONPATH /opt/python/v
```
mc-nv, line 1562: To save some space.
```suggestion
ENV PYVER={} PYTHONPATH=/opt/python/v
```

build.py
```diff
@@ -1514,6 +1540,36 @@ def add_cpu_libs_to_linux_dockerfile(backends, target_machine):
...
+RUN ln -sf ${{PYENV_ROOT}}/versions/${{PYVER}}* ${{PYTHONPATH}}
+ENV PYBIN ${{PYTHONPATH}}/bin
+ENV PYTHON_BIN_PATH ${{PYBIN}}/python${{PYVER}}
+ENV PATH ${{PYBIN}}:${{PATH}}
```
mc-nv, line 1566: Hard to say how big is our dockerfile in terms on lines trying to make sure it compact.
```suggestion
ENV PYBIN=${{PYTHONPATH}}/bin PYTHON_BIN_PATH=${{PYBIN}}/python${{PYVER}} PATH=${{PYBIN}}:${{PATH}}
```

mc-nv: Wondering why did you use a a notation `${{var}}` vs `${var}`

fpetrini15: Because this code is written in a Python script that uses `format`, we need to escape the `{}` otherwise it'll interpret it as a variable instead of treating it like a literal like we want.

Context: https://stackoverflow.com/questions/5466451/how-do-i-escape-curly-brace-characters-characters-in-a-string-while-using

fpetrini15: Local testing demonstrates that we cannot define a variable using another variable that is defined in the same line. For example, `PYTHON_BIN_PATH` and `PATH` here both depend on the definition of `PYBIN`. It must be defined before we use it to set these dependent values.

mc-nv (approved):

build.py
```diff
@@ -1514,6 +1540,34 @@ def add_cpu_libs_to_linux_dockerfile(backends, target_machine):
     return df
 
 
+def change_default_python_version_rhel(version):
+    df = """
+# The python library version available for install via 'yum install python3.X-devel' does not
+# match the version of python inside the RHEL base container. This means that python packages
+# installed within the container will not be picked up by the python backend stub process pybind
+# bindings. It must instead must be installed via pyenv.
+ENV PYENV_ROOT=/tmp/pyenv_build
```
mc-nv, line 1549: I don't know if it good idea to store it in `/tmp` folder. 
This folder is subject to store short-lived temporary files and may be cleaned up on OS restart.

fpetrini15: Where would you prefer I store it? Traditionally this is where we've stored our dependencies, such as `boost`.

Additionally, I think it is meant to be short-lived. Recall that we're only interested in gathering the python libs, which we extract with the other binaries we ship. Since this is container-based, I'm not concerned about a restart.

mc-nv: `/opt` is a first thing that coming to my mind.

mc-nv (approved):

PR #7752, opened 2024-10-29
test: Test model_metrics config and document histogram buckets override

yinggeh:
#### What does the PR do?
Tests new model_metrics message in pbtxt. Add example to override histogram buckets per-family.

#### Checklist

#### Commit Type:
- [x] docs
- [x] test

#### Related PRs:
https://github.com/triton-inference-server/core/pull/405
https://github.com/triton-inference-server/common/pull/126

#### Test plan:
L0_metrics

- CI Pipeline ID:
20106057

#### Background
Default histogram buckets does not satisfy all use cases.


qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -165,14 +171,38 @@ def check_existing_metrics(model_name, wait_secs_per_req, delta):
...
             )
             self.assertNotIn(non_decoupled_model_count, histogram_dict)
             self.assertNotIn(non_decoupled_model_sum, histogram_dict)
 
+    def test_buckets_override(self):
+        model_name = "async_execute_decouple"
+        metrics = self.get_metrics()
+        override_buckets = [
+            x for x in os.environ.get("OVERRIDE_BUCKETS").split(",")
+        ] + ["+Inf"]
```
GuanLuo, line 187: get range from model config to avoid hard-coded values in multiple places

qa/L0_metrics/test.sh
```diff
@@ -460,17 +461,117 @@ check_unit_test "${expected_tests}"
...
+
+run_server
+if [ "$SERVER_PID" != "0" ]; then
+    echo -e "*** FAILED: unexpected success starting $SERVER" >> $CLIENT_LOG
+    RET=1
+    kill_server
+else
+    export SERVER_LOG_ERROR="model_control must specify 'metric_identifier'"
+    python3 ${HISTOGRAM_PYTEST} HistogramMetricsTest.test_server_log_error 2>&1 | tee ${CLIENT_LOG}
+    check_unit_test
```
GuanLuo, line 525: Test this in L0_model_config and follow the style there.

GuanLuo: What is the behavior for specifying family name that doesn't exist / can't be customized? Add tests for those.

qa/L0_metrics/histogram_metrics_test.py
yinggeh: Updated

qa/L0_metrics/test.sh
yinggeh: Updated comments

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -165,14 +171,29 @@ def check_existing_metrics(model_name, wait_secs_per_req, delta):
...
+                FIRST_RESPONSE_HISTOGRAM, non_decoupled_model_name, "1", "count"
             )
             non_decoupled_model_sum = get_histogram_metric_key(
-                first_response_family, non_decoupled_model_name, "1", "sum"
+                FIRST_RESPONSE_HISTOGRAM, non_decoupled_model_name, "1", "sum"
             )
             self.assertNotIn(non_decoupled_model_count, histogram_dict)
             self.assertNotIn(non_decoupled_model_sum, histogram_dict)
 
+    def test_buckets_override(self):
```
rmccorm4, line 182: Do you have a test to see what happens when you pass some invalid or non-existent family in the new `config.pbtxt` metric_control fields?

Some examples:
1. Specify valid `nv_inference_first_response_histogram_ms` but histograms are disabled so it shouldn't be present -- what happens?
2. Specify invalid non-existent metric `abc123` - what happens?

rmccorm4: Looks like these might be in `unknown_metric_family` tests in `L0_model_config`, but double checking

rmccorm4: It looks like (1) might still be a valid case to test, because it looks like this metric is in the `metric_map_` by default [here](https://github.com/triton-inference-server/core/blob/e550d6b915ec8efd11bead0b0fc2e3d739c9d3ef/src/metric_model_reporter.h#L86-L87), so this [check](https://github.com/triton-inference-server/core/blob/e550d6b915ec8efd11bead0b0fc2e3d739c9d3ef/src/metric_model_reporter.cc#L85) will always pass.

Looks like it might be harmless - but sharing in case it would be unexpected to hit this code path if the metric family didn't actually exist (ex: with histograms disabled).

qa/L0_metrics/histogram_metrics_test.py
```diff
@@ -40,31 +40,38 @@
...
     if metric_type in ["count", "sum"]:
         return f'{metric_family}_{metric_type}{{model="{model_name}",version="{model_version}"}}'
+    elif metric_type == "bucket":
+        return f'{metric_family}_{metric_type}{{model="{model_name}",version="{model_version}",le="{le}"}}'
     else:
         return None
 
 
-class TestHistogramMetrics(tu.TestResultCollector):
+class HistogramMetricsTest(tu.TestResultCollector):
```
rmccorm4, line 57: Please keep `Test` prefix in class name, it's required to be discovered by pytest, which we are gradually moving towards using for it's automatic test reports: https://docs.pytest.org/en/7.1.x/getting-started.html#group-multiple-tests-in-a-class. If someone tries to call this test with `pytest` in the future, I think it will fail to find any tests and they may miss that.

rmccorm4: example pytest usage: https://github.com/triton-inference-server/server/blob/4af437e4902ac2b540727e4985620e019f063161/qa/L0_metrics/test.sh#L146

qa/L0_metrics/histogram_metrics_test.py
yinggeh: Case 1 is fine. I can add a test case to make sure the server produces no errors.

yinggeh: Added test for valid model_metrics config with histogram disabled

qa/L0_metrics/histogram_metrics_test.py
yinggeh: Updated

qa/L0_metrics/test.sh
```diff
@@ -460,16 +461,57 @@ check_unit_test "${expected_tests}"
...
+        buckets: [ -1, 0.0, 1, 2.5 ]
+      }
+    }
+  ]
+}
+EOL
+
+SERVER_ARGS="--model-repository=${MODELDIR} --model-control-mode=explicit --load-model=${decoupled_model} --metrics-config histogram_latencies=true --log-verbose=1"
+run_and_check_server
+export OVERRIDE_BUCKETS="-1,0,1,2.5,+Inf"
```
GuanLuo, line 504: You are still hard-coding the buckets value..

qa/L0_model_config/model_metrics/valid_config_with_warn/unknown_metric_family/expected
```diff
@@ -0,0 +1 @@
+Metric family 'nv_inference_request_success' in 'metric_identifier' is not a Triton core metric.
```
GuanLuo, line 1: This error message is misleading, should mention that the metrics is not customizable metrics

yinggeh: Updated

docs/user_guide/metrics.md
```diff
@@ -239,7 +239,34 @@ nv_inference_first_response_histogram_ms{model="my_model",version="1", le="5000"
...
+        buckets: [ 1, 2, 4, 8 ]
+      }
+    }
+  ]
+}
+```
+
+> **Note**
+>
+> To apply changes to metric options dynamically, the model must be completely
```
rmccorm4, line 263: Nice addition 🚀

yinggeh:
Verified CI has passed without introducing new failures. @rmccorm4 @GuanLuo

rmccorm4 (approved):

PR #7756, opened 2024-10-30
ci: Adding tests for `numpy>=2`

KrishnanPrash:
#### What does the PR do?
Modifying `L0_backend_python/examples/test.sh` to test the python backend when `numpy>=2` is installed in the environment. This is achieved by:
- Installing updated versions of `numpy>=2.0`, `torch==2.5.0`, and `torchvision==0.20.0` at the start of the subtest
- Running through the existing subtest in `L0_backend_python/examples/test.sh`
- Uninstalling `numpy>=2` and re-installing `numpy<2` so future subtests will still use `numpy 1.x`

Reason for limited scope:
Support for `numpy>=2` proved to be more complicated than intended because older version of packages like `torch`, `torchvision`, and `tensorflow` do not support for `numpy 2.x`.

#### Related PRs:
Python Backend PR: [Link](https://github.com/triton-inference-server/python_backend/pull/384)

- CI Pipeline ID: 19960045


qa/L0_backend_python/examples/test.sh
```diff
@@ -38,10 +38,12 @@ rm -fr *.log python_backend/
...
 # Install torch
 pip3 uninstall -y torch
+pip3 uninstall -y numpy
+pip3 install "numpy>=2"
 if [ "$TEST_JETSON" == "0" ] && [[ ${TEST_WINDOWS} == 0 ]]; then
-    pip3 install torch==2.0.0+cu117 -f https://download.pytorch.org/whl/torch_stable.html torchvision==0.15.0+cu117
+    pip3 install torch==2.5.0 -f https://download.pytorch.org/whl/torch_stable.html torchvision==0.20.0
 else
-    pip3 install torch==2.0.0 -f https://download.pytorch.org/whl/torch_stable.html torchvision==0.15.0
+    pip3 install torch==2.5.0 -f https://download.pytorch.org/whl/torch_stable.html torchvision==0.20.0
```
rmccorm4, line 46: Note that the `if` condition installed a `+cu` version, and the `else` did not. In your change you have both if/else installing the same thing. Please either match the install pattern with the correct `+cuXXX` variant, or double check whether this if/else is required if not. CC @fpetrini15 who might know a bit more about the if/else split here.

fpetrini15: @rmccorm4 I think this is a code segment we discussed a few months back. It seems like since Jacky has fixed my original error here; https://github.com/triton-inference-server/server/commit/23a6c2189db716bc9d00e4ba346969be6ed69ba2. I think for non-Jetson Linux tests, we want to continue installing the cuda-enabled version. That said, these tests were passing fine even with my logical error from months ago, so it's unclear to me what we achieve by installing the cuda-enabled version.

rmccorm4: If this doesn't break windows/jetson, I'm happy to keep it. If it might, we could also try making the change only on the linux non-jetson block. ex:

```
if not jetson/windows:
  pip install numpy 2
  pip install torch 2.5
else: # if windows/jetson
  pip install numpy 1
  pip install torch 2.0
fi
``` 

Let me know if you have a preference for minimal blast radius @fpetrini15

fpetrini15: I think we should align with the original intention of the script and install the cuda-enabled version.

CC @Tabrizian if you have any insight into why we preferred it originally?

qa/L0_backend_python/examples/test.sh
```diff
@@ -440,4 +442,7 @@ else
     echo -e "\n***\n*** Example verification test FAILED.\n***"
 fi
 
+pip3 uninstall -y numpy
+pip3 install "numpy<2"
```
KrishnanPrash, line 446: In L0_backend_python, the way `examples/test.sh` is called is along the lines of:
```
    setup_virtualenv

    set +e
    (cd ${TEST} && bash -ex test.sh)

    if [ $? -ne 0 ]; then
        echo "Subtest ${TEST} FAILED"
        RET=1
    fi
    set -e

    deactivate_virtualenv
```
So re-installing a lower version of numpy is unnecessary here because the virtualenv created for the examples subtest is thrown away and a new one is created for future next subtest.

qa/L0_backend_python/examples/test.sh
Tabrizian: I believe PyTorch with CUDA was not supported on Jetson. We also don't support GPU tensors on Jetson with Python backend so it could be related to that.

qa/L0_backend_python/examples/test.sh
```diff
@@ -38,10 +38,12 @@ rm -fr *.log python_backend/
 
 # Install torch
 pip3 uninstall -y torch
+pip3 uninstall -y numpy
+pip3 install "numpy>=2"
```
rmccorm4, line 42:
```suggestion
# NOTE: Using this subtest as a test case that involves using a python model with 
# numpy 2.X without changing the environments used in all the other test cases.
pip3 install "numpy>=2"
```

qa/L0_backend_python/examples/test.sh
```diff
@@ -38,10 +38,14 @@ rm -fr *.log python_backend/
 
 # Install torch
 pip3 uninstall -y torch
+pip3 uninstall -y numpy
+# NOTE: Using this subtest as a test case that involves using a python model with 
+# numpy 2.X without changing the environments used in all the other test cases.
```
rmccorm4, line 43: hopefully fix precommit
```suggestion
# NOTE: Using this subtest as a test case that involves using a python model with
# numpy 2.X without changing the environments used in all the other test cases.
```

KrishnanPrash: just manually ran pre-commit and pushed fix.

rmccorm4 (commented):
LGTM but let's make sure the variants like python3.8, etc don't fail from this change. If they do, we might need to better handle which torch version is installed for each python version.

rmccorm4 (approved):
LGTM but let's make sure the variants like python3.8, etc don't fail from this change. If they do, we might need to better handle which torch version is installed for each python version.

qa/L0_backend_python/examples/test.sh
```diff
@@ -38,10 +38,25 @@ rm -fr *.log python_backend/
 
 # Install torch
 pip3 uninstall -y torch
+pip3 uninstall -y numpy
+# NOTE: Using this subtest as a test case that involves using a python model with
+# numpy 2.X without changing the environments used in all the other test cases.
 if [ "$TEST_JETSON" == "0" ] && [[ ${TEST_WINDOWS} == 0 ]]; then
-    pip3 install torch==2.0.0+cu117 -f https://download.pytorch.org/whl/torch_stable.html torchvision==0.15.0+cu117
+    if [ ${PYTHON_ENV_VERSION} == "8" ]; then
```
rmccorm4, line 45: Comment why you're splitting it for future readers. ex: python 3.8 doesn't support a version of torch compatible with numpy 2.x

krishung5 (commented):
LGTM. +1 on adding the comment for python 3.8.

rmccorm4 (approved):

PR #7775, opened 2024-11-08
feat: Adding RestrictedFeatures Support to the Python Frontend Bindings

KrishnanPrash:
#### What does the PR do?
This PR provides [Restricted Features](https://github.com/triton-inference-server/server/blob/r24.11/docs/customization_guide/inference_protocols.md#limit-endpoint-access-beta) support to the python frontend bindings, a.k.a `tritonfrontend`. 
This PR introduces 3 new classes to the `tritonfrontend` package:
1. `Feature`: 1-to-1 mapping of [RestrictedCategory Enum](https://github.com/triton-inference-server/server/blob/8b5f33ed333ad3ff7b5463974652bb5750e4d784/src/restricted_features.h#L38C1-L50C3)
2. `FeatureGroup`: Pydantic `@dataclass` that stores (key, value, features) information while performing type validation.
3. `RestrictedFeatures`: Store collections of `FeatureGroup` instances, apply an additional layer of validation to check for collisions (ensuring that each feature belongs to only one group), and serialize the data into a JSON string.

With these changes, we can pass a `RestrictedFeatures` to the `KServeHttp` and `KServeGrpc` frontends, which looks along the lines of:
```python3
>>> import tritonserver
>>> from tritonfrontend import KServeHttp, KServeGrpc
>>> from tritonfrontend import RestrictedFeatures
>>> server = tritonserver.Server(...).start(wait_until_ready=True)

>>> rf = RestrictedFeatures(...)

>>> http_options = KServeHttp.Options(restricted_features=rf)
>>> http_service = KServeHttp(server, http_options)
>>> http_service.start()

>>> grpc_options = KServeGrpc.Options(restricted_features=rf)
>>> grpc_service = KServeGrpc(server, grpc_options)
>>> grpc_service.start()
```

A `RestrictedFeatures` instance can be created with:
```python3
>>> from tritonfrontend import Feature, FeatureGroup, RestrictedFeatures
>>> admin_group = FeatureGroup(key="admin-key", value="admin-value", features=[Feature.HEALTH, Feature.METADATA])
>>> infer_group = FeatureGroup("infer-key", "infer-value", [Feature.INFERENCE])

>>> rf = RestrictedFeatures([admin_group, infer_group])

>>> model_group = FeatureGroup("model-key", "model-value", [Feature.MODEL_REPOSITORY])
>>> rf.add_feature_group(model_group)

>>> rf.create_feature_group("stat-key", "stat-value", [Feature.STATISTICS])
>>> rf
[  
   {"key": "admin-key", "value": "admin-value", "features": ["health"]},
   {"key": "admin-key", "value": "admin-value", "features": ["metadata"]},  
   {"key": "infer-key", "value": "infer-value", "features": ["inference"]}, 
   {"key": "model-key", "value": "model-value", "features": ["model-repository"]}, 
   {"key": "stat-key", "value": "stat-value", "features": ["statistics"]}
]
```
Take a look at `api/_restricted_features.py` and `tritonfrontend.h`

#### Test plan:
In `L0_python_api/test_kserve.py` added:
- `test_correct_rf_parameters()` and `test_wrong_rf_parameters()`: Tests for valid/invalid arguments passed/selected for `Feature`, `FeatureGroup` and `RestrictedFeature`
To test if restricted features are working correctly with the frontends:
-  `test_restricted_features()`: Restricts inference. Then, sends an inference request with a correct and incorrect header to verify proper functionality.

CI Pipeline ID: 20189649

#### Background
For more information about restricted features support in Triton, please take a look at this: [[Link](https://github.com/triton-inference-server/server/blob/r24.11/docs/customization_guide/inference_protocols.md#limit-endpoint-access-beta)] 

#### Additonal Changes Made in this PR
- Cleaned up includes in `tritonfrontend.h` and `tritonfrontend_pybind.cc`.
- Added support to `@handle_triton_error` for non-void functions.
- Added support to pass headers in `send_and_test_inference_identity()` so that we can re-use utility function for restricted features testing.
- Skipping tests that use `KServeGrpc` and `tritonclient.grpc` because of lack of `fork()` support provided by python `cygrpc`. More information in DLIS-7215.


src/python/tritonfrontend/_api/_restricted_features.py
```diff
@@ -0,0 +1,126 @@
...
+class FeatureGroup:
+    key: str
+    value: str
+    features: List[Feature]
+
+    @field_validator("features", mode="before")
+    def validate_features(features: List[Feature]) -> List[Feature]:
+        invalid_features = [item for item in features if not isinstance(item, Feature)]
+        if invalid_features:
+            raise tritonserver.InvalidArgumentError(
```
KrishnanPrash, line 62: Currently investigating why `@field_validator` is incompatible with `@handle_triton_server` nested underneath. To unblock, directly raised `tritonserver` equivalent error.

src/python/tritonfrontend/_api/_kservegrpc.py
```diff
@@ -81,13 +82,19 @@ class Options:
         ] = Grpc_compression_level.NONE
         infer_allocation_pool_size: int = Field(8, ge=0)
         forward_header_pattern: str = ""
-        # DLIS-7215: Add restricted protocol support
-        # restricted_protocols: str = ""
+        restricted_protocols: RestrictedFeatures = RestrictedFeatures()
```
rmccorm4, line 85: @GuanLuo thoughts on calling it something uniform for both GRPC and HTTP?

ex: instead of `restricted_protocols` and `restricted_apis`, how about using `restricted_apis` or `restricted_features` for both?

GuanLuo: Protocol / feature is a bit general but it is fine

KrishnanPrash: Modified `KServeHttp.Options` and `KServeGrpc.Options` use `Options(restricted_features=...)`

qa/L0_python_api/test_kserve.py
```diff
@@ -108,9 +115,63 @@ def test_wrong_http_parameters(self):
...
+            rf.create_feature_group(
+                key="key", value="val", features=[Feature.METADATA, Feature.HEALTH]
+            )
+            rf.create_feature_group(key="key2", value="val", features=[Feature.HEALTH])
+
+
 HTTP_ARGS = (KServeHttp, httpclient, "localhost:8000")  # Default HTTP args
 GRPC_ARGS = (KServeGrpc, grpcclient, "localhost:8001")  # Default GRPC args
 METRICS_ARGS = (Metrics, "localhost:8002")  # Default Metrics args
+RESTRICTED_FEATURE_ARG = RestrictedFeatures()
```
GuanLuo, line 169: Is this the pytest way to set test suit level parameters? Can you point me to the guide on that?

src/python/tritonfrontend/_api/_error_mapping.py
```diff
@@ -54,7 +54,8 @@
 def handle_triton_error(func):
     def error_handling_wrapper(*args, **kwargs):
         try:
-            func(*args, **kwargs)
+            return_val = func(*args, **kwargs)
+            return return_val
```
GuanLuo, line 58:
```suggestion
            return func(*args, **kwargs)
```

qa/L0_python_api/test_kserve.py
```diff
@@ -357,6 +418,31 @@ def test_metrics_update(self, frontend, url):
...
+        headers = {key_prefix + "fake-key": "fake-value"}
+        with pytest.raises(
+            InferenceServerException,
+            match=f"expecting header '{key_prefix}infer-key'",
+        ):
+            utils.send_and_test_inference_identity(client_type, url, headers)
+
+        utils.teardown_service(service)
+        utils.teardown_server(server)
+
```
GuanLuo, line 445: test for:
1. multiple feature groups
2. overwrite feature groups
3. remove feature groups
4. duplicate feature in different groups

src/python/tritonfrontend/_api/_restricted_features.py
```diff
@@ -0,0 +1,192 @@
...
+            >>> rf = RestrictedFeatures()
+            >>> rf.create_feature_group("infer-key", "infer-value", [Feature.INFERENCE])
+            >>> rf
+            [{"key": "infer-key", "value": "infer-value", "features": ["inference"]}]
+        """
+        group = FeatureGroup(key, value, features)
+        self.add_feature_group(group)
+
+    @handle_triton_error
+    def _gather_rest_data(self) -> dict:
```
GuanLuo, line 165: what is `rest_data`?

src/python/tritonfrontend/_api/_restricted_features.py
KrishnanPrash: The function signature gets overwritten when using the `@handle_triton_error` decorator. And with the change in function signature, pydantic's `field_validator()` does not consider the underlying function it is trying to wrap to have a valid function signature. 


To fix this we use `@functools.wraps(func)` internally within `@handle_triton_error` it copies over the original function signature. In case you are curious, this is done by overwriting the function attributes of this final function. [Link to `functools.update_wrapper()` and `functools.wraps()`](https://github.com/python/cpython/blob/49f15d8667e4755876698a8daa13bab6acee5fa1/Lib/functools.py#L35C1-L77C55).

qa/L0_python_api/test_kserve.py
KrishnanPrash: The only suit level parameter initialization and testing I could find was through fixtures and since we haven't used fixtures throughout this test, for the sake of consistency I moved the `RestrictedFeatures` object to inside the only function that was testing it. If we would like to fixtures, I can look into that as well.

qa/L0_python_api/test_kserve.py
KrishnanPrash: 1. Added a test for multiple feature groups (`TestRestrictedFeatures::test_multiple_groups()`)
2. Overwriting was complicating the logic for validating types and tracking duplicates, so I decided to set `@dataclass(frozen=True)` for `FeatureGroup` to prevent any post-initialization setting of attributes. So, currently if someone wants to overwrite a certain attribute of a `FeatureGroup` instance, they will have to remove the `FeatureGroup` instance from the `RestrictedFeatures` object and make a new `FeatureGroup` object to reflect their choices.
3. Tests for this have been added to `TestRestrictedFeatureOptions`
4. Tests for this have been added to `TestRestrictedFeatureOptions`

qa/L0_python_api/test_kserve.py
```diff
@@ -108,6 +117,62 @@ def test_wrong_http_parameters(self):
...
+            rf = RestrictedFeatures(groups=[feature_group])
+            rf.create_feature_group(key="key2", value="val", features=[Feature.HEALTH])
+
+        with pytest.raises(tritonserver.InvalidArgumentError):
+            rf = RestrictedFeatures(groups=[feature_group])
+            rf.create_feature_group(
+                key="key", value="val", features=[Feature.METADATA, Feature.HEALTH]
+            )
+            rf.create_feature_group(key="key2", value="val", features=[Feature.HEALTH])
```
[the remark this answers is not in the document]
KrishnanPrash: Removed the redundant test.

qa/L0_python_api/test_kserve.py
```diff
@@ -357,28 +450,95 @@ def test_metrics_update(self, frontend, url):
...
-    #     utils.teardown_client(http_client)
-    #     utils.teardown_service(http_service)
+        # Invalid headers sent with inference request
+        headers = {key_prefix + "fake-key": "fake-value"}
+        with pytest.raises(
+            InferenceServerException,
+            match=f"expecting header '{key_prefix}infer-key'",
+        ):
+            utils.send_and_test_inference_identity(client_type, url, headers)
+
```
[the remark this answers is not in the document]
KrishnanPrash: Added to `test_restrict_inference()`. The error message for all these cases are identical:
```
tritonclient.utils.InferenceServerException: [403] This API is restricted, expecting header 'infer-key'
```

src/python/tritonfrontend/_api/_restricted_features.py
```diff
@@ -0,0 +1,283 @@
...
+        >>> rf
+            [
+                {"key": "infer-key", "value": "infer-value", "features": ["inference"]},
+                {"key": "meta-key", "value": "meta-value", "features": ["metadata", "health"]},
+            ]
+        """
+        group = FeatureGroup(key, value, features)
+        self.add_feature_group(group)
+
+    def has_feature(self, feature: Feature):
```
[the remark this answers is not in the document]
KrishnanPrash: Created a `get_feature_groups()` and `update_feature_groups(feature, key, value)` function now.

qa/L0_python_api/test_kserve.py
KrishnanPrash: This should be possible with `update_feature_group(Feature, key, value)` now.

qa/L0_python_api/test.sh
```diff
@@ -51,7 +51,11 @@ fi
 
 
 FRONTEND_TEST_LOG="./python_kserve.log"
-python -m pytest --junitxml=test_kserve.xml test_kserve.py > $FRONTEND_TEST_LOG 2>&1
+# TODO: [DLIS-7215] Run tritonclient.grpc as separate process
+# Currently, when we run tritonclient.grpc with tritonserver in the same process,
+# When a fork() is called by tritonserver on model load, lack of support in cygrpc
+# causes the python process to abort/crash without being able to be caught by pytest.
```
rmccorm4, line 57: Can you elaborate or share references? I don't understand how the client is related to server/model fork, and what "lack of support in cygrpc" means

KrishnanPrash: Added a comment providing more context into how `cygrpc` relates to directly imported python packages. Added link for further reading and reference.

qa/L0_python_api/test.sh
```diff
@@ -51,7 +51,14 @@ fi
 
 
 FRONTEND_TEST_LOG="./python_kserve.log"
-python -m pytest --junitxml=test_kserve.xml test_kserve.py > $FRONTEND_TEST_LOG 2>&1
+# TODO: [DLIS-7215] Run tritonclient.grpc as separate process
```
rmccorm4, line 54: Is 7215 the right ticket? It has no description and is in-review

rmccorm4: Apply everywhere else you see 7215

qa/L0_python_api/test.sh
```diff
@@ -51,7 +51,14 @@ fi
...
+# Currently, running tritonclient.grpc with tritonserver in the same process,
+# it will non-deterministically abort/crash without being able to be caught by pytest.
+# This is because fork() is called by tritonserver on model load,
+# which attempts to fork the imported libraries and their internal states,
+# and cygrpc (dependency of tritonclient.grpc) does not officially support fork().
```
rmccorm4, line 59: > However, **if the application only instantiate gRPC Python objects after calling fork(), then fork() will work normally**, since there is no C extension binding at this point.

If the claim is that we're calling `fork()` during model load, aren't we instantiating grpc (client) after the `fork()` if server/service have already started up?

Can you ever reproduce this by only running a single test case repeatedly?

qa/L0_python_api/test.sh
```diff
@@ -51,7 +51,14 @@ fi
...
+# Reference: https://github.com/grpc/grpc/blob/master/doc/fork_support.md
```
GuanLuo, line 60: Couldn't you just set env var as mentioned here so that the tests work with fork? https://github.com/grpc/grpc/blob/master/doc/fork_support.md#current-status

qa/L0_python_api/test_kserve.py
```diff
@@ -108,8 +117,99 @@ def test_wrong_http_parameters(self):
...
+        # Features List needs to be an element from tritonfrontend.Feature
+        with pytest.raises(tritonserver.InvalidArgumentError):
+            rf.create_feature_group(key="", value="", features=["health"])
+
+        # key and value need to be of type string
+        with pytest.raises(Exception):
+            rf.create_feature_group(
+                key=42, value="Secret to the Universe", features=[Feature.HEALTH]
+            )
+        with pytest.raises(Exception):
```
GuanLuo, line 185: Check if exception enclose informative message

qa/L0_python_api/test.sh
KrishnanPrash: Right ticket is DLIS-7735. Thank you for the catch. Updated all references.

src/python/examples/example_model_repository/identity/config.pbtxt
```diff
@@ -25,20 +25,26 @@
...
-    dims: [ -1 ]
+    dims: [ 1 ]
   }
 ]
+instance_group [
+  {
+    count: 1
+    kind : KIND_CPU
+  }
+]
```
GuanLuo, line 50:
```suggestion
]

```

qa/L0_python_api/test.sh
KrishnanPrash: Attempted to set and run test suite locally, but the test cases were still failing.

qa/L0_python_api/test.sh
KrishnanPrash: From, my understanding it is an internal state set upon import:

```python

while True:
     server = utils.startup_server()
     sleep(1)
     utils.teardown_server(server)
```

PR #7779, opened 2024-11-08
test: OpenAI frontend invalid chat tokenizer network issue WAR

kthui:
#### What does the PR do?
Allow test_chat_completions_invalid_chat_tokenizer test to be skipped if there is a network issue when accessing the invalid tokenizer.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

N/A

#### Test plan:
The affected test should continue to pass on existing pipeline, and skipped when there is a network issue.

- CI Pipeline ID: N/A

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/python/openai/tests/test_chat_completions.py
+++ b/python/openai/tests/test_chat_completions.py
@@ -564,9 +564,15 @@ class TestChatCompletionsTokenizers:
 
         # Pick a tokenizer with no chat template defined
         invalid_chat_tokenizer = "gpt2"
-        app = setup_fastapi_app(
-            tokenizer=invalid_chat_tokenizer, server=server, backend=backend
-        )
+        try:
+            app = setup_fastapi_app(
+                tokenizer=invalid_chat_tokenizer, server=server, backend=backend
+            )
+        except OSError as e:
+            expected_msg = f"We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like {invalid_chat_tokenizer} is not the path to a directory containing a file named config.json."
+            if expected_msg in str(e):
+                pytest.skip("HuggingFace network issues")
+            raise e
         with TestClient(app) as client:
             response = client.post(
                 "/v1/chat/completions",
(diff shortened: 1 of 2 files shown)


kthui:
When there is a network issue:
```
tests/test_chat_completions.py::TestChatCompletionsTokenizers::test_chat_completions_invalid_chat_tokenizer
...
transformers.__version__='4.45.1'
SKIPPED (HuggingFace network issues)
```

python/openai/tests/test_chat_completions.py
```diff
@@ -564,9 +564,15 @@ def test_chat_completions_invalid_chat_tokenizer(
...
         invalid_chat_tokenizer = "gpt2"
-        app = setup_fastapi_app(
-            tokenizer=invalid_chat_tokenizer, server=server, backend=backend
-        )
+        try:
+            app = setup_fastapi_app(
+                tokenizer=invalid_chat_tokenizer, server=server, backend=backend
+            )
+        except OSError as e:
+            expected_msg = f"We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like {invalid_chat_tokenizer} is not the path to a directory containing a file named config.json."
```
rmccorm4, line 572: This is fine for now, but just be wary if the message changes slightly in the future it might not be an exact match.

rmccorm4 (approved):
Thanks for the fix Jacky!

For future reference, this error is probably due to some env var or change to transformers build to disable network communication with HF and only allow pre-downloaded local files, such as via `HF_HUB_OFFLINE`

python/openai/tests/test_chat_completions.py
kthui: Yes. I want to be cautious on this first, to avoid skipping the test for anything else.

If the message varies in the future, I think we can update the condition to detect if the network communication is disabled, i.e. `HF_HUB_OFFLINE`.

PR #7787, opened 2024-11-13
fix: Resolve integer overflow in Load API file decoding

pskiran1:
#### What does the PR do?
This PR addresses the following issue:

- When large data is sent to `/v2/repository/models/<model_name>/load`, the server converts the `param_len` variable from `size_t` to `const int` to pass it to `base64_decode_block`. This conversion can cause an integer overflow in `base64_decode_block`, leading to a server crash with a segmentation fault.
- A similar issue occurs with `/v2/cudasharedmemory/region/{}/register` when large data is sent for the `raw_handle` parameter.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 20358255

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -580,8 +580,8 @@ class CudaSharedMemoryTestRawHttpRequest(unittest.TestCase):
 
     def _generate_mock_base64_raw_handle(self, data_length):
         original_data_length = data_length * 3 // 4
-        large_data = b"A" * original_data_length
-        encoded_data = base64.b64encode(large_data)
+        random_data = b"A" * original_data_length
+        encoded_data = base64.b64encode(random_data)
 
         assert (
             len(encoded_data) == data_length
@@ -607,7 +607,8 @@ class CudaSharedMemoryTestRawHttpRequest(unittest.TestCase):
         return response
 
     def test_exceeds_cshm_handle_size_limit(self):
-        byte_size = 2147483648
+        # byte_size greater than INT_MAX
+        byte_size = 1 << 31
         device_id = 0
         shm_name = "invalid_shm"
 
--- a/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
+++ b/qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
@@ -30,6 +30,7 @@ import sys
 
 sys.path.append("../common")
 
+import base64
 import os
 import time
 import unittest
@@ -37,6 +38,7 @@ from functools import partial
 
 import infer_util as iu
 import numpy as np
+import requests
 import test_util as tu
 import tritonclient.grpc as grpcclient
 import tritonclient.http as httpclient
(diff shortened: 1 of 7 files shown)


qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
```diff
@@ -564,5 +566,107 @@ def callback(user_data, result, error):
...
+            raw_handle, device_id, byte_size, shm_name
+        )
+        self.assertEqual(response.status_code, 200)
+
+        # Verify shared memory status
+        status = self.client.get_cuda_shared_memory_status()
+        self.assertEqual(len(status), 1)
+        self.assertEqual(status[0]["name"], shm_name)
+
+
```
GuanLuo, line 670: I don't think the (CUDA) shared memory changes are relevant in this PR?

qa/L0_http/http_test.py
```diff
@@ -231,6 +236,43 @@ def test_descriptive_status_code(self):
...
+    def test_loading_large_invalid_model(self):
+        # Generate large base64 encoded data
+        data_length = 2147483648
+        original_data_length = data_length * 3 // 4
+        random_data = b"A" * original_data_length
+        encoded_data = base64.b64encode(random_data)
+
+        assert (
+            len(encoded_data) == data_length
+        ), "Encoded data length does not match the required length."
```
GuanLuo, line 248: I think you just need to make sure that the encoded data size is larger than INT_MAX?
so you can:
```suggestion
        data_length = 1 << 32 # 32 is sizeof(int)
        int_max = (1 << 31) -1
        random_data = b"A" * data_length
        encoded_data = base64.b64encode(random_data)

        assert (
            len(encoded_data) > int_max
        ), "Encoded data length does not match the required length."
```

pskiran1: Using `data_length = 1 << 31` to avoid JSON read errors that occur with larger value `1 << 32`
`failed to parse the request JSON buffer: Missing a closing quotation mark in string. at 4294967242
`

qa/L0_cuda_shared_memory/cuda_shared_memory_test.py
pskiran1: I also found that sending a large amount of random data in place of the `raw_handle` for the CUDA shared memory register (a raw HTTP request) causes the server to crash with a segmentation fault at [this line](https://github.com/triton-inference-server/server/pull/7787/files#diff-2d773636d85b98207338699c3cc2b9fc7d31f38e45131ca127dca73f45d4bbb2L2451). Since this issue has the same root cause, I added a validation check in the CUDA shared memory registration code and created relevant test cases to verify the fix. I have updated the PR description to include details on the CUDA shared memory issue.

src/http_server.cc
```diff
@@ -2443,6 +2453,14 @@ HTTPAPIServer::HandleCudaSharedMemory(
           }
 
           if (err == nullptr) {
+            if (b64_handle_len > INT_MAX) {
```
tanmayv25, line 2456: Can we create a wrapper function on top of `base64_decode_block`? It can take the string_len as size_t, perform this check on the length, and pass it to the `base64_decode_block` as const int. 
All the setup logic like base64_init_decodestate can be part of this wrapper too.

This change will enforce DRY by keeping the check and decode logic in a single place and hence making the code more maintainable.

tanmayv25 (approved):

PR #7837, opened 2024-11-27
ci: Fix Windows CI Errors

fpetrini15:
#### What does the PR do?

Fixes accumulated errors in Windows CI.

#### Checklist

#### Commit Type:

- [ ] build
- [x] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

CI Pipeline ID:

20833929

### DIFF
--- a/qa/L0_batcher/test.sh
+++ b/qa/L0_batcher/test.sh
@@ -85,6 +85,7 @@ if [[ -v WSL_DISTRO_NAME ]] || [[ -v MSYSTEM ]]; then
     BACKEND_DIR=${BACKEND_DIR:=C:/tritonserver/backends}
     SERVER=${SERVER:=/mnt/c/tritonserver/bin/tritonserver.exe}
     export WSLENV=$WSLENV:TRITONSERVER_DELAY_SCHEDULER
+    export USE_GRPC=1
 else
     MODELDIR=${MODELDIR:=`pwd`}
     DATADIR=${DATADIR:="/data/inferenceserver/${REPO_VERSION}"}
@@ -601,7 +602,7 @@ done
 TEST_CASE=test_multi_batch_preserve_ordering
 
 # Skip test for Windows. Trace file concats at 8192 chars on Windows.
-if  [[ ! -v WSL_DISTRO_NAME ]] || [[ ! -v MSYSTEM ]]; then
+if  [[ ! -v WSL_DISTRO_NAME ]] && [[ ! -v MSYSTEM ]]; then
     rm -fr ./custom_models && mkdir ./custom_models && \
         cp -r ../custom_models/custom_zero_1_float32 ./custom_models/. && \
         mkdir -p ./custom_models/custom_zero_1_float32/1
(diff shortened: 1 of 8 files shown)


qa/common/infer_util.py
```diff
@@ -702,6 +702,7 @@ def inferAndCheckResults(
                                 if cidx == expected0_sort_idx[b][idx]:
                                     tester.assertEqual(
                                         ctuple[2],
+                                        ctuple[2].strip("\r"),
                                         "label{}".format(expected0_sort_idx[b][idx]),
```
rmccorm4, line 706: `assertEqual(a, b, msg=None)`: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual


Should this be replacing line 704 instead of adding a new line on 705?
```suggestion
                                    tester.assertEqual(
                                        ctuple[2].strip("\r"),
                                        "label{}".format(expected0_sort_idx[b][idx]),
```

fpetrini15: Yes, good catch! I was updating this based on a text file of a patch and clearly blurred things together.

qa/L0_trt_plugin/test.sh
```diff
@@ -135,7 +135,7 @@ SERVER_LD_PRELOAD=$CUSTOMPLUGIN
 SERVER_ARGS=$SERVER_ARGS_BASE
 SERVER_LOG="./inference_server_$LOG_IDX.log"
 
-if  [[ ! -v WSL_DISTRO_NAME ]] || [[ ! -v MSYSTEM ]]; then
+if  [[ ! -v WSL_DISTRO_NAME ]] && [[ ! -v MSYSTEM ]]; then
```
rmccorm4, line 138: Not familiar with MSYSTEM - but what's the scenario where one is set and not the other?

mc-nv: Variable `WSL_DISTRO_NAME` applicable to WSL environment only. <br>
Variable `MSYSTEM` is applicable for GNU on Windows. 

Afraid condition will be always `false`

fpetrini15: My understanding is that we use `MINGW64` as our shell interpreter for Windows, which _should_ always set `MYSYSTEM`. In fact, if you examine our runners, `MYSYSTEM` has a value, however, `WSL_DISTRO_NAME` does not. I'm not sure of a scenario where the opposite would be true and `WSL_DISTRO_NAME` has a value but `MSYSTEM` does not. Perhaps @mc-nv knows?

fpetrini15: The original intention of this if-condition was to skip the enclosed test on a Windows system. On a Linux system, neither variable will be set, so the enclosed test will be run. On a Windows system, we anticipate that at least one of these will be set, ensuring that we skip the test.

qa/L0_batcher/test.sh
```diff
@@ -85,6 +85,7 @@ if [[ -v WSL_DISTRO_NAME ]] || [[ -v MSYSTEM ]]; then
     BACKEND_DIR=${BACKEND_DIR:=C:/tritonserver/backends}
     SERVER=${SERVER:=/mnt/c/tritonserver/bin/tritonserver.exe}
     export WSLENV=$WSLENV:TRITONSERVER_DELAY_SCHEDULER
+    export USE_GRPC=1
```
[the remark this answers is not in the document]
rmccorm4: Can you leave a comment above it for future reference? I think it's too cryptic on it's own to understand why it's there later.

qa/L0_trt_plugin/test.sh
rmccorm4: nit: In that case, can we either:

(a) match the if statement at the top for consistency?
```bash
if [[ -v WSL_DISTRO_NAME ]] || [[ -v MSYSTEM ]]; then
  echo "Skip this section on Windows"
else
  ...
fi
```

(b) or set some syntactic sugar naming ex:
```bash
IS_WINDOWS=...

if [ IS_WINDOWS ] 
# or
if [ ! IS_WINDOWS ]
```

rmccorm4 (approved):
🚀 MTTLG

PR #7839, opened 2024-11-27
feat: ORCA Format KV Cache Utilization in Inference Response Header

BenjaminBraunDev:
#### What does the PR do?
This PR adds code to `HTTPAPIServer::GenerateRequestClass::StartResponse` inside `src/http_server.cc` to add both `kv_cache_utilization` and `max_token_capacity` metrics composed from the existing prometheus metrics in TensorRT-LLM Backend's `nv_trt_llm_kv_cache_block_metrics` metric family.

This is acomplished by parsing the serialized prometheus metrics text object provided to the Triton Sever frontend by the Triton Core libraries into a structured vector of metrics for a specific metric family.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

Changes are contained to 2 files:
 - `src/http_server.cc`
 - `src/http_server.h` (the former's header file)

The changes start in `HTTPAPIServer::GenerateRequestClass::StartResponse()` where the request header is checked and the response header is written. There are 3 other functions in `orca_http.cc`: `MetricFamilyExtractor()` which parses serialized prometheus metrics into a vector of `PromMetric` (which have a map of their metric labels), `and ExtractKVMetrics()` that pulls the values from the structured metrics and calculates the composite kv metrics, and finally `OrcaKVMetricHeader()` which forms the metrics into an `endpoint-load-metrics` header in the ORCA format specified in the `endpoint-load-metrics-format` request header. If there are no TensorRT-LLM Backend metrics, no metrics found for the header, or an invalid format type the header is simply not written.

The valid values for `endpoint-load-metrics-format` are documented in the feature request (related issue linked below) and comments in `StartResponse()`

#### Test plan:
The feature won't do anything unless the inference request has the `endpoint-load-metrics-format` header. If not present, the feature is effectively disabled. Beyond that, the changes have been manually tested to not cause issue if either the queried metrics are not present (such as if TensorRT-LLM is not being used as the backend), or if the ORCA header metric type is invalid. In either case, nothing is parsed and no header is written. All code changes are wrapped in an `#ifdef` and are only included if metrics are enabled during the Triton Server build.

#### Caveats:
1. This feature only works on Triton Inference Server running with TensorRT-LLM Backend, as otherwise the KV-cache metrics are not included in the server metrics.

2. This change only implements the kv-cache utilization metics, but the functions it adds allows other metrics to be added easily (including metrics that don't require TensorRT-LLM Backend).

#### Background
[This doc](https://docs.google.com/document/d/18VRJ2ufZmAwBZ2jArfvGjQGaWtsQtAP6_yF2Xn6zcms/edit?tab=t.0#heading=h.sw2xdf66jh6) captures the overall requirements for model servers to integrate with llm instance gateway. More details in the Feature Request below.

#### Related Issues:
- Resolves Feature Request [GitHub issue: #7865](https://github.com/triton-inference-server/server/issues/7865)

#### Screenshots
Response header before changes (or if `endpoint-load-metrics-format` request header is not present):
orca_before

Response header with request header `endpoint-load-metrics-format: json`:
orca_json

Response header with request header `endpoint-load-metrics-format: text`:
orca_http

cc @yinggeh @krishung5 @jbkyang-nvi


src/http_server.cc
```diff
@@ -3225,6 +3225,42 @@ HTTPAPIServer::HandleGenerate(
       req, RestrictedCategory::INFERENCE, restricted_apis_);
 
   AddContentTypeHeader(req, "application/json");
+#ifdef TRITON_ENABLE_METRICS
+  // logic to add kv_cache metrics to response header
+  // Get the metrics in Prometheus format
+  if (std::getenv("ORCA_HEADER_METRIC_TYPE") != nullptr) {
```
liu-cong, line 3231: Change the name of the env var to `ORCA_METRIC_FORMAT`, and pls document the usage of it, also link the ORCA proposal doc here for reference

src/http_server.cc
```diff
@@ -3225,6 +3225,42 @@ HTTPAPIServer::HandleGenerate(
...
+            ExtractKVMetrics(kv_utilization, orca_type);
+        if (!extracted_kv_metrics.empty()) {
+          evhtp_headers_add_header(
+              req->headers_out,
+              evhtp_header_new(
+                  "endpoint-load-metrics", extracted_kv_metrics.c_str(), 1, 1));
+        }
+      }
+    }
+    TRITONSERVER_MetricsDelete(metrics);
```
liu-cong, line 3255: why deleting the metric?

src/http_server.cc
```diff
@@ -3225,6 +3225,42 @@ HTTPAPIServer::HandleGenerate(
...
+    const std::string orca_type = std::getenv("ORCA_HEADER_METRIC_TYPE");
+    TRITONSERVER_Metrics* metrics = nullptr;
+    TRITONSERVER_Error* err =
+        TRITONSERVER_ServerMetrics(server_.get(), &metrics);
+    if (err == nullptr) {
+      const char* base;
+      size_t byte_size;
+      err = TRITONSERVER_MetricsFormatted(
+          metrics, TRITONSERVER_METRIC_PROMETHEUS, &base, &byte_size);
+      if (err == nullptr) {
```
liu-cong, line 3241: can we log an error if err is not null?

src/http_server.cc
```diff
@@ -3225,6 +3225,42 @@ HTTPAPIServer::HandleGenerate(
...
+        std::string kv_utilization(base, byte_size);
+        // Extract the KV utilization metrics from the Prometheus formatted
+        // string.
+        std::string extracted_kv_metrics =
+            ExtractKVMetrics(kv_utilization, orca_type);
+        if (!extracted_kv_metrics.empty()) {
```
liu-cong, line 3247: can we print a log if metric is empty?

src/http_server.cc
```diff
@@ -3381,6 +3417,120 @@ HTTPAPIServer::HandleGenerate(
...
 }
 
+#ifdef TRITON_ENABLE_METRICS
+std::vector<HTTPAPIServer::PromMetric>
+HTTPAPIServer::MetricFamilyExtractor(
+    const std::string& input, const std::string& metricFamily)
+{
+  std::vector<PromMetric> metrics;
+  // Construct the regex pattern using the provided metricFamily
+  std::string patternStr = metricFamily + R"((?:{(.*?)})?\s+(\d+(?:\.\d+)?))";
```
liu-cong, line 3427: Explain how the regex works. Given an example input and output

src/http_server.cc
```diff
@@ -3381,6 +3417,120 @@ HTTPAPIServer::HandleGenerate(
...
+  std::string labelString;
+  std::string metric_value;
+
+  while (re2::RE2::FindAndConsume(
+      &inputPiece, pattern, &labelString, &metric_value)) {
+    PromMetric metric;
+
+    // Extract labels if they exist
+    if (!labelString.empty()) {
+      re2::RE2 labelPattern(R"((\w+)=\"([^\"]+)\")");
```
liu-cong, line 3440: Explain the regex

src/http_server.cc
```diff
@@ -3381,6 +3417,120 @@ HTTPAPIServer::HandleGenerate(
...
+
+  // Calculate derived metrics
+  double kv_cache_utilization = 0;
+  if (max_blocks > 0) {
+    kv_cache_utilization = used_blocks / max_blocks;
+  }
+  uint64_t max_token_capacity =
+      static_cast<uint64_t>(max_blocks * tokens_per_block);
+
+  // Logic to construct and format response header
```
liu-cong, line 3495: Consider making the rest helper functions, something like AddOrcaMetricJson(name, value) and AddOrcaMetricText(name, value)

src/http_server.cc
BenjaminBraunDev: The [TRITONSERVER_ServerMetrics()](https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L3256) call returns a [TritonServerMetrics](https://github.com/triton-inference-server/core/blob/main/src/tritonserver.cc#L166) object that we use to get the serialized metrics. The [caller is responsible for calling delete](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonserver.h#L2577) on it once they're done using it; it [doesn't actually delete the metric](https://github.com/triton-inference-server/core/blob/main/include/triton/core/tritonserver.h#L656) just the metrics object.

src/http_server.cc
BenjaminBraunDev: We could, however this will always be the case if the server backend is not TensorRT-LLM. My goal was to make it unintrusive if the backend metrics weren't present since Triton can use different backends, however it could be argued that if the feature flag environment variable is set we are expecting the metrics to be present, and thus it should be fine to treat this as a warning state and print a log. I can add logging here similar to how I log an error if the environment variable is set to an invalid type.

src/http_server.cc
BenjaminBraunDev: Usage documented in related [feature request issue](https://github.com/triton-inference-server/server/issues/7865) and in code comments along with link.

src/http_server.cc
BenjaminBraunDev: It is [logged below](https://github.com/triton-inference-server/server/pull/7839/commits/74492a8b2d2838b3ce06d1bdb60e117941503f9e#diff-2d773636d85b98207338699c3cc2b9fc7d31f38e45131ca127dca73f45d4bbb2R3258), but it is confusingly written so I moved the the delete metric function below everything and made it and if/else instead of a double if.

src/http_server.h
```diff
@@ -558,7 +566,17 @@ class HTTPAPIServer : public HTTPServer {
   void HandleGenerate(
       evhtp_request_t* req, const std::string& model_name,
       const std::string& model_version_str, bool streaming);
-
+#ifdef TRITON_ENABLE_METRICS
+  // Helper function to set get the KV-cache utilization metrics for the
+  // infer response header
```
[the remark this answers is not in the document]
BenjaminBraunDev: Done.

src/http_server.h
```diff
@@ -558,7 +566,17 @@ class HTTPAPIServer : public HTTPServer {
   void HandleGenerate(
       evhtp_request_t* req, const std::string& model_name,
       const std::string& model_version_str, bool streaming);
-
+#ifdef TRITON_ENABLE_METRICS
+  // Helper function to set get the KV-cache utilization metrics for the
```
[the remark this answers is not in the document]
BenjaminBraunDev: Done.

src/http_server.cc
BenjaminBraunDev: Broke this regex down into into named variables and explained in comments with example.

src/http_server.cc
BenjaminBraunDev: Explained in comments with example.

src/http_server.cc
BenjaminBraunDev: This is a bit awkward due to how the fields are nested, but I broke out the header string creation process into a separate function that takes in the reporting format and calculated KV metrics. This separates out responsibility and makes it more readable.

src/http_server.cc
coolkp: Can you add a log for feature enabled with non compatible environments "Ignoring ORCA_METRIC_FORMAT, backend X not supported" Only log this once, not part of this, at the entrypoint of server setup

src/http_server.cc
coolkp: + also consider a seperate a library like https://github.com/triton-inference-server/server/blob/r24.10/src/vertex_ai_server.cc

src/http_server.cc
BenjaminBraunDev: I'm not sure if there's a simple way to get the backend from the frontend without just checking the metrics and seeing if the tensorrt ones are present (there is a "model" parameter which could be set to TensorRT as that's what llama was provisioned with, but in many cases it will be hidden as "ensamble" as there are many layers to the model one of which is tensorrt-llm). I added such that if ORCA_METRIC_FORMAT is set but tensorrt metrics aren't available, it will log an error describing that, though it's at inference time when we get the metrics.

jbkyang-nvi:
@nnshah1 @indrajit96 are we merging into the already frozen `r24.10`? Or should this be onto `main` instead?

nnshah1:
> @nnshah1 @indrajit96 are we merging into the already frozen `r24.10`? Or should this be onto `main` instead?

good catch we should target main

BenjaminBraunDev:
@jbkyang-nvi @nnshah1 Is Ubuntu 24.04 required to build the main branch, with container version 24.11? I ported my changes over but am having trouble building the server image. I see in the nvidia container support matrix Ubuntu 24.04 is required for CUDA Deep-Learning base container for 24.11, but I also see in the Release 2.52.0 Known Issues that TRT-LLM backend is built from TensorRT-LLM version 0.15.0 and built out of [nvcr.io/nvidia/tritonserver:24.10-py3-min](http://nvcr.io/nvidia/tritonserver:24.10-py3-min).

To build, I'm using a base image of `nvcr.io/nvidia/tritonserver:24.10-trtllm-python-py3` as a base image (ran into errors using 24.11 base image) but I'm getting an error while building target `grpc-repo` without much of an error log. Is this the right base image or should I update my Ubuntu and and use 24.11?

For context the full build command I'm using is:
```
TRTLLM_BASE_IMAGE=nvcr.io/nvidia/tritonserver:24.10-trtllm-python-py3
./build.py -v --no-container-interactive --enable-logging --enable-stats --enable-tracing \
              --enable-metrics --enable-gpu-metrics --enable-cpu-metrics \
              --filesystem=gcs \
              --endpoint=http --endpoint=grpc --endpoint=sagemaker --endpoint=vertex-ai \
              --backend=ensemble --enable-gpu --no-container-pull \
              --repoagent=checksum --cache=local --cache=redis \
              --image=base,${TRTLLM_BASE_IMAGE}
```

BenjaminBraunDev:
Follow up, I got the image built, but when trying to run it I get a version error:
```
/opt/tritonserver/bin/tritonserver: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.36' not found (required by /opt/tritonserver/bin/tritonserver)
/opt/tritonserver/bin/tritonserver: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /opt/tritonserver/bin/tritonserver)
/opt/tritonserver/bin/tritonserver: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /opt/tritonserver/bin/tritonserver)
/opt/tritonserver/bin/tritonserver: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /opt/tritonserver/bin/../lib/libtritonserver.so)
/opt/tritonserver/bin/tritonserver: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found (required by /opt/tritonserver/bin/../lib/libtritonserver.so)
```

From what I looked up I don't think I can get `GLIBC_2.36` on Ubuntu 22.04?

nnshah1:
@nv-kmcgill53 , @nvda-mesharma 

After the holiday can we help with updated instructions on building? I think we are moving everything to 24.04 - but may be mistaken.

BenjaminBraunDev:
NOTE: The force push here is moving the contents of my forked main branch -> forked r24.10 branch. The head branch name `r24.10` is a misnomer, in reality it's the base `main` branch + my changes.

BenjaminBraunDev:
I've added a manual test to the qa common folder that can be run so long as triton is running on some port. For example, if you run the tritonserver container, something like this would work:

```
export ORCA_PATH=/triton-inference-server-repo/qa/common/orca_header_test.py
docker run --rm -ti -v `pwd`:/mnt -w /mnt -v ~/.cache/huggingface:/root/.cache/huggingface -v ${ORCA_PATH}:/orca_header_test.py  --gpus all tritonserver:latest bash

python3 /orca_header_test.py http://localhost:8000/v2/models/ensemble/generate
```

Let me know if you have any more feedback for the PR. @yinggeh @indrajit96 @nnshah1 @jbkyang-nvi

rmccorm4:
Closes #7865

BenjaminBraunDev:
Added a testing script for functionality, could you take a look for review/approval? @yinggeh @indrajit96 @nnshah1 @jbkyang-nvi 

cc: @liu-cong @coolkp

qa/common/orca_header_test.py
```diff
@@ -0,0 +1,133 @@
...
+    try:
+        if orca_format == "json":
+            # Parse the header in JSON format
+            data = json.loads(header.replace("JSON ", ""))
+            if METRIC_KEY in data:
+                return data[METRIC_KEY]
+            else:
+                print(f"No key '{METRIC_KEY}' in header data: {data}")
+                return None
+            return data
```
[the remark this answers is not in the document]
BenjaminBraunDev: Fixed

src/http_server.cc
```diff
@@ -4238,6 +4239,62 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+  // "json"
+  //
+  // Any other value will have behavior equivalent to being unset while also
+  // logging and error.
+  //
+  // For specifics on the different formats for the load reporting formats, see:
+  // https://docs.google.com/document/d/1C1ybMmDKJIVlrbOLbywhu9iRYo4rilR-cT50OTtOFTs/edit?tab=t.0#heading=h.do9yfa1wlpk8
+  auto server = infer_request->EvHtpServer();
+  if (std::getenv("TRITON_ORCA_METRIC_FORMAT") != nullptr && server != nullptr) {
+    const std::string orca_type = std::getenv("TRITON_ORCA_METRIC_FORMAT");
```
indrajit96, line 4263: Can we have a macro for TRITON_ORCA_METRIC_FORMAT

src/http_server.cc
```diff
@@ -4238,6 +4239,62 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+        std::string formatted_metrics(base, byte_size);
+        // Extract the KV utilization metrics from the Prometheus formatted
+        // string.
+        std::string extracted_kv_metrics =
+            ExtractKVMetrics(formatted_metrics, orca_type);
+        if (!extracted_kv_metrics.empty()) {
+          evhtp_headers_add_header(
+              req->headers_out,
+              evhtp_header_new(
+                  "endpoint-load-metrics", extracted_kv_metrics.c_str(), 1, 1));
```
indrajit96, line 4281: Can we have a macro for endpoint-load-metrics

src/http_server.cc
```diff
@@ -4247,6 +4304,173 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+  }
+
+  return metrics;
+}
+
+std::string
+HTTPAPIServer::GenerateRequestClass::ExtractKVMetrics(
+    const std::string& prometheus_metrics, const std::string& orca_type)
+{
+  std::string metric_family = "nv_trt_llm_kv_cache_block_metrics";
```
indrajit96, line 4391: Replace text with macro

src/http_server.cc
```diff
@@ -4247,6 +4304,173 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+  std::vector<PromMetric> kv_cache_metrics =
+      MetricFamilyExtractor(prometheus_metrics, metric_family);
+
+  double tokens_per_block = -1;
+  double used_blocks = -1;
+  double max_blocks = -1;
+
+  for (const auto& metric : kv_cache_metrics) {
+    if (metric.labels.count("kv_cache_block_type") > 0) {
```
indrajit96, line 4400: Replace text with macro

src/http_server.cc
```diff
@@ -4247,6 +4304,173 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+      std::string type = metric.labels.at("kv_cache_block_type");
```
indrajit96, line 4401: Replace text with macro

qa/common/orca_header_test.py
```diff
@@ -0,0 +1,132 @@
...
+import os
+import requests
+
+# To run the test, have tritonserver running and run this script with the endpoint as a flag.
+#
+# Example:
+# ```
+# python3 orca_header_test.py http://localhost:8000/v2/models/ensemble/generate
+# ```
+def get_endpoint_header(url, data):
```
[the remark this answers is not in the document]
BenjaminBraunDev: [line 122](https://github.com/triton-inference-server/server/pull/7839/files/9fd10984136035a9e3ef382b49edec1dd5659e14#diff-063d73ca7270e71cd4806fbb656189f847c0f211c68dd3f7dde5b2e0886d3f47R122) checks if the header is empty and if the orca is set it will complain, but if it's not set then it will simply state is it unset as expected.

src/http_server.h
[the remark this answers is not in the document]
BenjaminBraunDev: We are planning to add this later once we get http and json working.

qa/common/orca_header_test.py
```diff
@@ -0,0 +1,132 @@
...
+        return False
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Make a POST request to generate endpoint to test the ORCA metrics header.")
+    parser.add_argument("url", help="The model URL to send the request to.")
+    args = parser.parse_args()
+
+    TEST_DATA = json.loads('{"text_input": "hello world", "max_tokens": 20, "bad_words": "", "stop_words": ""}')
+    ORCA_ENVVAR = "TRITON_ORCA_METRIC_FORMAT"
+    orca_format = os.environ.get(ORCA_ENVVAR, "")
```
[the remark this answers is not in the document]
BenjaminBraunDev: That's a good idea, making those changes now.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
```diff
@@ -4247,6 +4304,173 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+
+  // Calculate derived metrics
+  double kv_cache_utilization = 0;
+  if (max_blocks > 0) {
+    kv_cache_utilization = used_blocks / max_blocks;
+  }
+  uint64_t max_token_capacity =
+      static_cast<uint64_t>(max_blocks * tokens_per_block);
+
+  return OrcaKVMetricHeader(
```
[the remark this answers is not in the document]
BenjaminBraunDev: Done.

qa/common/orca_header_test.py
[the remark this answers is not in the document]
BenjaminBraunDev: Done, updated python test to reflect this and test all valid formats.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
```diff
@@ -4247,6 +4318,174 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+      MetricFamilyExtractor(prometheus_metrics, metric_family);
+
+  double tokens_per_block = -1;
+  double used_blocks = -1;
+  double max_blocks = -1;
+
+  for (const auto& metric : kv_cache_metrics) {
+    if (metric.labels.count(KV_CACHE_BLOCK_TYPE) > 0) {
+      std::string type = metric.labels.at(KV_CACHE_BLOCK_TYPE);
+      if (type == "tokens_per") {
```
indrajit96, line 4416: Replace with Macro

src/http_server.cc
```diff
@@ -4247,6 +4318,174 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+        tokens_per_block = metric.value;
+      } else if (type == "used") {
```
indrajit96, line 4418: Replace with Macro

src/http_server.cc
```diff
@@ -4247,6 +4318,174 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+        used_blocks = metric.value;
+      } else if (type == "max") {
```
indrajit96, line 4420: Replace with Macro

indrajit96 (commented):
LGTM.
Please fix the nit-picks
Meanwhile I will start a CI pipeline based on these changes.
Will update once the CI passes so we can merge.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
BenjaminBraunDev: Done.

src/http_server.cc
BenjaminBraunDev: Done.

BenjaminBraunDev:
> LGTM. Please fix the nit-picks Meanwhile I will start a CI pipeline based on these changes. Will update once the CI passes so we can merge.

Sounds good, fixed the nits and tested everything. Let me know when the pipeline finishes.

BenjaminBraunDev:
If the pipeline has finished successfully, can this get approval for merge? @indrajit96

qa/common/orca_header_test.py
```diff
@@ -0,0 +1,148 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+
+sys.path.append("../common")
+
+import argparse
+import json
+import os
```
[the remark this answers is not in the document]
BenjaminBraunDev: Removed

qa/common/orca_header_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+def request_header(orca_format):
+    return {"endpoint-load-metrics-type": orca_format} if orca_format else None
+
+def test_header_type(url, data, orca_format):
+    req_header = request_header(orca_format)
+    response_header = get_endpoint_header(args.url, TEST_DATA, req_header)
+
+    desired_keys = {"kv_cache_utilization", "max_token_capacity"}  # Just the keys, no need to initialize with None
+    passed = None
```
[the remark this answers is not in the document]
BenjaminBraunDev: Removed

qa/common/orca_header_test.py
rmccorm4, line 1: @indrajit96 let's make sure to remove this `qa/common` file if moving it into it's own test `qa/L0_orca` in your PR: https://github.com/triton-inference-server/server/pull/8009

src/http_server.cc
```diff
@@ -4238,6 +4249,69 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+      const char* base;
+      size_t byte_size;
+      err = TRITONSERVER_MetricsFormatted(
+          metrics, TRITONSERVER_METRIC_PROMETHEUS, &base, &byte_size);
+      if (err == nullptr) {
+        std::string formatted_metrics(base, byte_size);
+        // Extract the KV utilization metrics from the Prometheus formatted
+        // string.
+        std::string extracted_kv_metrics =
+            ExtractKVMetrics(formatted_metrics, orca_type);
```
rmccorm4, line 4293: @indrajit96 @kthui - getting all the metrics, serializing them, doing a few regex calls, etc - this seems like it might be a bit of a heavy operation on every request, probably more prevalent at high concurrency. Looks like it's only called on the first response, even in the streaming case. 

1. @indrajit96 @kthui Have we done any benchmarks to get a gauge of how much this would impact TTFT?
2. @indrajit96 Where are we documenting this feature and how to opt-in/out of it (HTTP/REST API markdown docs? metrics docs?)
3. Looks like if the request header isn't provided, this will always log an error "Failed to get KV metrics" at the start of every request's first response? So if user doesn't opt-in with this header, will their server output get spammed with an error on every request?

src/http_server.cc
```diff
@@ -4247,6 +4321,174 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+      std::string type = metric.labels.at(KV_CACHE_BLOCK_TYPE);
+      if (type == KV_CACHE_BLOCK_TYPE_TOKENS_PER) {
+        tokens_per_block = metric.value;
+      } else if (type == KV_CACHE_BLOCK_TYPE_USED) {
+        used_blocks = metric.value;
+      } else if (type == KV_CACHE_BLOCK_TYPE_MAX) {
+        max_blocks = metric.value;
+      }
+    }
+  }
```
rmccorm4, line 4427: @indrajit96 @krishung5  The TRTLLM backend supports returning these per-response stats about the KV Cache as outputs already, right? https://github.com/triton-inference-server/tensorrtllm_backend/blob/565008d551867282e99e5a8dba57263f4da26ec2/all_models/inflight_batcher_llm/tensorrt_llm/config.pbtxt#L495-L509

Can we just grab the already prepared output tensors instead of parsing the metrics?

rmccorm4: If so - does this feature generalize to "allow returning select outputs as json headers"? User provides a list of outputs they want to receive in a header, and the name of the header they want to receive it in - and the rest is generic logic, not specific to ORCA?

rmccorm4: Or can the triton client be a middle man that prepares an HTTP response converting output tensors to ORCA HTTP header? I'm not currently seeing why the ORCA specific logic needs to be embedded directly in the HTTP Server code like this, rather than something either done on a client side, or a more generic server-side feature.

CC @GuanLuo

src/http_server.cc
indrajit96: @rmccorm4 I think perf wise we should be OK as it's guarded by a custom header in the http request and this being an "OPT-IN" feature.
Currently we do not have any benchmarks on impact for TTFT.
I will add a PR for the docs changes on how to opt-in for the ORCA metrics.
@BenjaminBraunDev can you get the TTFT numbers WITH and WITHOUT the ENDPOINT_LOAD_METRICS_TYPE header?
You can use
https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/perf_analyzer/genai-perf/README.html to get TTFT metrics.

src/http_server.cc
indrajit96: From my understanding
1. ORCA metrics are derived metrics from the ones TRT-LLM exposes hence forwarding them as-is won't help.
https://github.com/triton-inference-server/server/pull/7839/files#diff-2d773636d85b98207338699c3cc2b9fc7d31f38e45131ca127dca73f45d4bbb2R4435
Another alternative could be we could add the metrics needed by ORCA directly into our backend and then forward them as is.
2. The end goal of ORCA is to help with lead balnacing with GCP 
"GCP ALB backends are expected to provide a per-request load report using the ORCA format, either as an HTTP response header or if using gRPC, in trailer metadata"
I am not sure how we can use http client to achieve this in a performant way.
CC @BenjaminBraunDev @nnshah1

rmccorm4: If custom logic is needed, something like a customizable shared library approach might make sense following existing patterns in Triton. How we do Custom batching might be a good minimal example for such shared library pattern: https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/user_guide/batcher.html#custom-batching

src/http_server.cc
```diff
@@ -4247,6 +4321,174 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+
+    named_metrics.AddDouble(kv_util_key.c_str(), metrics[kv_util_key]);
+    named_metrics.AddUInt(max_token_key.c_str(), metrics[max_token_key]);
+    orca_metrics.Add(named_metrics_key.c_str(), std::move(named_metrics));
+
+    triton::common::TritonJson::WriteBuffer buffer;
+    orca_metrics.Write(&buffer);
+    header_contents = std::string("JSON ") + buffer.Contents();
+
+  } else if (orca_type == "http") {
```
liu-cong, line 4474: I suggest renaming this to 'text', or support both. "TEXT" is what will be set as the prefix to the metrics in ORCA standard 

https://docs.google.com/document/d/1C1ybMmDKJIVlrbOLbywhu9iRYo4rilR-cT50OTtOFTs/edit?tab=t.0#heading=h.bo9zqknd00wx

src/http_server.cc
liu-cong: Agree with both of you. While the current implementation is not the most performant, it's guarded by the client. It's more important to enable this so the client can try this feature. And this helps getting feedback on whether it's important to further optimize it.

src/http_server.cc
krishung5: > The TRTLLM backend supports returning these per-response stats about the KV Cache as outputs already, right?

Correct, as long as the [`return_kv_cache_reuse_stats`](https://github.com/triton-inference-server/tensorrtllm_backend/blob/565008d551867282e99e5a8dba57263f4da26ec2/all_models/inflight_batcher_llm/tensorrt_llm/config.pbtxt#L279) is set in the inputs, it will return those metrics as outputs.

> something like a customizable shared library approach might make sense following existing patterns in Triton

I think this would also give the flexibility when more ORCA metrics need to be added, without the need to update the hard-coded logic.

src/http_server.cc
BenjaminBraunDev: Done.

BenjaminBraunDev:
If there are no issues, can we get this merged so we are able to try it with our load balancing? Feedback from that process should efficiently guide further optimization (e.g. benchmarking if there is any significant latency when trying to pull the metrics, etc.) @indrajit96 @rmccorm4 @krishung5

liu-cong:
Thank you @BenjaminBraunDev . The change looks good to me, and looking forward to trying this out!

indrajit96:
Hi @BenjaminBraunDev can we have a refactor in the existing PR?
Can we move all ORCA related changes from http_server.cc to a new file orca_http.cc orca_http.h create a new class in these files and make sure that http_server.cc and http_server.h have minimal changes?
Appreciate your help and patience in this.
We can merge post this.

BenjaminBraunDev:
I have refactored the changes by adding the `orca_http.cc` and `orca_http.h` files. `http_server.cc` only has the logic that checks the request header while the rest of the logic for extraction is in the orca files. Functionality is the same, tested both manually and with the test script. Let me know if this looks good to you and if there are any further changes, thank you. @indrajit96 @rmccorm4 @krishung5

BenjaminBraunDev:
Refactor applied, could you take a look at this?
@indrajit96 @rmccorm4 @krishung5

src/http_server.cc
```diff
@@ -4238,6 +4239,69 @@ HTTPAPIServer::GenerateRequestClass::StartResponse(
...
+  auto server = infer_request->EvHtpServer();
+  const char* orca_metric_format = nullptr;
+  evhtp_header_t* metric_format_header =
+      evhtp_headers_find_header(req->headers_in, ENDPOINT_LOAD_METRICS_TYPE);
+
+  if (metric_format_header != nullptr) {
+    orca_metric_format = metric_format_header->val;
+  }
+  if (orca_metric_format != nullptr && server != nullptr) {
+    const std::string orca_type = orca_metric_format;
```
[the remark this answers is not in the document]
BenjaminBraunDev: done

BenjaminBraunDev:
Finished last encapsulation change, let me know if there's any further feedback.
@indrajit96 @rmccorm4 @krishung5

indrajit96:
LGTM, will approve and merge pending CI pipeline.
Thanks a lot for this @BenjaminBraunDev !!

indrajit96 (approved):
LGTM, pre-commit and CI passes

rmccorm4 (approved):
Thanks for the contribution @BenjaminBraunDev !

I'm going to merge this since @indrajit96 verified CI passed.

@indrajit96 please do the needful to make sure CI and jobs like L0_build_variants look good after merge. Also please follow up to remove the qa/common test file added here if it's redundant in favor of the tests you added in a separate PR.

PR #7842, opened 2024-11-28
build: enabled build behind proxy

dtrawins:
#### What does the PR do?
Adds support to build behind proxy by mount docker/config.json into the container running the build.

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:
run build.py with .docker/config.json - confirm successful build behind proxy
run build.py without .docker/config.json   - confirm build fails with apt install error

- closes GitHub issue: #xxx


build.py
```diff
@@ -1772,6 +1772,8 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
             runargs += ["-v", "\\\\.\pipe\docker_engine:\\\\.\pipe\docker_engine"]
         else:
             runargs += ["-v", "/var/run/docker.sock:/var/run/docker.sock"]
+            if os.path.exists(os.path.expanduser("~/.docker/config.json")):
+                runargs += ["-v", os.path.expanduser("~/.docker/config.json:/root/.docker/config.json")]
```
mc-nv, line 1776: I don't really like this change. 
It's kind of we allow to add something without validation. 
We have no control of `~/.docker/config.json` entries there why we may be forced to review some false positive issues related to user environment.

I may need a bit more context, sound like user looking for a proxy config during the build process, and I may be mistaken here but the issue happen when we run Docker-out-of-Docker configuration using `build.py` and user is behind the proxy.

I suppose it's better to add a flag for users `--use-user-docker-config` and add value accordingly if flag is given.

nnshah1: I agree a flag may be more appropriate here to be clear on the intent

build.py
```diff
@@ -1234,7 +1234,7 @@ def create_dockerfile_linux(
 
 WORKDIR /opt/tritonserver
 COPY --chown=1000:1000 NVIDIA_Deep_Learning_Container_License.pdf .
-
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
```
mc-nv, line 1237: This line should be outdated. 
This change should cover it
https://github.com/triton-inference-server/server/pull/7876

dtrawins:
change already merged

PR #7855, opened 2024-12-06
build: Adding b64 dependency to relevant targets (fix L0_build_variants)

KrishnanPrash:
#### What does the PR do?
After #7787, L0_build_variants started failing with the following error message:
```
...
=== Building wheel
=== Output wheel file is in: /workspace/build/tritonserver/build/triton-server/python/generic
[ 98%] Built target frontend-server-wheel
[100%] Linking CXX executable tritonserver
/usr/bin/ld: CMakeFiles/main.dir/common.cc.o: in function `triton::server::DecodeBase64(char const*, unsigned long, std::vector<char, std::allocator<char> >&, unsigned long&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
common.cc:(.text+0x76a): undefined reference to `base64_init_decodestate'
/usr/bin/ld: common.cc:(.text+0x77b): undefined reference to `base64_decode_block'
collect2: error: ld returned 1 exit status
gmake[5]: *** [CMakeFiles/main.dir/build.make:205: tritonserver] Error 1
gmake[4]: *** [CMakeFiles/Makefile2:411: CMakeFiles/main.dir/all] Error 2
gmake[3]: *** [Makefile:136: all] Error 2
gmake[2]: *** [CMakeFiles/triton-server.dir/build.make:86: triton-server/src/triton-server-stamp/triton-server-build] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:192: CMakeFiles/triton-server.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
error: build failed
***
*** No-Endpoint/No-Filesystem FAILED
***  
```
If we comment out the `tritonfrontend` wheel building, we get the following error message:
```
...
[ 89%] Linking CXX executable tritonserver
/usr/bin/ld: CMakeFiles/main.dir/common.cc.o: in function `triton::server::DecodeBase64(char const*, unsigned long, std::vector<char, std::allocator<char> >&, unsigned long&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/server/src/common.cc:125: undefined reference to `base64_init_decodestate'
/usr/bin/ld: /server/src/common.cc:128: undefined reference to `base64_decode_block'
collect2: error: ld returned 1 exit status
make[5]: *** [CMakeFiles/main.dir/build.make:205: tritonserver] Error 1
make[4]: *** [CMakeFiles/Makefile2:359: CMakeFiles/main.dir/all] Error 2
make[3]: *** [Makefile:136: all] Error 2
make[2]: *** [CMakeFiles/triton-server.dir/build.make:86: triton-server/src/triton-server-stamp/triton-server-build] Error 2
make[1]: *** [CMakeFiles/Makefile2:193: CMakeFiles/triton-server.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

With the b64 dependency being introduced to `common.h`, it needs to be linked to whatever executable/library/object that uses `common.cc/common.h`. Hence, this PR links the b64 dependency to the tritonserver and py-bindings targets that both directly add link the b64 dependency.

#### Related PRs:
#7787

- CI Pipeline ID: 21154897

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -97,12 +97,10 @@ add_executable(
   main
   classification.cc
   command_line_parser.cc
-  common.cc
   main.cc
   shared_memory_manager.cc
   triton_signal.cc
   classification.h
-  common.h
   shared_memory_manager.h
   triton_signal.h
 )
@@ -590,6 +588,42 @@ if(${TRITON_ENABLE_TRACING})
       tracing-library
   )
 endif() # TRITON_ENABLE_TRACING
+
+# Common Library
+add_library(common_obj OBJECT common.cc)
+
+if(${TRITON_ENABLE_GPU})
+  target_compile_definitions(
+      common_obj
+      PRIVATE TRITON_ENABLE_GPU=1
+  )
+endif()
+
+
+if (WIN32)
+  find_library(B64_LIBRARY NAMES b64)
+  target_link_libraries(
+      common_obj
+      PUBLIC
+        ${B64_LIBRARY}
+        triton-core-serverapi  # from repo-core
+  )
+else()
+  target_link_libraries(
+      common_obj
+      PUBLIC
+        b64
+        triton-core-serverapi  # from repo-core
+    )
+endif()
+
+set_target_properties(
+    common_obj
+    PROPERTIES
+      POSITION_INDEPENDENT_CODE ON
+)
+
+target_link_libraries(main PRIVATE common_obj)
 
 if (NOT WIN32)
   #
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -97,10 +97,12 @@ add_executable(
   main
   classification.cc
   command_line_parser.cc
+  common.cc
   main.cc
   shared_memory_manager.cc
   triton_signal.cc
   classification.h
+  common.h
   shared_memory_manager.h
   triton_signal.h
 )
@@ -116,7 +118,7 @@ if (NOT WIN32)
 endif()
 
 target_compile_features(main PRIVATE cxx_std_${TRITON_MIN_CXX_STANDARD})
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+if(WIN32)
   message("Using MSVC as compiler, default target on Windows 10. "
     "If the target system is not Windows 10, please update _WIN32_WINNT "
     "to corresponding value.")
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -591,6 +589,42 @@ if(${TRITON_ENABLE_TRACING})
   )
 endif() # TRITON_ENABLE_TRACING
 
+# Common Library
+add_library(common_obj OBJECT common.cc)
```
fpetrini15, line 593: Where is `common.h`?

GuanLuo (dismissed):

src/CMakeLists.txt
```diff
@@ -591,6 +589,57 @@ if(${TRITON_ENABLE_TRACING})
   )
 endif() # TRITON_ENABLE_TRACING
 
+# Common Library
+add_library(common_obj OBJECT common.h common.cc)
+
+target_compile_features(common_obj PRIVATE cxx_std_${TRITON_MIN_CXX_STANDARD})
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
nv-kmcgill53, line 596: For consistency, 
```suggestion
if (WIN32)
```

src/CMakeLists.txt
```diff
@@ -591,6 +589,57 @@ if(${TRITON_ENABLE_TRACING})
...
+
+if(${TRITON_ENABLE_GPU})
+  target_compile_definitions(
+      common_obj
+      PRIVATE TRITON_ENABLE_GPU=1
+  )
+endif()
+
+
+if (WIN32)
```
nv-kmcgill53, line 618: Is there a reason why we can't merge these two if-else block together?

nv-kmcgill53 (changes requested):
> With the b64 dependency being introduced to common.h, it needs to be linked to whatever executable/library/object that uses common.cc/common.h. 

Why can't we just add the b64 lib in the compiler command line? Something like adding `-lb64` or something when the lib or executable is getting linked? This would save us from having to do this complicated common library stuff, no?

fpetrini15:
> Why can't we just add the b64 lib in the compiler command line? Something like adding -lb64 or something when the lib or executable is getting linked? This would save us from having to do this complicated common library stuff, no?

@nv-kmcgill53 I had suggested to Prasanth to make this it's own object since common owned the dependency. I figured that down the line it would help developers trace where the dependency of b64 is coming from--much like we do with the endpoint libraries. Open to your thoughts here.

nv-kmcgill53:
I guess I don't quite see how this block of cmake makes it clearer where the b64 dependency is coming from and who is using it. If it's a library linked on the command line then it's clear that lib is needed to build everything in the repository. If the b64 lib is needed to build a different part of the repository then it's more intuitive to add the lib as a dependency to that part of the code rather than having this intermediary object get created. 

From another angle, if we were to scale your reasoning out from this single instance of `common.h` to all files of this repository, we would end up explicitly saying which libs are needed for every source file. We would then construct the command line for every source file and build independently. This is much more complicated then declaring the lib to be used for the entire repository.

Hopefully I am hearing you right, if not let's get together offline for higher bandwidth

fpetrini15:
> We would then construct the command line for every source file and build independently. This is much more complicated then declaring the lib to be used for the entire repository.

Yeah this convinced me. Thanks for the wisdom 🙇

nv-kmcgill53 (approved):
This looks a lot better.

fpetrini15 (approved):
LGTM

PR #7859, opened 2024-12-06
ci: RHEL8 L0_backend_python Support

fpetrini15:
#### What does the PR do?

Support the L0_backend_python test for our RHEL8 asset.

#### Checklist

#### Commit Type:
- [ ] build
- [x] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/python_backend/pull/389
https://github.com/triton-inference-server/core/pull/418

- CI Pipeline ID:

21140314


qa/L0_backend_python/env/test.sh
```diff
@@ -116,9 +119,14 @@ conda deactivate
...
 create_conda_env "3.12" "python-3-12"
 conda install -c conda-forge libstdcxx-ng=14 -y
+TF_VERSION="2.16.2"
 conda install numpy=1.26.4 -y
-conda install tensorflow=2.16.2 -y
-PY312_VERSION_STRING="Python version is 3.12, NumPy version is 1.26.4, and Tensorflow version is 2.16.2"
+if [ $TRITON_RHEL -eq 1 ]; then
+    TF_VERSION="2.17.0"
+fi
+conda search -f tensorflow
```
nv-kmcgill53, line 127: What is the point of this search?

qa/L0_backend_python/env/test.sh
nv-kmcgill53, line 1: I see you didn't absorb My changes from https://github.com/triton-inference-server/server/pull/7838 which means there is still a silent failure in job 126856344 where the locale is still not getting passed through correctly for one of the python versions. 

The issue in this sub-test is that the assertion attempts to find one line of `Locale` being set when it should be testing that 4 lines of `Locale` are being set in the logs. If you go back to the Ci job and search on `Locale is`, you will find some tests with 4 print outs and some with 3. As listed in this sub-test, https://github.com/triton-inference-server/python_backend/pull/260 gives the reason for pinning the pybind version in the first place. My investigation showed that the silent failure started on this PR: https://github.com/triton-inference-server/python_backend/pull/384. So there is a regression in the way we use pybind11 since the 24.10 release.

qa/L0_backend_python/env/test.sh
fpetrini15: This was an artifact left behind when I was trying to confirm that version of tensorflow we were looking for was not available. I will remove it, thanks.

qa/L0_backend_python/env/test.sh
fpetrini15: I can run a test locally to see about the other subtests, but seeing 3/4 `Locale` printouts sounds like it could be related to the fact that we delete one of the models in the last run: https://github.com/triton-inference-server/server/blob/main/qa/L0_backend_python/env/test.sh#L243-L245, leaving a model for python3.6, 3.7.1, and 3.12. Could it be related to what you are describing?

fpetrini15: I ran the test locally and you're right. I don't see the `Locale is ('en_US', 'UTF-8')` printout for Python3.12

nv-kmcgill53: This failure is captured in DLIS-7725 and will be inspected later. This failure also doesn't block this PR as discussed offline.

nv-kmcgill53 (approved):

PR #7874, opened 2024-12-12
fix: Handle TRT-LLM models with max_batch_size of zero in OpenAI frontend

rmccorm4:
Pass model config read at load time through to check max_batch_size when forming requests.

Currently, the conversion logic creates inputs of shape `[-1, 1]` for most inputs, which assumes a `max_batch_size > 0`. If `max_batch_size == 0`, then the inputs should be of shape `[1]` without the extra batch dimension, based on the current reference model configs of TRT-LLM models.

In the future, this could probably be made more advanced to actually double check the shapes and presence of each input when converting from OpenAI -> Triton request format, but for now it will be kept more lightweight for simplicity until needed.

Before:
```
# Model with max_batch_size == 0
$  curl http://localhost:9000/v1/chat/completions ...
{"detail":"[request id: <id_unknown>] unexpected shape for input 'temperature' for model 'mock'. Expected [1], got [1,1]. "}
```

After:
```
# Model with max_batch_size == 0
$  curl http://localhost:9000/v1/chat/completions ...
{"id":"cmpl-4e98a5c2-b820-11ef-8bf9-04d4c4933ecf","choices":[{"finish_reason":"stop","index":0,"message":{"content":"..."}], ...}
```

### DIFF
--- a/python/openai/openai_frontend/engine/triton_engine.py
+++ b/python/openai/openai_frontend/engine/triton_engine.py
@@ -68,6 +68,8 @@ class TritonModelMetadata:
     backend: str
     # Triton model object handle
     model: tritonserver.Model
+    # Triton model config object
+    config: dict
     # Tokenizers used for chat templates
     tokenizer: Optional[Any]
     # Time that model was loaded by Triton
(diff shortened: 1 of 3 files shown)


rmccorm4:
Closing as it appears that TRTLLM doesn't support setting max batch size to 0 out of the box currently, and I don't want to introduce this shape check/fix to the hot path if not needed.

PR #7879, opened 2024-12-13
refactor: Update the response queue in the server to reuse response slots

pskiran1:
#### What does the PR do?

The current response queue allocates memory for each response.
This PR aims to enhance the response queue by reusing response slots across multiple responses within the same request once they have been written (completed) to the network. This may help to reduce the active memory utilization.
- In the `PopResponse()` function, we clear the response content and return it to the reusable pool.
- In the `AllocateResponse()` function, we check for any available responses in the reusable pool; if present, we use it, otherwise, we allocate a new response.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [x] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 21564131

- closes GitHub issue: #xxx

### DIFF
--- a/src/grpc/infer_handler.h
+++ b/src/grpc/infer_handler.h
@@ -1,4 +1,4 @@
-// Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2023-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -102,9 +102,9 @@ struct RequestReleasePayload final {
 //
 // ResponseQueue
 //
-// A simple queue holding the responses to be written. Uses a
-// vector of persistent message objects to prevent allocating
-// memory for each response to be written.
+// This class implements a queue to manage responses that need to be written.
+// It internally uses a reusable pool of persistent message objects to avoid
+// allocating memory for each response individually.
 //
 template <typename ResponseType>
 class ResponseQueue {
@@ -113,19 +113,29 @@ class ResponseQueue {
 
   ~ResponseQueue()
   {
+    // Delete all responses in the reusable pool
+    for (auto response : reusable_pool_) {
+      delete response;
+    }
+
+    // Delete all responses currently in the queue
     for (auto response : responses_) {
       delete response;
     }
   }
 
-  // Resets the queue
+  // Resets the queue to its initial state
   void Reset()
   {
+    std::lock_guard<std::mutex> lock(mtx_);
     alloc_count_ = 0;
     ready_count_ = 0;
-    current_index_ = 0;
-    for (auto response : responses_) {
-      response->Clear();
+    pop_count_ = 0;
+
+    while (!responses_.empty()) {
+      responses_.front()->Clear();
+      reusable_pool_.push_back(responses_.front());
+      responses_.pop_front();
     }
   }
 
@@ -137,17 +147,27 @@ class ResponseQueue {
     std::lock_guard<std::mutex> lock(mtx_);
     alloc_count_ = 1;
     if (responses_.size() < 1) {
-      responses_.push_back(new ResponseType());
+      if (!reusable_pool_.empty()) {
+        responses_.push_back(reusable_pool_.front());
+        reusable_pool_.pop_front();
+      } else {
+        responses_.push_back(new ResponseType());
+      }
     }
     return responses_[0];
   }
 
-  // Allocates a response on the head of the queue
+  // Allocates a response at the end of the queue
   void AllocateResponse()
   {
     std::lock_guard<std::mutex> lock(mtx_);
     alloc_count_++;
-    if (responses_.size() < alloc_count_) {
+
+    // Use a response from the reusable pool if available
+    if (!reusable_pool_.empty()) {
+      responses_.push_back(reusable_pool_.front());
+      reusable_pool_.pop_front();
+    } else {
       responses_.push_back(new ResponseType());
     }
   }
@@ -156,12 +176,15 @@ class ResponseQueue {
   ResponseType* GetLastAllocatedResponse()
   {
     std::lock_guard<std::mutex> lock(mtx_);
-    if (responses_.size() < alloc_count_) {
+
+    // Ensure that the requested response has been allocated
+    if ((responses_.size() + pop_count_) < alloc_count_) {
       LOG_ERROR
           << "[INTERNAL] Attempting to access the response not yet allocated";
       return nullptr;
     }
-    return responses_[alloc_count_ - 1];
+
+    return responses_.back();
   }
 
   // Marks the next non-ready response complete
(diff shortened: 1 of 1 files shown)


src/grpc/infer_handler.h
```diff
@@ -178,64 +201,93 @@ class ResponseQueue {
...
+      return;
+    }
+
+    // Clear and move the current response to the reusable pool
+    auto response = responses_.front();
+    response->Clear();
+    reusable_pool_.push_back(response);
+    responses_.pop_front();
+    pop_count_++;
   }
```
kthui, line 260: Just checking my understanding:
Previously, we just increment the `current_index_`/`pop_count_` when a response is popped from the `responses_` queue, while the `index` always refer to the actual location on the `responses_` queue. Now, a popped response is removed from the `responses_` queue, changing the index of the remaining elements, so an "offset"/`pop_count_` is necessary to correctly access the elements, given the callers are unaware of the change when providing the `index`.

I assume the response is fully written to gRPC and there will be no more access to the response object by the time `PopResponse()` is called? In other words, the response object can be safely reused/modified after it is popped? Given the previous implementation keeps the popped response objects intact, until the `~ResponseQueue()` is destructed.

pskiran1: Yes, @kthui. Once we call PopResponse(), the response object will no longer be needed.
cc: @tanmayv25

pskiran1:
The PR was automatically closed on forced rebase with the main, reopened with a new commit.

src/grpc/infer_handler.h
```diff
@@ -178,64 +207,97 @@ class ResponseQueue {
...
     std::lock_guard<std::mutex> lock(mtx_);
-    return (ready_count_ > current_index_);
+    return (ready_count_ > pop_count_);
   }
 
  private:
-  std::vector<ResponseType*> responses_;
+  // Stores responses that need to be written. The front of the queue indicates
+  // the current response, while the back indicates the last allocated response.
+  std::deque<ResponseType*> responses_;
```
indrajit96, line 290: Can we have smart pointers `unique_ptr` instead of raw pointers.
Just a suggestion if it's not a lot of work.

src/command_line_parser.cc
```diff
@@ -1438,6 +1444,14 @@ TritonParser::Parse(int argc, char** argv)
         case OPTION_GRPC_INFER_ALLOCATION_POOL_SIZE:
           lgrpc_options.infer_allocation_pool_size_ = ParseOption<int>(optarg);
           break;
+        case OPTION_GRPC_MAX_RESPONSE_POOL_SIZE:
+          lgrpc_options.max_response_pool_size_ = ParseOption<int>(optarg);
+          if (lgrpc_options.max_response_pool_size_ <= 0) {
```
indrajit96, line 1449: Do we also need to have an upper limit on the check?
Code below suggests the max is {INT_MAX};

qa/L0_decoupled/test.sh
```diff
@@ -127,6 +127,45 @@ for trial in $TRIALS; do
 
   kill $SERVER_PID
   wait $SERVER_PID
+
+  SERVER_ARGS="--model-repository=$MODELDIR --grpc-max-response-pool-size=1"
```
indrajit96, line 131: Can you add a test plan in the description on what we are trying to test here?
Why have we set `--grpc-max-response-pool-size` only to `1`?
Also can we add a test to confirm performance increase with and without using `--grpc-max-response-pool-size` ?
You can refer to `L0_perf_nomodel` to use PA.
Ideally we should see PA numbers improve when using `--grpc-max-response-pool-size`

indrajit96:
Can we also update docs for `--grpc-max-response-pool-size`

src/command_line_parser.cc
```diff
@@ -536,6 +537,11 @@ TritonParser::SetupOptions()
        "allocated for reuse. As long as the number of in-flight requests "
        "doesn't exceed this value there will be no allocation/deallocation of "
        "request/response objects."});
+  grpc_options_.push_back(
```
tanmayv25, line 540: Why do we need an extra argument for this? Why no reuse `"grpc-infer-allocation-pool-size` above?

pskiran1: @tanmayv25, We use the `grpc-infer-allocation-pool-size` parameter as the limit for state bucket and state reuse, with a default value of only 8. In contrast, for the response queue threshold, we need to have a higher maximum value to achieve better performance by default(current behavior).
If we apply the same option for the response queue threshold also, we must always specify the `grpc-infer-allocation-pool-size` to improve performance by default. However, a major drawback is that if this value is set too high, it will also affect the size of the state bucket, resulting in states not being deleted. For example, if we set it to 200, it will keep 200 states alive for reuse.
Please correct me if my understanding is wrong. Thank you.

pskiran1: Considering the above, it may be better to maintain a separate flag for controlling the response queue limit without affecting the state bucket.

src/grpc/infer_handler.h
pskiran1: As we discussed offline, the ResponseQueue class has an [explicit destructor](https://github.com/triton-inference-server/server/pull/7879/files#diff-49172908070e57597f28f4bd7feea976592a4d3f3675644e64630309cb3c6a3fL114-L119) that ensures all responses stored in the queue are deleted when the object is destroyed. And, when we reset the queue, we only need to clear its contents. Therefore, I believe we may not need a smart pointer for the queue.
Marking the thread as resolved for now, please feel free to unresolve it if needed. Thank you.

src/command_line_parser.cc
pskiran1: Thanks for the note. I verified and confirmed that the CLI parser performs limit checks and generates an error if it exceeds the upper limit.
```
Bad option: "--grpc-max-response-pool-size".
Provided option value is out of bound. Got 3000000000
````

qa/L0_decoupled/test.sh
pskiran1: > Also can we add a test to confirm memory footprint decreses with using --grpc-max-response-pool-size VS not using --grpc-max-response-pool-size ?

I have added a new test case in L0_memory to evaluate memory utilization when running the server with different values for `--grpc-max-response-pool-size` (1, 25, and 50), as well as without this flag.

> Why have we set --grpc-max-response-pool-size only to 1?

Regarding setting `--grpc-max-response-pool-size` to 1, I included this specific test to evaluate the lowest possible value. And, running the decoupled model tests takes a long time, with additional pool sizes it is leading to timeouts. The new test case covers different values.

pskiran1: Also, updated description with higher level details about tests. Please let me know if we are missing something.

pskiran1:
> Can we also update docs for `--grpc-max-response-pool-size`

I have updated the documentation to include details about both the `--grpc-infer-allocation-pool-size` and `--grpc-max-response-pool-size` options. @tanmayv25, please review and confirm if the content is correct. Thank you.

tanmayv25 (approved):

tanmayv25:
Nice work @pskiran1 !

PR #7882, opened 2024-12-14
build: Adding b64 dependency to relevant targets (fix L0_build_variants) (#7855)

yinggeh:
Cherry-pick https://github.com/triton-inference-server/server/pull/7855 into r24.12
Pipeline ID: 21450412

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -118,7 +118,7 @@ if (NOT WIN32)
 endif()
 
 target_compile_features(main PRIVATE cxx_std_${TRITON_MIN_CXX_STANDARD})
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+if(WIN32)
   message("Using MSVC as compiler, default target on Windows 10. "
     "If the target system is not Windows 10, please update _WIN32_WINNT "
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -130,13 +130,30 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
   target_compile_definitions(main
     PRIVATE
       NOMINMAX)
+
+  # Dependency from common.h
+  find_library(B64_LIBRARY NAMES b64)
```
mc-nv, line 135: What will happen if library not found?

mc-nv: Can we try to be consistent in our scenarios and make change similar to. 
https://github.com/triton-inference-server/perf_analyzer/commit/0c589f85dbab0b55d709cb8e2baf8fbfc27f7847

yinggeh: @KrishnanPrash

KrishnanPrash: If it's acceptable, would it be possible to get this in as a follow-up PR?

mc-nv (approved):

PR #7920, opened 2025-01-07
Update variable name

mc-nv:
This PR is fixing missed models generation.

### DIFF
--- a/qa/common/gen_qa_model_repository
+++ b/qa/common/gen_qa_model_repository
@@ -286,7 +286,7 @@ python3 $VOLUME_SRCDIR/gen_qa_dyna_sequence_implicit_models.py --onnx --onnx_ops
 chmod -R 777 $VOLUME_DYNASEQIMPLICITDESTDIR
 python3 $VOLUME_SRCDIR/gen_qa_ragged_models.py --onnx --onnx_opset=$ONNX_OPSET --models_dir=$VOLUME_RAGGEDDESTDIR
 chmod -R 777 $VOLUME_RAGGEDDESTDIR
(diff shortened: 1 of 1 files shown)


qa/common/gen_qa_model_repository
```diff
@@ -286,7 +286,7 @@ python3 $VOLUME_SRCDIR/gen_qa_dyna_sequence_implicit_models.py --onnx --onnx_ops
 chmod -R 777 $VOLUME_DYNASEQIMPLICITDESTDIR
 python3 $VOLUME_SRCDIR/gen_qa_ragged_models.py --onnx --onnx_opset=$ONNX_OPSET --models_dir=$VOLUME_RAGGEDDESTDIR
 chmod -R 777 $VOLUME_RAGGEDDESTDIR
-python3 $VOLUME_SRCDIR/gen_qa_ort_scalar_models.py --onnx_opset=$ONNX_OPSET --models_dir=$SCALARMODELSDESTDIR
+python3 $VOLUME_SRCDIR/gen_qa_ort_scalar_models.py --onnx_opset=$ONNX_OPSET --models_dir=$VOLUME_SCALARMODELSDESTDIR
 chmod -R 777 $VOLUME_RAGGEDDESTDIR
```
rmccorm4, line 290: Should the next `chmod` line 290 also be updated?
```
chmod -R 777 $VOLUME_SCALARMODELSDESTDIR
```

mc-nv: For consistency probably you are right.

rmccorm4 (approved):
It would be nice to update the scripts to catch these types of errors earlier with something like:
```python
if not FLAGS.models_dir:
    raise ValueError("--models_dir must be provided")
```

But it's fine to just do the fix if there's no bandwidth right now.

PR #7929, opened 2025-01-13
ci: Stabilize L0_pinned_memory flakiness

rmccorm4:
L0_pinned_memory fails intermittently due to PA unstable measurements that are not critical to the test functionality - this is not a performance test.

This PR adds the use of `--request-count` in PA to remove PA stability criteria and make L0_pinned_memory test more consistent

---

In reality, we probably don't need to be using PA for this test, and can swap it out with a simpler client that doesn't measure performance.

For simplicity, we can just relax the stability requirements on PA to avoid this error with minimal changes for now.

### DIFF
--- a/qa/L0_pinned_memory/test.sh
+++ b/qa/L0_pinned_memory/test.sh
@@ -38,8 +38,9 @@ if [ ! -z "$TEST_REPO_ARCH" ]; then
     REPO_VERSION=${REPO_VERSION}_${TEST_REPO_ARCH}
 fi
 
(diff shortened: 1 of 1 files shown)


indrajit96 (approved):
LGTM, pending pipeline.

rmccorm4:
Jobs passing on multiple re-runs. Made a small change to request count number to reduce test time a bit. Should be faster and pass more consistently now.

yinggeh (approved):

PR #7930, opened 2025-01-13
ci: Fix error-masking bug and improve debugability in L0_trace

rmccorm4:
Changes:
- Remove invalid check for trace_mode after expected error that always produced `RET=1` because the `curl.out` only contained `{"error": ...}` but we checked for `trace_mode` count to be non-zero
- Remove RET=0 reset in middle of script - this was masking the error above
- Add 'cat curl.out' to various places that check the contents and may fail for easier visibility to error in logs. I had to repro the first 2 bullets locally to understand what was happening, so adding this to dump out the `curl` responses so we can better determine what


oandreeva-nv (approved):
lgtm, conditional on ci

rmccorm4:
Job passed ✅

PR #7933, opened 2025-01-14
ci: Fix L0_lifecycle server shutdown

kthui:
#### What does the PR do?
There are four tests in L0_lifecycle that the server shutdown is triggered by the py test, which may not return by the time the server completes its shutdown, causing the failure that the test script is trying to shutdown a stopped server.

#### Checklist

#### Commit Type:
- [ ] build
- [x] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

N/A

#### Test plan:
L0_lifecycle should pass.

- CI Pipeline ID: 22497034

#### Caveats:
N/A

#### Background
N/A

N/A

### DIFF
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2018-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -1576,7 +1576,7 @@ if [ `grep -c "Model 'custom_zero_1_float32' (version 1) has 1 in-flight inferen
     RET=1
 fi
 
-kill $SERVER_PID
+kill $SERVER_PID || true
 wait $SERVER_PID
 
 rm -f $CLIENT_LOG
@@ -1614,7 +1614,7 @@ if [ `grep -c "Model 'custom_sequence_int32' (version 1) has 1 in-flight inferen
     RET=1
 fi
 
-kill $SERVER_PID
+kill $SERVER_PID || true
 wait $SERVER_PID
 
 rm -f $CLIENT_LOG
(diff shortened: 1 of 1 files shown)


qa/L0_lifecycle/test.sh
rmccorm4, line 1: > There are four tests in L0_lifecycle that the server shutdown is triggered by the py test, **which may not return by the time the server completes its shutdown**, causing the failure that the test script is trying to shutdown a stopped server.

Aren't the errors in the job logs the opposite? I thought I saw that the test script was trying to kill the process but the process was already dead?
```
kill $SERVER_PID
# error: no process found
```

In other words - are the processed supposed to be already dead before `test.sh` calls `kill $SERVER_PID`? Or is there some kind of crash happening during a test that causes the server process to exit before `tesh.sh` tries to kill it?

kthui: > are the processed supposed to be already dead before test.sh calls kill $SERVER_PID?

yes

for example:
1. test.sh: start the server
2. test.sh: run py test on lifecycle_test.py
3. lifecycle_test.py: [stop the server](https://github.com/triton-inference-server/server/blob/4adcffff4522dcecc3598bedd55dc6fabd17e6df/qa/L0_lifecycle/lifecycle_test.py#L2724)
4. lifecycle_test.py: test return
5. test.sh: [stop the server again](https://github.com/triton-inference-server/server/blob/4adcffff4522dcecc3598bedd55dc6fabd17e6df/qa/L0_lifecycle/test.sh#L1617)

rmccorm4: Got it - thanks for highlighting the test cases that are killing the server. That makes it a bit confusing to track when it should be dead/alive in the bash script - but this should be a fine WAR for now.

rmccorm4 (approved):

qa/L0_lifecycle/test.sh
kthui: A "cleaner" alternative is to remove the `kill $SERVER_PID...` from those four tests, but if the py test failed before the server is killed, the `wait $SERVER_PID` could block forever, which prevents testing the remaining test cases, so I think it would be easier for now to keep the `kill $SERVER_PID` while allowing it to fail for the four test cases.

PR #7940, opened 2025-01-15
fix: Fix L0_implicit_state and it's variants

indrajit96:
#### What does the PR do?
Update the config file with `max_sequence_idle_microseconds` to avoid sequence timeouts in server

#### Checklist
- [] Related issues are referenced.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [X] test

#### Related PRs:
NA

Config file

#### Test plan:
NA

- CI Pipeline ID:
https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/22507783

#### Background
Server waits for `max_sequence_idle_microseconds` before closing the request sequence and expects the client to start a NEW sequence for ANY subsequent request.
Test was failing because the `max_sequence_idle_microseconds` was set to low causing the server to believe the sequence is over and reject any further requests with SAME `sequence_id` without the `START` flag. Causing the test to fail.


qa/L0_implicit_state/test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2021-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2021-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 2: Don't update copyright if file isn't modified

qa/L0_implicit_state/models/growable_memory/config.pbtxt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 1: Copyright 2023-2025

rmccorm4 (commented):
LGTM, just copyright nits, thanks for the fix!

Also, please expand on the bug and the fix more in the description to educate future readers. 

Ex: if sequence times out mid sequence, it will interpret a request as the start of a new sequence, but client won't have set the START flag because it thinks it's still in the middle of a sequence, so an error will be raised -- but in nicer words

qa/L0_implicit_state/models/growable_memory/config.pbtxt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright (c) 2023 - 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 1:
```suggestion
# Copyright 2023-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

rmccorm4: needs to be a specific format to not exceed 80 char line length

indrajit96: Pre-Commit hook doesn't like 2023 - 2025 it replaced it with 2025.

rmccorm4: Even if you remove the space? `2023-2025`

yinggeh:
Please clean up the commit history.

rmccorm4:
The original change was so small, I'd probably just make a fresh branch for simplicity. I think that force push from Pavithra messed up the history.

indrajit96:
> The original change was so small, I'd probably just make a fresh branch for simplicity. I think that force push from Pavithra messed up the history.

Yup :)

PR #7941, opened 2025-01-15
fix: Fix L0_implicit_state and it's variants

indrajit96:
#### What does the PR do?
Update the config file with `max_sequence_idle_microseconds` to avoid sequence timeouts in server

#### Checklist
- [] Related issues are referenced.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [X] test

#### Related PRs:
NA

Config file

#### Test plan:
NA

- CI Pipeline ID:
https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/22507783

#### Background
Server waits for `max_sequence_idle_microseconds` before closing the request sequence and expects the client to start a NEW sequence for ANY subsequent request.
Test was failing because the `max_sequence_idle_microseconds` was set to low causing the server to believe the sequence is over and reject any further requests with SAME `sequence_id` without the `START` flag. Causing the test to fail.

### DIFF
--- a/qa/L0_implicit_state/models/growable_memory/config.pbtxt
+++ b/qa/L0_implicit_state/models/growable_memory/config.pbtxt
@@ -28,6 +28,7 @@ name: "growable_memory"
 backend: "implicit_state"
 max_batch_size: 0
 sequence_batching {
+  # Set large idle timeout to avoid inter-request timeouts for test consistency
   max_sequence_idle_microseconds: 10000000
   control_input [
     {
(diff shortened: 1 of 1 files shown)


qa/L0_implicit_state/models/growable_memory/config.pbtxt
```diff
@@ -28,6 +28,7 @@ name: "growable_memory"
 backend: "implicit_state"
 max_batch_size: 0
 sequence_batching {
+  max_sequence_idle_microseconds: 10000000
```
yinggeh, line 31: Why 10 sec? Is there a delayed request in the test?

indrajit96: Currently the test uses `max_sequence_idle_microseconds: 1000000` causing it the be flaky.
RCA explained in PR description.
To make test reliable increasing from 1 to 10 secs

rmccorm4, line 31:
```suggestion
  # Set large idle timeout to avoid inter-request timeouts for test consistency 
  max_sequence_idle_microseconds: 10000000
```

rmccorm4 (approved):

PR #7945, opened 2025-01-16
build: Downgrade ORT to 1.19.2 until OpenVINO NPU EP issue is resolved

rmccorm4:
Downgrade to working version (1.19.2 used in r24.12) for stable support of ONNX Runtime OpenVino Execution Provider until we address the issues with NPU support.

### DIFF
--- a/build.py
+++ b/build.py
@@ -74,7 +74,7 @@ DEFAULT_TRITON_VERSION_MAP = {
     "release_version": "2.54.0dev",
     "triton_container_version": "25.01dev",
     "upstream_container_version": "24.12",
-    "ort_version": "1.20.1",
+    "ort_version": "1.19.2",
     "ort_openvino_version": "2024.4.0",
     "standalone_openvino_version": "2024.4.0",
     "dcgm_version": "3.3.6",


yinggeh:
I am fine with downgrading ORT verison but need @mc-nv to confirm.

mc-nv:
I"m afraid that in `r24.12` we been using version `1.20.1`: 
```
https://github.com/triton-inference-server/server/blob/main/build.py#L77
```
Container data: 
```
$ docker run --rm -it nvcr.io/nvidia/tritonserver:24.12-py3 readelf -V /opt/tritonserver/backends/onnxruntime/libonnxruntime.so  | grep gnu.version_d -A3
Version definition section '.gnu.version_d' contains 2 entries:
 Addr: 0x00000000000058f0  Offset: 0x000058f0  Link: 4 (.dynstr)
  000000: Rev: 1  Flags: BASE  Index: 1  Cnt: 1  Name: libonnxruntime.so.1
  0x001c: Rev: 1  Flags: none  Index: 2  Cnt: 1  Name: VERS_1.20.1
```

rmccorm4:
> Believe our fix for this issue was using of previous version of the models based on OpenVINO 2023.3
https://github.com/triton-inference-server/server/blob/r24.12/qa/common/gen_qa_model_repository#L58

This is a runtime issue and not a model generation issue. It passed with the previous versions in build.py for [r24.11](https://github.com/triton-inference-server/server/blob/5815d218664f277e866146fb05ef0cd9913883d4/build.py#L76-L77) (confirmed with a new pipeline 22573279 for this PR as well that all ONNX jobs pass), but failed when we updated ORT to 1.20 in [r24.12](https://github.com/triton-inference-server/server/blob/1344e8f410122b2621c79dc7356a8b5f8c8fa353/build.py#L77-L78) as well. 

I'm happy to keep/upgrade the ORT version if we can fix the OpenVino EP issues - but I think it's silly to keep it broken if we don't have clear bandwidth to fix it.

rmccorm4:
Closing for now since r24.12 was released with this issue - can revisit.

PR #7965, opened 2025-01-24
ci: Pin awscli version due to botocore incompatibility to fix L0_storage_swiftstack

rmccorm4:
Fix `L0_storage_swiftstack` due to [botocore 1.36 incompatibility](https://github.com/boto/boto3/issues/4392). The solution is to pin `awscli<=1.36.40` to make sure we use `botocore<1.36` or else we'd run into the following error when uploading to 3rd party s3 clone:
```
An error occurred (InvalidArgument) when calling the PutObject operation: x-amz-content-sha256 must be UNSIGNED-PAYLOAD, or a valid sha256 value
```


indrajit96 (approved):
LGTM!! contingent on passing pipeline

rmccorm4:
Test passed in pipeline 22959162

PR #7969, opened 2025-01-27
feat: Add graceful shutdown timer to GRPC frontend

mattwittwer:
#### What does the PR do?

This PR adds a shutdown timer to the gRPC endpoint for both infer and streaming infer requests. Inflight requests will be allowed to complete before the timer expires and new requests made after shutdown has started will be rejected.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

N/A

Start with the new functions in src/grpc/grpc_server.cc and the flow in src/main.cc
GracefulStop();
unload the models in core 
Stop();

#### Test plan:

Review L0_lifecycle and grpc related tests. 
Updated shutdown behavior changes the tested responses for some lifecycle tests

- CI Pipeline ID: 27245783

#### Caveats:

The updated shutdown process includes three responses. The CANCELLED state cannot be removed as it is part of the GRPC endpoint shutdown behavior:
 - "GRPC server is shutting down and has stopped accepting new requests."
 - "CANCELLED"
 - "failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:8001: Failed to connect to remote host: connect: Connection refused (111)"

#### Background

Shutdown behavior for the gRPC endpoint could lead to unexpected errors when unloading the models from the core

N/A


src/grpc/infer_handler.cc
```diff
@@ -753,14 +753,20 @@ ModelInferHandler::Process(
       StartNewRequest();
     }
 
-    if (ExecutePrecondition(state)) {
+    if (accepting_new_conn_ && ExecutePrecondition(state)) {
```
krishung5, line 756: Do we need a lock for `accepting_new_conn_`?

tanmayv25, line 756: Alongwith reading requests and returning error to the client, I would advice we also prevent calling 

```
service_->RequestModelInfer(
      state->context_->ctx_.get(), &state->request_,
      state->context_->responder_.get(), cq_, cq_, state);
```
in L671. 
The idea is that we should avoid additional activity on the completion queue once the server shutdown is detected. We should just go through the inflight requests and drain their state objects before exiting the server to prevent any memory leak.

> When the “Graceful shutdown function” is called, the server immediately notifies all clients to stop sending new RPCs. Then after the clients have received that notification, the server will stop accepting new RPCs.

Source: https://grpc.io/docs/guides/server-graceful-stop/

mattwittwer: Added lock: `conn_mtx_`
For both `accepting_new_conn_` and `cq_shutdown_`

mattwittwer: Moved the forceful shutdown until after the core unloads. This prevents responses from returning from the core once the forceful shutdown has initiated. Completion queues are unloaded only after the forceful shutdown is complete

src/main.cc
```diff
@@ -324,11 +335,9 @@ StopEndpoints()
 {
   bool ret = true;
 
-  // TODO: Add support for 'exit_timeout_secs' to the endpoints below and move
-  // them to the 'StopEndpoints(uint32_t* exit_timeout_secs)' function above.
-
 #ifdef TRITON_ENABLE_GRPC
   if (g_grpc_service) {
+    // Forceful shutdown of GRPC service
```
krishung5, line 340: Question - why would we still need the forceful shutdown of GRPC while we have the graceful shutdown?

src/grpc/infer_handler.cc
```diff
@@ -681,12 +681,14 @@ ModelInferHandler::Process(
...
 {
   // There are multiple handlers registered in the gRPC service.
-  // Hence, there we can have a case where a handler thread is
+  // Hence, we can have a case where a handler thread is
   // making progress in the state machine for a request and the
   // other thread is issuing cancellation on the same request.
   // Need to protect the state transitions for these cases.
   std::lock_guard<std::recursive_mutex> lock(state->step_mtx_);
 
+  std::shared_lock<std::shared_mutex> lk1(*conn_mtx_);
```
krishung5, line 690: This mutex is for `accepting_new_conn_`, right? Could we move it closer to the accepting_new_conn_` section?

src/main.cc
mattwittwer: The graceful shutdown `GracefulStop()` rejects any new GRPC requests but does not stop ongoing requests. It waits until the timeout completes and then returns but the GRPC server may still exist. Unloading the triton server core first prevents writing back to unloaded states in the GRPC server. 

The forceful shutdown in `Stop()` then cancels any remaining requests and stops the GRPC server if it hasn't stopped already.

src/grpc/infer_handler.cc
mattwittwer: Yes, updated the conn_mtx_ location

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2018-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
indrajit96, line 3: As discussed can we keep the testing logic of existing tests as is and ADD new test for the graceful shutdown.

indrajit96 (commented):
1. Let's create a TEP for this as discussed.
2. Test a simple no-model perf with and without changes to see if any perf regression was introduced

src/grpc/grpc_server.h
```diff
@@ -153,6 +160,11 @@ class Server {
 
   int bound_port_{0};
   bool running_{false};
+
+  std::thread graceful_shutdown_thread;
+  std::shared_mutex conn_mtx_;
```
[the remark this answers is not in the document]
mattwittwer: Updated to include comments explaining the new variables

src/grpc/infer_handler.h
```diff
@@ -1324,6 +1329,8 @@ class InferHandler : public HandlerBase {
       const std::shared_ptr<StateContext>& context,
       Steps start_step = Steps::START)
   {
+    IncrementConnectionCount();
+
```
indrajit96, line 1333: CC @pskiran1 
IRRC, state creation was enhanced by you recently.
Can you have a look at the PR as well

src/grpc/grpc_server.cc
```diff
@@ -2569,9 +2596,14 @@ Server::Stop()
         TRITONSERVER_ERROR_UNAVAILABLE, "GRPC server is not running.");
   }
 
-  // Always shutdown the completion queue after the server.
-  server_->Shutdown();
+  // Forcefully cancel remaining RPC connections
+  server_->Shutdown(std::chrono::system_clock::now());
 
+  if (graceful_shutdown_thread.joinable()) {
```
pskiran1, line 2602: Can we modify it to `graceful_shutdown_thread_`?

src/grpc/infer_handler.cc
```diff
@@ -753,14 +753,23 @@ ModelInferHandler::Process(
       StartNewRequest();
     }
 
-    if (ExecutePrecondition(state)) {
+    std::shared_lock<std::shared_mutex> lk1(*conn_mtx_);
```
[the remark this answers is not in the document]
mattwittwer: Unlock added to reduce scope covered by lk1 here: https://github.com/triton-inference-server/server/blob/mwittwer/gRPC-shutdown-timer/src/grpc/infer_handler.cc#L793

qa/L0_lifecycle/lifecycle_test.py
mattwittwer: Added additional test case to separate branch here: https://github.com/triton-inference-server/server/commit/77921ab29f35cbd7ba2dbbc7b47587c79497b366 

Updated TEP with discussion on this test case and a pointer to the commit.

indrajit96 (approved):
LGTM, make sure you attach perf numbers to the ticket and the update the latest CI pipeline on the reviews thread.
Nice work!

PR #7970, opened 2025-01-27
build: Dockerfiles generated by build.py for RHEL require base image specification

saturley-hall:
#### What does the PR do?
Images built for RHEL which did not specify a base image defaulted to Ubuntu 24.04 which caused error during the docker build stage as the `yum` package manager was unavailable. This change causes `build.py` to error out with a message should the RHEL target be specified without a base image also being specified.

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

- CI Pipeline ID:

#### Caveats:

Specification of a RHEL base image does _not_ guarantee that the container will build as there are internal dependencies that may be missing.

### DIFF
--- a/build.py
+++ b/build.py
@@ -1644,7 +1644,9 @@ def create_build_dockerfiles(
     if "base" in images:
         base_image = images["base"]
         if target_platform() == "rhel":
-            print("warning: RHEL is not an officially supported target and you will probably experience errors attempting to build this container.")
+            print(
+                "warning: RHEL is not an officially supported target and you will probably experience errors attempting to build this container."
+            )
     elif target_platform() == "windows":
         base_image = "mcr.microsoft.com/dotnet/framework/sdk:4.8"
     elif target_platform() == "rhel":
--- a/build.py
+++ b/build.py
@@ -1645,6 +1645,10 @@ def create_build_dockerfiles(
         base_image = images["base"]
     elif target_platform() == "windows":
         base_image = "mcr.microsoft.com/dotnet/framework/sdk:4.8"
+    elif target_platform() == "rhel":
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -1643,8 +1643,15 @@ def create_build_dockerfiles(
 ):
     if "base" in images:
         base_image = images["base"]
+        if target_platform() == "rhel":
+            print(
+                "warning: RHEL is not an officially supported target and you will probably experience errors attempting to build this container."
+            )
     elif target_platform() == "windows":
         base_image = "mcr.microsoft.com/dotnet/framework/sdk:4.8"
```
mc-nv, line 1651: that's interesting: 
https://github.com/triton-inference-server/server/blob/main/Dockerfile.win10.min#L29

build.py
```diff
@@ -1643,8 +1643,15 @@ def create_build_dockerfiles(
...
+    elif target_platform() == "rhel":
+        # Disallow RHEL target unless base image is specified
```
mc-nv, line 1653: This comment is redundant.

mc-nv: I mean if you think the exception notice is not informative enough may be it better to add one more sentence?

saturley-hall:
```suggestion
```

build.py
mc-nv: No it's just an observation.

mc-nv (approved):

dmitrytokarev (approved):
LGTM

PR #8038, opened 2025-02-26
feat: Add multi-LoRA support to OpenAI frontend

kthui:
#### What does the PR do?
Add support for selecting LoRA to OpenAI frontend. The user can optionally provide the `--model-and-lora-name-separator=<lora_separator>` command line argument when starting the OpenAI frontend. If the argument is provided and non-empty, the OpenAI model name maybe `<triton_model_name><lora_separator><lora_name>` if the `<lora_separator>` is found in the OpenAI model name. If found, the LoRA parameters will be set to the `<lora_name>` during inference.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

Start with the test case, and then move on to the implementation.

#### Test plan:
A new test under the L0_openai is added. It will test for selecting LoRA with/without the `--model-and-lora-name-separator` argument set.

- CI Pipeline ID: 26950793

#### Caveats:
The implementation currently only support vLLM, TRT-LLM will be added in a follow-up PR.

#### Background
N/A

N/A


python/openai/openai_frontend/main.py
```diff
@@ -108,6 +108,12 @@ def parse_args():
         choices=["vllm", "tensorrtllm"],
         help="Manual override of Triton backend request format (inputs/output names) to use for inference",
     )
+    triton_group.add_argument(
+        "--model-and-lora-name-separator",
```
rmccorm4, line 112: `--lora-separator` for conciseness?

python/openai/tests_lora/lora_test.py
```diff
@@ -0,0 +1,198 @@
...
+            f.write(
+                json.dumps(
+                    {
+                        "model": "unsloth/gemma-2b",
+                        "enable_lora": True,
+                        "max_lora_rank": 32,
+                    }
+                )
+            )
+        with open(f"models/{self._model_name}/1/multi_lora.json", "w") as f:
```
rmccorm4, line 76: Can we attempt to read the `multi_lora.json` if found in OpenAI frontend to support listing available loras under the `{model}{separator}{lora}` scheme via `/v1/models`?

python/openai/openai_frontend/main.py
kthui: Sure, updated: [Rename --model-and-lora-name-separator to --lora-separator](https://github.com/triton-inference-server/server/pull/8038/commits/09a0868a2c1c000d1beee4f6fbfeb83090accd8c)

python/openai/tests_lora/server_wrapper.py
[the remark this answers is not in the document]
kthui: It was intended for keeping the new test separate from the existing tests, but it is possible to reuse the `class OpenAIServer` from the existing test: [Reuse OpenAIServer from tests](https://github.com/triton-inference-server/server/pull/8038/commits/b7897f682f63bc1216ee8fbe3622c27874cb065b)

python/openai/tests_lora/lora_test.py
kthui: Yes, we can add support for listing model with LoRA name via `v1/models`.

Can we do it on a separate PR, since this is an added scope from supporting LoRA inference?

rmccorm4: Separate PR works for me

python/openai/tests_lora/server_wrapper.py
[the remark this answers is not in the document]
rmccorm4: Yeah let's try to re-use it if possible. 

Does it need to be in a separate `lora_tests` folder? Can't you just have `tests/test_lora.py`?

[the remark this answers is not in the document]
kthui: This is interesting! The original goal of separating the tests was to avoid the LoRA tests from starting the [module scoped OpenAI server on the conftest.py](https://github.com/triton-inference-server/server/blob/jacky-openai-lora-vllm/python/openai/tests/conftest.py#L82-L101), which is incompatible with the LoRA tests. It turns out that the conftest.py will be ignored by pytest if the test_<something>.py does not import pytest, which is the case on the LoRA tests, so it is possible to keep the test_lora.py in the tests directory.

[Move LoRA tests to openai/tests](https://github.com/triton-inference-server/server/pull/8038/commits/31e889b2f2a4dac223e234abe2cf9e8b985aa0b6)

python/openai/tests/test_lora.py
```diff
@@ -0,0 +1,198 @@
...
+            client = server.get_client()
+            # Check selecting LoRA works
+            self._test_completions(client, "")
+            self._test_completions(client, "doll")
+            self._test_completions(client, "sheep")
+            self._test_chat_completion(client, "")
+            self._test_chat_completion(client, "doll")
+            self._test_chat_completion(client, "sheep")
+            # Check selecting unknown LoRA results in LoRA not found
+            # TODO: Server hangs when shutting down if LoRA not found.
```
krishung5, line 185: Will this be included in follow-up PRs, or is it within the scope of this one?

kthui: It is unlikely to be in scope of either PR, because this is either an issue in the vLLM backend or how the OpenAI frontend / Python binding interact with Triton core, but the follow-up PR may early terminate the inference if LoRA can be determined and the specified LoRA is not a part of the loaded LoRAs, allowing the test to avoid the problematic part and pass.

python/openai/tests_lora/server_wrapper.py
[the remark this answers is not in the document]
rmccorm4: >  It turns out that the conftest.py will be ignored by pytest if the test_<something>.py does not import pytest, which is the case on the LoRA tests, so it is possible to keep the test_lora.py in the same tests directory.

My understanding is that the `module="scope"` fixture should be called once per test module `test_{file}.py` that uses it. So given `tests/test_existing.py` and `tests/test_lora.py`, if both tests used that module scope fixture defined in conftest.py, it would be called twice, once per module.

Here's a simple example:
```bash
rmccormick@ced35d0-lcedt:~/triton/distributed/jacky_pytest$ cat conftest.py 
import pytest

@pytest.fixture(scope="module")
def scope_module():
    print("module scope fixture called!")

rmccormick@ced35d0-lcedt:~/triton/distributed/jacky_pytest$ cat test_mod1.py 
def test_foo(scope_module):
    print("Test foo in module1!")

rmccormick@ced35d0-lcedt:~/triton/distributed/jacky_pytest$ cat test_mod2.py 
def test_bar(scope_module):
    print("Test bar in module2!")
```
and

```
rmccormick@ced35d0-lcedt:~/triton/distributed/jacky_pytest$ pytest -s
========================================================================== test session starts ===========================================================================
platform linux -- Python 3.10.12, pytest-8.1.1, pluggy-1.4.0
rootdir: /home/rmccormick/triton/distributed/jacky_pytest
plugins: timeout-2.3.1, asyncio-0.23.8, anyio-3.7.1, cov-4.1.0, mock-3.14.0
asyncio: mode=strict
collected 2 items                                                                                                                                                        

test_mod1.py module scope fixture called!
Test foo in module1!
.
test_mod2.py module scope fixture called!
Test bar in module2!
.

=========================================================================== 2 passed in 0.01s ============================================================================
```

---

> The original goal of separating the tests was to avoid the LoRA tests from starting the [module scoped OpenAI server on the conftest.py](https://github.com/triton-inference-server/server/blob/jacky-openai-lora-vllm/python/openai/tests/conftest.py#L82-L101), which is incompatible with the LoRA tests.

Following the details above, simply having the the lora tests in their own module, and not using the fixture anywhere in that module, should achieve the desired goal, as you've done here.

Does this all match your expectations?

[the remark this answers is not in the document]
kthui: Thanks for pointing out an example! I think a `server` variable has likely find its way into my initial testing when trying out the pytest, which caused the impression that the default server is always started for all test classes in the tests directory. This matches my expectation given each individual test case has control over what is started before the test.

richardhuo-nv (approved):
LGTM! Thanks!

krishung5 (approved):

rmccorm4 (approved):

PR #8040, opened 2025-02-27
ci: Add OpenAI frontend testing for LLM API backend

krishung5:
#### What does the PR do?

Add OpenAI frontend testing for LLM API backend.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/python/openai/tests/conftest.py
+++ b/python/openai/tests/conftest.py
@@ -1,4 +1,4 @@
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -31,8 +31,10 @@ import pytest
 from fastapi.testclient import TestClient
 from tests.utils import OpenAIServer, setup_fastapi_app, setup_server
 
-
 ### TEST ENVIRONMENT SETUP ###
+LLMAPI_SETUP = os.environ.get("LLMAPI_SETUP", 0)
+
+
 def infer_test_environment():
     # Infer the test environment for simplicity in local dev/testing.
     try:
@@ -48,7 +50,10 @@ def infer_test_environment():
         import tensorrt_llm as _
 
         backend = "tensorrtllm"
-        model = "tensorrt_llm_bls"
+        if LLMAPI_SETUP:
+            model = "tensorrt_llm"
+        else:
+            model = "tensorrt_llm_bls"
         return backend, model
     except ImportError:
         print("No tensorrt_llm installation found.")
@@ -57,7 +62,10 @@ def infer_test_environment():
 
 
 def infer_test_model_repository(backend):
-    model_repository = str(Path(__file__).parent / f"{backend}_models")
+    if LLMAPI_SETUP:
+        model_repository = str(Path(__file__).parent / f"{backend}_llmapi_models")
+    else:
+        model_repository = str(Path(__file__).parent / f"{backend}_models")
     return model_repository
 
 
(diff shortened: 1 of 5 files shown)


richardhuo-nv (dismissed):
LGTM! Thanks!

python/openai/tests/conftest.py
```diff
@@ -48,7 +50,10 @@ def infer_test_environment():
         import tensorrt_llm as _
 
         backend = "tensorrtllm"
-        model = "tensorrt_llm_bls"
+        if LLMAPI_SETUP:
+            model = "tensorrt_llm"
+        else:
+            model = "tensorrt_llm_bls"
```
rmccorm4, line 56: Does it make sense to set `backend = "llmapi"` and use the `backend` fixture to check if llmapi in the tests, following similar patterns for vllm/trtllm in the existing tests?

krishung5: The config.pbtxt `backend` field will be `python` for LLM API backend.

I thought that the LLM API backend is one of the tensorrtllm backends, so hence didn't have another "llmapi backend" fixture. 

I think avoiding having an ENV makes more sense, however, right now the logic to determine which backend to use is to import tensorrtllm or vllm module and see if there's an import error, then set the backend and model name accordingly.
https://github.com/triton-inference-server/server/blob/205f13c13b0cbb8b0933b18a3cb3085048fb8c40/python/openai/tests/conftest.py#L47-L54

Both general tensorrtllm and llmapi backend cases will import the tensorrtllm module successfully. I think some sort of flags would still be required to know if this is a general trtllm model or a llmapi one.

rmccorm4: > The config.pbtxt backend field will be python for LLM API backend.

Do we plan to add a python-based backend like vllm is?

krishung5: Do you mean making it `backend: llmapi` in the config.pbtxt? I don't think there's a plan to do so right now. I could see it be less confusing having this after the pivot is more finalized, right now it's more like adding another python model just like the original one.

rmccorm4: Oh right because they're not using the `runtime` config.pbtxt feature currently, they just swapped the `backend` to python instead, gotcha - makes sense.

Will the llmapi `model.py` replace the existing `model.py` in trtllm backend? Hopefully we're not planning to support 3 implementations for now 😅

python/openai/tests/conftest.py
```diff
@@ -31,8 +31,10 @@
 from fastapi.testclient import TestClient
 from tests.utils import OpenAIServer, setup_fastapi_app, setup_server
 
-
 ### TEST ENVIRONMENT SETUP ###
+LLMAPI_SETUP = os.environ.get("LLMAPI_SETUP", 0)
```
rmccorm4, line 35: Would be nice to clean up some of the `if/else` everywhere for llmapi vs trtllm stuff, but hopefully we can consolidate the supported backend implementations once its a bit more mature.

rmccorm4 (approved):

python/openai/tests/conftest.py
krishung5: Agree - confirming w/ TRT-LLM team on this.

krishung5:
I'll be able to merge this PR only after the LLM API backend MR is merged. Also pending on CI - having some issues with the build which are not related to the openai changes.  I'll ask for final review once we're ready to merge. Thanks!

rmccorm4:
> I'll be able to merge this PR only after the LLM API backend MR is merged. Also pending on CI - having some issues with the build which are not related to the openai changes. I'll ask for final review once we're ready to merge. Thanks!

Marking as draft for clarity - feel free to unmark it when ready

PR #8052, opened 2025-03-04
feat: Add OpenAI frontend multi-LoRA model listing

kthui:
#### What does the PR do?
Add support for listing/retrieving vLLM models that has LoRA(s) specified at the backend to OpenAI frontend.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
Merging into https://github.com/triton-inference-server/server/pull/8038

Start with the related PR, and then look into the feature and then test on this PR.

#### Test plan:
The L0_openai_vllm test_lora.py is enhanced with new tests that list and retrieve models via OpenAI endpoint, and assert the listed and retrieved models are correct given the LoRA configuration.

- CI Pipeline ID: T.B.D.

#### Caveats:
N/A

#### Background
N/A

N/A


python/openai/tests/test_lora.py
```diff
@@ -91,10 +94,81 @@ def _create_model_repository(self):
...
+        os.makedirs(f"models/{self._model_name}/1", exist_ok=True)
+        with open(f"models/{self._model_name}/config.pbtxt", "w") as f:
+            f.write('backend: "vllm"')
+        with open(f"models/{self._model_name}/1/model.json", "w") as f:
+            f.write(json.dumps({"model": "unsloth/gemma-2b"}))
+
+    def _create_model_repository_mock_llm(self):
+        shutil.rmtree("models", ignore_errors=True)
+        os.makedirs(f"models/{self._model_name}/1", exist_ok=True)
+        with open(f"models/{self._model_name}/config.pbtxt", "w") as f:
```
rmccorm4, line 108: If not making any modifications to it, why not `shutil.copy` mock_llm's config.pbtxt as well?

python/openai/README.md
```diff
@@ -230,6 +230,32 @@ pip install -r requirements-test.txt
 pytest -v tests/
 ```
 
+### LoRA Adapters
+
+If the command line argument `--lora-separator=<separator_string>` is provided
```
rmccorm4, line 235: Quick question - for easy UX purposes - why not provide a default lora separator and just load them if detected (ex: user prepared the multi_lora.json)? Seems tedious to make the user specify a separator to get lora support, but not sure what other solutions do.

python/openai/tests/test_lora.py
kthui: The [mock_llm's config.pbtxt](https://github.com/triton-inference-server/server/blob/jacky-openai-lora-vllm-model-list/python/openai/tests/test_models/mock_llm/config.pbtxt#L34-L45) doesn't work when the `--backend=vllm` is set, because it is missing the sampling_parameters and exclude_input_in_output inputs for vLLM backend. The [provided config](https://github.com/triton-inference-server/server/pull/8052/files#diff-92cadbd7d17864f2149b4ad239ba6f1bf3eeb3eea6888004ca677c742a51787cR114-R135) is to fill in the missing inputs for the inference to work. (The backend model.py doesn't have to read the additional input though)

python/openai/README.md
kthui: I think the main benefit is saving existing users from having to add `--lora-separator` to their start command, providing they already have working models with LoRA adapters over KServe, after upgrading their frontend version.

I have two concerns with a default separator:
1. We don't have control over what model/LoRA names are used. For example, a user trying out LoRA separator may name their model `experiment_model_with_lora_adapters` or LoRA adapter `gemmadoll-2b-dolly-lora-tune` (which is adapter HF repo name used in testing in lower case). If the default separator is named `_lora_`, then it will force the user to either rename their model/LoRA or set the `--lora-separator` to a different string.
2. If the LoRA name list is not detected (i.e. the model is from cloud storage), users should still be able to inference using well-formed model name and LoRA pairs, so we cannot turn off the default LoRA separator if the multi_lora.json is not detected. This complicates things for users not using LoRA, i.e. any model name containing the default LoRA separator will either require the user to set the `--lora-separator` flag or change their model name.

Given OpenAI endpoint API does not officially support LoRA, I think we should always let the user opt-in to a feature that is not 100% compatible with any well-formed request under the official standard (i.e. by explicitly setting the `--lora-separator=<str>`), instead of having them opt-out if they later find out it does not work for them.

Maybe there is a separator that will never conflict with any model/LoRA name that I am missing?

cc @nnshah1 @krishung5 @richardhuo-nv if you have any thought?

rmccorm4: Thanks for sharing more details - let's keep it opt-in for now then, and if we find it can be improved or made default later, we can do so later.

kthui: Filed ticket DLIS-8193

rmccorm4 (approved):

python/openai/README.md
```diff
@@ -230,6 +230,32 @@ pip install -r requirements-test.txt
...
+curl -s http://localhost:9000/v1/completions -H 'Content-Type: application/json' -d '{
+  "model": "<model_name>_separator_<lora_name>",
+  "prompt": "Machine learning is"
+}'
+```
+
+When listing or retrieving model(s), the model id will include the LoRA name in
+the same `<model_name><separator_string><lora_name>` format for each LoRA
+adapter listed on the `multi_lora.json`. Note: The LoRA name inclusion is
+limited to locally stored models, inference requests are not limited though.
```
rmccorm4, line 253: I think it would be nice to show an example output of calling `/v1/models` with a lora for users to get a quick understanding of how it would look and be used.

kthui: Sure, added the example:
[docs: Enhance example showing different outputs with different LoRAs](https://github.com/triton-inference-server/server/pull/8052/commits/8492018c52689f3c69cf2241a1566e90222c21b4)
[docs: Collapse OpenAI LoRA example](https://github.com/triton-inference-server/server/pull/8052/commits/fe8047e577ad354ae5719054848d170b03acb882)

rmccorm4: > When listing or retrieving model(s), the model id will include the LoRA name in the same <model_name><separator_string><lora_name> format for each LoRA adapter listed on the multi_lora.json. Note: The LoRA name inclusion is limited to locally stored models, inference requests are not limited though.

Can you add collapsed example output of `curl .../v1/models` showing the lora models in the output?

PR #8054, opened 2025-03-04
Reverse order of GAP and PA installation to avoid installing outdated PA

matthewkotila:
* SDK container is supposed to contain the version of PA and GAP that was built during its job
* `pip`-installing the GAP wheel built in the SDK job causes [tritonclient](https://pypi.org/project/tritonclient/) from PyPI to be installed
* [tritonclient](https://pypi.org/project/tritonclient/) from PyPI contains its own `perf_analyzer` binary that it installs to `/usr/local/bin`
* [tritonclient](https://pypi.org/project/tritonclient/) from PyPI is going to contain last month's `perf_analyzer` binary (older than the `perf_analyzer` binary built in the SDK job)
* This SDK Dockerfile was inadvertently installing its manually-built `perf_analyzer` binary into `/usr/local/bin` and _then_ `pip`-installing the manually-built GenAI-Perf wheel, causing the previously installed `perf_analyzer` to get overwritten with the last-month's `perf_analyzer` binary packaged with [tritonclient](https://pypi.org/project/tritonclient/) from PyPI
* To fix this, I propose reversing the order of GAP and PA installations
  * Install GAP wheel first, which installs last month's PA into `/usr/local/bin`
  * Install manual/latest PA next into `/usr/local/bin` overwriting outdated last-month's PA

### DIFF
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -238,6 +238,8 @@ COPY --from=sdk_build /workspace/client/src/python/library/tests/* qa/python_cli
 # Install an image needed by the quickstart and other documentation.
 COPY qa/images/mug.jpg images/mug.jpg
 
+RUN pip3 install install/python/genai_perf-*.whl
+
 # Install the dependencies needed to run the client examples. These
 # are not needed for building but including them allows this image to
 # be used to run the client examples.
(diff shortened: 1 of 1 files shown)


dyastremsky (approved):
Thank you for the clear PR title and description. Great fix.

Issue #7650, opened 2024-09-24
Big performance drop when using ensemble model over separate calls


jcuquemelle:
**Description**
We have an ensemble of 2 models chained together (description of models below).

Calling only the "preprocessing" model yields a max throughput of 21500 QPS @ 6 Cpu cores usage
Calling only the "inference" model yields a max throughput of 44000 QPS @ 6 Cpu + 0.7 Gpu usage
Calling the "ensemble" model yields a max throughput of less than 8500 QPS @ 10 Cpu + 0.7 Gpu usage

Why is there such a performance drop when using the ensemble ? given the figures for individual models, I'd expect the output of the ensemble to be around the same as the individual model with the less throughput i.e. 21500 QPS  

The server-side metrics don't show any sign of blocking step in the inference pipeline although client-side the latency increases over time: 
```
Request Rate: 8500 inference requests per seconds                                                                                       
  Pass [1] throughput: 8500.57 infer/sec. p75 latency: 3735 usec                                                                        
  Pass [2] throughput: 8499.32 infer/sec. p75 latency: 6115 usec
  Pass [3] throughput: 8500.86 infer/sec. p75 latency: 1298 usec
  Pass [4] throughput: 8282.7 infer/sec. p75 latency: 74570 usec
  Pass [5] throughput: 8384 infer/sec. p75 latency: 133184 usec
  Pass [6] throughput: 8425.3 infer/sec. p75 latency: 165502 usec
  Pass [7] throughput: 8255.55 infer/sec. p75 latency: 191599 usec
  Client: 
    Request count: 92706
    Throughput: 8356.54 infer/sec
    Avg client overhead: 18.98%
    p50 latency: 149491 usec
    p75 latency: 178315 usec
    p90 latency: 190544 usec
    p95 latency: 213366 usec
    p99 latency: 263968 usec
    Avg gRPC time: 153756 usec (marshal 5 usec + response wait 153751 usec + unmarshal 0 usec)
  Server: 
    Inference count: 92705
    Execution count: 92705
    Successful request count: 92705
    Avg request latency: 924 usec (overhead 265 usec + queue 90 usec + compute 569 usec)

  Composing models: 
  inference, version: 1
      Inference count: 92705
      Execution count: 73789
      Successful request count: 92705
      Avg request latency: 562 usec (overhead 107 usec + queue 67 usec + compute input 81 usec + compute infer 278 usec + compute output
 27 usec)

  preprocessing, version: 1
      Inference count: 92704
[...]

PR #8061, opened 2025-03-10
feat: Configurable grpc infer thread count

yinggeh:
#### Checklist

#### Commit Type:
- [x] feat

#### Test plan:
L0_sequence_stress

- CI Pipeline ID:
25216718

#### Background
In ensemble model, the bottleneck of model maximum throughput is bounded by the gRPC infer handling rate.

- Fixes GitHub issue: #7650


src/command_line_parser.cc
```diff
@@ -1435,6 +1440,14 @@ TritonParser::Parse(int argc, char** argv)
         case OPTION_GRPC_ADDRESS:
           lgrpc_options.socket_.address_ = optarg;
           break;
+        case OPTION_GRPC_INFER_THREAD_COUNT:
+          lgrpc_options.infer_thread_count_ = ParseOption<int>(optarg);
+          if (lgrpc_options.infer_thread_count_ < 2) {
+            throw ParseException(
```
indrajit96, line 1446: Do we also want a upper bound on the infer_thread_count_?

indrajit96 (commented):
Update the docs to reflect new optional param

src/command_line_parser.cc
yinggeh: Good point. I added 256 as upper bound because CPU with maximum cores on the market is around 128.

tanmayv25 (dismissed):

Issue #8034, opened 2025-02-25
Segment fault crash due to race condition of request cancellation (with fix proposal)


lunwang-ttd:
**Description**
There is a race condition causes segment fault crash when a request is cancelled.

**Triton Information**
r25.01

Are you using the Triton container or did you build it yourself?
I built a RelWithDebInfo for debugging, and the issue can also be reproduced with the official Triton container.

**To Reproduce**
Start triton server in GDB with the simple model (`docs/examples/model_repository/simple/config.pbtxt`), suppose the model repository is `/models`

```
#!/bin/bash

# Delays to simulate the race condition
export TRITONSERVER_DELAY_GRPC_NOTIFICATION=5
export TRITONSERVER_DELAY_GRPC_ENQUEUE=2
export TRITONSERVER_DELAY_GRPC_PROCESS=2

gdb -ex=r --args tritonserver \
--log-verbose=0 \
--model-repository=/models \
--grpc-infer-allocation-pool-size=8
```

Send requests to the server.

```
#!/usr/bin/env python
# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions
# are met:
#  * Redistributions of source code must retain the above copyright
#    notice, this list of conditions and the following disclaimer.
#  * Redistributions in binary form must reproduce the above copyright
#    notice, this list of conditions and the following disclaimer in the
#    documentation and/or other materials provided with the distribution.
#  * Neither the name of NVIDIA CORPORATION nor the names of its
#    contributors may be used to endorse or promote products derived
#    from this software without specific prior written permission.
#
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
[...]

PR #8078, opened 2025-03-20
fix: Fix gRPC cancellation race condition

yinggeh:
### Problem description:
This PR fixes a race condition where a cancellation request is received before InferResponseComplete, but the server processes it only after the state has been released, leading to a potential segmentation fault.

1. **Thread A**: Server receives cancellation and pops the notification state from the completion queue. It sets `SetReceivedNotification(true);`.
2. **Thread B**: Before Thread A enters `Process()`, `InferResponseComplete` is triggered, detecting the cancellation and putting the state back into the queue `PutTaskBackToQueue`.
3. **Thread C**: Pops the state from step 2, enters `Process()`, and releases the state via `StateRelease(state)`.
4. **Thread A**: Finally enters `Process()`, but since the state is already released, calling `IsGrpcContextCancelled()` results in a segmentation fault.

### Why This Happens:
- There's no guarantee about the execution order of Process() across different threads.
- A naive solution using a reference counter led to potential deadlocks when multiple requests were canceled simultaneously.

### Solution in This PR:
Calls `SetReceivedNotification(true);` inside a mutex lock, ensuring a cancellation at `InferResponseComplete` falls into one of two cases:

1. Before InferResponseComplete: The cancellation is processed and no response is sent to the client.
2. After InferResponseComplete: It's too late to cancel the request as model has already finished execution. The cancellation is ignored.

#### Checklist

#### Commit Type:
- [x] fix

#### Test plan:
L0_request_cancellation

- CI Pipeline ID:
25756852

- closes GitHub issue: #8034


src/grpc/infer_handler.cc
```diff
@@ -697,6 +697,26 @@ ModelInferHandler::Process(
         std::chrono::milliseconds(state->delay_process_ms_));
   }
 
+  // If notification arrives after InferResponseComplete is invoked, simply
+  // ignore it and release the state as if the request is completed.
+  if (is_notification) {
+    bool rn_prev = state->context_->ReceivedNotification();
```
lunwang-ttd, line 703: In which cases, `rn_prev` is true? Can gRPC framework deliver multiple cancellation tags for a single request?

src/grpc/infer_handler.cc
```diff
@@ -697,6 +697,26 @@ ModelInferHandler::Process(
...
+    state->context_->SetReceivedNotification(true);
+    if (state->IsGrpcContextCancelled()) {
+      // this is the cancellation notification
+      LOG_VERBOSE(1) << "Cancellation notification received for " << Name()
+                     << ", rpc_ok=" << rpc_ok << ", context "
+                     << state->context_->unique_id_ << " step "
+                     << state->context_->step_ << ", state "
+                     << state->unique_id_ << " step " << state->step_;
+      if (state->step_ == Steps::COMPLETE || state->step_ == Steps::FINISH) {
```
lunwang-ttd, line 712: With this logic, it seems the branch in `IssueRequestCancellation` is not needed and will never enter anymore:
```
          } else if (state->step_ == Steps::COMPLETE) {
            // The RPC is complete and no callback will be invoked to retrieve
            // the object. Hence, need to explicitly place the state on the
            // completion queue.
            PutTaskBackToQueue(state);
          }
```

src/grpc/infer_handler.cc
yinggeh: Ideally `rn_prev` should be `false` before notification is received. The implementation here was just to be cautious.

src/grpc/infer_handler.cc
yinggeh: Not necessary. Because `state` in `IssueRequestCancellation` may not be the same state in `HandleCancellation`.

src/grpc/infer_handler.h
```diff
@@ -1522,14 +1509,25 @@ InferHandler<
...
       }
+
       LOG_VERBOSE(2) << "Grpc::CQ::Next() "
                      << state->context_->DebugString(state);
       if (!Process(state, ok, is_notification)) {
         LOG_VERBOSE(1) << "Done for " << Name() << ", " << state->unique_id_;
         state->context_->EraseState(state);
         StateRelease(state);
       } else {
+        // In non-streaming infer mode, there is no guarantee that
```
tanmayv25, line 1529: In non-streaming infer mode which has multiple request handler, there is no guarantee that state->context_ is valid beyond this line.

yinggeh: Updated

src/grpc/infer_handler.cc
```diff
@@ -697,9 +697,25 @@ ModelInferHandler::Process(
         std::chrono::milliseconds(state->delay_process_ms_));
   }
 
+  // If notification arrives after InferResponseComplete is invoked, simply
```
lunwang-ttd, line 700: This commend is confusing, it's deduplicated with comment at L736

src/grpc/infer_handler.cc
```diff
@@ -715,10 +731,17 @@ ModelInferHandler::Process(
       // thread, and cancellation at step START was not reproducible in a
       // single thread scenario.
       StartNewRequest();
+    } else if (
+        state->step_ == Steps::COMPLETE || state->step_ == Steps::FINISH) {
+      // If the request is completed, simply ignore the cancellation and
+      // restore the previous notification state.
```
lunwang-ttd, line 737: `restore the previous notification state` is confusing, I think it should also be removed after you removed the `prev_state` variable

qa/L0_request_cancellation/grpc_cancellation_test.py
```diff
@@ -218,13 +240,34 @@ def test_grpc_async_infer_cancellation_during_response_complete(self):
...
-        time.sleep(self._model_delay + 2)
+        time.sleep(self._model_delay + delay_process_entry_sec + 2)
         future.cancel()
         time.sleep(
-            delay_notification_sec + delay_response_completion_sec
+            delay_response_completion_sec
         )  # ensure the cancellation is processed
         self._assert_callback_cancelled()
 
+    def test_grpc_async_infer_cancellation_before_response_complete_and_process_after_final_response(
```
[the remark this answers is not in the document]
lunwang-ttd: I checked the codes again, actually after states are processed in `InferResponseComplete`, they are erased from inflight states set, which means these states won't be released with either COMPLETE or FINISH step.

qa/L0_request_cancellation/grpc_cancellation_test.py
```diff
@@ -218,13 +240,34 @@ def test_grpc_async_infer_cancellation_during_response_complete(self):
...
+            int(os.getenv("TRITONSERVER_DELAY_GRPC_NOTIFICATION")) / 1000
+        )
+        future = self._client.async_infer(
+            model_name=self._model_name,
+            inputs=self._inputs,
+            callback=self._callback,
+            outputs=self._outputs,
+        )
+        # ensure the cancellation is received before InferResponseComplete checking cancellation
+        time.sleep(self._model_delay + 2)
```
lunwang-ttd, line 266: Should this be `self._model_delay - 2`?

qa/L0_request_cancellation/grpc_cancellation_test.py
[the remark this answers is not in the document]
lunwang-ttd: For streaming infer, states with async notification [are not erased](https://github.com/triton-inference-server/server/blob/7337c139f04cfe0395356036b5cb61bad9e88f0b/src/grpc/stream_infer_handler.cc#L655) from inflight set, so they will be handled in `IssueRequestCancellation`, but the `FINISH` state is not considered, so I think there is also a chance of state leak.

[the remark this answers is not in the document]
yinggeh: Hi @lunwang-ttd. Can you elaborate your questions? Are you pointing out a potential bug that needs to be fixed or is fixed by this PR?

> For this case, the old implementation will cause leak of state

This test is used to replicate https://github.com/triton-inference-server/server/issues/8034. It won't cause a memory leak but a segmentation fault in the old implementation.

qa/L0_request_cancellation/grpc_cancellation_test.py
yinggeh: Nice catch

qa/L0_request_cancellation/grpc_cancellation_test.py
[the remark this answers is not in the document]
lunwang-ttd: @yinggeh I misunderstood the purpose of this case, I thought it was to simulate cancellation notification processing when step is in COMPLETE or FINISH, for this case, state will be leaked.

This PR also fixes this potential leak.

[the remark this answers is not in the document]
yinggeh: @lunwang-ttd I understand your concern and appreciate your suggestions. The two places that set state to FINISH in (non-shutting down) stream mode without immediately releasing the state after returning from [Process](https://github.com/triton-inference-server/server/blob/8e5d9f8fc61a40d1436bd4f896b3845fb9e915ef/src/grpc/stream_infer_handler.cc#L133) are
1. https://github.com/triton-inference-server/server/blob/8e5d9f8fc61a40d1436bd4f896b3845fb9e915ef/src/grpc/infer_handler.h#L776
2. https://github.com/triton-inference-server/server/blob/8e5d9f8fc61a40d1436bd4f896b3845fb9e915ef/src/grpc/stream_infer_handler.cc#L447

However, both state transitions only occur when client closes stream where the lifecycle of states are handled properly base on what I found. I believe you that there exists memory leaks in gRPC logics and I am more than happy to help if you kindly provide a simple reproduction code/steps.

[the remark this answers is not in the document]
lunwang-ttd: Hi @yinggeh,
I only reproduced the memory leak issue with the InferHandler (see below), and I have confirmed that this PR fixes the issue, thanks!

For StreamInferHandler, I didn't try to reproduce the memory leak issue because we are not using stream infer on production, I only judged based on my understanding on related codes, and I might be wrong. Feel free to ignore it if you think it's not an issue, thanks!


### Reproduction of memory leak with InferHandler
Step 1: start triton server with the simple model (docs/examples/model_repository/simple/config.pbtxt), suppose the model repository is /models

```
#!/bin/bash

# Delays to simulate the race condition
export TRITONSERVER_DELAY_GRPC_NOTIFICATION=2
export TRITONSERVER_DELAY_GRPC_ENQUEUE=1

tritonserver \
--log-verbose=0 \
--model-repository=/models
```

Step 2: on the same node, run the attached Python script repeatedly.

```bash
watch -n 1 ./client_simple_model.py
```

client_simple_model.py is below.

```python
#!/usr/bin/env python
# Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions
# are met:
#  * Redistributions of source code must retain the above copyright
#    notice, this list of conditions and the following disclaimer.
#  * Redistributions in binary form must reproduce the above copyright
#    notice, this list of conditions and the following disclaimer in the
#    documentation and/or other materials provided with the distribution.
#  * Neither the name of NVIDIA CORPORATION nor the names of its
#    contributors may be used to endorse or promote products derived
#    from this software without specific prior written permission.
#
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT

import sys
import time
from functools import partial

import numpy as np
from tritonclient.utils import InferenceServerException

if __name__ == "__main__":
    # 2 ms cancellation timeout
    client_timeout = 0.002
    url = "localhost:8001"

    try:
        triton_client = grpcclient.InferenceServerClient(
            url=url
        )
    except Exception as e:
        print("context creation failed: " + str(e))
        sys.exit()

    model_name = "simple"

    # Infer
    inputs = []
    outputs = []
    inputs.append(grpcclient.InferInput("INPUT0", [1, 16], "INT32"))
    inputs.append(grpcclient.InferInput("INPUT1", [1, 16], "INT32"))

    # Create the data for the two input tensors. Initialize the first
    # to unique integers and the second to all ones.
    input0_data = np.arange(start=0, stop=16, dtype=np.int32)
    input0_data = np.expand_dims(input0_data, axis=0)
    input1_data = np.ones(shape=(1, 16), dtype=np.int32)

    # Initialize the data
    inputs[0].set_data_from_numpy(input0_data)
    inputs[1].set_data_from_numpy(input1_data)

    outputs.append(grpcclient.InferRequestedOutput("OUTPUT0"))
    outputs.append(grpcclient.InferRequestedOutput("OUTPUT1"))

    # Define the callback function. Note the last two parameters should be
    # result and error. InferenceServerClient would povide the results of an
    # inference as grpcclient.InferResult in result. For successful
    # inference, error will be None, otherwise it will be an object of
    # tritonclientutils.InferenceServerException holding the error details
    def callback(user_data, result, error):
        if error:
            user_data.append(error)
        else:
            user_data.append(result)

    # list to hold the results of inference.
    user_data = []

    # Inference call
    for i in range(1000):
        triton_client.async_infer(
            model_name=model_name,
            inputs=inputs,
            callback=partial(callback, user_data),
            outputs=outputs,
            client_timeout=client_timeout,
        )
        time.sleep(0.001)

    # Wait until the results are available in user_data
    time_out = 20
    while (len(user_data) == 0) and time_out > 0:
        time_out = time_out - 1
        time.sleep(1)

    print("results: ", len(user_data))

    # Display and validate the available results
    count = 1
    for result_data in user_data:
        # Check for the errors
        if type(result_data) == InferenceServerException:
            print("request -", count, result_data.message())
        else:
            print("request -", count, "PASS")
        count += 1

    print("PASS: Async infer")
```

PR #8089, opened 2025-03-22
docs: change server README to only load densenet_onnx since tf model cannot load after 25.03

ziqif-nv:
This PR is needed for 25.03, since in 25.03, all TF models would not be loaded. So in the guide, we have to use extra flags to load the onnx model only.

Tested manually and the new cmd works in 25.02.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,12 +89,11 @@ cd server/docs/examples
 ./fetch_models.sh
 
 # Step 2: Launch triton from the NGC Triton container
-docker run --gpus=1 --rm --net=host -v ${PWD}/model_repository:/models nvcr.io/nvidia/tritonserver:25.02-py3 tritonserver --model-repository=/models
+docker run --gpus=1 --rm --net=host -v ${PWD}/model_repository:/models nvcr.io/nvidia/tritonserver:25.02-py3 tritonserver --model-repository=/models --model-control-mode explicit --load-model densenet_onnx
 
 # Step 3: Sending an Inference Request
 # In a separate console, launch the image_client example from the NGC Triton SDK container
-docker run -it --rm --net=host nvcr.io/nvidia/tritonserver:25.02-py3-sdk
-/workspace/install/bin/image_client -m densenet_onnx -c 3 -s INCEPTION /workspace/images/mug.jpg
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -89,12 +89,11 @@ cd server/docs/examples
 ./fetch_models.sh
 
 # Step 2: Launch triton from the NGC Triton container
-docker run --gpus=1 --rm --net=host -v ${PWD}/model_repository:/models nvcr.io/nvidia/tritonserver:25.02-py3 tritonserver --model-repository=/models
+docker run --gpus=1 --rm --net=host -v ${PWD}/model_repository:/models nvcr.io/nvidia/tritonserver:25.02-py3 tritonserver --model-repository=/models --model-control-mode explicit --load-model densenet_onnx
```
kthui, line 92: Can we remove the extra
```bash
--model-control-mode explicit --load-model densenet_onnx
```
flags, for keeping the instruction short?

It is because the `/models` repository should only contain the `densenet_onnx` model given the `inception_graphdef` model is removed from the `./fetch_models.sh`.

ziqif-nv: > Can we remove the extra
> 
> ```shell
> --model-control-mode explicit --load-model densenet_onnx
> ```
> 
> flags, for keeping the instruction short?
> 
> It is because the `/models` repository should only contain the `densenet_onnx` model given the `inception_graphdef` model is removed from the `./fetch_models.sh`.

in some of the container, like QA, the simple* models would be copied into /models such that it would cause error

I am adding the extra flags to make sure it would not load those unnecessary tf models

ziqif-nv: also, as you can see `./fetch_models.sh` would only download the model weights, while the needed `config.pbtxt` and `densenet_labels.txt` must exist

`config.pbtxt` and `densenet_labels.txt` should be already copied into /models as well as other simple* tf models

krishung5: > I am adding the extra flags to make sure it would not load those unnecessary tf models

I might have missed some context here, but I thought we'd like to remove all the tf models, including the ones you mention here? In this case, we won't need those extra flags.

krishung5: I see, thanks for the context! I have no objections so I'll let you decide which way to do.

krishung5 (approved):

PR #8099, opened 2025-03-24
docs: rename userguide AI Agents section to Features | add spec decoding to Features

ziqif-nv:
Renaming `AI Agents` to `Features` since `Constrained Decoding` and `Function Calling` are more like features. 

Also, vLLM side is using Features as well for `Tool Calling` and `Speculative Decoding`

### DIFF
--- a/docs/contents.rst
+++ b/docs/contents.rst
@@ -50,10 +50,11 @@
 
 .. toctree::
    :hidden:
-   :caption: AI Agents
+   :caption: Features
 
    Constrained Decoding <../tutorials/AI_Agents_Guide/Constrained_Decoding/README.md>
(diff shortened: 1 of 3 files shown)


krishung5 (approved):

harryskim:
How about LLM Features instead of Features?

ziqif-nv:
> How about LLM Features instead of Features?

Updated. Thanks

oandreeva-nv:
Should we change tutorials structure as well?

ziqif-nv:
> Should we change tutorials structure as well?

Good question. We should.

So should I change tutorials side first before making changing here?

statiraju (approved):

PR #8124, opened 2025-04-03
Fix L0_response_cache test

tanmayv25:
TF model would dump its response data on the CPU even if the model was GPU. However, ORT dumps data in GPU memory for the next step to consume it. It only copies the data to CPU if the response data is required. As response cache only supports data from CPU memory we were not populating the cache. Hence seeing no cache hit.

### DIFF
--- a/qa/L0_response_cache/ensemble_cache_test.py
+++ b/qa/L0_response_cache/ensemble_cache_test.py
@@ -78,7 +78,7 @@ class EnsembleCacheTest(tu.TestResultCollector):
                 with open(config_file, "w") as f:
                     config_data += config_to_add
                     f.write(config_data)
-    
+
     def _add_instance_group_cpu(self, config_file):
         # Utility function to add instance group of kind CPU to the config file
         with open(config_file, "r") as f:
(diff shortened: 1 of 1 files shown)


qa/L0_response_cache/ensemble_cache_test.py
```diff
@@ -252,6 +261,8 @@ def test_ensemble_composing_model_cache_enabled(self):
         self._update_config(
             self.composing_config_file, RESPONSE_CACHE_PATTERN, RESPONSE_CACHE_CONFIG
         )
+        # Currently, response cache is supported only for tensors on CPU.
+        self._add_instance_group_cpu(self.composing_config_file)
```
ziqif-nv, line 265: what is the config.pbtxt contains `instance_group` with `kind: KIND_GPU`? should we change it to `KIND_CPU`?

ziqif-nv (commented):
Thank you so much!

qa/L0_response_cache/ensemble_cache_test.py
tanmayv25: Yes. For this case we don't have the instance_group setting for model config. Hence, I just added it directly.

ziqif-nv: I c. I saw you have the logic `if "instance_group" not in config_data:` so I wonder my mentioned case is missed or not

ziqif-nv (approved):
awesome!

PR #8134, opened 2025-04-07
feat: add the tool calling to the openai frontend

richardhuo-nv:
#### What does the PR do?
1. Add the tool calling parsers implementation to openai frontend, the available parsers are llama3 and mistral. Most of the implementation is from the vllm. A user could use the --tool-call-parser arguments to specify the tool parser. 
2. Add the --chat-template {chat template file path} argument to allow the user use the customized template to better tune the prompt for tool calling.
3. Add the guided decoding backend integration with the tool calling to enable the named tool calling and required tool calling functionalities.

Please check more detail in the change of README.md

All changes in `python/openai/openai_frontend/engine/utils/tool_call_parsers` are from the vLLM with some minor compatibility changes. 
https://github.com/vllm-project/vllm/tree/main/vllm/entrypoints/openai/tool_parsers

Example for the tool calling:
```python 
import json
from openai import OpenAI


def get_current_weather(city: str, state: str, unit: "str"):
    return (
        "The weather in Dallas, Texas is 85 degrees fahrenheit. It is "
        "partly cloudly, with highs in the 90's."
    )

available_tools = {"get_current_weather": get_current_weather}

openai_api_key = "EMPTY"
openai_api_base = "http://localhost:9000/v1"

client = OpenAI(
    api_key=openai_api_key,
    base_url=openai_api_base,
)

model = "llama-3.1-8b-instruct" # change this to the model in the repository

tools = [
    {
        "type": "function",
        "function": {
            "name": "get_current_weather",
            "description": "Get the current weather in a given location",
            "parameters": {
                "type": "object",
                "properties": {
                    "city": {
                        "type": "string",
                        "description": "The city to find the weather for, e.g. 'San Francisco'",
                    },
                    "state": {
                        "type": "string",
                        "description": "the two-letter abbreviation for the state that the city is"
                        " in, e.g. 'CA' which would mean 'California'",
                    },
                    "unit": {
                        "type": "string",
                        "description": "The unit to fetch the temperature in",
                        "enum": ["celsius", "fahrenheit"],
                    },
                },
                "required": ["city", "state", "unit"],
            },
        },
    }
]

messages = [
    {
        "role": "system",
        "content": "You're a helpful assistant! Answer the users question best you can.",
    },
    {"role": "user", "content": "What is the weather in Dallas, Texas in Fahrenheit?"},
]

tool_calls = client.chat.completions.create(
    messages=messages, model=model, tools=tools, max_tokens=128
)
print(tool_calls)
function_name = tool_calls.choices[0].message.tool_calls[0].function.name
function_arguments = tool_calls.choices[0].message.tool_calls[0].function.arguments

print(f"function name: " f"{function_name}")
print(f"function arguments: {function_arguments}")
print(f"tool calling result: {available_tools[function_name](**json.loads(function_arguments))}")
```
Example Outputs:
```
ChatCompletion(id='cmpl-fde5e58e-13e9-11f0-8e66-a3cfd3802b07', choices=[Choice(finish_reason='tool_calls', index=0, logprobs=None, message=ChatCompletionMessage(content='', refusal=None, role='assistant', audio=None, function_call=None, tool_calls=[ChatCompletionMessageToolCall(id='cmpl-fe1ddd67-13e9-11f0-a6c8-a3cfd3802b07', function=Function(arguments='{"city": "Dallas", "state": "TX", "unit": "fahrenheit"}', name='get_current_weather'), type='function')]))], created=1744055615, model='llama-3.1-8b-instruct', object='chat.completion', service_tier=None, system_fingerprint=None, usage=None)
function name: get_current_weather
function arguments: {"city": "Dallas", "state": "TX", "unit": "fahrenheit"}
tool calling result: The weather in Dallas, Texas is 85 degrees fahrenheit. It is partly cloudly, with highs in the 90's.
```
You can find the `too_calls` field in the ChatCompletionMessage is filled with the function name and arguments.

Example response before adding the tool calling functionality :
```
ChatCompletion(id='cmpl-53f8493f-13e9-11f0-a6e6-6b1a0bd73b9d', choices=[Choice(finish_reason='stop', index=0, logprobs=None, message=ChatCompletionMessage(content='Unfortunately, I\'m a large language model, I don\'t have real-time access to current weather conditions. However, I can suggest some options to help you find out the current weather in Dallas, Texas.\n\nYou can check the following websites for the current weather in Dallas, Texas:\n\n1.  National Weather Service (weather.gov) - Enter "Dallas, TX" in the search bar to get the current weather conditions.\n2.  Weather.com (weather.com) - Enter "Dallas, TX" in the search bar to get the current weather conditions.\n3.  AccuWeather (accuweather.com) - Enter "Dallas,', refusal=None, role='assistant', audio=None, function_call=None, tool_calls=None))], created=1744055330, model='llama-3.1-8b-instruct', object='chat.completion', service_tier=None, system_fingerprint=None, usage=None)
```
You can find that the `too_calls` field in the ChatCompletionMessage is None. 

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/merge_requests/1491

The reviewer can start with the chat method in the triton_engine.py, where the core logics happen. 

- CI Pipeline ID:
26830934

#### Caveats:
Tensorrt LLM backend's guided decoding is missing in the latest official release 0.18.0, the tests can only path with Tensorrt LLM build on main branch.

#### Background
One of our partners requested the tool calling interface in the openai frontend to be integrated with NIM.


.gitignore
rmccorm4, line 1: Thanks Richard!

1. Can you share a couple high level examples of the new functionality in the PR description? Maybe a before/after of a request/response with and without tool calling to highlight the differences?
2. Can we add docs on using these features? How-to, examples, limitations, etc.

oandreeva-nv: +1 on @rmccorm4 . 

Are there any limitations expected from this implementation?

python/openai/README.md
```diff
@@ -342,3 +342,228 @@ For more information on the `tritonfrontend` python bindings, see the docs
     - Set the following environment variable: `export TRTLLM_ORCHESTRATOR=1`
 - [ ] TensorRT-LLM ([Leader Mode](https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#leader-mode))
     - Not currently supported
+
+## Tool Calling
+
+The OpenAI frontend supports `tools` and `tool_choice` in the `v1/chat/completions` API. Please refer to the OpenAI API reference for more details about these parameters.
+  [tools](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools)
+  [tool_choice](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice)
```
rmccorm4, line 350:
```suggestion
The OpenAI frontend supports `tools` and `tool_choice` in the `v1/chat/completions` API. Please refer to the OpenAI API reference for more details about these parameters:
  [tools](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools),
  [tool_choice](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice)
```

python/openai/openai_frontend/engine/chat_utils.py
rmccorm4, line 1: nit: Can this be `utils/chat.py` to be grouped with other utilities?

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -90,6 +105,14 @@ def __init__(
         # now, and won't account for dynamically loading/unloading models.
         self.create_time = int(time.time())
         self.model_metadata = self._get_model_metadata()
+        self.tool_call_parser = (
+            ToolParserManager.get_tool_parser_cls(tool_call_parser)
+            if tool_call_parser
+            else None
+        )
+        self.chat_template = (
+            load_chat_template(chat_template) if chat_template else None
```
rmccorm4, line 114: What's the context for this extra chat_template arg instead of using the chat template attached to the tokenizer? Can you add an inline comment about it?

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -299,44 +394,170 @@ def _get_first_streaming_chat_response(
...
+                    tool_calls=[
+                        ChatCompletionMessageToolCallChunk(
+                            index=0,
+                            function=Function2(
+                                name=tool_choice_required_function_name,
+                                arguments=delta_text,
+                            ),
+                        )
+                    ]
+                )
```
rmccorm4, line 464: These first if two blocks have the same result right? Can it be simplified to something like this? (may not be 100% correct):
```python
tool_fn_name = tool_choice_function_name or tool_choice_required_function_name

if tool_fn_name:
  <call>
elif tool_choice_auto:
  ...
else:
  ...
```

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -152,16 +192,71 @@ async def chat(
...
                     index=0,
-                    message=ChatCompletionResponseMessage(
-                        content=text, role=role, function_call=None
-                    ),
+                    message=response_message,
                     logprobs=None,
-                    finish_reason=ChatCompletionFinishReason.stop,
+                    finish_reason=ChatCompletionFinishReason.tool_calls
+                    if auto_tools_called
+                    else ChatCompletionFinishReason.stop,
```
rmccorm4, line 259: Why is the finish reason `tool_calls` only on "auto" `tool_choice_auto`, but `stop` for `tool_choice` or `tool_choice_required`?

src/python/tritonfrontend/_api/_kservegrpc.pyi
```diff
@@ -57,6 +57,7 @@ class KServeGrpc:
         max_connection_age_grace_ms: int
         infer_compression_level: int | Grpc_compression_level
         infer_allocation_pool_size: int
+        max_response_pool_size: int
```
rmccorm4, line 60: Thanks for adding these changes to the type hints! Not relevant to this PR, but just to spread knowledge - I think we can apply a change similar to this PR to remove automate generation of these stub `.pyi` files rather than manually keeping them in sync: https://github.com/triton-inference-server/core/pull/414

qa/L0_openai/generate_engine.py
```diff
@@ -0,0 +1,56 @@
...
+from argparse import ArgumentParser
+
+from tensorrt_llm import LLM, BuildConfig
+from tensorrt_llm.plugin import PluginConfig
+
+
+def generate_model_engine(model: str, engines_path: str):
+    config = BuildConfig(plugin_config=PluginConfig.from_dict({"_gemm_plugin": "auto"}))
+
+    engine = LLM(
```
rmccorm4, line 35: Nice use of LLM API to simplify engine building!

python/openai/tests/utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2024-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
rmccorm4, line 1: Unnecessary change - can revert
```suggestion
# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/openai/tests/test_tool_calling.py
```diff
@@ -0,0 +1,571 @@
...
+        assert role_sent
+        assert finish_reason_count == 1
+        assert len(chunks)
+
+        # FIXME: tensorrt_llm streaming and non-streaming requests are not generating exact same text response in this case
+        # example:
+        #   streaming: The current temperature in Dallas, TX is 98°F. Is there anything else I can assist you with?
+        #   non-streaming: It's currently 98°F in Dallas.
+        if backend != "tensorrtllm":
+            assert "".join(chunks) == choice.message.content
```
rmccorm4, line 327: Can you instead do something like this so we can automate detecting when TRTLLM fixes/changes this behavior?
```python
if backend == "tensorrtllm":
  # FIXME: tensorrt_llm streaming and non-streaming requests are not generating exact same text response in this case
  # NOTE: If this assertion consistently does not get raised, then this means the TRTLLM behavior 
  # is changed and this special case can be removed.
  with pytest.raises ... # or pytest.xfail or something
    assert "".join(chunks) == choice.message.content
else:
  assert "".join(chunks) == choice.message.content

File: python/openai/tests/test_tool_calling.py [hunk 9]
```diff
@@ -0,0 +1,571 @@
...
+        assert function_name == tool_calls[0].function.name
+        assert isinstance(function_args_str, str)
+
+        # validate arguments
+        streamed_args = json.loads(function_args_str)
+        assert isinstance(streamed_args, Dict)
+        assert isinstance(streamed_args.get("city"), str)
+        assert isinstance(streamed_args.get("state"), str)
+        assert streamed_args.get("city") == "Dallas"
+        assert streamed_args.get("state") == "TX"
```
rmccorm4, line 254: I think this whole assertion chunk is copied a few times, and can probably be moved to a helper to reduce code dupe (feel free to change the function name to what makes most sense):
```python
self.validate_weather_tool_outputs()
```

python/openai/tests/test_openai_client.py
```diff
@@ -243,8 +243,3 @@ async def test_chat_streaming(
         assert len(chunks) > 1
         streamed_output = "".join(chunks)
         assert streamed_output == output
-
-    @pytest.mark.skip(reason="Not Implemented Yet")
-    @pytest.mark.asyncio
-    async def test_openai_client_function_calling(self):
-        pass
```
rmccorm4, line 250: Thanks for removing the placeholder!

python/openai/openai_frontend/main.py
```diff
@@ -120,6 +120,18 @@ def parse_args():
...
+        type=str,
+        default=None,
+        help="Specify the parser for handling tool calling related response text. Options include: 'llama3' and 'mistral'.",
+    )
+    triton_group.add_argument(
+        "--chat-template",
+        type=str,
+        default=None,
+        help="The path of the jinja chat template file.",
+    )
```
rmccorm4, line 134: Why do we need an extra chat template arg? Just for supporting models with no built-in chat template? Are we using that in a test or example somewhere?

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Given two strings, extract the difference in the middle between two strings
+    that are known to have a common prefix and/or suffix.
+
+    This function is provided as a UTILITY for extracting information from JSON
+    generated by partial_json_parser, to help in ensuring that the right tokens
+    are returned in streaming, so that close-quotes, close-brackets and
+    close-braces are not returned prematurely. The order of arguments IS
+    important - the new version of the partially-parsed JSON must be the first
+    argument, and the secnod argument must be from the previous generation.
```
rmccorm4, line 109:
```suggestion
    argument, and the second argument must be from the previous generation.
```

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
```diff
@@ -0,0 +1,129 @@
...
+        -> 'ple'
+
+    """
+    suffix = find_common_suffix(curr, old)
+
+    old = old[::-1].replace(suffix[::-1], "", 1)[::-1]
+    prefix = find_common_prefix(curr, old)
+    diff = curr
+    if len(suffix):
+        diff = diff[::-1].replace(suffix[::-1], "", 1)[::-1]
```
rmccorm4, line 123: These are some pretty complex lines (reverse, remove first instance of reversed suffix, then reverse again?) -- is there a reference that these came from, or some simple quick description/example you could add inline with it?

Alternatively, would it be easy to simplify these lines? It's fine if not.

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
```diff
@@ -0,0 +1,129 @@
...
+import json
+from typing import Any
+
+import partial_json_parser
+from partial_json_parser.core.options import Allow
+
+
+# partial_json_parser doesn't support extra data and
+# JSONDecorder.raw_decode doesn't support partial JSON
+def partial_json_loads(input_str: str, flags: Allow) -> tuple[Any, int]:
```
rmccorm4, line 40: Can you add some context on the partial json handling? Is it specifically because of streaming mode where the full response will be json, but when streamed as tokens, each is partial?

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
rmccorm4: Ah I see it looks like they came from vllm: https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/tool_parsers/utils.py

python/openai/openai_frontend/engine/utils/tool_call_parsers/llama_tool_call_parser.py
```diff
@@ -0,0 +1,289 @@
...
+        # case -- if a tool call token is not present, return a text response
+        if not (full_text.startswith(self.bot_token) or full_text.startswith("{")):
+            return ChatCompletionResponseMessage(
+                tool_calls=None, content=full_text, role=role
+            )
+
+        original_full_text = full_text
+        try:
+            # FIXME: tensorrt_llm backend might generate some unnecessary text messages
+            # after the tool call json text starting with "assistant\n\n".
```
rmccorm4, line 85: Is this a bug we need to file against trtllm?

python/openai/tests/test_tool_calling.py
```diff
@@ -0,0 +1,571 @@
...
+        assert streamed_args.get("unit") == "fahrenheit"
+
+        # make sure everything matches non-streaming except for ID
+        assert function_name == tool_calls[0].function.name
+        assert choice.message.role == role_name
+        assert choice.message.tool_calls[0].function.name == function_name
+
+    @pytest.mark.asyncio
+    async def test_tool_call_with_required_tool_choice(
```
rmccorm4, line 453: Do we have any negative tests like requesting or requiring a tool that doesn't exist?

python/openai/openai_frontend/engine/chat_utils.py
```diff
@@ -0,0 +1,166 @@
...
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import json
+from functools import partial
+from typing import Dict, Iterable, List, Optional, Required, TypedDict, Union, cast
+
+from openai.types.chat import ChatCompletionMessageToolCallParam
+from openai.types.chat.chat_completion_message_tool_call_param import Function
```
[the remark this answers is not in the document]
richardhuo-nv: The ConversationMessage will be used in the `apply_chat_template` of the tokenizer, which essentially accepts the dict type, and apply the dict into the jinja based chat template, and all `openai.types.chat` class are defined as TypeDict . But classes defined in the `schemas.openai` are BaseModels, which raise exceptions when deserialize as a dict.

python/openai/openai_frontend/engine/triton_engine.py
richardhuo-nv: Sure, will add comments. 

Sometimes the original chat templates with model is not as good in some cases, so vLLM actually tuned bunch of chat templates to help the user get higher quality tool call prompt. 
https://github.com/vllm-project/vllm/tree/main/examples

I thought maybe if the user didn't get good enough result, we can suggest them to try some other templates, rather than wait for the next release to enable this feature.

python/openai/openai_frontend/engine/utils/tool_call_parsers/llama_tool_call_parser.py
```diff
@@ -0,0 +1,289 @@
...
+            if backend != "vllm":
```
oandreeva-nv, line 86: if the behavior is trtllm specifi, do we want to change this to `backend == "tensorrtllm"` ?

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
oandreeva-nv: If we're re-using the vllm code, does it make sense to import those functions? In case something is changing, we'll simply inherit it from vllm

qa/L0_openai/generate_engine.py
richardhuo-nv: Yeah, I think we could. The caveat is tensorrt_llm._torch has a small portion of model architectures supported. 
https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/models/__init__.py
But seems all popular models are covered.

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
richardhuo-nv: Yeah, that would be ideal. But one issue is when we are using the trtllm image, the vllm package is not installed and the import will fail.

python/openai/openai_frontend/engine/utils/tool_call_parsers/utils.py
richardhuo-nv: Yeah, it's essentially for streaming. It basically checks if at some point, the tokens generated can not be part of json, which means the generation should be treated as normal text.

python/openai/openai_frontend/engine/utils/tool_call_parsers/llama_tool_call_parser.py
richardhuo-nv: Yeah, actually I use `backend == "tensorrtllm"` at the beginning, but find it doesn't work, and then find since the tensorrt_llm backend is not just one model, but 4 models (preprocessing, postprocessing, tensorrt_llm, tensorrt_llm_bls), only  tensorrt_llm has the backend "tensorrtllm". We are calling the model with tensorrt_llm_bls which is actually "python" backend. Unless we have some indicator to say the "python" backend tensorrt_llm_bls is actually a "tensorrtllm" model, I guess we have to stick to these kind of unclear if statement for a while.

python/openai/openai_frontend/engine/triton_engine.py
richardhuo-nv: https://community.openai.com/t/function-call-with-finish-reason-of-stop/437226
This is somehow a behavior from OpenAI, and seems they are not prioritize it for a fix. vLLM also adapted this behavior since a lot of agentic code already accept it and added the workaround. 
https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/serving_chat.py#L828

python/openai/openai_frontend/engine/chat_utils.py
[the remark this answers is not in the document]
rmccorm4: OK, if we keep the mix of types for simplicity for now, please add something like this for future:
```python
# FIXME: Converge on single set of types in either schemas.openai or openai.types
from openai.types.chat import ...
```

richardhuo-nv:
@rmccorm4 I found the issue that Tensorrt LLM streaming and non-streaming result not matching is likely caused by the latest 0.19.0dev version of Tensorrt LLM, even the openai client completion tests are start failing. When I keep the trtllm version as 0.18.0, the issue is gone. I will keep the testing version of trtllm and 0.18.0 and reproduce and file a bug to trtllm. In the meantime, I disabled the tests that require the guided decoding, which is not supported in 0.18.0 yet.

python/openai/openai_frontend/engine/triton_engine.py
oandreeva-nv: still need a comment?

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -299,44 +374,145 @@ def _get_first_streaming_chat_response(
...
+                        if auto_tools_called
+                        else 0
+                    )
+                else:
+                    index = 0
+
+                # check to make sure we haven't "forgotten" to stream
+                #   any tokens that were generated but previously
+                #   matched by partial json parsing, such as '}'.
+                # only happens if we are NOT using guided decoding
```
oandreeva-nv, line 446: nit:
```suggestion
                # check to make sure we haven't "forgotten" to stream
                # any tokens that were generated but previously
                # matched by partial json parsing, such as '}'.
                # only happens if we are NOT using guided decoding
```

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -299,44 +374,145 @@ def _get_first_streaming_chat_response(
...
+                    if auto_tools_called
+                    else ChatCompletionFinishReason.stop
+                )
+            else:
+                finish_reason = None
+
+            # if the response delta is None (e.g. because it was a
+            # "control token" for tool calls or the parser otherwise
+            # wasn't ready to send a token, then
+            #   get the next token without streaming a chunk
```
oandreeva-nv, line 500:
```suggestion
            # get the next token without streaming a chunk
```

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -371,6 +547,40 @@ def _validate_chat_request(
...
+
+        if (
+            request.tool_choice is None
+            and request.tools
+            and self.tool_call_parser is None
+        ):
+            raise Exception(
+                "having tools in the request requires --tool-call-parser to be set"
+            )
+
```
oandreeva-nv, line 583: Does it make sense to put this into a helper function, i.e. `verify_tool_call_settings` ?

python/openai/openai_frontend/engine/utils/chat.py
```diff
@@ -0,0 +1,167 @@
...
+
+    name: Optional[str]
+    """The name of the function to call"""
+
+    tool_calls: Optional[Iterable[ChatCompletionMessageToolCallParam]]
+    """The tool calls generated by the model, such as function calls."""
+
+
+_AssistantParser = partial(cast, ChatCompletionRequestAssistantMessage)
+_ToolParser = partial(cast, ChatCompletionRequestToolMessage)
```
oandreeva-nv, line 62: We only use each of those in one place, do we want to keep them global ?

python/openai/openai_frontend/engine/utils/chat.py
```diff
@@ -0,0 +1,167 @@
...
+    role: str, parts: List[ChatCompletionRequestMessageContentPart]
+) -> ConversationMessage:
+    content = list[Dict]()
+
+    for part in parts:
+        if part.root.type == Type1.text or part.root.type == "text":
+            parse_res = {"type": "text", "text": part.root.text}
+            content.append(parse_res)
+        else:
+            raise Exception(f"only text message is supported, but got {part.root.type}")
```
oandreeva-nv, line 88: nit: raise `TypeError`? https://docs.python.org/3/library/exceptions.html#TypeError

python/openai/openai_frontend/engine/utils/chat.py
```diff
@@ -0,0 +1,167 @@
...
+    for msg in messages:
+        sub_message = _parse_chat_message_content(msg)
+        conversation.append(sub_message)
+
+    _postprocess_messages(conversation)
+
+    return conversation
+
+
+def load_chat_template(chat_template) -> str:
```
oandreeva-nv, line 162:
```suggestion
def load_chat_template(chat_template) -> Optional[str]:
```

oandreeva-nv: seems like return None is also possible

python/openai/README.md
```diff
@@ -342,3 +342,228 @@ For more information on the `tritonfrontend` python bindings, see the docs
...
     - Not currently supported
+
+## Tool Calling
+
+The OpenAI frontend supports `tools` and `tool_choice` in the `v1/chat/completions` API. Please refer to the OpenAI API reference for more details about these parameters:
+  [tools](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools),
+  [tool_choice](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice)
+
+To enable the tool-calling feature, add the `--tool-call-parser {parser_name}` flag when starting the server. The two available parsers are `llama3` and `mistral`.
+The `llama3` parser supports tool-calling features for LLaMA 3.1, 3.2, and 3.3 models, while the `mistral` parser supports tool-calling features for the Mistral Instruct model.
```
oandreeva-nv, line 353: By any chance, have you tried llama4?

python/openai/openai_frontend/engine/triton_engine.py
oandreeva-nv: by the way, I think you can just 
```
self.chat_template = load_chat_template(chat_template)
```
seems like `load_chat_template` already handles case when `chat_template` is none

oandreeva-nv: and we also can add clarification to `load_chat_template` description

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -152,16 +191,52 @@ async def chat(
...
+            )
+            auto_tools_called = (
+                response_message.tool_calls is not None
+                and len(response_message.tool_calls.root) > 0
+            )
+        else:
+            response_message = ChatCompletionResponseMessage(
+                content=text, role=role, tool_calls=None
+            )
+
```
oandreeva-nv, line 229: I'd suggest to extract this logic to a separate function. This method is already big enough

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -152,16 +191,52 @@ async def chat(
         response = responses[0]
         text = _get_output(response)
 
+        auto_tools_called = False
+        tool_function_name = self._get_named_function_name(request=request)
+
+        response_message: Optional[ChatCompletionResponseMessage]
```
oandreeva-nv, line 197: can response_message be `None` ? I don't think in schema definition it's stated as `Optional`

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -299,44 +374,145 @@ def _get_first_streaming_chat_response(
...
+        tool_function_name = self._get_named_function_name(request=request)
+
+        # Determine whether tools are in use with "auto" tool choice
+        tool_choice_auto = (
+            tool_call_parser
+            and not tool_function_name
+            and self._should_stream_with_auto_tool_parsing(request)
+        )
+
+        previous_text = "" if tool_choice_auto else None
```
oandreeva-nv, line 398: Can we remove `if ...` given that we're only using this when `tool_choice_auto` is defined

python/openai/openai_frontend/engine/triton_engine.py
```diff
@@ -299,44 +374,145 @@ def _get_first_streaming_chat_response(
...
+
+            if finish_reason and response_delta is None:
+                response_delta = ChatCompletionStreamResponseDelta(content="")
+
+            choice = ChatCompletionStreamingResponseChoice(
+                index=0,
+                delta=response_delta,
+                logprobs=None,
+                finish_reason=finish_reason,
+            )
```
oandreeva-nv, line 512: I'd also suggest to extract the above logic to `_get_rsponse_delta` and `_get_finis_reason`, or whatever seems right

python/openai/openai_frontend/engine/utils/chat.py
oandreeva-nv: are we keeping this as str?

python/openai/openai_frontend/engine/utils/chat.py
oandreeva-nv: TypeError ?

rmccorm4 (commented):
Have Olga's comments been addressed? What's the status here?

richardhuo-nv:
> Have Olga's comments been addressed? What's the status here?

I think I addressed them. I've request Olga for a re-review.

rmccorm4:
Can you comment on her open threads above saying they've been addressed or add details otherwise?

richardhuo-nv:
> Can you comment on her open threads above saying they've been addressed or add details otherwise?

I thought I commented them. Is it somehow not show
image

python/openai/openai_frontend/engine/triton_engine.py
oandreeva-nv: Could you clarify this a bit if it was resolved?

rmccorm4:
They weren't submitted, see the `Pending` next to your name on each comment. You need to submit the batch of comments:
image

richardhuo-nv:
> They weren't submitted, see the `Pending` next to your name on each comment. You need to submit the batch of comments: image

omg, I see what's happening here...

richardhuo-nv:
@oandreeva-nv WARNING added for langchain. Thanks!

rmccorm4 (approved):

oandreeva-nv (approved):

PR #8174, opened 2025-04-28
fix: Update handling of large array sizes

mattwittwer:
#### What does the PR do?

This PR adds an additional parameter to customize the maximum size allocatable by incoming requests. This can be useful in limiting very large input arrays

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

Start with the new check added [here](https://github.com/triton-inference-server/server/blob/mwittwer/shape_array_vulnerability/src/http_server.cc#L2812-L2822). Then review the additional input added to main to set this value. 

#### Test plan:

New test cases added for json and raw binary inputs

- CI Pipeline ID: 29618182

- closes GitHub issue: #xxx


yinggeh:
Test case?

src/command_line_parser.cc
```diff
@@ -1393,6 +1397,9 @@ TritonParser::Parse(int argc, char** argv)
         case OPTION_HTTP_THREAD_COUNT:
           lparams.http_thread_cnt_ = ParseOption<int>(optarg);
           break;
+        case OPTION_HTTP_MAX_INPUT_SIZE:
+          lparams.http_max_input_size_ = ParseOption<int>(optarg);
```
rmccorm4, line 1401: Should probably test and check for negative and zero

src/http_server.cc
```diff
@@ -2757,6 +2759,17 @@ HTTPAPIServer::ParseJsonTritonIO(
...
           }
 
+          // Check if byte_size is larger than max_input_size_
+          if (byte_size > max_input_size_) {
+            return TRITONSERVER_ErrorNew(
+                TRITONSERVER_ERROR_INVALID_ARG,
+                ("input '" + std::string(input_name) +
+                 "' has a byte_size that exceeds the maximum allowed value "
+                 "of " +
+                 std::to_string(max_input_size_))
```
yinggeh, line 2769: Suggest user to increase option `--http-max-input-size`.

src/command_line_parser.cc
mattwittwer: Added check to ensure only positive values can be provided: https://github.com/triton-inference-server/server/blob/mwittwer/shape_array_vulnerability/src/command_line_parser.cc#L1402

src/http_server.cc
mattwittwer: Updated error message to include a prompt for this value: https://github.com/triton-inference-server/server/blob/mwittwer/shape_array_vulnerability/src/http_server.cc#L2813-L2823

qa/L0_http/test.sh
```diff
@@ -753,6 +753,100 @@ set -e
...
+SERVER_ARGS="--model-repository=${MODELDIR} --http-max-input-size=-1024"
+SERVER_LOG="./inference_server_negative_limit.log"
+CLIENT_LOG="./http_input_size_limit_negative.log"
+run_server
+if [ "$SERVER_PID" != "0" ]; then
+    echo -e "\n***\n*** Server should not start with negative max input size\n***"
+    kill $SERVER_PID
+    wait $SERVER_PID
+    RET=1
+fi
```
rmccorm4, line 848: Can you grep for partial expected error message in server log in these two zero/negative test cases? ex: `--http-max-input-size must be greater than 0`

Wouldn't want them to pass the test just because of a typo with bad model repo or something causing server startup to fail for some other reason.

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -0,0 +1,315 @@
...
+            result["outputs"][0]["shape"][1],
+            f"Expected shape {[1, shape_size]}, got {result['outputs'][0]['shape']}",
+        )
+
+    def test_large_input_raw_binary(self):
+        """Test raw binary input larger with custom limit set"""
+        model = "onnx_zero_1_float32"
+
+        # Test case 1: Input just over our configured limit (should fail)
+        large_input = np.ones(2**25 + 32, dtype=np.float32)  # Just over 128MB
```
rmccorm4, line 183: In `test.sh` it looks like this is set to `2**27` (128MB) but here in the pytest on this line, it looks like `2**25`=32MB + ~32bytes. 

Why is this failing?

rmccorm4: Are you actually seeing this "test case 1" fail when sending a payload of ~32MB when the limit is custom set to 128MB? Or am I misreading something?

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -0,0 +1,315 @@
...
+        # Verify error message contains size limit info
+        error_msg = response.content.decode()
+        self.assertIn(
+            "exceeds the maximum allowed value",
+            error_msg,
+            "Expected error message about exceeding max input size",
+        )
+
+        # Test case 2: Input just under our configured limit (should succeed)
+        small_input = np.ones(2**25 - 32, dtype=np.float32)  # Just under 128MB
```
rmccorm4, line 209: Same here 2^25 == 32MB, but limit set to 2^27 == 128MB? In this "just under limit" scenario, it should still work, but should update the values to actually be "just under" the limit and match the server-side config of ~2^27 instead of ~2^25.

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -0,0 +1,315 @@
...
+            np.array_equal(output_array, small_input),
+            "Response data does not match input data",
+        )
+
+    def test_large_input_json(self):
+        """Test JSON input larger with custom limit set"""
+        model = "onnx_zero_1_float32"
+
+        # Test case 1: Input just over our configured limit (should fail)
+        shape_size = 2**25 + 32  # Just over 128MB
```
rmccorm4, line 241: Same here 2^25 vs 2^27

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -0,0 +1,315 @@
...
+        # Verify error message contains size limit info
+        error_msg = response.content.decode()
+        self.assertIn(
+            "exceeds the maximum allowed value",
+            error_msg,
+            "Expected error message about exceeding max input size",
+        )
+
+        # Test case 2: Input just under our configured limit (should succeed)
+        shape_size = 2**25 - 32  # Just under 128MB
```
rmccorm4, line 277: Same here 2^25 vs 2^27

qa/L0_http/http_input_size_limit_test.py
rmccorm4: Oh - I think this is because its 32MB elements * 4bytes (fp32) each. Please update the comment to make this more clear.

qa/L0_http/http_input_size_limit_test.py
rmccorm4: https://github.com/triton-inference-server/server/pull/8174/files#r2116850655

qa/L0_http/http_input_size_limit_test.py
rmccorm4: https://github.com/triton-inference-server/server/pull/8174/files#r2116850655

qa/L0_http/http_input_size_limit_test.py
rmccorm4: https://github.com/triton-inference-server/server/pull/8174/files#r2116850655

rmccorm4 (approved):
PR looks functionally sound - left some comment on improving test stability and readability

qa/L0_http/http_input_size_limit_test.py
mattwittwer: https://github.com/triton-inference-server/server/pull/8174#discussion_r2117012165

Updated: https://github.com/triton-inference-server/server/blob/mwittwer/shape_array_vulnerability/qa/L0_http/http_input_size_limit_test.py#L242-L243

rmccorm4 (approved):

yinggeh:
Just a thought, people will set `--http-max-input-size` mostly because they need payload size greater than 128MB. From UX perspective to make their life easier, shouldn't the unit be MB instread of bytes?

mattwittwer:
> Just a thought, people will set `--http-max-input-size` mostly because they need payload size greater than 128MB. From UX perspective to make their life easier, shouldn't the unit be MB instread of bytes?

@yinggeh - This is definitely a good option and may improve readability. Setting the minimum unit size to be MB would mean the user wont be able to set smaller amounts but that probably isn't a common use case. I'll work on making this change

rmccorm4:
All of our other fields use bytes and it has the most flexibility in case users want to set small sizes. Let's keep as bytes for now just to keep this field consistent with all other size related fields please. @mattwittwer @yinggeh 

We can consider MB support as a separate follow-up if users request it (for example `--http-max-input-size 128MB` and parse the size qualifier if any, but I'd like to stay consistent for now if there's no explicit request.

mattwittwer:
@rmccorm4 - Restored the input value to bytes

rmccorm4 (approved):

PR #8191, opened 2025-05-07
test: L0_orca_trtllm fixed

indrajit96:
#### What does the PR do?
Fix L0_orca_trtllm which was broken due to new changes in the trtllm directory structure.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [X] test

test.sh

#### Test plan:
NA

- CI Pipeline ID:
27767229

#### Caveats:
NA

#### Background
Test broke due to 
1. CI runners were suddenly unable to find func definitions (RCA unknown)
2. Change in TRTLLM directory structure which resulted in trtllm-build failing.

### DIFF
--- a/qa/L0_orca/test.sh
+++ b/qa/L0_orca/test.sh
@@ -36,7 +36,7 @@ MODEL_NAME="gpt2_tensorrt_llm"
 NAME="tensorrt_llm_benchmarking_test"
 MODEL_REPOSITORY="$(pwd)/triton_model_repo"
 TENSORRTLLM_BACKEND_DIR="/workspace/tensorrtllm_backend"
-GPT_DIR="$TENSORRTLLM_BACKEND_DIR/tensorrt_llm/examples/gpt"
+GPT_DIR="$TENSORRTLLM_BACKEND_DIR/tensorrt_llm/examples/models/core/gpt"
 TOKENIZER_DIR="$GPT_DIR/gpt2"
 ENGINES_DIR="${BASE_DIR}/engines/inflight_batcher_llm/${NUM_GPUS}-gpu"
 TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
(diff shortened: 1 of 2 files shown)


qa/L0_orca/test.sh
```diff
@@ -138,6 +145,42 @@ function kill_server {
...
+        --gemm_plugin float16 \
+        --workers "${NUM_GPUS}" \
+        --output_dir "${ENGINES_DIR}"
+
+    cd ${BASE_DIR}
+}
+
 clone_tensorrt_llm_backend_repo
 build_gpt2_base_model
 build_gpt2_tensorrt_engine
```
rmccorm4, line 186: How did this call to `build_gpt2_tensorrt_engine` previously work if the function didn't exist until this PR?

indrajit96: I tried finding an answer to this, but was unable to
Checked with @mc-nv if anything changed in the way CI runs or mounts. No luck.
Double checked if the test did actually run as expected before it broke (It did)
Test started breaking when we merged 24.04 release branch into main.

rmccorm4: > They were originally ONLY defined in L0_perf_tensorrt_llm/test.sh and prev tests used to pick the defn from there

Do we do a `source L0_perf_tensorrt_llm/test.sh` somewhere when running this test?

yinggeh: Maybe because there is no `+e` so bash just ignore the error.

mc-nv (dismissed):

qa/L0_orca/test.sh
```diff
@@ -138,6 +145,42 @@ function kill_server {
     done
 }
 
+function clone_tensorrt_llm_backend_repo {
```
yinggeh, line 148: In the previous PR, I suggested moving these functions to `qa/common/util.sh` but somehow they were deleted instead of moved.
https://github.com/triton-inference-server/server/pull/8009#discussion_r1956768205

PR #8220, opened 2025-05-23
fix: Improve cancellation handling for gRPC non-decoupled inference

pskiran1:
#### What does the PR do?

This PR is intended to fix the following issue:
- In gRPC inference for a non-decoupled model, where the response and final flag are sent separately, if the client cancels the request before the final flag is sent (while sending the response), the related `response_release_payload` will be deleted during the cancellation handling. This causes a problem because the next call to `InferResponseComplete` (sending the final flag) will attempt to use a state that no longer exists.

Made changes to the existing test cases to cover more test data. Also added new test cases, `test_grpc_async_cancel` and `test_grpc_async`, to test the code changes.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 28983978

- closes GitHub issue: #xxx

### DIFF
--- a/src/grpc/infer_handler.cc
+++ b/src/grpc/infer_handler.cc
@@ -1068,12 +1068,14 @@ ModelInferHandler::InferResponseComplete(
     state->cb_count_++;
   }
 
+  bool is_complete = (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) != 0;
+
   LOG_VERBOSE(1) << "ModelInferHandler::InferResponseComplete, "
                  << state->unique_id_ << " step " << state->step_;
 
   // Allow sending 1 response and final flag separately, only mark
   // non-inflight when seeing final flag
-  if (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) {
+  if (is_complete || state->IsGrpcContextCancelled()) {
     state->context_->EraseInflightState(state);
   }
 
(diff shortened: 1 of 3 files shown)


src/grpc/infer_handler.cc
```diff
@@ -1068,12 +1068,14 @@ ModelInferHandler::InferResponseComplete(
     state->cb_count_++;
   }
 
+  bool is_complete = (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) != 0;
```
yinggeh, line 1071:
```suggestion
  bool is_final_response = (flags & TRITONSERVER_RESPONSE_COMPLETE_FINAL) != 0;
```

src/grpc/infer_handler.cc
```diff
@@ -1085,30 +1087,30 @@ ModelInferHandler::InferResponseComplete(
         TRITONSERVER_InferenceResponseDelete(iresponse),
         "deleting GRPC inference response");
 
-    state->context_->EraseInflightState(state);
```
yinggeh, line 1088: Let's revert this change and keep cancellation handling logic inside one block.

Issue #7796, opened 2024-11-14
Problems with the response of the OpenAI-Compatible Frontend for Triton Inference Server


DimadonDL:
Hi,

i have installed Triton with vllm backend and also the OpenAI-Compatible Frontend for Triton Inference Server (Beta). The model is meta-llama/Llama-3.1-8B-Instruct. Now when I call the Endpoint for example like this:

MODEL="llama-3.1-8b-instruct"
curl -s http://localhost:9000/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model": "'${MODEL}'",
  "messages": [{"role": "user", "content": "Why is the sky blue?"}]
}' | jq

The Response is:

`{
  "id": "cmpl-276d1a84-a293-11ef-b088-d404e69cb4ea",
  "choices": [
    {
      "finish_reason": "stop",
      "index": 0,
      "message": {
        "content": "The sky appears blue because of a phenomenon called Rayleigh scattering, named after the",
        "tool_calls": null,
        "role": "assistant",
        "function_call": null
      },
      "logprobs": null
    }
  ],
  "created": 1731593837,
  "model": "vllm_model",
  "system_fingerprint": null,
  "object": "chat.completion",
  "usage": null
}`

As you can see the content is cropped. I have played with the config but I don't know what's the problem. With Python the response is fine:

`from openai import OpenAI

client = OpenAI(
    base_url="http://localhost:9000/v1",
    api_key="EMPTY",
)

model = "vllm_model"
completion = client.chat.completions.create(
    model=model,
    messages=[
        {
            "role": "system",
            "content": "You are a helpful assistant.",
        },
        {"role": "user", "content": "Why is the sky blue?"},
    ],
    max_tokens=4096,
)

print(completion.choices[0].message.content)`

The response here is:

The sky appears blue due to a phenomenon called Rayleigh scattering. This is a scientific explanation:

1.  **Sunlight and the Atmosphere**: When sunlight enters the Earth's atmosphere, it encounters tiny molecules of gases such as nitrogen (N2) and oxygen (O2). These molecules are much smaller than the wavelength of light.

2.  **Scattering of Light**: According to the Rayleigh scattering theory, when light travels through the atmosphere, it encounters these tiny molecules. The shorter (blue) wavelengths of light are scattered more than the longer (red) wavelengths. This scattering of light in all directions is what gives the sky its blue color.

[...]

PR #8226, opened 2025-05-30
fix: Support max_completion_tokens option in OpenAI frontend

pskiran1:
#### What does the PR do?

- Added support for `max_completion_tokens`.  
- Did not remove `max_tokens` support for the time being to maintain backward compatibility.  
- Added test cases.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 30370160

#### Background

As mentioned in the API documentation, `max_tokens` is deprecated, and it is recommended to use `max_completion_tokens` instead. https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens

- Relates to GitHub issue: #7796


python/openai/tests/test_chat_completions.py
```diff
@@ -199,14 +201,21 @@ def test_chat_completions_invalid_sampling_parameters(
         assert response.status_code == 422
 
     # Simple tests to verify max_tokens roughly behaves as expected
+    @pytest.mark.parametrize(
+        "param_key",
```
rmccorm4: Let's call this something a little more intuitive to read, like `max_tokens_key` or something.

Reading this isn't clear at a glance what kind of param it is for:
```python
        payload[param_key] = 1
```

python/openai/openai_frontend/schemas/openai.py
```diff
@@ -850,11 +842,15 @@ class CreateChatCompletionRequest(BaseModel):
         None,
         description="An integer between 0 and 20 specifying the number of most likely tokens to return at each token position, each with an associated log probability. `logprobs` must be set to `true` if this parameter is used.",
     )
-    # TODO: Consider new max_completion_tokens field in the future: https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens
-    max_tokens: Optional[conint(ge=0)] = Field(
-        16,
+    max_completion_tokens: Optional[conint(ge=0)] = Field(
+        None,
```
richardhuo-nv, line 846: Will trtllm backend or vllm backend has a default number for this? What's the behavior of if we leave max_tokens and max_completion_tokens both as None? Will it only generating one token? If so, I think it's better to have a default value here so that at least the user can get something.

python/openai/openai_frontend/schemas/openai.py
```diff
@@ -81,7 +81,7 @@ class CreateCompletionRequest(BaseModel):
     )
     best_of: Optional[conint(ge=0, le=20)] = Field(
         1,
-        description='Generates `best_of` completions server-side and returns the "best" (the one with the highest log probability per token). Results cannot be streamed.\n\nWhen used with `n`, `best_of` controls the number of candidate completions and `n` specifies how many to return – `best_of` must be greater than `n`.\n\n**Note:** Because this parameter generates many completions, it can quickly consume your token quota. Use carefully and ensure that you have reasonable settings for `max_tokens` and `stop`.\n',
+        description='Generates `best_of` completions server-side and returns the "best" (the one with the highest log probability per token). Results cannot be streamed.\n\nWhen used with `n`, `best_of` controls the number of candidate completions and `n` specifies how many to return – `best_of` must be greater than `n`.\n\n**Note:** Because this parameter generates many completions, it can quickly consume your token quota. Use carefully and ensure that you have reasonable settings for `max_completion_tokens` and `stop`.\n',
     )
```
richardhuo-nv: Since completion API is still only using the max_tokens, maybe we can keep the description as max_tokens.

python/openai/openai_frontend/schemas/openai.py
pskiran1: For `vLLM`, there is no issue since the [framework has a default value](https://docs.vllm.ai/en/v0.6.4/dev/sampling_params.html#:~:text=max_tokens%3A%20int%20%7C%20None%20%3D%2016).
However, for `TRT-LLM`, the `max_tokens` [field is mandatory](https://github.com/triton-inference-server/server/pull/8226/commits/5e2756ee6fff74a8948b5ed84201e622596f2e4a), so we need to establish a default value.

If I am not wrong, configuring the default value in the [schema](https://github.com/triton-inference-server/server/blob/main/python/openai/openai_frontend/schemas/openai.py) may create confusion in identifying whether the user provided `max_tokens` or `max_completion_tokens`, as both fields would always have values. To avoid this ambiguity and ensure that the request field is correctly identified, I moved the default value for `max_tokens` to the command line arguments. commit: https://github.com/triton-inference-server/server/pull/8226/commits/5e2756ee6fff74a8948b5ed84201e622596f2e4a
This change also allows users to customize the default value when starting the server.

Please let me know if this approach is feasible or if you have any suggestions.
Thank you.
cc: @rmccorm4

richardhuo-nv (dismissed):

richardhuo-nv (approved):

PR #8297, opened 2025-07-16
TPRD-1627: update validation condition

mc-nv:
This change motivation is to resolve the exception that may appear during the condition validation.

### DIFF
--- a/qa/common/test_util.py
+++ b/qa/common/test_util.py
@@ -310,9 +310,9 @@ def check_gpus_compute_capability(min_capability):
         bool
     """
 
-    import sys
+    import importlib.util
 
-    if "cuda" in sys.modules:
+    if importlib.util.find_spec("cuda") is not None:
         import cuda.core.experimental as cuda_core_experimental
 
         devices = cuda_core_experimental.system.devices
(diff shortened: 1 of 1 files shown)


qa/common/test_util.py
```diff
@@ -312,8 +312,13 @@ def check_gpus_compute_capability(min_capability):
 
     import importlib.util
 
-    if importlib.util.find_spec("cuda.core.experimental"):
-        import cuda.core.experimental as cuda_core_experimental
+    if importlib.util.find_spec("cuda") is not None:
+        try:
+            import cuda.core.experimental as cuda_core_experimental
+        except ImportError:
+            print(
```
yinggeh: I think it's better to throw exception than slience the import error.

qa/common/test_util.py
```diff
@@ -325,8 +330,11 @@ def check_gpus_compute_capability(min_capability):
             if compute_capability < min_capability:
                 return False
 
-    elif importlib.util.find_spec("pycuda.driver"):
-        import pycuda.driver as cuda
+    elif importlib.util.find_spec("pycuda") is not None:
```
yinggeh, line 328: Need an else case to catch neither library is installed case, and return an error/throw exception.

yinggeh (approved):
LGTM!

PR #8327, opened 2025-08-05
Update versions for 25.08

mc-nv:


qa/common/gen_jetson_trt_models
```diff
@@ -142,6 +142,7 @@ docker pull $TENSORRT_IMAGE
 
 docker run $DOCKER_GPU_ARGS \
    --rm -v $DOCKER_VOLUME:/mnt \
+   -e TRT_VERBOSE \
```
yinggeh, line 145: Why do we need this?

qa/common/gen_qa_trt_data_dependent_shape.py
```diff
@@ -45,7 +45,13 @@ def create_data_dependent_modelfile(
     trt_input_dtype = np_to_trt_dtype(input_dtype)
 
     # Create the model
-    TRT_LOGGER = trt.Logger(trt.Logger.INFO)
+    TRT_LOGGER = (
+        trt.Logger(trt.Logger.INFO)
+        if os.environ.get("TRT_VERBOSE") != "1"
+        else trt.Logger(trt.Logger.VERBOSE)
+        if os.environ.get("TRT_VERBOSE") != "1"
+        else trt.Logger(trt.Logger.VERBOSE)
```
yinggeh: Remove duplicates

build.py
```diff
@@ -1494,7 +1494,7 @@ def dockerfile_prepare_container_linux(argmap, backends, enable_gpu, target_mach
             cp -r nvpl_slim_24.04/include/* /usr/local/include && \\
             rm -rf nvpl_slim_24.04.tar nvpl_slim_24.04; \\
         fi \\
-        && pip3 install --no-cache-dir --progress-bar on --index-url $VLLM_INDEX_URL -r /run/secrets/requirements \\
```
yinggeh, line 1497: Why was `--progress-bar on` removed?

mc-nv: Because it make no sense there.

qa/common/gen_jetson_trt_models
mc-nv: Interesting question, aren't particular flag is self explaining?

mc-nv: This value [Set environment variables (-e, --env, --env-file)](https://docs.docker.com/reference/cli/docker/container/run/#env) and purpose to be able pass that variable to the build if necessary.

yinggeh (approved):

PR #8346, opened 2025-08-13
Introduce global variable to handle CMake scope issue

mc-nv:


qa/L0_client_build_variants/test.sh
```diff
@@ -54,6 +54,8 @@ TRITON_BACKEND_REPO_TAG=${TRITON_BACKEND_REPO_TAG:="main"}
 TRITON_CORE_REPO_TAG=${TRITON_CORE_REPO_TAG:="main"}
 TRITON_COMMON_REPO_TAG=${TRITON_COMMON_REPO_TAG:="main"}
 
+export CMAKE_POLICY_VERSION_MINIMUM=3.5
```
yinggeh, line 57: Is there a better way than adding this line to each individual test?

mc-nv: It's a quick fix here, like everything we do in our projects, unfortunately. 
Reason for current failure is that we have CMake processes that may come out of parent scope, and it picks global variables. 
By defining global variable it should work as expected but let's see.

PR #8353, opened 2025-08-15
test: Fix CuPy CUDA library link

kthui:
Fix for CuPy CUDA 13 library link.

This PR targets the r25.08 branch directly. Pipeline: 33318849

### DIFF
--- a/python/openai/README.md
+++ b/python/openai/README.md
@@ -35,7 +35,7 @@
 ## Pre-requisites
 
 > [!WARNING]
-> **CuPy CUDA 13 Compatibility Issue**: The Triton Inference Server Image has been upgraded to CUDA 13. The latest CuPy wheel released at the time of writing is not fully compatible with CUDA 13. If you encounter issues with CuPy, they may be fixed by running the following commands before starting the Frontend:
+> **CuPy CUDA 13 Compatibility Issue**: The Triton Inference Server Image has been upgraded to CUDA 13. You may encounter issues when using CuPy before it officially supports CUDA 13 (see [this issue](https://github.com/cupy/cupy/issues/9286) requesting CUDA 13 support). Some issues may be resolved by linking CUDA 12 shared objects to CUDA 13, for example:
 > ```bash
 > ln -sf /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.13.0.48 /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.12
 > export LD_LIBRARY_PATH="/usr/local/cuda/targets/x86_64-linux/lib:$LD_LIBRARY_PATH"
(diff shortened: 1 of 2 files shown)


qa/L0_python_api/test.sh
```diff
@@ -27,6 +27,11 @@
 
 pip3 install pytest-asyncio==0.23.8
 
+# Create CUDA compatibility symlink for CuPy
+# TODO: Remove patch once CuPy supports CUDA 13
+ln -sf /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.13.0.48 /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.12
```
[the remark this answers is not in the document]
kthui: This is temporary, we expect CuPy will release a new version that supports CUDA 13 in the near future.

python/openai/README.md
```diff
@@ -34,6 +34,13 @@
 
 ## Pre-requisites
 
+> [!WARNING]
+> **CuPy CUDA 13 Compatibility Issue**: The Triton Inference Server Image has been upgraded to CUDA 13. The latest CuPy wheel released at the time of writing is not fully compatible with CUDA 13. If you encounter issues with CuPy, they may be fixed by running the following commands before starting the Frontend:
+> ```bash
+> ln -sf /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.13.0.48 /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.12
```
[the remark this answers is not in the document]
kthui: This is temporary, we expect CuPy will release a new version that supports CUDA 13 in the near future.

mc-nv, line 40: This probably won't work if job using compiler operations. 
Compiler will look for `libnvrtc.so.12` SONAME in order to resolve dependency but library `libnvrtc.so.13.0.48` will have name `libnvrtc.so.13`: 
```
root@ad4f880c5b30:/opt/tritonserver/qa/L0_batcher_cudashm# readelf -a /usr/local/cuda-13.0/targets/x86_64-linux/lib/libnvrtc.so.13.0.48 | grep SON
 0x000000000000000e (SONAME)             Library soname: [libnvrtc.so.13]
```

kthui: We don't compile CuPy, it comes from a pip wheel installed only when we are installing tritonfrontend wheel. We just need to make sure it links during runtime and that is sufficient here.

mc-nv: Then may be it better to download necessary file and make it available?
https://developer.download.nvidia.com/compute/cuda/redist/cuda_nvrtc/linux-x86_64/

kthui: Can you apply the fix?

mc-nv: Let me try it locally.

PR #8360, opened 2025-08-19
fix: WAR for Python CUDA library unknown race condition

GuanLuo:
Thanks for submitting a PR to Triton!
Please go the the `Preview` tab above this description box and select the appropriate sub-template:

* [PR description template for Triton Engineers](?expand=1&template=pull_request_template_internal_contrib.md)
* [PR description template for External Contributors](?expand=1&template=pull_request_template_external_contrib.md)

If you already created the PR, please replace this message with one of
* [External contribution template](https://raw.githubusercontent.com/triton-inference-server/server/main/.github/PULL_REQUEST_TEMPLATE/pull_request_template_external_contrib.md)
* [Internal contribution template](https://raw.githubusercontent.com/triton-inference-server/server/main/.github/PULL_REQUEST_TEMPLATE/pull_request_template_internal_contrib.md)

and fill it out.

PR #8361, opened 2025-08-19
fix: WAR for Python CUDA library unknown race condition

mc-nv:
Picking changes from https://github.com/triton-inference-server/server/pull/8360

### DIFF
--- a/qa/common/shm_util.py
+++ b/qa/common/shm_util.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2018-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


qa/common/shm_util.py
```diff
@@ -94,64 +97,69 @@ def create_set_shm_regions(
...
 
-    if use_system_shared_memory:
-        shm_ip0_handle = shm.create_shared_memory_region(
-            shm_region_names[0] + "_data", "/" + shm_region_names[0], input0_byte_size
-        )
-        shm_ip1_handle = shm.create_shared_memory_region(
-            shm_region_names[1] + "_data", "/" + shm_region_names[1], input1_byte_size
-        )
+    with CREATION_LOCK:
+        if use_system_shared_memory:
```
yinggeh, line 101: Also left the same comment in the original PR https://github.com/triton-inference-server/server/pull/8360. Why do we change the behavior of system shared memory if there is no issue?

GuanLuo (approved):

krishung5 (approved):

PR #8374, opened 2025-09-05
feat: Static key authentication for OpenAI frontend

yinggeh:
#### What does the PR do?
A similar feature to Triton [limit-endpoint-access](https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/customization_guide/inference_protocols.html#limit-endpoint-access-beta).

### Limit Endpoint Access

The OpenAI-compatible server supports restricting access to specific API endpoints through authentication headers. This feature allows you to protect sensitive endpoints while keeping others publicly accessible.

#### Configuration

Use the `--openai-restricted-api` command-line argument to configure endpoint restrictions:

```
--openai-restricted-api <API_1>,<API_2>,... <restricted-key> <restricted-value>
```

- **`API`**: A comma-separated list of APIs to be included in this group. Note that currently a given API is not allowed to be included in multiple groups. The following protocols / APIs are recognized:
  - **inference**: Chat completions and text completions endpoints
    - `POST /v1/chat/completions`
    - `POST /v1/completions`
  - **model-repository**: Model listing and information endpoints
    - `GET /v1/models`
    - `GET /v1/models/{model_name}`
  - **metrics**: Server metrics endpoint
    - `GET /metrics`
  - **health**: Health check endpoint
    - `GET /health/ready`

- **`restricted-key`**: The HTTP request header to be checked when a request is received.
- **`restricted-value`**: The header value required to access the specified protocols.

#### Checklist

#### Commit Type:
- [x] feat

#### Test plan:
L0_openai_vllm--base

- CI Pipeline ID:
34741599

- closes GitHub issue: #xxx


python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
```diff
@@ -0,0 +1,271 @@
...
+
+                # Check for duplicate API categories across different arguments
+                if self.IsRestricted(endpoint):
+                    raise ValueError(
+                        f"restricted api '{endpoint}' can not be specified in multiple config groups"
+                    )
+
+                self.Insert(endpoint, (key, value))
+
+    def RestritionDict(self) -> dict[str, tuple[str, str]]:
```
nnshah1:
```suggestion
    def RestrictionDict(self) -> dict[str, tuple[str, str]]:
```

python/openai/openai_frontend/main.py
```diff
@@ -162,6 +164,13 @@ def parse_args():
         choices=["debug", "info", "warning", "error", "critical", "trace"],
         help="log level for uvicorn",
     )
+    openai_group.add_argument(
```
nnshah1, line 166: since we are using arg parse - could we potentially use a syntax for lists or maybe a different way of parsing to avoid needing to do our own string processing - ?

python/openai/openai_frontend/frontend/fastapi_frontend.py
```diff
@@ -107,3 +115,11 @@ def _add_cors_middleware(self, app: FastAPI):
             allow_methods=["*"],
             allow_headers=["*"],
         )
+
+    def _add_api_restriction_middleware(self, app: FastAPI):
+        app.add_middleware(
+            APIRestrictionMiddleware, restricted_apis=self.restricted_apis
+        )
+        print(
```
nnshah1, line 128: should we use logging?

yinggeh: That's a good point. Currently we don't have logger in OAI frontend outside "tritonserver" and "FastAPI". It was also annotated here https://github.com/triton-inference-server/server/blob/f0b3d9c14a19fb55de3e4c1cf5161ae9254ddf5c/python/openai/openai_frontend/frontend/fastapi_frontend.py#L109-L110 I can create a ticket to track the effort to bring in logger to OAI frontend but that's outside the scope of this PR.

python/openai/openai_frontend/main.py
yinggeh: Hi @nnshah1. I am not really following. Can you give an example?
If you mean providing a custom parsing function to `type` like
```
openai_group.add_argument(
    "--openai-restricted-api",
    type=parse_restriction_config,
    action="append",
)
```
then it's possible. However, each config argument is still parsed individually. Eventually we still need to traverse all configs again to report any conflicts (e.g. --openai-restricted-api="inference:key1=value1" --openai-restricted-api="inference:key2=value2").

nnshah1: I was thinking of also using nargs 2

```
mport argparse

parser = argparse.ArgumentParser(description='Process pairs of values.')

parser.add_argument(
    '--pair',
    nargs=2,
    action='append',
    metavar=('KEY', 'VALUE'),
    help='Specify a key-value pair. Can be given multiple times.'
)

args = parser.parse_args()

if args.pair:
    print("Received pairs:")
    for key, value in args.pair:
        print(f"  Key: {key}, Value: {value}")
else:
    print("No pairs provided.")
```

nnshah1: just to avoid the string parsing - but not a blocker

yinggeh: Used nargs=3 to avoid all parsing work. The expected input arg will be `--openai-restricted-api <API_1>,<API_2>,... <restricted-key> <restricted-value>`

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
```diff
@@ -0,0 +1,224 @@
...
+
+        Args:
+            request: The incoming HTTP request
+            auth_header: Tuple of (expected_header_key, expected_header_value)
+
+        Returns:
+            dict: {"valid": bool, "message": str} - Authentication result and error message if invalid
+        """
+        expected_key, expected_value = auth_header
+        print(f"expected_key: {expected_key}")
```
nnshah1: I think we would leave this out until we get logging levels - @rmccorm4 , @whoisj  what do you think?

nnshah1 (approved):
LGTM - @kthui , @whoisj

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
yinggeh: Sorry that was for debugging. I have removed that.

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
```diff
@@ -0,0 +1,222 @@
...
+        for (
+            restricted_api,
+            auth_spec,
+        ) in self.restricted_apis.RestrictionDict().items():
+            # Check each endpoint in the API
+            for restricted_endpoint in ENDPOINT_MAPPING[restricted_api]:
+                restricted_method, restricted_path = restricted_endpoint.split(" ")
+
+                # Match both HTTP method and path prefix
+                if request_method == restricted_method and request_path.startswith(
```
whoisj, line 156: URL are case insensitive, how is that being handled?

if I restrict `v1/models` and the incoming request is for `v1/MODELS` how is that treated?
in theory, the HTTP server should honor the request regardless of casing.
does that mean it is not restricted?

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # Check authentication for the matching restricted endpoint
+        auth_result = self._check_authentication(request, auth_header)
+        if auth_result["valid"]:
+            # Authentication passed, allow request to proceed
+            return await call_next(request)
+        else:
+            # Authentication failed, return 401 error
+            return JSONResponse(
+                status_code=401,
```
whoisj, line 188: `HTTP401` means authentication required.
whereasd `HTTP403` means authentication provieded, but insufficient.

either we return the correct code here, or take the safer approach and return `HTTP404` when either unathenticated or incorrectly authenticated.

whoisj (commented):
Mostly looks good, but I have a couple of questions before approval.

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
yinggeh: I don't think our FastAPI supports URL case-insensitivity.
```
INFO:     127.0.0.1:33054 - "GET /v1/models HTTP/1.1" 200 OK
INFO:     127.0.0.1:33068 - "GET /V1/MODELS HTTP/1.1" 404 Not Found
```

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
yinggeh: In Triton server frontend, we send response `This API is restricted, expecting header '<restricted-key>'` in either case (user did not send correct header key or set the wrong header value). I will vote for 401 and 403 to differentiate them in OAI frontend.
https://github.com/triton-inference-server/server/blob/4ebd170feedbe2f011876083b6826feaa705b6db/src/http_server.cc#L4831-L4837

yinggeh: Actually, we don't disable an endpoint. When do we return a 403?

401 Unauthorized:
This status code means that the client (user or application) has not provided valid authentication credentials for the requested resource. The server requires authentication, but the client either didn't send any credentials or sent invalid ones (e.g., incorrect username/password, expired token).

whoisj: if this is how Triton Server behaves today, then I am 100% fine with replicating that here.

typically, the HTTP response codes are as follows:
- `HTTP401` authentication required.
- `HTTP403` invalid authentication provided.

OR 

- `HTTP404` resource not found

The use of `HTTP404` instead of one of the more "useful" codes is avoid API scanning. When ever unauthenticated request returns the "not found" status, there's no mechanism for scanning which secure APIs exist.

python/openai/openai_frontend/frontend/fastapi/middleware/api_restriction.py
whoisj: yes, I did my research here and by default both FastAPI and OpenAI's URL implementations are case-sensitive.

thanks for following up.

PR #8377, opened 2025-09-08
Update "cuda-cupy" package

mc-nv:
`cuda-cupy` package has been released with support of CUDA 13. 
Updating configuration in order to confirm that the package can be used.

related to: https://github.com/triton-inference-server/core/pull/450

### DIFF
--- a/python/openai/README.md
+++ b/python/openai/README.md
@@ -34,13 +34,6 @@
 
 ## Pre-requisites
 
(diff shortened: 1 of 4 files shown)


yinggeh:
We should revert https://github.com/triton-inference-server/server/pull/8361 if the race condition has been fixed.

PR #8400, opened 2025-09-23
fix: Ensure model name cannot be used to escape model repository

whoisj:
This change adds a check to ensure that when model name is appended to the model-repository path that the combined path does not escape the model repository.

Linear [TRI-54](https://linear.app/nvidia/issue/TRI-54/os-access-violation-via-path-access)
CI Pipeline #35390548

### DIFF
--- a/python/openai/openai_frontend/engine/utils/triton.py
+++ b/python/openai/openai_frontend/engine/utils/triton.py
@@ -362,6 +362,11 @@ def _get_vllm_lora_names(
         repo_paths = [repo_paths]
     for repo_path in repo_paths:
         model_path = os.path.join(repo_path, model_name)
+        if os.path.normpath(model_path) != model_path:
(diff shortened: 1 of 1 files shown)


python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -362,6 +362,12 @@ def _get_vllm_lora_names(
         repo_paths = [repo_paths]
     for repo_path in repo_paths:
         model_path = os.path.join(repo_path, model_name)
+        # Ensure model name doesn't attempt to escape the model repository path.
```
mc-nv: I don't think this comment should be here, your code self explain the action you propose.

whoisj: removed. Looks like I'll need the approval reapplied because of. 😔

PR #8402, opened 2025-09-23
ci: Fix L0_infer_valgrind--base

yinggeh:
Recommand select "Hide whitespace" option for ease of review.

### DIFF
--- a/python/openai/openai_frontend/engine/utils/triton.py
+++ b/python/openai/openai_frontend/engine/utils/triton.py
@@ -362,6 +362,11 @@ def _get_vllm_lora_names(
         repo_paths = [repo_paths]
     for repo_path in repo_paths:
         model_path = os.path.join(repo_path, model_name)
+        if os.path.normpath(model_path) != model_path:
+            raise ValueError(
+                f"Invalid model name: '{model_name}'. Model names must be valid file-system-path segment names."
+            )
+        model_path = os.path.normpath(model_path)
         if not os.path.isdir(model_path):
             # Cloud path?
             return None
--- a/python/openai/requirements.txt
+++ b/python/openai/requirements.txt
@@ -29,7 +29,7 @@ fastapi==0.115.6
 # Fix httpx version to avoid bug in openai library:
 # https://community.openai.com/t/error-with-openai-1-56-0-client-init-got-an-unexpected-keyword-argument-proxies/1040332/3
 httpx==0.27.2
-openai==1.60.0
+openai==1.107.3
 partial-json-parser # used for parsing partial JSON outputs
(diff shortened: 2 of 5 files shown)


qa/L0_infer/infer_test.py
```diff
@@ -163,31 +163,6 @@ def _infer_exact_helper(
                 ):
                     ensemble_prefix.append(prefix)
 
-        if tu.validate_for_onnx_model(
```
yinggeh, line 166: Remove duplicate function call

qa/L0_infer/infer_test.py
```diff
@@ -666,16 +641,59 @@ def test_mix_iff(self):
                 swap=False,
             )
 
-    def test_raw_version_latest_1(self):
-        input_size = 16
-        tensor_shape = (1, input_size)
+    if not VALGRIND_TESTS:
+
+        def test_raw_version_latest_1(self):
```
yinggeh, line 646: These tests were not designed to be part of valgrind specific tests, which is why they are keeping failing.

qa/L0_infer/infer_test.py
```diff
@@ -666,16 +641,59 @@ def test_mix_iff(self):
...
+                        np.int8,
+                        model_version=2,
+                        swap=True,
+                        use_http=USE_HTTP,
+                        use_grpc=USE_GRPC,
+                        use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
+                        use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY,
+                    )
+                except InferenceServerException as ex:
+                    self.assertTrue(
```
whoisj, line 693: is this indentation correct? 😕

qa/L0_infer/infer_test.py
```diff
@@ -747,299 +763,211 @@ def test_raw_version_latest_2(self):
...
+                        ex.message().startswith("Request for unknown model")
+                    )
 
-        if "onnx" in BACKENDS:
-
-            def test_ensemble_mix_type(self):
-                for bs in (1, 8):
+                try:
                     iu.infer_exact(
```
whoisj, line 876: what happens if this doesn't raise?

whoisj (approved):
LGTM, 1 quesion

qa/L0_infer/infer_test.py
yinggeh, line 693: Which line? GitHub pre-commit should already taken care of code formattings.

qa/L0_infer/infer_test.py
yinggeh: I think it will move on where it should fail. We should fix that in a separate PR. Good catch!

PR #8405, opened 2025-09-24
ci: Fix CI failures

pskiran1:
#### What does the PR do?

- Updated the Dockerfile to install the `perf_analyzer` Python package.  
- Modified test scripts to use the pip-installed `perf_analyzer` instead of the `clients/perf_analyzer` binary.

#### Checklist

#### Commit Type:
- [ ] build
- [x] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 35450013

- closes GitHub issue: #xxx


Dockerfile.QA
```diff
@@ -348,7 +348,7 @@ RUN rm -f /usr/bin/python && \
 
 RUN pip3 install --upgrade "numpy<2" pillow attrdict future "grpcio<1.68" requests gsutil \
                            "awscli<=1.36.40" six "grpcio-channelz<1.68" prettytable virtualenv \
-                           check-jsonschema
+                           check-jsonschema perf_analyzer
```
mc-nv: I would probably vote here to move necessary installation towards to the test scenario in `test.sh` file despite a of the duplication. 
Reason for it is that we should think on removing this fat QA image from the CI scenario and focus on final image testing. 

Sure it's only my point of view but would be nice to start moving towards to it, it will help us a lot in future.

mc-nv (approved):

PR #8411, opened 2025-09-25
fix: Replace Use of `alloca`

whoisj:
This change replaces the use of `alloca` which performs stack allocation with `std::unique_ptr<T[]> holder{new T[count]}` because it is safer to perform unknown size allocation in the heap vs stack.

[TRI-69](https://linear.app/nvidia/issue/TRI-69/psirt-buffer-overflow-v2modelsmodel-nameinfer) 
[DLIS-8500](https://jirasw.nvidia.com/browse/DLIS-8500)

CI-Pipeline: [35590031](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/35590031)

### DIFF
--- a/src/data_compressor.h
+++ b/src/data_compressor.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, NVIDIA CORPORATION. All rights reserved.
+// Copyright (c) 2021-2025, NVIDIA CORPORATION. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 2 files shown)


pskiran1 (approved):
LGTM. Will wait for @yinggeh's review as well.
Thank you.

src/data_compressor.h
```diff
@@ -93,11 +93,12 @@ class DataCompressor {
         &stream, deflateEnd);
 
     // Get the addr and size of each chunk of memory in 'source'
+    std::unique_ptr<struct evbuffer_iovec[]> buffer_array_holder;
```
yinggeh, line 96: Maybe my issue but smart pointer makes me overthinking. Can you use `std::vector` instead of smart pointer here?

whoisj: I honestly do not see the difference, especially since `std::vector` uses a smart pointer internally.

whoisj: @yinggeh are you good w/ the change as-is? I'd like to get this merged asap.

PR #8326, opened 2025-08-04
feat: Add support for `usage` in the OpenAI frontend TRT-LLM backend

pskiran1:
#### What does the PR do?

This PR adds support for usage metrics in the OpenAI frontend when using the TRT-LLM backend. Previously, usage reporting (prompt_tokens, completion_tokens, total_tokens) was only available for the vLLM backend.

Key Changes:

- Enabled usage token reporting for TRT-LLM backend by requesting token counts in requests and parsing INT32 data types in responses
- Removed backend-specific checks that limited usage reporting to vLLM only
- Updated tests to expect usage metrics for all backends instead of conditionally based on backend type

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [x] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [x] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [x] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan: 

- CI Pipeline ID: 38287206

#### Background

By default, the OpenAI API supports `usage` for non-streaming requests. For streaming requests, we need to enable `stream_options: {"include_usage": true}`.
https://platform.openai.com/docs/api-reference/chat/create
https://platform.openai.com/docs/api-reference/chat-streaming

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- closes GitHub issue: #xxx

PR #8430, opened 2025-09-30
fix: Update L0_openai to adapt TRT-LLM v1.0

whoisj:
This change moves the import of `LLM` from `tensorrt_llm` to `tensorrt_llm._tensorrt_engine` to adapt the API changes in TRT-LLM v1.0.

[TRI-71](https://linear.app/nvidia/issue/TRI-71/2509-ci-failure-l0-openai-trtllm-base)


mc-nv:
add @pskiran1  to the review as this change may interfere with his work
https://github.com/triton-inference-server/server/pull/8326

PR #8431, opened 2025-10-01
test: Fix L0_simple_lib--base

yinggeh:
#### What does the PR do?
Fix race condition. The `barrier` object is destructed before access in request release callbacks, leading to segfault.

#### Checklist

#### Commit Type:
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/core/pull/452

#### Test plan:

- CI Pipeline ID:
35850970

- closes GitHub issue: #xxx

### DIFF
--- a/src/simple.cc
+++ b/src/simple.cc
@@ -1,4 +1,4 @@
-// Copyright 2020-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2020-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
@@ -876,6 +876,10 @@ main(int argc, char** argv)
     FAIL_IF_ERR(
         TRITONSERVER_InferenceResponseDelete(completed_response),
         "deleting inference response");
+
+    // We need to make sure that the previous request was released before
+    // reusing it.
+    request_release_future.get();
   }
 
   // The TRITONSERVER_InferenceRequest object can be reused for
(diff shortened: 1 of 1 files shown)


src/simple.cc
```diff
@@ -936,6 +936,10 @@ main(int argc, char** argv)
     FAIL_IF_ERR(
         TRITONSERVER_InferenceResponseDelete(completed_response),
         "deleting inference response");
+
+    // We need to make sure that the previous request was released before
+    // reusing it.
+    request_release_future.get();
```
whoisj, line 942: why not use its `.wait()` function instead of `.get()` which can throw?

src/simple.cc
```diff
@@ -993,6 +997,10 @@ main(int argc, char** argv)
     FAIL_IF_ERR(
         TRITONSERVER_InferenceResponseDelete(completed_response),
         "deleting inference response");
+
+    // Make sure the request is released before deleting it. If not, release
+    // callback will segfault after barrier is destructed.
+    request_release_future.get();
```
whoisj, line 1003: why not use its .wait() function instead of .get() which can throw?

src/simple.cc
```diff
@@ -876,6 +876,10 @@ main(int argc, char** argv)
     FAIL_IF_ERR(
         TRITONSERVER_InferenceResponseDelete(completed_response),
         "deleting inference response");
+
+    // We need to make sure that the previous request was released before
+    // reusing it.
+    request_release_future.get();
```
whoisj, line 882: why not use its .wait() function instead of .get() which can throw?

yinggeh: I'm not the owner of this test. My guess is for tests we might want it to throw so that we know what's the error.

whoisj (approved):

pskiran1 (approved):

PR #8432, opened 2025-10-01
fix: Improve input size validation for large HTTP JSON requests

pskiran1:
#### What does the PR do?

This change addresses an issue where an excessively large HTTP JSON payload could cause an unhandled `std::length_error` exception and terminate the server unexpectedly.
A validation check is now performed in the `EVBufferToJson` function. It compares the incoming JSON data size against the configured `--http-max-input-size limit`. Requests that exceed this limit are now gracefully rejected.

- A new test case has been added to test the failing scenario.
- Existing test cases have been updated, ensuring that the entire request JSON is within the `--http-max-input-size` limit.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 36107617

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_http/http_input_size_limit_test.py
+++ b/qa/L0_http/http_input_size_limit_test.py
@@ -40,6 +40,7 @@ import test_util as tu
 # Each FP32 value is 4 bytes, so we need to divide target byte sizes by 4 to get element counts
 BYTES_PER_FP32 = 4
 MB = 2**20  # 1 MB = 1,048,576 bytes
+GB = 2**30  # 1 GB = 1,073,741,824 bytes
 DEFAULT_LIMIT_BYTES = 64 * MB  # 64MB default limit
 INCREASED_LIMIT_BYTES = 128 * MB  # 128MB increased limit
 
@@ -185,7 +186,7 @@ class InferSizeLimitTest(tu.TestResultCollector):
             ]
         }
         # Verify we're actually under the 64MB limit
-        self.assertLess(len(json.dumps(payload)), DEFAULT_LIMIT_BYTES)
+        self.assertLess(len(json.dumps(payload).encode("utf-8")), DEFAULT_LIMIT_BYTES)
 
         response = requests.post(
             self._get_infer_url(model), headers=headers, json=payload
@@ -340,7 +341,7 @@ class InferSizeLimitTest(tu.TestResultCollector):
             ]
         }
         # Verify we're actually under the 128MB limit
-        self.assertLess(len(json.dumps(payload)), INCREASED_LIMIT_BYTES)
+        self.assertLess(len(json.dumps(payload).encode("utf-8")), INCREASED_LIMIT_BYTES)
 
         response = requests.post(
             self._get_infer_url(model), headers=headers, json=payload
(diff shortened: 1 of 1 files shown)


qa/L0_http/http_input_size_limit_test.py
```diff
@@ -180,9 +184,8 @@ def test_default_limit_rejection_json(self):
                 }
             ]
         }
-        assert (
-            shape_size * BYTES_PER_FP32 < 64 * MB
-        )  # Verify we're actually under the 64MB limit
+        # Verify we're actually under the 64MB limit
+        self.assertLess(len(json.dumps(payload)), DEFAULT_LIMIT_BYTES)
```
mattwittwer: Can you add the following encode value to ensure we aren't allowing special characters to be counted as a single byte: .encode('utf-8')

Full line:
self.assertLess(len(json.dumps(payload).encode('utf-8')), DEFAULT_LIMIT_BYTES)

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -360,6 +365,54 @@ def test_large_input_json(self):
             f"Expected shape {[1, shape_size]}, got {result['outputs'][0]['shape']}",
         )
 
+    def test_large_string_in_json(self):
+        """Test JSON request with large string input"""
+        model = "simple_identity"
+
+        # Create a string that is larger (large payload about 2GB) than the default limit of 64MB
+        large_string_size = 2**31 + 64
```
mattwittwer: Please create a GB variable in this file similar to line 42:
MB = 2**20  # 1 MB = 1,048,576 bytes

This will help in readability so that the expression can be:
large_string_size = 2 * GB + 64

It may also help to explain the calculation in the comments:
(2^31 + 64) elements * 1 bytes = 2GB + 64 bytes = 2,147,483,712 bytes

whoisj:
LGTM and Matt already commented on everything that I noticed 👍

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -320,8 +323,11 @@ def test_large_input_json(self):
         )
 
         # Test case 2: Input just under the 128MB configured limit (should succeed)
```
yinggeh, line 326: I saw two size limits, 128MB and 64MB. What's the difference?

src/http_server.cc
```diff
@@ -3060,6 +3060,16 @@ HTTPAPIServer::EVBufferToJson(
     triton::common::TritonJson::Value* document, evbuffer_iovec* v, int* v_idx,
     const size_t length, int n)
 {
+  if (length > max_input_size_) {
```
yinggeh, line 3063: Need to add similar check for `EVBufferToInput`?

qa/L0_http/http_input_size_limit_test.py
pskiran1: The default value is 64MB for `--http-max-input-size`, and 128 MB is a test scenario `--http-max-input-size=<128MB in bytes>`.
cc: @mattwittwer

src/http_server.cc
pskiran1: Based on the current functionality, I think it is not needed, as the `header_length` is only used within the `EVBufferToInput()` function to pass to `EVBufferToJson()`. We have added validation to this function.

PR #8441, opened 2025-10-03
Remove software-properties-common from base container to mitigate GPL…

mc-nv:
Removing package that introduces GBL-2 license. 
Can impact tests scenarios


whoisj (approved):
Seems fine, so long as it's unnecessary.

dmitry-tokarev-nv:
@mc-nv software-properties-common also present in https://github.com/triton-inference-server/server/blob/r25.09/Dockerfile.sdk

dmitry-tokarev-nv (approved):

PR #8478, opened 2025-10-28
rollback: Reverting TensorRT-LLM Backend changes.

mc-nv:


build.py
```diff
@@ -1282,10 +1282,6 @@ def create_dockerfile_linux(
     pip3 install --no-cache-dir  grpcio-tools==1.64.0 && \\
     pip3 uninstall -y setuptools
 ENV LD_LIBRARY_PATH=/usr/local/tensorrt/lib/:/opt/tritonserver/backends/tensorrtllm:$LD_LIBRARY_PATH
-
-RUN apt-get update && apt-get install -y --no-install-recommends \
```
whoisj, line 1286: is this not required to generate the C++ bindings for the gRPC interface?

mc-nv: 1. We source gRPC and install it separately via `third_party`. 
2. This part is related to runtime image. 
3. 25.09 version didn't include this installation...

PR #8489, opened 2025-11-03
Add Integer Overflow Sanity Check

whoisj:
This change adds a sanity check prior to a static cast of `int32_t` to `int64_t` to use as an offset.

While the current code cannot enter this state, this protects against future changes.

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -606,8 +606,14 @@ ReadDataFromJsonHelper(
         const char* cstr;
         size_t len = 0;
         RETURN_IF_ERR(tensor_data.AsString(&cstr, &len));
-        if (static_cast<int64_t>(*counter + len + sizeof(uint32_t)) >
-            expected_cnt) {
+        // Quick sanity check to ensure we don't write beyond `expected_cnt`.
+        int32_t actual_cnt = *counter + len + sizeof(uint32_t);
+        if (actual_cnt < 0) {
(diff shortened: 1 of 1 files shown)


src/http_server.cc
```diff
@@ -606,8 +606,14 @@ ReadDataFromJsonHelper(
         const char* cstr;
         size_t len = 0;
         RETURN_IF_ERR(tensor_data.AsString(&cstr, &len));
-        if (static_cast<int64_t>(*counter + len + sizeof(uint32_t)) >
-            expected_cnt) {
+        // Quick sanity check to ensure we don't write beyond `expected_cnt`.
+        int32_t value = *counter + len + sizeof(uint32_t);
```
yinggeh: Just a thought, this pattern probably resides in many places. Maybe use a wrapper function

whoisj: where would the wrapper function live?

do we have a `random_utils.h` somewhere? honest question, not trying to be difficult - just really do not know.

whoisj: actually, looking at the file: we only do this kind of thing in one place.

yinggeh:
```suggestion
        int32_t actual_cnt = *counter + len + sizeof(uint32_t);
```

yinggeh (approved):

yinggeh (approved):

PR #8497, opened 2025-11-10
fix: Clone TRT Repository when Necessary

whoisj:
This change tests for the full path before changing directory and clones the TensorRT directory when the full path is unavailable.


qa/common/gen_qa_model_repository
yinggeh:
```suggestion
if [ -d "/usr/src/tensorrt" ] && [ -d "/usr/src/tensorrt/samples/python" ]; then
    cd /usr/src/tensorrt/samples/python/onnx_custom_plugin
else
    git clone -b release/${TENSORRT_VERSION} https://github.com/NVIDIA/TensorRT.git
    cd /workspace/TensorRT/samples/python/onnx_custom_plugin
fi
```

whoisj: updated.

PR #8544, opened 2025-11-25
models: restructure models generation script.

mc-nv:
We need to made changes to the models generation in order to be able mitigate build behavior in absence of docker.


dmitry-tokarev-nv (approved):
Approving to unblock progress.

PR #8568, opened 2025-12-10
Update models generation script (#8567)

mc-nv:


qa/common/gen_qa_model_repository
```diff
@@ -344,7 +344,7 @@ chmod -R 777 $TRITON_MDLS_QA_TRT_DATA_DEPENDENT_MODEL
 if [ -d "/usr/src/tensorrt/samples/python/onnx_custom_plugin" ]; then
     cd /usr/src/tensorrt/samples/python/onnx_custom_plugin
 else
-    git clone -b release/\${TRT_VERSION%.*.*} --depth 1 https://github.com/NVIDIA/TensorRT.git /workspace/TensorRT
+    git clone -b release/\$( echo \$TRT_VERSION | cut -d . -f -2) --depth 1 https://github.com/NVIDIA/TensorRT.git /workspace/TensorRT
```
dmitry-tokarev-nv, line 347: why not just ${TRT_VERSION%.*} ?
```suggestion
    git clone -b release/\${TRT_VERSION%.*}--depth 1 https://github.com/NVIDIA/TensorRT.git /workspace/TensorRT
```

dmitry-tokarev-nv (approved):
Works but could be simpler imo

qa/common/gen_qa_model_repository
mc-nv: There is two reason. 
1. The branch name.
2. The variable value

```bash
# expected value: "release/x.y"
# 25.11 TRT_VERSION
$ export TRT_VERSION=10.14.1.48
$ echo ${TRT_VERSION%.*}
10.14.1
# 25.12 TRT_VERSION
$ export TRT_VERSION=10.14.1.48+cuda13.0
$ echo ${TRT_VERSION%.*}
10.14.1
# using "cut" utility
$ echo $(echo $TRT_VERSION | cut -d . -f -2)
10.14

```

dmitry-tokarev-nv: ahh so it's not x.y.z. Thank you for explanation!

PR #8569, opened 2025-12-10
feat: Support multi-LoRA in TensorRT-LLM backend in OpenAI-compatible API frontend

yinggeh:
#### What does the PR do?
Support multi-LoRA in TensorRT-LLM backend in OpenAI-compatible API frontend

#### Checklist

#### Commit Type:
- [x] feat

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx

### DIFF
--- a/python/openai/README.md
+++ b/python/openai/README.md
@@ -310,6 +310,8 @@ The path of LoRA adapter in `multi_lora.json` is the directory of
 
 <details>
 <summary>For example</summary>
+
+model repository
 ```
 inflight_batcher_llm
 ├── postprocessing
@@ -336,8 +338,9 @@ inflight_batcher_llm
     |   └── multi_lora.json
     └── config.pbtxt
 ```
-```
+
 multi_lora.json
+```
 {
   "doll": "inflight_batcher_llm/tensorrt_llm_bls/1/luotuo-lora-7b-0.1-weights",
   "sheep": "inflight_batcher_llm/tensorrt_llm_bls/1/Japanese-Alpaca-LoRA-7b-v0-weights"


python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -135,8 +146,8 @@ def _create_vllm_generate_request(
             request_logprobs = True
     inputs["return_logprobs"] = np.bool_([request_logprobs])
 
-    if lora_name is not None:
-        sampling_parameters["lora_name"] = lora_name
+    if lora_config is not None:
```
whoisj, line 149: I think this needs a `lora_config.name is not None` as well.

python/openai/openai_frontend/engine/utils/triton.py
```diff
@@ -167,15 +178,10 @@ def _create_trtllm_generate_request(
     model,
     prompt,
     request: CreateChatCompletionRequest | CreateCompletionRequest,
-    lora_name: str | None,
+    lora_config: TritonLoraConfig | None,
     echo_tensor_name: str | None,
     default_max_tokens: int,
 ):
-    if lora_name is not None:
```
whoisj, line 174: do we need if `lora_config.name is not None` instead of just removing the check?

yinggeh: No. `name` must not be None in `TritonLoraConfig` when it's get parsed from `multi_lora.json`.

python/openai/openai_frontend/engine/utils/triton.py
yinggeh: Answered in the other thread.

whoisj (dismissed):

PR #8570, opened 2025-12-10
fix: Set memory decompression limit

mattwittwer:
#### What does the PR do?

This PR adds additional size validation to compressed inputs.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

The main changes are in: src/data_compressor.h please review the logic to calculate the size of the decompressed buffer and then review the testing changes.

#### Test plan:

Yes specific tests for this decompression function were added here: src/test/data_compressor_test.cc
And additional end to end tests were added here: qa/L0_http/http_input_size_limit_test.py

- CI Pipeline ID:

40402527

- closes GitHub issue: #xxx

### DIFF
--- a/src/data_compressor.h
+++ b/src/data_compressor.h
@@ -311,6 +311,11 @@ class DataCompressor {
                     "encountered inconsistent stream state during "
                     "decompression");
               }
+              // Break if decompression is complete, even if buffer is exactly
+              // full
+              if (ret == Z_STREAM_END) {
+                break;
+              }
             } while (stream.avail_out == 0);
           }
           // Make sure the last buffer is committed


src/data_compressor.h
```diff
@@ -264,11 +301,25 @@ class DataCompressor {
...
+            }
+
             RETURN_MSG_IF_ERR(
                 CommitEVBuffer(
                     decompressed_data, &current_reserved_space,
-                    output_buffer_size - stream.avail_out),
-                "unexpected error committing output buffer for compression: ");
+                    final_chunk_size),
+                "unexpected error committing output buffer for "
+                "decompression: ");
```
whoisj: this sentence ends in ": ", is that correct?

src/data_compressor.h
```diff
@@ -227,7 +233,12 @@ class DataCompressor {
               reinterpret_cast<unsigned char*>(current_reserved_space.iov_base);
           stream.avail_out = output_buffer_size;
 
-          // Compress until end of 'source'
+          // Track total decompressed size and current buffer size for limit
+          // checking
+          size_t total_decompressed = 0;
+          size_t current_buffer_size = output_buffer_size;
```
whoisj, line 241: why "current" and not "remaining"?

src/data_compressor.h
```diff
@@ -175,6 +176,11 @@ class DataCompressor {
                                     ? source_byte_size
                                     : (1 << 20 /* 1MB */);
 
+    if (max_decompressed_size > 0 &&
```
yinggeh, line 181: Please comment with the purpose/rationale behind.

qa/L0_http/test.sh
```diff
@@ -775,14 +775,14 @@ fi
 
 set +e
 # Run test to verify that large inputs fail with default limit
-python http_input_size_limit_test.py InferSizeLimitTest.test_default_limit_rejection_raw_binary >> $CLIENT_LOG 2>&1
+python http_input_size_limit_test.py InferSizeLimitTest.test_default_limit_raw_binary >> $CLIENT_LOG 2>&1
```
yinggeh, line 778: Why renaming tests here?

yinggeh:
Please also update [command_line_parser.cc](https://github.com/triton-inference-server/server/blob/90b6675887beb363d0b6973a8949fc2bd43afb1d/src/command_line_parser.cc#L503).

src/data_compressor.h
mattwittwer: Good catch, yes all print statements on this page include an unnecessary ':'. The RETURN_MSG_IF_ERR from src/common.h:81 is used to format this message and appends a colon already.

I manually triggered this failure to see error message. This shows the double colon failure:
`unexpected error allocating output buffer for decompression: : failed to reserve 1048576 bytes in evbuffer`

Updated all instances of the double colon in data_compressor.h

qa/L0_http/test.sh
mattwittwer: I updated the names here to better standardize the test naming convention in this file. The _rejection_ portion of the name is not necessary and makes the first two tests appear different than the rest. While in reality they are very similar to the large input tests and the compressed tests. If it would be cleaner I can move this rename to a separate PR. 

# Previous test names:
def test_default_limit_rejection_raw_binary(self):
def test_default_limit_rejection_raw_binary(self):
def test_large_input_raw_binary(self):
def test_large_input_json(self):
def test_large_string_in_json(self):

# Updated names for the existing _default_ limit tests:
def test_default_limit_raw_binary(self):
def test_default_limit_json(self):
def test_large_input_raw_binary(self):
def test_large_input_json(self):
def test_large_string_in_json(self):

# New Tests for this PR:
def test_default_limit_compressed(self):
def test_large_input_compressed(self):

mattwittwer:
> Please also update [command_line_parser.cc](https://github.com/triton-inference-server/server/blob/90b6675887beb363d0b6973a8949fc2bd43afb1d/src/command_line_parser.cc#L503).

Updated the description of http-max-input-size in command_line_parser.cc.

src/data_compressor.h
mattwittwer: Updated to include a description for this check

src/data_compressor.h
mattwittwer: This is just tracking the current amount of buffer available for decompression at each step in the inflate() loop. So generally this would be 1MB, or more, based on the value of output_buffer_size set [here](https://github.com/triton-inference-server/server/blob/mwittwer/set_memory_decompression_limit/src/data_compressor.h#L175)
And limited by the remaining size before the limit is hit [here](https://github.com/triton-inference-server/server/blob/mwittwer/set_memory_decompression_limit/src/data_compressor.h#L281)

src/data_compressor.h
whoisj: LGTM, thanks @mattwittwer

whoisj (approved):
LGTM :shipit:

mattwittwer:
Encountered error with base case where decompression size == decompression limit failed to allocate:
```
[ RUN      ] DataCompressorTest.DecompressionSizeLimitGzip
/opt/tritonserver/server/src/test/data_compressor_test.cc:698: Failure
Value of: (err == nullptr)
  Actual: false
Expected: true
64 MB data should decompress at exact 64 MB limit: Decompressed data size exceeds the maximum allowed value of 67108864 bytes. Use --http-max-input-size to increase the limit.
[  FAILED  ] DataCompressorTest.DecompressionSizeLimitGzip (1476 ms)
[ RUN      ] DataCompressorTest.DecompressionSizeLimitDeflate
/opt/tritonserver/server/src/test/data_compressor_test.cc:776: Failure
Value of: (err == nullptr)
  Actual: false
Expected: true
64 MB data should decompress at exact 64 MB limit: Decompressed data size exceeds the maximum allowed value of 67108864 bytes. Use --http-max-input-size to increase the limit.
[  FAILED  ] DataCompressorTest.DecompressionSizeLimitDeflate (1510 ms)
[----------] 12 tests from DataCompressorTest (5481 ms total)

[----------] Global test environment tear-down
[==========] 12 tests from 1 test suite ran. (5481 ms total)
[  PASSED  ] 10 tests.
[  FAILED  ] 2 tests, listed below:
```

Resolved with an additional check:
If there is nothing more to decompress. Break the inflate() loop. No need to generate a new buffer. This is especially important when there is no more room to allocate additional buffer.
https://github.com/triton-inference-server/server/blob/mwittwer/set_memory_decompression_limit/src/data_compressor.h#L314-L318

PR #8590, opened 2026-01-06
Update condition for manilinux scenarios.

mc-nv:
- **Update conditions**
- **Update condition**


yinggeh:
Please also review logic in https://github.com/triton-inference-server/server/blob/75588dc03703d7a911995d4172f7bc7facda1187/qa/L0_infer/install_and_test.sh

mc-nv:
> Please also review logic in https://github.com/triton-inference-server/server/blob/75588dc03703d7a911995d4172f7bc7facda1187/qa/L0_infer/install_and_test.sh

Thanks @yinggeh , provided logic pretty similar to what we introduced in this change, in future we may align it.

PR #8612, opened 2026-01-20
fix: Improve bounds checking on integer conversions

whoisj:
This change improves the handling of values which could cause int32 values of over/underflow leading to undefined behavior.

[TRI-557](https://linear.app/nvidia/issue/TRI-557)

### DIFF
--- a/src/http_server.cc
+++ b/src/http_server.cc
@@ -609,7 +609,8 @@ ReadDataFromJsonHelper(
         if (len > INT64_MAX) {
           return TRITONSERVER_ErrorNew(
               TRITONSERVER_ERROR_INTERNAL,
-              "Unable to parse 'data' field: string length exceeds size limitation");
+              "Unable to parse 'data' field: string length exceeds size"
+              " limitation");
         }
         // Quick sanity check to ensure we don't write beyond `expected_cnt`.
         int64_t tmp_cnt = static_cast<int64_t>(*counter) +


src/http_server.cc
```diff
@@ -603,21 +603,29 @@ ReadDataFromJsonHelper(
...
+        int64_t tmp_cnt = static_cast<int64_t>(*counter) +
+                          static_cast<int64_t>(len) +
+                          static_cast<int64_t>(sizeof(uint32_t));
+        if (tmp_cnt < 0) {
           return TRITONSERVER_ErrorNew(
               TRITONSERVER_ERROR_INTERNAL,
               "Unable to parse 'data' field: string length is negative");
         }
-        if (static_cast<int64_t>(actual_cnt) > expected_cnt) {
+        if (tmp_cnt > expected_cnt) {
```
yinggeh: You can combine this if statement with `if (tmp_cnt < 0) {`

src/http_server.cc
```diff
@@ -603,21 +603,29 @@ ReadDataFromJsonHelper(
...
           return TRITONSERVER_ErrorNew(
               TRITONSERVER_ERROR_INTERNAL,
               "Shape does not match true shape of 'data' field");
         }
+        int32_t actual_cnt = static_cast<int32_t>(tmp_cnt);
```
yinggeh, line 629: If `tmp_cnt <= expected_cnt` and `expected_cnt` is `int64_t`, how can you make sure it's a lossless cast?

src/http_server.cc
whoisj: different error messages, but I guess combining them won't hurt.

src/http_server.cc
whoisj: fixed up

src/http_server.cc
yinggeh: `tmp_cnt` < 0 simply means integer overflow IMO

PR #8656, opened 2026-02-12
Increase batcher test timeout

mc-nv:
Updating timeout to support longer time for models workload.


mudit-eng:
Does new timeout guarantee that tests would pass? How do we know there is not an underlying regression causing longer response time?

mc-nv:
> Does new timeout guarantee that tests would pass? How do we know there is not an underlying regression causing longer response time?

It's experimental result, currently we have a QA complain on failed test. 
We can try to warm up the models before starting the server, it could speed up. 
Indeed this change is questionable, but without submitting it we won't have a conversation about it.

PR #8659, opened 2026-02-13
fix: Path Traversal Bug in SageMaker Server

whoisj:
This change prevents the /dev/, /proc/, or /sys/ folders being used to load models from in `SagemakerAPIServer` to avoid issues with interacting with pseudo-filesystem paths.

### DIFF
--- a/src/sagemaker_server.cc
+++ b/src/sagemaker_server.cc
@@ -300,8 +300,8 @@ SagemakerAPIServer::ParseSageMakerRequest(
                                      // redundant components.
   auto url_abspath = url_path.string();
 
-  if (repo_path.find("/dev/") == 0 || repo_path.find("/proc/") == 0 ||
-      repo_path.find("/sys/") == 0) {
+  if (url_abspath.find("/dev/") == 0 || url_abspath.find("/proc/") == 0 ||
+      url_abspath.find("/sys/") == 0) {
     LOG_ERROR << "Invalid URL: " << url_string
               << ". \"url\" property value cannot start with /dev/, /proc/, or "
                  "/sys/."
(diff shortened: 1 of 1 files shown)


yinggeh:
Please keep linear ticket private. It's not supposed to be exposed to the public.

src/sagemaker_server.cc
```diff
@@ -291,6 +292,15 @@ SagemakerAPIServer::ParseSageMakerRequest(
     }
   }
 
+  if (url_string.find("/dev/") == 0 || url_string.find("/proc/") == 0 ||
```
yinggeh: Is `url_string` a normalized absolute path? One can do `/home/xxx/model_repository/../../dev`, etc.

yinggeh: > Not sure if it's a case, isn't we should check that the path is canonical?
> If we talking about the system isn't can be that  /etc , /usr , /boot and other system folders are also could be targeted?

Same question asked by Misha.

whoisj: this is waiting on @GuanLuo for clarification.

whoisj: good point. the value can literally be anything 😒 

I suppose we could pull the core APIs that provide better path handling into this code and use those to resolve things.

whoisj: updated the change with your recommendation @yinggeh

src/sagemaker_server.cc
```diff
@@ -291,6 +294,22 @@ SagemakerAPIServer::ParseSageMakerRequest(
     }
   }
 
+  std::filesystem::path url_path(url_string);
+  url_path = std::filesystem::absolute(
+      url_path.lexically_normal());  // Normalize the path to remove any
+                                     // redundant components.
+  auto repo_path = url_path.string();
```
yinggeh:
```suggestion
  auto url_abspath = url_path.string();
```

src/sagemaker_server.cc
```diff
@@ -891,10 +910,26 @@ SagemakerAPIServer::SageMakerMMELoadModel(
...
-  std::string repo_path = parse_map.at("url");
+  std::string url_string = parse_map.at("url");
   std::string model_name_hash = parse_map.at("model_name_hash");
   std::string target_model = parse_map.at("target_model");
 
+  std::filesystem::path url_path(url_string);
+  url_path = std::filesystem::absolute(
+      url_path.lexically_normal());  // Normalize the path to remove any
+                                     // redundant components.
+  std::string repo_path = url_path.string();
```
yinggeh:
```suggestion
  std::string url_abspath = url_path.string();
```

src/sagemaker_server.cc
yinggeh: Any update?

whoisj: None. 😔

Let's not worry about additional paths, the identified issue was related to accessing virtual paths which this PR resolves.

PR #8665, opened 2026-02-18
fix: Block path traversal attacks via the deployment API for MLflow-Triton

mudit-eng:
#### What does the PR do?
This code change sanitizes the model name in the MLflow-Triton deployment plugin to prevent path traversal.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

Added a unit test and verified locally. Also, tested in CI.

- CI Pipeline ID:
44324776, 45253345

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_mlflow/test.sh
+++ b/qa/L0_mlflow/test.sh
@@ -256,7 +256,7 @@ if [ $? -ne 0 ]; then
     echo -e "\n***\n*** Python Test Failed\n***"
     RET=1
 else
-    check_test_results $TEST_RESULT_FILE 2
+    check_test_results $TEST_RESULT_FILE 3
     if [ $? -ne 0 ]; then
         cat $PY_LOG
         echo -e "\n***\n*** Test Result Verification Failed\n***"


yinggeh:
Please use template https://github.com/triton-inference-server/server/blob/main/.github/PULL_REQUEST_TEMPLATE/pull_request_template_internal_contrib.md

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
```diff
@@ -513,6 +516,12 @@ def _validate_flavor(self, flavor):
         if flavor not in self.supported_flavors:
             raise Exception("{} model flavor not supported by Triton".format(flavor))
 
+    def _validate_model_name(self, name):
+        if not name:
+            raise Exception("Please provide a model name for the deployment")
+        if '/' in name or '\\' in name or '..' in name:  # Path traversal protection
+            raise Exception("Path traversal is not allowed in model's name: {}".format(name))
```
yinggeh: A more commonly used is `f"Path traversal is not allowed in model's name: {name}"`

qa/L0_mlflow/plugin_test.py
```diff
@@ -116,6 +115,62 @@ def test_onnx_flavor_with_files(self):
             filecmp.cmp(config_path, "./models/onnx_model_with_files/config.pbtxt")
         )
 
+    def test_path_traversal_model_name(self):
```
yinggeh:
```suggestion
    def test_invalid_path_traversal_model_name(self):
```

qa/L0_mlflow/plugin_test.py
```diff
@@ -116,6 +115,62 @@ def test_onnx_flavor_with_files(self):
...
+
+        model_uri = "models:/onnx_model_with_files/1"
+
+        model_name_normal = "onnx_model_123"
+        self.client_.create_deployment(model_name_normal, model_uri, flavor="onnx")
```
yinggeh: Why you creating then deleting a deployment?

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
```diff
@@ -513,6 +516,12 @@ def _validate_flavor(self, flavor):
         if flavor not in self.supported_flavors:
             raise Exception("{} model flavor not supported by Triton".format(flavor))
 
+    def _validate_model_name(self, name):
+        if not name:
+            raise Exception("Please provide a model name for the deployment")
+        if '/' in name or '\\' in name or '..' in name:  # Path traversal protection
```
yinggeh: Why isn't `\\` allowed?

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
mudit-eng: I will change it. I just followed the convention in this file where .format is used.

qa/L0_mlflow/plugin_test.py
mudit-eng: Earlier, my unit test was failing. I wanted to rule out a potential resource leak, so I added a delete step after a successful create. I don't think it is needed.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
mudit-eng: This is for path traversal in Windows. Since we don't support Windows, I will remove it.

yinggeh: What's double backsplash `\\` on windows?

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
yinggeh: I see. In this file it's fine. I've seen the other way more commonly used in our code.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
mudit-eng: This translates to backslash traversal. Something like this: C:\Windows\System32

yinggeh: Got it. The first backslash is the escape character.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
whoisj: for the record the `\` character is a legal character on POSIX systems.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
```diff
@@ -513,6 +516,14 @@ def _validate_flavor(self, flavor):
         if flavor not in self.supported_flavors:
             raise Exception("{} model flavor not supported by Triton".format(flavor))
 
+    def _validate_model_name(self, name):
```
yinggeh, line 519: Just a note no need to fix today, there are multiple similar pattern @whoisj added to address PSIRT issues in server. We could refactor them into one common utility fuction.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
```diff
@@ -99,6 +99,9 @@ def create_deployment(self, name, model_uri, flavor=None, config=None):
         """
         self._validate_flavor(flavor)
 
+        # Validate model name
+        self._validate_model_name(name)
+
         # Verify model does not already exist in Triton
```
yinggeh, line 105: This check should also be moved into`_validate_model_name`.

qa/L0_mlflow/plugin_test.py
```diff
@@ -116,6 +116,41 @@ def test_onnx_flavor_with_files(self):
             filecmp.cmp(config_path, "./models/onnx_model_with_files/config.pbtxt")
         )
 
+    def test_invalid_path_traversal_model_name(self):
```
yinggeh: Not sure if it works but can you try pytest decorator https://github.com/triton-inference-server/server/blob/0844c371fe62c9fb8da9e47e7d76808bdd0ccc58/python/openai/tests/test_lora.py#L43-L56 to make the test cleaner?

yinggeh:
```suggestion
    def test_invalid_model_name(self):
```

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
mudit-eng: Which check should move into `_validate_model_name`?

qa/L0_mlflow/plugin_test.py
mudit-eng: I refactored the code to use the pytest decorator. While the test change itself is straightforward, it also requires environment updates. We would need to modify the test.sh file to make it work with pytest instead of unittest. This increases the scope of the change, so I recommend keeping it as a separate update.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
yinggeh: I was referring to L105-110. On the second thought, either should be fine. Your call it.
```
        # Verify model does not already exist in Triton
        if self._model_exists(name):
            raise Exception(
                "Unable to create deployment for name %s because it already exists."
```

qa/L0_mlflow/plugin_test.py
mudit-eng: I have made this change

yinggeh: Fair enough. Can you add a "TODO" comment linking to a linear ticket?

yinggeh: Actually, can you refactor into a simple for-loop instead of using pytest?

qa/L0_mlflow/plugin_test.py
```diff
@@ -116,6 +116,41 @@ def test_onnx_flavor_with_files(self):
...
+        model_uri = "models:/onnx_model_with_files/1"
+
+        model_name_empty = ""
```
yinggeh: Please also test `model_name_empty = "  "`

qa/L0_mlflow/plugin_test.py
mudit-eng: I don't think we need a Linear ticket for pytest. This file is rarely touched and there is no real impact of converting this to pytest. We have other important areas in Triton to work on. If I create a ticket also, TBH, it will just remain in the backlog.
For the for-loop suggestion, what simplification are you looking for - readability or maintainability or something else? We just have 3 cases to tests and code is straightforward. And as we know, this file is rarely touched. I don't think it is worth spending time going back and forth with pytest decorators or for-loop, unless there is a real benefit.

qa/L0_mlflow/plugin_test.py
mudit-eng: What would be the reason for testing 'spaces'? Spaces are allowed in the model name. If we go in this direction, then we can test lot of cases like ".", "#", "      ", and so on. Scope of this PR is to do security hardening by preventing path traversal.

yinggeh: Because we've seen server error with empty model names. The point of tests is not testing that what we know will work. The point of tests is to discover as many corner cases before thery are reported by our customers and any orgs, like what you have spent time fixing now. It also prevents anyone in the future who accidentally touchs this code and causes regression.

> then we can test lot of cases like ".", "#", " ", and so on

And yes, I don't mind having more test cases if they are necessary. See how @pskiran1 did https://github.com/triton-inference-server/server/pull/8672/changes#diff-46efca2b90e4866165c6eda38961abdc21ab0bbcbe81f1dfa916489139fc08fe

qa/L0_mlflow/plugin_test.py
yinggeh: I don’t see functional risk in keeping this as-is. My only suggestion is around long-term maintainability — parameterized tests tend to scale better when validating path validation logic like this. This could reduce duplication and improve scalability if more cases are added later. If we expect this area to evolve, it might be worth adopting that pattern now while the test surface is small.

qa/L0_mlflow/plugin_test.py
mudit-eng: I totally understand what you are saying @yinggeh .

I can add non-path traversal cases, but this PR is primarily focused on blocking path traversal as reported in PSIRT. We can expand the scope of the ticket to include model name validation if needed. I just want to be mindful of potential scope creep. 

I have seen Sai’s PR. @pskiran1 is still testing path traversal scenarios. Although he has added test cases covering spaces, non-printable characters, etc., he is still validating for path traversal patterns (.., /) within those cases.
I don't think he is explicitly testing `model_name_empty = "  "`.

What are other invalid model names I can test with? Are there other characters that are not allowed? Let's have a list so that both PRs (Sai's and mine) can include those.

@whoisj

qa/L0_mlflow/plugin_test.py
mudit-eng: Fair point. Adding for-loop should be trivial. I can do that. But I don't think we should go to pytest direction for the same reasons I mentioned earlier.

qa/L0_mlflow/plugin_test.py
whoisj: I'm 100% fine with keeping the scope of this PR being kept minimal, so long as we create a tracking ticket internally for the work being put off for a future PR.

mudit-eng: I have made the change to check that model name is not empty and does not only contains whitespace, tabs, or newlines.
For any other validation, we can have a tracking ticket.

deploy/mlflow-triton-plugin/mlflow_triton/deployments.py
```diff
@@ -513,6 +516,16 @@ def _validate_flavor(self, flavor):
         if flavor not in self.supported_flavors:
             raise Exception("{} model flavor not supported by Triton".format(flavor))
 
+    def _validate_model_name(self, name):
+        # Check if the model name is empty or only contains whitespace, tabs, or newlines
+        if name.strip() == "":
+            raise Exception("Please provide a valid model name for the deployment")
```
yinggeh:
```suggestion
            raise Exception("Empty model name is not allowed. Please provide a valid model name for the deployment")
```

qa/L0_mlflow/plugin_test.py
```diff
@@ -116,6 +116,66 @@ def test_onnx_flavor_with_files(self):
...
+                str(e.exception),
+            )
+            self.assertNotIn(
+                f"Path traversal is not allowed in model's name: {model_name}",
+                str(e.exception),
+            )
+            self.assertIn(
+                f"Registered Model with name={model_name} not found",
+                str(e.exception),
+            )
```
yinggeh:
```suggestion
            self.client_.create_deployment(model_name, model_uri, flavor="onnx")
```

yinggeh (changes requested):

qa/L0_mlflow/plugin_test.py
yinggeh: The test fails if exception is raised.

mudit-eng: Rewrote the code to change asserts.

yinggeh (approved):
Nice work

PR #8667, opened 2026-02-19
Update variables for models generation. Addressing legacy variables values.

mc-nv:
Addressing legacy variables values.

### DIFF
--- a/qa/common/gen_jetson_trt_models
+++ b/qa/common/gen_jetson_trt_models
@@ -42,7 +42,7 @@ TENSORRT_IMAGE=${TENSORRT_IMAGE:=nvcr.io/nvidia/tensorrt:$TRITON_VERSION-py3-igp
 UBUNTU_IMAGE=${UBUNTU_IMAGE:=ubuntu:24.04}
 
 # Set CI specific parameters
-DOCKER_GPU_ARGS=${DOCKER_GPU_ARGS:-$([[ $RUNNER_GPUS =~ ^[0-9] ]] && eval $NV_DOCKER_ARGS || echo "--gpus $NVIDIA_VISIBLE_DEVICES" )}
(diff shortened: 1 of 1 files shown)


qa/common/gen_qa_model_repository
```diff
@@ -80,7 +80,7 @@ PROJECT_NAME=${PROJECT_NAME:=tritonserver}
 log_message.status "define: Docker engine parameters"
 DOCKER_VOLUME=${DOCKER_VOLUME:=volume.gen_qa_model_repository.${CI_JOB_ID}}
 DOCKER_VOLUME_CONTAINER=${DOCKER_VOLUME}.gen_qa_model_repository.${CI_JOB_ID}
-DOCKER_GPU_ARGS=${DOCKER_GPU_ARGS:-$([[ $RUNNER_GPUS =~ ^[0-9] ]] && eval $NV_DOCKER_ARGS || echo "--gpus device=$CUDA_DEVICE" )}
+DOCKER_GPU_ARGS=${DOCKER_GPU_ARGS:-$([[ -v RUNNER_GPUS && $RUNNER_GPUS =~ ^[0-9] ]] && eval $NV_DOCKER_ARGS || echo "--runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=$NVIDIA_VISIBLE_DEVICES" )}
```
whoisj, line 83: using `NVIDIA_VISIBLE_DEVICES` is the direct solution, but it's nvidia-runtime only (which is probably fine). the `--gpus device=1` is the official way to do this with docker regardless of custom runtime.

have you proven that this resolves correctly? the existing problem wasn't the mechanism, it was the resolution of this statement.

mc-nv: I've tested it, locally and on multi node cluster, in given scenarios it went trough.

PR #8676, opened 2026-02-24
fix: Avoid overflows when reading json inputs

whoisj:
Check input from HTTP+JSON sizes before allocating memory for them.

Protect against integer overflows while we're at it.

### DIFF
--- a/qa/L0_http/generate_endpoint_test.py
+++ b/qa/L0_http/generate_endpoint_test.py
@@ -276,7 +276,7 @@ class GenerateEndpointTest(tu.TestResultCollector):
             self.generate_stream_expect_failure(self._model_name, inputs, error_msg)
 
     def test_json_dtype_size_expansion_exceeds_limit_error(self):
-        '''
+        """
         Test that when the client sends a JSON input of byte[], that when it
         expands to dtype[], it exceeds the maximum allowed input size and
         returns an appropriate error message. The test sends a large base64
@@ -287,7 +287,7 @@ class GenerateEndpointTest(tu.TestResultCollector):
         exceeds the limit.
         This is important to prevent clients from sending inputs that could
         cause excessive memory usage on the server.
-        '''
+        """
 
         input_data = [1] * (
             64 * 1024 * 1024
--- a/qa/L0_http/http_input_size_limit_test.py
+++ b/qa/L0_http/http_input_size_limit_test.py
@@ -65,7 +65,7 @@ class InferSizeLimitTest(tu.TestResultCollector):
         return f"http://localhost:8000/v2/models/{model_name}/infer"
 
     def test_json_dtype_size_expansion_exceeds_limit_error(self):
-        '''
+        """
         Test that when the client sends a JSON input of byte[], that when it
         expands to dtype[], it exceeds the maximum allowed input size and
         returns an appropriate error message. The test sends a large base64
@@ -76,7 +76,7 @@ class InferSizeLimitTest(tu.TestResultCollector):
         exceeds the limit.
         This is important to prevent clients from sending inputs that could
         cause excessive memory usage on the server.
-        '''
+        """
         model = "onnx_zero_1_float32"
 
         # Provided data is 64MB of int8, but the model expects FP32,
@@ -115,6 +115,9 @@ class InferSizeLimitTest(tu.TestResultCollector):
             f"Expected error code for type/size mismatch, got: {response.status_code}",
         )
         error_msg = response.content.decode()
+        print(
+            f"Error message: {error_msg}", flush=True
+        )  # Print the error message for debugging
         self.assertIn(
             "Request JSON size of ",
             error_msg,
@@ -158,6 +161,9 @@ class InferSizeLimitTest(tu.TestResultCollector):
             f"Expected error code for type/size mismatch, got: {response.status_code}",
         )
         error_msg = response.content.decode()
+        print(
+            f"Error message: {error_msg}", flush=True
+        )  # Print the error message for debugging
         self.assertIn(
             "request JSON size of ",
             error_msg,


mudit-eng:
LGTM

mattwittwer:
One edge case: 
max_input_size_ is checked per-input rather than cumulatively, allowing up to a 4x memory limit bypass.
This PR checks `byte_size > max_input_size_` to prevents excessive memory allocation. However, because this check is evaluated on individual inputs rather than the cumulative total of the entire HTTP request, a single request can still cause the server to allocate up to 4x the memory limit.

The scenario:
The upstream EVBufferToJson limits the physical size of the incoming HTTP JSON payload to max_input_size_.
However, the most compact JSON representation of an element is 0, (2 bytes), while an FP64 element allocates 8 bytes in memory. 
With a limit set to --http-max-input-size=128000000 (128 MB), a single 128 MB JSON payload formatted like this can allocate 512MB:

```
{
  "parameters": {
    "in_1": {"datatype": "FP64", "data": [ ... 16 million zeros (32 MB JSON) ... ]},
    "in_2": {"datatype": "FP64", "data": [ ... 16 million zeros (32 MB JSON) ... ]},
    "in_3": {"datatype": "FP64", "data": [ ... 16 million zeros (32 MB JSON) ... ]},
    "in_4": {"datatype": "FP64", "data": [ ... 16 million zeros (32 MB JSON) ... ]}
  }
}
```

What happens:
The upstream check passes because the total HTTP payload is exactly 128 MB.
ExactMappingInput parses in_1. The calculated byte_size is 128 MB. The check if (byte_size > max_input_size_) passes. Triton allocates 128 MB. This repeats for in_2, in_3, and in_4. Resulting in the server allocating 512 MB of memory for a single request, bypassing the 128 MB max_input_size_ limit.

whoisj:
> max_input_size_ is checked per-input rather than cumulatively, allowing up to a 4x memory limit bypass.

Agreed about exceeding the overall limit. I'll create a follow up to address that separately. Thanks for pointing it out.

src/http_server.cc
```diff
@@ -3593,7 +3593,48 @@ HTTPAPIServer::GenerateRequestClass::ExactMappingInput(
...
+    // allowed input size.
+    if (byte_size > max_input_size_) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INVALID_ARG,
+          ("Input '" + name + "' has a byte_size (" +
+           std::to_string(byte_size) +
+           " bytes) that exceeds the maximum allowed value of " +
+           std::to_string(max_input_size_) +
+           " bytes. Use --http-max-input-size to increase the limit.")
+              .c_str());
```
yinggeh, line 3658:
```suggestion
```

yinggeh: Based on @mattwittwer feedback, we should remove the size check and come up with a proper solution.

whoisj: why remove this check?

yinggeh: Correct me if I am wrong. Do we need this check here? What's the point of checking individual input if we check for sum of all input sizes after all?

whoisj: If we remove this check then this PR does nothing. Are you suggesting the PR should be scrapped?

whoisj: @yinggeh updated. Let me know if this more along what you were thinking or not?

src/http_server.cc
```diff
@@ -3597,46 +3600,32 @@ HTTPAPIServer::GenerateRequestClass::ExactMappingInput(
...
-                    .c_str());
-          }
-          break;
-        default:
-          return TRITONSERVER_ErrorNew(
-              TRITONSERVER_ERROR_INVALID_ARG,
-              (std::string("input '") + name + "' has unsupported datatype " +
-               value)
-                  .c_str());
+      if (element_cnt > (SIZE_MAX / element_size)) {
```
yinggeh, line 3627: Not sure, `element_size` could be 0 if `==TRITONSERVER_TYPE_INVALID`.

src/http_server.cc
```diff
@@ -3597,46 +3600,32 @@ HTTPAPIServer::GenerateRequestClass::ExactMappingInput(
...
           ("Input '" + name + "' has a byte_size (" +
            std::to_string(byte_size) +
            " bytes) that exceeds the maximum allowed value of " +
-           std::to_string(max_input_size_) +
+           std::to_string(max_input_size_ - consumed_input_byte_size) +
            " bytes. Use --http-max-input-size to increase the limit.")
               .c_str());
     }
 
+    consumed_input_byte_size += byte_size;
```
yinggeh, line 3661: Possible overflow here?

yinggeh:
PR title isn't following the convention.

yinggeh:
I think test is required for such code change. Whether verify its functionality with existing ones or by creating a new test.

src/http_server.cc
whoisj: unfortunately `0 == invalid` AND `0 == bytes` 🤦‍♂️

fixed up.

src/http_server.cc
whoisj: invalid scope was the actual problem, fixed up.

qa/common/gen.ONNXRuntime.gen_qa_model_repository.docker.v2.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
yinggeh: Are they needed?

src/http_server.cc
yinggeh: It won't be BYTE since it's inside `dtype != TRITONSERVER_TYPE_BYTES` case.

qa/common/gen.ONNXRuntime.gen_qa_model_repository.docker.v2.sh
whoisj: nope, they got picked up on accident.

whoisj: removed

whoisj:
> I think test is required for such code change. Whether verify its functionality with existing ones or by creating a new test.

Added 2 new tests. Let me know if they look good to you.

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -58,7 +62,88 @@
...
+        )
+
+        self.assertEqual(
+            400,
+            response.status_code,
+            f"Expected error code for type/size mismatch, got: {response.status_code}",
+        )
+        error_msg = response.content.decode()
+        self.assertEqual(
+            '{"error":"Request JSON size of 89478576 bytes exceeds the maximum allowed value of 67108864 bytes. Use --http-max-input-size to increase the limit."}',
```
yinggeh: Would like to see tests that passing with size 67108864 but failed with 67108864+1.

qa/L0_http/test.sh
```diff
@@ -802,6 +802,14 @@ if [ $? -ne 0 ]; then
     echo -e "\n***\n*** Default Input Size Limit Test Failed for compressed input\n***"
     RET=1
 fi
+
+# Run test to verify that large inputs fail with default limit
+python http_input_size_limit_test.py InferSizeLimitTest.test_type_size_explosion >> $CLIENT_LOG 2>&1
```
yinggeh: I am confused with the naming here. Can you use a more descriptive name?

whoisj: do you have a suggestion?

whoisj:
Converted to a draft after hitting another SEGFAULT bug. Will reopen once the tests are clean.

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -58,7 +62,95 @@
...
+            f"Expected error code for type/size mismatch, got: {response.status_code}",
+        )
+        error_msg = response.content.decode()
+        self.assertIn(
+            "Request JSON size of ",
+            error_msg,
+        )
+        self.assertIn(
+            " bytes exceeds the maximum allowed input size"
+            error_msg,
```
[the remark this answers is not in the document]
whoisj: incorrect, there's no syntax error.

qa/L0_http/http_input_size_limit_test.py
whoisj: changed the format.

src/http_server.cc
```diff
@@ -2685,12 +2700,12 @@ HTTPAPIServer::ParseJsonTritonIO(
           infer_req->serialized_data_.emplace_back();
           std::vector<char>& serialized = infer_req->serialized_data_.back();
           serialized.resize(byte_size);
-
+          char* serialized_base = &serialized[0];
```
mudit-eng: `&serialized[0]` is undefined behavior if vector is empty. Safer to do something like:

char* serialized_base = serialized.data();

or 

char* serialized_base = serialized.empty() ? nullptr : serialized.data();

src/http_server.h
```diff
@@ -461,6 +465,7 @@ class HTTPAPIServer : public HTTPServer {
     const MappingSchema* request_schema_{nullptr};
     const MappingSchema* response_schema_{nullptr};
     const bool streaming_{false};
+    const size_t max_input_size_{0};
```
mudit-eng, line 466: HTTPAPIServer already has protected member:`size_t max_input_size_;`. Can this create confusion?

src/http_server.cc
```diff
@@ -3084,12 +3099,13 @@ HTTPAPIServer::EVBufferToJson(
...
+    auto overrun = length - max_input_size_;
     return TRITONSERVER_ErrorNew(
         TRITONSERVER_ERROR_INVALID_ARG,
-        ("Request JSON size of " + std::to_string(length) +
-         " bytes exceeds the maximum allowed value of " +
-         std::to_string(max_input_size_) +
-         " bytes. Use --http-max-input-size to increase the limit.")
+        ("Request JSON size of " + std::to_string(length) + " + " +
+         std::to_string(overrun) +
+         " bytes exceeds the maximum allowed input size. "
```
mudit-eng: Not sure if this message is intuitive. This does not convey what is the maximum input size.

src/http_server.cc
```diff
@@ -3563,7 +3583,8 @@ TRITONSERVER_Error*
 HTTPAPIServer::GenerateRequestClass::ExactMappingInput(
     const std::string& name,
     triton::common::TritonJson::Value& generate_request,
-    std::map<std::string, triton::common::TritonJson::Value>& input_metadata)
+    std::map<std::string, triton::common::TritonJson::Value>& input_metadata,
```
mudit-eng, line 3585: Can we use `unordered_map` here?

src/http_server.h
```diff
@@ -428,6 +429,8 @@ class HTTPAPIServer : public HTTPServer {
     const MappingSchema* RequestSchema() { return request_schema_; }
     const MappingSchema* ResponseSchema() { return response_schema_; }
 
+    size_t MaxInputSize() { return max_input_size_; }
```
mudit-eng: this is not being called anywhere

src/http_server.cc
```diff
@@ -3587,15 +3608,59 @@ HTTPAPIServer::GenerateRequestClass::ExactMappingInput(
...
+      RETURN_IF_ERR(TRITONSERVER_InferenceRequestAddInput(
+          triton_request_.get(), name.c_str(), dtype, nullptr, 0));
+      return nullptr;
+    }
+
+    // Ensure that the resulting array size in bytes does not exceed the maximum
+    // allowed input size.
+    if (byte_size + consumed_input_byte_size > max_input_size_ ||
+        byte_size + consumed_input_byte_size < consumed_input_byte_size) {
+      auto overrun = byte_size + consumed_input_byte_size - max_input_size_;
```
mudit-eng: Can `byte_size + consumed_input_byte_size` result in an overflow?

src/http_server.cc
whoisj: sure. I was just using the existing code, but I'll fix up regardless.

src/http_server.cc
whoisj: the updated format was a request from @yinggeh. Can you two decide between you which you prefer?

src/http_server.cc
whoisj: no idea, but that would be changing existing code and I'd rather not because I feel that is outside the scope of this PR.

src/http_server.h
whoisj: I'll remove.

src/http_server.cc
whoisj: no because of the check immediately previous.

src/http_server.h
whoisj: I don't see how, they're both private fields.

src/http_server.cc
mudit-eng: Yes, you checked for overflow here: `byte_size + consumed_input_byte_size < consumed_input_byte_size)` but can't there be an overflow here: `auto overrun = byte_size + consumed_input_byte_size - max_input_size_`

whoisj: while true, this value is only to produce an error message.

not sure what we would do with an overrun of a `size_t` value.

src/http_server.cc
mudit-eng: Difference would be:

New log: `Request JSON size of 1044 + 20 bytes exceeds the maximum allowed input size.`

Old log: `Request JSON size of 1044 bytes exceeds the maximum allowed value of 1024.`

I talked to Yingge. May be there is a misunderstanding. We can keep the old message. 

Plus, there is a bug in the new code. We should have printed `max_input_size_ + overrun` instead of 'length + overrun`. Can we validate the numbers too in the test?

whoisj:
@mudit-eng updated the error message per your comment. Updated tests to match.

CI-Pipeline-ID: 47658956

src/http_server.cc
whoisj: Updated. Thanks for the clarity.

src/http_server.cc
whoisj: fixed up.

qa/L0_http/generate_endpoint_test.py
```diff
@@ -274,6 +275,24 @@ def test_invalid_input_types(self):
             self.generate_expect_failure(self._model_name, inputs, error_msg)
             self.generate_stream_expect_failure(self._model_name, inputs, error_msg)
 
+    def test_type_size_explosion(self):
```
yinggeh:
```suggestion
    def test_json_dtype_size_expansion_exceeds_limit_error(self):
```

yinggeh: Please add a description comment to explain what's happening here.

whoisj: done.

whoisj: done.

src/http_server.cc
whoisj: fixed up.

mudit-eng:
Thanks for addressing all the feedbacks. :)

PR #8686, opened 2026-03-06
fix: apply HTTP restrictions to the SageMaker endpoint

mattwittwer:
#### What does the PR do?

Implement the core HTTP server’s restricted API and max input size settings in the SageMaker endpoint. 

Add regression testing for restricted operations and payload-size enforcement for the Sagemaker endpoint.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

The main changes to pass these parameters are in: 
src/main.cc
src/sagemaker_server.h

Then the implementation for the restricted endpoint checks is in:
src/sagemaker_server.cc

#### Test plan:

- CI Pipeline ID:

45550957

- closes GitHub issue: #xxx

### DIFF
--- a/src/sagemaker_server.cc
+++ b/src/sagemaker_server.cc
@@ -41,14 +41,14 @@ namespace triton { namespace server {
     }                                                 \
   } while (false)
 
-#define RETURN_AND_RESPOND_IF_RESTRICTED(REQ, RESTRICTED_CATEGORY)          \
-  do {                                                                      \
-    auto const& is_restricted_api =                                         \
-        restricted_apis_.IsRestricted(RESTRICTED_CATEGORY);                 \
-    auto const& restriction = restricted_apis_.Get(RESTRICTED_CATEGORY);    \
-    if (is_restricted_api && RespondIfRestricted((REQ), restriction)) {     \
-      return;                                                               \
-    }                                                                       \
+#define RETURN_AND_RESPOND_IF_RESTRICTED(REQ, RESTRICTED_CATEGORY)       \
+  do {                                                                   \
+    auto const& is_restricted_api =                                      \
+        restricted_apis_.IsRestricted(RESTRICTED_CATEGORY);              \
+    auto const& restriction = restricted_apis_.Get(RESTRICTED_CATEGORY); \
+    if (is_restricted_api && RespondIfRestricted((REQ), restriction)) {  \
+      return;                                                            \
+    }                                                                    \
   } while (false)
 
 namespace {


src/sagemaker_server.cc
```diff
@@ -173,11 +173,25 @@ SagemakerAPIServer::Handle(evhtp_request_t* req)
         if (multi_model_name.empty()) {
           LOG_VERBOSE(1) << "SageMaker request: LIST ALL MODELS";
 
+          if (restricted_apis_.IsRestricted(
```
yinggeh: Should create a new macro similar to `RETURN_AND_RESPOND_IF_RESTRICTED`

qa/L0_sagemaker/test.sh
```diff
@@ -612,6 +612,196 @@ set -e
...
+    -d '{"inputs":[{"name":"INPUT0","datatype":"INT32","shape":[1,1],"data":[1]},{"name":"INPUT1","datatype":"INT32","shape":[1,1],"data":[1]}]}'`
+if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** Failed. Expected small payload to succeed on SageMaker endpoint (got $code)\n***"
+    RET=1
+fi
+
+# Large payload over 256 bytes should be rejected
+rm -f ./curl.out
+LARGE_PAYLOAD='{"inputs":[{"name":"INPUT0","datatype":"INT32","shape":[1,16],"data":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16]},{"name":"INPUT1","datatype":"INT32","shape":[1,16],"data":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16]}]}'
```
yinggeh, line 779: @pskiran1 @whoisj Is there a way to test large data size without hardcoding the payload? IMO I don't think testing 256 bytes is practical.

whoisj: we could have a `.data` file that contains the tensor data and gets loaded at test runtime.

mattwittwer: So my thought with this check was to keep the SageMaker test simple, and not duplicate the full `--http-max-input-size` coverage that exists in [qa/L0_http/http_input_size_limit_test.py](https://github.com/triton-inference-server/server/blob/main/qa/L0_http/http_input_size_limit_test.py).

The `L0_http` tests already exercise the boundary behavior for JSON / binary / compressed payloads and different limit values. So this regression test just proves that `/invocations` goes through the same input-size enforcement path, and that the same configured limit also applies to the core HTTP endpoint.

If you'd prefer, I can definitely switch this to generate the oversized payload similar to some or all of the L0_http tests.

whoisj: sounds reasonable. Honestly, we just need a better overall test packaging solution - and that's very much outside the scope of this PR.

yinggeh:
This should be feat

PR #8688, opened 2026-03-08
feat: allow override of SageMaker single-model repository path

### LINKED ISSUE #8668: Provided serve file in container should allow override of model repository
**Is your feature request related to a problem? Please describe.**
When running triton in sagemaker, it uses the `/usr/bin/serve` script as a entrypoint. It seems to be hardcoded to use `/opt/ml/model` as the model repository. This is fine IF you are using sagemaker to download the triton model. If you try to use the sagemaker model to download the actual model, it will fail as its attempting to download and load a triton ensemble/model instead.

**Describe the solution you'd like**
Change `SAGEMAKER_SINGLE_MODEL_REPO=/opt/ml/model/` to `SAGEMAKER_SINGLE_MODEL_REPO=${SAGEMAKER_TRITON_MODEL_REPOSITORY:-/opt/ml/model}`

**Describe alternatives you've considered**
There isn't really any other straight forward alternatives. I do not wish to package the model and triton model inside and have sagemaker download it. 

**Additional context**
There isn't really any reason to not do this, it provides more control with no downside.

cluster2600:
#### What does the PR do?

The SageMaker `serve` script hardcodes the single-model repository path to `/opt/ml/model/`. This makes it impossible to use a custom model directory without replacing the entrypoint entirely.

This PR introduces the `SAGEMAKER_TRITON_MODEL_REPOSITORY` environment variable, which allows users to override the default model repository path whilst keeping full backwards compatibility — when the variable is unset the behaviour is identical to before.

This follows the same pattern already used by other configurable values in the script (e.g. `SAGEMAKER_TRITON_OVERRIDE_PING_MODE`, `SAGEMAKER_BIND_TO_PORT`, etc.).

#### Checklist
- [x] I have read the [Contribution guidelines](#../../CONTRIBUTING.md) and signed the [Contributor License Agreement](https://github.com/NVIDIA/triton-inference-server/blob/master/Triton-CCLA-v1.pdf)

#### Commit Type:
- [x] feat

#### Related PRs:
None

`docker/sagemaker/serve` — the one-line change on line 28 is the core of this PR.

#### Test plan:
1. Existing `L0_sagemaker` tests pass unchanged (the env var is not set so the default `/opt/ml/model/` is used).
2. A new test block in `qa/L0_sagemaker/test.sh` copies the model to a seperate directory, sets `SAGEMAKER_TRITON_MODEL_REPOSITORY` to that path, starts `serve`, and verifies inference succeeds.
3. After inference the env var is unset and the temporary directory is removed so subsequent tests are unaffected.

#### Caveats:
- Only the single-model endpoint path is made configurable. The multi-model endpoint path (`/tmp/sagemaker`) is a dummy placeholder used when `SAGEMAKER_MULTI_MODEL=true` and does not benefit from the same treatment.

#### Background
When deploying custom models on SageMaker that are not packaged in the standard Triton model-repository layout at `/opt/ml/model/`, users currently have no straightforward way to point `serve` at a different directory. The only workaround is to override the container entrypoint altogether, which is cumbersome and error-prone.

#### Related Issues:
- Closes #8668

### DIFF
--- a/docker/sagemaker/serve
+++ b/docker/sagemaker/serve
@@ -25,7 +25,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-SAGEMAKER_SINGLE_MODEL_REPO=/opt/ml/model/
+SAGEMAKER_SINGLE_MODEL_REPO=${SAGEMAKER_TRITON_MODEL_REPOSITORY:-/opt/ml/model/}
 
 # Use 'ready' for ping check in single-model endpoint mode, and use 'live' for ping check in multi-model endpoint model
 # https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/rest_predict_v2.yaml#L10-L26
--- a/qa/L0_sagemaker/test.sh
+++ b/qa/L0_sagemaker/test.sh
@@ -165,6 +165,48 @@ set -e
 kill $SERVER_PID
 wait $SERVE_PID
 
+# Test custom model repository via SAGEMAKER_TRITON_MODEL_REPOSITORY
+CUSTOM_REPO_DIR=`pwd`/custom_repo
+mkdir -p ${CUSTOM_REPO_DIR}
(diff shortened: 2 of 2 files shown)


cluster2600:
Closing — no response after 3 weeks.

PR #8696, opened 2026-03-12
test: Test safe GetElementCount and GetByteSize APIs at common and co…

yinggeh:


src/test/CMakeLists.txt
```diff
@@ -78,43 +78,53 @@ if(${TRITON_ENABLE_HTTP} OR ${TRITON_ENABLE_METRICS} OR
...
-# Unit test for Backend API
+# Unit test for Backend + Common + Core tensor size APIs (GetElementCount,
+# GetByteSize). Core template headers are used directly; the few core symbols
+# not available via libtritonserver.so (hidden by linker script) are defined
+# in the test source.
 #
-add_executable(
-  backend_tensor_size_test
-  backend_tensor_size_test.cc
-)
```
mc-nv, line 86: Are these files were removed?

src/test/CMakeLists.txt
```diff
@@ -78,43 +78,53 @@ if(${TRITON_ENABLE_HTTP} OR ${TRITON_ENABLE_METRICS} OR
...
-    INSTALL_RPATH_USE_LINK_PATH FALSE
-    INSTALL_RPATH ""
-)
+  set_target_properties(
+    tensor_size_test
+    PROPERTIES
+      SKIP_BUILD_RPATH TRUE
+      BUILD_WITH_INSTALL_RPATH TRUE
+      INSTALL_RPATH_USE_LINK_PATH FALSE
+      INSTALL_RPATH "$ORIGIN/../lib"
```
mc-nv, line 98: I never had chance to evaluate `ORIGIN` value at glance it's something we never declare. 
Can we omit it?

src/test/CMakeLists.txt
```diff
@@ -78,43 +78,53 @@ if(${TRITON_ENABLE_HTTP} OR ${TRITON_ENABLE_METRICS} OR
...
-  PRIVATE
-    ${CMAKE_CURRENT_SOURCE_DIR}/..
-    ${GTEST_INCLUDE_DIRS}
-)
+  target_include_directories(
+    tensor_size_test
+    PRIVATE
+      ${CMAKE_CURRENT_SOURCE_DIR}/..
+      ${GTEST_INCLUDE_DIRS}
+      ${repo-core_SOURCE_DIR}/src
```
mc-nv, line 106: Why entire source code directory has been added?

src/test/CMakeLists.txt
```diff
@@ -78,43 +78,53 @@ if(${TRITON_ENABLE_HTTP} OR ${TRITON_ENABLE_METRICS} OR
...
+    tensor_size_test
+    PRIVATE
+      triton-backend-utils
+      triton-core-serverapi
+      triton-core-serverstub
+      triton-common-model-config
+      proto-library
+      protobuf::libprotobuf
+      GTest::gtest
+      GTest::gtest_main
```
mc-nv, line 120: I see there was a target condition used with generation expression, why this part doesn't use same pattern?

src/test/tensor_size_test.cc
```diff
@@ -35,10 +35,25 @@
 #include <string>
 #include <vector>
 
+#include "model_config_utils.h"
+#undef RETURN_IF_ERROR  // core (Status) vs backend (TRITONSERVER_Error*); avoid
```
mc-nv, line 39: Where this definition is coming from?

PR #8727, opened 2026-04-08
build:  remove vllm flags

mc-nv:
<sub>
Closes: TRI-926 <br>
Related: #8698
</sub>

----

## Description 
Removing unused vllm flags as container now sourcing pre-installed version from DLFW vLLM image

### DIFF
--- a/build.py
+++ b/build.py
@@ -78,7 +78,6 @@ DEFAULT_TRITON_VERSION_MAP = {
     "ort_openvino_version": "2026.0.0",
     "standalone_openvino_version": "2026.0.0",
     "dcgm_version": "4.5.3-1",
-    "vllm_version": "0.16.0",
     "rhel_py_version": "3.12.3",
 }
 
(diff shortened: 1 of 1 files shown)


whoisj (approved):

yinggeh:
I use `build.py` as a reference to vLLM version. If it's removed from here, where can I easily find the vLLM version without pulling and running container.

mc-nv:
> I use `build.py` as a reference to vLLM version. If it's removed from here, where can I easily find the vLLM version without pulling and running container.

You bring a good case @yinggeh. <br>
The right way to seek which version we are using is by referring to DLFW vLLM image itself instead of the `build.py` 
For instance current version we are removing is: 
```bash
$ docker run --rm -it nvcr.io/nvidia/vllm:26.02-py3  bash -c "pip3 list " | grep vllm
vllm                               0.15.1+nv26.2                        44440029
$ docker run --rm -it nvcr.io/nvidia/vllm:26.03-py3  bash -c "pip3 list " | grep vllm
vllm                                     0.17.1+a03ca76a.nv26.3.46967107.cu132

```

geraldstanje1:
@mc-nv @whoisj since triton has delay to update to latest vllm version - how can someone upgrade to e.g. vllm 0.20.0 if vllm_version is removed?

looks like this is the right way? https://github.com/triton-inference-server/vllm_backend

also where i see the build process for building vLLM image nvcr.io/nvidia/vllm:26.02-py3 ?

PR #8733, opened 2026-04-10
fix: Avoid Reusing Closed File Descriptor

whoisj:
This change removes caching of file descriptors that are immediately closed. Removing the caching, removes the temptation to use the cached values to use the closed file descriptor.

CI Pipeline ID: 48236947

### DIFF
--- a/src/shared_memory_manager.cc
+++ b/src/shared_memory_manager.cc
@@ -1,4 +1,4 @@
-// Copyright 2019-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2019-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions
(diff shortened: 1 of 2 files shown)


src/shared_memory_manager.cc
```diff
@@ -367,23 +367,10 @@ SharedMemoryManager::RegisterSystemSharedMemory(
   }
 
   // register
-  void* mapped_addr;
-  int shm_fd = -1;
-
-  // don't re-open if shared memory is already open
-  for (auto itr = shared_memory_map_.begin(); itr != shared_memory_map_.end();
```
yinggeh, line 374: Can't we evict cached entry if fd is closed?

whoisj: what would be the point? The fd is closed immediately.

yinggeh: So the caching itself is pointless?

whoisj: Correct.

yinggeh (approved):
LGTM. Make sure CI is happy

PR #8737, opened 2026-04-14
fix: async model load/unload to prevent evhtp thread starvation

### LINKED ISSUE #8635: HTTP Connection Distribution Imbalance in evhtp Causes Sequential Request Processing
**Description**
Triton Inference Server experiences unbalanced connection distribution across worker threads when using the HTTP endpoint, resulting in sequential request processing despite having multiple worker threads available.

The evhtp library used by Triton has a main thread that blocks on an event base listening for events on the input socket. When connections arrive, they are dispatched to worker threads through a pipe (a pair of sockets) mechanism.

Under certain conditions, the following sequence occurs:
1. Main thread handles incoming connection and sends it to pipe
2. Thread 1 accepts connection from pipe
3. Main thread handles next incoming connection and sends it to pipe
4. Thread 1 accepts connection again (same thread)

When the same thread repeatedly wins the race to accept connections from the pipe, requests are processed sequentially on that single thread, even when other worker threads are idle.

**Triton Information**
Image: `nvcr.io/nvidia/tritonserver:25.06-py3`
Backend: `python`

**To Reproduce**
1. Create a docker container with a triton server and the `add_sub` model by following [this guide](https://github.com/triton-inference-server/python_backend?tab=readme-ov-file#quick-start)
2. Add a delay to the initialize method of the add_sub model to simulate long model loading:
```
def initialize(self, args):
     # Sleep for one minute
     time.sleep(60)
```
3. Copy the add_sub model 6 times add_sub1, add_sub2, add_sub3 ... add_sub6
4. Start a triton server with the following command:
```
tritonserver \
    --model-repository=<path_to_models_directory> \
    --model-control-mode=explicit \
    --load-model=* \
    --model-load-thread-count=10 \
    --http-thread-count=10
```
5. Touch each model.py to simulate that the models have changed:
```
#!/bin/bash

for x in {1..6}; do
    touch "/models/add_sub${x}/1/model.py"
    echo "Touched models/add_sub${x}/1/model.py"
done
```

6.  Make concurrent requests to Triton to load all models:
```
#!/bin/bash

for x in {1..6}; do
    echo "Loading model: add_sub${x}"
    curl -X POST "localhost:8000/v2/repository/models/add_sub${x}/load" &
    echo ""
done
wait
```

**Expected behavior**
All 6 models should begin loading simultaneously, with logs appearing at the same time:
```
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub1:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub2:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub3:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub4:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub5:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub6:1"
````

**Actual Behavior**
Some models start loading with a one-minute delay (matching the delay set in the initialize method), indicating sequential processing:
```
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub1:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub2:1"
I0203 09:10:40.123456 666 model_lifecycle.cc:473] "loading: add_sub3:1"
I0203 09:11:40.435898 666 model_lifecycle.cc:473] "loading: add_sub4:1"
I0203 09:11:40.435898 666 model_lifecycle.cc:473] "loading: add_sub5:1"
I0203 09:11:40.435898 666 model_lifecycle.cc:473] "loading: add_sub6:1"
```

**Possible solution**
The core of this problem is that libevhtp was designed to have lightweight handlers that execute very quickly. When executing large blocking tasks, they should be rescheduled to a separate thread pool.

A potential solution is to use the [thread pool from the common library](https://github.com/triton-inference-server/common/blob/main/src/thread_pool.cc) with a limit for a queue size (or even rewrite it as a small workstealing tp) and add it as a field in the HTTPAPIServer class, where blocking handlers would be executed asynchronously.

### PR DESCRIPTION
**Pull request by itsnothuy**
#### What does the PR do?

Fixes synchronous model load/unload calls blocking evhtp worker threads
in `HandleRepositoryControl()`, causing HTTP thread starvation under
concurrent load/unload traffic. Health probes, inference requests, and
metadata queries are starved until all load/unload operations complete.

The fix applies the same `evhtp_request_pause` → detached `std::thread`
→ `evthr_defer` async pattern already used by `InferRequestClass` for
inference requests. A new `ControlRequestClass` (modelled exactly on
`InferRequestClass`) handles the lifecycle: capture the owning evhtp
thread, pause the request to free the worker, run the blocking
`TRITONSERVER_ServerLoadModelWithParameters` /
`TRITONSERVER_ServerUnloadModel` call on a detached thread, then post
the reply back via `evthr_defer`. An `std::atomic<int>` counter gates
concurrency and returns HTTP 503 when the limit is reached, preventing
thread explosion.

---

#### Checklist

---

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

---

#### Related PRs:
None. This change is self-contained to src/http_server.h, src/http_server.cc, and src/main.cc. It does not touch the gRPC, SageMaker, or Vertex AI servers, or any backend.

---

1. **`src/http_server.h`** — `ControlRequestClass` (new inner class,
   ~65 lines after `InferRequestClass`'s enclosing section). Compare
   side-by-side with `InferRequestClass` to verify the async pattern is
   identical.

2. **`src/http_server.cc`** — `HandleRepositoryControl()` (~line 1447).
   The load path and unload path both follow the same structure:
   parse synchronously → gate with `fetch_add` → `new ControlRequestClass(req)`
   → `std::thread(...).detach()` → inside thread: call API, `fetch_sub`,
   `evthr_defer(ReplyCallback)`.

3. **`src/main.cc`** — `StartHttpService()` (~line 138). One-liner:
   `model_load_thread_count_` inserted as the new
   `control_request_concurrency` argument to `HTTPAPIServer::Create()`.

4. **Constructor init** in `src/http_server.cc` (~line 1167):
   `max_control_requests_ = control_request_concurrency` + log line.

---

#### Test plan:

**Manual reproduction of the bug (before this fix):**

```bash
# 1. Start Triton with a slow-loading Python model (time.sleep(60) in initialize())
tritonserver \
  --model-repository=/models \
  --model-control-mode=explicit \
  --http-thread-count=2 \
  --model-load-thread-count=2

# 2. In parallel: issue 2 concurrent load requests + 1 health probe
curl -X POST localhost:8000/v2/repository/models/slow_model_1/load &
curl -X POST localhost:8000/v2/repository/models/slow_model_2/load &
# Before fix: this hangs for 60s — all evhtp threads are blocked:
curl localhost:8000/v2/health/live

# After fix: health probe returns 200 immediately even while loads are in-flight
```

**Existing test coverage:**

- No new L0 test added in this PR. The fix follows the identical pattern
  as `InferRequestClass` which is covered by existing `qa/L0_http/`
  inference tests (verifying the `evhtp_request_pause` /
  `evthr_defer` lifecycle is exercised).
- Existing `qa/L0_http/` tests should pass without regression; the
  async path is only taken when `--model-load-thread-count > 0`
  (default 4).
  
> **Note**: I do not yet have access to a Linux + CUDA build environment
> to run the full CI locally. I am happy to add an L0 test in a follow-up
> PR once the approach is confirmed acceptable, or if a reviewer can
> advise on the test infrastructure setup.

---

#### Caveats:

1. **`--model-load-thread-count` dual use**: This flag now controls both
   Triton's internal model loading parallelism
   (`TRITONSERVER_ServerOptionsSetModelLoadThreadCount`) and the HTTP
   concurrency limit for load/unload requests. Setting it to `0` disables
   the async path and falls back to the original synchronous behaviour.
   Documentation update for this dual use is pending and can be done in a
   follow-up.

2. **Shutdown safety**: Detached threads hold raw pointers to
   `HTTPAPIServer` members (`control_request_cnt_`, `server_.get()`).
   If the server is destroyed while a model load/unload is in-flight, a
   use-after-free is theoretically possible. This is the **same
   pre-existing risk** accepted by `InferRequestClass`; the fix does not
   introduce a new hazard. A comprehensive fix (e.g., `shared_ptr`
   to the counter, or joining on shutdown) should address all async paths
   simultaneously in a separate PR.

3. **`evthr_defer` return value unchecked**: Same as `InferRequestClass`
   line 4017 in the existing code. If the event loop has stopped,
   `ctrl_req` will leak. Not introduced by this PR.

4. **Second `Create()` overload** (options-map path, used in some
   embedded/SageMaker deployments): hardcodes `control_concurrency = 4`
   and reads from an options key `"control_request_concurrency"` that is
   not currently set by any caller. Normal `main.cc` startup is
   unaffected. Happy to align this in a follow-up if the reviewer prefers.

---

#### Background

Reported in issue #8635 by @aleksn7. Root cause confirmed:
`HandleRepositoryControl()` in `src/http_server.cc` calls
`TRITONSERVER_ServerLoadModelWithParameters()` synchronously on the evhtp
worker thread with no `evhtp_request_pause()` before the call. The evhtp
thread is fully blocked for the duration of model initialization
(potentially minutes for large models). With `--http-thread-count=N`,
N concurrent load requests exhaust the entire thread pool, starving
inference and Kubernetes health probes.

@aleksn7 provided explicit guidance on the approach in the issue:

> "another approach is to skip the thread pool entirely and create a new
> thread for each model_repository call … don't forget to limit the
> concurrent request count … you could use `--model-load-thread-count`
> CLI parameter."

This PR implements exactly that approach.

The `InferRequestClass` async pattern (lines 3931–3944, 3887–3910 of
`src/http_server.cc`) was used as the template. `ControlRequestClass`
is a simplified version: no inference cancellation in the fini hook
(not needed — there is no in-flight inference to cancel), no response
count tracking. All other mechanics (`evhtp_request_pause`,
`htpconn->thread` capture, `evhtp_hook_on_request_fini`,
`evthr_defer`, `evhtp_send_reply` + `evhtp_request_resume` in the
callback) are identical.

---

#### Related Issues:

- Fixes #8635

---


tnsahr2580:
Hi — following this PR since we hit the same evhtp thread starvation during
concurrent model load/unload. It's been open since April without a review.
Is something blocking it (missing L0 tests? design still under discussion?),
or is it just waiting in the queue?

PR #8768, opened 2026-05-06
fix: ignore SIGPIPE to prevent server crash on S3 idle connection timeout

### LINKED ISSUE #8767: Bug: tritonserver crashes with Exit Code 141 (SIGPIPE) when loading models from S3 due to idle connection timeout
**Description**
The `tritonserver` process crashes and the container restarts in an infinite loop exactly **~300 seconds** into the boot sequence while loading models sequentially from an S3 bucket. No error is logged and the process simply vanishes with **Exit Code 141** (`128 + 13`, i.e., killed by `SIGPIPE`).

This occurs because S3 silently drops idle keep-alive TCP connections after exactly 300 seconds. If Triton is busy initializing a slow model (e.g., 5+ minutes), the HTTP keep-alive connection sits idle. When initialization finishes, the AWS SDK calls `write()` on the dead socket to fetch the next model. Because `tritonserver` does not handle `SIGPIPE`, the OS instantly terminates the entire process.

**Triton Information**
What version of Triton are you using?
`24.05` (Reproducible on all recent releases)

Are you using the Triton container or did you build it yourself?
Official NGC Triton container (`nvcr.io/nvidia/tritonserver:<tag>-py3`)

**To Reproduce**
Steps to reproduce the behavior.

1. Create two models in an S3 bucket (`model_A` and `model_B`).
2. Start the Triton container loading models from S3:
   ```bash
   docker run \
     -e AWS_ACCESS_KEY_ID=... \
     -e AWS_SECRET_ACCESS_KEY=... \
     nvcr.io/nvidia/tritonserver:<tag>-py3 \
     tritonserver --model-repository s3://<bucket>/models
   ```
3. Triton will initialize `model_A` and sleep. S3 will silently kill the idle TCP connection at the 300-second mark.
4. When `model_A` finishes sleeping, Triton attempts to use the dead S3 connection to download `model_B`. The server instantly crashes with `Exit Code 141`.

Describe the models (framework, inputs, outputs), ideally include the model configuration file (if using an ensemble include the model configuration file for that as well).

The easiest way to reproduce this deterministically without a heavy GPU model is to use the `python` framework to artificially simulate a 5-minute model initialization.
In `model_A/1/model.py`, add a `time.sleep(310)` to the `initialize()` function to artificially force the S3 connection to sit idle and time out:
```python
import time
class TritonPythonModel:
    def initialize(self, args):
        print("Sleeping for 310 seconds to force S3 idle timeout...")
        time.sleep(310) 
    def execute(self, requests): return requests
```

**Expected behavior**
The AWS C++ SDK is designed to gracefully handle `EPIPE` and automatically reconnect. Triton should globally ignore `SIGPIPE` at the executable level so that the `write()` system call can return `-1` with `errno=EPIPE`, rather than letting the OS instantly terminate the process.

Its-Tanay:
#### What does the PR do?
This PR globally sets the `SIGPIPE` disposition to `SIG_IGN` inside the `tritonserver` executable. 

This prevents an unhandled `SIGPIPE` from crashing the entire server with Exit Code 141 when the AWS C++ SDK attempts to write to an S3 keep-alive connection that has timed out during a slow model initialization. By ignoring the signal, the `write()` system call correctly returns `EPIPE`, allowing the AWS SDK to gracefully catch the error and reconnect to S3.

#### Checklist
- [x] I have read the [Contribution guidelines](#../../CONTRIBUTING.md) and signed the Contributor License Agreement

#### Commit Type:
- [x] fix

#### Related PRs:
N/A

`src/triton_signal.cc` -> `RegisterSignalHandler()`

#### Test plan:
Because this is an OS-level signal handler fix for a network idle timeout, standard unit tests cannot reliably cover it. I have verified this manually, and the exact deterministic reproduction steps using a dummy Python model (which forces the required 300s S3 idle timeout) are thoroughly documented in the attached GitHub issue.

#### Caveats:
None. The fix is explicitly placed in the standalone `tritonserver` executable (`triton_signal.cc`) rather than `triton-core` to ensure there are no global signal side-effects for users embedding Triton as a C++ library.

#### Background
This was discovered while loading a sequence of models from an S3 model repository, where one model took > 5 minutes to initialize on the GPU, causing the AWS connection to sit idle and be killed by the network layer.

#### Related Issues: 
- Fixes #8767


Its-Tanay:
Hey folks, were you able to review this ? @pskiran1 @whoisj @yinggeh (sorry to bother)

whoisj:
The change looks reasonable to me. Need to go through our internal lab's validation. I'll try to get that done today. Thank you for your patience.

Its-Tanay:
Hey @whoisj, any updates on this ?

whoisj:
CI Pipeline ID: 51849071

Its-Tanay:
Thanks for running it @whoisj! Let me know if anything comes up. 🙇

whoisj (approved):

whoisj:
This change will be part of the Triton Inference Server 26.06 release.

PR #8769, opened 2026-05-06
fix: Pre-allocate serialized buffer for gRPC BYTES input

pskiran1:
#### What does the PR do?

When a gRPC `ModelInferRequest` carries `BYTES` data via the explicit `contents.bytes_contents` field, `InferGRPCToInput` previously appended each element into an empty `std::string` without reserving capacity. As the destination grew, the string had to repeatedly reallocate and copy its existing contents into a larger buffer, which is expensive in CPU and memory and can fragment the heap when `bytes_contents` carries a large number of elements. This change pre-computes the total serialized size and calls `reserve()` once before the append loop, so the buffer grows in a single allocation. Adds a regression test in `L0_grpc/python_unit_test.py` that exercises the many-element path.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 50441852

- closes GitHub issue: #xxx

### DIFF
--- a/qa/L0_grpc/python_unit_test.py
+++ b/qa/L0_grpc/python_unit_test.py
@@ -113,9 +113,7 @@ class GrpcTest(unittest.TestCase):
         response = grpc_stub.ModelInfer(request, timeout=120)
 
         self.assertEqual(list(response.outputs[0].shape), [element_count])
-        self.assertEqual(
-            len(response.raw_output_contents[0]), expected_serialized_size
-        )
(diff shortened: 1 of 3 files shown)


whoisj (approved):
LGTM

yinggeh (approved):

PR #8770, opened 2026-05-07
fix: Cap chunked HTTP request chunk count at 65536 to prevent memory exhaustion

yinggeh:
#### What does the PR do?
Caps the number of non-empty chunks per chunked HTTP request at 65536 (`1 << 16`) to mitigate a memory-exhaustion vector where a client streams many tiny chunks and libevhtp retains each one in its evbuffer, growing server RSS without bound.

Implementation:
- Hooks `evhtp_hook_on_headers` (`ChunkCountReset`) and `evhtp_hook_on_new_chunk` (`ChunkCountIncrement`) per connection.
- Counter is `thread_local` on the worker thread and reset per request.
- On exceeding the cap: emits a JSON 400 error, drops keepalive, and calls `bufferevent_disable(req->conn->bev, EV_READ)` so the server stops ingesting bytes while it flushes the error response (otherwise memory continues growing during the flush).

#### Checklist

#### Commit Type:
- [x] fix

#### Related PRs:
https://github.com/triton-inference-server/third_party/pull/75

#### Test plan:
- `L0_http` — `http_request_many_chunks.py` covers under-limit (accepted) and over-limit (rejected with 400, connection closed) paths.
- `L0_sagemaker` — equivalent over-limit coverage on the SageMaker endpoint.

- CI Pipeline ID: 50576040

#### Caveats:
- Limit is a compile-time constant (`kMaxChunkedChunks = 1 << 16`); not configurable at runtime.
- Counter is per-request, not per-connection — keepalive reuse is unaffected.

#### Background
Triton Inference Server high memory usage when receiving HTTP requests composed of a large number of small chunks. libevhtp accumulates each chunk in its evbuffer, so an attacker can drive RSS up without sending much total bytes. Capping the chunk count bounds the fan-out and the early `EV_READ` disable bounds memory while the error response drains.

### DIFF
--- a/qa/L0_http/http_request_many_chunks.py
+++ b/qa/L0_http/http_request_many_chunks.py
@@ -78,6 +78,8 @@ class HTTPRequestManyChunksTest(unittest.TestCase):
             try:
                 s.sendall(b"0\r\n\r\n")
             except (BrokenPipeError, ConnectionResetError):
+                # Server may close/reset early after deciding on an error response.
+                # Ignore send failure here and continue reading any available response bytes.
                 pass
 
             response = b""
--- a/qa/L0_http/http_request_many_chunks.py
+++ b/qa/L0_http/http_request_many_chunks.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python
-# Copyright 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -26,23 +26,34 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import socket
+import sys
 import unittest
+
+sys.path.append("../common")
+from test_util import MIB, get_server_process_from_env, wait_for_stable_rss
 
 
 class HTTPRequestManyChunksTest(unittest.TestCase):
     def setUp(self):
-        self._model_name = "simple"
         self._local_host = "localhost"
         self._http_port = 8000
-        self._malicious_chunk_count = (
-            1000000  # large enough to cause a stack overflow if using alloca()
-        )
-        self._parse_error = (
-            "failed to parse the request JSON buffer: Invalid value. at 0"
+        self._model_name = "simple"
+        # Must match server kMaxChunkedChunks (http_server.cc).
+        self._k_max_chunked_chunks = 65536
+        self._over_max_chunks_error = f"Chunked request body exceeds maximum of {self._k_max_chunked_chunks} non-empty chunks. Send fewer or larger HTTP chunks."
+
+    def _infer_chunked_header(self):
+        return (
+            f"POST /v2/models/{self._model_name}/infer HTTP/1.1\r\n"
+            f"Inference-Header-Content-Length: 0\r\n"
         )
 
     def send_chunked_request(
-        self, header: str, chunk_count: int, expected_response: str
+        self,
+        header: str,
+        chunk_count: int,
+        expected_response: str,
+        expected_http_status=400,
     ):
         s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         header = (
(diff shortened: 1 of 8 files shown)


qa/L0_http/http_request_many_chunks.py
```diff
@@ -60,92 +71,104 @@
...
-                s.send(b"1\r\nA\r\n")
-            # End chunked encoding
-            s.sendall(b"0\r\n\r\n")
+                try:
+                    s.send(b"1\r\nA\r\n")
+                except (BrokenPipeError, ConnectionResetError):
+                    break
+            try:
+                s.sendall(b"0\r\n\r\n")
+            except (BrokenPipeError, ConnectionResetError):
```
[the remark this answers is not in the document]
yinggeh: Fixed

qa/L0_sagemaker/sagemaker_request_many_chunks.py
```diff
@@ -54,11 +67,16 @@
...
-                s.send(b"1\r\nA\r\n")
-            # End chunked encoding
-            s.sendall(b"0\r\n\r\n")
+                try:
+                    s.send(b"1\r\nA\r\n")
+                except (BrokenPipeError, ConnectionResetError):
+                    break
+            try:
+                s.sendall(b"0\r\n\r\n")
+            except (BrokenPipeError, ConnectionResetError):
```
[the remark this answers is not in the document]
yinggeh: fixed

src/http_server.cc
```diff
@@ -48,6 +50,13 @@
 
 namespace triton { namespace server {
 
+// Transfer-Encoding: chunked — count non-empty chunks per request (worker
+// thread); reset in ChunkCountReset, increment in ChunkCountIncrement. Caps
+// evbuffer fan-out / RSS.
+thread_local uint64_t tls_http_chunk_n = 0;
```
whoisj: use of `thread_local` always makes me nervous as it often doesn't result in what you expect.

Copilot correctly points out a major risk using it as a single thread could handle multiple connections in parallel. Do we know that a thread never handles more than a single connection at a time?

yinggeh: Fixed

qa/common/test_util.py
```diff
@@ -37,6 +38,45 @@
...
+        rss = server.memory_info().rss
+        if last_rss is not None and abs(rss - last_rss) < rss_tolerance_bytes:
+            stable_count += 1
+        else:
+            stable_count = 0
+        last_rss = rss
+        if stable_count >= stable_threshold:
+            break
+        time.sleep(0.1)
+    return
```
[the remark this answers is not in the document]
yinggeh: Push back for fixes

src/http_server.cc
```diff
@@ -48,6 +50,13 @@
...
+constexpr uint64_t kMaxChunkedChunks =
+    1 << 16;  // reject on chunk count > kMaxChunkedChunks (65536)
```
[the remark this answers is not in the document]
yinggeh: fixed

src/http_server.cc
```diff
@@ -288,12 +295,50 @@ HTTPServer::NewConnection(evhtp_connection_t* conn, void* arg)
...
+    EVBufferAddErrorJson(req->buffer_out, msg.c_str());
+    // Force connection close after flushing this error response.
+    req->flags &= ~EVHTP_REQ_FLAG_KEEPALIVE;
+    if ((req->conn != nullptr) && (req->conn->bev != nullptr)) {
+      // Stop ingesting request bytes once over limit so memory won't keep
+      // growing while we flush the error response.
+      bufferevent_disable(req->conn->bev, EV_READ);
+    }
+    evhtp_send_reply(req, EVHTP_RES_BADREQ);
+    return EVHTP_RES_OK;
```
whoisj, line 334: is `EVHTP_RES_OK` the correct return code? isn't this an error condition?

src/http_server.cc
```diff
@@ -288,12 +295,50 @@ HTTPServer::NewConnection(evhtp_connection_t* conn, void* arg)
...
+{
+  (void)arg;
+  if ((req->flags & EVHTP_REQ_FLAG_FINISHED) != 0) {
+    return EVHTP_RES_OK;
+  }
+
+  if (chunk_len == 0) {
+    return EVHTP_RES_OK;
+  }
+  if (++req->chunk_count > kMaxChunkedChunks) {
```
whoisj, line 319: where is `req->chunk_count` initialized to zero?

yinggeh: Answered in https://github.com/triton-inference-server/third_party/pull/75#discussion_r3203972595

src/http_server.cc
yinggeh: - The chunk hook’s return is not “send this HTTP status”; libevhtp treats != EVHTP_RES_OK (200) from on_new_chunk as fatal parse failure.
- At the end of the same read callback, if the parser is in an error state, libevhtp evhtp_connection_free’s the connection.
- evhtp_send_reply only queues output (bufferevent_write_buffer); freeing the bufferevent does not guarantee the peer saw the full 400 + JSON.

whoisj (approved):
LGTM

PR #8771, opened 2026-05-08
test: Add Torch AOTI Tests

whoisj:
This change:
- Creates a new L0_torch_aoti test suit.
- Adds complex Torch AOTI model generation to qa/common/gen_qa_models.py.
- Cleans up existion AOTI model generation in qa/common/gen_qa_models.py.
- Enabled torchvision AOTI model generation in qa/common/gen_qa_model_repository.

### DIFF
--- a/qa/L0_torch_aoti/.gitignore
+++ b/qa/L0_torch_aoti/.gitignore
@@ -1,7 +1,0 @@
-models/
-
-*.log
-
-1
-2
-test_results
--- a/qa/L0_torch_aoti/.gitignore
+++ b/qa/L0_torch_aoti/.gitignore
@@ -1,0 +1,7 @@
+models/
+
+*.log
+
+1
+2
+test_results
--- a/qa/L0_torch_aoti/test.sh
+++ b/qa/L0_torch_aoti/test.sh
@@ -1,0 +1,147 @@
+#!/bin/bash
+# Copyright 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#  * Neither the name of NVIDIA CORPORATION nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+source ../common/util.sh
+
+if [[ "${DEBUG}" == "true" ]]; then
+    set -x
+else
+    set +x
+fi
+
+COLOR_DARK="\033[90m"
+COLOR_ERROR="\033[31m"
+COLOR_INFO="\033[94m"
+COLOR_RESET="\033[0m"
+COLOR_STATUS="\033[36m"
+COLOR_SUCCESS="\033[32m"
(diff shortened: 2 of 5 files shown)


qa/L0_torch_aoti/.gitignore
```diff
@@ -0,0 +1,7 @@
+models/
```
yinggeh: I would remove `.gitignore`.

whoisj: I wouldn't because when doing local testing this is highly valuable to avoid accidentally checking in files that should not be there.

yinggeh: It'll also be difficult to see which files are generated by the test. Eventually leaving garbage files in the test directory. In some cases, tests will fail unless starting at a clean state. IMO, I strongly object to ignore them based on my dev experience.

whoisj: clean up is not the responsibility of .gitignore. I'll remove the file, but I believe your assertion is wrong.

qa/common/gen_qa_models.py
yinggeh: Function name need to be more specific since it's for torch I/O.

qa/common/gen_qa_models.py
```diff
@@ -1298,69 +1299,44 @@ def generate_sample_inputs(
     input_shape = [abs(ips) for ips in input_shape]
 
     if input_dtype == np.int8:
```
yinggeh: Why not simply create a numpy to torch dtype map?

qa/common/gen_qa_models.py
```diff
@@ -1298,69 +1299,44 @@ def generate_sample_inputs(
...
-        input0 = torch.randint(-128, 127, input_shape, dtype=torch.int8, device=device)
-        input1 = torch.randint(-128, 127, input_shape, dtype=torch.int8, device=device)
+        input0 = torch.zeros(input_shape, dtype=torch.int8, device=device)
```
yinggeh: What's the reasoning behind this change? How do you tell backend is working correctly for a zero-value output tensor?

qa/common/gen_qa_models.py
```diff
@@ -1418,7 +1394,7 @@ def create_torch_aoti_modelfile(
         )
         return False
 
-    model_version_dir = f"{models_dir}/{model_name}/{model_version}"
+    model_version_dir = os.path.join(models_dir, model_name, str(model_version))
```
yinggeh, line 1384: Good use case of `os.path`

qa/common/gen_qa_models.py
```diff
@@ -1684,14 +1812,170 @@ def create_torch_aoti_modelconfig(
...
+    label_path = os.path.join(config_dir, label_filename)
+
+    with open(label_path, "w") as file:
         for l in range(output_label_cnt):
             file.write(f"label{l}\n")
-        print(f"Created {config_dir}/{label_filename}")
+        print(f"Created {label_path}")
+
+
+def create_torch_aoti_complex_modelconfig(
```
yinggeh:
```suggestion
def create_torch_aoti_complex_model_config(
```

qa/L0_torch_aoti/test.sh
```diff
@@ -0,0 +1,148 @@
...
+    set +x
+fi
+
+COLOR_DARK="\033[90m"
+COLOR_ERROR="\033[31m"
+COLOR_INFO="\033[94m"
+COLOR_RESET="\033[0m"
+COLOR_STATUS="\033[36m"
+COLOR_SUCCESS="\033[32m"
+COLOR_WARNING="\033[33m"
```
yinggeh, line 42: Consider moving to `qa/common/util.sh` in the future.

qa/common/gen_qa_models.py
whoisj: what do you mean? This function generates sample inputs, not translate types. Sorry, just confused.

qa/common/gen_qa_models.py
whoisj: fixed up.

qa/common/gen_qa_models.py
whoisj: these are just sample values used by PyTorch to compile the model. These aren't used for inference.

qa/common/gen_qa_models.py
whoisj: all other similar functions are named `modelconfig` not `model_config`, I'm just following the existing pattern.

yinggeh: I see. "Model config" should really be two words.

whoisj: agreed. I can change them all, what do you think?

yinggeh: For now I am fine with only changing this new method.

qa/common/gen_qa_models.py
yinggeh: Something like
```
np_to_torch_dtype_dict = {
    np.int8: torch.int8,
    ...
}

...

if input_dtype in np_to_torch_dtype_dict:
    input0 = torch.zeros(input_shape, dtype=np_to_torch_dtype_dict[input_dtype], device=device)
    input1 = torch.zeros(input_shape, dtype=np_to_torch_dtype_dict[input_dtype], device=device)
else:
    print(
        f"{_color_yellow}warning: dtype {input_dtype} is unsupported; falling back to torch.int32{_color_reset}"
    )
    input0 = torch.zeros(input_shape, dtype=torch.int32, device=device)
    input1 = torch.zeros(input_shape, dtype=torch.int32, device=device)
```

qa/common/gen_qa_models.py
whoisj: updated.

qa/common/gen_qa_models.py
whoisj: fixed up.

qa/common/gen_qa_models.py
```diff
@@ -1289,78 +1290,36 @@ def forward(self, INPUT0, INPUT1):
...
-        input1 = torch.randint(
-            0, 4294967295, input_shape, dtype=torch.uint32, device=device
-        )
-    elif input_dtype == np.uint64:
-        input0 = torch.randint(
-            0, 18446744073709551615, input_shape, dtype=torch.uint64, device=device
-        )
-        input1 = torch.randint(
-            0, 18446744073709551615, input_shape, dtype=torch.uint64, device=device
+    np_to_dtype = {
```
yinggeh:
```suggestion
    np_to_torch_dtype = {
```

whoisj: done.

yinggeh (approved):

PR #8787, opened 2026-05-18
feat: Add HTTP request body size limit to OpenAI frontend

pskiran1:
#### What does the PR do?

Adds an HTTP request body size limit to the Triton OpenAI frontend to reject oversized requests before JSON parsing. The PR introduces a FastAPI/ASGI middleware that:
- Rejects requests early when `Content-Length` exceeds the configured limit.
- Counts streamed/chunked request bodies and rejects them once they exceed the limit.
- Returns a structured OpenAI-style error response with HTTP `413 content_too_large`.
- Rejects malformed `Content-Length` headers with HTTP `400 invalid_content_length`.

The default limit is `67108864` bytes / 64 MiB, and it can be configured with the new `--http-max-input-size` CLI option.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID: 52210748

- closes GitHub issue: #xxx


python/openai/openai_frontend/utils/utils.py
```diff
@@ -26,6 +26,9 @@
 
 from enum import IntEnum
 
+# Default maximum allowed HTTP request input size in bytes (64 MiB).
```
yinggeh:
```suggestion
# Default maximum allowed HTTP request input size in bytes (64 MiB).
# Same as HTTP_DEFAULT_MAX_INPUT_SIZE in src/common.h.
```

python/openai/openai_frontend/utils/utils.py
```diff
@@ -44,4 +48,15 @@ class StatusCode(IntEnum):
     CLIENT_ERROR = 400
     AUTHORIZATION_ERROR = 401
     NOT_FOUND = 404
+    CONTENT_TOO_LARGE = 413
```
yinggeh, line 51: Good point. We should keep C++ http server consistent. Can you create a linear ticket for it?

yinggeh:
Should be `feat` not `fix`

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
```diff
@@ -0,0 +1,158 @@
...
+    handlers that never read the body. The buffered body is released the
+    moment the application consumes it, so the middleware contributes no
+    sustained memory overhead.
+    """
+
+    def __init__(self, app: ASGIApp, http_max_input_size: int) -> None:
+        self.app = app
+        self.http_max_input_size = validate_positive_int(http_max_input_size)
+        self._content_too_large_message = (
+            "Request content size exceeds the maximum allowed input size of "
```
yinggeh: Let's keep the message consistent with C++ server.

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
```diff
@@ -0,0 +1,158 @@
...
+            f"{self.http_max_input_size} bytes. "
+            "Use --http-max-input-size to increase the limit."
+        )
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
```
yinggeh, line 55: What's this?

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
```diff
@@ -0,0 +1,158 @@
...
+                    StatusCode.CONTENT_TOO_LARGE,
+                    "content_too_large",
+                    self._content_too_large_message,
+                )
+                return
+            body_chunks.append(chunk)
+            if not message.get("more_body", False):
+                break
+
+        # Assemble body (single-chunk fast path avoids a copy) and replay to app.
```
yinggeh: What's "single-chunk fast path"?

python/openai/tests/test_request_size.py
```diff
@@ -0,0 +1,161 @@
...
+
+sys.path.append(
+    os.path.join(str(Path(__file__).resolve().parent.parent), "openai_frontend")
+)
+
+from frontend.fastapi.middleware.request_size import RequestSizeLimitMiddleware
+from tests.utils import setup_fastapi_app, setup_server
+from utils.utils import HTTP_DEFAULT_MAX_INPUT_SIZE
+
+_LIMIT = HTTP_DEFAULT_MAX_INPUT_SIZE  # 64 MiB
```
yinggeh: Why introducing a new variable?

python/openai/tests/test_request_size.py
```diff
@@ -0,0 +1,161 @@
...
+_OVERFLOW = 16  # bytes
```
yinggeh: Is it possible to overflow 1 byte?

python/openai/tests/test_request_size.py
pskiran1: It is not needed, Thank you.
I updated the test to use `HTTP_DEFAULT_MAX_INPUT_SIZE`

python/openai/tests/test_request_size.py
pskiran1: Yes, I updated the test to try with 1 byte extra than the limit.

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
pskiran1: It was intended for cases where the user sends a single large chunk to avoid using .join(), but I don’t think it’s necessary. I have updated the code to remove it.

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
pskiran1: I updated the error message.

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
pskiran1: This is a pure `ASGI` middleware. Unlike `BaseHTTPMiddleware`, which buffers the entire request body into memory before any handler logic runs, a pure `ASGI` middleware intercepts the raw `receive()` callable directly, allowing this middleware to count body bytes as they stream in and reject oversized requests before any Python object is allocated.
`ASGI` apps receive three scope types: `http`, `lifespan`, and `websocket`. The `lifespan` scope carries Uvicorn's startup and shutdown events with different attributes. Without this passthrough, it may crash or hang when the middleware attempted  for the Content-Length check. The guard ensures only HTTP requests are subject to the size limit while all other scope types are forwarded to the application unchanged.

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
```diff
@@ -0,0 +1,160 @@
...
+                # the app processes it, instead of being held by this closure.
+                message, body_message = body_message, None
+                return message
+            # Body already delivered — delegate to the original receive() so
+            # streaming responses can wait for the real client disconnect.
+            return await receive()
+
+        await self.app(scope, replay_receive, send)
+
+    def _content_too_large_message(self, actual_bytes: int) -> str:
```
yinggeh:
```suggestion
    def _oversized_request_message(self, actual_bytes: int) -> str:
```

python/openai/tests/test_request_size.py
```diff
@@ -0,0 +1,163 @@
...
+
+def _assert_content_too_large(response) -> None:
+    assert response.status_code == 413
+    body = response.json()
+    assert set(body) == {"error"}
+    error = body["error"]
+    assert error["type"] == "invalid_request_error"
+    assert error["code"] == "content_too_large"
+    assert str(HTTP_DEFAULT_MAX_INPUT_SIZE) in error["message"]
+    assert "--http-max-input-size" in error["message"]
```
yinggeh: Can you use `_oversized_request_message` util function for assertion?

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
```diff
@@ -0,0 +1,160 @@
...
+        self.app = app
+        self.http_max_input_size = validate_positive_int(http_max_input_size)
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        # Stage 1: reject on Content-Length before reading any body bytes.
+        for name, value in scope["headers"]:
```
whoisj, line 58: `scope["headers"]` should return a `dict[str, list[str]]`

would it make more sense to
```python
  if "headers" in scope:
    scope_headers = scope["headers"]
    if "content-length" in scope_headers:
      for value in scope_headers["content-length"]:
```
instead of looping over all of the key-value pairs (which could be dozens) in headers?

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # Assemble the buffered body and replay it to the app.
+        body_message: Message = {
+            "type": "http.request",
+            "body": b"".join(body_chunks),
+            "more_body": False,
+        }
+        del body_chunks
+
+        async def replay_receive() -> Message:
```
whoisj, line 124: this is a static function that's not part of the containing class. 

why is structured as part of the class?

python/openai/openai_frontend/utils/utils.py
```diff
@@ -26,6 +26,10 @@
 
 from enum import IntEnum
 
+# Default maximum allowed HTTP request input size in bytes (64 MiB).
+# Same as HTTP_DEFAULT_MAX_INPUT_SIZE in src/common.h.
+HTTP_DEFAULT_MAX_INPUT_SIZE: int = 1 << 26
```
whoisj, line 31: 64MiB is really the maximum? (🔽 low priority concern 🔽)

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
pskiran1: Thank you for noting it. In ASGI, `scope["headers"]` is actually `Iterable[[bytes, bytes]]`, a list of (name, value) byte pairs, not a dict. For reference [ASGI HTTP spec](https://asgi.readthedocs.io/en/latest/specs/www.html#http-connection-scope).

python/openai/openai_frontend/utils/utils.py
pskiran1: Good catch. The old comment was ambiguous. 64 MiB is the default for the `--http-max-input-size` CLI flag, not a hard cap.

python/openai/openai_frontend/frontend/fastapi/middleware/request_size.py
pskiran1: It is a closure that captures two per-request locals from `__call__:` the buffered body_message (mutated via nonlocal) and the original ASGI receive callable. Since a single middleware instance is shared across all concurrent requests, storing this state on self would risk requests interfering with each other.

whoisj (approved):
LGTM, but I am not a FastAPI expert. I'll leave it to @pskiran1 and @yinggeh to decide if this is correct or not.

python/openai/openai_frontend/utils/utils.py
pskiran1: Created TRI-1312 for tracking this. Thank you.

yinggeh (approved):

PR #8794, opened 2026-05-20
fix: Replace std:atoi with std:stoi

whoisj:
Fixes an issue where exception handling was relied upon to detect errors when exceptions would not be generated, allowing errors to occur unnoticed.


mudit-eng (approved):
Title needs to be fixed: Replace 'std:atoi' with 'std:stoi'

yinggeh:
Since `atoi` doesn't throw, why we `try catch` it before?

yinggeh (approved):

PR #8800, opened 2026-05-23
test: Special handling for ONNX bf16 models in tests

yinggeh:
- Skip bf16 models in L0_infer (unsupported by ORT CPU EP and Python backend).
- Add ml_dtypes to L0_http_fuzz virtual_env pip install.


whoisj (approved):
LGTM

mc-nv (approved):

PR #8812, opened 2026-06-02
Remove Windows server build support

Vinya567:
## Summary

This removes the Windows-specific server build path and keeps the server build flow focused on the supported Unix-style targets.

The change updates `build.py` to reject Windows explicitly as a target platform, removes PowerShell script generation and Windows container build handling, and simplifies the generated build scripts to always use the bash path. It also removes Windows-specific CMake branches from the server build files, including MSVC compile options, Windows protobuf and OpenTelemetry path handling, Windows install renaming for `tritonserver.exe`, and `NOT WIN32` guards around tracing, examples, Python frontend packaging, and tests.

## Rationale

Windows is no longer a supported server build target, so keeping parallel Windows build branches makes the build scripts harder to maintain and creates paths that are no longer expected to work. Failing early in `build.py` makes the unsupported platform behavior explicit, while simplifying CMake keeps the remaining build configuration aligned with the supported Linux/RHEL/iGPU targets.

## Validation

- Ran `python3 -m py_compile build.py` to verify the Python build script still parses.
- Ran `git diff --check origin/main...HEAD` to verify the committed changes do not introduce whitespace errors.
- Searched the updated build entry points for remaining Windows build conditionals in `build.py`, `CMakeLists.txt`, `src/CMakeLists.txt`, and `src/grpc/CMakeLists.txt`.

### DIFF
--- a/build.py
+++ b/build.py
@@ -118,6 +118,8 @@ def fail_if(p, msg):
 def target_platform():
     # When called by compose.py, FLAGS will be None
     if FLAGS and FLAGS.target_platform is not None:
+        if FLAGS.target_platform == "windows":
+            fail("Windows is no longer a supported target platform.")
         return FLAGS.target_platform
     platform_string = platform.system().lower()
     if platform_string == "linux":
@@ -128,6 +130,8 @@ def target_platform():
             return "linux"
         else:
             return "rhel"
+    elif platform_string == "windows":
+        fail("Windows is no longer a supported target platform.")
     else:
         return platform_string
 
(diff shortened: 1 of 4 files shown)


build.py
```diff
@@ -2461,13 +2233,13 @@ def enable_all():
         "--container-memory",
         default=None,
         required=False,
-        help="Value for Docker --memory argument. Used only for windows builds.",
```
whoisj, line 2464: should this arg just be removed completely?

whoisj (commented):
Looks good. Had 1 question.

build.py
Vinya567: yes. --container-memory was only used by the Windows Docker build paths, so I removed the argument completely in the latest commit.

whoisj (approved):
LGTM

PR #8819, opened 2026-06-05
test: Fix flaky L0_http test_large_string_in_json

Vinya567:
#### What does the PR do?
Fixes a deterministic failure in `qa/L0_http/http_input_size_limit_test.py::test_large_string_in_json`.

The test built a ~2 GiB JSON body to exercise the server's `--http-max-input-size` guard, but that body produces a `Content-Length` header beyond `INT32_MAX`. After the recent `std::atoi` → `std::stoi` change in `Content-Length` parsing (#8794), the server now rejects the request at the header-parse step (with an INT32 range error) before ever reaching the JSON-size check the test asserts on, so the test fails on every nightly.

The fix shrinks the request body to byte-precise boundaries around the JSON-size limit, so the test actually exercises the guard it claims to.

#### Checklist
- [x] PR title reflects the change and is of format `<type>: <description>`

#### Commit Type:
- [x] test

#### Related PRs:

- #8794 — the upstream `std::atoi` → `std::stoi` change in `Content-Length` parsing that exposed the latent issue. Before this change, `atoi` silently returned a wrong-but-non-throwing value on overflow, so the test happened to still reach the JSON-size check. After, `stoi` throws and the server returns a different error than the test expects.
- #8432 — the original PR that introduced `test_large_string_in_json`. /cc @spolisetty for design intent on the original 2 GiB sizing.

`qa/L0_http/http_input_size_limit_test.py` — the change is contained to one file. Look at the new `_build_payload_of_json_size` helper and the rewritten `test_large_string_in_json`.

#### Test plan:

Verified locally in a `triton-devel` container against `main`:

- `InferSizeLimitTest.test_large_string_in_json` — 10/10 pass
- Full `qa/L0_http/test.sh` — passes, no other regressions

The new test asserts both sides of the boundary:
- Body of exactly `DEFAULT_LIMIT_BYTES` (64 MiB) → must succeed (200)
- Body of `DEFAULT_LIMIT_BYTES + 1` → must fail with the JSON-size error

Body length is asserted byte-exact in the helper via `assertEqual(len(json.dumps(payload)), target_size)` before being posted with `requests.post(json=payload)`, so the boundary is byte-precise and not skewed by JSON wrapper overhead.

- CI Pipeline ID: 53762086 — targeted run via `RULE_WORKFLOW` + `EXTRA_BUILD_BACKENDS=python onnxruntime` per reviewer tips. L0_http--base job retried for commit `f9b4e0b`: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/334699597 — passed in 8 min 29 sec, no rebuild needed for test-only change.

`test_chunked_infer_over_max_chunks_reject_with_bounded_rss_growth` (in `qa/L0_http/http_request_many_chunks.py`) also occasionally fails in nightly runs with intermittent RSS-growth assertions (`4.2 MiB > 1 MiB`). That failure is intermittent, unrelated to this fix, and should be tracked separately.

#### Background

`test_large_string_in_json` was added to verify the server's `--http-max-input-size` JSON-body limit. The original test built a body large enough to clearly exceed the default 64 MiB limit, but also large enough that the resulting `Content-Length` header overflows a signed 32-bit integer (`Content-Length: 2147483796` vs `INT32_MAX = 2147483647`).

Before #8794, `Content-Length` was parsed with `std::atoi`, which silently returned an undefined-but-non-throwing value on overflow, so the body still reached the JSON-size check. After #8794 switched to `std::stoi`, that parse now throws `std::out_of_range` and the server returns 400 with:

```
Unable to parse Content-Length, value is out of range [ -2147483648, 2147483647 ], got: 2147483796
```

which doesn't match the JSON-size error the test expects. The fix keeps the test focused on the JSON-size guard by constructing a body just at and just above the limit (well below INT32 Content-Length).

#### Related Issues: 
- Relates to #8794
- Relates to #8432

### DIFF
--- a/qa/L0_http/http_input_size_limit_test.py
+++ b/qa/L0_http/http_input_size_limit_test.py
@@ -499,8 +499,8 @@ class InferSizeLimitTest(TestResultCollector):
             f"Expected shape {[1, shape_size]}, got {result['outputs'][0]['shape']}",
         )
 
-    def _build_json_string_body_of_size(self, target_size):
-        """Build a JSON inference request body that is exactly target_size bytes."""
+    def _build_payload_of_json_size(self, target_size):
+        """Build an inference request payload whose JSON serialization is exactly target_size bytes."""
         payload = {
             "inputs": [
                 {
@@ -513,21 +513,17 @@ class InferSizeLimitTest(TestResultCollector):
         }
         overhead = len(json.dumps(payload))
         pad = target_size - overhead
-        assert pad >= 0, "target_size smaller than payload overhead"
+        self.assertGreaterEqual(pad, 0, "target_size smaller than payload overhead")
         payload["inputs"][0]["data"] = ["A" * pad]
-        body = json.dumps(payload)
-        assert len(body) == target_size, (len(body), target_size)
-        return body
+        self.assertEqual(len(json.dumps(payload)), target_size)
+        return payload
 
     def test_large_string_in_json(self):
         """Verify the server's JSON size limit at the byte boundary."""
         model = "simple_identity"
-        headers = {"Content-Type": "application/json"}
 
-        body_over = self._build_json_string_body_of_size(DEFAULT_LIMIT_BYTES + 1)
-        response = requests.post(
-            self._get_infer_url(model), data=body_over, headers=headers
-        )
+        payload_over = self._build_payload_of_json_size(DEFAULT_LIMIT_BYTES + 1)
+        response = requests.post(self._get_infer_url(model), json=payload_over)
         self.assertEqual(
             400,
             response.status_code,
@@ -540,10 +536,8 @@ class InferSizeLimitTest(TestResultCollector):
         self.assertIn(" bytes exceeds the maximum allowed input size of ", error_msg)
         self.assertIn("Use --http-max-input-size to increase the limit.", error_msg)
 
-        body_at = self._build_json_string_body_of_size(DEFAULT_LIMIT_BYTES)
-        response = requests.post(
-            self._get_infer_url(model), data=body_at, headers=headers
-        )
+        payload_at = self._build_payload_of_json_size(DEFAULT_LIMIT_BYTES)
+        response = requests.post(self._get_infer_url(model), json=payload_at)
         self.assertEqual(
             200,
             response.status_code,
--- a/qa/L0_http/http_input_size_limit_test.py
+++ b/qa/L0_http/http_input_size_limit_test.py
@@ -499,54 +499,65 @@ class InferSizeLimitTest(TestResultCollector):
             f"Expected shape {[1, shape_size]}, got {result['outputs'][0]['shape']}",
         )
 
-    def test_large_string_in_json(self):
-        """Test JSON request with large string input"""
-        model = "simple_identity"
-
-        # Create a string larger than the default 64MB limit but well below INT32
-        # Content-Length max (~2GB) so the server validates JSON size, not header
-        # parsing.
-        large_string_size = DEFAULT_LIMIT_BYTES + OFFSET_ELEMENTS
-        large_string = "A" * large_string_size
-
+    def _build_string_body_of_size(self, target_size):
+        """Build a JSON inference request body that is exactly target_size bytes."""
         payload = {
             "inputs": [
                 {
                     "name": "INPUT0",
                     "datatype": "BYTES",
                     "shape": [1, 1],
-                    "data": [large_string],
+                    "data": [""],
                 }
             ]
         }
+        overhead = len(json.dumps(payload))
+        pad = target_size - overhead
+        assert pad >= 0, "target_size smaller than payload overhead"
+        payload["inputs"][0]["data"] = ["A" * pad]
+        body = json.dumps(payload)
+        assert len(body) == target_size, (len(body), target_size)
+        return body
 
+    def test_large_string_in_json(self):
+        """Verify the server's JSON size limit at the byte boundary.
+
+        Uses a body well below the INT32 Content-Length max (~2GB) so the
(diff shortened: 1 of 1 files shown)


qa/L0_http/http_input_size_limit_test.py
```diff
@@ -499,53 +499,65 @@ def test_large_input_json(self):
...
+        assert len(body) == target_size, (len(body), target_size)
+        return body
 
+    def test_large_string_in_json(self):
+        """Verify the server's JSON size limit at the byte boundary.
+
+        Uses a body well below the INT32 Content-Length max (~2GB) so the
+        server's JSON-size check fires deterministically rather than failing
+        in Content-Length header parsing.
+        """
```
yinggeh:
```suggestion
        """Verify the server's JSON size limit at the byte boundary."""
```

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -499,53 +499,65 @@ def test_large_input_json(self):
...
-    def test_large_string_in_json(self):
-        """Test JSON request with large string input"""
-        model = "simple_identity"
-
-        # Create a string that is larger (large payload about 2GB) than the default limit of 64MB
-        # (2^31 + 64) elements * 1 bytes = 2GB + 64 bytes = 2,147,483,712 bytes
-        large_string_size = 2 * GIB + 64
-        large_string = "A" * large_string_size
-
+    def _build_string_body_of_size(self, target_size):
```
yinggeh:
```suggestion
    def _build_json_string_body_of_size(self, target_size):
```

qa/L0_http/http_input_size_limit_test.py
yinggeh: The comment is PR specific. Not related to the test itself.

yinggeh:
If you are using agent generating PR description, make sure tell it to use https://github.com/triton-inference-server/server/blob/main/.github/PULL_REQUEST_TEMPLATE/pull_request_template_internal_contrib.md as the template.

yinggeh:
Again, do not put linear tickets in public PRs.

qa/L0_http/http_input_size_limit_test.py
Vinya567: Done in `6a74ad1`.

qa/L0_http/http_input_size_limit_test.py
Vinya567: Done in `6a74ad1`.

Vinya567:
Both addressed:
- PR description rewritten to use `pull_request_template_internal_contrib.md` all template sections filled out.
- Scrubbed all Linear / internal references (description, title, etc.).

Sorry for the slip ,added both rules to my onboarding checklist so they don't happen again. Thanks for the catches!

whoisj (dismissed):

pskiran1 (dismissed):
LGTM

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -499,53 +499,57 @@ def test_large_input_json(self):
...
-        self.assertIn(
-            " bytes exceeds the maximum allowed input size of ",
-            error_msg,
+        self.assertIn("request JSON size of ", error_msg)
+        self.assertIn(" bytes exceeds the maximum allowed input size of ", error_msg)
+        self.assertIn("Use --http-max-input-size to increase the limit.", error_msg)
+
+        body_at = self._build_json_string_body_of_size(DEFAULT_LIMIT_BYTES)
+        response = requests.post(
+            self._get_infer_url(model), data=body_at, headers=headers
```
yinggeh: Can be more explicit by passing json object to this method. `data` argument accepts a dictionary, list of tuples, bytes or a file object.
```
        response = requests.post(
            self._get_infer_url(model), headers=headers, json=payload
        )
```

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -499,53 +499,57 @@ def test_large_input_json(self):
...
                     "datatype": "BYTES",
                     "shape": [1, 1],
-                    "data": [large_string],
+                    "data": [""],
                 }
             ]
         }
+        overhead = len(json.dumps(payload))
+        pad = target_size - overhead
+        assert pad >= 0, "target_size smaller than payload overhead"
```
yinggeh: Use `unittest.TestCase` assertions, e.g. `self.assertGreaterEqual`

yinggeh (approved):

yinggeh (commented):
Left some comments

qa/L0_http/http_input_size_limit_test.py
```diff
@@ -499,53 +499,57 @@ def test_large_input_json(self):
...
+        payload["inputs"][0]["data"] = ["A" * pad]
+        body = json.dumps(payload)
+        assert len(body) == target_size, (len(body), target_size)
```
yinggeh: Same here

yinggeh (changes requested):

qa/L0_http/http_input_size_limit_test.py
Vinya567: Done in f9b4e0b.

qa/L0_http/http_input_size_limit_test.py
Vinya567: Done in f9b4e0b.

qa/L0_http/http_input_size_limit_test.py
Vinya567: Addressed in review comments in f9b4e0b.

yinggeh (approved):
LGTM

PR #8827, opened 2026-06-10
test: fix: L0_server_status

whoisj:
Fixes the L0_server_status test by using the correct expected value for `model_stats` count.

### DIFF
--- a/qa/L0_server_status/server_status_test.py
+++ b/qa/L0_server_status/server_status_test.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-# Copyright 2018-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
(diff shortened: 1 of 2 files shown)


qa/L0_server_status/server_status_test.py
```diff
@@ -428,6 +428,9 @@ def test_model_specific_infer(self):
                 self.assertTrue(False, "unexpected error {}".format(ex))
 
 
+EXPECTED_MODEL_STATS: int = 144
```
yinggeh, line 431: Did you figure out what was changed?

whoisj: yeah, the number of stats reported.

there's another condition that I was hitting that happens after running `./test.sh` multiple times where the model repository gets corrupted. I spent time chasing that problem when it wasn't the real problem.

yinggeh: > yeah, the number of stats reported.

Was it because I genereated 2 BF16 ONNX models https://github.com/triton-inference-server/server/pull/8782/changes#diff-b0b691449b31f9b36929a5f1dd1edc000ea86264a4078d75318de9cbfb23cbc0?

whoisj: No, it's because the test isn't idempotent and requires manual clean up of `L0_server_status` folder before running the test again.

yinggeh (approved):

Vinya567 (approved):

PR #8850, opened 2026-06-24
test: Move initializer_as_input ONNX test into qa/ and drop cpdir

Vinya567:
#### What does the PR do?
Moves the `initializer_as_input` ONNX QA test from `onnxruntime_backend/test/` into `qa/L0_initializer_as_input/` so the test lives in the same repo where it's consumed by the QA harness (and follows the existing `L0_` tier naming convention under `qa/`). Drops the now-redundant `cpdir(onnxruntime_backend/test/*, qa/)` line from `cibase_build()` in `build.py` — the files are already part of the `qa/` checkout, so the cross-repo wildcard copy is unnecessary.

This addresses the `[WIP] other way than wildcard?` comment that was left when the cross-repo `cpdir` was originally introduced.

#### Checklist

#### Commit Type:
- [x] test

#### Related PRs:
- Follow-up PR in `triton-inference-server/onnxruntime_backend` to remove the now-duplicated `test/initializer_as_input/` directory (will be opened after this PR merges so the source-of-truth in `qa/` is live first).

1. `build.py` — the 3-line deletion in `cibase_build()` is the actual behavioral change. Confirm nothing else in that function depends on `qa/` being repopulated from the backend tree.
2. `qa/L0_initializer_as_input/` — the migrated test. Files are byte-identical to the originals in `onnxruntime_backend/test/initializer_as_input/`; no logic changes. Diff is effectively a cross-repo directory move + `L0_` prefix to match the `qa/` naming convention.

#### Test plan:
1. **Local run** of `test.sh` inside `nvcr.io/nvidia/tritonserver:26.04-py3` with the new `qa/L0_initializer_as_input/` mounted as the working directory:
   - `tritonserver` starts, loads `add_with_initializer`
   - `test.py` issues an inference request and verifies output
   - Result: `PASS` (exit code 0)
2. **CI run** with the cross-repo `cpdir` removed and the test sourced from `qa/`:
   - `py3-devel` builds successfully (proves `build.py` change does not break the dev image build)
   - `L0_initializer_as_input--ONNXRuntime` passes, pulling files from the `qa/` checkout (no cross-repo wildcard)
   - All other jobs in the gating set are green

- CI Pipeline ID: 55844627

#### Caveats:
- NVIDIA's downstream RHEL build currently copies `onnxruntime_backend/test/` into `qa/` at image-build time (added recently as a hotfix to unblock the RHEL pipeline). That copy becomes a no-op once this PR merges, since `qa/L0_initializer_as_input/` is already in the build context. A follow-up cleanup will remove it from the downstream Dockerfile.
- The old `onnxruntime_backend/test/initializer_as_input/` directory is intentionally kept in place until this PR lands, so the existing cross-repo flow stays functional during review.

#### Background
The `cpdir` line being removed was originally added in #5630 with a `[WIP] other way than wildcard?` comment, flagging that the cross-repo wildcard copy was a temporary stopgap. Since `qa/` is the canonical location for QA tests consumed by Triton server's QA harness, the cleanest fix is to land the test where it's consumed and drop the wildcard.

#### Related Issues:
N/A — no public github issue tracking this cleanup.


qa/L0_initializer_as_input/models/add_with_initializer/1/model.onnx
mc-nv, line 1: Would be nice to have a consistency in naming  and preserve prefix`L0_`

qa/L0_initializer_as_input/test.sh
```diff
@@ -0,0 +1,67 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+export CUDA_VISIBLE_DEVICES=0
```
whoisj, line 28: @mc-nv @yinggeh @Vinya567 does anyone know why we do this in the test.sh file?

This makes it very difficult to control which GPUs are assigned to which tests from outside of the test itself.

I know that this line is here because it's nearly every other test. My question is how did this get started and is it something we should continue doing?

qa/L0_initializer_as_input/models/add_with_initializer/config.pbtxt
```diff
@@ -0,0 +1,48 @@
...
+    optional: true
+  }
+]
+output [
+  {
+    name: "OUTPUT"
+    data_type: TYPE_FP32
+    dims: [ 5, 5]
+  }
+]
```
whoisj: missing end-of-file new line here.

qa/L0_initializer_as_input/generate_test_model.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
whoisj, line 1: if this is a new file, why is the copyright 2023-2026?

qa/L0_initializer_as_input/models/add_with_initializer/config.pbtxt
```diff
@@ -0,0 +1,48 @@
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
whoisj, line 1: if this is a new file, why is the copyright 2023-2026?

qa/L0_initializer_as_input/README.md
```diff
@@ -0,0 +1,33 @@
+<!--
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
whoisj, line 2: if this is a new file, why is the copyright 2023-2026?

qa/L0_initializer_as_input/test.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
whoisj, line 2: if this is a new file, why is the copyright 2023-2026?

qa/L0_initializer_as_input/test.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
whoisj, line 2: if this is a new file, why is the copyright 2023-2026?

whoisj (changes requested):
LGTM, concerned about the copyright on each file. Please address (explain or correct).

qa/L0_initializer_as_input/models/add_with_initializer/1/model.onnx
Vinya567: Renamed to qa/L0_initializer_as_input/ to match the L0_ convention used by the other 137 test dirs in qa/. New SHA d0de15451 ! diff is a pure directory move + tier prefix, no file content changes. Thanks for catching this!

qa/L0_initializer_as_input/models/add_with_initializer/config.pbtxt
Vinya567: fixed in d0de15451!Thanks.

qa/L0_initializer_as_input/test.sh
Vinya567: Yeah, I copied this file as-is from onnxruntime_backend/test/initializer_as_input/test.sh ! didn't change any of the test logic, just moved it.

You're right that this pattern is everywhere in qa/L0_* tests. Worth digging into the history, but probably as a separate cleanup since it's not specific to this test. I can take a look once we figure out the right convention ! want to open an issue to track it, or kick off a thread elsewhere?

qa/L0_initializer_as_input/models/add_with_initializer/config.pbtxt
Vinya567: These aren't actually new , they're moves from onnxruntime_backend/test/initializer_as_input/ where they've existed since 2023. So the 2023 in the header is the original authorship year, and 2026 is when I moved them here.

Worth noting: the pre-commit add-license hook produced this 2023-2026 automatically when I amended ,it picks up the original year and extends to today. The tool itself follows the same pattern (its own header at tools/add_copyright.py line 1 reads "Copyright (c) 2024-2026").

That said , if the team's convention is "new path = new file = year resets to today", I can change all 5 to just 2026. Which way do we want to go, @mc-nv @yinggeh?

yinggeh: Fine for me since the content wasn't changed.

qa/L0_initializer_as_input/test.sh
yinggeh: My guess is this hard-coded value implies this test must use single-GPU.

Vinya567: yeah that's my read too ! single-GPU isolation. probably worth a separate cleanup pass 
across all qa/L0_* tests though, not just this one.

Vinya567:
> LGTM, concerned about the copyright on each file. Please address (explain or correct).

These aren't actually new , they're moves from onnxruntime_backend/test/initializer_as_input/ where they've existed since 2023. So the 2023 in the header is the original authorship year, and 2026 is when I moved them here.

Worth noting: the pre-commit add-license hook produced this 2023-2026 automatically when I amended ,it picks up the original year and extends to today. The tool itself follows the same pattern (its own header at tools/add_copyright.py line 1 reads "Copyright (c) 2024-2026").

build.py
```diff
@@ -2094,9 +2094,6 @@ def cibase_build(
                 os.path.join(ort_install_dir, "test", "custom_op_test.onnx"),
                 os.path.join(ci_dir, "qa", "L0_custom_ops"),
             )
-        # [WIP] other way than wildcard?
-        backend_tests = os.path.join(build_dir, "onnxruntime", "test", "*")
-        cmake_script.cpdir(backend_tests, os.path.join(ci_dir, "qa"))
```
mc-nv, line 2099: Can yo help me find out what was the motivation on this change?

Vinya567: yeah good q! those lines pulled the onnx qa tests from the backend repo 
(onnxruntime_backend/test/) into qa/ at build time, since that's where they 
used to live.

now that the test sits directly in qa/L0_initializer_as_input/, the cpdir 
isn't needed , and honestly the # [WIP] other way than wildcard? comment 
above it kinda hinted it was meant to be temporary anyway.

followup pr to onnxruntime_backend will delete the now-duplicate source 
files once this one merges.

qa/L0_initializer_as_input/test.sh
mc-nv: Majority of our tests are legacy ones. 
Usage of the `CUDA_VISIBLE_DEVICES` looks like kind of additional control for the test, which is can be redundant.

The one of the reasons I would see it had been added is that eventually there was CI runners with multiple GPU devices of different type in order to get proper enumeration or limit number of GPU's that may serve the request.

Vinya567: thanks for the context misha! yeah multi-gpu legacy era makes sense. since 
this pr is just a move (no logic change), leaving the line as-is for now 
to keep the diff clean. happy to file a separate ticket for the broader 
cuda_visible_devices cleanup across qa/L0_* if you want to tackle it.

whoisj (approved):

yinggeh (approved):

PR #8886, opened 2026-07-15
feat(build): experimental build presets (--build-presets-file)

mc-nv:
#### What does the PR do?
Adds an experimental, opt-in **build presets** mechanism to `build.py`, gated behind `TRITON_BUILD_EXPERIMENTAL=1` (off by default — no behavioral change when unused):

- **`--build-presets-file <json>`** pins per-component build settings — git tag, clone organization, library path, and extra/override CMake args — for `core` / `backends` / `repoagents` / `caches`. Command-line flags take precedence over the file. A per-backend `org` sets **both** the `git clone` URL and `-DTRITON_REPO_ORGANIZATION`, which the global `--github-organization` cannot do per backend.
- On **`--dryrun`**, `build.py` writes `build_presets.json` — a provenance-annotated snapshot of every resolved cmake `-D` flag each component receives (matching `cmake_build`), each labeled with its `source`: `cli` / `preset` / `default`. That file reloads via `--build-presets-file` to reproduce the build.

All load/validate/precedence/dump logic lives in `tools/build/build_presets.py`; the `build.py` integration is intentionally minimal (an argparse flag, a guarded apply block, one line in the backend build loop, and the `--dryrun` snapshot block).

Also included:
- **`qa/L2_build_presets`** — a self-contained pytest suite validating the documented scenarios via `build.py --dryrun` (no GPU / container / real build). Runs in-tree, via `TRITON_BUILD_PY`, or self-clones in a bare container; `test.sh` installs deps and writes a log + JUnit report.
- **Incremental SPDX license-header adoption** — a migrate-on-touch pre-commit hook (`tools/add_spdx_header.py`) that converts each *changed* source file's legacy NVIDIA BSD header to the two-line SPDX form (maintaining / replacing-in-place / inserting), coexisting with `add_copyright.py`.

#### Checklist
- [ ] Added succinct git squash message before merging.

#### Commit Type:
- [x] feat

#### Related PRs:
None — single repo (`server`).

- `build.py` — the integration points: the `--build-presets-file` arg, the guarded apply block, the per-backend clone-org line in the backend build loop, and the `--dryrun` snapshot block.
- `tools/build/build_presets.py` — all preset load / validate / precedence / dump logic.
- `qa/L2_build_presets/build_presets_test.py` — the scenarios covered.
- `tools/add_spdx_header.py` + `.pre-commit-config.yaml` — the SPDX migrate-on-touch hook.

#### Test plan:
- `cd qa/L2_build_presets && bash test.sh` → 11 pytest scenarios pass: snapshot sections, per-flag source annotation, `CMAKE_INSTALL_PREFIX` exclusion, provenance (cli vs default), extra-channel routing, env gate, example apply, round-trip reload, and validation errors.
- Regression: the documented CI `build.py` invocations produce byte-identical `cmake_build` with the feature unused (verified via `--dryrun` diff).

- CI Pipeline ID:

#### Caveats:
- Experimental and off by default (gated behind `TRITON_BUILD_EXPERIMENTAL=1`).
- Reload pins cmake `-D` values but does not re-derive conditionally-emitted flags; reload alongside the same top-level flags (or `--enable-all`) for exact reproduction.
- Repoagent/cache `cmake_args` are snapshot-only (not re-pinned on load); their `tag` is applied.
- The SPDX migrate-on-touch hook changes pre-commit behavior repo-wide (files migrate as they are touched); it can be split into its own PR if preferred.

#### Background
Follow-up to the TRI-829 `build.py` CLI usability work: lets users declaratively pin per-backend build settings (including a per-backend clone org) and inspect exactly which cmake flags each component receives.

#### Related Issues:
- Relates to TRI-829 (Linear)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,28 +1,5 @@
-# Copyright 2023-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-#
-# Redistribution and use in source and binary forms, with or without
-# modification, are permitted provided that the following conditions
-# are met:
-#  * Redistributions of source code must retain the above copyright
-#    notice, this list of conditions and the following disclaimer.
-#  * Redistributions in binary form must reproduce the above copyright
-#    notice, this list of conditions and the following disclaimer in the
-#    documentation and/or other materials provided with the distribution.
-#  * Neither the name of NVIDIA CORPORATION nor the names of its
-#    contributors may be used to endorse or promote products derived
-#    from this software without specific prior written permission.
-#
-# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS ``AS IS'' AND ANY
-# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
-# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
-# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
-# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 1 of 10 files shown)


whoisj (approved):
LGTM. @yinggeh will this conflict with your recent change?

PR #8888, opened 2026-07-17
build(deps): Align protobuf and related dependencies with the gRPC v1.81.1 bump

mc-nv:
#### What does the PR do?
Fixes the nightly server build against protobuf v33 (gRPC v1.81.1 third_party bump):
- Compute `_FINDPACKAGE_UTF8_RANGE_CONFIG_DIR` and pass `utf8_range_DIR` into the triton-server ExternalProject cache args — protobuf v33's CMake package config now requires the `utf8_range` package config to resolve.
- Replace the removed `google::protobuf::stringpiece_internal::StringPiece` with `absl::string_view` in the gRPC model-config handler, and surface the previously ignored `JsonStringToMessage()` status as an INTERNAL error. (The license header of `grpc_server.cc` was converted to the SPDX form by the repo pre-commit hooks.)

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Related PRs:
- triton-inference-server/client#908
- triton-inference-server/common#160
- triton-inference-server/core#516
- triton-inference-server/perf_analyzer#466
- triton-inference-server/third_party#79

#### Where should the reviewer start?
- `CMakeLists.txt` — `_FINDPACKAGE_UTF8_RANGE_CONFIG_DIR` / `utf8_range_DIR` cache arg

#### Test plan:
Nightly build pipeline on internal GitLab CI.

- CI Pipeline ID: 58489960

#### Caveats:
None.

#### Background
The gRPC v1.81.1 / protobuf v33 update in triton-inference-server/third_party#76 broke the nightly build across the Triton repos; this PR chain repairs it.

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
- Resolves: TRI-1608

<sup>
CI (internal): [#58489960](http://tritonserver.local/ci/pipelines/58489960)<br/>
</sup>

PR #8889, opened 2026-07-17
test: fix L0_lifecycle test_shutdown_dynamic exit timeout

Vinya567:
#### What does the PR do?

Fix `L0_lifecycle` failures on Luna where `LifeCycleTest.test_shutdown_dynamic` passes pytest but the QA script aborts during server cleanup.

#### Root cause

- `test_shutdown_dynamic` starts the server without `--exit-timeout-secs`, so Triton uses the default 30s shutdown window.
- The test sends 6 async inferences with `execute_delay_ms=5000`, then SIGINT while they are in flight; draining can exceed 30s on Luna (~80% failure rate).
- The server logs `Exit timeout expired` and exits non-zero; `wait $SERVER_PID` under `set -e` aborts the rest of `L0_lifecycle` (silent failure in the main log).

#### Fix

- `qa/L0_lifecycle/test.sh`:
  - Set `--exit-timeout-secs=${SERVER_TIMEOUT}` (120s) for the `test_shutdown_dynamic` server launch so draining has room to complete.
  - Use `wait $SERVER_PID || true` after cleanup so a non-zero server exit does not abort the suite.

Previous revisions of this branch also touched `qa/L0_lifecycle/lifecycle_test.py` (gRPC error-string assertions and pre-existing flake8 lints). Those changes are redundant with #8888, which landed on main after this PR was opened, so this PR now carries only the `test.sh` fix.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:

- #8888 (aligned gRPC error-string assertions in `lifecycle_test.py` for gRPC v1.81.1)

- `qa/L0_lifecycle/test.sh` — `test_shutdown_dynamic` server args (`--exit-timeout-secs`) and cleanup `wait`.

#### Test plan:

- GitLab Luna: `L0_lifecycle--base` green at commit `721e9c07` (job [368596258](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/368596258), pipeline [58809167](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/pipelines/58809167)); log shows `test_shutdown_dynamic` ran and `*** Test Passed ***`.
- Earlier Luna runs on same pipeline: `d41ea55` failed on `(111)` assertion mismatch (job [368365178](https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/368365178)); `8ba9e9fb` and `38c233da` also green.
- CI used `TRITON_SERVER_BRANCH_NAME=vinyak/nvbug-6450088-l0-lifecycle-shutdown-timeout` and `TRITON_THIRD_PARTY_REPO_TAG=4d8c8cec` to work around a transient `master` py3 build break (gRPC patch); that pin is CI-only, not part of this PR.
- CI Pipeline ID: 58809167

#### Caveats:

- Targeted Luna pipeline pinned `third_party` for CI verification only; `main` py3 builds were failing independently (gRPC `CMakeLists.txt` patch mismatch).
- After #8888 landed on main, this PR was rebased and force-pushed; the earlier Luna verification runs remain valid because the `test.sh` fix is unchanged.

#### Background

- Linear TRI-1580 / NVBUG 6450088 / DLIS-8700
- Same shutdown-timeout family as Jetson lifecycle work (TRI-1330), but a different test (`L0_lifecycle`) and platform (Luna x86).

- Relates to NVBUG 6450088

### DIFF
--- a/qa/L0_lifecycle/test.sh
+++ b/qa/L0_lifecycle/test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2018-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2018-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions
@@ -1563,7 +1563,7 @@ cp -r ../custom_models/custom_zero_1_float32 models/. && \
         echo "{ key: \"execute_delay_ms\"; value: { string_value: \"5000\" }}" >> config.pbtxt && \
         echo "]" >> config.pbtxt)
 
-SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=1"
+SERVER_ARGS="--model-repository=`pwd`/models --log-verbose=1 --exit-timeout-secs=${SERVER_TIMEOUT}"
 SERVER_LOG="./inference_server_$LOG_IDX.log"
 run_server
 if [ "$SERVER_PID" == "0" ]; then
@@ -1585,7 +1585,7 @@ if [ `grep -c "Found 1 gRPC service connections and inference handlers" $SERVER_
 fi
 
 kill $SERVER_PID || true
-wait $SERVER_PID
+wait $SERVER_PID || true
 
 rm -f $CLIENT_LOG
 


qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2629,6 +2629,32 @@ def test_file_override_security(self):
...
                 self.assertNotEqual(contents, new_contents)
 
+    _GRPC_CONNECTION_REFUSED_PREFIX = (
+        "failed to connect to all addresses; last error: UNKNOWN: "
+        "ipv4:127.0.0.1:8001: Failed to connect to remote host:"
+    )
+    _GRPC_CONNECTION_REFUSED_SUFFIXES = (
+        "Connection refused",
+        "connect: Connection refused",
+        "Connection refused (111)",
```
yinggeh: Isn't `Connection refused (111)` the only occurance of the suffix?

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2629,6 +2629,32 @@ def test_file_override_security(self):
...
+    def _assert_grpc_connection_refused(self, ex):
+        msg = ex.message()
+        self.assertTrue(
+            msg.startswith(self._GRPC_CONNECTION_REFUSED_PREFIX),
+            msg,
+        )
+        suffix = msg[len(self._GRPC_CONNECTION_REFUSED_PREFIX) :].strip()
+        self.assertIn(suffix, self._GRPC_CONNECTION_REFUSED_SUFFIXES, msg)
+
+    def _assert_shutdown_rejects_new_request(self, ex):
```
yinggeh: Since it's only called once, I'm not sure this function is necessary.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2732,29 +2754,21 @@ def callback(user_data, result, error):
...
+            # The first request received by the gRPC endpoint while shutting down may return
+            # CANCELLED or Connection refused depending on shutdown timing.
+            self._assert_shutdown_rejects_new_request(ex)
         # 2: New sequence with existing sequence ID
         try:
             triton_client.infer(model_name, inputs, sequence_id=1, sequence_start=True)
             self.assertTrue(False, "expected error for new inference during shutdown")
         except InferenceServerException as ex:
-            self.assertIn(
-                "failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:8001: "
```
yinggeh, line 2752: I think simply creating and reusing a string variable for the error message is better than creating helper functions which over-complicates the tests.

qa/L0_lifecycle/lifecycle_test.py
Vinya567: You're right — on `main` the expected message ends with `connect: Connection refused (111)`, and that is still the primary string (`_GRPC_CONNECTION_REFUSED_ERROR`).

The only reason for a second constant is what we saw on Luna CI with commit `d41ea55` (only the `test.sh` timeout fix, no `lifecycle_test.py` changes). `test_shutdown_dynamic` failed on the assertion because gRPC returned the same shutdown rejection without the errno suffix:

`... Failed to connect to remote host: Connection refused`

Same behavior (new request rejected while shutting down), different client wording. I dropped the suffix list from the earlier commit and kept one alternate full string for that Luna case.

If you prefer to keep only the original `(111)` string, I'm happy to remove `_GRPC_CONNECTION_REFUSED_ERROR_NO_ERRNO` and re-run `L0_lifecycle--base` on Luna to confirm.

qa/L0_lifecycle/lifecycle_test.py
Vinya567: Agreed , it was only used in one place, so the helper wasn't worth it. Removed in `721e9c07`; `test_shutdown_sequence` now inlines the check (allow `CANCELLED`, otherwise expect the connection-refused error string), same as the other call sites.

qa/L0_lifecycle/lifecycle_test.py
Vinya567: Agreed , updated in `721e9c07`:

- Removed the helper functions
- Added `_GRPC_CONNECTION_REFUSED_ERROR` (same full string as `main`) and reused it at each call site
- Kept one alternate string only for the Luna CI variant discussed above

Each site now does a simple inline check against those constants instead of a shared helper.

qa/L0_lifecycle/lifecycle_test.py
yinggeh: I think it's okay to only assertin the error prefix as the suffix is subject to change based on your observation.

Vinya567: Good point , that makes sense. Luna showed the suffix can vary, so checking the prefix is enough to prove "can't connect during shutdown."

Re-ran Luna on `721e9c07` and `L0_lifecycle--base` passed (job 368596258). Happy to simplify to prefix-only if you'd like.

yinggeh: Which GPU is Luna?

Vinya567: Luna is 8× A100 40GB (x86, Ubuntu 24.04) , that's the platform where we saw the suffix difference in CI.

yinggeh: > I think it's okay to only assertin the error prefix as the suffix is subject to change based on your observation.

Still waiting for this change.

Vinya567: Done ,switched to prefix-only checks with `_GRPC_CONNECTION_REFUSED_PREFIX` and removed the full-string suffix variants.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2668,11 +2650,8 @@ def callback(user_data, result, error):
             triton_client.infer(model_name, inputs)
             self.assertTrue(False, "expected error for new inference during shutdown")
         except InferenceServerException as ex:
-            self.assertIn(
-                "failed to connect to all addresses; last error: UNKNOWN: ipv4:127.0.0.1:8001: "
-                + "Failed to connect to remote host: connect: Connection refused (111)",
-                ex.message(),
-            )
+            msg = ex.message()
+            self.assertTrue(msg.startswith(self._GRPC_CONNECTION_REFUSED_PREFIX), msg)
```
yinggeh: Consider using `self.assertStartsWith`

Vinya567: Good call,updated to `self.assertStartsWith` in `c9bedf42`.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -1,7 +1,32 @@
...
+# PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# flake8: noqa
```
yinggeh: Why do we need this?

Vinya567: Pre-commit #8897 fixed flake8 arg parsing, and it now enforces rules that were silently disabled before. That surfaced ~30 pre-existing lints in this file (unused vars, imports after sys.path.append, etc.) ,none from this fix. Added # flake8: noqa to unblock CI without turning this PR into a lint cleanup. Happy to drop it and clean them up in a separate PR if you'd rather.

qa/L0_lifecycle/lifecycle_test.py
```diff
@@ -2715,23 +2740,21 @@ def callback(user_data, result, error):
             # each subsequent request returns Connection refused
             msg = ex.message()
             if "CANCELLED" not in msg:
-                self.assertTrue(
-                    msg.startswith(self._GRPC_CONNECTION_REFUSED_PREFIX), msg
-                )
+                self.assertStartsWith(self._GRPC_CONNECTION_REFUSED_PREFIX, msg)
```
yinggeh: I just realized `self.assertStartsWith` is available in 3.14. Our test environment might not be supporting it.

Vinya567: So I'll revert back to self.assertTrue(msg.startswith(...)) - okay?

yinggeh: Or `assertIn`, up to you.

qa/L0_lifecycle/lifecycle_test.py
yinggeh: Can you clean it up in this PR?

yinggeh: If you install pre-commit locally, does it automatically reformat for you?

Vinya567: No  isort and black auto-fix formatting, but flake8 is a reporter. E402/F841/E721/E712/E711/F401 all need manual edits.

qa/L0_lifecycle/lifecycle_test.py
Vinya567: Went with self.assertTrue(msg.startswith(...)) in f36a9bd8.

qa/L0_lifecycle/lifecycle_test.py
Vinya567: > Can you clean it up in this PR?

Cleaned up in f36a9bd8 — removed the # flake8: noqa and fixed the pre-existing lints (E402/F841/E266/E721/E712/E711/F401).

pskiran1 (dismissed):
LGTM

pskiran1 (approved):

yinggeh:
I removed your NVBUGS ID from title.

PR #8896, opened 2026-07-22
docs: Update LLM getting-started guide to PyTorch backend

faradawn:
Deprecating the engine build in favor of pytorch / llmapi backend. Workflow modernized: removes the convert_checkpoint.py + trtllm-build + inflight_batcher_llm ensemble path and replaces it with a single model.yaml-driven LLM API workflow; 

Upgrading the example model from Phi-3 to DeepSeek R1 MVFP4. 

Tested on B200x8.


whoisj:
When do these changes take effect? If they already apply, then I can approve now. If Triton needs to update to a newer version of TRTLLM, then we should wait until then.

faradawn:
These changes are ready. Confirming that DSR1 works on the current Triton and TRTLLM version.

whoisj (approved):

PR #8901, opened 2026-07-23
build: Fix py3-sdk maven dependency

mc-nv:
#### What does the PR do?
Fixes two failure modes in the `py3-sdk` client/SDK image build (`Dockerfile.sdk`):

1. **Maven dependency resilience.** Replaces the `wget` fetch of Apache Maven (which flaked with `exit code: 4` / `Network is unreachable` against archive.apache.org) with `curl -fSLk --retry 3`, and switches the pinned version to a full URL build-arg (`JAVA_BINDINGS_MAVEN_URL`, Maven 3.9.16) that CI already points at the internal Artifactory mirror. The extracted dir is resolved with a glob-based symlink and the install is verified with `mvn --version`.
2. **Build OOM.** Caps client-build parallelism at `--parallel 16`. A bare `cmake --build --parallel` emits an unbounded `make -j`, which spawned one `cc1plus` per ready gRPC translation unit (1–2+ GB each) and tripped the OOM killer (`cannot allocate memory`). 16 matches the RHEL Dockerfile's existing `-j16`.

#### Checklist

#### Commit Type:
- [x] build

#### Related PRs:
N/A — server-only change.

`Dockerfile.sdk` — the "Install Maven 3.9+ manually" block (`ARG JAVA_BINDINGS_MAVEN_URL` + the `curl … | tar | ln -s | mvn --version` RUN) and the client build `RUN … cmake --build . -v --parallel 16 …`.

#### Test plan:
- Rebuild `py3-sdk` and confirm the Maven step pulls 3.9.16 from Artifactory and `mvn --version` succeeds. Verified in pipeline 59299689 (`py3-sdk` green — Maven step + capped client build both pass).
- Confirm the client build no longer OOMs on the `dl325g11-*` builder class.

- CI Pipeline ID: 59299689

#### Caveats:
`curl -k` skips TLS verification for the tarball fetch (retains prior lenient behavior on the internal build network); `--retry 3` absorbs transient drops. The credentialed `JAVA_BINDINGS_MAVEN_URL` build-arg is not exposed in the published image (multi-stage build discards the `sdk_build` stage history — verified against the built SDK image).

#### Background
CI `py3-sdk` / `py3-sdk-rhel` intermittently failed fetching apache-maven from archive.apache.org (`wget … exit code: 4`), and `py3-sdk` separately OOM'd during the gRPC client compile. This makes both steps resilient.

#### Related Issues:
- Resolves: TRI-1621

<sup>
CI (internal): [#59299689](http://tritonserver.local/ci/pipelines/59299689)<br/>
</sup>

### DIFF
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -39,6 +39,7 @@ ARG TRITON_CLIENT_REPO_TAG=main
 ARG TRITON_THIRD_PARTY_REPO_TAG=main
 ARG TRITON_ENABLE_GPU=ON
 ARG JAVA_BINDINGS_JAVACPP_PRESETS_TAG=1.5.8
+ARG JAVA_BINDINGS_MAVEN_URL=https://archive.apache.org/dist/maven/maven-3/3.9.16/binaries/apache-maven-3.9.16-bin.tar.gz
 # DCGM version to install for Model Analyzer
 ARG DCGM_VERSION=4.5.3-1
 
@@ -88,7 +89,7 @@ ENV CMAKE_POLICY_MINIMUM_REQUIRED=3.5
 
 # Install Maven 3.9+ manually -- the distro-provided maven is too old to
 # honor the MAVEN_ARGS environment variable (added in Maven 3.9.0, MNG-7038).
-ARG JAVA_BINDINGS_MAVEN_URL=https://archive.apache.org/dist/maven/maven-3/3.9.9/binaries/apache-maven-3.9.9-bin.tar.gz
+ARG JAVA_BINDINGS_MAVEN_URL
 RUN curl -fSLk --retry 3 -o /tmp/maven.tar.gz ${JAVA_BINDINGS_MAVEN_URL} && \
     tar -xzf /tmp/maven.tar.gz -C /opt && \
     find /opt -maxdepth 1 -name "apache-maven-*" -type d -exec ln -s {} /opt/maven \; && \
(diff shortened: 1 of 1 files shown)


mc-nv:
@greptileai

yinggeh (dismissed):

Dockerfile.sdk
```diff
@@ -89,12 +89,13 @@ ENV CMAKE_POLICY_MINIMUM_REQUIRED=3.5
...
-RUN wget -nv -O /tmp/maven.tar.gz \
-        https://archive.apache.org/dist/maven/maven-3/${JAVA_BINDINGS_MAVEN_VERSION}/binaries/apache-maven-${JAVA_BINDINGS_MAVEN_VERSION}-bin.tar.gz && \
+ARG JAVA_BINDINGS_MAVEN_URL
+RUN --mount=type=secret,id=maven_url,env=MAVEN_URL,required=false \
+    [ -n "${MAVEN_URL}" ] || MAVEN_URL="${JAVA_BINDINGS_MAVEN_URL}"; \
+    curl -fSLk --retry 3 -o /tmp/maven.tar.gz "${MAVEN_URL}" && \
     tar -xzf /tmp/maven.tar.gz -C /opt && \
-    ln -s /opt/apache-maven-${JAVA_BINDINGS_MAVEN_VERSION} /opt/maven && \
-    rm /tmp/maven.tar.gz
+    find /opt -maxdepth 1 -name "apache-maven-*" -type d -exec ln -s {} /opt/maven \; && \
```
nv-rinig, line 102: could there be other versions here or should we explicitly add the version to the search instead of generic wildcard?

mc-nv: The version should be 3.9+, I don't want to hardcode version of the tool, if it's a concern I have other way to implement it.

nv-rinig: can i know what the other implementation would look like? my concern is if there's a possibility of other maven versions that could get inadvertently linked here. if we do clean up/always start from a clean state then it's not a concern, but i would love for you to confirm.

mc-nv: > can i know what the other implementation would look like? 

In that case  we can use strip component flag during the extract and predefine specific. 
I prefer approach with symbolic link as of today. 

> my concern is if there's a possibility of other maven versions that could get inadvertently linked here. if we do clean up/always start from a clean state then it's not a concern, but i would love for you to confirm.

As of this writing there is no other installation in given location.

nv-rinig: ack.

nv-rinig (approved):

PR #8923, opened 2026-08-10
Fix SageMaker MME cross-model invoke/unload via X-Amzn-SageMaker-Target-Model header (TRI-1563)

akhilraj9:
## Summary
The SageMaker multi-model endpoint (MME) adapter authorized invoke and unload
requests using the model identity in the URL (`/models/<hash>`), but performed
the actual inference / unload using the `X-Amzn-SageMaker-Target-Model` header,
without verifying the two referred to the same model. A request scoped to one
loaded model could therefore invoke or unload a *different* loaded model, and the
unload path erased the URL-hash registry entry while unloading the header-selected
model — leaving runtime and repository state inconsistent.

This is a confused-deputy / incorrect-authorization issue (CWE-441 / CWE-863). In
multi-tenant or shared-model deployments it breaks tenant isolation (cross-model
inference, cross-model unload/DoS) and corrupts model lifecycle bookkeeping.

## Root cause
The registry (`sagemaker_models_list_`) stored only `hash → repo_path`, discarding
which model each hash was actually loaded as. At invoke/unload time there was
nothing to validate the header against, so the header was trusted as the action
target.

## Fix — bind identity at load, validate on use
- Extend the registry value from `repo_path` to `{ repo_path, target_model }`
  (new `SageMakerModelInfo`), recording the model each hash was loaded under.
- On invoke and unload, act on the **registered** `target_model`, never on the
  header. If the request carries an `X-Amzn-SageMaker-Target-Model` header that
  does not match the registered value, reject with `400`.
- Unload now evicts the registered model and erases that same entry, so runtime
  and repository state stay consistent.

Legitimate traffic is unaffected: SageMaker sets the header consistently with the
loaded model (or omits it), so the mismatch path only triggers on a contradictory
request.

## Changes
- `src/sagemaker_server.{h,cc}` — the fix
- `qa/L0_sagemaker/sagemaker_multi_model_test.py` — regression test
  (`test_sm_5b_identity_confusion_poc`): mismatched header on invoke and unload
  must return `400`, and the other model must remain loaded
- `qa/L0_sagemaker/test.sh` — bump expected MME test count 7 → 8

## Testing
`L0_sagemaker` passes (8/8), including the new regression test. Verified end-to-end
on the internal CI pipeline (build + `L0_sagemaker` green).
Pipeline ID: 65892700


src/sagemaker_server.cc
```diff
@@ -697,34 +715,53 @@ void
...
+              << target_model_header
+              << "' does not match the model registered under URL hash '"
+              << model_name_hash << "'";
+    TRITONSERVER_Error* err = TRITONSERVER_ErrorNew(
+        TRITONSERVER_ERROR_INVALID_ARG,
+        "X-Amzn-SageMaker-Target-Model header does not match the model "
+        "requested in the URL");
+    EVBufferAddErrorJson(req->buffer_out, err);
+    evhtp_send_reply(req, EVHTP_RES_BADREQ); /* 400 */
+    TRITONSERVER_ErrorDelete(err);
```
yinggeh: Can use macro `HTTP_RESPOND_IF_ERR`

src/sagemaker_server.cc
```diff
@@ -1121,8 +1156,12 @@ SagemakerAPIServer::SageMakerMMELoadModel(
   } else {
     std::lock_guard<std::mutex> lock(models_list_mutex_);
 
-    /* Use model name hash as expected in SageMaker MME contract */
-    sagemaker_models_list_.emplace(model_name_hash, repo_parent_path);
+    /* Use model name hash as the key, as expected in the SageMaker MME
+     * contract, and remember the target_model the repository was actually
+     * loaded under so invoke/unload can validate the request header against it
+     * (TRI-1563). */
```
yinggeh: I wouldn't put ticket number unless there is future work.

src/sagemaker_server.h
```diff
@@ -169,8 +169,20 @@ class SagemakerAPIServer : public HTTPAPIServer {
...
   const std::string inference_type_;
 
-  /* Maintain list of loaded models */
-  std::unordered_map<std::string, std::string> sagemaker_models_list_;
+  /* Information tracked for each loaded MME model, keyed by the model-name
+   * hash from the request URL. */
+  struct SageMakerModelInfo {
+    // Parent path of the registered model repository
+    // (e.g. /opt/ml/models/<hash>). Needed to unregister the repo on unload.
+    std::string repo_path;
```
yinggeh: const

qa/L0_sagemaker/sagemaker_multi_model_test.py
```diff
@@ -320,6 +320,112 @@ def test_sm_5_model_unload(self):
             "Expected status code 404, received {}".format(r.status_code),
         )
 
+    def test_sm_5b_identity_confusion_poc(self):
```
yinggeh: What's sm_5b?

qa/L0_sagemaker/sagemaker_multi_model_test.py
```diff
@@ -29,16 +29,16 @@
...
+import time  # noqa: E402
+import unittest  # noqa: E402
 
-import numpy as np
-import requests
-import test_util as tu
-import tritonclient.http as httpclient
+import numpy as np  # noqa: E402
+import requests  # noqa: E402
+import test_util as tu  # noqa: E402
```
yinggeh, line 40: Move `sys.path.append("../common")` before this line for fewer ` # noqa: E402`.

src/sagemaker_server.cc
akhilraj9: Good call — switched both the invoke and unload rejection paths to HTTP_RESPOND_IF_ERR. Same 400 + JSON body, a lot less boilerplate. Done.

src/sagemaker_server.cc
akhilraj9: Removed the ticket number here and in the two related comments. Done.

src/sagemaker_server.h
akhilraj9: Made both repo_path and target_model const — they're set once at load and only read after. Done.

qa/L0_sagemaker/sagemaker_multi_model_test.py
akhilraj9: sm = sagemaker; 5b just orders it right after test_sm_5_model_unload (unittest runs tests lexicographically). Renamed to test_sm_5b_target_model_header_mismatch with a docstring so it's self-explanatory, and dropped "poc" — it's a permanent regression test now, not a one-off.

qa/L0_sagemaker/sagemaker_multi_model_test.py
akhilraj9: Done — moved the stdlib/third-party imports above sys.path.append("../common") so only test_util needs # noqa: E402 now (8 → 1). Matches the existing pattern in e.g. L0_cuda_graph/trt_cuda_graph_test.py.

whoisj (approved):
LGTM. @yinggeh are you happy with this now?

yinggeh (approved):

PR #8925, opened 2026-08-10
fix: catch std::out_of_range when parsing inference header length in HTTP frontends

akhilraj9:
#### What does the PR do?
`GetInferenceHeaderLength` parses the `Inference-Header-Content-Length` request
header (and the SageMaker/Vertex `json-header-size` MIME parameter) with
`std::stoi`. `std::stoi` throws `std::out_of_range` for values outside the
`int` range, but the three frontend implementations only caught
`std::invalid_argument`. An uncaught `std::out_of_range` unwinds out of the
handler into the libevhtp C event loop, which has no C++ landing pad, so the
runtime calls `std::terminate()` and the server process aborts.

This applies the same `catch (const std::out_of_range&)` pattern already used by
the sibling `GetContentLength` in `http_server.cc`, returning a 400 instead of
terminating. Fixed in all three HTTP frontends:
- `HTTPAPIServer::GetInferenceHeaderLength` (src/http_server.cc)
- `SagemakerAPIServer::GetInferenceHeaderLength` + its Content-Length parse (src/sagemaker_server.cc)
- `VertexAiAPIServer::GetInferenceHeaderLength` (src/vertex_ai_server.cc)

#### Commit Type:
- [x] fix

`src/http_server.cc` — compare the new `catch (const std::out_of_range& oor)`
block in `GetInferenceHeaderLength` against the existing one in the sibling
`GetContentLength` (same file); they now match.

#### Test plan:
Added one regression test per frontend that sends a header value of
`99999999999` (exceeds INT_MAX) and asserts the server returns 400 and stays
live:
- `qa/L0_http/http_test.py::test_inference_header_content_length_out_of_range`
- `qa/L0_sagemaker/sagemaker_test.py::test_malformed_binary_header_out_of_range`
- `qa/L0_vertex_ai/vertex_ai_test.py::test_malformed_binary_header_out_of_range`

Expected unit-test counts in the corresponding `test.sh` harnesses were bumped
(17 / 10 / 9). L0_http, L0_sagemaker and L0_vertex_ai all pass.

- CI Pipeline ID: 61802272

#### Caveats:
None. Change is limited to exception handling on an existing error path.

#### Background
Regression/incomplete-fix residual of CVE-2026-24210: the `out_of_range` catch
was added to `GetContentLength` but not to the three `GetInferenceHeaderLength`
variants when `std::atoi` was replaced with `std::stoi`. Reported via internal
security tracking.

#### Related Issues:
- Relates to internal tracking DLIS-8702 / NVBug 6522755 / TRI-1636


pskiran1 (dismissed):
LGTM!

pskiran1:
@akhilraj9, please fix pre-commit errors

whoisj (approved):

yinggeh (approved):

yinggeh:
@pskiran1 Do we have a pipeline running?

pskiran1:
@akhilraj9, please update the PR with the latest Pipeline ID, we will review the pipeline results and merge the PR.

akhilraj9:
> @akhilraj9, please update the PR with the latest Pipeline ID, we will review the pipeline results and merge the PR.

Thanks for the review. The Pipeline ID is already available in the PR description. For quick reference: 61802272.

PR #8926, opened 2026-08-10
improve(Build): Add build script capabilities

jcuquemelle:
#### What does the PR do?
This PR adds several parameters to the main build script (and/or forwards them to ORT bakend build) to add configurability:
- forward the main build parallelism to ORT nested build (this is partly addressed by https://github.com/triton-inference-server/server/pull/8873, but having the parameter forwarded makes it unified on nested build, and it allows to fix issues when the 2GB / concurrent compiler does not hold)
- Define a CUDA arch list to support by the built image (and forward to ORT build). This allows to speed up the build and produce lighter artifacts when one doesn't need to support the full set of compute capabilities.
- Allow to use a local directory as `--github-organization`. This enables having a local clone of needed triton repos and build against them instead of fetching remote repositories, even in nested builds. 

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/onnxruntime_backend/pull/354

Only 1 file changed in the PR

#### Test plan:
Run the build with related options 
Tested on internal CI

#### Caveats:
Only tested forwarding to onnx runtime, other backends could need adaption for these options to be active

#### Background
We needed this for adaption to our build system's constraints, and to iterate faster by removing unneeded CPs in nested buids, ORT in particular

None

### DIFF
--- a/build.py
+++ b/build.py
@@ -1761,6 +1761,11 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
             if os.environ.get(var):
                 runargs += ["-e", f"{var}={os.environ[var]}"]
 
+        # Forward the CUDA arch list into the buildbase container so nested
+        # backend builds honor --cuda-arch-list.
+        if FLAGS.cuda_arch_list is not None:
+            runargs += ["-e", f'"CUDA_ARCH_LIST={FLAGS.cuda_arch_list}"']
+
         runargs += ["tritonserver_buildbase"]
 
         runargs += ["./cmake_build"]
--- a/build.py
+++ b/build.py
@@ -1724,7 +1724,7 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
         if not FLAGS.no_container_interactive:
             runargs += ["-it"]
 
-        runargs += ["-v", "/var/run/docker.sock:/var/run/docker.sock"]
+        runargs += docker_runargs()
         if FLAGS.use_user_docker_config:
             if os.path.exists(FLAGS.use_user_docker_config):
                 runargs += [
@@ -2243,6 +2243,29 @@ def min_cuda_arch(arch_list_str):
             val = val / 10.0
         mins.append(val)
     return f"{min(mins):.1f}" if mins else None
+
+
+def docker_runargs():
+    # Docker run args for the host-mounted build container: how to reach the
+    # daemon socket, and whether the host network is needed.
+    #
+    # DOCKER_HOST unset → system docker; mount the default socket. The default
+    # bridge has DNS, so no `--network host`.
+    #
+    # DOCKER_HOST set → rootless or remote daemon. Forward the env var so the
+    # in-container `docker build` reaches the same daemon. Rootless's default
+    # bridge has no DNS, so add `--network host` to let FetchContent resolve
+    # github.com. (We don't probe `docker info --format .Rootless` because if
+    # DOCKER_HOST is set we're already in the rootless/remote-daemon case.)
+    docker_host = os.getenv("DOCKER_HOST")
+    if docker_host is None:
+        return ["-v", "/var/run/docker.sock:/var/run/docker.sock"]
+
+    runargs = ["--network", "host", "-e", f"DOCKER_HOST={docker_host}"]
+    if docker_host.startswith("unix://"):
+        socket_path = docker_host[len("unix://") :]
+        runargs += ["-v", f"{socket_path}:{socket_path}"]
+    return runargs
 
 
 def enable_all():
--- a/build.py
+++ b/build.py
@@ -1730,9 +1730,12 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
         # (TRITON_REPO_ORGANIZATION), so pointing it at a local mirror is the
         # only way to resolve deeply-nested clones (e.g. core -> common)
         # offline; FETCHCONTENT_SOURCE_DIR overrides do not reach those builds.
+        # FLAGS.github_organization is already normalized as an absolute path
         if FLAGS.github_organization and os.path.isdir(FLAGS.github_organization):
-            org_abs = os.path.abspath(FLAGS.github_organization)
-            runargs += ["-v", f"{org_abs}:{org_abs}"]
+            runargs += [
+                "-v",
+                f"{FLAGS.github_organization}:{FLAGS.github_organization}",
+            ]
         if FLAGS.use_user_docker_config:
             if os.path.exists(FLAGS.use_user_docker_config):
                 runargs += [
@@ -2755,6 +2758,12 @@ if __name__ == "__main__":
         FLAGS.extra_backend_cmake_arg = []
     if FLAGS.build_secret is None:
         FLAGS.build_secret = []
+
+    # When --github-organization is a local directory, resolve it to an absolute
+    # path, so nested clones and CMake's TRITON_REPO_ORGANIZATION properly reference it.
+    # Absolute values and remote URLs are left unchanged.
+    if FLAGS.github_organization and os.path.isdir(FLAGS.github_organization):
+        FLAGS.github_organization = os.path.abspath(FLAGS.github_organization)
 
     FLAGS.boost_url = os.getenv(
         "TRITON_BOOST_URL",
--- a/build.py
+++ b/build.py
@@ -668,12 +668,6 @@ def onnxruntime_cmake_args(images, library_paths):
             else FLAGS.ort_version,
         )
     ]
-    if FLAGS.build_parallel_was_explicit:
-        cargs.append(
-            cmake_backend_arg(
-                "onnxruntime", "TRITON_BUILD_PARALLEL", None, FLAGS.build_parallel
-            )
-        )
     # TRITON_ENABLE_GPU is already set for all backends in backend_cmake_args()
     if FLAGS.enable_gpu:
         # TODO: TPRD-712 TensorRT is not currently supported by our RHEL build for SBSA.
@@ -2954,6 +2948,11 @@ if __name__ == "__main__":
         )
         log('CMake core override "-D{}={}"'.format(parts[0], parts[1]))
         OVERRIDE_CORE_CMAKE_FLAGS[parts[0]] = parts[1]
+
+    if FLAGS.build_parallel_was_explicit and "onnxruntime" in backends:
+        FLAGS.extra_backend_cmake_arg.append(
+            f"onnxruntime:TRITON_BUILD_PARALLEL={FLAGS.build_parallel}"
+        )
 
     for cf in FLAGS.extra_backend_cmake_arg:
         parts = cf.split(":", 1)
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -2723,9 +2812,30 @@ def enable_all():
...
 
+    FLAGS.build_parallel_was_explicit = FLAGS.build_parallel is not None
     if FLAGS.build_parallel is None:
         FLAGS.build_parallel = default_build_parallel()
 
+    # CUDA arch list: export into the environment so backends that read
+    # $CUDA_ARCH_LIST directly (triton-backend's define.cuda_architectures.cmake)
+    # see the same value the ONNX Runtime backend gets via TRITON_CUDA_ARCH_LIST.
+    if FLAGS.cuda_arch_list is not None:
+        os.environ["CUDA_ARCH_LIST"] = FLAGS.cuda_arch_list
```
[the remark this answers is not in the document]
jcuquemelle: Arch list is passed as an env var to the buildBase docker build, so is accessible to all nested builds

build.py
```diff
@@ -1714,7 +1731,15 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
...
-        runargs += ["-v", "/var/run/docker.sock:/var/run/docker.sock"]
+        runargs += docker_runargs()
+        # If --github-organization is a local directory (a repo mirror), mount
+        # it too. The org propagates into every nested component configure
+        # (TRITON_REPO_ORGANIZATION), so pointing it at a local mirror is the
+        # only way to resolve deeply-nested clones (e.g. core -> common)
+        # offline; FETCHCONTENT_SOURCE_DIR overrides do not reach those builds.
+        if FLAGS.github_organization and os.path.isdir(FLAGS.github_organization):
+            org_abs = os.path.abspath(FLAGS.github_organization)
+            runargs += ["-v", f"{org_abs}:{org_abs}"]
```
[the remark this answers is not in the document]
jcuquemelle: fixed in https://github.com/triton-inference-server/server/pull/8926/commits/40d2b6e6fc48e64b3efbce7e4b37fc2c4d9f8a88

Vinya567:
Hi @jcuquemelle -

Thank you for the PR!

Confirmed on our side: this adds optional `build.py` knobs — `--cuda-arch-list` (via `$CUDA_ARCH_LIST`), `DOCKER_HOST` for rootless/remote Docker, and local `--github-organization` as a bind-mounted mirror.

Validated through our internal CI (Pipeline ID: 64574726, branch `vinyak/tri-1722-server-8926` with your `build.py` changes on current main). `py3-base` ran `./build.py -v` without `--cuda-arch-list` or `--build-parallel`, and `DOCKER_HOST` was unset. `py3-devel` and the existing ORT L0s passed.

build.py
```diff
@@ -82,6 +82,11 @@
     "rhel_py_version": "3.12.3",
 }
 
+# Minimum CUDA compute capability Triton will load models for. Used as the
+# default for --min-compute-capability when neither that flag nor
+# --cuda-arch-list is given.
+DEFAULT_MIN_COMPUTE_CAPABILITY = "6.0"
+
```
mc-nv: This value can't be used currently we support 7.5 and higher and we no longer rely on it in favor of `CUDA_ARCH_LIST`
https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#gpu-feature-list

build.py
```diff
@@ -663,7 +668,12 @@ def onnxruntime_cmake_args(images, library_paths):
             else FLAGS.ort_version,
         )
     ]
-
+    if FLAGS.build_parallel_was_explicit:
+        cargs.append(
+            cmake_backend_arg(
+                "onnxruntime", "TRITON_BUILD_PARALLEL", None, FLAGS.build_parallel
+            )
+        )
```
mc-nv: I would suggest use `--extra-backend-cmake-arg` instead.

build.py
```diff
@@ -1714,7 +1724,18 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
...
+        # it too. The org propagates into every nested component configure
+        # (TRITON_REPO_ORGANIZATION), so pointing it at a local mirror is the
+        # only way to resolve deeply-nested clones (e.g. core -> common)
+        # offline; FETCHCONTENT_SOURCE_DIR overrides do not reach those builds.
+        # FLAGS.github_organization is already normalized as an absolute path
+        if FLAGS.github_organization and os.path.isdir(FLAGS.github_organization):
+            runargs += [
+                "-v",
+                f"{FLAGS.github_organization}:{FLAGS.github_organization}",
+            ]
```
mc-nv, line 1732: I don't think it make sense to mess up `git` url prefixes with local mounts.

build.py
```diff
@@ -1743,6 +1764,11 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
             if os.environ.get(var):
                 runargs += ["-e", f"{var}={os.environ[var]}"]
 
+        # Forward the CUDA arch list into the buildbase container so nested
+        # backend builds honor --cuda-arch-list.
+        if FLAGS.cuda_arch_list is not None:
+            runargs += ["-e", f'"CUDA_ARCH_LIST={FLAGS.cuda_arch_list}"']
+
```
mc-nv, line 1765: It can be added to override some system defined variable.
Need more context on it.

build.py
```diff
@@ -2208,6 +2234,56 @@ def finalize_build(cmake_script, install_dir, ci_dir):
...
+            val = float(tok)
+        except ValueError:
+            continue
+        # "86" means 8.6; "8.6" stays; "100" means 10.0
+        if val >= 100:
+            val = val / 10.0
+        elif val >= 20 and val.is_integer():
+            val = val / 10.0
+        mins.append(val)
+    return f"{min(mins):.1f}" if mins else None
```
mc-nv, line 2255: Seams like a CMake codes for NVCC compiler, what is the motivation to add it here?

build.py
```diff
@@ -2208,6 +2234,56 @@ def finalize_build(cmake_script, install_dir, ci_dir):
...
+    # DOCKER_HOST is set we're already in the rootless/remote-daemon case.)
+    docker_host = os.getenv("DOCKER_HOST")
+    if docker_host is None:
+        return ["-v", "/var/run/docker.sock:/var/run/docker.sock"]
+
+    runargs = ["--network", "host", "-e", f"DOCKER_HOST={docker_host}"]
+    if docker_host.startswith("unix://"):
+        socket_path = docker_host[len("unix://") :]
+        runargs += ["-v", f"{socket_path}:{socket_path}"]
+    return runargs
```
mc-nv, line 2278: This need to be tested.

build.py
mc-nv, line 1: @jcuquemelle could you please share your `build.py` execution call so we can better understand the needs.

build.py
jcuquemelle: that was a stale value from L2384
parser.add_argument(
        "--min-compute-capability",
        type=str,
        required=False,
        default="6.0",

Will update but it could be dropped, see discussion on https://github.com/triton-inference-server/server/pull/8926/changes#r3867237489

build.py
jcuquemelle: this function reads a list of supported archs string, and derives the minimum one. This is needed to override the default value in the build script (L2384 before this PR), since it would be faulty if a build is run with e.g. `--cuda-arch-list 8.0,8.9,...`

Another way to handle this could be to totally remove the `--min-compute-capability` in this script (fully rely on `--cuda-arch-list`): the min compute capability is only forwarded to the `core` build, so resolving the min capability from cuda arch list could be done here

build.py
jcuquemelle: CUDA_ARCH_LIST env var is already used by `backend` and `onnxruntime-backend` builds. This addition merely adds it in the buildbase container so when these backends are built nested from build.py script with an explicit `--cuda-arch-list`, the list is properly forwarded to the containerized nested builds. 
For host builds, the env var is directly exported (L2830)

build.py
jcuquemelle: For more context: We have a single github organization for all third party forked projects (https://github.com/criteo-forks), hence all our clones of triton repos are prefixed with `triton-`

Since the build.py (and nested builds) rely on a proper naming, at build time we clone all repos locally, and link them with the canonical names, which allows to pass the folder containing these links as `--github-organization`; this also allows a cross-repo development to be built locally with a single `build.py` invocation without needing to push them first on remotes (which we can't actually do since the `triton-` prefix does not allow that). 

the build.py invocation: python3 ./build.py -v \
  --no-container-interactive \
  --github-organization ./.triton-org-mirror \
  --container-prebuild-command "git config --system --add safe.directory '*'" \
  --target-platform linux \
  --build-type Release \
  --cuda-arch-list 8.6 .... \
  --enable-gpu \
  --enable-logging --enable-stats --enable-metrics --enable-gpu-metrics --enable-cpu-metrics --enable-tracing --enable-nvtx \
  --endpoint grpc --endpoint http --filesystem s3 \
  --backend onnxruntime \
  --override-backend-cmake-arg "onnxruntime:TRITON_ENABLE_ONNXRUNTIME_OPENVINO=OFF" \
  --override-backend-cmake-arg "onnxruntime:TRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON" \
  -j 24

build.py
jcuquemelle: See my reply to your comment here: https://github.com/triton-inference-server/server/pull/8926#discussion_r3913723002 for more context

build.py
mc-nv: afraid removing PTX will drop forward compatibility.

mc-nv: It may not be an issue for specific scenario but could be a problem for long living binaries.

build.py
```diff
@@ -2208,6 +2228,56 @@ def finalize_build(cmake_script, install_dir, ci_dir):
...
+    # bridge has DNS, so no `--network host`.
+    #
+    # DOCKER_HOST set → rootless or remote daemon. Forward the env var so the
+    # in-container `docker build` reaches the same daemon. Rootless's default
+    # bridge has no DNS, so add `--network host` to let FetchContent resolve
+    # github.com. (We don't probe `docker info --format .Rootless` because if
+    # DOCKER_HOST is set we're already in the rootless/remote-daemon case.)
+    docker_host = os.getenv("DOCKER_HOST")
+    if docker_host is None:
+        return ["-v", "/var/run/docker.sock:/var/run/docker.sock"]
```
mc-nv, line 2272: Thinking what to do with `docker context` in such a way.

build.py
jcuquemelle: This is only to resolve the min supported CC from the list. If PTX, then this function returns None, and this will resolve to the default min CC as it's the case without this PR (see L2830)

jcuquemelle:
Removed the rootless docker compatibility from this PR

PR #8927, opened 2026-08-11
feat: add a Python driver that manifests, archives and publishes QA models

mc-nv:
#### What does the PR do?

Adds a Python driver for QA model generation beside the existing shell one, and gives every
generated model a manifest describing how it was built. Models can then be packed into
per-framework (or per-model) archives and published, so a consumer fetches the backend it
needs rather than the whole corpus.

The shell driver is unchanged and still works; nothing here is required to use it.

Three problems this addresses, each of which previously failed silently:

- **A rebuild cost the whole corpus.** Stages are now selectable (`--onnx --pytorch`, or
  `TRITON_MODELS_FRAMEWORKS`), and an unrecognised framework name is refused rather than
  skipped — a typo that generated nothing used to surface hours later as a test failure.
- **Nothing recorded what built an artifact.** Each model gets a `manifest.json` with the
  container, platform, framework version, GPU and format-specific fields (ONNX opset, IR
  version, TensorRT version). Those flatten into properties an archive can be selected on.
- **The GPU was exposed to every stage.** The nvidia runtime mounts `nvidia-smi` into any
  image, so the CPU-only OpenVINO and ONNX stages probed a real GPU and stamped a compute
  capability onto artifacts that do not depend on one. GPU access is now per stage.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [x] docs
- [x] feat
- [x] fix
- [ ] perf
- [x] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

None in other GitHub repositories. The CI change that consumes these drivers lives in an
internal repository and is linked from the tracking issue.

`qa/common/gen_qa_model_repository.py` — the driver. The stage table near the top is the
whole model: four stages, each with an image, a privilege level and a `needs_gpu` flag.

Then `qa/common/gen_manifest.py` for what gets recorded, and `qa/common/README.md`, which
documents the flags and the reasoning behind the layout choices.

`qa/common/gen_qa_model_repository` (the shell driver) is worth a look to confirm the
changes there are confined to version handling.

#### Test plan:

Verified by running real generations rather than by inspection:

- Each stage individually, docker runtime — `openvino` (16 models), `onnx` (735 models),
  `pytorch` + `tensorrt` (197 models, 100 `.plan`, 101 `.pt`).
- GPU gating — after restricting the GPU to the stages that need it, the ONNX stage's
  manifests record `gpu: null` across all 735 models, including those from
  `gen_qa_dyna_sequence_models.py`, while `pytorch` and `tensorrt` record `cc=8.9`.
  Before the change the same ONNX models carried `compute_capability: 8.9`.
- Archiving and the layout it produces — `manifest-summary.json` at the tree root,
  `archives/` beside it holding the tarballs, per-bundle manifests and `index.json`.
- The shell driver still runs, which is the compatibility bar for this PR.
- `pre-commit` on every changed file.

Not covered: the `L0_*` suites have not been run against a corpus produced by the Python
driver. The archives it produces are byte-identical in content to the shell driver's output
for the same inputs, but that equivalence is argued rather than measured.

- CI Pipeline ID: 62039930

#### Caveats:

- Archiving warns rather than fails on error. By that point the models are generated and
  correct, and failing a multi-hour run over packaging would discard them.
- `gen_archive.py` run standalone has no fallback to `build.py` for the semver, so it can
  produce an archive name with that field missing. The driver always passes it explicitly,
  so this only affects direct invocation.
- The `tensorrt` bundle records no `driver_version`; that image has `cuda-python` but no
  `pynvml`.

#### Background

The QA corpus was published as one artifact per repository directory, fetched whole over a
CIFS mount. A test run needing one backend pulled everything, and an unstable link made that
expensive to retry. Splitting the corpus into described, individually selectable archives is
what makes a partial fetch possible.

- Resolves: TRI-1105

<sup>
CI (internal): [#62039930](http://tritonserver.local/ci/pipelines/62039930)<br/>
</sup>


qa/L0_backend_python/python_based_backends/test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
yinggeh: It is a mistake?

qa/common/gen_archive.py
```diff
@@ -0,0 +1,423 @@
...
+                )
+    data = raw.getvalue()
+    if not compress:
+        return data
+    # mtime=0: gzip stamps the current time in its header otherwise, which would
+    # defeat reproducibility even with a deterministic tar inside.
+    out = io.BytesIO()
+    with gzip.GzipFile(fileobj=out, mode="wb", compresslevel=6, mtime=EPOCH) as gz:
+        gz.write(data)
+    return out.getvalue()
```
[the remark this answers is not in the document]
mc-nv: Confirmed and fixed — `perf: stream framework bundles instead of buffering them` (0784be249, pending push).

`build_bundle_bytes` is replaced by `write_bundle(model_dirs, tree, stream, compress)`, which packs straight into a caller-supplied stream and returns `(bytes_written, sha256)` from a wrapper that hashes and counts as bytes pass through. The wrapper sits **outside** the gzip layer, so a compressed bundle still hashes to its compressed bytes — which is what the `X-Checksum-Sha256` header commits to.

Measured on a 200 MiB stand-in for the 12.78 GiB model:

```
OLD buffered peak:  217.5 MiB  (1.09x input)
NEW streamed peak:    ~0 MiB
```

One correction to the finding: the cost was ~1.09x, not several archive-sized allocations. `raw.getvalue()` does not hold both copies simultaneously in CPython — peak never exceeded the tar build itself. Still an archive-sized ceiling worth removing, just a smaller multiple than stated.

Output is **byte-identical** to the previous implementation on both the plain and gzip paths (verified by running the old function alongside the new one and diffing), so archives stay reproducible and checksums already in Artifactory still match.

Two things outside the finding that the fix had to handle:

- Streaming writes as it goes, so a mid-pack failure would leave a truncated archive where a valid one belongs — and the upload step trusts what it finds on disk. Now packed under a `.partial` name and renamed once complete.
- `--dry-run` still reports `archive_bytes` and `sha256`, so it packs to a discarding stream rather than skipping the work. Verified dry-run and real run agree on both values.

qa/common/gen_artifactory.py
```diff
@@ -0,0 +1,430 @@
...
+        target = build_target(args.url, bundle_path(args, framework), name, properties)
+
+        if args.dry_run:
+            # The full target, properties and all: a dry run is for checking
+            # where a thing would land, which the name alone does not say.
+            _log("would upload {}".format(target))
+            uploaded += 1
+            continue
+        try:
+            payload = source.read_bytes()
```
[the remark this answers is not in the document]
mc-nv: Confirmed and fixed — `perf: stream archive uploads from disk` (769dc5f99, pending push).

`upload_archive` now takes a path and hands `urllib` an open file object instead of `source.read_bytes()`. You were right that the allocation outlives the attempt that uses it: the `Request` was built once at line 205, **outside** the `for attempt in range(...)` loop, so the bytes stayed pinned across every retry rather than just the one sending them.

Two consequences the conversion had to handle, both verified against a local HTTP server:

- **`Content-Length` is now set explicitly.** `urllib` cannot size a file object, so `_get_content_length()` returns `None` and it falls back to `Transfer-Encoding: chunked`. Confirmed on the wire: `Content-Length=67108864`, `Transfer-Encoding=None`.
- **The request is built inside the retry loop, on a freshly opened handle.** A file object is consumed by the attempt that sends it, so hoisting the request out of the loop — which the bytes version did safely — would replay an exhausted stream on every retry after the first. A 503 → 503 → 200 sequence sends the full body on all three attempts.

Peak allocation for a 64 MiB archive is 2.0 MiB (0.032x). Checksum and auth headers are unchanged.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1909 @@
...
+            "        build-essential \\\n"
+            "        cmake \\\n"
+            "        libprotobuf-dev \\\n"
+            "        protobuf-compiler \\\n"
+            "        python3 \\\n"
+            "        python3-dev \\\n"
+            "        python3-pip \\\n"
+            "        wget \\\n"
+            "        gnupg2 \\\n"
+            "        software-properties-common",
```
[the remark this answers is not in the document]
mc-nv: Fixed in aebec9c81 — joined with explicit `+`.

This was a deliberate line wrap of one long shell command, not a missing comma. Rewriting it as explicit concatenation keeps the resulting string identical while removing the ambiguity the rule exists to flag: inside a list literal, adjacent string literals and a forgotten comma look the same.

Verified nothing changed: every string constant reachable from the module's code objects is identical before and after. That holds because CPython folds `"a" + "b"` into a single constant at compile time, exactly as it does adjacent literals. Worth checking rather than eyeballing here — these strings *are* the shell scripts the generator emits, so silent drift would change what CI runs.

Site: the openvino stage'\''s `apt-get install` package list.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1909 @@
...
+        image=ctx.ubuntu_image,
+        needs_root=True,
+        needs_gpu=False,
+        prelude=[],
+        setup=[
+            "export DEBIAN_FRONTEND=noninteractive",
+            "apt-get update && \\\n"
+            "        apt-get install -y --no-install-recommends build-essential cmake"
+            " libprotobuf-dev \\\n"
+            "                protobuf-compiler python3 python3-dev python3-pip",
```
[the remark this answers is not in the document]
mc-nv: Fixed in aebec9c81 — joined with explicit `+`.

This was a deliberate line wrap of one long shell command, not a missing comma. Rewriting it as explicit concatenation keeps the resulting string identical while removing the ambiguity the rule exists to flag: inside a list literal, adjacent string literals and a forgotten comma look the same.

Verified nothing changed: every string constant reachable from the module's code objects is identical before and after. That holds because CPython folds `"a" + "b"` into a single constant at compile time, exactly as it does adjacent literals. Worth checking rather than eyeballing here — these strings *are* the shell scripts the generator emits, so silent drift would change what CI runs.

Site: the onnx stage'\''s `apt-get install` line.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1909 @@
...
+            "ln -s /usr/bin/python3 /usr/bin/python",
+            # TODO: Remove the pins, DLIS-3838. Kept verbatim for parity: the
+            # protobuf cap contradicts onnx's own protobuf>=4.25.1 and is
+            # overwritten by the --upgrade below, and numpy<=1.23.5 has no
+            # cp312 wheel, so both only survive because the default image is
+            # ubuntu:22.04 with Python 3.10.
+            'pip3 install "protobuf<=3.20.1" "numpy<=1.23.5" "ml_dtypes<=0.5.4"'
+            "  # TODO: Remove current line DLIS-3838",
```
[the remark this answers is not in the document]
mc-nv: Fixed in aebec9c81 — joined with explicit `+`.

This was a deliberate line wrap of one long shell command, not a missing comma. Rewriting it as explicit concatenation keeps the resulting string identical while removing the ambiguity the rule exists to flag: inside a list literal, adjacent string literals and a forgotten comma look the same.

Verified nothing changed: every string constant reachable from the module's code objects is identical before and after. That holds because CPython folds `"a" + "b"` into a single constant at compile time, exactly as it does adjacent literals. Worth checking rather than eyeballing here — these strings *are* the shell scripts the generator emits, so silent drift would change what CI runs.

Site: the pip3 pin line, where the second fragment appends a trailing `# TODO: Remove current line DLIS-3838` shell comment.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1909 @@
...
+
+def render_stage_script(stage, ctx, final=False):
+    """The bash script a stage runs inside its container."""
+    lines = [
+        "#!/bin/bash",
+        "# Generated by gen_qa_model_repository.py -- do not edit.",
+        "# Make all generated files accessible outside of container",
+        "umask 0000",
+        "nvidia-smi --query-gpu=compute_cap,compute_mode,driver_version,name,index"
+        " --format=csv || true",
```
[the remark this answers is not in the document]
mc-nv: Fixed in aebec9c81 — joined with explicit `+`.

This was a deliberate line wrap of one long shell command, not a missing comma. Rewriting it as explicit concatenation keeps the resulting string identical while removing the ambiguity the rule exists to flag: inside a list literal, adjacent string literals and a forgotten comma look the same.

Verified nothing changed: every string constant reachable from the module's code objects is identical before and after. That holds because CPython folds `"a" + "b"` into a single constant at compile time, exactly as it does adjacent literals. Worth checking rather than eyeballing here — these strings *are* the shell scripts the generator emits, so silent drift would change what CI runs.

Site: the `nvidia-smi --query-gpu` invocation and its `--format=csv || true` tail.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1909 @@
...
+    lines.append("set -e")
+    lines += list(stage.setup)
+    lines.append("")
+    for step in stage.steps:
+        lines += step.render(ctx)
+    if final:
+        lines += [
+            "",
+            "# Phase 2: sizes are only final once every stage has written its"
+            " artifacts.",
```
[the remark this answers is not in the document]
mc-nv: Fixed in aebec9c81 — joined with explicit `+`.

This was a deliberate line wrap of one long shell command, not a missing comma. Rewriting it as explicit concatenation keeps the resulting string identical while removing the ambiguity the rule exists to flag: inside a list literal, adjacent string literals and a forgotten comma look the same.

Verified nothing changed: every string constant reachable from the module's code objects is identical before and after. That holds because CPython folds `"a" + "b"` into a single constant at compile time, exactly as it does adjacent literals. Worth checking rather than eyeballing here — these strings *are* the shell scripts the generator emits, so silent drift would change what CI runs.

Site: the phase-2 banner comment emitted into the generated script.

qa/common/gen_manifest.py
```diff
@@ -0,0 +1,1414 @@
...
+    Field 3 is the directory within the filesystem that is mounted, field 4 the
+    mount point. See Documentation/filesystems/proc.rst.
+    """
+    try:
+        with open("/proc/self/mountinfo") as handle:
+            for line in handle:
+                fields = line.split()
+                if len(fields) > 4 and fields[4] == "/":
+                    return fields[3]
+    except OSError:
```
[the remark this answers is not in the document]
mc-nv: Documented in f28a2a417.

Best-effort by design, so the fix is to say so rather than to change behaviour — the rule fires on `pass` *without an explanatory comment*, and the sibling handler a few lines below that already carried one was never flagged.

`/proc/self/mountinfo` being absent or unreadable is not an error here: the backing path is descriptive metadata, and `""` is the honest answer when it cannot be determined.

qa/common/gen_manifest.py
```diff
@@ -0,0 +1,1414 @@
...
+        # Any driver-level failure means we cannot describe the GPU; fall back.
+        return None
+
+    bindings = _import_optional("cuda.bindings.driver")
+    if bindings is not None:
+        try:
+            _, version = bindings.cuDriverGetVersion()
+            details["cuda_runtime"] = str(version)
+            details["cuda_version"] = format_cuda_version(version)
+        except Exception:
```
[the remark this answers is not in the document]
mc-nv: Documented in f28a2a417.

Enrichment only. The `details` dict assembled above already describes the GPU; a `cuda.bindings.driver` failure leaves `cuda_runtime` and `cuda_version` at their `None` defaults rather than discarding everything gathered so far.

qa/common/gen_manifest.py
```diff
@@ -0,0 +1,1414 @@
...
+    # expose it, and cuDriverGetVersion reports something different.
+    nvml = _import_optional("pynvml")
+    if nvml is not None:
+        try:
+            nvml.nvmlInit()
+            try:
+                details["driver_version"] = nvml.nvmlSystemGetDriverVersion()
+            finally:
+                nvml.nvmlShutdown()
+        except Exception:
```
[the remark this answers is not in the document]
mc-nv: Documented in f28a2a417.

`pynvml` imported but unusable — no driver loaded, or no permission to talk to it. `driver_version` stays absent, which is preferable to substituting the `cuDriverGetVersion` value: that is a different number and would be recorded under a key that claims to mean the display driver version.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1909 @@
...
+        if not ctx.cleanup:
+            return
+        log_status("enroot: removing containers and imported images")
+        for container in self.containers:
+            run_command(["enroot", "remove", "-f", container], ctx.dry_run, check=False)
+        if not ctx.dry_run:
+            for sqsh in self.images.values():
+                try:
+                    os.unlink(sqsh)
+                except OSError:
```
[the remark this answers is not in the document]
mc-nv: Documented in f28a2a417.

Best-effort cleanup of enroot squashfs images. The image may already be gone, or sit on a read-only mount. Neither is worth failing a run whose actual work has already completed.

qa/common/gen_archive.py
```diff
@@ -0,0 +1,507 @@
...
+        # leave a truncated archive sitting where a valid one belongs -- and
+        # the upload step trusts what it finds on disk.
+        partial = target.with_name(target.name + ".partial")
+        try:
+            with open(partial, "wb") as handle:
+                archive_bytes, sha256 = write_bundle(model_dirs, tree, handle, compress)
+        except OSError as error:
+            try:
+                partial.unlink()
+            except OSError:
```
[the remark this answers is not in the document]
mc-nv: Fixed in 4040f83ec — and this one was self-inflicted: the handler arrived with the streaming archiver in 0784be249, so the alert is new on this branch rather than pre-existing.

Same class as the four above. It tidies up the `.partial` file after a pack that has already failed and is about to raise. If the removal itself fails, the partial stays behind under a name nothing consumes — not worth masking the original packing error with.

qa/common/gen_qa_pytorch_model.py
```diff
@@ -28,6 +28,7 @@
 import argparse
 import os
 
+import gen_manifest
```
dmitry-tokarev-nv, line 31: **Blocking — this breaks `L0_backend_python/python_based_backends` unconditionally.**

`qa/L0_backend_python/python_based_backends/test.sh` copies this generator out of `qa/common` and runs it from its own directory:

```sh
cp ${GEN_PYTORCH_MODEL_PY} ./gen_qa_pytorch_model.py   # test.sh:63
GEN_PYTORCH_MODEL_PY=./gen_qa_pytorch_model.py
```

`gen_manifest.py` is not copied along with it, and `sys.path[0]` is the test directory rather than `qa/common`, so the new import dies before `main()` is ever reached:

```
File "./gen_qa_pytorch_model.py", line 31, in <module>
    import gen_manifest
ModuleNotFoundError: No module named 'gen_manifest'
```

`test.sh` treats a non-zero exit as fatal. This is the only generator copied standalone, so it is the only one affected — but it is an unconditional L0 failure.

No inline suggestion, because the minimal fix lives in a file this PR does not touch:

```diff
 cp ${GEN_PYTORCH_MODEL_PY} ./gen_qa_pytorch_model.py
+cp ../../common/gen_manifest.py ./gen_manifest.py
```

The alternative is to make this import optional and guard the two call sites at L123 and L132.

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1915 @@
...
+        self.nvidia_visible_devices = (
+            args.nvidia_visible_devices or DEFAULT_NVIDIA_VISIBLE_DEVICES
+        )
+
+        # Filled in once the engine is known: the mount root differs, and the
+        # build directory hangs off it.
+        self.runtime_name = None
+        self.build_dir = args.build_dir
+        self.source_dir = None
+        self.docker_gpu_args = resolve_gpu_args(args)
```
dmitry-tokarev-nv, line 1226: **Blocking — the empty-to-default fallback immediately above is bypassed on the docker path.**

L1217 sanitizes an empty value to `DEFAULT_NVIDIA_VISIBLE_DEVICES`, but this line passes the raw `args` namespace, so `resolve_gpu_args()` reads the *un-sanitized* value at L1283:

```
--nvidia-visible-devices ''  ->  ctx.nvidia_visible_devices = 'all'
                                 ctx.docker_gpu_args = ['--runtime=nvidia', '-e', 'NVIDIA_VISIBLE_DEVICES=']
```

Empty means `void` to the container toolkit — no GPUs, runtime behaves as `runc`. So docker gets no GPU while enroot gets `all`, and the PyTorch/TensorRT stages fail minutes in inside torch. That is precisely the accident the comment at L1211-L1216 describes: a job forwarding `--nvidia-visible-devices "${NVIDIA_VISIBLE_DEVICES}"` with the variable never defined by its runner.

`resolve_gpu_args()` only reads `.nvidia_visible_devices`, so passing `self` is sufficient:

```suggestion
        self.docker_gpu_args = resolve_gpu_args(self)
```

qa/common/gen_qa_model_repository.py
```diff
@@ -0,0 +1,1915 @@
...
+        default=None,
+        help="remove containers, volumes and imported images afterwards",
+    )
+    behaviour.add_argument(
+        "--no-cleanup", dest="cleanup", action="store_false", help=argparse.SUPPRESS
+    )
+    behaviour.add_argument(
+        "--clean-build-dir",
+        action="store_true",
+        default=True,
```
dmitry-tokarev-nv, line 1595: **Blocking — this turns on a destructive `rmtree` by default, on a code path that has never actually run.**

`action="store_true"` combined with `default=True` means `clean_build_dir` is True unless `--no-clean-build-dir` is explicitly passed. So `EnrootRuntime.prepare()` at L1028-L1035 *always* `shutil.rmtree`s `ctx.build_dir`, and `build_dir` comes from `TRITON_MDLS_BLD_DIR` when that is set. A SLURM/enroot job pointing it at a shared or pre-populated directory loses its contents.

The comment at L1029-L1032 is explicit that the shell driver intended this but could not do it — bash discards the prefix assignment made to a function call, so the variable is empty by the time its `rm -rf` runs. This is therefore a delete that has never happened in practice, being switched on by default in the same change that ports the driver. Explicit opt-in is the safer default; refusing paths outside the mount root would be safer still.

```suggestion
        default=False,
```

dmitry-tokarev-nv (commented):
Review scoped to critical / blocking issues only; style, naming and latent-but-inactive divergences are left out deliberately.

Five findings, each reproduced against `4040f83` rather than read off the diff:

1. `gen_qa_pytorch_model.py` — the new `import gen_manifest` breaks `L0_backend_python/python_based_backends`, which copies that generator standalone. Unconditional L0 failure.
2. `gen_qa_model_repository.py` — `resolve_gpu_args()` receives the raw namespace instead of `self`, so the empty-to-`all` GPU fallback never reaches the docker args.
3. `gen_qa_model_repository.py` — `--clean-build-dir` defaults to `True`, enabling an `rmtree` of `TRITON_MDLS_BLD_DIR` that the shell driver never actually performed.
4. `gen_qa_model_repository` — the output tree moved from `$TRITON_VERSION` to `$TRITON_CONTAINER_VERSION`; deliberate, but out-of-repo consumers need to move with it or they go silently green.
5. `gen_manifest.py` — `BACKEND_TO_MODULE` is keyed on backend names but looked up with stage keys, so `framework_version` is `null` for pytorch and tensorrt.

The rest checked out: removed dead assignments have no remaining readers, the `gen_ensemble_model_utils` to `gen_common` move leaves no dangling `emu.` references, the four rendered stage scripts match the shell heredocs step-for-step and pass `bash -n`, and `gen_archive.py` round-trips both granularities with reproducible member ordering.

<sub>Automated review (Claude Code), verified by checkout and execution.</sub>

qa/common/gen_qa_model_repository
```diff
@@ -66,14 +66,25 @@ log_message.status "Changing working directory to the script directory to: " "${
 cd ${TRITON_MDLS_BASE_SCRIPT_DIR}
 
 log_message.status "define: default values"
-TRITON_VERSION=${TRITON_VERSION:=26.07}
+# The three versions, matching build.py's DEFAULT_TRITON_VERSION_MAP entries
+# triton_container_version, upstream_container_version and release_version.
+# Bump them here when that map moves. TRITON_VERSION is deliberately gone: it
+# named the container train here and the semver in CI, so reading it was wrong
+# half the time.
+TRITON_CONTAINER_VERSION=${TRITON_CONTAINER_VERSION:=26.08dev}
```
dmitry-tokarev-nv, line 74: **Please confirm the CI side lands together with this.**

This commit moves all 36 `TRITON_CONTAINER_VERSION` references — including the 23 `TRITON_MDLS_QA_*` paths and the closing `docker cp ... /tmp/` — off `$TRITON_VERSION`. The rationale in L71-L73 is clear and I am not disputing it; the concern is the rollout. Nothing in this repository reads those paths: the consumers are the GitLab CI jobs. Any job still computing the model-repository path from `TRITON_VERSION` will `cd` into a directory that does not exist, and because docker and the shell create missing directories on the way, that surfaces as an empty loop and a **green** job — the exact silent-success failure mode this PR is meant to eliminate.

Separately, the `26.08dev` default here does not match `build.py`'s `triton_container_version`, which is `26.09dev` at this SHA — so the "matching build.py" note at L69-L70 is already stale.

qa/common/gen_manifest.py
```diff
@@ -0,0 +1,1424 @@
...
+    ".xml": "openvino",
+}
+
+# The importable module whose __version__ describes each backend's artifacts.
+BACKEND_TO_MODULE = {
+    "onnx": "onnx",
+    "openvino": "openvino",
+    "libtorch": "torch",
+    "plan": "tensorrt",
+}
```
dmitry-tokarev-nv, line 153: **`model.framework_version` is always `null` for the pytorch and tensorrt stages.**

This map is keyed on the *backend* vocabulary, but `probe_framework_version()` is called with the *stage key*. `stage_environment()` in `gen_qa_model_repository.py:776` exports `TRITON_MODEL_GEN_FRAMEWORK=<stage.key>`, and those keys are `openvino | onnx | pytorch | tensorrt`. Neither `pytorch` nor `tensorrt` is in this dict, so `.get()` returns `None` and L789-L790 bail out before importing anything.

The effect is that the two stages whose library version matters most — the torch release and the TensorRT release — record `model.framework_version: null`, and `_log_properties` prints `framework version -`. The old shell driver, which sets no such variable, resolves per model via `config.provenance` and gets this right.

```suggestion
BACKEND_TO_MODULE = {
    "onnx": "onnx",
    "openvino": "openvino",
    "libtorch": "torch",
    "plan": "tensorrt",
    # Stage keys: TRITON_MODEL_GEN_FRAMEWORK carries the framework vocabulary
    # rather than the backend one, and probe_framework_version() sees both.
    "pytorch": "torch",
    "tensorrt": "tensorrt",
}
```

qa/common/gen_qa_model_repository.py
mc-nv: those args are intentionally as we are getting set of pre-defined vars in our CI

qa/common/gen_qa_model_repository.py
mc-nv: our ci cleans the directory I don't see reason why we need to clean up the build dir by defult or cusomer

qa/common/gen_qa_model_repository
mc-nv: Not following this comment, it doesn't involves build.py

qa/common/gen_manifest.py
mc-nv: redundant comment as long as we don't use those keys in CI

qa/common/gen_qa_pytorch_model.py
mc-nv: I would rather look into it after the change, intentionally left it behind.

qa/common/gen_qa_model_repository.py
dmitry-tokarev-nv: discussed on a call - agreeed

qa/common/gen_qa_pytorch_model.py
dmitry-tokarev-nv: sounds good

qa/common/gen_qa_model_repository
dmitry-tokarev-nv: discussed on a call, this is as designed.

dmitry-tokarev-nv (approved):
Discussed on a call all concerns were addressed.

PR #8939, opened 2026-08-26
build: Add build secrets and authenticate GitHub clones

mc-nv:
#### What does the PR do?
Adds a build secret mechanism to `build.py` and uses it to authenticate the
GitHub clones the build performs and to route package installs through a
package mirror.

`--docker-build-secret` takes a Docker build secret spec and forwards it to
`docker build --secret` unchanged, so every form Docker accepts works:
`id=<id>`, `id=<id>,src=<path>` and `id=<id>,env=<var>`. A spec may also carry
`target=<path>`; Docker rejects that key on the command line because it belongs
to the Dockerfile, so `build.py` strips it from the build command and applies it
when generating the Dockerfile.

Two ids carry meaning. `apt_sources` is mounted over the apt source list for the
duration of every apt step in the generated `Dockerfile` and
`Dockerfile.buildbase`, and on the corresponding steps of `Dockerfile.sdk` and
`Dockerfile.QA`. `GITHUB_TOKEN` is an environment variable rather than a build
secret: the component and backend repos are cloned by `cmake_build` while the
build runs inside the container, not while an image is built, so a build secret
cannot reach them. `build.py` configures a git credential helper in the build
base and forwards the token to the build container instead. The helper holds no
credential of its own, only a reference, so it is safe in a layer and in the
pushed build base image, and it keeps the token out of `.git/config`.

Every mount is `required=false` and the helper prints nothing when the token is
absent, so a build without either secret behaves exactly as it does today: the
generated Dockerfiles are byte-identical and clones stay unauthenticated.

#### Checklist

#### Commit Type:
- [x] build
- [ ] ci
- [x] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

A follow-up PR stacked on this branch bumps the fastapi and starlette pins in
the openai frontend; it is listed here once opened.

`build.py` — `secret_spec_id`, `parse_secret_spec`, `secret_build_args`,
`apt_sources_secret_mount` and `mount_apt_sources_secret`, then the two
`docker build` argument lists and the `docker run` invocation that forwards
`GITHUB_TOKEN`.

#### Test plan:
Verified with `--dryrun`, by inspecting the generated build scripts, and with
real BuildKit builds.

- `./build.py --enable-all --dryrun` with no secret: generated Dockerfiles are
  byte-identical to a build without this feature, and no `--secret` reaches
  `docker build`.
- `./build.py --enable-all --dryrun --docker-build-secret id=apt_sources,src=<file>`:
  both `docker build` invocations carry `--secret`, and every apt step of
  `Dockerfile` and `Dockerfile.buildbase` carries the matching mount.
- `target=` is absent from the build command and present on the Dockerfile mount.
- Real BuildKit build with an apt source list mounted: packages resolve through
  the mirror, and the secret is absent from the image, from `docker history`,
  from `docker inspect` and from the exported filesystem.
- Real BuildKit build with `GITHUB_TOKEN`: an authenticated clone succeeds, the
  token does not appear in `.git/config`, and an unauthenticated build still
  clones a public repo successfully.

- CI Pipeline ID:

#### Caveats:
The apt source list carries credentials, which is why it is passed as a secret
and mounted per step rather than copied in. On a base image without CA
certificates the first apt step of `Dockerfile.buildbase` reports a failed TLS
handshake, still exits 0, and installs from the distribution repositories, so
that step behaves as it did before the mount was added.

#### Background
The build clones several repositories from GitHub and installs a large number of
packages. Both were unauthenticated, which is subject to rate limiting and
cannot reach a private repository or an authenticated mirror.

- Resolves: TRI-1736

### DIFF
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -124,18 +124,9 @@ WORKDIR /workspace
 COPY TRITON_VERSION .
 COPY ${TRITON_CLIENT_REPO_SUBDIR} client
 
-# Authenticate the GitHub clones cmake performs for the component repos. The
-# helper stores no credential, only a reference to GITHUB_TOKEN, so it is safe
-# in a layer and in the pushed image. Each step that clones supplies the value
-# with --mount=type=secret,id=github_token. The helper prints nothing when the
-# token is absent, leaving git to clone anonymously as before.
-RUN git config --global credential."https://github.com".helper \
-      '!f() { test -n "${GITHUB_TOKEN}" || exit 0; \
-              echo username=x-access-token; \
-              echo "password=${GITHUB_TOKEN}"; }; f'
-
 WORKDIR /workspace/client_build
-RUN --mount=type=secret,id=github_token,env=GITHUB_TOKEN,required=false \
+RUN --mount=type=secret,id=gitconfig,target=/run/secrets/gitconfig,required=false \
+    export GIT_CONFIG_GLOBAL=/run/secrets/gitconfig && \
     cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
           -DTRITON_VERSION=`cat /workspace/TRITON_VERSION` \
           -DTRITON_REPO_ORGANIZATION=${TRITON_REPO_ORGANIZATION} \
@@ -149,13 +140,15 @@ RUN --mount=type=secret,id=github_token,env=GITHUB_TOKEN,required=false \
           -DTRITON_ENABLE_EXAMPLES=ON -DTRITON_ENABLE_TESTS=ON \
           -DTRITON_ENABLE_GPU=${TRITON_ENABLE_GPU} /workspace/client
 RUN --mount=type=secret,id=maven_settings,target=/run/secrets/maven_settings,required=false \
-    --mount=type=secret,id=github_token,env=GITHUB_TOKEN,required=false \
+    --mount=type=secret,id=gitconfig,target=/run/secrets/gitconfig,required=false \
+    export GIT_CONFIG_GLOBAL=/run/secrets/gitconfig && \
     if [ -f /run/secrets/maven_settings ]; then export MAVEN_ARGS="--settings /run/secrets/maven_settings"; fi && \
     cmake --build . -v --parallel ${TRITON_CLIENT_BUILD_PARALLEL} --target cc-clients java-clients python-clients
 
 # Install Java API Bindings
 RUN --mount=type=secret,id=maven_settings,target=/run/secrets/maven_settings,required=false \
-    --mount=type=secret,id=github_token,env=GITHUB_TOKEN,required=false \
+    --mount=type=secret,id=gitconfig,target=/run/secrets/gitconfig,required=false \
+    export GIT_CONFIG_GLOBAL=/run/secrets/gitconfig && \
     if [ -f /run/secrets/maven_settings ]; then export MAVEN_ARGS="--settings /run/secrets/maven_settings"; fi && \
     if [ "$TARGETPLATFORM" = "linux/amd64" ]; then \
         source /workspace/client/src/java-api-bindings/scripts/install_dependencies_and_build.sh \
@@ -271,13 +264,8 @@ RUN rm -f /usr/bin/python && \
 # Install Model Analyzer
 ARG TRITON_MODEL_ANALYZER_REPO_TAG
 ARG TRITON_MODEL_ANALYZER_REPO="${TRITON_REPO_ORGANIZATION}/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}"
-# Separate stage from the client build, so the credential helper is configured
-# again here. It stores no credential, only a reference to GITHUB_TOKEN.
-RUN git config --global credential."https://github.com".helper \
-      '!f() { test -n "${GITHUB_TOKEN}" || exit 0; \
-              echo username=x-access-token; \
-              echo "password=${GITHUB_TOKEN}"; }; f'
-RUN --mount=type=secret,id=github_token,env=GITHUB_TOKEN,required=false \
+RUN --mount=type=secret,id=gitconfig,target=/run/secrets/gitconfig,required=false \
+    export GIT_CONFIG_GLOBAL=/run/secrets/gitconfig && \
     pip3 install "git+${TRITON_MODEL_ANALYZER_REPO}"
 
 # Entrypoint Banner
--- a/Dockerfile.sdk
+++ b/Dockerfile.sdk
@@ -124,8 +124,19 @@ WORKDIR /workspace
 COPY TRITON_VERSION .
 COPY ${TRITON_CLIENT_REPO_SUBDIR} client
 
+# Authenticate the GitHub clones cmake performs for the component repos. The
+# helper stores no credential, only a reference to GITHUB_TOKEN, so it is safe
+# in a layer and in the pushed image. Each step that clones supplies the value
+# with --mount=type=secret,id=github_token. The helper prints nothing when the
+# token is absent, leaving git to clone anonymously as before.
+RUN git config --global credential."https://github.com".helper \
+      '!f() { test -n "${GITHUB_TOKEN}" || exit 0; \
+              echo username=x-access-token; \
+              echo "password=${GITHUB_TOKEN}"; }; f'
+
 WORKDIR /workspace/client_build
-RUN cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
+RUN --mount=type=secret,id=github_token,env=GITHUB_TOKEN,required=false \
+    cmake -DCMAKE_INSTALL_PREFIX=/workspace/install \
           -DTRITON_VERSION=`cat /workspace/TRITON_VERSION` \
           -DTRITON_REPO_ORGANIZATION=${TRITON_REPO_ORGANIZATION} \
           -DTRITON_COMMON_REPO_TAG=${TRITON_COMMON_REPO_TAG} \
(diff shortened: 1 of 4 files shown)


build.py
```diff
@@ -2623,15 +2755,25 @@ def enable_all():
         help="This flag sets the Python version for RHEL platform of Triton Inference Server to be built. Default: the latest supported version.",
     )
     parser.add_argument(
-        "--build-secret",
+        "--docker-build-secret",
```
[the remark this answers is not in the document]
mc-nv: It's addressed in complement MR

build.py
```diff
@@ -1122,6 +1231,23 @@ def create_dockerfile_buildbase(ddir, dockerfile_name, argmap):
...
+    # being built, so a docker build secret cannot reach it. Configure a
+    # credential helper instead: it holds no credential itself, only a
+    # reference to GITHUB_TOKEN, so it is safe in a layer and in the pushed
+    # build base image. The value is supplied by 'docker run -e GITHUB_TOKEN'.
+    #
+    # The helper exits without printing anything when GITHUB_TOKEN is empty,
+    # which leaves git to fall back to unauthenticated access rather than
+    # offering an empty password and failing a clone that used to work.
+    df += """
+RUN git config --global credential."https://github.com".helper \\
```
dmitry-tokarev-nv: **Blocking — this mechanism cannot fix the rate-limiting problem the PR targets.**

Git only invokes a credential helper *after* the server answers `401`. Cloning a **public** repo over HTTPS returns `200` with no auth challenge, so the helper is never consulted and `GITHUB_TOKEN` is never sent. Every default component repo (`common`, `core`, `backend`, `thirdparty`, and the backends) is public, so a build on a shared CI egress IP still clones anonymously and still hits GitHub's unauthenticated limits. Only private-repo clones — which do get a `401` — actually benefit.

To authenticate public clones you have to force the credential onto the request rather than wait to be asked:

```
RUN git config --global url."https://x-access-token:${GITHUB_TOKEN}@github.com/".insteadOf "https://github.com/"
```

(or an `http.extraHeader` carrying the bearer token). Worth verifying with a helper that logs its own invocation before merging — `git ls-remote` against any public repo will show the helper is never called.

build.py
```diff
@@ -878,6 +878,115 @@ def install_dcgm_libraries(dcgm_version, target_machine):
...
+    the step resolves through the mirror like any other. On one that does not,
+    apt reports the failed TLS handshake, still exits 0, and installs from the
+    distribution repositories, so the step behaves as it did before the mount
+    was added.
+    """
+    mount = apt_sources_secret_mount()
+    if not mount:
+        return df
+
+    marker = "RUN apt-get update"
```
dmitry-tokarev-nv, line 993: **Blocking — the DCGM step never receives the mount, so a GPU build fails on a mirror-only network.**

This matches only the literal string `RUN apt-get update`. `install_dcgm_libraries()` emits `apt install` / `apt update -qq`, never `RUN apt-get update`, so it is silently skipped. In the generated `build/Dockerfile.buildbase` and `build/Dockerfile` the DCGM step runs `apt update -qq && apt install ... datacenter-gpu-manager-4-core` with no `--mount=type=secret`. On the exact network this flag exists for — only the authenticated mirror reachable — `--enable-gpu` dies there.

The flag help and `docs/customization_guide/build.md` both claim *every* apt step is covered; today it's only the ones spelled `apt-get update`.

Suggest matching both spellings, e.g. a regex over `RUN apt(-get)? update`, and asserting the substitution count is non-zero so a future rename fails loudly instead of silently.

Dockerfile.sdk
```diff
@@ -124,8 +124,19 @@ WORKDIR /workspace
 COPY TRITON_VERSION .
 COPY ${TRITON_CLIENT_REPO_SUBDIR} client
 
+# Authenticate the GitHub clones cmake performs for the component repos. The
+# helper stores no credential, only a reference to GITHUB_TOKEN, so it is safe
+# in a layer and in the pushed image. Each step that clones supplies the value
+# with --mount=type=secret,id=github_token. The helper prints nothing when the
+# token is absent, leaving git to clone anonymously as before.
+RUN git config --global credential."https://github.com".helper \
```
dmitry-tokarev-nv: **Severe — protocol mismatch means this helper is never offered.**

The helper is registered under `credential."https://github.com"`, but `TRITON_REPO_ORGANIZATION` defaults to `http://github.com/triton-inference-server` (line 35; `Dockerfile.QA:30` is the same). Git matches credential sections on **protocol as well as host** — `git credential fill` with `protocol=https,host=github.com` fires the helper, `protocol=http,host=github.com` does not.

So with the default ARG, the cmake clones and `pip3 install "git+http://github.com/.../model_analyzer@..."` (line 280) request `protocol=http` and never match this section. The token is only reachable via GitHub's http→https redirect *plus* a 401.

Either bump the ARG defaults to `https://`, or register the helper for both protocols.

Dockerfile.sdk
```diff
@@ -167,7 +180,8 @@ ARG TRITON_CORE_REPO_TAG
 ARG TARGETPLATFORM
 ARG TRITON_ENABLE_GPU
 
-RUN apt-get update && \
+RUN --mount=type=secret,id=apt_sources,target=/etc/apt/sources.list.d/nvidia-artifactory-ubuntu.list,required=false \
```
dmitry-tokarev-nv, line 176: **Blocking for the SDK image — this is the only apt step that got the mount.**

The `sdk_build` stage's apt step (line 64) and the DCGM install in this runtime stage (line 256) have no `apt_sources` mount. On a mirror-only network `docker build -f Dockerfile.sdk` fails on the *first* stage, before ever reaching this line.

Same fix as `build.py`'s `mount_apt_sources_secret`: add `--mount=type=secret,id=apt_sources,target=/etc/apt/sources.list.d/nvidia-artifactory-ubuntu.list,required=false` to lines 64 and 256.

Dockerfile.sdk
```diff
@@ -257,7 +271,14 @@ RUN rm -f /usr/bin/python && \
 # Install Model Analyzer
 ARG TRITON_MODEL_ANALYZER_REPO_TAG
 ARG TRITON_MODEL_ANALYZER_REPO="${TRITON_REPO_ORGANIZATION}/model_analyzer@${TRITON_MODEL_ANALYZER_REPO_TAG}"
-RUN pip3 install "git+${TRITON_MODEL_ANALYZER_REPO}"
+# Separate stage from the client build, so the credential helper is configured
+# again here. It stores no credential, only a reference to GITHUB_TOKEN.
+RUN git config --global credential."https://github.com".helper \
```
dmitry-tokarev-nv: **Severe — this credential helper ships in the released SDK image.**

This is a standalone `RUN` in the final (published) stage, so `/root/.gitconfig` persists into the shipped image carrying a helper that unconditionally forwards whatever `GITHUB_TOKEN` is in the container's environment to github.com. A user running the SDK container in a CI job that exports `GITHUB_TOKEN` for an unrelated purpose will have it sent on any github.com clone.

Unlike the build-base copy (`build.py:1245`), which genuinely must persist for `cmake_build`, this one only needs to exist for the very next step. Folding it into the same `RUN` as the `pip3 install` on line 280 keeps it out of the shipped layer entirely.

build.py
```diff
@@ -1729,6 +1861,13 @@ def create_docker_build_script(script_name, container_install_dir, container_ci_
         # or explicit --release-version). Dev / pre-release builds leave it
         # unset so build_wheel.py reads the in-tree TRITON_VERSION file and
         # takes the PEP 817 variant path.
+        # Forward the GitHub token to the clones cmake_build performs in this
+        # container. Passed by name rather than as NAME=value: 'docker run -e
+        # VAR' takes the value from the ambient environment, so the token stays
+        # out of this script, which CI publishes as a build artifact.
+        if os.environ.get("GITHUB_TOKEN"):
```
dmitry-tokarev-nv: **Severe — this bakes the *generation-time* environment into a script that runs later.**

The comment above correctly explains that `docker run -e VAR` (name only) keeps the token out of the published artifact. But this `if` defeats that goal: whether the `-e GITHUB_TOKEN` line is emitted at all is decided when `build.py` runs, not when `build/docker_build` runs.

The PR body says this script is published as a CI artifact and executed in a later job. In any pipeline where the token is injected only into the *executing* job, the line is absent, the container never sees the token, and clones silently run unauthenticated (or fail outright on a private repo).

`docker run -e VAR` with `VAR` unset is a no-op, so the guard buys nothing:

```python
runargs += ["-e", "GITHUB_TOKEN"]
```

dmitry-tokarev-nv (commented):
Review of the current head (`133660f7`). Posting only the blocking / severe items; several lower-severity ones (dead `declared_secret_ids()`, silent no-op on a typo'd secret id, `target=` honored only for `apt_sources`) are left out.

The headline concern is the first comment on `build.py:1245`: **the credential-helper design cannot fix the GitHub rate-limiting problem the PR is built to solve**, because git never consults a helper for a public-repo clone. That one determines whether the PR achieves its stated goal, so it's worth resolving before the rest.

Two more items that don't sit on a changed line:

- **`create_dockerfile_buildbase_rhel` gets neither the credential helper nor the apt-sources mount.** `mount_apt_sources_secret()` is called only from `create_dockerfile_buildbase` and `create_dockerfile_linux`, and the helper `RUN` is added only in the former — yet `create_docker_build_script` still forwards `-e GITHUB_TOKEN` for every platform. `--target-platform rhel` therefore clones anonymously with no warning. The same gap applies to `--no-container-build`, where `cmake_build` runs on the host with no helper configured, while the docs say only "Set `GITHUB_TOKEN` in the environment".

- **Two claims in the PR description don't hold as written.** Generated Dockerfiles are *not* byte-identical when no secret is supplied — the credential-helper `RUN` at `build.py:1245` is emitted unconditionally. And "every apt step" is not covered (see the `build.py:986` and `Dockerfile.sdk:183` comments).

Separately, on the `--build-secret` → `--docker-build-secret` rename already raised by the bot and answered with "addressed in complement MR": the same hunk *also* drops `--secret id=req/VLLM_INDEX_URL/PYTORCH_TRITON_URL/NVPL_SLIM_URL` and `--build-arg BUILD_PUBLIC_VLLM` from the final image build. Nothing in-tree consumes those, but out-of-tree vLLM build flows will break — worth confirming the companion MR covers that removal too, and calling both out in the commit message / release notes since neither appears in the PR description.

(Fixes are given as plain code blocks rather than `suggestion` blocks — most span multiple lines or belong on lines this PR doesn't touch.)

build.py
whoisj, line 266: is this something to actually be concerned about?

whoisj (approved):
LGTM. I only had the one question (above).

PR #8948, opened 2026-09-01
test: add torch_aoti smoke coverage to L0_infer

Vinya567:
#### What does the PR do?
Adds a light torch AOTI pass to L0_infer: copy the existing `torch_aoti_float32_float32` add model in the GPU phase and infer `ARGS[0]`/`ARGS[1]` → `RESULT` over HTTP and gRPC. AOTI I/O names are not libtorch-shaped, so this does not reuse `infer_exact`. Nested/sequence/negative AOTI cases stay in `L0_torch_aoti`.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:
N/A

`qa/L0_infer/test.sh` (GPU-only copy + expected test counts), then `qa/L0_infer/infer_test.py` (`test_torch_aoti_float32`).

#### Test plan:
- Local: generated `torch_aoti_float32_float32`, started tritonserver, ran `python3 -m unittest infer_test.InferTest.test_torch_aoti_float32` — **ok** (HTTP and gRPC).
- Full `L0_infer/test.sh` against QA DATADIR: GitLab CI after this PR.

- CI Pipeline ID:[#65592447]
GPU-only (packaged `.pt2` is CUDA). The new test skips if that model is not loaded. Expected test counts bump by one; CPU-only and missing-DATADIR phases still skip.

#### Background
L0_infer is the shared backend matrix (including Dynamo `SERVER_LAUNCH_MODE=dynamo`). It never loaded a torch AOTI model. This is one existing add model, not a rewrite of `L0_torch_aoti`.

- N/A

### DIFF
--- a/qa/L0_infer/infer_test.py
+++ b/qa/L0_infer/infer_test.py
@@ -36,6 +36,8 @@ import unittest
 import infer_util as iu
 import numpy as np
 import test_util as tu
+import tritonclient.grpc as grpcclient
+import tritonclient.http as httpclient
 from tritonclient.utils import *
 
 TEST_SYSTEM_SHARED_MEMORY = bool(int(os.environ.get("TEST_SYSTEM_SHARED_MEMORY", 0)))
@@ -1181,6 +1183,66 @@ class InferTest(tu.TestResultCollector):
                                 use_system_shared_memory=TEST_SYSTEM_SHARED_MEMORY,
                                 use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY,
                             )
+
+    def test_torch_aoti_float32(self):
+        # Light AOTI coverage: one existing add model (ARGS[0]/ARGS[1] -> RESULT).
+        # Not libtorch-shaped; infer_exact cannot be reused.
+        model_name = "torch_aoti_float32_float32"
+        input0 = np.ones((1, 16), dtype=np.float32)
+        input1 = np.full((1, 16), 2.0, dtype=np.float32)
+        expected = input0 + input1
+
+        def _is_ready_http():
+            with httpclient.InferenceServerClient("localhost:8000") as client:
+                return client.is_model_ready(model_name)
+
+        def _is_ready_grpc():
+            with grpcclient.InferenceServerClient("localhost:8001") as client:
+                return client.is_model_ready(model_name)
+
+        ready = False
+        if USE_HTTP:
+            ready = _is_ready_http()
+        elif USE_GRPC:
+            ready = _is_ready_grpc()
+        if not ready:
+            self.skipTest(f"{model_name} not loaded in this L0_infer phase")
+
+        if USE_HTTP:
+            with httpclient.InferenceServerClient("localhost:8000") as client:
+                inputs = [
+                    httpclient.InferInput("ARGS[0]", input0.shape, "FP32"),
+                    httpclient.InferInput("ARGS[1]", input1.shape, "FP32"),
+                ]
+                inputs[0].set_data_from_numpy(input0)
+                inputs[1].set_data_from_numpy(input1)
+                result = client.infer(
+                    model_name,
+                    inputs,
+                    outputs=[httpclient.InferRequestedOutput("RESULT")],
+                )
+                self.assertTrue(
+                    np.allclose(result.as_numpy("RESULT"), expected),
+                    "torch_aoti HTTP RESULT mismatch",
+                )
+
+        if USE_GRPC:
+            with grpcclient.InferenceServerClient("localhost:8001") as client:
+                inputs = [
+                    grpcclient.InferInput("ARGS[0]", input0.shape, "FP32"),
+                    grpcclient.InferInput("ARGS[1]", input1.shape, "FP32"),
+                ]
+                inputs[0].set_data_from_numpy(input0)
+                inputs[1].set_data_from_numpy(input1)
+                result = client.infer(
+                    model_name,
+                    inputs,
+                    outputs=[grpcclient.InferRequestedOutput("RESULT")],
+                )
+                self.assertTrue(
+                    np.allclose(result.as_numpy("RESULT"), expected),
+                    "torch_aoti gRPC RESULT mismatch",
+                )
 
 
 if __name__ == "__main__":
--- a/qa/L0_infer/infer_test.py
+++ b/qa/L0_infer/infer_test.py
@@ -30,15 +30,15 @@ import sys
 
 sys.path.append("../common")
 
-import os
-import unittest
+import os  # noqa: E402
+import unittest  # noqa: E402
 
-import infer_util as iu
-import numpy as np
-import test_util as tu
-import tritonclient.grpc as grpcclient
-import tritonclient.http as httpclient
-from tritonclient.utils import *
+import infer_util as iu  # noqa: E402
+import numpy as np  # noqa: E402
+import test_util as tu  # noqa: E402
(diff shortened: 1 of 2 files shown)


qa/L0_infer/test.sh
```diff
@@ -320,6 +320,12 @@ for TARGET in cpu gpu; do
 
     generate_model_repository
 
+    # Light torch_aoti coverage for Dynamo/L0_infer: one existing add model.
+    # GPU-only — the packaged .pt2 is CUDA. Skip CPU target so the server still starts.
```
yinggeh:
```suggestion
```

qa/L0_infer/test.sh
```diff
@@ -320,6 +320,12 @@ for TARGET in cpu gpu; do
...
+    if [ "$TARGET" == "gpu" ] && [ -d "${DATADIR}/qa_model_repository/torch_aoti_float32_float32" ]; then
```
yinggeh:
```suggestion
    if [ "$TARGET" == "gpu" ]; then
```

yinggeh: Need to assert the model exists.

qa/L0_infer/test.sh
```diff
@@ -392,6 +398,10 @@ if [ "$TEST_VALGRIND" -eq 1 ]; then
         cp -fr ./models/onnx_float32_float32_float32 ./nobatch_models/.
         cp -fr ./models/custom_zero_1_float32 ./nobatch_models/.
       fi
+      if [ "$TARGET" == "gpu" ] && [ -d "${DATADIR}/qa_model_repository/torch_aoti_float32_float32" ]; then
```
yinggeh:
```suggestion
      if [ "$TARGET" == "gpu" ]; then
```

qa/L0_infer/test.sh
```diff
@@ -73,9 +73,9 @@ if [ "$TEST_VALGRIND" -eq 1 ]; then
 fi
 
 if [ "$TEST_SYSTEM_SHARED_MEMORY" -eq 1 ] || [ "$TEST_CUDA_SHARED_MEMORY" -eq 1 ]; then
-    EXPECTED_NUM_TESTS=${EXPECTED_NUM_TESTS:="33"}
+    EXPECTED_NUM_TESTS=${EXPECTED_NUM_TESTS:="34"}
 else
-    EXPECTED_NUM_TESTS=${EXPECTED_NUM_TESTS:="46"}
+    EXPECTED_NUM_TESTS=${EXPECTED_NUM_TESTS:="47"}
```
mc-nv, line 78: These values we also control with CI variables,  it something  we need to pay attention in this change.

whoisj (dismissed):

qa/L0_infer/infer_test.py
```diff
@@ -1182,6 +1184,66 @@ def test_ensemble_label_lookup(self):
...
+        input0 = np.ones((1, 16), dtype=np.float32)
+        input1 = np.full((1, 16), 2.0, dtype=np.float32)
+        expected = input0 + input1
+
+        def _is_ready_http():
+            with httpclient.InferenceServerClient("localhost:8000") as client:
+                return client.is_model_ready(model_name)
+
+        def _is_ready_grpc():
+            with grpcclient.InferenceServerClient("localhost:8001") as client:
```
mc-nv: I'm not certain about hardcoded values for `localhost`, and reason for is that test run Docker-out-of-Docker. 
For sure it may work but because we don't specify network settings and passing docker engine socket downstream to execution container this value may route packages in unexpected way and introduce instability.

qa/L0_infer/test.sh
```diff
@@ -372,7 +376,7 @@ done
 # separately to reduce the loading time.
 if [ "$TEST_VALGRIND" -eq 1 ]; then
   TESTING_BACKENDS="python python_dlpack onnx"
-  EXPECTED_NUM_TESTS=36
+  EXPECTED_NUM_TESTS=37
```
mc-nv: This is a strict hardcode, and I’m not sure if it’s something we should tolerate.

qa/L0_infer/test.sh
Vinya567: yeah the default only applies if CI leaves the var empty. jobs that pin EXPECTED_NUM_TESTS in GitLab wont get the +1 and the new test still counts even if it skips. is bumping those GitLab pins the right way or would you rather the test.sh default always win?

qa/L0_infer/infer_test.py
Vinya567: fair. this is QA only, same as the rest of L0. I’ll use TRITONSERVER_IPADDR (default localhost) like infer_util. ports stay 8000/8001.

qa/L0_infer/test.sh
Vinya567: that overwrite was already in test.sh (36) while GitLab still says 42. this PR only moved it 36 to 37 for the extra test. should we keep overwriting in test.sh or honor the GitLab pin and update that instead?

qa/L0_infer/test.sh
mc-nv: I only point you tot he potential issue, you can decide on your own what is the best solution would be.

qa/L0_infer/test.sh
mc-nv: If valgrind tests uses `EXPECTED_NUM_TESTS` variable in definition this value should be changed. 
But by any mean my suggestion was the following unless there is an overlap with the same variable for other scenarios: 
```suggestion
  EXPECTED_NUM_TESTS=${EXPECTED_NUM_TESTS:="37"} 
```

Vinya567: yep, that’s the overlap called out.

same var, two counts: first loop is 43 with aoti (GitLab 42), this python/onnx split is 37 because VALGRIND_TESTS=1 drops the version tests. :=37 won’t assign (already 47 or 42) so this loop would fail. keeping the overwrite. GitLab should be 42 → 43, not 37.??

qa/L0_infer/infer_test.py
```diff
@@ -1182,6 +1184,66 @@ def test_ensemble_label_lookup(self):
                                 use_cuda_shared_memory=TEST_CUDA_SHARED_MEMORY,
                             )
 
+    def test_torch_aoti_float32(self):
```
yinggeh, line 1189: Why did you only test single dtype instead of following the convention for other backends, e.g. test_raw_hhh, test_raw_bbb, test_raw_iii?

Vinya567: this was a single-model smoke for L0_infer. AOTI doesn’t share infer_exact (ARGS[0]/RESULT), and L0_torch_aoti already covers the other dtypes. I can add float16 as a second GPU smoke if you want more here.

qa/L0_infer/test.sh
mc-nv: What we going to do with this overlap, probably better to have it somehow guarded. 
If it's not a big deal we can introduce new var.

mc-nv: It's not blocking comment feel free to resolve

mc-nv (approved):

qa/L0_infer/test.sh
Vinya567: done, 24a180d0

PR #8951, opened 2026-09-02
fix: skip tritonclient.http import in dynamo L0_infer

Vinya567:
#### What does the PR do?
Stop importing `tritonclient.http` at L0_infer module load. Import it only inside `USE_HTTP` paths in the AOTI smoke. gRPC-only images (Dynamo: `tritonclient[grpc]`, no gevent) can collect tests; native Triton still runs HTTP when `USE_HTTP` is on.

#### Checklist

#### Commit Type:
- [x] fix

#### Related PRs:
Follow-up to https://github.com/triton-inference-server/server/pull/8948

`qa/L0_infer/infer_test.py` — no module-level HTTP import; lazy import in `_is_ready_http` and the HTTP infer block.

#### Test plan:
- Native Triton L0_infer still uses HTTP when `USE_HTTP` is set.
- With `SERVER_LAUNCH_MODE=dynamo` (`USE_HTTP=False`), `infer_test.py` imports without gevent.
- Re-run Dynamo Post-Merge CI `triton-runtime / QA Test` after merge.

#### Caveats:
Does not add HTTP extras to the Dynamo image. Dynamo still uses gRPC only.

#### Background
#8948 imported HTTP at module load. Dynamo QA failed with `ModuleNotFoundError: No module named 'gevent'` before any test ran.

- Relates to #8948

### DIFF
--- a/qa/L0_infer/infer_test.py
+++ b/qa/L0_infer/infer_test.py
@@ -37,8 +37,14 @@ import infer_util as iu  # noqa: E402
 import numpy as np  # noqa: E402
 import test_util as tu  # noqa: E402
 import tritonclient.grpc as grpcclient  # noqa: E402
-import tritonclient.http as httpclient  # noqa: E402
 from tritonclient.utils import InferenceServerException  # noqa: E402
+
+# Dynamo images ship gRPC-only tritonclient (no gevent/http extra). Match
+# infer_util: skip the HTTP import so L0_infer can load in dynamo mode.
+if os.environ.get("SERVER_LAUNCH_MODE") == "dynamo":
+    httpclient = None
+else:
+    import tritonclient.http as httpclient  # noqa: E402
 
 TEST_SYSTEM_SHARED_MEMORY = bool(int(os.environ.get("TEST_SYSTEM_SHARED_MEMORY", 0)))
 TEST_CUDA_SHARED_MEMORY = bool(int(os.environ.get("TEST_CUDA_SHARED_MEMORY", 0)))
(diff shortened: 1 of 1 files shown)


qa/L0_infer/infer_test.py
```diff
@@ -37,7 +37,6 @@
 import numpy as np  # noqa: E402
 import test_util as tu  # noqa: E402
 import tritonclient.grpc as grpcclient  # noqa: E402
-import tritonclient.http as httpclient  # noqa: E402
```
whoisj, line 40: why is this necessary? 😕

Vinya567: Dynamo’s image is tritonclient[grpc] only (no gevent). This import ran at module load after #8948 and QA died before any test. Native Triton still uses HTTP ,  we just import it inside the USE_HTTP paths. Dynamo already sets USE_HTTP=False so it never hits those.

whoisj (approved):

yinggeh:
I recommend using `DYNAMO = os.environ.get("SERVER_LAUNCH_MODE") == "dynamo"` to check instead of import at multiple locations in the file.

Vinya567:
yes will file a follow-up for this , gate the HTTP import on DYNAMO like infer_util instead of importing in two places. also need to add the extra AOTI dtypes you suggested (beyond the one float32 smoke). I’ll put both on that ticket.

yinggeh:
> yes will file a follow-up for this , gate the HTTP import on DYNAMO like infer_util instead of importing in two places. also need to add the extra AOTI dtypes you suggested (beyond the one float32 smoke). I’ll put both on that ticket.

Thanks. Also, can you move new AOTI tests after `def test_raw_ioi(self):` and before `if not (TEST_SYSTEM_SHARED_MEMORY or TEST_CUDA_SHARED_MEMORY):`?

Vinya567:
yep will move the AOTI tests to right after test_raw_ioi (before the shm class-output skip). extra dtypes + DYNAMO HTTP import in that same follow-up.

PR #8936, opened 2026-08-20
fix: graceful UNAVAILABLE instead of NULL-deref crash on shared-memory requests (TRI-1698)

akhilraj9:
#### What does the PR do?
When a frontend is started through the in-process Python bindings (`tritonfrontend`),
`HTTPAPIServer`/gRPC are constructed with a **null `SharedMemoryManager`** (TODO DLIS-7194).
An unauthenticated client could crash the whole server process with a single inference
(or shared-memory control) request referencing shared memory, because the request handlers
dereferenced `shm_manager_` with no null check (CWE-476 / remote DoS).

This guards every untrusted-input-reachable `shm_manager_` dereference and returns
`TRITONSERVER_ERROR_UNAVAILABLE` (HTTP 503 / gRPC UNAVAILABLE) instead of dereferencing.
No behavior change on the standard `tritonserver` binary path (manager is non-null there).

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [ ] Populated github labels field
- [x] Added test plan and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added succinct git squash message before merging.
- [x] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
- [x] fix

#### Related PRs:
None.

#### Where should the reviewer start?
- `src/http_server.cc` — `ParseJsonTritonIO` (infer input + output guards) and `HandleSystemSharedMemory` / `HandleCudaSharedMemory`
- `src/grpc/infer_handler.{cc,h}` — `InferGRPCToInput` and `InferAllocatorPayload`
- `src/grpc/grpc_server.cc` — the six `CommonHandler` shm control lambdas
- `src/common.h` — shared `kSharedMemoryManagerUnavailableErrorStr`
- `qa/L0_python_api/test_shared_memory_frontend.py` — regression test

#### Test plan:
- Reproduced on `nvcr.io/nvidia/tritonserver:26.03-py3` (v2.67.0): server healthy → one inference with `shared_memory_region` → `Fatal Python error: Segmentation fault`, process exit 139.
- New subprocess-isolated e2e test `qa/L0_python_api/test_shared_memory_frontend.py` (wired into `test.sh`): fails on the unpatched build (detects the crash), asserts a graceful 503 + live server on the patched build.
[...]

PR #8952, opened 2026-09-03
refactor: extract macro for gRPC shared-memory error reporting

akhilraj9:
#### What does the PR do?
Applies the review suggestion from PR #8936: the six gRPC shared-memory control
handlers in `src/grpc/grpc_server.cc` each repeated the same block —
create a `TRITONSERVER_Error`, report it via `GrpcStatusUtil::Create`, delete it,
and `return`. This extracts that into a single macro, `RESPOND_WITH_ERR_AND_RETURN`,
following the `do{…}while(false)` idiom of the existing `common.h` error macros.

- New `RESPOND_WITH_ERR_AND_RETURN(STATUS, ERR)` macro (1 definition).
- The 6 shared-memory-manager null guards collapse from a 6-line block to one call.
- Pure refactor — no behavior change; the macro expands to the identical code.

#### Checklist
- [ ] Added succinct git squash message before merging.

#### Commit Type:
- [x] refactor

#### Related PRs:
#8936 (introduced the shared-memory null guards; this addresses its review feedback).

- `src/grpc/grpc_server.cc` — the `RESPOND_WITH_ERR_AND_RETURN` macro definition (just below the includes) and the 6 call sites in the `CommonHandler::Register{System,Cuda}SharedMemory{Status,Register,Unregister}` handlers.

#### Test plan:
- Mechanical, behavior-preserving refactor: the macro expands to the exact code it replaces, so existing coverage applies (`L0_python_api` shared-memory-frontend test + the gRPC shared-memory control tests exercise these handlers).
- No new tests needed; CI build confirms it compiles.
- Pipeline ID: 66045430

#### Caveats:
Scope is limited to the exact pattern flagged in review — the 6 shared-memory handlers that create a fresh error and return early. The ~18 other `GrpcStatusUtil::Create + TRITONSERVER_ErrorDelete` sites in this file have no early `return` (they fall through at end-of-handler), so they'd need a separate no-return variant; intentionally left out to keep this change focused.

#### Background
Follow-up to review comment on PR #8936: "create a macro for this error reporting pattern."

Relates to TRI-1815.

### DIFF
--- a/src/grpc/grpc_server.cc
+++ b/src/grpc/grpc_server.cc
@@ -62,6 +62,15 @@
 #include "../tracer.h"
 #endif  // TRITON_ENABLE_TRACING
 
+// Reports a gRPC error to STATUS, releases it, and returns from the handler.
+#define RESPOND_WITH_ERR_AND_RETURN(STATUS, ERR) \
+  do {                                           \
+    TRITONSERVER_Error* err__ = (ERR);           \
+    GrpcStatusUtil::Create((STATUS), err__);     \
+    TRITONSERVER_ErrorDelete(err__);             \
+    return;                                      \
+  } while (false)
+
 namespace triton { namespace server { namespace grpc {
 
 namespace {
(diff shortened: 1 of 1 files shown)


src/grpc/grpc_server.cc
```diff
@@ -62,6 +62,15 @@
 #include "../tracer.h"
 #endif  // TRITON_ENABLE_TRACING
 
+// Reports a gRPC error to STATUS, releases it, and returns from the handler.
+#define RESPOND_WITH_ERR_AND_RETURN(STATUS, ERR) \
```
yinggeh, line 66:
```suggestion
#define RETURN_AND_RESPOND_WITH_ERR(STATUS, ERR) \
```

yinggeh: Macro name matches http_server.cc

yinggeh, line 66: Should define inside namespace `triton::server::grpc`

akhilraj9: Good point, renamed to RETURN_AND_RESPOND_WITH_ERR in 22de540.

akhilraj9: Done in 22de540. Moved the definition to just after the namespace opens,

yinggeh (approved):

PR #8815, opened 2026-06-03
perf: Add completion queue sharding

geobeau:
#### What does the PR do?
This PR adds a way to shard and configure more than one completion queue.
Per the commit message, see *background* for the motivation.

```    
Add --grpc-infer-cq-count to control the number of inference completion queues. Default is 1 (single shared CQ, same behavior as before).
 Set to 0 for one CQ per handler thread, or N>1 for N sharded CQs, to reduce contention at high throughput.
```

#### Checklist
- [ ] I have read the [Contribution guidelines](#../../CONTRIBUTING.md) and signed the [Contributor License
Agreement](https://github.com/NVIDIA/triton-inference-server/blob/master/Triton-CCLA-v1.pdf)
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ ] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [ ] I ran pre-commit locally (`pre-commit install, pre-commit run --all`)
- [ ] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [x] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Caveats:
Increasing the number of queues might make the load a bit unbalanced.

#### Background

When running above 100k QPS, GRPC threads will be the bottleneck to scale further we need to add more GRPC threads.
Unfortunately, adding more GRPC will have dimishing returns because there is a big contention on the futex required to access the completion queue. By sharding the completion queue into N part, we can reduce the contention drastically.

PR #8959, opened 2026-09-04
feat: add completion queue sharding for gRPC inference (unary and streaming)

akhilraj9:
#### What does the PR do?
Extends #8815 ("Add completion queue sharding for gRPC inference", TRI-1743)
from the unary `ModelInfer` path to the streaming `ModelStreamInfer` path.

That PR added `--grpc-infer-cq-count` and sharded the **unary** inference completion
queue, but left streaming on a single shared CQ. Under `max_inflight_requests`
backpressure with many concurrent decoupled streams, that one CQ/handler serializes
every stream — a laggard starves and is cancelled at the ~30s gRPC deadline, returning
0 responses (root cause of the `L0_simple_ensemble` flake, TRI-1711).

This PR shards the streaming CQ the same way, with **one handler per streaming CQ (1:1)**
so each stream stays single-writer (gRPC forbids concurrent writes on one stream).
No new flags — driven by the existing `--grpc-infer-cq-count` (`0` = one CQ per handler thread).

#### Checklist

#### Commit Type:
- [x] feat

#### Related PRs:
- #8815— TRI-1743, "Add completion queue sharding for gRPC inference" (this PR is stacked on it; set that branch as the base, not `main`).

- `src/grpc/grpc_server.cc` — the streaming CQ build loop and, most importantly, the
  handler-creation loop: **1 handler : 1 CQ using index `[i]`**, not the unary path's
  `[i % cq_count]`. This preserves the per-stream single-writer invariant and is the
  crux of the change.
- `src/grpc/grpc_server.h` — `model_stream_infer_cq_` (single) → `model_stream_infer_cqs_` (vector).
- `qa/L0_simple_ensemble/test.sh` — the `--grpc-infer-cq-count` validation + boundary-startup block.

#### Test plan:
Functional (streaming sharding):
- `test_concurrent_requests_with_different_limits` runs the server with
  `--grpc-infer-cq-count=0 --grpc-infer-thread-count=16` (16 concurrent streams over 16
  sharded CQs). All requests return their full response set with zero cancellations
  (previously flaked to 0/8).

Command-line value validation (server must fail to start):
- `--grpc-infer-cq-count=-1`  → log contains `Must be in the range 0 to 128`
- `--grpc-infer-cq-count=129` → log contains `Must be in the range 0 to 128`
- `--grpc-infer-cq-count=abc` → log contains `Invalid option value`

Boundary startup (server must start cleanly):
- `1` (legacy single-CQ behavior)
- `128` with `--grpc-infer-thread-count=128`, so it actually constructs 128 CQs/handlers
  and proves the per-CQ handler loop holds at max fan-out.

Diagnostics:
- The test asserts on gRPC errors before the response count, so a real failure surfaces
  its true status (e.g. `CANCELLED`) instead of the misleading "expected N, got 0".

e2e tests added: yes (all of the above in `qa/L0_simple_ensemble`).

- CI Pipeline ID: 66067932

#### Caveats:
- Stacked on the unmerged #8815; merge that first (or review only the commits
  above it). Base branch must be its branch, not `main`.
- The concurrent-streaming test is intentionally not run at `--grpc-infer-cq-count=1`
  (the single-CQ shape that reproduces the flake) to avoid reintroducing flakiness.
- Streaming shards strictly 1:1 (handler:CQ); unlike the unary path it cannot share a
  CQ across handlers, because a stream must be written by a single thread.

#### Background
`L0_simple_ensemble` began flaking after the server-side C++ gRPC bump (v1.54.3 → v1.81.1):
under backpressure + 16 concurrent streams, the single streaming CQ could no longer
service all streams fairly, starving one until it hit the gRPC deadline and was cancelled.
#<geobeau-PR> already introduced CQ sharding for unary; this applies the same fix to the
streaming path that was actually failing.

#### Related Issues:
- Fixes TRI-1711
- Relates to TRI-1743
- closes GitHub issue: #xxx

### DIFF
--- a/docs/customization_guide/inference_protocols.md
+++ b/docs/customization_guide/inference_protocols.md
@@ -145,6 +145,7 @@ For client-side documentation, see [Client-Side GRPC Status Codes](https://githu
 In general, using 2 threads per completion queue seems to give the best performance, see [gRPC Performance Best Practices](https://grpc.io/docs/guides/performance/#c). However, in cases where the performance bottleneck is at the request handling step (e.g. ensemble models), increasing the number of gRPC inference handler threads may lead to a higher throughput.
 
 * `--grpc-infer-thread-count`: 2 by default.
+* `--grpc-infer-cq-count`: 1 by default (a single shared completion queue). Set to `0` to create one completion queue per handler thread (capped at `--grpc-infer-thread-count`), which reduces contention when many requests or decoupled streams are processed concurrently. Valid range is 0-128.
 
 Note: More threads don't always mean better performance.
 
--- a/docs/customization_guide/inference_protocols.md
+++ b/docs/customization_guide/inference_protocols.md
@@ -145,7 +145,9 @@ For client-side documentation, see [Client-Side GRPC Status Codes](https://githu
 In general, using 2 threads per completion queue seems to give the best performance, see [gRPC Performance Best Practices](https://grpc.io/docs/guides/performance/#c). However, in cases where the performance bottleneck is at the request handling step (e.g. ensemble models), increasing the number of gRPC inference handler threads may lead to a higher throughput.
 
 * `--grpc-infer-thread-count`: 2 by default.
-* `--grpc-infer-cq-count`: 1 by default (a single shared completion queue). Set to `0` to create one completion queue per handler thread (capped at `--grpc-infer-thread-count`), which reduces contention when many requests or decoupled streams are processed concurrently. Valid range is 0-128.
+* `--grpc-infer-cq-count`: 1 by default (a single shared completion queue).
+  Set to `0` to create one completion queue per handler thread (capped at `--grpc-infer-thread-count`), which reduces contention when many requests or decoupled streams are processed concurrently.
+  Valid range is `0-128`.
 
 Note: More threads don't always mean better performance.
 
--- a/qa/L0_simple_ensemble/ensemble_backpressure_test.py
+++ b/qa/L0_simple_ensemble/ensemble_backpressure_test.py
@@ -30,18 +30,18 @@ import sys
 
 sys.path.append("../common")
 
-import os
-import queue
-import threading
-import time
-import unittest
-from contextlib import ExitStack
-from functools import partial
+import os  # noqa: E402
+import queue  # noqa: E402
+import threading  # noqa: E402
+import time  # noqa: E402
+import unittest  # noqa: E402
+from contextlib import ExitStack  # noqa: E402
+from functools import partial  # noqa: E402
 
-import numpy as np
-import test_util as tu
-import tritonclient.grpc as grpcclient
-from tritonclient.utils import InferenceServerException
+import numpy as np  # noqa: E402
+import test_util as tu  # noqa: E402
+import tritonclient.grpc as grpcclient  # noqa: E402
+from tritonclient.utils import InferenceServerException  # noqa: E402
 
 SERVER_URL = "localhost:8001"
 DEFAULT_RESPONSE_TIMEOUT = 60
@@ -77,6 +77,14 @@ def prepare_infer_args(input_value, enable_batching=False):
     infer_input[0].set_data_from_numpy(input_data)
     outputs = [grpcclient.InferRequestedOutput("OUT")]
     return infer_input, outputs
+
+
+def _describe_error(err):
+    """Concise, readable description of a tritonclient error for logs and asserts."""
+    try:
+        return f"{err.status()}: {err.message()}"
+    except Exception:
+        return str(err)
 
 
 def collect_responses(user_data, timeout=DEFAULT_RESPONSE_TIMEOUT):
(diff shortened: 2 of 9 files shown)


docs/customization_guide/inference_protocols.md
```diff
@@ -145,6 +145,7 @@ For client-side documentation, see [Client-Side GRPC Status Codes](https://githu
 In general, using 2 threads per completion queue seems to give the best performance, see [gRPC Performance Best Practices](https://grpc.io/docs/guides/performance/#c). However, in cases where the performance bottleneck is at the request handling step (e.g. ensemble models), increasing the number of gRPC inference handler threads may lead to a higher throughput.
 
 * `--grpc-infer-thread-count`: 2 by default.
+* `--grpc-infer-cq-count`: 1 by default (a single shared completion queue). Set to `0` to create one completion queue per handler thread (capped at `--grpc-infer-thread-count`), which reduces contention when many requests or decoupled streams are processed concurrently. Valid range is 0-128.
```
whoisj:
```suggestion
* `--grpc-infer-cq-count`: 1 by default (a single shared completion queue).
  Set to `0` to create one completion queue per handler thread (capped at `--grpc-infer-thread-count`), which reduces contention when many requests or decoupled streams are processed concurrently.
  Valid range is `0-128`.
```

src/grpc/grpc_server.cc
```diff
@@ -2445,8 +2445,20 @@ Server::Server(
   }
 
   common_cq_ = builder_.AddCompletionQueue();
-  model_infer_cq_ = builder_.AddCompletionQueue();
-  model_stream_infer_cq_ = builder_.AddCompletionQueue();
+
+  int cq_count =
+      (options.infer_cq_count_ > 0)
+          ? std::min(options.infer_cq_count_, options.infer_thread_count_)
```
mc-nv: std::min is used here but neither grpc_server.cc nor grpc_server.h explicitly includes <algorithm>. It currently compiles only because some other transitively-included header pulls it in on this toolchain — worth adding `#include <algorithm>` explicitly to avoid a build break if that transitive include ever changes.

src/command_line_parser.cc
```diff
@@ -1479,6 +1484,15 @@ TritonParser::Parse(int argc, char** argv)
                 "the range 2 to 128.");
           }
           break;
+        case OPTION_GRPC_INFER_CQ_COUNT:
+          lgrpc_options.infer_cq_count_ = ParseOption<int>(optarg);
+          if (lgrpc_options.infer_cq_count_ < 0 ||
+              lgrpc_options.infer_cq_count_ > 128) {
+            throw ParseException(
+                "invalid argument for --grpc_infer_cq_count. Must be in "
```
mc-nv: This error message uses underscores (`--grpc_infer_cq_count`) but the actual flag is hyphenated (`--grpc-infer-cq-count`). Most other option-validation messages in this file use the real hyphenated flag name — this appears to copy the pre-existing `--grpc_infer_thread_count` message, which has the same issue. Worth fixing to `--grpc-infer-cq-count` while touching this code (cosmetic; does not affect the L0 test, which only greps for the range substring).

qa/L0_simple_ensemble/ensemble_backpressure_test.py
```diff
@@ -30,18 +30,18 @@
...
-import time
-import unittest
-from contextlib import ExitStack
-from functools import partial
-
-import numpy as np
-import test_util as tu
-import tritonclient.grpc as grpcclient
-from tritonclient.utils import InferenceServerException
+import os  # noqa: E402
```
mc-nv: This PR bundles a `# noqa: E402` sweep across all pre-existing imports in this file, alongside the actual CQ-sharding test changes. Not wrong, but its incidental to the feature and mixes an unrelated lint-only cleanup into a `feat:` PR — worth splitting out per "one concern per PR" if there's a next revision.

src/grpc/grpc_server.cc
```diff
@@ -2547,6 +2561,8 @@ Server::GetOptions(Options& options, UnorderedMapType& options_map)
 
   RETURN_IF_ERR(GetValue(
       options_map, "infer_thread_count", &options.infer_thread_count_));
+  RETURN_IF_ERR(
+      GetValue(options_map, "infer_cq_count", &options.infer_cq_count_));
```
mc-nv: Flagging the open Greptile P1 thread here for visibility: `GetOptions` now unconditionally requires an `infer_cq_count` key in the options map. Within this PR's own commit it's safe — `_kservegrpc.py`'s `Options` dataclass always sets a default of `1`, so the key is always present when going through the current Python bindings. The risk is version skew: an older `tritonfrontend` wheel (built before this PR) paired with this newer `.so` would be missing the key and fail hard with `INVALID_ARG` instead of falling back to the documented default. That said, this matches an existing pre-PR pattern (`infer_thread_count` has the same unconditional-requirement behavior), so it's not a new class of bug introduced here — just worth a decision on whether to harden `GetValue` fallback behavior for new options going forward.

src/command_line_parser.cc
mc-nv: Isn't this value different tot he documentation `--grpc-infer-cq-count` /

akhilraj9: fixed, the message now says --grpc-infer-cq-count to match the flag/docs (3f1e0d6c)

src/grpc/grpc_server.cc
akhilraj9: Good catch — made infer_cq_count optional in GetOptions (dbfe394d): a missing key now falls back to the default 1 (single CQ) instead of INVALID_ARG, so an older tritonfrontend wheel still works against a newer .so

akhilraj9: happy to do a broader GetValue-with-default pass as a follow-up.

qa/L0_simple_ensemble/ensemble_backpressure_test.py
mc-nv, line 33: I'm confused why we need `# noqa: E402` to the `import os` ?

whoisj (changes requested):

whoisj (approved):
lgtm

qa/L0_simple_ensemble/ensemble_backpressure_test.py
mc-nv: This is not blocking comment but, looks like a hack for `pre-commit` rule.

Vinya567 (approved):
LGTM

PR #8975, opened 2026-09-15
test: expand torch_aoti coverage in L0_infer

Vinya567:
#### What does the PR do?
Expand L0_infer torch AOTI from one fp32 add model to int8, int32, fp16, and fp32. Share one helper (`ARGS[0]`/`ARGS[1]` → `RESULT`), copy all four model dirs on GPU (fail if a dir is missing), and import `tritonclient.http` only when not in Dynamo mode so Dynamo QA can collect the suite without gevent.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Related PRs:
- Follow-up to #8948 and #8951
- ai-dynamo/dynamo#14835 (fp16/bf16 tensor path; AOTI fp16 is not skipped under Dynamo)

`qa/L0_infer/infer_test.py` (`_infer_torch_aoti` and the four tests after `test_raw_ioi`), then `qa/L0_infer/test.sh` (`copy_torch_aoti_models` and expected counts 37/50, valgrind 40).

#### Test plan:
- Import check in a Dynamo image: `USE_HTTP=False`, four AOTI tests collected, no gevent import.
- Full GPU `L0_infer` against QA DATADIR: GitLab CI after this PR (count pins in a follow-up MR).

- CI Pipeline ID:

#### Caveats:
GPU-only model copy (packaged `.pt2` is CUDA). Tests `skipTest` if the model is not loaded; skips still count in `testsRun`. GitLab jobs that pin `EXPECTED_NUM_TESTS` need a +3 bump.

#### Background
#8948 added fp32 smoke coverage. Review asked for the remaining add-model dtypes already in the QA store, a Dynamo-safe HTTP import, and a hard fail if those model dirs are missing.

#### Related Issues: (use one of the action keywords Closes / Fixes / Relates to)
- Relates to #8948
- Relates to #8951


qa/L0_infer/infer_test.py
```diff
@@ -532,6 +536,110 @@ def test_raw_ioi(self):
             swap=False,
         )
 
+    def _infer_torch_aoti(self, model_name, input0, input1, expected, dtype_str):
+        # AOTI add models: ARGS[0]/ARGS[1] -> RESULT. Not libtorch-shaped.
+        def _is_ready_http():
```
whoisj, line 541:
```python
def _is_ready_http():

    with httpclient.InferenceServerClient(
        f"{TRITONSERVER_IPADDR}:8000"
    ) as client:
        return client.is_model_ready(model_name)
```
Seems like calling `_is_ready_http()` is going to cause problems. What am I missing?

qa/L0_infer/infer_test.py
```diff
@@ -61,8 +61,12 @@
 USE_GRPC = os.environ.get("USE_GRPC", 1) != "0"
 USE_HTTP = os.environ.get("USE_HTTP", 1) != "0"
 # The Dynamo frontend exposes only the KServe gRPC endpoint, not HTTP/REST.
+# Dynamo images ship gRPC-only tritonclient (no gevent). Match infer_util.
 if DYNAMO:
     USE_HTTP = False
+    httpclient = None
```
whoisj, line 67: so with `httpClient = None` what happens w/ `httpclient.InferenceServerClient`?

whoisj (commented):
Couple of questions:

qa/L0_infer/infer_test.py
Vinya567: `_is_ready_http `only runs when `USE_HTTP` is true. Dynamo sets `USE_HTTP=False`, so we never call it.

qa/L0_infer/infer_test.py
Vinya567: Dynamo never uses `httpclient.InferenceServerClient.` `USE_HTTP is False`, so we only use gRPC. Native Triton already imported httpclient in the else at the top.

qa/L0_infer/infer_test.py
whoisj: nowhere in `_is_ready_http` do I see an environment variable being checked.

Vinya567: right — `_is_ready_http` itself doesn’t check any env var. the caller does:

```python
        ready = False
        if USE_HTTP:                 # L554
            ready = _is_ready_http() # L555
        elif USE_GRPC:
            ready = _is_ready_grpc()

whoisj: fair. let's not do anything about this in the PR, but I do think this is a risky design overall. not your fault, you're just inheriting it.

whoisj (approved):

